PR #13, opened 2017-04-10
Introduce a concept guide for writing JSON and YAML based on a protobuf message

ZackButcher:
Unfortunately I had to hard-code our tables in HTML because I wanted multi-line code snippets in the table. Github markdown only supports a single line of content in a tabel's cell when the table is declared in markdown with pipes.

### DIFF
--- a/docs/concepts/writing-config.md
+++ b/docs/concepts/writing-config.md
@@ -36,7 +36,7 @@ message Metric {
     <td>
 <pre>
 {
-  "descriptor_name": "request_count",
+  "descriptorName": "request_count",
   "value": "1",
   "labels": {
       "source": "origin.ip",
@@ -61,7 +61,7 @@ message Metric {
   </tr>
   <tr>
     <td>
-<pre>
+<pre>`
 message Metric {
         string descriptor_name = 1;
         string value = 2;
@@ -78,14 +78,14 @@ message MetricsParams {
 {
   "metrics": [
     {
-      "descriptor_name": "request_count",
+      "descriptorName": "request_count",
       "value": "1",
       "labels": {
             "source": "origin.ip",
             "target": "target.service",
       },
     }, {
-      "descriptor_name": "request_latency",
+      "descriptorName": "request_latency",
       "value": "response.duration",
       "labels": {
             "source": "origin.ip",
@@ -100,7 +100,7 @@ message MetricsParams {
 </tbody>
 </table>
 
-*Note that we never name the outer-most message (the `MetricsParams`), but we do need the outer set of braces marking our JSON as an object.*
+*Note that we never name the outer-most message (the `MetricsParams`), but we do need the outer set of braces marking our JSON as an object. Also note that we shift from snake_case to lowerCamelCase; both are accepted but lowerCamelCase is idiomatic.*
 
 ### Proto `enum` fields uses the name of the value:
 
@@ -131,7 +131,7 @@ message AttributeDescriptor {
 <pre>
 {
   "name": "request.duration",
-  "value_type": "INT64",
+  "valueType": "INT64",
 }
 </pre>
       </td>
@@ -178,10 +178,10 @@ message MonitoredResourceDescriptor {
   "labels": [
     {
       "name": "label one",
-      "value_type": "STRING",
+      "valueType": "STRING",
     }, {
       "name": "second label",
-      "value_type": "DOUBLE",
+      "valueType": "DOUBLE",
     },
   ],
 }
@@ -223,18 +223,18 @@ message Metric {
     <td>
 <pre>
 {
-  "descriptor_name": "request_count",
+  "descriptorName": "request_count",
   "value": "1",
   "labels": {
       "source": "origin.ip",
-        "target": "target.service",
+      "target": "target.service",
   },
 }  
 </pre>
     </td>
     <td>
 <pre>
-descriptor_name: request_count
+descriptorName: request_count
 value: "1"
 labels:
   source: origin.ip
@@ -274,14 +274,14 @@ message MetricsParams {
 {
   "metrics": [
     {
-      "descriptor_name": "request_count",
+      "descriptorName": "request_count",
       "value": "1",
       "labels": {
             "source": "origin.ip",
             "target": "target.service",
       },
     }, {
-      "descriptor_name": "request_latency",
+      "descriptorName": "request_latency",
       "value": "response.duration",
       "labels": {
             "source": "origin.ip",
@@ -295,12 +295,12 @@ message MetricsParams {
     <td>
 <pre>
 metrics:
-- descriptor_name: request_count
+- descriptorName: request_count
   value: "1"
   labels:
     source: origin.ip
     target: target.service
-- descriptor_name: request_latency
+- descriptorName: request_latency
   value: response.duration
   labels:
     source: origin.ip
@@ -340,14 +340,14 @@ message AttributeDescriptor {
 <pre>
 {
   "name": "request.duration",
-  "value_type": "INT64",
+  "valueType": "INT64",
 }
 </pre>
       </td>
       <td>
 <pre>
 name: request.duration
-value_type: INT64
+valueType: INT64
 </pre>
 
 or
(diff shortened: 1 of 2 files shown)


docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
...
+</pre>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+
+## Translating to YAML
+
+There is no canonical mapping between protobufs and YAML; instead protobuf defines a canonical mapping to JSON, and YAML defines a canonical mapping to JSON. To ingest YAML as a proto we convert it to JSON then to  protobuf.
```
kyessenov: This is an implementation detail. YAML is a strict superset of JSON.

ZackButcher:
Updated to correct (lowerCamelCase) casing in the JSON/YAML examples.

We provide both because, as you say, YAML is a superset of JSON so a user _can_ provide us config as JSON. Further, the mixer's config API accepts both JSON and YAML payloads so both formats can be used. Finally, it helps provide more context for the transformations: one example may not be clear but seeing the same example in multiple related forms can help drive home the core concept.

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
+---
+title: Writing Istio Configuration
```
geeknoid: Make the title shorter so it tends to fix in a tab. The tab will always have Istio / in it, so no need to repeat Istio again.

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
...
+headline: Writing Istio Configuration
```
geeknoid: Nuke "Istio"

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
...
+sidenav: doc-side-concepts-nav.html
+bodyclass: docs
+layout: docs
+type: markdown
+---
+
+{% capture overview %}
+This document describes how to write configuration that conforms to Istio's schemas. All configuration schemas in Istio are defined as protobuf messages. When in doubt, search for the protos.
```
geeknoid: Link protobuf to https://developers.google.com/protocol-buffers/docs/proto3

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
...
+{% endcapture %}
+
+{% capture body %}
+## Translating to JSON
+The protobuf language guide defines [the canonical mapping between protobuf and JSON](https://developers.google.com/protocol-buffers/docs/proto3#json), refer to it as the source of truth. We provide a few specific examples that appear in the Mixer's configuration.
```
geeknoid: Suggestion: Please refer to [the canonical mapping between protobuf and JSON] as the authoritative source for representing protobufs as YAML.

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
...
+
+### Proto `map` and `message` translate to JSON objects:
```
geeknoid: How do these headers look in the righthand autogenerated table of contents? Seem maybe a bit long..

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
...
+</pre>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+*Notice we don't provide a `description`: fields with no value can be omitted.*
+
+### Proto `message` fields can contain other messages:
+<table>
```
geeknoid: Nit: insert blank line above.

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
...
+</pre>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+
+## Translating to YAML
+
+There is no canonical mapping between protobufs and YAML; instead protobuf defines a canonical mapping to JSON, and YAML defines a canonical mapping to JSON. To ingest YAML as a proto we convert it to JSON then to  protobuf.
```
geeknoid: Add links to where the two forms of canonical mappings are defined.

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,433 @@
...
+
+**Important things to note:**
+- YAML fields are implicitly strings
+- JSON arrays map to YAML lists; each element in a list is prefixed by a dash (`-`)
+- JSON objects map to a set of field names all at the same indentation level in YAML
+- YAML is whitespace sensitive and must use spaces; tabs are never allowed
+
+### Proto `map` and `message` fields:
```
geeknoid: Why do the tables below show the JSON translations again?

docs/concepts/writing-config.md
rshriram: Zack, you can fork the repo, then enable your personal github page (zackbutcher.github.io) via your GH settings. Then you enable GH pages in your fork of istio.github.io. Once you do these two, you can view the page via zackbutcher.github.io/istio.github.io/ 
(We should probably document this step. Not sure if we already did)

geeknoid:
Tasks and Tutorials are for things that have step-by-step instructions.
Reference will ultimately primarily be for auto-generated authoritative
reference material like API docs and CLI command-line options.

So to me, this does belong in Concepts. It just doesn't belong at the top
of the table of contents. When we start filling in more docs, we can nest
this doc deeper, under the overall Config section.

How's that sound?

rshriram:
So the way I was viewing this was as a Task, with title, "Writing mixer configuration files" [I guess its a grey area here. I can go either way between concept and Task]. But I do agree that this does not belong in the ToC for concepts. It should be nested under the mixer.

ZackButcher:
This is more general than just mixer, though: proxy config is also yaml-conforming-to-proto-schema, as is manager config, right? I attempted to pull out anything mixer specific (with the exception that the examples use mixer protos).

docs/concepts/writing-config.md
ZackButcher: I thought it'd be more helpful to have everything side-by-side to make clear both the mapping from proto to yaml but also json to yaml. I can remove the JSON here but I thought it added value.

docs/concepts/writing-config.md
ZackButcher: Already have, although my site is set up to use the master branch for zackbutcher.github.io/istio.github.io/; I'll go update it to use this branch so we can see the nicely rendered page.

docs/concepts/writing-config.md
geeknoid: Feels redundant, and it doesn't fit on my screen without side scrolling... I'd nuke the redundant column.

frankbu:
I think the routing rule config doc (https://istio.io/docs/reference/rule-dsl.html) needs to tie into this one. Maybe in the first paragraph where it says that it's base on a proto 3 schema, it can point to this document for details on the mapping. Notice, however, that the routing language is described in YAML (not JSON), so I tend to agree with @kyessenov that it might be better for this doc to leave out the JSON examples and focus on YAML instead.

The reverse linkage may also be good, that is, this doc could include a link to https://istio.io/docs/reference/rule-dsl.html as an example of where the mapping is used.

ZackButcher:
PTAL

docs/concepts/writing-config.md
```diff
@@ -0,0 +1,194 @@
...
+## Translating to YAML
+There is no canonical mapping between protobufs and YAML; instead [protobuf defines a canonical mapping to JSON](https://developers.google.com/protocol-buffers/docs/proto3#json), and [YAML defines a canonical mapping to JSON](http://yaml.org/spec/1.2/spec.html#id2759572). To ingest YAML as a proto we convert it to JSON then to  protobuf.
+
+**Important things to note:**
+- YAML fields are implicitly strings
+- Proto `repeated` fields map to YAML lists; each element in a YAML list is prefixed by a dash (`-`)
+- Proto `message`s map to JSON objects; in YAML objects are field names all at the same indentation level
+- YAML is whitespace sensitive and must use spaces; tabs are never allowed
+
+### Proto `map` and `message` fields:
```
geeknoid: Can you change these headers? They don't look good in the right-hand side table of contents. I think you can generally remove the word "Proto" on each of these, and drop the : at the end of the line.

PR #21, opened 2017-04-14
Call out special cases in the proto-JSON mapping from protobuf spec

ZackButcher:
Specifically call out types used in our configs that do not conform to the standard protobuf-json object mapping. In particular, the specification declares that `Duration` and `Timestamp`s are serialized very differently than their protobuf structure would imply.


ZackButcher:
We should also go update the examples to remove `LabelDescriptor` now that we've gotten rid of it in our real config, I suppose.

PR #22, opened 2017-04-16
Changes related to new logo

jasminejaksic:
Updated the logo with the new image, edited size, padding, etc and made other minor changes


jasminejaksic:
Before and after screenshot
screen shot 2017-04-16 at 12 56 23 am

_includes/hero.html
```diff
@@ -1,6 +1,6 @@
 <div class="hero">
   <div class="container">
-    <img class="logo" src="img/logo.jpg" alt="istio">
+    <img class="logo" src="{{ site.baseurl }}/img/logo.png" alt="istio">
```
geeknoid: istio -> Istio

geeknoid:
Note, we also need to replace all the imagery in the favicons directory. This is what gets displayed as icons in tabs and other similar places.

rshriram:
@jasminejaksic I have uploaded favicons

jasminejaksic:
@rshriram Should we use the blue version of the icon, since white is a bit difficult to see?

PR #53, opened 2017-04-24
egress envoy documentation

GregHanson:


_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
+---
+title: Accessing a Service with Egress
```
rshriram: Egress Envoy

rshriram: Change everywhere. Istio proxy or Egress proxy to Egress Envoy

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+layout: docs
+type: markdown
+---
+
+
+This task describes how to configure Istio to expose an external service to a Kubernetes cluster. You'll learn how 
+to create an Egress proxy, define an external service and make requests to the service from within the cluster.
+
+
+## Doing ...
```
rshriram: delete

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+
+This task assumes you have deployed Istio on Kubernetes.  If you have not done so, please first complete
+the [Installation Steps]({{site.bareurl}}/docs/tasks/istio-installation.html).
+
+This task also assumes you have a publicly accessible service to call from within the cluster 
+(or [httpbin.org](http://httpbin.org) can be used as an example). 
+
+### Setup the environment
+
+Create the Istio Egress proxy.
```
rshriram: Create the Egress Envoy using the following YAML snippet

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+metadata:
+  name: httpsgoogle
+spec:
+  type: ExternalName
+  externalName: www.google.com
+  ports:
+  - port: 443
+    name: https
+```
+
```
rshriram: May be mention that all access to external name services must be over http? e.g., http://httpsgoogle/foo ?

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+Deploy your app(s) using the [istioctl kube-inject]({{site.bareurl}}/docs/reference/istioctl.html#kube-inject) command.
+You can use your own app, or try one of the example apps from [demos](https://github.com/istio/istio/tree/master/demos) 
+directory. Each app directory contains an associated README.md providing more details.
+
+```bash
+kubectl apply -f <(istioctl kube-inject -f {resource.yaml})
+```
+
```
rshriram: I think the above bit (deploy your apps..) is not needed, as it has already been covered elsewhere.

rshriram: We should probably have a snippet for a debian/ubuntu image that will have both bash and curl. Then we could use this as a simple client for people to try accessing destination services.

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: istio-egress
+spec:
+  ports:
+  - port: 80
+    name: "80"
```
kyessenov: remove this `name` field. I don't think it's needed.

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+        args: ["proxy", "egress", "-v", "2"]
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: metadata.namespace
+```
+
+Create the external service definition for you external service or use one of the samples below.  The `metadata.name`
```
kyessenov: "for your"

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: metadata.namespace
+```
+
+Create the external service definition for you external service or use one of the samples below.  The `metadata.name` 
+field is the url your internal apps will use when calling the external service.  The `spec.ExternalName` should be the
```
kyessenov: `spec.externalName`

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: httpbin
+spec:
+  type: ExternalName
+  externalName: httpbin.org
+  ports:
+  - port: 80
+    name: http
```
kyessenov: can also drop the port name. don't think it makes a difference and may confuse.

_docs/tasks/egress.md
kyessenov: yes, definitely.

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: httpsgoogle
+spec:
+  type: ExternalName
+  externalName: www.google.com
+  ports:
+  - port: 443
+    name: https
```
kyessenov: remove this `https` as well.

_docs/tasks/egress.md
kyessenov: can we use the debug base image that Jason created? it has curl and wrk which is all we really need.

_docs/tasks/egress.md
```diff
@@ -0,0 +1,131 @@
...
+
+```bash
+$ kubectl exec -it {APP_POD_NAME} curl httpbin/headers
+.. response ..
+```
+
+For external services of type HTTPS, the port must be specified in the request:
+
+```bash
+$ kubectl exec -it {APP_POD_NAME} curl httpsgoogle:443
```
kyessenov: add "http://" prefix to make it extra clear. The port is technically there with "https" but due to the change of the protocol, it must be explicit here.

_docs/tasks/egress.md
rshriram: Yep, that would work well. All we need to do is to add a lua script for wrk to send load to the bookinfo service at a rate of 10 qps (greater than the rate limit).

_docs/tasks/egress.md
rshriram: I would also rename the httpsgoogle to something else, such as securegoogle.

_docs/tasks/egress.md
GregHanson: @rshriram @kyessenov I was wondering about this, should it be part of the `istio.yaml` in istio/istio repo?  Then it would be deployed as part of the istio-installation here: https://github.com/istio/istio.github.io/blob/master/_docs/tasks/istio-installation.md

Or should the egress proxy not be created by default like ingress?

_docs/tasks/egress.md
GregHanson: @rshriram do you foresee the "deploy your apps" step happening [here](https://github.com/istio/istio.github.io/blob/master/_docs/tasks/integrating-services-into-istio.md)?  If so, I should link to it in the `Before you begin` section

GregHanson: @kyessenov what is the debug base image, [this one](https://hub.docker.com/r/istio/app_debug/tags/) ?

ayj: yes, app_debug has these sorts of packages pre-installed.

GregHanson: @ayj is there a `debug.yaml` defined someplace I should use or should I just define it inline in the task?

ayj: Not that I'm aware. You can use istio.io/demos/apps/echo/logic-app.yaml as a reference. Be sure to name the service port with `http` prefix if you define it inline otherwise we'll only proxy at L4.

PR #65, opened 2017-04-25
Generated docs for our protos

ZackButcher:
Unfortunately this did require some manual touch-up, but nothing that'll
be impossible to automate I think.

See live examples of all the pages:
https://zackbutcher.github.io/istio.github.io/docs/reference/api/mixer-service.html
https://zackbutcher.github.io/istio.github.io/docs/reference/api/mixer-config.html
https://zackbutcher.github.io/istio.github.io/docs/reference/api/enovy-config.html

I still need to grab the mixer aspect config protos, which is a bit harder because they live in the mixer repo. @geeknoid do we want to generate docs for adapter config protos as well?

### DIFF
--- a/_docs/reference/api/enovy-config.md
+++ b/_docs/reference/api/enovy-config.md
@@ -1,0 +1,908 @@
+---
+category: Reference
+title: Envoy Configuration Schema
+overview: Generated documentation for Envoy's Configuration Schema
+              
+parent: API and Configuration Schemas
+order: 60
+
+bodyclass: docs
+layout: docs
+type: markdown
+---
+<a name="rpc_istio.proxy.v1alpha.config"></a>
+## Package istio.proxy.v1alpha.config
+
+<a name="rpc_istio.proxy.v1alpha.config_index"></a>
+### Index
+
+* [CircuitBreaker](#istio.proxy.v1alpha.config.CircuitBreaker)
+(message)
+* [CircuitBreaker.SimpleCircuitBreakerPolicy](#istio.proxy.v1alpha.config.CircuitBreaker.SimpleCircuitBreakerPolicy)
+(message)
+* [DestinationPolicy](#istio.proxy.v1alpha.config.DestinationPolicy)
+(message)
+* [DestinationWeight](#istio.proxy.v1alpha.config.DestinationWeight)
+(message)
+* [HTTPFaultInjection](#istio.proxy.v1alpha.config.HTTPFaultInjection)
+(message)
+* [HTTPFaultInjection.Abort](#istio.proxy.v1alpha.config.HTTPFaultInjection.Abort)
+(message)
+* [HTTPFaultInjection.Delay](#istio.proxy.v1alpha.config.HTTPFaultInjection.Delay)
+(message)
+* [HTTPRetry](#istio.proxy.v1alpha.config.HTTPRetry)
+(message)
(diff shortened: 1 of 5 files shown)


kyessenov:
Looks pretty good. I wonder if we can use YAML names instead of proto identifiers.

ZackButcher:
Also, right nav floating over content when it's too wide is a known issue, right? It's pretty bad on these pages.

@kyessenov it would be nice to have camelCase names...I'll see about filing a feature request for it. We _might_ be able to do it in post processing (we already have to do a bit unfortunately) but I'm not sure we can just blindly replace all `a_b` with `aB`.

ZackButcher:
Looks like in a few spots (mixer's cfg.proto and metric_descriptor.proto) we have inline YAML snippets as examples, but don't mark them as code in markdown syntax. I'll write up a PR to istio/api to change those.

_docs/reference/api/enovy-config.md
```diff
@@ -0,0 +1,908 @@
+---
```
geeknoid: The file name has a typo...

frankbu: This isn't really envoy config anyway, it's the istio proxy config (influenced by envoy).
So `proxy-config.md` would be a better name.

frankbu:
I notice in the envoy config file there a few "/" characters in the generated output. For example: "/L7 fault injection policy applies to Http traffic".

Also should things marked @exclude be left out? "/@exclude Faults can be injected into ..."

ZackButcher:
`@exclude` directives were used by one of the other open source doc generation tools, as are the extra `/`'s in the proxy config. They should all be removed. The syntax for an excluded comment is `(-- --)`, inside of a normal proto comment, and they can span multiple lines, e.g.:

```proto
// A comment that will appear in the docs
// (-- This comment will not appear
// nor will this line
// nor this one --)
// But this one will.
```

@frankbu w.r.t. naming: is this config truly independent of Envoy? If it is, then we should call it `Proxy Config`, but if it's not we should keep it as `Envoy Config` (I believe that most closely follows the doc guidelines we laid out).

ZackButcher:
@rshriram et al are planning on another rev of the proxy config before alpha; I'm going to put this PR on hold till that's complete. Processing the proxy config file as-is today is too much manual work to do repeatedly (removing the `/`'s, the `@exclude` stuff, etc.).

geeknoid:
Can you just check in as-is? You can then update it later. In the meantime,
it'll put something in the doc set.

frankbu:
I agree with @geeknoid that you should check it in as-is, so we have a placeholder doc that we can reference it from other docs. @ZackButcher: I believe the proxy config is intended to be proxy independent, even though it aligns directly with envoy config in many places. You might want to confirm with @rshriram. However, the main point is that this config is used to generate the actual envoy config file, which is different, so calling this "Envoy Config" would be confusing.

ZackButcher:
SG - I chatted with @geeknoid out of band and decided we'll check in mixer stuff (after I finish my PR updating really out of date comments), but leave out proxy for now since it requires too much manual processing. I'll include proxy as soon as it's updated.

If we're fine with the partially broken proxy docs I'm more than happy to check them in, though.

frankbu:
@ZackButcher : I'm OK with broken proxy doc, so I can reference it from other docs I'm working on. Have we agreed to name it `proxy-config.html`?

ZackButcher:
Yup, I'll rename to proxy-config and check in all three.

PR #67, opened 2017-04-26
Auto-generate reference documentation for istioctl. 

ayj:
Auto-generate istioctl reference documentation. This depends on https://github.com/istio/manager/pull/597 for raw content generation plus an additional script to prepend the necessary template headers and fix up links. The new auto-generated files are under `_docs/reference/istioctl-autogen/` to avoid conflict with the existing `_docs/reference/istioctl.md` file. `_docs/reference/istioctl.md` will be removed and `_docs/reference/istioctl-autogen/` renamed once content is copied over from istioctl.md to the `istio.io/manager/cmd/istioctl/*.go`. https://github.com/istio/istio.github.io/issues/70 tracks moving the content.

Usage:

```bash
# Use istioctl that is in PATH
./scripts/auto-generate-istioctl.sh

# Use a specific version of istioctl, e.g. private build from sibling manager repo.
ISTIOCTL=../manager/bazel-bin/cmd/istioctl/istioctl ./scripts/auto-generate-istioctl.sh
```


_docs/reference/istioctl-autogen/istioctl.md
```diff
@@ -0,0 +1,37 @@
+---
+category: Reference
+title: Istioctl
+bodyclass: docs
```
geeknoid: Need to output the overview: and index: fields too.

_docs/reference/istioctl-autogen/istioctl.md
```diff
@@ -0,0 +1,37 @@
...
+```
+  -c, --kubeconfig string                Use a Kubernetes configuration file instead of in-cluster configuration
+      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
+  -n, --namespace string                 Select a Kubernetes namespace (default "default")
+  -v, --v Level                          log level for V logs
+      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
+```
+
+### SEE ALSO
+* [istioctl collateral](istioctl_collateral.html)	 - Generate istioctl collateral files
```
geeknoid: We should mark this collateral option as a hidden option in the cobra stuff which I think will prevent it from being listed when dumping out this material. This is not something we need people to use or care about, it's just pollution within these docs IMO.

_docs/reference/istioctl-autogen/istioctl.md
geeknoid: Is there an issue tracking moving the content over from the .md to the .go file?

ayj: Added `overview` with the same contents as `title`. I used _docs/reference/contribute/ and didn't see `index` in any of that markdown. 

Created https://github.com/istio/istio.github.io/issues/70 to track moving the content.

ayj:
Thanks Martin. I don't have write access to istio.github.io so someone else will need to merge.

PR #75, opened 2017-04-28
Avoid serving javascript and favicons over http 

ayj:
Commit 41145599 changed the base path for referencing resources based
on site.github.environment. As a result, most of the javascript
resources and some images were served over http instead of https which
some browsers won't load (e.g. Chrome). Fix the problem by stripping
the protocol prefix from site.github.url such that content
is always served over the original protocol, i.e. https. Fix taken 
from https://github.com/github/pages-gem/issues/238#issuecomment-206964532.

Note:  The value of `home` various based on whether we're using istio.io or a staging site. This means I couldn't test 100% that this fixes the problem. However, I was able to test the reverse by replacing `https` with `http` in my staging site and verifying my staging site starting serving javascript/images over http.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -2,9 +2,9 @@
 layout: default
 ---
 {% assign home = "" %}
-{% if site.github.environment == "dotcom" %}
-    {% assign home = site.github.url %}
-{% endif %}
+<!-- {% if site.github.environment == "dotcom" %} -->
+    {% assign home = site.github.url | replace: 'http://', //' %}
+<!-- {% endif %} -->
 
 <!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


_layouts/base.html
```diff
@@ -2,9 +2,9 @@
 layout: default
 ---
 {% assign home = "" %}
-{% if site.github.environment == "dotcom" %}
-    {% assign home = site.github.url %}
-{% endif %}
+<!-- {% if site.github.environment == "dotcom" %} -->
```
geeknoid: I believe if you comment this out, then the site will not work when served from a local Jekyll running on your machine. That's why I needed the conditional.

geeknoid:
I incorporated this change in my latest commit.

PR #80, opened 2017-04-28
traffic management doc improvements

frankbu:


_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -148,6 +148,96 @@ route:
...
+```yaml
+destination: "ratings.default.svc.cluster.local"
+route:
+- tags:
+    version: v1
+httpReqTimeout:
+  simpleTimeout:
+    timeoutSeconds: 10
+```
+
```
rshriram: can you also mention this: This value can be overridden on a per-request basis if the application adds 
x-envoy-upstream-rq-timeout-ms header on outbound requests. The timeout must be specified in millisecond units.

_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -148,6 +148,96 @@ route:
...
+```yaml
+destination: "ratings.default.svc.cluster.local"
+route:
+- tags:
+    version: v1
+httpReqRetries:
+  simpleRetry:
+    attempts: 3
+```
+
```
rshriram: Also mention:
The maximum number of retries can also be controlled on a per-request basis by providing the x-envoy-max-retries. A retry policy should be in place in order for this header field to be honored. 
In addition, setting x-envoy-upstream-rq-per-try-timeout-ms header (value in milliseconds) on outbound will cause Envoy to set a per try timeout. This allows a caller to set a tight per try timeout to allow for retries while maintaining a reasonable overall timeout. The per-try timeout specified via HTTP header must be <= the global route timeout or it will be ignored.

_docs/tasks/request-timeouts.md
```diff
@@ -0,0 +1,120 @@
...
+the message: Sorry, product reviews are currently unavailable for this book.
+This was the result of it recieving the timeout error from the `reviews` service.
+
+If you check out the [fault injection task](./fault-injection.html), you'll find out that the `productpage`
+microservice also has its own application-level timeout (3 seconds) for calls to the `reviews` microservice.
+Notice that in this task we used an Istio route rule to set the timeout to 1 second.
+Had we instead set the timeout to something greater than 3 seconds (e.g., 4 seconds) the timeout
+would have had no effect since the more restrictive of the two will take precedence.
+More details can be found [here](/docs/concepts/traffic-management/handling-failures.html#faq).
+
```
rshriram: May be we could move the text about per-request timeout (using http headers) into this page? like Additional information or whatever.. Same applies to retries.

_docs/concepts/traffic-management/rules-configuration.md
frankbu: mentioned it and added a link to the reference doc

_docs/concepts/traffic-management/rules-configuration.md
frankbu: I think these details should go in the reference doc which I've linked to

PR #86, opened 2017-05-02
Fix index mis-alignment

wattli:
Currently the index is mis-aligned https://istio.io/docs/concepts/network-and-auth/auth.html


wattli:
@geeknoid , please take a look and this is how it looks now https://screenshot.googleplex.com/McCP6hBRDd6

PR #89, opened 2017-05-02
Update integrating-services-into-istio.md

ayj:


_docs/tasks/integrating-services-into-istio.md
```diff
@@ -1,36 +1,207 @@
...
 order: 20
 
 layout: docs
 type: markdown
 ---
 
-This task shows how to do X in a Kubernetes cluster. You'll learn
-how to ...
-
+This task shows how to integrate applications with Istio into a
```
rshriram: applications running on kubernetes with istio.

_docs/tasks/integrating-services-into-istio.md
```diff
@@ -1,36 +1,207 @@
...
+Kubernetes cluster. You'll learn how to manually inject the Envoy side
```
rshriram: del manually. Typo: Envoy side. Should be Envoy sidecar

_docs/tasks/integrating-services-into-istio.md
```diff
@@ -1,36 +1,207 @@
...
+into deployments using [istioctl
+kube-inject]({{site.bareurl}}/docs/reference/istioctl/istioctl_kube-inject.html)
 
 ## Before you begin
-* Do this.
-* Do this too.
 
-## Doing ...
+This task assumes you have deployed Istio on Kubernetes.  If you have
```
rshriram: Deployed istio control plane

_docs/tasks/integrating-services-into-istio.md
```diff
@@ -1,36 +1,207 @@
...
+not done so, please first complete the [Installation
+Steps]({{site.bareurl}}/docs/tasks/istio-installation.html).
+
+## Injecting Envoy sidecar into a deployment
 
-1. Do this.
-1. Do this next. Possibly read this [related explanation](...).
+Example deployment and service to demonstrate this task. Save this as
+echo.yaml.
```
rshriram: Okay, now we need to decide on what we have to do with the apps. All tasks use bookinfo. This one uses something simple and server side. It doesn't convince me that istio didn't break anything. 

May be do a container list per pod?
May be have our debug app from manager be used as client and our app from manager test be used as a server?

_docs/tasks/integrating-services-into-istio.md
ayj: FWIW, this wording was taken from _docs/task/{egress,ingress}.md.

_docs/tasks/integrating-services-into-istio.md
ayj: > This one uses something simple and server side.

I added a separate deployment for the client instead of reusing echo. I also added a reference to the bookinfo sample under "what's next". I think using bookinfo itself would complicate this task since it has 6 deployments and 5 services.

> May be do a container list per pod?

I already have "Use `kubectl get deployment echo -o yaml` to inspect the modified
deployment and look for the following"  with additional notes of what the user should look for.

> May be have our debug app from manager be used as client and our app from manager test be used as a server?

We know there is no istio-specific logic in our debug apps but the user might not. To make it obvious that injection doesn't depend on apps being istio-aware I thought it better to use  well known non-istio specific images.

_docs/tasks/integrating-services-into-istio.md
rshriram: Yikes. I think @frankbu is splitting up those docs and cleaning them up.

_docs/tasks/integrating-services-into-istio.md
```diff
@@ -1,36 +1,224 @@
...
+The Envoy sidecar does _not_ intercept container-to-container traffic
+within the same pod when traffic is routed via localhost. This is by
+design.
 
+```bash
+$ kubectl exec -it ${SERVER} -c echo -- curl localhost:8080 | grep x-request-id
+```
+
+## Understanding what happened
+
```
rshriram: Are we following this title convention everywhere? Can we retitle this as "Under the hood"?

_docs/tasks/integrating-services-into-istio.md
```diff
@@ -1,36 +1,224 @@
...
+admission controller. Use `kubectl get deployment echo -o yaml` to
+inspect the modified deployment and look for the following:
+
+* A proxy container which includes the Envoy proxy and agent to manage
+  local proxy configuration.
+
+* An [init-container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/)
+  to program [iptables](https://en.wikipedia.org/wiki/Iptables).
+
+The proxy container includes the Envoy proxy and corresponding agent
```
rshriram: This line is repeated.

_docs/tasks/integrating-services-into-istio.md
```diff
@@ -1,36 +1,224 @@
...
+    "securityContext":{
+      "privileged":true
+    }
+  }
+]'
+```
+
+The `enable-core-dump` init-container is also added to enable GDB core
+dumps for the Envoy proxy. This can be disabled by passing
+`--coreDump=False` to `istioctl kube-inject`.
```
rshriram: We should remove the core dump part. This is not very comforting to an end user 😉

_docs/tasks/integrating-services-into-istio.md
ayj: This wording is used in request-routing.md, request-timeouts.md, and fault-injection.md

ayj:
Thanks for the review @rshriram. Can you merge once you're satisfied with the changes?

_docs/tasks/integrating-services-into-istio.md
rshriram: You missed this one..

_docs/tasks/integrating-services-into-istio.md
rshriram: and the core dump part.

ayj:
Sorry, pushed the last set of changes to the wrong repo. Fixed now. PTAL.

PR #99, opened 2017-05-02
Automate generation of mixer CLI collateral docs

ZackButcher:
Big difference between this and istioctl generation is that we mung all the docs for subcommands into a single markdown file, rather than generate a ton of files. @ayj we should probably do the same with istioctl too.


ZackButcher:
For istioctl at least sub-subcommands can be grouped together, even if we want separate docs per command (since unlike the mixer CLIs, istioctl is a single binary).

PR #103, opened 2017-05-03
Add Understanding rate limits section

mandarjog:


_docs/tasks/rate-limiting.md
```diff
@@ -101,10 +101,58 @@ The following configuration applies a `5qps` rate limit only to version `v3` of
...
+
+Consider the following example
+```yaml
+descriptorName: RequestCount
+maxAmount: 5000
+expiration: 5s
+labels:
+  label1: target.service
+```
+This defines a counter with a limit of `5000` that resets every `5 seconds`.
```
geeknoid: This is not the right description, here and elsewhere.

We use a rolling window algorithm, not a fixed timer reset algorithm. So the maxAmount and expiration indicate the total number of requests allowed in any specific time interval.

_docs/tasks/rate-limiting.md
```diff
@@ -101,10 +101,58 @@ The following configuration applies a `5qps` rate limit only to version `v3` of
...
+of the configuration. Here we introduce the notion of labels that enable creation of more granular counter keys.
+When a request arrives at Mixer with `target.service=ratings` it forms the following counter key.
+
+``` RequestCount;maxAmount=5000;expiration=5s;label1=ratings ```
+
+Using `target.service` in the counter key enables independent rate limits for every service. 
+In absence of `target.service` as part of the key, the same counter location is used by all services resulting in 
+combined rate limit of `5000` requests per `5 seconds`
+
+Mixer supports arbitrary number of labels by defining `QuotaDescriptors`.
```
geeknoid: an arbitrary

_docs/tasks/rate-limiting.md
```diff
@@ -101,10 +101,58 @@ The following configuration applies a `5qps` rate limit only to version `v3` of
...
+```yaml
+name: RequestCount
+rate_limit: true
+labels:
+   label1: 1 # STRING
+```
+Here we define `RequestCount` Quota descriptor that takes 1 string label. We recommend using meaningful label names
```
geeknoid: define the
Quota -> quota

_docs/tasks/rate-limiting.md
mandarjog: This is not a description of the MemQuota adapter specifically. Are we requiring that all quota adapters support rolling window?

geeknoid: This will be the preferred approach I believe, since it avoids the '11th hour' problem where actual QPS can be twice as high for a given time interval than what is stated.

mandarjog:
addressed comments. PTAL.

PR #106, opened 2017-05-03
Move istioctl, mixc, and mixs into a Command subcategory.

geeknoid:


ZackButcher:
FYI this will make the two shell scripts unusable. Both will need to be updated to use the new directory, to not delete everything in the target directory, and to not produce an index.md file.

geeknoid:
Does that mean we should actually just have a single script then?

ZackButcher:
Yeah, we should merge them. Since we're going to combine istioctl subcommands together info a single md file too, all the logic can be shared anyway. There just need to be some minor changes made to the mixer shell script.

PR #113, opened 2017-05-04
Swagger doc first pass

liamawhite:
Spec is currently just an object as it can be either a route-rule or destination-policy spec schema; which I don't think we can completely document.

Signed-off-by: Liam White <liamwhite@uk.ibm.com>


liamawhite:
@rshriram not sure whats happening with my CLA?

PR #115, opened 2017-05-05
Egress task

frankbu:
Here is an enhanced version of the egress task.
I didn't change the actual example, just the description.
However, when I tried it, it didn't actually work for me (got 404 back from curl).
Maybe I didn't have something setup right, @GregHanson?
@kyessenov : let me know if you want me to change anything. Otherwise, if you want to
just approve/merge it, and then take another pass at improving it, that's fine with me.

### DIFF
--- a/_docs/tasks/egress.md
+++ b/_docs/tasks/egress.md
@@ -35,7 +35,7 @@ from within your Istio cluster. In this task we will use
 1. Register an external HTTP service:
 
    ```bash
-   cat <<EOF | kubectl create -f -
+   $ cat <<EOF | kubectl create -f -
    apiVersion: v1
    kind: Service
    metadata:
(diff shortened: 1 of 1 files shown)


kyessenov:
Frank, it is not working since istio-install does not have egress proxy for some reason. Let me push a change to installation YAML to add egress proxy.

kyessenov:
Ok, I take my words back, the egress proxy should be set-up by default. I'm not sure who maintains istio installation YAMLs but we need to synchronize the new YAMLs in istio/manager with what we have in the website. The smoketest does not exercise a lot of features, so I trust what we have in istio/manager a lot more than istio/istio.

@andraxylia can you please synchronize the manager YAML templates with installation templates? I see that auth is not enabled on ingress/egress and some ports are missing from a brief look.

kyessenov:
I must say I'm lost in istio-install. There are over 20 YAML files there and I don't know where to begin making changes since some of them are generated...

frankbu:
Merging now. We can iterate further in another PR

PR #116, opened 2017-05-05
Avoid - in attribute name for user agent.

wora:


_docs/reference/api/attribute-vocabulary.md
```diff
@@ -39,16 +39,16 @@ deployments will have agents (Envoy or Mixer adapters) that produce these attrib
 | target.labels | map | A map of key-value pairs attached to the target. | |
 | target.user | string | The user running the target application. | service-account |
 | request.headers | map | A map of HTTP headers attached to the request. | |
-| request.id | string | A unique ID for the request, which can be propagated to downstream systems. This should be a guid or a psuedo-guid with a low probability of collision in a temporal window measured in days or weeks. | |
+| request.id | string | A unique ID for the request, which should be propagated to downstream systems. It should have a low probability of collision within a service in a temporal window measured in days. | |
 | request.path | string | The HTTP URL path including query string | |
 | request.host | string | The HTTP Host header. | |
 | request.method | string | The HTTP method. | |
-| request.reason | string | The system parameter for auditing reason. It is required for cloud audit logging and GIN logging | |
+| request.reason | string | The request reason used for auditing system. | |
```
ZackButcher: auditing system -> auditing system**s**

_docs/reference/api/attribute-vocabulary.md
```diff
@@ -39,16 +39,16 @@ deployments will have agents (Envoy or Mixer adapters) that produce these attrib
 | target.labels | map | A map of key-value pairs attached to the target. | |
 | target.user | string | The user running the target application. | service-account |
 | request.headers | map | A map of HTTP headers attached to the request. | |
-| request.id | string | A unique ID for the request, which can be propagated to downstream systems. This should be a guid or a psuedo-guid with a low probability of collision in a temporal window measured in days or weeks. | |
+| request.id | string | A unique ID for the request, which should be propagated to downstream systems. It should have a low probability of collision within a service in a temporal window measured in days. | |
```
ZackButcher: Not directly related to your PR, but I don't think we've defined the meaning of should/must/may etc. in the context of this doc (and in the context of Istio docs more generally). We should note what we mean by these words.

_docs/reference/api/attribute-vocabulary.md
ZackButcher: nit: and maybe it should read "used by auditing systems" instead of "used for"; not as sure about that.

_docs/reference/api/attribute-vocabulary.md
kyessenov: I find it strange to qualify the probability by "days". No proxy can guarantee that for unrestricted amount of traffic per day :)

wora: I think we should follow https://cloud.google.com/apis/design/.

wora:
Please merge this PR. I don't have write access.

PR #118, opened 2017-05-05
Regenerate proto docs 

ZackButcher:
We regenerate all of the docs in istio/api as well as the protos defined in istio/mixer. See the example at https://zackbutcher.github.io/istio.github.io/docs/reference/index.html

@rshriram PTAL at the proxy docs and make sure they line up with what you wanted (especially the page titles, urls).

@geeknoid how do you feel about the mixer aspect and adapter stuff?

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -1,5 +1,5 @@
 ---
-#layout: compress
+layout: compress
 ---
 {% include home.html %}
 


geeknoid:
I think I'd prefer splitting up the Aspect page into one page per aspect, so it would look more like the Mixer Adapters section.

Also, the denyChecker Config page has docs for the rpc.Status struct for some reason. That shouldn't be in there.

rshriram:
This looks good @ZackButcher . Thanks for doing this.. Could you please move the routing-rule to the top (set order to something like 990), so that we have
route rule
destination policy
fault injection
mesh configuration

Either way, once you push the docs, I will have to manually edit one of them (route rule) to redorder the content

PR #119, opened 2017-05-05
Prevent client busybox container from exiting prematurely

ayj:


frankbu:
This fixes busybox, but the documented task still doesn't work. First thing I noticed is that in the curl command is wrong: `kubectl exec -it ${CLIENT} -c echo -- curl echo:80`. It should be `-c busybox`. After changing that, however, it still doesn't work (curl returns 404)

frankbu:
Works for me

ayj:
Thanks for verifying Frank. I opened https://github.com/istio/manager/issues/683 to track the issue of k8s service being required on the client-side to use non-FQDN.

PR #122, opened 2017-05-05
Regenerate proxy proto docs

ZackButcher:
Splitting this out from https://github.com/istio/istio.github.io/pull/118 per @rshriram's request since we're still iterating on mixer things.

@rshriram should it be "manager config" or "mesh config"?


ZackButcher:
This PR also puts routing-rules first in the set of proxy docs.

PR #131, opened 2017-05-09
use type in proxy rule output

ayj:
Proxy config uses `type` instead of `kind`. Update the output examples in documentation in case a user copies that as an example. The PR for the corresponding update to istioctl is [here](https://github.com/istio/manager/pull/695).

### DIFF
--- a/_docs/tasks/request-routing.md
+++ b/_docs/tasks/request-routing.md
@@ -1,7 +1,7 @@
 ---
 title: Configuring Request Routing
 overview: This task shows you how to configure dynamic request routing based on weights and HTTP headers.
-            
+
 order: 50
 
 layout: docs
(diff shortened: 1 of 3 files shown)


_docs/tasks/request-routing.md
```diff
@@ -99,7 +99,7 @@ route requests to all available versions of a service in a random fashion.
    ```bash
    istioctl create -f route-rule-reviews-test-v2.yaml
    ```
-   
+
```
kyessenov: can you change `Cookie` to `cookie`?

ayj: Sure - PTAL. Can you merge when ready?

kyessenov: not authorized

PR #134, opened 2017-05-09
Add support for alpha and beta RBAC

andraxylia:
On some GKE cluster, it does not work to apply alpha if cluster is beta.


rshriram:
there are conflicts. And can you please provide more context on the difference between alpha/beta RBAC?

andraxylia:
I fixed the conflicts.

The difference is that GKE complains with an error if we try to apply the alpha config on the cluster that knows beta, so the safest approach is to have different configs.

The two files are different just  in the version field, alpha vs beta.

PR #147, opened 2017-05-10
Installation - Added `cd istio` command

saturnism:
Added `cd istio` to so that there is continuity to the next `cd install/kubernetes` step


_docs/tasks/installing-istio.md
```diff
@@ -36,6 +36,7 @@ clone Istio's [GitHub](https://github.com/istio/istio) repository:
 
    ```bash
    git clone --depth=1 https://github.com/istio/istio
+   cd istio
```
frankbu: Wouldn't it be better to just change step 2 to `cd istio/install/kubernetes`?

frankbu:
This has been fixed in #179.

PR #149, opened 2017-05-10
Installation - added sudo commands

saturnism:
Added sudo commands to install istioctl


_docs/tasks/installing-istio.md
```diff
@@ -90,6 +90,12 @@ clone Istio's [GitHub](https://github.com/istio/istio) repository:
    curl ${ISTIOCTL_URL}/istioctl-osx > /usr/local/bin/istioctl
    chmod +x /usr/local/bin/istioctl
    ```
+   
+   If your need to use `sudo`:
```
geeknoid: your -> you

Could you indicate why one might want/need to use sudo?

andraxylia:
This change is no longer needed, because the installation archive will contain istioctl binary. I removed those instructions completely, let's close this.

PR #176, opened 2017-05-10
fixed ingress task instructions

frankbu:


_docs/tasks/ingress.md
```diff
@@ -134,10 +151,55 @@ to configure ingress behavior.
...
+    ```bash
+    export SECURE_INGRESS_URL=169.47.243.100:32254
+ 
+    ```
+    
 1. Access the secured httpbin service using _curl_:
 
    ```bash
-   curl -k https://$INGRESS_URL/html
+   curl -k https://$SECURE_INGRESS_URL/html
```
kyessenov: it should be same ingress for secure and insecure.

_docs/tasks/ingress.md
```diff
@@ -60,34 +60,51 @@ to configure ingress behavior.
...
      "headers": {
-       "Accept": "*/*", 
-       "Content-Length": "0", 
-       "Host": "httpbin.default.svc.cluster.local:8000", 
-       "User-Agent": "curl/7.35.0", 
-       "X-Envoy-Expected-Rq-Timeout-Ms": "15000", 
-       "X-Request-Id": "59cf4fce-72e0-4470-ade5-f59149705944"
-     }
-   }
+   ...
```
kyessenov: any reason for removal?

_docs/tasks/ingress.md
frankbu: Yes, I know it turns out to be the same in the "load balancer" case, but in the nodePort case, it's different (<ip>:<nodeport for port 80> vs <ip>:<nodeport for port 443>). Wasn't sure how best to unify the instructions for the two cases, so this is what I came up with. I'll merge this for now, and see if we can come up with a better way to document it later.

_docs/tasks/ingress.md
frankbu: The exact output is not relevant for this task, only that there is some output, so I thought it was best to remove the extra noise.

Issue #157, opened 2017-05-10
Use inline scripts to fetch external IPs in tasks


ahmetb:
See the screenshot below:

- if I'm using NodePort, it provides me a handy script
- if I'm using external IP, it tells me to copy paste.

Why is there such a difference? I believe the latter can be achieved with a handy inline script too.

![image](https://cloud.githubusercontent.com/assets/159209/25913353/4aaa63e2-356f-11e7-8c29-96f97c4123d5.png)

ldemailly:
this is on 
https://github.com/istio/istio.github.io/blob/master/_docs/tasks/ingress.md

ldemailly:
it seems we are not doing this. if you feel it's needed reopen

PR #178, opened 2017-05-10
command to auto extract INGRESS_URL

ldemailly:
as per issue #157
(this change is a bit more complicated (/ugly) so I'm submitting it for review instead of just pushing it this time)

### DIFF
--- a/_docs/tasks/ingress.md
+++ b/_docs/tasks/ingress.md
@@ -58,13 +58,20 @@ to configure ingress behavior.
    ```bash
    kubectl get ingress -o wide
    ```
-   
    ```
    NAME      HOSTS     ADDRESS                 PORTS     AGE
    gateway   *         130.211.10.121          80        1d
-   export INGRESS_URL=130.211.10.121:80
    ```
-
+   In this example we would want to `export INGRESS_URL=130.211.10.121:80`, so:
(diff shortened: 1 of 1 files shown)


_docs/tasks/ingress.md
```diff
@@ -58,13 +58,20 @@ to configure ingress behavior.
...
-
+   In this example we would want to `export INGRESS_URL=130.211.10.121:80`, so:
+   ```bash
+   export INGRESS_URL=`kubectl get ingress -o wide | tail -1 | awk '{printf("%s:%s", $3,$4)}'`
+   echo INGRESS_URL=$INGRESS_URL
+   ```
+   which outputs
+   ```
+   INGRESS_URL=130.211.10.121:80
+   ```
```
frankbu: The :80 part was a bug ... see latest version of the task. 

The instructions for the "load balancers are not supported" case has become more complicated, so I'm gone in the opposite direction and made those instructions manual.

I think that if we want to make the instructions to simply call a command that sets the env variable automatically, we should figure it out for all cases.

rshriram:
do we still need this? @ldemailly ?

ldemailly:
the page changed since that feedback was given in #157 so I'll abandon this for now

PR #191, opened 2017-05-11
Add curl debugging instructions for users to play with Auth

wattli:


wattli:
@geeknoid @myidpt @lookuptable @wlu2016 please take a look, thanks.

_docs/samples/bookinfo.md
```diff
@@ -162,6 +162,49 @@ This application is polyglot, i.e., the microservices are written in different l
...
+   ```bash
+   ls /etc/certs/ 
+   ````
+   ```bash
+   cert-chain.pem   key.pem 
+   ```` 
+   
+   Step 4: send requests to another service, for example, productpage.
+   ```bash
+   curl https://productpage:9080 -v --key /etc/certs/key.pem --cert /etc/certs/cert-chain.pem -k
```
kyessenov: doesn't `-k` negate all cert validation and `--key` is then not used?

wattli: -k is for client not to check server cert,

--key is to present client to server

wattli:
@geeknoid , could you help to merge please?

_docs/samples/bookinfo.md
```diff
@@ -162,6 +162,49 @@ This application is polyglot, i.e., the microservices are written in different l
    200
    ```
 
+1. If you enabled auth and want to play with it, you can use curl from one envoy to send request to other services. For example, you want to ssh into the envoy container of details service, and send request to other services by curl. There are several steps:
+   
+   Step 1: get the details pod name
+   ```bash
+   kubectl get pods | grep details
```
lookuptable: nit: can you use label selector like `-l app=details`?

_docs/samples/bookinfo.md
```diff
@@ -162,6 +162,49 @@ This application is polyglot, i.e., the microservices are written in different l
...
+   ```bash
+   kubectl exec -it details-v1-4184313719-5mxjc -c proxy /bin/bash 
+   ```
+
+   Step 3: make sure the key/cert is in /etc/certs/ directory
+   ```bash
+   ls /etc/certs/ 
+   ````
+   ```bash
+   cert-chain.pem   key.pem
```
lookuptable: There should be also `root-cert.pem`

_docs/samples/bookinfo.md
```diff
@@ -162,6 +162,49 @@ This application is polyglot, i.e., the microservices are written in different l
...
+   ls /etc/certs/ 
+   ````
+   ```bash
+   cert-chain.pem   key.pem 
+   ```` 
+   
+   Step 4: send requests to another service, for example, productpage.
+   ```bash
+   curl https://productpage:9080 -v --key /etc/certs/key.pem --cert /etc/certs/cert-chain.pem -k
+   ````
```
lookuptable: extra ` here

_docs/samples/bookinfo.md
wattli: I intentionally hide root-cert.pem since it is not used below to avoid confusion.

_docs/samples/bookinfo.md
wattli: Will do in a follow up PR

_docs/samples/bookinfo.md
wattli: Will do in a follow up PR

_docs/samples/bookinfo.md
kyessenov: gotcha

PR #203, opened 2017-05-12
Address Ingress hackathon issues

### LINKED ISSUE #186: Issues with docs/tasks/ingress.html
From hackathon feedback:

- [x] Doc should include instructions for httpbin:
```
cd demos/apps/httpbin
kubectl apply -f <(istioctl kube-inject -f httpbin.yaml)
```
- [x] sample response is missing (in place of current: .. response ..)
- [ ] Use of `cat <<EOF | kubectl create -f -` is inconsistent with previous tutorials, where config was saved to files first.
- [x] $INGRESS_URL is exported with port :80, but then it’s being used in secured-ingress example, where connections should be made over port :443 for HTTPS.
- [x] /html endpoint is terrible example to use with curl, since it just emits snippet from “Moby Dick”, which is very confusing output to see.
- [ ] timeout example should include upstream request timeout in the output.
- [x] https://istio.io/docs/tasks/ingress.html#understanding- is missing content.
- [x] Missing cleanup instructions to remove services installed in this tutorial.

frankbu:
Fixes #186


_docs/tasks/ingress.md
```diff
@@ -52,48 +59,48 @@ to configure ingress behavior.
...
 
-   If your cluster is running in an environment that supports external load balancers,
-   use the ingress' external address:
+   * If your cluster is running in an environment that supports external load balancers,
+     use the ingress' external address:
 
-   ```bash
-   kubectl get ingress -o wide
-   ```
+     ```bash
```
ijsnellf: Stylistically I think it is easier to read when the command and expected results of executing the command are in the same block.

_docs/tasks/ingress.md
```diff
@@ -153,53 +167,58 @@ to configure ingress behavior.
    
 1. Determine the secure ingress URL:
  
-    If your cluster is running in an environment that supports external load balancers,
-    use the ingress' external address:
+   * If your cluster is running in an environment that supports external load balancers,
```
ijsnellf: Could we avoid repeating some of these steps somehow?

_docs/tasks/ingress.md
ijsnellf: Ah, I guess it is nice if they are separate for the `copy` button.

_docs/tasks/ingress.md
frankbu: There's probably a better way to document this, but I think the whole procedure may change in the future, so IMO it's good enough as is for now.

_docs/tasks/ingress.md
frankbu: Yes, this is one of the conventions we've decided to use in all the docs.

PR #206, opened 2017-05-12
added guidance for non-linux users using minikube

christian-posta:


README.md
```diff
@@ -15,3 +15,40 @@ To run the site locally with Docker, use the following command:
 ```bash
 docker run --rm --label=jekyll --volume=$(pwd):/srv/jekyll  -it -p 127.0.0.1:4000:4000 jekyll/jekyll jekyll serve
 ```
+
+
+
+#### Side note for those on non-linux machines
+
```
rshriram: question: do I need minikube to view the site locally? won't docker-machine work?
Specifically, how will you share files from host to minikube? Don't you need to setup shared folders etc. in virtual box?

christian-posta: Don't *need* it per-se; Docker machine would work. 

Minikube + xhyve automatically sets up the shared folders properly (i.e., under /Users/<username>) and configures the permissions for making changes locally and syncing them to the docker image, etc. VirtualBox under minikube tries to do that, but it doesn't work reliably. 

So I was intending to find a nice solution for people who are developing locally with istio+updating the docs; IMHO minikube is great for this. Otherwise, it's up to you to figure out the docker-machine, shared folders, sync the shared folders with the volume mounts in the docker images, etc and that can be a pain. Although, for docker-machine with xhyve i suspect it would work just like it does with minikube though i've not yet tried. Want me to change some of that text to just be "Minikube" specific instead?

rshriram: Oh reading the PR again, why doesn't docker for mac work on a Mac? I don't mind minikube, but I see that as a very useful thing for folks on Windows, but not on Mac, since the docker for Mac thing works pretty smoothly. 

But if your workflow is just minikube, I am okay with this. Do please fix the text to state that minikube on OSX is one option (standard docker for mac is another).

christian-posta: Yah, docker on mac works great. But the version of minikube i was using (0.17.1) when combined with virtual box didn't quite work well. Using xhyve (which is IMHO way better than virtual box .. for either minikube OR docker for mac) worked better for Minikube. I'll update the PR to clarify that.

christian-posta:
Made some clarifications. Let me know if that's better, or if i should further break out the instructions. Also let me know if you'd like me to squash those commits

ldemailly:

README.md
```diff
@@ -20,7 +20,45 @@ Make sure you are not introducing html errors or bad links:
...
 ```
 ```
+#### Side note for those on non-linux machines
+ 
+If you're developing locally but not on a Linux machine, you have a couple options. 
+You can opt to use [Docker for Mac](https://docs.docker.com/docker-for-mac/) / [Docker for Windows](https://docs.docker.com/docker-for-windows/). This will give you a docker environment from which to run the above docker container (which has all the of the correct Jekyll dependencies and Ruby versions installed). 
 
+
+If doing Istio development on Kubernetes and doing so locally with  [minikube](https://kubernetes.io/docs/getting-started-guides/minikube/)we highly recommend you use your OS's native virtualization. For example, to start minikube with native virtualization for Mac OSX try:
```
ldemailly: missing space before "we highly"

README.md
```diff
@@ -20,7 +20,45 @@ Make sure you are not introducing html errors or bad links:
...
+```
+
+You can see more about this command and how to install the [xhyve](https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#xhyve-driver) drivers by [taking a look at the xhyve driver documentation](https://github.com/zchee/docker-machine-driver-xhyve#install)
+
+We've had some difficulties getting virtualbox to configure shared folder permissions correctly when using minikube+virtualbox. 
+
+To build and run the website with minikube, bind to your `minikube ip` like this:
+
+```bash
+docker run --rm --label=jekyll  --volume=$(pwd):/srv/jekyll  -it -p $(minikube ip):4000:4000 jekyll/jekyll jekyll serve
```
ldemailly: this is very similar to the above (with $(minikube ip) replacing localhost)

README.md
ldemailly: This sounds like general advice that should be on the docs or in istio but maybe not in this file which is more about how to update/run the website?

christian-posta: maybe we should have a "developing istio" or "developers" area on the istio.io site? at least describing the various repos and how to contribute? and maybe setting up a local dev environment?

christian-posta: thanks for catching this; it's been corrected

PR #207, opened 2017-05-12
Rename demos to samples and use paths relative to root of the installation

andraxylia:


_docs/tasks/installing-istio.md
```diff
@@ -37,14 +37,20 @@ default namespace. They can be modified for deployment in a different namespace.
...
     * sample apps
     * the `istioctl` client binary, needed to inject Envoy as a sidecar proxy, and useful for creating routing rules and policies.
-    * a text file containing the Istio version.
+    * the istio.VERSION configuration file.
 
-3. Run the following command to determine if your cluster has [RBAC (Role-Based Access Control)](https://kubernetes.io/docs/admin/authorization/rbac/)
+1. Add the `istioctl` client to your PATH. For example, run the following commands on a Mac system:
+
+   ```bash
+   sudo ln -s $PWD/istioctl /usr/local/bin/
```
ldemailly: this is kinda ugly. maybe we can just add the directory to the path instead
```shell
export PATH=$PATH:`pwd`
```
that would work if we put more tools later too

this being said maybe we should but the isoctl binary in a `bin/` sub directory

andraxylia: There are other ways to do it. For instance, i have "." in my PATH. 
bin is a good idea. If the installation archive ends up being different than the instructions, please update the instructions.

ldemailly: 👍 , I will once we have the new layout in 0.1.2+

PR #218, opened 2017-05-15
adding mention of git.io/getIstio

ldemailly:


_docs/tasks/installing-istio.md
```diff
@@ -34,7 +34,11 @@ This page shows how to install and configure Istio in a Kubernetes cluster.
 For the {{ site.data.istio.version }} release, Istio must be installed in the same Kubernetes namespace as the applications. Instructions below will deploy Istio in the
 default namespace. They can be modified for deployment in a different namespace.
 
-1. Go to the [Istio release](https://github.com/istio/istio/releases) page, to download the installation file corresponding to your OS.
+1. Go to the [Istio release](https://github.com/istio/istio/releases) page, to download the installation file corresponding to your OS or run 
+   ```bash
+   curl -L https://git.io/getIstio | sh -
+   ``` 
+   to get the latest automatically downloaded and extracted (on MacOS and Ubuntu).
```
rshriram: to download and extract the latest release.

ldemailly: opsa !

ldemailly:
@rshriram some people may be (rightfully) wary of a `curl ...| sh -` installation so I think it's ok to first mention the manual download page (also there you get to pick which version to install) and the curl thing is more 'cool' and gadgety for now so only mentioned in passing for now is ok ?

Issue #226, opened 2017-05-16
broken links etc


ldemailly:
Add `gem "html-proofer"` and then:

```bash
ldemailly-macbookpro:istio.github.io ldemailly$ htmlproofer --assume-extension ./_site
Running ["ImageCheck", "ScriptCheck", "LinkCheck"] on ["./_site"] on *.html... 


Checking 185 external links...
Ran on 92 files!


- ./_site/blog/0.1-announcement.html
  *  External link https://grpc.io failed: response code 0 means something's wrong.
             It's possible libcurl couldn't connect to the server or perhaps the request timed out.
             Sometimes, making too many requests at once also breaks things.
             Either way, the return message (if any) from the server is: Peer certificate cannot be authenticated with given CA certificates
- ./_site/blog/index.html
  *  image /img/rss.gif does not have an alt attribute (line 1)
- ./_site/docs/concepts/network-and-auth/auth.html
  *  image ./img/auth/auth.svg does not have an alt attribute (line 1)
- ./_site/docs/concepts/policy-and-control/mixer-aspect-config.html
  *  internally linking to /docs/reference/api/mixer/adapters/denyChecker.html, which does not exist (line 35)
     <a href="/docs/reference/api/mixer/adapters/denyChecker.html">denyChecker</a>
  *  internally linking to /docs/reference/api/mixer/adapters/genericListChecker.html, which does not exist (line 35)
     <a href="/docs/reference/api/mixer/adapters/genericListChecker.html">genericListChecker</a>
  *  internally linking to /docs/reference/api/mixer/adapters/ipListChecker.html, which does not exist (line 35)
     <a href="/docs/reference/api/mixer/adapters/ipListChecker.html">ipListChecker</a>
  *  internally linking to /docs/reference/api/mixer/adapters/memQuota.html, which does not exist (line 35)
     <a href="/docs/reference/api/mixer/adapters/memQuota.html">memQuota</a>
  *  internally linking to /docs/reference/api/mixer/adapters/prometheus.html, which does not exist (line 35)
     <a href="/docs/reference/api/mixer/adapters/prometheus.html">prometheus</a>
  *  internally linking to /docs/reference/api/mixer/adapters/redisQuota.html, which does not exist (line 35)
     <a href="/docs/reference/api/mixer/adapters/redisQuota.html">redisQuota</a>
  *  internally linking to /docs/reference/api/mixer/adapters/statsd.html, which does not exist (line 35)
[...]

PR #227, opened 2017-05-16
Add html-proofer

ldemailly:
```
bundle install
rake test
```
To see the issues (#226)


ldemailly:
once it all passes we can probably put it as a commit hook

PR #228, opened 2017-05-16
zipkin addon in install - verified

rshriram:


ldemailly:
Wasn't there some issue with zipkin dependencies ? are those solved ?

rshriram:
Yep.. turns out there was no issue.

PR #229, opened 2017-05-16
zipkin-tracing task and bookinfo sample cleanup

frankbu:


_docs/tasks/istio-auth.md
```diff
@@ -172,3 +172,62 @@ kubectl exec <app-pod> -c proxy -- cat /etc/envoy/envoy-revX.json | grep ssl_con
 # Expect ssl_context in the output.
 ```
 
+## Playing with auth
```
rshriram: cc @wattli @myidpt . This seems much more cohesive. Setting up auth and playing with auth. The bookinfo sample setup is cleaner as well.

PR #239, opened 2017-05-16
fault injection disclaimer

kyessenov:


_docs/tasks/ingress.md
```diff
@@ -284,6 +284,7 @@ to set a timeout rule on calls to the httpbin service.
    This time a 504 (Gateway Timeout) appears after 3 seconds.
    Although httpbin was waiting 5 seconds, Istio cut off the request at 3 seconds.
 
+_Note:_ HTTP fault injection (abort and delay) is not currently supported by ingress proxies.
```
frankbu: Shouldn't we make this restriction more visible? I would put something like this at the beginning of the section:

## Setting Istio rules on an edge service

Similar to inter-cluster requests, Istio 
[routing rules]({{home}}/docs/concepts/traffic-management/rules-configuration.html)
can also be set for edge services
that are called from outside the cluster.

> Note: HTTP fault injection (abort and delay) is not currently supported by ingress proxies.

To illustrate we will use [istioctl]({{home}}/docs/reference/commands/istioctl.html)
to set a timeout rule on calls to the httpbin service.

PR #245, opened 2017-05-17
ratelimiting task improvements

frankbu:
Note that this change depends on yaml file changes in ihttps://github.com/istio/istio/pull/285


frankbu:
Can't easily reuse yaml files in bookinfo sample directory because of smoketest issues, so closing this PR in favor of #260

PR #256, opened 2017-05-17
Update _override.scss

kyessenov:


kyessenov:
why? we don't have to be consistent for pre elements vs rest of the page. It's slight grey with this change, which is good enough for code snippets.

kyessenov:
can we merge this? I don't know the implications of changing _all_ background colors.

PR #260, opened 2017-05-17
rate limiting and basic access control task improvements

frankbu:


_docs/tasks/basic-access-control.md
```diff
@@ -18,54 +18,66 @@ This task shows how to use Istio to control access to a service.
...
+   indicating that the `ratings` service is being called by the "v2" version of the `reviews` service.
+   
+   If you log in as any other user (or logout) you should see red ratings stars with each review,
+   indicating that the `ratings` service is being called by the "v3" version of the `reviews` service.
+
+1. Explicitly deny access to version `v3` of the `reviews` service. 
 
    ```bash
-   istioctl mixer rule create global ratings.default.svc.cluster.local -f deny-reviews.yml
+   istioctl mixer rule create global ratings.default.svc.cluster.local -f deny-reviews.yaml
```
rshriram: So, we have these files in the istio folder right? If not, in a separate PR to istio/istio, I would suggest creating these files in istio repo or doing the stdin thing (not sure istioctl accepts stdin). 

Apart from that, I would suggest changing the order of the commands (ask user to create file and then run command with that file). I found myself getting confused repeatedly at this spot, since all other tasks use a file that is already present in the repo. The rate limit and ACL doesn't.

_docs/tasks/basic-access-control.md
```diff
@@ -18,54 +18,66 @@ This task shows how to use Istio to control access to a service.
...
    ```
-  where `deny-reviews.yml` is 
+   where `deny-reviews.yaml` is
```
rshriram: Thank you for changing yml to yaml. Another inconsistency, with same user experience as above. We use yaml throughout the docs.

_docs/tasks/basic-access-control.md
```diff
@@ -89,6 +101,25 @@ Using a whitelist is a two step process.
    `checkExpression` is evaluated and checked against the list `[v1, v2]`. The check behavior can be changed to a blacklist by specifying
    `blacklist: true`. The expression evaluator returns the value of the `version` label as specified by the `checkExpression` key.
 
+The alpha version of `istioctl` does not yet support
```
rshriram: Nit. We decided to stop using "alpha" throughout docs. We could just say the current version of istioctl.

_docs/tasks/basic-access-control.md
```diff
@@ -89,6 +101,25 @@ Using a whitelist is a two step process.
...
+pushing adapter configurations, although there is a
+[workaround]({{home}}/docs/concepts/policy-and-control/mixer-aspect-config.html#pushing-configuration)
+if you want to try it.
```
rshriram: Why do we even need this para ? Can we kill it?

_docs/tasks/basic-access-control.md
```diff
@@ -89,6 +101,25 @@ Using a whitelist is a two step process.
...
+
+## Cleanup
+
+* Remove the mixer configuration rule:
+
+  ```bash
+  istioctl mixer rule create global ratings.default.svc.cluster.local -f samples/apps/bookinfo/mixer-rule-empty-rule.yaml
```
rshriram: This seems strange doesn't it? Using a create with empty object. Not sure if we can use istioctl delete. If not, this is a thing that we can definitely fix in istioctl (@ayj @esnible)

_docs/tasks/rate-limiting.md
```diff
@@ -53,16 +62,18 @@ Using Istio we can ensure that `1qps` is not breached.
...
-   generator is running (i.e., generating more than 5 req/s), the traffic generated by
-   your browser will be rate limited. When `reviews` service is unable to access `ratings`
-   service you stop seeing stars on the UI.
+
+1. Refresh the `productpage` in your browser.
+
+   While the load generator is running (i.e., generating more than 1 req/s), the traffic generated by
+   your browser will be rate limited. 
+   Notice that if you log in as user "jason" or any other user, the `reviews` service is unable to
+   access the `ratings` service, so you stop seeing stars, red or black.
```
rshriram: General feedback on the story: while this may have been the current flow, how about we generate load with user Jason (use curl -b user=jason...). This way we just have black stars or no stars.

_docs/tasks/rate-limiting.md
```diff
@@ -89,17 +100,24 @@ The following configuration applies a `1qps` rate limit only to version `v3` of
...
 
-   Since the `reviews-v3` service is unable to access the `ratings` service freely, we should see `red` stars very infrequently.
+1. Refresh the `productpage` in your browser.
+
+   As in the previous example, while the load 
+   generator is running (i.e., generating more than 1 req/s), the traffic generated by
+   your browser will be rate limited, but this time only if the request is from `reviews:v3`.
+   Notice that this time if you log in as user "jason" (the `reviews:v2` user) you will continue to see
+   the black ratings stars. Only the other users will stop seeing
+   the red ratings stars while the load generator is running.
```
rshriram: Okay, this complicates the simple story i suggested earlier.

_docs/tasks/rate-limiting.md
```diff
@@ -157,6 +175,21 @@ the following config validation error if any labels are missing.
 * quotas[RequestCount_byService_byUser].labels: wrong dimensions: descriptor expects 2 labels, found 0 labels
 ```
 
+## Cleanup
+
+* Remove the mixer configuration rule:
+
+  ```bash
+  istioctl mixer rule create global ratings.default.svc.cluster.local -f samples/apps/bookinfo/mixer-rule-empty-rule.yaml
```
rshriram: Same comment as before for create.

_docs/tasks/basic-access-control.md
frankbu: Yes, very strange, but Mandar told me this is the only way you can do it currently. Maybe I should put a Note in to explain this is a temporary solution until istioctl is more functional. The istioctl mixer commands are really very limited right now, so all we can do is use what we've got.

_docs/tasks/basic-access-control.md
frankbu: I had intended to use/add yaml files in istio/istio (some are there already, some are there, but not exactly correct, some need to be added), but I abandoned that because the smoketest depends on them (Bazel does, even if the code isn't even calling some of them), so given the pain it created to try that (my changes broke the build in several places ... would need to rework the smoketest), I gave up and just left this as Mandar had it (user needs to create the file with copy/paste). That said, I think I'll take your suggested and change it to do the pipe from stdin.

_docs/tasks/rate-limiting.md
frankbu: Right. the first one was setting the stage for this one.

_docs/tasks/basic-access-control.md
frankbu: Mandar seemed to want to include the whitelist example, but the way he had it just left the reader hanging ... so how do I do it?  So I added this to explain it's not supported yet, but you can try the workaround if you want.

PR #265, opened 2017-05-17
Update index.md

kyessenov:


kyessenov:
{{home}} seems to be istio.io/community here, I need istio.io/

PR #273, opened 2017-05-18
Mark all code fragments with language in integrating task

kyessenov:


ldemailly:
is there a "language" for output ? maybe like "console" or "text", it's not really bash

geeknoid:
I checked in a change this morning that makes it so the default language is "unknown" which makes things render correctly.

So I don't think this change is necessary at this point.

PR #280, opened 2017-05-19
Formatting fixes

LisaFC:
Fixing list formatting


LisaFC:
It doesn't actually make the blank line lists look notably different:
http://lcarey0-l.dub.corp.google.com:4000/docs/concepts/what-is-istio/overview.html

The font size for lists wasn't affecting them anyway as they had paragraph styling instead.

Ultimately though I'd like to remove the blank lines, though, so if you think I should add that to this CL then I can do that...

LisaFC:
PR, I mean! This is what happens when you keep switching between version control systems...

geeknoid:
Yes, I think we should do all this as a single PR.

List handling is particularly challenging around PRE blocks. The worst case
is this doc:
http://lcarey0-l.dub.corp.google.com:4000/docs/tasks/installing-istio.html.
It's difficult to get the stars lined up such that the markdown processor
does he right thing.

For example, if you don't indent things properly, you'll get extra spaces
on the left hand side of every line in a PRE block, which isn't good. Or
markdown will get confused in paragraph numbering and reset the number to 1.

So if we're going to change things, we should make sure it works everywhere
before pulling the trigger.

LisaFC:
AAAAAAARRRRGHHHH MARKDOWN. It's really great for quickly marking up simple docs but it's *terrible* for doing any kind of complex structuring like nested lists. I've used it a lot for Cloudsite docs recently and I don't think I've ever managed to "get it right" first time, or at least not without a lot of swearing.

[mutter mutter goes back to hewing HTML on tablets of stone mutter mutter]

OK, I'll try and sort out all the lists in one fell swoop. Watch this space...

LisaFC:
OK, I've now set it up so that paragraphs inside list items don't have huge spacing around them, which I *think* means that it doesn't matter how you space them in your Markdown, they all come out looking more or less the same. So if it's easier to space everything out in a complicated nested list, you can do that and it won't affect the presentation. PTAL.

LisaFC:
Also I think a solution to Ye Horrore that is the massive nested procedure in Installing Istio is to actually break it up into separate procedures. I can see people losing the will to live somewhere around step 7....

geeknoid:
Your PR shows 27 commits in it. I think something funny is happening to
your local repo. Do you need to rebase perhaps?

LisaFC:
Hmm, definitely something odd going on. I'll try rebasing once I've committed/stashed everything.

_sass/base/_markdown.scss
```diff
@@ -26,13 +26,14 @@
...
     line-height: 24px;
+    margin: 5px;
   }
 
   // Lists
   ol,
   ul {
     list-style: initial;
-    font-size: 15px;
+    font-size: 16px;
```
geeknoid: Can we fix this to not specify absolute font sizes in pixels? This doesn't work well on screens of higher density. Please use one of the SASS variables defined in _sass/base/_config.scss instead. I've been trying to stomp out these absolute pixel values over time, so if we're changing things, let's change them for the better.

geeknoid: And anyway, why do we need to override font-size and font-weight for lists? Shouldn't those just be the same as the main text of the doc?

LisaFC: Done.

PR #285, opened 2017-05-19
moved the rss icon

ryangrahamnc:
moved the rss icon to the top-right, directly under the header, and aligned with the right of the content


ryangrahamnc:
https://github.com/istio/istio.github.io/issues/232

PR #287, opened 2017-05-19
279 fixed code segments overlapping on toc

ryangrahamnc:
fixed code segments overlapping on toc in http://localhost:4000/docs/reference/commands/istioctl.html

Before:
image

After:
image

Might also want to see how this looks on a non-mac, given how mac does scrollbars differently than everyone else. I dont have anything to test on tho...


ryangrahamnc:
https://github.com/istio/istio.github.io/issues/279

PR #291, opened 2017-05-20
Update service-mesh.md

kyessenov:


rshriram:
@ldemailly can you please fix this? There is literally a 2 word change in this PR (not related to links). Presubmit has failed with a stack trace. This is precisely what I was worried about.

ldemailly:
it's because a previous one got in and master is already broken (which is what I was worried about :-) )

PR #296, opened 2017-05-22
zindex of toc on docs pages

ryangrahamnc:
wrapped a div around the non-toc contents of the docs pages. used position:relative and zindex to make toc overlap other content

image


ryangrahamnc:
image

PR #302, opened 2017-05-22
updated designs to match mockups

ryangrahamnc:
image

image

image

image

image

https://github.com/istio/istio.github.io/issues/167


true-index.html
```diff
@@ -7,7 +7,10 @@
 <div class="nav-hero-container">
     <div class="hero">
         <div class="container">
+            <h1 class="hero-title">Istio</h1>
+            <img class="hero-logo" src="{{home}}/img/istio-logo.svg" />
```
ldemailly: - ./_site/true-index.html
  *  image /img/istio-logo.svg does not have an alt attribute (line 1)

rshriram: does it have to have an alt attribute?
Also, can the html proofer run independent of the CLA check? I waited for it to run for half hour.. Nothing happened and I had to merge it out of expediency.

ldemailly: yes it does for assistive/screen reader, I fixed it on master

I asked @sebastienvas it does seem to be depending on CLA check somehow

PR #305, opened 2017-05-22
Update servicegraph image

douglas-reid:


douglas-reid:
@frankbu that suggestion sounds good to me. Do you want to take a stab, or want me to add that?

frankbu:
@douglas-reid, you might as well push the change along with this PR. I would suggest something like changing the text:
"and executing the `curl` request to confirm it's working"
to
"and generating some load on the application (e.g., executing some `curl` requests in a `while` loop)"

douglas-reid:
SG. Altered text to match your suggestion.

PR #311, opened 2017-05-22
Update metrics task and images

douglas-reid:


douglas-reid:
Please hold review for a moment.  More updates to follow.

PR #313, opened 2017-05-23
292 fixed chevron when nav autocloses on mobile

ryangrahamnc:
https://github.com/istio/istio.github.io/issues/292


ryangrahamnc:
@googlebot I signed it!

PR #325, opened 2017-05-23
Fix bad table creation in troubleshooting doc

douglas-reid:


douglas-reid:

PR #329, opened 2017-05-24
Remove the step about sourcing istio.VERSION

andraxylia:


geeknoid:
I pushed a change last night that fixed how a lot of the PRE blocks were indented so they wouldn't render in HTML with blanks on their left and would have the right indent relative to their containing bullet point. I think some of this change ends up undoing some of these fixes.

rshriram:
Andra, I suggest you close this PR and create a new one to just remove the instructions about istio.version

PR #337, opened 2017-05-24
Fix broken link to concept explanation for Mixer

earth2marsh:
… was trying to pop one too many parent directories.


ldemailly:

PR #349, opened 2017-05-26
installation doc update (for 0.1.5)

ldemailly:
- fix for https://github.com/istio/issues/issues/18  (clusterroles.rbac.authorization.k8s.io "istio-manager" is forbidden)
we should also change the yaml ?

- port forward should be in background, adding &


andraxylia:
This is a temporary thing until we fix the rules and release 0.1.6.

PR #358, opened 2017-05-31
Get newer snapshot of istio

ldemailly:
Cc @geeknoid


ldemailly:
not sure how this can/should be automated ? also maybe should pull from 0.1 branch at some point

Issue #367, opened 2017-06-07
provide updatable snapshot of previous releases


ldemailly:
either a separate (virtual) host
releases.istio.io/v0.1
or under istio.io as a sub directory (could possibly be a submodule)

the goal is that there is a branch (like https://github.com/istio/istio.github.io/tree/release-0.1 ) or some otherway to easily update the snapshot of alpha docs (and ditto for 0.2 etc... later)

I tried to use forks (e.g. https://ldemailly.github.io/istio-0-1.github.io/ ) but it doesn't seem possible to fork inside the same org (I suppose we could create a istio-releases org but that seems heavy handed)

cc @mandarjog @ayj @geeknoid

geeknoid:
I suggest the following:

- We create a branch in istio.github.io for every release.

- The active branch under development is called 'next'. The current release's branch is called 'current'.

- When a release takes place, the 'current' branch is renamed to its explicit version number (V0.1), the next branch is renamed to current, and a new next branch is created.

- We have an automated build process that takes all these branches, runs Jekyll for all of them, does whatever sed processing is necessary to patch URLs, and then dumps the resulting HTML into istio.github.io and checks that in as is. The versioned branches all get subdirectories, while the current branch is at the root.

  istio.github.io/ 
  istio.github.io/versions/V0.1
  istio.github.io/versions/next

This seems like a composition of simple steps.

ldemailly:
simplification proposal: we probably only need docs.istio.io to point to the last release (aka alpha and the `release-0.1` branch for now) and istio.io to be current (`master`) and that's it for the next 3 months; after which we can also erase alpha docs and replace them by beta docs... we will only need to keep more than 1 after GA

And in `master` we add a link in the header to "0.1 release docs" and that's it

ldemailly:
We're going to go with mandar's solution but for reference I got a very quick "github pages only, nothing else required" hack going with https://istio-releases.github.io/v0.1/docs/ by creating an org for releases and forking the site there

geeknoid:
What  is Mandar's solution(tm) ?

ldemailly:
> What is Mandar's solution(tm) ?

linked above:  #368

ldemailly:
See also #406, #407 and #379

ldemailly:
all done except the post submit on release-0.1 is failing

PR #368, opened 2017-06-07
versioned docs: add build scripts to be driven thru jenkins

mandarjog:
publishes the checked out branch to firebase hosting.

https://istiodocs-v01.firebaseapp.com/v-0.1/

Firebase hosting was chosen because
1. It is hosted :-)
2. easy to setup
3. automagic ssl cert provisioning. 

This should be attached to the jenkins framework to work automagically. 

Every maintained branch build will be hosted such that fron't end routing rule should be 
`istio.io/v-{version}  --> https://istiodocs-v-{version}.firebaseapp.com/v-{version}/`

### DIFF
--- a/_docs/reference/contribute/writing-a-new-topic.md
+++ b/_docs/reference/contribute/writing-a-new-topic.md
@@ -228,15 +228,14 @@ display a config file or a test file. To do so, you can't use normal markup and 
 use direct HTML. For example:
 
 ```
-{% raw %}<pre data-src="{{home}}/repos/istio/BUILD"></pre>{% endraw %}
+{% raw %}<pre data-src="https://raw.githubusercontent.com/istio/istio/master/BUILD"></pre>{% endraw %}
 ```
 
 which produces the following result:
 
-<pre data-src="{{home}}/repos/istio/BUILD"></pre>
+<pre data-src="https://raw.githubusercontent.com/istio/istio/master/BUILD"></pre>
 
-The `data-src` attribute specifies the path to the file to display. This has to be a file within the
-current site, it cannot come from a different site.
+The `data-src` attribute specifies the path to the file to display.
 
 ### Highlighting lines
 
--- a/_docs/reference/contribute/writing-a-new-topic.md
+++ b/_docs/reference/contribute/writing-a-new-topic.md
@@ -235,7 +235,9 @@ which produces the following result:
 
 <pre data-src="https://raw.githubusercontent.com/istio/istio/master/BUILD"></pre>
 
-The `data-src` attribute specifies the path to the file to display.
+The `data-src` attribute specifies the path to the file to display. [PrismJS](http://prismjs.com/) fetches 
+ the file using XMLHttpRequest. If the file is from a different origin site, CORS should be enabled on that site.
+Note that the github raw content site (`raw.githubusercontent.com`) is CORS enabled so it may be used here.
 
 ### Highlighting lines
 
--- a/_includes/home.html
+++ b/_includes/home.html
@@ -1,4 +1,1 @@
-{% assign home = "" %}
-{% if site.github.environment == "dotcom" %}
 {% assign home = site.baseurl %}
-{% endif %}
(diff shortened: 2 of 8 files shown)


ayj:
Is the intent to create one firebase project per branch?

mandarjog:
Yes. That's the way firebase likes it :-(

Using "iPhone 9";

sebastienvas:
Is this a postsubmit ? Which project should be used and where can I get the token ?

ldemailly:
this starts to address #367, thank you!

where is the mapping from release-0.1 branch to v01 ?

mandarjog:
https://istiodocs-v01.firebaseapp.com/v-0.1/

Now the build is published at a context root /v-0.1  
We can tie this together at the top level proxy now.

@ayj @sebastienvas ptal.

ldemailly:
can we kill the - in v-0.1 ?

mandarjog:
@ldemailly firebase needs every release to be in it's own project, So I have tacked on the v-0.1 to it.

_docs/reference/contribute/writing-a-new-topic.md
```diff
@@ -228,15 +228,14 @@ display a config file or a test file. To do so, you can't use normal markup and
 use direct HTML. For example:
 
 ```
-{% raw %}<pre data-src="{{home}}/repos/istio/BUILD"></pre>{% endraw %}
+{% raw %}<pre data-src="https://raw.githubusercontent.com/istio/istio/master/BUILD"></pre>{% endraw %}
```
geeknoid: This won't work. The URL reference needs to be within the current site, otherwise the browser will block the fetch. This is why we have the repos/istio submodule within the repo.

_docs/reference/contribute/writing-a-new-topic.md
```diff
@@ -228,15 +228,14 @@ display a config file or a test file. To do so, you can't use normal markup and
...
 ```
 
 which produces the following result:
 
-<pre data-src="{{home}}/repos/istio/BUILD"></pre>
+<pre data-src="https://raw.githubusercontent.com/istio/istio/master/BUILD"></pre>
 
-The `data-src` attribute specifies the path to the file to display. This has to be a file within the
```
geeknoid: Unless you have magic somehow, you can't just remove this requirement.

_includes/home.html
```diff
@@ -1,4 +1 @@
-{% assign home = "" %}
```
geeknoid: Removing this will break some use cases. We need to be able to:

Host the site on istio.io
Host the site locally when someone runs Jekyll
Host the site on people's own repos (geeknoid.istio.io)

Making this change will break the third case I believe.

mandarjog: I have tested it with case 3. It works fine there too.
https://mandarjog.github.io/docs/tasks/metrics-logs.html

look for `https://raw.githubusercontent.com/istio/istio/release-0.1/samples/apps/bookinfo/mixer-rule-additional-telemetry.yaml` link there

_docs/reference/contribute/writing-a-new-topic.md
mandarjog: CORS is enabled on `raw.githubusercontent.com` so we don't have to do anything special here.
So long as we are linking files from github, this works ok.

mandarjog: Trying to replicate all repos locally in /repos is clunky.

_docs/reference/contribute/writing-a-new-topic.md
mandarjog: see comment below.

mandarjog:
I will remove the gitmodule.
We should keep {home} because it shorter than site.baseurl, and gives us the opportunity to map.

ldemailly:
> Also, given this change, how about you remove the nested repos since we won't need those anymore?

how are those 2 related ? specifically, how would someone include a yaml from istio/istio without the sub module ?

mandarjog:
@ldemailly you use the absolute link like this one.   so long as CORS is enabled on that site.

`<pre data-src="https://raw.githubusercontent.com/istio/istio/master/BUILD">`

CORS *is* enabled on `githubusercontent.com`

PR #369, opened 2017-06-07
deemphasize stackoverflow, highlight troubleshooting

ldemailly:
Many of the questions on stackoverflow are troubleshooting or issues
kind

I suspect this is because so is the first link on this page, so pushing
it down and highlighting (with new Creative Commons icon) the
troubleshooting/bug reporting page


ldemailly:
it looks like this with this PR:
screen shot 2017-06-07 at 11 51 21 am

geeknoid:
Nuke the "s" on troubles.

Looks good otherwise.

PR #371, opened 2017-06-09
remove stackoverflow from footer

ldemailly:
most questions on stackoverflow are support questions better suited for the mailing list or the issues

(people will still be able to reach so from community/)


ldemailly:
as mentioned, if you look at https://stackoverflow.com/questions/tagged/istio right now it has mostly misplaced questions

stackoverflow is supposed to encyclopedic definitive answers not a support channel for alpha products for which answer will be completely outdated in weeks times

I don't think we should waste resources on curating on stackoverflow at this time, we have enough with the mailing lists and the issues

PR #373, opened 2017-06-12
Add notes explaining when to use replace rather than create for routing rules

LisaFC:
Fixes (some of?) https://github.com/istio/istio.github.io/issues/185

### DIFF
--- a/_docs/tasks/fault-injection.md
+++ b/_docs/tasks/fault-injection.md
@@ -21,12 +21,15 @@ This task shows how to inject delays and test the resiliency of your application
 * Initialize the application version routing by either first doing the
   [request routing](./request-routing.html) task or by running following
   commands:
+  
+  > Note: This assumes you don't have any default routes set yet. If you've already created default route rules for the sample, you'll need to use `replace` rather than `create` in the following commands.
+
 
   ```bash
   istioctl create -f samples/apps/bookinfo/route-rule-all-v1.yaml
   istioctl create -f samples/apps/bookinfo/route-rule-reviews-test-v2.yaml
   ```
-
+  
 ## Fault injection
 
 To test our BookInfo application microservices for resiliency, we will _inject a 7s delay_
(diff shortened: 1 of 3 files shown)


_docs/tasks/fault-injection.md
```diff
@@ -21,12 +21,15 @@ This task shows how to inject delays and test the resiliency of your application
 * Initialize the application version routing by either first doing the
   [request routing](./request-routing.html) task or by running following
   commands:
+  
+  > Note: This assumes you don't have any default routes set yet. If you've already created default route rules for the sample, you'll need to use `replace` rather than `create` in the following commands.
```
frankbu: It's not just setting "default" rules in these commands, so I think it should reworded slightly:

Note: This assumes you don't have any routes set yet. If you've already created conflicting route rules for the sample, you'll need to use `replace` rather than `create` in one or both of the following commands.

_docs/tasks/request-routing.md
```diff
@@ -22,11 +22,13 @@ This task shows you how to configure dynamic request routing based on weights an
...
 Because the BookInfo sample deploys 3 versions of the reviews microservice,
 we need to set a default route.
-Otherwise if you access the application several times, you would notice that sometimes the output contains
+Otherwise if you access the application several times, you'll notice that sometimes the output contains
 star ratings.
 This is because without an explicit default version set, Istio will
 route requests to all available versions of a service in a random fashion.
 
+> Note: This task assumes you don't have any default routes set yet. If you've already created default route rules for the sample, you'll need to use `replace` rather than `create` in the following commands.
+
```
frankbu: Note: This task assumes you don't have any routes set yet. If you've already created conflicting route rules for the sample, you'll need to use `replace` rather than `create` in one or both of the following commands.

_docs/tasks/fault-injection.md
LisaFC: Done.

_docs/tasks/request-routing.md
LisaFC: Done.

PR #392, opened 2017-06-16
istioctl delete supports removing mixer rules

kelseyhightower:


frankbu:
Exact same change is required here: https://github.com/istio/istio.github.io/blob/master/_docs/tasks/rate-limiting.md

Can do it in a separate PR if too late.

ldemailly:

PR #395, opened 2017-06-20
additional link to envoy doc to explain circuit breaker maxConnections etc

ldemailly:
that explains why we have the fields we do unlike the previous link


ldemailly:
@ZackButcher kuat mentioned I should git blame (you wrote the first version) because if I understood correctly the outlier stuff isn't really related to circuitbreaker - maybe we should remove that or clarify the 2 concepts ?

ZackButcher:
These docs were generated from the proto files in istio/api - those comments are SoT and should either be changed alongside these doc changes (ideally), or only update proto comments and regenerate the docs. We've made incremental changes to these docs, re-doing doc gen the first time is going to be a pain, unfortunately.

ZackButcher:
So to your question - I didn't actually write any of these, I merely generated the md files. IIRC @rshriram wrote most of this.

PR #396, opened 2017-06-20
Indicate which companies are currently backing Istio

### LINKED ISSUE #364: provide some information about team size
I got a question from a friend about the rough size of the investment behind Istio,
while some of it is clear from the roadmap, blog posts, and the various repos github stats,
some aggregate numbers would be nice to have

a subset of this would be extracting contributors across repos

it also gives me a chance to write a toy yet somewhat useful first Go program

ldemailly:
Fixes #364
Also renamed file with typo in name

Needs update when https://github.com/istio/istio/pull/421 is merged
(To switch to “master”)

Any way to make it work for a div instead of a pre @geeknoid  ?


ldemailly:
it looks like this for now:
screen shot 2017-06-20 at 12 44 54 am

geeknoid:
Is there precedent for this kind of disclosure in common OSS projects?

My sense is that what matters is the overall rate of contribution to the repos, independent of who is committing those changes, as well as the number of stars, forks, and followers of the repos. This is how external people can judge the traction of our technology in the market.

So I think we should punt on this.

ldemailly:
Talked to @zinuga and we can use manually curated list for now - please let me know what you want the exact text to be like

_faq/which-companies-contribute.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Which organizations contribute to Istio?
+order: 51
+type: markdown
+---
+{% include home.html %}
+
+Istio is actively contributed to by dozens of engineers from Google, Ibm, Redhat and others. It also includes Lyft's Envoy.
```
geeknoid: Ibm -> IBM
Redhat -> RedHat

I think I'd just say:

Istio is actively contribute to by dozens of engineers from Google, IBM, RedHat, Lyft (via Envoy development), and others.

ldemailly:
who is behind Istio shall remain a mystery :-)

PR #397, opened 2017-06-20
Revert "additional link to envoy doc to explain circuit breaker maxConnections etc"

rshriram:
Reverts istio/istio.github.io#395

While in Envoy, the concepts carry their own custom meaning, from Istio perspective, this was Envoy's CB and outlier detection were intentionally bundled into one unit. Linking to just circuit breaker is totally incorrect and just throws off the user, as our CB configuration talks about outlier detection as well.


ldemailly:
The outlier link is still there, above, even if it doesn't explain maxconnection while the added link does

This came from an actual support question btw

rshriram:
I was going to fix those two links in a subsequent PR

PR #399, opened 2017-06-20
Rename how-what-istio-started.md to how-was-istio-started.md

ldemailly:


ldemailly:
salvaging the non controversial part #396

PR #405, opened 2017-07-10
Typo fix + updated gitignore + grafana url fix

ldemailly:
to match what the versioned build script scripts/build.sh does


ldemailly:
note: I sort of did this backward, probably should make a PR first on master from a branch other than 0.1... will do that next time

PR #406, opened 2017-07-10
Point to 0.1 archive of docs and site version fixes

ldemailly:
Point to 0.1 archive of docs and site version fixes:
site.data.istio.version was used where it should be hardcoded to 0.1
(like in “our first release was …”) and vice versa (like in the footer)

There isn’t really a navigation (as I don’t know how to do that) but at
least the docs/welcome page points to 0.1

### DIFF
--- a/_docs/index.md
+++ b/_docs/index.md
@@ -12,6 +12,8 @@ type: markdown
 
 # Welcome
 
+Here you will find the latest ({{site.data.istio.version}}) documentation, follow this link for the [documentation for the latest stable release (0.1)](https://istio.io/v-0.1/docs/)
+
 Welcome to Istio's documentation home page. From here you can learn all about Istio by following
 the links below:
 
--- a/_docs/index.md
+++ b/_docs/index.md
@@ -10,11 +10,9 @@ type: markdown
 ---
 {% include home.html %}
 
-# Welcome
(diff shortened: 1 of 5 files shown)


_docs/index.md
```diff
@@ -12,6 +12,8 @@ type: markdown
 
 # Welcome
 
+Here you will find the latest ({{site.data.istio.version}}) documentation, follow this link for the [documentation for the latest stable release (0.1)](https://istio.io/v-0.1/docs/)
```
geeknoid: This shouldn't be the first thing at the top of the Welcome page. How about this:

- Change the Welcome header to be Welcome to Istio <version number>

- Add a new paragraph at the end of the page that points to the stable version

ldemailly:

_docs/index.md
ldemailly: good idea. see new rev - how about now ? also any idea how/where to make a real nav bar - or is this enough?

ldemailly:

ldemailly:
I am waiting for the redirects to be in place before merging this so we don't pollute the search with duplicate blogs/ faq/ home page etc

ldemailly:
redirects are in place so pushing this

PR #407, opened 2017-07-10
Update for 0.1 with pointer to current docs

ldemailly:
Mirror of #406

### DIFF
--- a/_docs/index.md
+++ b/_docs/index.md
@@ -10,7 +10,7 @@ type: markdown
 ---
 {% include home.html %}
 
-# Istio's documentation for version 0.1
+# Welcome - Istio 0.1
 
 Welcome to Istio's documentation archive for the 0.1 release.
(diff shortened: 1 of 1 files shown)


ldemailly:

_docs/index.md
```diff
@@ -10,9 +10,11 @@ type: markdown
 ---
 {% include home.html %}
 
-# Welcome
```
geeknoid: Can we leave the title as Welcome? This matches the tab name, and matches what we do in all other sections like Concepts, Tasks, Reference. Maybe just:

# Welcome - Istio 0.1

ldemailly: I changed it to match what you asked in #406 and to make it extra clear this set is potentially put of date.

ldemailly: changed - PTAL

PR #411, opened 2017-07-13
updating roadmap

ldemailly:
updating hybrid entry

### DIFF
--- a/_docs/reference/release-roadmap.md
+++ b/_docs/reference/release-roadmap.md
@@ -27,7 +27,8 @@ Istio 0.2 will be our second release, and will work across multiple Kubernetes c
 ## Istio 0.3
 
 Istio 0.3 will be our third release, focused on performance, scale, and stability. The full feature set for 0.3 is still TBD, based on feedback from the community. Some of the additional features we are considering include:
-- Support for multiple clusters in a single mesh and Istio meshes without Kubernetes.
+- Support for multiple clusters in a single mesh.
+- Support for Istio meshes without Kubernetes.
 - Client-oriented telemetry collection and distribution.
 - Global load balancing with autoscaling.
 - Support for the Open Service Broker APIs.
--- a/_docs/reference/release-roadmap.md
+++ b/_docs/reference/release-roadmap.md
@@ -12,7 +12,7 @@ type: markdown
 
 Istio 0.2 will be our second release, and will work across multiple Kubernetes clusters and some basic hybrid deployments. Istio 0.2 is focused on expanding the coverage of Istio to more places, and making it easier for users and vendors to integrate into the system. Istio 0.2 will include the following features:
 - TCP telemetry collection and policy enforcement.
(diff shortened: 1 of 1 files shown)


_docs/reference/release-roadmap.md
```diff
@@ -12,7 +12,7 @@ type: markdown
 
 Istio 0.2 will be our second release, and will work across multiple Kubernetes clusters and some basic hybrid deployments. Istio 0.2 is focused on expanding the coverage of Istio to more places, and making it easier for users and vendors to integrate into the system. Istio 0.2 will include the following features:
 - TCP telemetry collection and policy enforcement.
-- Hybrid deployments (multiple clusters, VMs).
+- Initial Raw VM support.
```
rshriram: the term Raw VM doesn't rhyme. Our discussion in istio groups literally talked about integrating with a VM cluster. Its still hybrid. The roadmap is a high level doc and we don't have to be super specific.
Fyi, @GregHanson and @kimikowang are still doing Consul integration which would make Istio work on Docker/VM only cluster. So, i am not sure why this is being kicked out, without having any discussion.

ldemailly: we're not doing "multiple clusters" for 0.2

rshriram: Sure, but we are doing VM only mode, and k8s+VMs. That is still hybrid deployment.

smawson: How about we just say:

Support for VMs joining an Istio Mesh.

Or something benign like that.

_docs/reference/release-roadmap.md
```diff
@@ -27,6 +27,7 @@ Istio 0.2 will be our second release, and will work across multiple Kubernetes c
 ## Istio 0.3
 
 Istio 0.3 will be our third release, focused on performance, scale, and stability. The full feature set for 0.3 is still TBD, based on feedback from the community. Some of the additional features we are considering include:
+- Hybrid deployments (multiple clusters, VMs).
```
smawson: Here let's say: Support for multiple clusters in a single mesh.

ldemailly: k adding also k8s less istio (not that wording)

ldemailly:
ptal

_docs/reference/release-roadmap.md
```diff
@@ -27,6 +27,7 @@ Istio 0.2 will be our second release, and will work across multiple Kubernetes c
 ## Istio 0.3
 
 Istio 0.3 will be our third release, focused on performance, scale, and stability. The full feature set for 0.3 is still TBD, based on feedback from the community. Some of the additional features we are considering include:
+- Support for multiple clusters in a single mesh and Istio meshes without Kubernetes.
```
smawson: Break this into two parts, they're separate.

- Support for multiple clusters in a single mesh.
- Support for Istio meshes without Kubernetes.

PR #413, opened 2017-07-15
Added installation steps via helm charts.

gyliu513:
Fixed https://github.com/istio/istio.github.io/issues/412

### DIFF
--- a/_docs/tasks/installing-istio.md
+++ b/_docs/tasks/installing-istio.md
@@ -36,7 +36,9 @@ This page shows how to install and configure Istio in a Kubernetes cluster.
 * If you previously installed Istio on this cluster, please uninstall first by following the
   [uninstalling]({{home}}/docs/tasks/installing-istio.html#uninstalling) steps at the end of this page.
 
-## Installation steps
+You can use the [Istio Helm chart](https://github.com/kubernetes/charts/tree/master/incubator/istio) to install, or follow the steps below for manual install:
+
+## Manual installation steps
 
 For the {{ site.data.istio.version }} release, Istio must be installed in the same Kubernetes namespace as the applications. 
 Instructions below will deploy Istio in the
--- a/_docs/tasks/installing-istio.md
+++ b/_docs/tasks/installing-istio.md
@@ -36,7 +36,7 @@ This page shows how to install and configure Istio in a Kubernetes cluster.
 * If you previously installed Istio on this cluster, please uninstall first by following the
   [uninstalling]({{home}}/docs/tasks/installing-istio.html#uninstalling) steps at the end of this page.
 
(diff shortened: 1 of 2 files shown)


gyliu513:
I signed it!

gyliu513:
Thanks @ldemailly , how about make this on top of the installation page, so  that if people want a simple installation steps, they can just goto helm without reading the whole manual installation steps.

ldemailly:
The page you want to link https://github.com/kubernetes/charts/tree/master/incubator/istio
says:
screen shot 2017-07-16 at 1 35 18 am
with "here" being https://istio.io/docs/tasks/installing-istio.html#installation-steps which is this page, thus we would create a loop with your change

Can you first change https://github.com/kubernetes/charts/tree/master/incubator/istio to be self contained (and reference this page as "manual steps" if you want) ?

or again we can add a new section but not at the top - it'll show up in table of content box:

screen shot 2017-07-16 at 1 37 39 am

gyliu513:
Posted a PR in to fix istio helm chart https://github.com/kubernetes/charts/pull/1496 @ldemailly

ldemailly:
Please remove the separate header and instead add the single paragraph:
```
You can use the [istio helm chart](https://github.com/kubernetes/charts/tree/master/incubator/istio) to install, or follow the steps below:
```

ldemailly:
ping @gyliu513

ldemailly:
@gyliu513 thanks for make some of the changes, though I'm not exactly in love with the "manual install" part (should probably be "manual installation", install is a verb (or a colloquial noun/abbrev)

I'll defer to @geeknoid to PTAL

ldemailly:
ps: avoid doing force pushes as it erases the history in github

_docs/tasks/installing-istio.md
```diff
@@ -36,7 +36,9 @@ This page shows how to install and configure Istio in a Kubernetes cluster.
 * If you previously installed Istio on this cluster, please uninstall first by following the
   [uninstalling]({{home}}/docs/tasks/installing-istio.html#uninstalling) steps at the end of this page.
 
-## Installation steps
+You can use the [Istio Helm chart](https://github.com/kubernetes/charts/tree/master/incubator/istio) to install, or follow the steps below for manual installation:
+
+## Manual installation steps
```
geeknoid: I would move the paragraph above inside of the # Installation steps header, and leave the header name unchanged.

There should be steps in the Uninstalling section regarding Helm charts too.

gyliu513: Done, thanks @geeknoid

gyliu513:
Removed the update to istio auth, thanks @ldemailly

PR #414, opened 2017-07-17
No need to put `istioctl` client to PATH manually.

gyliu513:
/cc @ldemailly


gyliu513:
Thanks @ldemailly , when myself and my colleagues read this document, they will always do step 3 by adding the `istioctl` to `PATH` manually, but this is not needed if I download it via `curl -L https://git.io/getIstio | sh -`.

This patch is clarifying this is only need if you download the file from istio release directly.

ldemailly:
thx!

PR #415, opened 2017-07-18
RBAC clarification

andraxylia:


_docs/tasks/installing-istio.md
```diff
@@ -70,7 +70,11 @@ default namespace. They can be modified for deployment in a different namespace.
    ```
    * If the command displays an error, or does not display anything, it means the cluster does not support RBAC, and you can proceed to step 5 below.
    
-   * If the command displays 'beta' version, or both 'alpha' and 'beta', please apply istio-rbac-beta.yaml configuration:
+   * If the command displays 'beta' version, or both 'alpha' and 'beta', please apply istio-rbac-beta.yaml configuration as show below:
+   
+   *(Note: If you deploy Istio in another namespace than the "default" namespace, replace the "namespace: default" line
```
ldemailly: would does `default` work inside a *italic section, `testing`, 1,2,3*

looks like it does, try ? (check how it looks like on your local rendered version)

also once this is in you would cherry pick it on the release-0.1

_docs/tasks/installing-istio.md
```diff
@@ -84,6 +88,10 @@ default namespace. They can be modified for deployment in a different namespace.
    ```
    
    * If the command displays only 'alpha' version, please apply istio-rbac-alpha.yaml configuration:
+   
+   *(Note: If you deploy Istio in another namespace than the "default" namespace, replace the "namespace: default" line 
+      in all ClusterRoleBinding resources with the actual namespace.)*
```
ldemailly: ditto

andraxylia:
That command does not work for me, but my plugin IDE shows that at italics.

ldemailly:
so can you please put the command / keywords in backquote ?
ie 
```
   *(Note: If you deploy Istio in another namespace than the `default` namespace, replace the `namespace: default` line in all ClusterRoleBinding resources with the actual namespace.)*
```
what doesn't work about
`docker run --rm --label=jekyll --volume=$(pwd):/srv/jekyll  -it -p 127.0.0.1:4000:4000 jekyll/jekyll jekyll serve`

PR #429, opened 2017-08-15
Fix auth related doc

wattli:


_docs/tasks/istio-auth.md
```diff
@@ -54,23 +54,6 @@ Istio CA is up if the "AVAILABLE" column is 1.
 
    Istio Auth is enabled if the line `    authPolicy: MUTUAL_TLS` is uncommented.
 
-1. Check Istio Auth is enabled on Envoy proxies.
```
wattli: This is not needed since LDS in implemented, ssl_context will not be in envoy config file.

PR #431, opened 2017-08-16
Update inject container name from istio-proxy to proxy.

gyliu513:
Fixed https://github.com/istio/istio.github.io/issues/430


ayj:
`istio-proxy` is the correct container name in the master branch (see [here](https://github.com/istio/pilot/blob/master/platform/kube/inject/inject.go#L59)). `proxy` is only used by the 0.1 branch.

gyliu513:
@ayj I think that we should update document to `proxy`, as the document will enable user to get the released version which is `proxy`, comments?

ayj:
Can you rebase this fix against the release-0.1 branch which corresponds to the 0.1.x versions?

gyliu513:
v0.1 document https://istio.io/v-0.1/docs/tasks/istio-auth.html using `proxy` instead and should works. @ayj

PR #442, opened 2017-09-12
added openshift specific install steps

debianmaster:
Added openshift specific install steps. A sample screenshot is added here for your reference so you can see how it looks like after adding these steps.

screen shot 2017-09-12 at 12 45 26 pm

### DIFF
--- a/_docs/tasks/installing-istio.md
+++ b/_docs/tasks/installing-istio.md
@@ -29,7 +29,7 @@ This page shows how to install and configure Istio in a Kubernetes cluster.
   ```bash
   $(bx cs cluster-config <cluster-name>|grep "export KUBECONFIG")
   ```
-* If you are using [Openshift Origin](https://www.openshift.org) version 3.7 or later, Openshift by default does not allow containers running with uid 0, you can enable this to istio service accounts as follows:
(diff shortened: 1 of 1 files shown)


debianmaster:
I signed it!

_docs/tasks/installing-istio.md
```diff
@@ -29,6 +29,13 @@ This page shows how to install and configure Istio in a Kubernetes cluster.
   ```bash
   $(bx cs cluster-config <cluster-name>|grep "export KUBECONFIG")
   ```
+* If you are using [Openshift Origin](https://www.openshift.org) version 3.7 or later, Openshift by default does not allow containers running with uid 0, you can enable this for Istio's service accounts for ingress and egress as follows:
+
```
frankbu: ... uid 0. You ...

... enable this for Istio's ingress and egress service accounts as ...

_docs/tasks/installing-istio.md
```diff
@@ -29,6 +29,13 @@ This page shows how to install and configure Istio in a Kubernetes cluster.
...
+  ```bash
+ oc adm policy add-scc-to-user anyuid -z istio-ingress-service-account -n istio-system
+ oc adm policy add-scc-to-user anyuid -z istio-egress-service-account -n istio-system
+  ```  
+You can expose services either via istio ingress or using default openshift router.
```
frankbu: ... expose services using Istio ingress or the default Openshift router.

PR #443, opened 2017-09-12
update routing rule concepts and task docs

frankbu:
These changes require fix for https://github.com/istio/pilot/issues/1236

### DIFF
--- a/_docs/concepts/traffic-management/rules-configuration.md
+++ b/_docs/concepts/traffic-management/rules-configuration.md
@@ -22,29 +22,38 @@ service to version "v1" can be described using the Rules DSL as
 follows:
 
 ```yaml
-destination: reviews.default.svc.cluster.local
-route:
-- tags:
-    version: v1
-  weight: 100
+apiVersion: config.istio.io/v1alpha2
+kind: RouteRule
+metadata:
+  name: reviews-default
+spec:
+  destination:
+    name: reviews
+  route:
+  - labels:
+      version: v1
+    weight: 100
 ```
 
 The destination is the name of the service to which the traffic is being
-routed. In a Kubernetes deployment of Istio, the route *tag* "version: v1"
-corresponds to a Kubernetes *label* "version: v1".  The rule ensures that
-only Kubernetes pods containing the label "version: v1" will receive
-traffic. Rules can be configured using the
-[istioctl CLI]({{home}}/docs/reference/commands/istioctl.html). See the
+routed. The route *labels* identify the specific service instances that will
+recieve traffic. For example, in a Kubernetes deployment of Istio, the route
+*label* "version: v1" indicates that only pods containing the label "version: v1"
+will receive traffic.
+
+Rules can be configured using the
+[istioctl CLI]({{home}}/docs/reference/commands/istioctl.html), or in a Kubernetes
+deployment using the `kubectl` command instead. See the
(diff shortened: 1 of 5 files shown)


_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -53,58 +62,91 @@ following important aspects must be kept in mind while writing route rules:
...
+destination:
+  name: reviews
 ```
 
-The *destination* value SHOULD be a fully qualified domain name (FQDN). It
-is used by Pilot for matching rules to services. For example,
-in Kubernetes, a fully qualified domain name for a service can be
-constructed using the following format: *serviceName.namespace.dnsSuffix*.
+The *destination* value specifies, implicitly or explicitly, a fully qualified
+domain name (FQDN). It is used by Pilot for matching rules to services.
```
rshriram: Istio Pilot

_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -53,58 +62,91 @@ following important aspects must be kept in mind while writing route rules:
...
+
+The FQDN of the service is composed from three components: *name*,
+*namespace*, and *domain*:
```
rshriram: You might want to specify the alternative as well. name+namespace+domain OR service (it works now. and thats what Consul and egress rules use).

_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -53,58 +62,91 @@ following important aspects must be kept in mind while writing route rules:
...
+```
+
+More commonly, to simplify and maximize reuse of the rule (for example, to use
+the same rule in more than one namespace or domain), the rule destination
+specifies only the *name* field, relying on defaults for the other
+two.
+
+The default value for the *namespace* is the namespace of the rule
+itself, which can be specified in the *metadata* field of the rule, or
+at config time using the `-n` option of the CLI. The default value of
```
rshriram: or during rule install using the `istioctl -n <namespace> create -f` or `kubectl -n <namesapce> create -f` command.

_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -53,58 +62,91 @@ following important aspects must be kept in mind while writing route rules:
...
+  namespace: default
+  domain: svc.cluster.local
+```
+
+More commonly, to simplify and maximize reuse of the rule (for example, to use
+the same rule in more than one namespace or domain), the rule destination
+specifies only the *name* field, relying on defaults for the other
+two.
+
+The default value for the *namespace* is the namespace of the rule
```
rshriram: Lets also add that namespace and domain are kubernetes specific. Other platforms could use the domain.service field to specify the FQDN of the service.

_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -53,58 +62,91 @@ following important aspects must be kept in mind while writing route rules:
...
+itself, which can be specified in the *metadata* field of the rule, or
```
rshriram: Should we have a section for the metadata field as well? Also, I am thinking if its worth while to have a separate call out subsection on inheritance of namespace, domain fields. What do you think? (does not have to be in this PR, but can be in a subsequent PR).

rshriram:
merging and marking the rest as TODOs as people don't have a reference point.

PR #447, opened 2017-09-15
Trimming install task. remove redundant content

rshriram:


rshriram:
ignoring presubmit errors for lyft.github.io since we will make a separate pass for that.

rshriram:
170 and 193 are being removed right?
Also, I couldn't easily remove the others because there is no associated task for grafana/service graph

frankbu:
No, I mean the new lines 170 and 193, after your deletions. I fixed them.

ldemailly:
can you describe why the removal ?
cc @andraxylia who works on install and @douglas-reid for zipkin/jaeger

douglas-reid:
Is this because we have a task on Distributed Tracing as well?

PR #451, opened 2017-09-15
Corrections and improvements

frankbu:


_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -182,15 +219,20 @@ For example, the following rule will route 25% of traffic for the "reviews" serv
...
-- labels:
-    version: v2
-  weight: 25
-- labels:
-    version: v1
-  weight: 75
+apiVersion: config.istio.io/v1alpha2
+kind: RouteRule
+metadata:
+  name: reviews-bluegreen
```
rshriram: can we call this something else? Like canaryrollout ? Bluegreen usually implies atomic swap from one infra to another.. Or definitions vary.

frankbu: done

PR #454, opened 2017-09-17
Fix bad link error for new pages

### LINKED ISSUE #383: when adding new pages/docs/blogs the edit link gives an error
e.g.
  *  External link https://github.com/istio/istio.github.io/edit/master/_posts/2017-06-12-0.1-canary.md failed: 404 No error

there should be a way to configure jekyll to use the pr's url instead of master so no link 404s as that masks other errors when adding new content

the easiest 'fix' is to ignore all links to github but that would also mask other valid errors

ldemailly:
Fixes #383


ldemailly:
tested locally (including that it still catches bad link even into github) and passed the build here

merging so the other PR can be rebased and show real green/red status

btw it's also faster now as there is 138 external links vs 233 before

PR #465, opened 2017-09-18
Setup tab

rshriram:
Will be the home of all Istio installation related documents.


_docs/index.md
```diff
@@ -18,7 +20,12 @@ the links below:
 - [Concepts]({{home}}/docs/concepts/). Concepts explain some significant aspect of Istio. This
 is where you can learn about what Istio does and how it does it.
 
-- [Tasks]({{home}}/docs/tasks/). Tasks show you how to do a single directed activity with Istio. Starting with [downloading and installing]({{home}}/docs/tasks/installing-istio.html) Istio.
+- [Setup]({{home}}/docs/setup/). Setup contains instructions for installing
+  Istio control plane in various environments (e.g., Kubernetes, Consul,
```
Rachael-Graham: add "the": `... for installing the Istio control...`

_docs/index.md
```diff
@@ -18,7 +20,12 @@ the links below:
...
+  etc), as well as instructions for installing the sidecar in the
```
Rachael-Graham: etc.

_docs/setup/index.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Setup
+overview: Installation instructions for Istio control plane and sidecar on different platforms.
```
Rachael-Graham: I like the overview you have on the Welcome/index: "Setup contains instructions for installing the Istio control plane in various environments (e.g., Kubernetes, Consul, etc.), as well as instructions for installing the sidecar in the application deployment." Making them the same would also increase consistency.

ldemailly:
could you git mv the current content (except https://istio.io/docs/tasks/installing-istio.html which should stay, but should be updated to point to new locations of the sub task) so we see actual changes in content ?

cc @andraxylia

_docs/setup/install-kubernetes.md
```diff
@@ -0,0 +1,220 @@
+---
```
andraxylia: You should use git mv, this seems to be the old file?

rshriram:
I did this because I wanted to main the installing-istio link. If you don't mind nixing that link, I would happily git mv. Besides, I am also cleaning up that link.

_docs/setup/install-sidecar.md
```diff
@@ -0,0 +1,235 @@
+---
+title: Installing Istio Sidecar
```
andraxylia: We agreed to not using sidecar anymore, but rather proxy or Envoy?

rshriram: Other way around. No proxy. Use Envoy or Sidecar. And given that we have multiple sidecar technologies, it makes sense to keep this section generic for the moment.

ldemailly:
I do want to keep installing-istio - I'm talking about kubernetes and consul docs

rshriram:
I added redirects.

rshriram:
@ldemailly ^^^.

_docs/setup/install-kubernetes.md
```diff
@@ -127,108 +127,42 @@ default namespace. They can be modified for deployment in a different namespace.
      ```
      This command will install Pilot, Mixer, Ingress-Controller, and Egress-Controller, and the Istio CA (Certificate Authority).
 
-1. *Optional:* Install addons for metric collection and/or request tracing as described in the following sections.
-
-### Enabling metrics collection
```
ldemailly: that section should be green in some other doc in same PR no ?

_docs/setup/install-kubernetes.md
```diff
@@ -127,108 +127,42 @@ default namespace. They can be modified for deployment in a different namespace.
...
    istio-egress-1684034556-fhw89    1/1       Running   0          5h
    istio-ingress-1842462111-j3vcs   1/1       Running   0          5h
    istio-pilot-2275554717-93c43     2/2       Running   0          5h
    istio-mixer-2104784889-20rm8     1/1       Running   0          5h
-   prometheus-3067433533-wlmt2      1/1       Running   0          5h
-   servicegraph-3127588006-pc5z3    1/1       Running   0          5h
    ```
 
 
-## Deploy your application
```
ldemailly: same for this, should at least be a link ?

_docs/setup/install-kubernetes.md
rshriram: Nope. Its being removed. I spoke to @douglas-reid and he is going to rework this content in a different PR.

_docs/setup/install-kubernetes.md
rshriram: This is a setup tab. There is a dedicated install sidecar section. We will add next steps once this section is done. Cannot finish everything in one doc.

_docs/setup/install-kubernetes.md
ldemailly: IMO it's not a great experience to have it disappear completely even for a few days when it could be moved and linked

_docs/setup/install-kubernetes.md
ldemailly: I'm not a fan of disappearing content before the replacement is made

_docs/index.md
```diff
@@ -10,6 +10,8 @@ type: markdown
 ---
 {% include home.html %}
 
+Follow this link for the archive of [most recent stable release (0.1) documentation](https://istio.io/v-0.1/docs/).
```
ldemailly: 👍 on making it more proeminent (we may need to put it in all page nave if we could)

_docs/setup/install-kubernetes.md
rshriram: I mean in a different task

kyessenov: I assume the text is backed up in a branch...

_docs/setup/install-kubernetes.md
```diff
@@ -127,108 +127,42 @@ default namespace. They can be modified for deployment in a different namespace.
...
-
-You can now deploy your own application, or one of the sample applications provided with the installation,
```
andraxylia: This should be put back

rshriram: done.. As long as this is a quick start, I don't mind.

_docs/setup/install-kubernetes.md
rshriram: I added all the bloat again.

_docs/tasks/installing-istio.md
```diff
@@ -1,19 +0,0 @@
...
-
-layout: docs
-type: markdown
----
-
-{% include home.html %}
-
-This task describes the steps needed for istio installation on various platforms.
-
-* [Quick Start - Kubernetes Platforms]({{home}}/docs/tasks/install-kubernetes.html).
```
ldemailly: as I said a couple times, can we keep that page with links into `setup/` instead of `tasks/`

rshriram: I added redirect_from .. Please see the header. This is what we follow for all the blog articles as well. Any external links will get redirected to the correct page.

ldemailly: :+1: missed that in the new file/didn't know that feature in jekyll

_docs/setup/install-sidecar.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: Integrating Services into the Mesh
+title: Installing Istio Sidecar
```
ldemailly: maybe add the redirect from it's previous location for that one too, it's likely to be linked in

PR #467, opened 2017-09-18
Added link to Apache Mesos.

joerg84:


ldemailly:
if we started adding links we'd have to put links for every other platform ?

joerg84:
@ldemailly My initial concern was actually that the first mention on the site should use Apache Mesos (vs. just Mesos)
Still, I am happy to update all links.

ldemailly:
I guess I was too subtly trying to hint that imo we shouldn't have any link to "future" stuff at all and names are enough, but it's fine either way/I don't care that much about it - no need to change further, thanks!

PR #474, opened 2017-09-20
New samples

rshriram:
cc @frankbu FYI
They won't be visible on the page because they are marked as drafts. Once you remove that line, they become visible on the pages.
@douglas-reid / @ZackButcher PTAL at the rough setup in consul/intelligent routing.
The idea here is to describe all the necessary pieces for a user to go through a particular aspect of Istio namely
1. routing
2. telemetry & traces
3. policy enforcement
4. ...
Each aspect could have one or more tasks, and an associated setup/tweak. 

I am just putting it out here to see how this will pan out. If it doesn't feel right, we can take it off. Also, these step by step walk throughs make sense, only if we can gradually enable different aspects of Istio as per (https://docs.google.com/document/d/1veSJJJUU9vEh1NWeVuoDwiAUmSCvOjtxUP2sGEU-xdg) samples section. If we are going to ask people to change config maps, and do acrobatics, then this is a pointless effort and we might as well let people go through the usual task daisy chain.


_docs/samples/consul-integration.md
```diff
@@ -0,0 +1,75 @@
...
+layout: docs
+type: markdown
+draft: true
+---
+{% include home.html %}
+
+This sample deploys the Bookinfo application in a simple Docker Compose environment using Consul as the service registry, and demonstrates various features of the Istio service mesh on non-kubernetes platforms.
+
+## Before you begin
+* Setup Istio by following the instructions in the [Installation guide]({{home}}/docs/tasks/install-kubernetes.html).
```
ldemailly: it's setup/ not docs/

PR #481, opened 2017-09-21
switched consul steps to use istioctl context-create rather than kubectl

GregHanson:
no longer a requirement to use kubectl for consul steps

### DIFF
--- a/_docs/samples/consul-integration.md
+++ b/_docs/samples/consul-integration.md
@@ -10,11 +10,12 @@ draft: false
 ---
 {% include home.html %}
 
-This sample deploys the Bookinfo application in a simple Docker Compose environment using Consul as the service registry, and demonstrates various features of the Istio service mesh on non-kubernetes platforms.
+This sample deploys the Bookinfo application in a simple Docker Compose environment using Consul as the service 
+registry, and demonstrates various features of the Istio service mesh on non-kubernetes platforms.
 
 ## Before you begin
 * Setup Istio by following the instructions in the [Installation guide]({{home}}/docs/setup/install-kubernetes.html).
-  You will need `kubectl` and Steps 1-4 from `Installation step` only for this Task.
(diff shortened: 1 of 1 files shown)


_docs/samples/consul-integration.md
```diff
@@ -33,9 +33,7 @@ Note that Istio pilot agent is running inside each app container so as to coordi
 1. First step is to generate a configuration file which will be used by `istioctl` and Istio Discovery
 
     ```bash
-    kubectl config set-cluster local --server=http://172.28.0.13:8080
-    kubectl config set-context local --cluster=local
-    kubectl config use-context local
+    istioctl context-create --api-server http://172.28.0.13:8080
```
rshriram: We also need to tweak the "Before you begin" section, and remove the kubectl part there. Ideally, if we are bringing up the API server as part of docker compose, we don't need any reference to the kube installation instructions right?

GregHanson: you are right, I missed removing kubectl - updated

_docs/samples/consul-integration.md
```diff
@@ -10,11 +10,12 @@ draft: false
...
 {% include home.html %}
 
-This sample deploys the Bookinfo application in a simple Docker Compose environment using Consul as the service registry, and demonstrates various features of the Istio service mesh on non-kubernetes platforms.
+This sample deploys the Bookinfo application in a simple Docker Compose environment using Consul as the service 
+registry, and demonstrates various features of the Istio service mesh on non-kubernetes platforms.
 
 ## Before you begin
 * Setup Istio by following the instructions in the [Installation guide]({{home}}/docs/setup/install-kubernetes.html).
-  You will need `kubectl` and Steps 1-4 from `Installation step` only for this Task.
+  You will need Steps 1-3 from `Installation step` only for this Task.
```
rshriram: Remove this entire thing.. Copy the three steps here. Those are generic steps and have nothing to do with k8s.

PR #485, opened 2017-09-21
Fix typo and CRDs

andraxylia:


_docs/setup/install-kubernetes.md
```diff
@@ -104,7 +104,7 @@ and can manage micro-services from all other namespaces.
 
    Both options create the `istio-system` namespace along with the required RBAC permissions, and deploy Istio-Pilot, Istio-Mixer, Istio-Ingress, Istio-Egress, and Istio-CA (Certificate Authority).
 
-1. *Optional:* If you cluster has Kubernetes alpha features enabled, and you wish to enable transparent injection of sidecar, please install the Istio-Initializer:
+1. *Optional:* If your cluster has Kubernetes alpha features enabled, and you wish to enable transparent injection of sidecar, please install the Istio-Initializer:
```
ldemailly: I think lin suggested to have a link to which alpha feature we need

rshriram: Yes please. This sentence is otherwise a bit vague. Might be worth while to have a link to k8s docs or describe what you mean by alpha features..
Say something like, if you have enabled Kubernetes external admission controllers (alpha feature), ...

andraxylia: This is not a good idea, it could lead the user into the confusion they need to check if a particular alpha feature is needed.

This is an installation guide and should be kept simple, people interested into details can read the design docs.

PR #488, opened 2017-09-21
Install tweaks 2

ldemailly:
Install tweak: Keep the the upgrade note short and with a link to 0.1
uninstall


ldemailly:
before:
screen shot 2017-09-21 at 1 28 43 pm
after:
screen shot 2017-09-21 at 1 28 58 pm

_docs/setup/install-kubernetes.md
```diff
@@ -18,10 +18,10 @@ Quick Start instructions to install and configure Istio in a Kubernetes cluster.
...
 with [RBAC (Role-Based Access Control)](https://kubernetes.io/docs/admin/authorization/rbac/) enabled.
 
-  > Note: Istio 0.2.x is not backward compatible with 0.1.x. Uninstall any 0.1.x
-  > version of Istio before installing the newer version. Any applications
-  > using Istio sidecar from 0.1.x version must upgrade to using the
-  > sidecar supplied by 0.2.x release.
+  > Note: If you installed Istio 0.1.x please
+  > [uninstall](https://istio.io/v-0.1/docs/tasks/installing-istio.html#uninstalling)
+  > it completely before installing the newer version (including the Istio sidecar
+  > for all Istio enabled application pods).
```
kyessenov: does this apply to previous 0.2.x release too? i suspect it does.

_docs/setup/install-kubernetes.md
```diff
@@ -18,10 +18,10 @@ Quick Start instructions to install and configure Istio in a Kubernetes cluster.
 The following instructions assume you have access to a Kubernetes **1.7.4 or newer** cluster
 with [RBAC (Role-Based Access Control)](https://kubernetes.io/docs/admin/authorization/rbac/) enabled.
 
-  > Note: Istio 0.2.x is not backward compatible with 0.1.x. Uninstall any 0.1.x
-  > version of Istio before installing the newer version. Any applications
-  > using Istio sidecar from 0.1.x version must upgrade to using the
-  > sidecar supplied by 0.2.x release.
+  > Note: If you installed Istio 0.1.x please
```
rshriram: Remove please.

Rachael-Graham: remove please: "... 0.1.x, uninstall ..."

_docs/setup/install-kubernetes.md
rshriram: This is also conveying the impression that istio is not backward compatible for future cases as well. That's why I qualified my statement earlier saying it applies only to 0.1 to 0.2 but not from 0.2.x. Your call.

ldemailly: yes but I don't want to muddy it with non released release - if someone used master they would hopefully guess that they need a full cleanup before reporting an issue ?

ZackButcher: +1, optimize for post release, since people will only have issues with 0.2.x versions that we're producing right now before we've actually officially released 0.2. I think we only need to talk about 0.1 -> 0.2.

ldemailly: the sentence actually just mentions what to do with 0.1 (and this being 0.2, it means 0.1->0.2)
it also avoids the negative "Istio 0.2.x is not backward compatible with 0.1.x" (though if we want to get technical, 0.2.4 is incompatible with 0.2.2 but we warned about that in the prerelase notes...)

_docs/setup/install-kubernetes.md
```diff
@@ -18,10 +18,10 @@ Quick Start instructions to install and configure Istio in a Kubernetes cluster.
...
+  > [uninstall](https://istio.io/v-0.1/docs/tasks/installing-istio.html#uninstalling)
+  > it completely before installing the newer version (including the Istio sidecar
```
Rachael-Graham: phrase in parentheses is vague--does this mean 'you need to uninstall the sidecar too', or 'you need to uninstall Istio before installing the newer version and/or the sidecar'

_docs/setup/install-kubernetes.md
Rachael-Graham: ha, sry bout that having problems on my end

ldemailly: you had another comment about the part in parenthesis, but I lost it - can you repeat it ?

Rachael-Graham: yep: does phrase in parentheses mean 'uninstall the sidecar too', or 'uninstall before you install the newer version and/or the sidecar'

ldemailly:
@rshriram PTAL

PR #491, opened 2017-09-21
Simplify and make reproducible the website setup

ldemailly:
We still need to make an image so it doesn’t take forever but at least
this is reproducible and working


README.md
```diff
@@ -13,51 +13,26 @@ familiar with these before editing.
...
-```bash
-docker run --rm --label=jekyll --volume=$(pwd):/srv/jekyll  -it  jekyll/jekyll sh -c "bundle install && rake test"
-```
+The `rake test` part is to make sure you are not introducing html errors or bad links, you should see
 ```
 ```
+in the output
 
-#### Side note for those on non-linux machines
```
ldemailly: I don't think we need to tell people how to install docker and the minikube part is just noise for this docs repo

README.md
```diff
@@ -13,51 +13,26 @@ familiar with these before editing.
 To run the site locally with Docker, use the following command:
 
 ```bash
-docker run --rm --label=jekyll --volume=$(pwd):/srv/jekyll  -it -p 127.0.0.1:4000:4000 jekyll/jekyll sh -c "bundle install && jekyll serve"
+docker run --label=jekyll --volume=$(pwd):/srv/jekyll  -it -p 127.0.0.1:4000:4000 jekyll/jekyll:3.5.2 sh -c "bundle install && rake test && jekyll serve"
```
rshriram: Why did you remove the --rm ?

_config.yml
```diff
@@ -61,3 +61,5 @@ exclude:
   - repos/*.html
   - repos/*.md
 
+repository:
+  istio/istio.github.io
```
rshriram: What's this for?

ldemailly: this is to avoid some NGO liquid error @kyessenov was getting because his git origin wasn't set

kyessenov: breaks on my machine without it. although i don't know if it's ok to check-in.

ldemailly:
@rshriram see (and test !) the new version - it works well for me, subsequent start are much faster as they don't reinstall the gems only check them

ldemailly:
@andraxylia you had issues with jekyll can you please try these updated instructions - it should work out of the box (slow for first time and fast for 2nd onward)

PR #492, opened 2017-09-21
WIP: early version of raw vm installation

costinm:
Please comment or/and update (or merge into your doc) - I copied from the original doc, but updated to use the new scripts, and tried to make it a bit more modular.

### DIFF
--- a/_docs/setup/install-extensions.md
+++ b/_docs/setup/install-extensions.md
@@ -24,19 +24,24 @@ _This document is under construction._
 
 ## Installation steps
 
-Example functions to help with the setup are available in install/tools/istio_vm_common.sh -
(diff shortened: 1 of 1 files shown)


ldemailly:
Would be nice to have started from the branch I created for that purpose on tuesday
https://github.com/istio/istio.github.io/tree/cluster-ext
though it needed some conflict resolution that I just did (you can ask if that was the issue) but we can merge the 2

I'll take your changes and put them there and then add my comments if that works ?

ldemailly:
let's continue this work in
https://github.com/istio/istio.github.io/pull/493

ldemailly:
see https://github.com/istio/istio.github.io/pull/493#issuecomment-331326224

PR #496, opened 2017-09-22
Authentication subsection renamed to Security

rshriram:


rshriram:
Also moving basic access control to security.

PR #497, opened 2017-09-22
document egress rules

frankbu:


_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -47,14 +47,15 @@ deployment using the `kubectl` command instead. See the
 [configuring request routing task]({{home}}/docs/tasks/request-routing.html) for
 examples.
 
-There are two types of rules in Istio, **Route Rules** and **Destination
-Policies** (these are not the same as Mixer policies). Both types of rules
-control how requests are routed to a destination service.
+There are three kinds of traffic management rules in Istio, **Route Rules**, **Destination
```
Rachael-Graham: "...rules in Istio: **Route .... "

_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -47,14 +47,15 @@ deployment using the `kubectl` command instead. See the
...
+Policies** (these are not the same as Mixer policies), and **Egress Rules**. All three
+kinds of rules control how requests are routed to a destination service.
 
 ## Route Rules
 
-Route rules control how requests are routed to different versions of a
-service. Requests can be routed based on the source and destination, HTTP
+Route rules control how requests are routed within an Istio service mesh,
+for example, to different versions of a service.
```
Rachael-Graham: ".... within an Istio service mesh. For example, a route rule could route requests to different versions of a service."

_docs/tasks/traffic-management/egress.md
```diff
@@ -91,27 +84,86 @@ HTTP or port `443` for HTTPS.
...
 
-3. For external services of type HTTPS, the port must be specified in the request.
-   App clients should make the request over HTTP since the Egress Envoy will initiate HTTPS 
-   with the external service:
+1. Wait a few seconds, then issue the _curl_ request again:
 
    ```bash
-   curl http://securegoogle:443
+   kubectl exec -it $SOURCE_POD -c sleep bash
+   time curl -o /dev/null -s -w "%{http_code}\n" http://httpbin.org/delay/5
```
rshriram: Make the delay larger so that people don't see the 3 vs 5 as noise. Like use httpbin.org/delay/7 for example.

_docs/concepts/traffic-management/rules-configuration.md
```diff
@@ -554,3 +555,61 @@ without any rules being set, as soon as version discrimination is desired
...
+      protocol: http
+    - port: 443
+      protocol: https
+```
+
+The destination of an egress rule is specified using the *service* field, which
+can be either a fully qualified or wildcard domain name.
+It represents a white listed set of one or more external services that services
+in the mesh are allowed to access. The supported wildcard syntax can be found
+[here]({{home}}/docs/reference/config/traffic-rules/egress-rules.html).
```
rshriram: nit.. this is egress-rule.html

_docs/tasks/traffic-management/egress.md
frankbu: Using `time curl ...` as the command and printing the http response, so the difference is very clear in the output when you run it (esp the 504 vs 200 response). I think waiting 5 secs for the command to return is already annoying enough :-)

rshriram:
Only error is 
```
  *  internally linking to /docs/reference/config/traffic-rules/egress-rule.html, which does not exist (line 253)
     <a href="/docs/reference/config/traffic-rules/egress-rule.html">here</a>
```
which is fine because we don't have any reference docs. And this is the reference URL.

PR #504, opened 2017-09-23
fixing the crd delete

ldemailly:
jsonpath outputs space separated, we need the grep to apply per line instead

using fgrep avoids need to escape the dots


ldemailly:
are pilot rules also labelled with istio ?

kyessenov:
No they are not. The labeling is ad-hoc.

PR #509, opened 2017-09-25
Setup cleanup

rshriram:
image


_docs/index.md
```diff
@@ -10,9 +10,7 @@ type: markdown
 ---
 {% include home.html %}
 
-Follow this link for the archive of [most recent stable release (0.1) documentation](https://istio.io/v-0.1/docs/).
```
ldemailly: why was this removed; 0.2 isn't yet stable

PR #511, opened 2017-09-25
update bookinfo.md

frankbu:


_docs/guides/bookinfo.md
```diff
@@ -45,34 +39,63 @@ The end-to-end architecture of the application is shown below.
...
+<figcaption>BookInfo Application</figcaption></figure>
 
-## Start the application
+All of the microservices will be packaged with an Envoy sidecar that intercepts incoming
+and outgoing calls for the services, providing the hooks needed to externally control,
+via the Istio control plane, routing, telemetry collection, and policy enforcement
+for the application as a whole.
+
+Before you begin, if you haven't already done so, setup Istio by following the instructions
+corresponding to your platform [installation guide]({{home}}/docs/setup/) and then
```
ldemailly: link directly https://istio.io/docs/setup/kubernetes/quick-start.html
(this doc is already k8s scoped, no need to have users need 2 more clicks to get to it)

_docs/guides/bookinfo.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Bookinfo on Kubernetes
```
ldemailly: Let's keep "on Kubernetes" or drop the consul one

frankbu: Direction proposed by this PR is the latter. This guide just describes the sample and gives basic run/stop instructions for all platforms.

ldemailly: Not sure it will flow well to not specify the platform but if you can pull it off, sounds good

removal of "Bookinfo with Consul" should be in this same PR then

ldemailly:
are you doing a follow up ? as mentioned above I'd like the consul part to be settled / right now it looks off

PR #519, opened 2017-09-25
fix download url

ldemailly:
uses the downloader that downloads 0.2.4


_docs/setup/kubernetes/quick-start.md
```diff
@@ -65,9 +65,9 @@ Starting with the {{ site.data.istio.version }} release, Istio is installed in i
 
  1. Go to the [Istio release](https://github.com/istio/istio/releases) page to download the
     installation file corresponding to your OS. If you are using a MacOS or Linux system, you can also
-    run the following command to download and extract the latest stable release automatically:
```
costinm: We still need to leave the old script documented. 

Otherwise - remove it ( but than why do we need 2 ) ? 

Long term - we should stop doing this, the script just downloads a zip file - and adds security risks.

ldemailly: we will keep 2, 1 stable, 1 unstable, both are used - until we have a better (more secure) solution

PR #524, opened 2017-09-25
Add "Istio Addons" Task

douglas-reid:
This moves the content out from the installation guide into a distinct Task. This might make more sense to divide into 3 tasks, or to name more conceptually, but this provides a first stab at content.

This PR also cleans up the bad `istioctl` command from metrics-logs.md.

There should be a follow-on PR to add a link from metrics-logs.md to this task. Pending outcome of this PR.


_docs/tasks/istio-addons.md
```diff
@@ -0,0 +1,257 @@
...
+include a Prometheus server with config to pull metrics from Istio endpoints, a
+Grafana server preconfigured with a default dashboard for monitoring an Istio
+service mesh, and an example ServiceGraph service that can generate a visual
+representation of an Istio service mesh.
+
+The [BookInfo]({{home}}/docs/samples/bookinfo.html) sample application is used
+as the example application throughout this task.
+
+## Before you begin
+* [Install Istio](./installing-istio.html) in your cluster and deploy an
```
ldemailly: remove ./ ?

_docs/tasks/istio-addons.md
```diff
@@ -0,0 +1,257 @@
+---
```
ldemailly: not sure if github would do anything better but could you cp + commit + edit in an attempt to see a diff instead of full green ?

douglas-reid: A diff of what?  The removed stuff ?

This is a fully new doc, so it would just be a red block followed by this long green block.

_docs/tasks/istio-addons.md
douglas-reid: good catch. this clued me into the fact that we fundamentally changed the structure of the docs repo recently.

_docs/tasks/istio-addons.md
ldemailly: oh there is nothing common between these instructions and the ones that were in quick-start.md ?

_docs/tasks/istio-addons.md
ldemailly: yeah it keeps changing...  make sure "rake test" is happy and/or navigation works when using your local jekyll

ditto for how the images render

PR #526, opened 2017-09-25
Fix istioctl create command

mandarjog:
1. Fix create command
2. Remove extra `---` at the end of the document.


mandarjog:
Addresses https://github.com/istio/issues/issues/76

PR #528, opened 2017-09-26
switch version from pre0.2 to 0.2.6

ldemailly:
we can consider maybe just "0.2" too once we get into weekly cadence


ldemailly:
this is expecting we cut 0.2.5 soon

ldemailly:
note: do not merge before https://github.com/istio/istio/releases
shows 0.2.5

PR #529, opened 2017-09-26
have a more direct link to download and install

ldemailly:
like in 0.1


_docs/index.md
```diff
@@ -22,6 +22,8 @@ is where you can learn about what Istio does and how it does it.
   the Istio control plane in various environments (e.g., Kubernetes, Consul,
   etc.), as well as instructions for installing the sidecar in the
   application deployment.
+  Jump directly to [downloading and installing](https://istio.io/docs/setup/kubernetes/quick-start.html) Istio for Kubernetes,
+  if it is your platform of choice.
```
ldemailly: happy to change the english around it as long as there is a directly link to download & install (for k8s)

ldemailly:
also adding back (but now at the bottom and without mention of stable) the link to 0.1 archive of docs

PR #550, opened 2017-09-27
first round of faq docs for traffic management task

GregHanson:


rshriram:
cc @kyessenov and @ijsnellf to pile on other limitations.

PR #551, opened 2017-09-27
use istio-system as default namespace

linsun:
as istio.yaml indicated for mixer deployment config.


rshriram:
@douglas-reid I remember you objecting to this change in istio/istio when @andraxylia made a similar change. I haven't kept up. So please comment/approve.

PR #558, opened 2017-09-28
crd have moved to istio-system namespace since #909

markeijsermans:
istio-config-default namespace no longer exists in the standard kube install manifests

referenced PR: https://github.com/istio/istio/pull/909


rshriram:
@douglas-reid are these changes valid?

PR #559, opened 2017-09-28
fix kubectl get svc istio-ingress commands

tomgco:
This addresses the issue pointed out in https://github.com/istio/issues/issues/67

### DIFF
--- a/_docs/guides/bookinfo.md
+++ b/_docs/guides/bookinfo.md
@@ -155,14 +155,14 @@ run the application using the instructions in one of the following sections.
    > If, however, the cluster has a firewall, you will also need to create a firewall rule to allow TCP traffic to the NodePort.
    > In GKE, for instance, you can create a firewall rule using the following command:
    > ```bash
-   > gcloud compute firewall-rules create allow-book --allow tcp:$(kubectl get svc istio-ingress -o jsonpath='{.spec.ports[0].nodePort}')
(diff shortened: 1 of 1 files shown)


_docs/guides/bookinfo.md
```diff
@@ -155,14 +155,14 @@ run the application using the instructions in one of the following sections.
...
-   > gcloud compute firewall-rules create allow-book --allow tcp:$(kubectl get svc istio-ingress -o jsonpath='{.spec.ports[0].nodePort}')
+   > gcloud compute firewall-rules create allow-book --allow tcp:$(kubectl --namespace istio-system get svc istio-ingress -o jsonpath='{.spec.ports[0].nodePort}')
    > ```
 
    If your deployment environment does not support external load balancers (e.g., minikube), the ADDRESS field will be empty.
    In this case you can use the service NodePort instead:
    
    ```bash
-   export GATEWAY_URL=$(kubectl get po -n istio-system -l istio=ingress -o 'jsonpath={.items[0].status.hostIP}'):$(kubectl get svc istio-ingress -n istio-system -o 'jsonpath={.spec.ports[0].nodePort}')
+   export GATEWAY_URL=$(kubectl --namespace istio-system get po -n istio-system -l istio=ingress -o 'jsonpath={.items[0].status.hostIP}'):$(kubectl --namespace istio-system get svc istio-ingress -n istio-system -o 'jsonpath={.spec.ports[0].nodePort}')
```
gyliu513: We already have `-n istio-system` so no need to add `--namespace istio-system` again

PR #562, opened 2017-09-28
vm guide page updates

ldemailly:
also fixed typo in link


frankbu:
See https://github.com/istio/istio.github.io/pull/563. I added more disclaimers. I think it's good to at least show people what's coming.

PR #566, opened 2017-09-28
Clarification of enable/disable mTLS

andraxylia:


_docs/tasks/security/faq.md
```diff
@@ -9,17 +9,20 @@ type: markdown
 ---
 {% include home.html %}
 
+* _How can I enable/disable mTLS encryption after I installed Istio?_
+
+  Enabling/disabling mTLS requires uninstalling and installing Istio,
+  as well as re-deploying your applications.
```
ldemailly: you don't need to redeploy do you ?

ldemailly: Let's put something like this as the answer:

```
kubectl edit configmap -n istio-system istio
```
comment out or uncomment out `authPolicy: MUTUAL_TLS`
and then
```
kubectl delete pods -n istio-system -l istio=pilot
```
to restart pilot, after a few seconds (depending on your `*RefreshDelay`) your envoy will have picked up the change

andraxylia: I removed the requirement to re-deploy the apps, ptal.

ldemailly: can you add the above in a "short cut" or "advanced users" section
(you asked me if I personally verified it works; I did spend the time to do so, it'd be nice to have it there)

andraxylia: Agree, let me add it an a separate PR, since this one got merged.

ldemailly: thanks!

PR #569, opened 2017-09-29
Create cluster-admin rolebinding

andraxylia:
Added instructions specific to GKE.

### DIFF
--- a/_docs/setup/kubernetes/quick-start.md
+++ b/_docs/setup/kubernetes/quick-start.md
@@ -17,7 +17,7 @@ Quick Start instructions to install and configure Istio in a Kubernetes cluster.
 
 The following instructions assume you have access to a Kubernetes **1.7.4 or newer** cluster
 with [RBAC (Role-Based Access Control)](https://kubernetes.io/docs/admin/authorization/rbac/) enabled.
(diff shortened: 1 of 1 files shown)


_docs/setup/kubernetes/quick-start.md
```diff
@@ -28,24 +28,30 @@ If you wish to enable [transparent injection of sidecar]({{home}}/docs/setup/kub
...
     ```
-    Replace `<cluster-name>` with the name of the cluster you want to use and `<zone>`
-    with the zone that cluster is in.
+    where `<cluster-name>` is the name of the cluster you want to use, and `<zone>` is the zone that cluster is in.
 
-  * If you are using [IBM Bluemix Container Service](https://www.ibm.com/cloud-computing/bluemix/containers), retrieve your
-    credentials for kubectl:
+    2. Grant cluster admin permissions to the current user - admin permissions are required to create the necessary RBAC rules for Istio:
+    ```bash
+        kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=$(gcloud config get-value core/account)
```
rshriram: this seems overly indented.. shouldn't it start from the same column as backtick?

andraxylia: yup, the indentation is wrong everywhere.

ldemailly: looks fine everywhere except here as far as I can see

ldemailly:
merged so I can close
https://github.com/istio/issues/issues/4

PR #571, opened 2017-09-29
Change proxy to istio-proxy on tasks description

ayj:
Fixes https://github.com/istio/issues/issues/70

### DIFF
--- a/_docs/setup/kubernetes/manual-sidecar-inject.md
+++ b/_docs/setup/kubernetes/manual-sidecar-inject.md
@@ -201,7 +201,7 @@ applications which are redirected to proxy.
           apiVersion: v1
           fieldPath: metadata.namespace
     -
-      name: INSTANCE_IP
+      name: POD_IP
       valueFrom:
         fieldRef:
           apiVersion: v1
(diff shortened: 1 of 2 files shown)


_docs/setup/kubernetes/manual-sidecar-inject.md
```diff
@@ -181,17 +201,23 @@ applications which are redirected to proxy.
           apiVersion: v1
           fieldPath: metadata.namespace
     -
-      name: POD_IP
+      name: INSTANCE_IP
```
andraxylia: in the future we can rename this to POD_IP, it seems to be an unwritten standard.

_docs/setup/kubernetes/manual-sidecar-inject.md
```diff
@@ -204,20 +230,22 @@ traffic to the proxy. An init-container is used for two reasons:
...
-     }
-   },
+   ```yaml
+      initContainers:
+      - args:
+        - -p
+        - "15001"
+        - -u
+        - "1337"
+        image: docker.io/istio/proxy_init:unittest
```
andraxylia: unittest?

_docs/setup/kubernetes/manual-sidecar-inject.md
ayj: Changed to reflect the actual value used [here](https://github.com/istio/pilot/blob/master/platform/kube/inject/inject.go#L426)

PR #573, opened 2017-09-29
Update the docs with new template docs

guptasu:


rshriram:
Are these changes also present in istio/api ? Because the next time we regenerate these docs, all these changes will get blown away.

_docs/reference/config/mixer/template/listentry.md
```diff
@@ -20,6 +20,22 @@ type: markdown
 ListEntry is used to verify the presence/absence of a string
 within a list.
 
+When writing the configuration, the value for the fields associated with this template can either be a
+literal or an [expression](https://istio.io/docs/reference/config/mixer/expression-language.html). Please note that if the datatype of a field is not istio.mixer.v1.config.descriptor.ValueType,
```
Rachael-Graham: istio.mixer.v1.config.descriptor.ValueType should probably be in code--same for every instance of this

guptasu:
@rshriram Yes. These changed are generated from mixer repo protos, and they are already checked in.

_docs/reference/config/mixer/template/listentry.md
guptasu: Hi @Rachael-Graham , if you don't feel strongly about it, I would prefer to leave this for now. The documentation is generated from proto files that are checked in inside the mixer repo and given the code freeze, it will be extremely hard now to submit a PR in mixer that just adds some `code` tags.

Let me know, if you strongly feel for making this change, I can manually edit thse md file for the time being and post 0.2 I can edit the protos.

ZackButcher:
Just edit by hand. Next release, tools will be open source and we can re-evaluate, but atm lets just hand edit to get the docs where we need them to be for 0.2 release.

There's no regular process that regenerates the docs or anything, so there's no risk of manual edits being overwritten.

_docs/reference/config/mixer/template/listentry.md
Rachael-Graham: Nope sounds good to me--definitely can leave it for now. a very minor change!

PR #592, opened 2017-10-02
Removing stray template doc file. This was causing a bad link in the index file

guptasu:


guptasu:
@douglas-reid 's PR is fixing this issue along with other issues.

PR #596, opened 2017-10-03
removed unnecessary step and bug fix

GregHanson:
bug fix for a doc link
added hardocded kubeconfig file, so can simplify the docs to only one `context-create` command


linsun:
Thanks for catching this, Greg!

PR #609, opened 2017-10-04
Improve the instructions on access control

myidpt:


rshriram:
@myidpt please rebase. i renamed the istio-auth.md to mutual-tls.md and the auth document to mutual-tls.md to correctly reflect things.. Also network-and-auth renamed to security, so that it matches with the rest of tasks, etc. 

I also cleaned up all mention of auth and used mutual TLS Authentication for a clearer explanation everywhere.

myidpt:
@rshriram Got it. Rebased.

PR #611, opened 2017-10-04
Add initial FAQ for telemetry

douglas-reid:
Here is a first-pass stab at FAQ entries vaguely related to Mixer. The questions here are based on my experience of answering questions via Slack, email, and in-person. There are surely more questions and answers that can be included here, but I think this is a decent start.


_docs/tasks/telemetry/faq.md
```diff
@@ -9,3 +9,178 @@ type: markdown
 ---
 {% include home.html %}
```
geeknoid: As discussed earlier:

- This should be part of the one-and-only FAQ. We can create different sections in this FAQ if needed. I don't think we want distributed FAQs within the site.

- Whether or not we keep this FAQ separate, we should follow the same model as the main FAQ. That is, one file per question which we then combine together via Jekyll voodoo. faq/index.md and _faq/* is how this can be done. I can help set this up if we want to keep distinct FAQ pages.

douglas-reid: @rshriram I believe you added the individual FAQ pages. Are you fine with the move to _faq ?

@geeknoid I was just trying to add content first and worry about any refactoring later. I'm fine moving this content to wherever now that it exists.

geeknoid:
​Oh, I hadn't noticed all the other tasks also have similar FAQs.

A "Tasks" section, which is where you go to get step-by-step instructions,
is not the right place for FAQs in general.​

rshriram:
Yeah I added the faq per section in tasks. The big faq seemed to be buried deep in the webpage. And it was pretty high level. Some section specific questions could not be put there - the ux flow involves someone searching for the big faq and then skimming through the entire thing to find out the related faq. 

A simple example is “why isn’t each version being hit uniformly”? This kind of stuff makes sense under the traffic Management faq, where someone is playing with route rules and finds something not working. They go to the docs (typically traffic Management section, or occasionally the reference), don’t find what they are looking for. Now imagine that person having to look for the faq link in istio, (in a different part of the site altogether, and skimming the list to find what they want). In contrast, a curated list of gotchas, common issues, limitations, etc is available right in the section specific faq where the user is already present. 

I was hoping that we would pile up a lot of faqs, config snippets for commonly seen how-do-i?, etc. faq is bad naming but seemed the closest.

I have no preference with regard to using the Jekyll voodoo to generate faq automatically in each section instead of using faq.md.

geeknoid:
Buried? FAQ is a top-level link at the top of each page on the site, right
next to the search box.

Would you be happy with multiple sections in the main FAQ page?

rshriram:
I didn't notice the top level FAQ at all! So, what about the other point that I had? [FAQ is badly named here]. Its more of a kitchen sink, "here are all the other nuances, tips, tricks that you might find handy/encounter in your journey with Istio". I had intended all this to go into the (badly named)FAQ page.. That would include code snippets, troubleshooting instructions, some best practices here and there, etc.

geeknoid:
So how about we do this:

- Introduce sections into the main FAQ page so we can partition the
content. We can then move some of the FAQ-ish material from the various
tasks/FAQ files there.

- We rename the tasks/FAQ files to simply "XXX Insights". So for example,
in the traffic management tasks, the page would be "Traffic Management
Insights"

How's that?

rshriram:
Sure that sounds good by me.. Insights is a bit too fancy for the kinds of things we want to put in that page.. ;) .. 
May be something like "Notes" ? or Additional Information or something?

geeknoid:
"More on Traffic Management"
"Traffic Management Tips and Tricks"

?

douglas-reid:
So, should I move this content to the global FAQ (adopting its format)?

geeknoid:
How about you go ahead with this checkin, and I'll do a clean on all those
tasks FAQ files.

PR #612, opened 2017-10-05
Add automatic injection to release-notes

andraxylia:


linsun:
good catch, i was thinking this one is missed earlier but forgot to point it out.

PR #617, opened 2017-10-05
minor issue found during testing

linsun:
and some clarifications


_docs/tasks/traffic-management/traffic-shifting.md
```diff
@@ -60,7 +60,7 @@ two steps: 50%, 100%.
 1. Refresh the `productpage` in your browser and you should now see *red* colored star ratings approximately 50% of the time.
 
    > Note: With the current Envoy sidecar implementation, you may need to refresh the `productpage` very many times
-   > to see the proper distribution. You can modify the rules to route 90% of the traffic to v3 to see red stars more often.
+   > to see the proper distribution. You may see all traffic goes to v1 in the initial ~15 refreshes. You can modify the rules to route 90% of the traffic to v3 to see red stars more often.
```
frankbu: How about this:

    > Note: With the current Envoy sidecar implementation, you may need to refresh the `productpage` very many times
   > to see the proper distribution. It may require 15 refreshes or more before you see any change. You can modify the rules to route 90% of the traffic to v3 to see red stars more often.

ldemailly: technically it's our configuration not the implementation https://github.com/envoyproxy/envoy/issues/1719

PR #626, opened 2017-10-05
Small tweaks for Security

wattli:


_docs/tasks/security/mutual-tls.md
```diff
@@ -92,7 +92,7 @@ There are several steps:
    
 1. send requests to another service, for example, details.
    ```bash
-   curl https://details:9080 -v --key /etc/certs/key.pem --cert /etc/certs/cert-chain.pem --cacert /etc/certs/root-cert.pem -k
+   curl https://productpage:9080 -v --key /etc/certs/key.pem --cert /etc/certs/cert-chain.pem --cacert /etc/certs/root-cert.pem -k
```
wattli: Currently somehow only productpage works, so this PR is for temporary workaround. I am looking into the problem.

_docs/tasks/security/mutual-tls.md
```diff
@@ -59,25 +59,25 @@ Istio CA is up if the "AVAILABLE" column is 1.
 When running Istio with mutual TLS authentication turned on, you can use curl in one service's
 envoy to send request to other services.
 For example, after starting the [BookInfo]({{home}}/docs/guides/bookinfo.html) 
-sample application you can ssh into the envoy container of `productpage` service, 
+sample application you can ssh into the envoy container of `ratings` service,
```
ldemailly: curious why the change ? should work the same no ?

_docs/tasks/security/mutual-tls.md
ldemailly: that seems wrong - makes it seem like it needs the ingress ?

what happens exactly when using another svc ?

_docs/tasks/security/mutual-tls.md
wattli: See my previous comment in this thread: "Currently somehow only productpage works, so this PR is for temporary workaround. I am looking into the problem."

wattli: It's reverted, and turns out the endpoint changes.

wattli:
@douglas-reid , this fixes the auth issue you found this morning.

Issue #637, opened 2017-10-06
Issue with _docs/reference/config/mixer/expression-language.md


ldemailly:
== is documented as supporting prefix/suffix matches with `*` when in fact one needs to use `match()` for that

ozevren:
This doesn't seem to be an issue anymore.

ldemailly:
thx

Issue #610, opened 2017-10-04
Update CEXL doc to include extension fns


douglas-reid:
I feel like we should document `ip()` and `match()` (and whatever else I've forgotten) somewhere.

geeknoid:
Can this be closed?

mandarjog:
yes.

PR #638, opened 2017-10-06
Add match and ip doc to CEXL

mandarjog:
Addresses #637 and #610


ldemailly:
I guess the answer is no ? only prefix/suffix?

Issue #640, opened 2017-10-07
Istio does not install correctly on Kubernetes 1.7.4 using instructions in Kubernetes > Quick Start


danifitz:
Hi,

I was following https://istio.io/docs/setup/kubernetes/quick-start.html to install Istio on IBM Bluemix Container Service (k8s v1.7.4_1502).

When it says to run `kubectl apply -f install/kubernetes/istio.yaml` I find that I get the following error:
```
namespace "istio-system" created
configmap "istio-mixer" created
service "istio-mixer" created
serviceaccount "istio-mixer-service-account" created
deployment "istio-mixer" created
configmap "istio" created
service "istio-pilot" created
serviceaccount "istio-pilot-service-account" created
deployment "istio-pilot" created
service "istio-ingress" created
serviceaccount "istio-ingress-service-account" created
deployment "istio-ingress" created
service "istio-egress" created
serviceaccount "istio-egress-service-account" created
deployment "istio-egress" created
serviceaccount "istio-ca-service-account" created
deployment "istio-ca" created
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRole" is registered for version "rbac.authorization.k8s.io/v1beta1"
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRole" is registered for version "rbac.authorization.k8s.io/v1beta1"
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRole" is registered for version "rbac.authorization.k8s.io/v1beta1"
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRole" is registered for version "rbac.authorization.k8s.io/v1beta1"
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRole" is registered for version "rbac.authorization.k8s.io/v1beta1"
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRoleBinding" is registered for version "rbac.authorization.k8s.io/v1beta1"
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRoleBinding" is registered for version "rbac.authorization.k8s.io/v1beta1"
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRoleBinding" is registered for version "rbac.authorization.k8s.io/v1beta1"
unable to decode "install/kubernetes/istio.yaml": no kind "ClusterRoleBinding" is registered for version "rbac.authorization.k8s.io/v1beta1"
[...]

PR #641, opened 2017-10-09
Istio does not install correctly on Kubernetes 1.7.4

linsun:
#640:  https://github.com/istio/istio.github.io/issues/640


ldemailly:
thanks though this was already there below 
"Install or upgrade the Kubernetes CLI kubectl to match the version supported by your cluster (version 1.7 or later for CRD support)."

but maybe people will read the top one :-)

PR #645, opened 2017-10-09
adding known issues

linsun:
https://github.com/istio/istio.github.io/issues/643

### DIFF
--- a/_docs/reference/release-notes.md
+++ b/_docs/reference/release-notes.md
@@ -88,5 +88,10 @@ identity provisioning. This agent runs on each node (VM / physical machine) and 
 - **Persistent CA Key/Certificate Storage**. Istio CA now stores signing key/certificates in
 persistent storage to facilitate CA restarts. 
 
+### Known Issues
+
+- **User may get periodical 404 when accessing the application**:  We have noticed that envoy doesn't get routes properly occasionally thus a 404 is returned to the user.  We are actively working on this [issue](https://github.com/istio/istio/issues/1038).
+- **Pilot reports ready before Pilot is actually ready**: You can check pods status in the `istio-system` namespace and wait a few seconds after all Istio Pods reach ready status.  We are actively working on this [issue](https://github.com/istio/istio/pull/1055).
+
 
 [Release notes for previous versions of Istio](./notes)
--- a/_docs/reference/release-notes.md
+++ b/_docs/reference/release-notes.md
@@ -90,8 +90,9 @@ persistent storage to facilitate CA restarts.
 
 ### Known Issues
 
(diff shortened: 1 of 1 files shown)


_docs/reference/release-notes.md
```diff
@@ -88,5 +88,10 @@ identity provisioning. This agent runs on each node (VM / physical machine) and
 - **Persistent CA Key/Certificate Storage**. Istio CA now stores signing key/certificates in
 persistent storage to facilitate CA restarts. 
 
+### Known Issues
+
+- **User may get periodical 404 when accessing the application**:  We have noticed that envoy doesn't get routes properly occasionally thus a 404 is returned to the user.  We are actively working on this [issue](https://github.com/istio/istio/issues/1038).
+- **Pilot reports ready before Pilot is actually ready**: You can check pods status in the `istio-system` namespace and wait a few seconds after all Istio Pods reach ready status.  We are actively working on this [issue](https://github.com/istio/istio/pull/1055).
```
ldemailly: replace Pilot by Istio-Ingress ?

_docs/reference/release-notes.md
```diff
@@ -88,5 +88,10 @@ identity provisioning. This agent runs on each node (VM / physical machine) and
 - **Persistent CA Key/Certificate Storage**. Istio CA now stores signing key/certificates in
 persistent storage to facilitate CA restarts. 
 
+### Known Issues
```
ldemailly: maybe we should also list here that with auth enabled you can't talk to k8s api server or non istio services
(and that for health checks you need to use a custom curl command)

_docs/reference/release-notes.md
```diff
@@ -88,5 +88,10 @@ identity provisioning. This agent runs on each node (VM / physical machine) and
...
+
+- **User may get periodical 404 when accessing the application**:  We have noticed that envoy doesn't get routes properly occasionally thus a 404 is returned to the user.  We are actively working on this [issue](https://github.com/istio/istio/issues/1038).
```
geeknoid: envoy -> Envoy

_docs/reference/release-notes.md
geeknoid: pods status -> pod status
Istio Pods -> Istio pods
ready status -> ready state

_docs/reference/release-notes.md
linsun: Thanks.

_docs/reference/release-notes.md
linsun: these are already documented in security FAQ, but I did add non istio service as that is pretty major.  talking to K8S API server isn't that common...

_docs/reference/release-notes.md
linsun: updated.

linsun: updated.  I kept status as that is what kubectl output shows when I call `kubectl get pods`

PR #663, opened 2017-10-13
Fix minor doc issues re: Sidecar Injection.

TimDumol:
Handles https://github.com/istio/issues/issues/75.

* Tweak description of how `istio-initializer` behaves.
* Fix wrong initializer name (`status.sidecar.istio.io`).

### DIFF
--- a/_docs/setup/kubernetes/sidecar-injection.md
+++ b/_docs/setup/kubernetes/sidecar-injection.md
@@ -233,8 +233,8 @@ It finds its configured name `sidecar.initializer.istio.io`
 as the first in the list of pending initializers.
 
 3) istio-initializer checks to see if it was responsible for
-initializing the namespace of the workload. No further work is done
-and the initializer ignores the workload if the initializer is
+initializing workloads in the namespace of the workload. No further work is done
+and the initializer ignores the workload if the initializer is not
 configured for the namespace. By default the initializer is
 responsible for all namespaces (see [configuration options](#configuration-options)).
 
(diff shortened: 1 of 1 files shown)


TimDumol:
I signed it!

_docs/setup/kubernetes/sidecar-injection.md
```diff
@@ -233,8 +233,8 @@ It finds its configured name `sidecar.initializer.istio.io`
 as the first in the list of pending initializers.
 
 3) istio-initializer checks to see if it was responsible for
-initializing the namespace of the workload. No further work is done
-and the initializer ignores the workload if the initializer is
+initializing workloads in the namespace of the workload. No further work is done
```
ldemailly: double workload ?

TimDumol: It was intentional, but in hindsight it sounds less awkward to say "initializing workloads in its namespace." Will do!

TimDumol: Actually, then again, "its" is unclear, since "it" could refer to the `istio-initializer`, which makes no sense. Perhaps, "initializing workloads in the namespace of the uninitialized workload."?

ldemailly: maybe
"istio-initializer checks to see it is configured to manage the namespace of the new workload." ...

but it's fine as is when reading it all at once, thanks

PR #666, opened 2017-10-13
Add api attributes to attribute vocabulary.

apicl:


_docs/reference/config/mixer/attribute-vocabulary.md
```diff
@@ -63,4 +63,8 @@ deployments will have agents (Envoy or Mixer adapters) that produce these attrib
 | connection.sent.bytes_total | int64 | Total number of bytes sent by a destination service during the lifetime of a connection. | |
 | connection.duration | duration | The total amount of time a connection has been open. | |
 | context.protocol | string | Protocol of the request or connection being proxied. | tcp |
-| context.time | timestamp | The timestamp of the Report() call. | |
\ No newline at end of file
+| context.time | timestamp | The timestamp of the Report() call. | |
+| api.service | string | The public service name. One API could be handled by multiple internal services. Or multiple APIs could be handled by a single internal service. | my-svc.com |
+| api.version | string | The API version. Not required. | v1alpha1 |
```
douglas-reid: i don't think we need "Not required." here. This isn't about how attributes are used or when, but just about their names and definitions.

_docs/reference/config/mixer/attribute-vocabulary.md
```diff
@@ -63,4 +63,8 @@ deployments will have agents (Envoy or Mixer adapters) that produce these attrib
...
+| api.operation | string | Unique string used to identify the operation. The id MUST be unique among all operations described in a specific service + version. | getPetsById |
```
douglas-reid: can we make this: "...described in a specific <*api.*service, *api*.version>" to be clearer about the uniqueness requirements?

_docs/reference/config/mixer/attribute-vocabulary.md
```diff
@@ -63,4 +63,8 @@ deployments will have agents (Envoy or Mixer adapters) that produce these attrib
...
+| api.protocol | string | The protocol type of the API call. Mainly for monitoring/analytics. Note that this is the frontend protocol exposed to the client, not the protocol implemented by the backend service. | "http", “https”, or "grpc" |
```
douglas-reid: side question about this general attribute: how is this distinct from `context.protocol` (which is also about the protocol exposed to the client, iiuc)?  And, if this is distinct, can we clarify the difference in this documentation?

_docs/reference/config/mixer/attribute-vocabulary.md
```diff
@@ -63,4 +63,8 @@ deployments will have agents (Envoy or Mixer adapters) that produce these attrib
 | connection.sent.bytes_total | int64 | Total number of bytes sent by a destination service during the lifetime of a connection. | |
 | connection.duration | duration | The total amount of time a connection has been open. | |
 | context.protocol | string | Protocol of the request or connection being proxied. | tcp |
-| context.time | timestamp | The timestamp of the Report() call. | |
\ No newline at end of file
+| context.time | timestamp | The timestamp of the Report() call. | |
```
geeknoid: @douglas-reid Can we find a better description for context.time? In particular, I'd like to not mention the Report() call here. This is an attribute, it can be used in cases other than Report. For example, it could be used in Check and could be used somewhat by a service calling Mixer.

_docs/reference/config/mixer/attribute-vocabulary.md
geeknoid: Yeah, this does seem redundant.

_docs/reference/config/mixer/attribute-vocabulary.md
douglas-reid: sure :). it is meant to be distinct from both `request.time` and `response.time`. It covers TCP long-lived connection scenarios (as well as batching scenarios) to allow tracking of when the client called Mixer. I'm sure it can be used in other scenarios too.

Would "timestamp of Mixer operation" be better?

geeknoid: Yeah, I think that would be an improvement. Chengliang, please update as part of this PR.

_docs/reference/config/mixer/attribute-vocabulary.md
apicl: Done.

_docs/reference/config/mixer/attribute-vocabulary.md
apicl: Done.

_docs/reference/config/mixer/attribute-vocabulary.md
apicl: API protocol is different from context.protocol. Context protocol can be http, tcp, etc. The api protocol currently only include "http", "https", "grpc".

geeknoid:
Are there cases where api.protocol is non-empty and is different than
context.protocol? If not, the api.protocol is completely redundant.

apicl:
PTAL.

_docs/reference/config/mixer/attribute-vocabulary.md
apicl: Done.

apicl:
PTAL.

api.protocol could be grpc, while context.protocol is "http".

_docs/reference/config/mixer/attribute-vocabulary.md
```diff
@@ -63,4 +63,8 @@ deployments will have agents (Envoy or Mixer adapters) that produce these attrib
 | connection.sent.bytes_total | int64 | Total number of bytes sent by a destination service during the lifetime of a connection. | |
 | connection.duration | duration | The total amount of time a connection has been open. | |
 | context.protocol | string | Protocol of the request or connection being proxied. | tcp |
-| context.time | timestamp | The timestamp of the Report() call. | |
\ No newline at end of file
+| context.time | timestamp | The timestamp of Mixer operation. | |
+| api.service | string | The public service name. One API could be handled by multiple internal services. Or multiple APIs could be handled by a single internal service. | my-svc.com |
+| api.version | string | The API version. | v1alpha1 |
+| api.operation | string | Unique string used to identify the operation. The id MUST be unique among all operations described in a specific <service, version>. | getPetsById |
```
geeknoid: I actually don't understand the comment about MUST be unique. An attribute is basically a parameter to a function. There is no uniqueness requirement on this parameter. There might be a uniqueness requirement when the operation is declared in an API spec, but its not relevant to this document.

So I'd remove the whole second sentence.

_docs/reference/config/mixer/attribute-vocabulary.md
```diff
@@ -63,4 +63,8 @@ deployments will have agents (Envoy or Mixer adapters) that produce these attrib
 | connection.sent.bytes_total | int64 | Total number of bytes sent by a destination service during the lifetime of a connection. | |
 | connection.duration | duration | The total amount of time a connection has been open. | |
 | context.protocol | string | Protocol of the request or connection being proxied. | tcp |
-| context.time | timestamp | The timestamp of the Report() call. | |
\ No newline at end of file
+| context.time | timestamp | The timestamp of Mixer operation. | |
+| api.service | string | The public service name. One API could be handled by multiple internal services. Or multiple APIs could be handled by a single internal service. | my-svc.com |
```
geeknoid: Perhaps this should be phrased in a similar manner to api.protocol:

The public service name. This is different than the in-mesh service identity and reflects the name of the service exposed to the client.

apicl: Done.

_docs/reference/config/mixer/attribute-vocabulary.md
apicl: The policies are defined based on the operation. This uniqueness documentation helps clarify the assumptions.

apicl:
PTAL.

_docs/reference/config/mixer/attribute-vocabulary.md
geeknoid: I still think you could remove that. But if you insist on keeping the sentence, then MUST should be changed to 'is'.

apicl:
PTAL.

PR #667, opened 2017-10-13
Add liveness/readiness to security faq

wattli:


_docs/tasks/security/faq.md
```diff
@@ -43,9 +43,22 @@ type: markdown
...
+  for health check, e.g., one can install curl in the service pod and curl itself
+  within the pod. The Istio team is actively working on a solution.
+
+  An example of readinessProbe:
+
+  ```
+  livenessProbe:
+    exec:
+      command:
+      - /bin/sh
```
ldemailly: don't do the /bin/sh it's unecessary for one command

_docs/tasks/security/faq.md
```diff
@@ -43,9 +43,22 @@ type: markdown
...
+      - -c
+      - curl -f http://localhost:{$PORT}
```
ldemailly: add /health to the url for example

_docs/tasks/security/faq.md
```diff
@@ -43,9 +43,22 @@ type: markdown
 
   If Istio Auth is enabled, http and tcp health check from kubelet will not
   work since they do not have Istio Auth issued certs. A workaround is to
-  use command option for health check, e.g., one can install curl in the
-  service pod and curl itself within the pod. The Istio team is actively
-  working on a solution.
+  use [command option](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-command)
```
ldemailly: use a [liveness command](...)

wattli: done

_docs/tasks/security/faq.md
wattli: done

_docs/tasks/security/faq.md
wattli: in reality, many people just simply check localhost:port, i.e., they don't use /health for health check. I feel the current way will be easier to understand (for those who use /health, they will simply add /health, but if we go the other way, i.e., add /health to the url, for those who do not use /health, they will not sure what to do).

_docs/tasks/security/faq.md
```diff
@@ -43,9 +43,22 @@ type: markdown
...
+
+  An example of readinessProbe:
+
+  ```
+  livenessProbe:
+    exec:
+      command:
+      - curl
+      - -f
+      - http://localhost:{$PORT}
```
ldemailly: the {$PORT} is strange (that’s meant as an example/ for user to change to the actual port right?)
what about : 
```
   - http://localhost:8080/health # replace with your actual port and health endpoint
```

PR #670, opened 2017-10-17
Revise a few docs to make it more clear

wattli:


_docs/setup/kubernetes/mesh-expansion.md
```diff
@@ -79,7 +81,7 @@ install/tools/setupMeshEx.sh generateDnsmasq
 Example generated file:
 
 ```bash
-cat /etc/dnsmasq.d/kubedns
+cat kubedns
```
ldemailly: thx!

_docs/setup/kubernetes/mesh-expansion.md
```diff
@@ -100,11 +102,13 @@ and install the setup:
...
 # On a Mac, either brew install base64 or set BASE64_DECODE="/usr/bin/base64 -D"
 export GCP_OPTS="--zone MY_ZONE --project MY_PROJECT"
+```
+```bash
 install/tools/setupMeshEx.sh machineSetup VM_NAME
 ```
 
 Or the equivalent manual steps:
-
+------- Manual setup steps begin -------
```
ldemailly: how does this look like with jekyll ? screen shot ?

wattli: http://0.0.0.0:4000/docs/setup/kubernetes/mesh-expansion.html

Just in case my server dies, sorry for non-google reviewers, who would not be able to access this page:
https://screenshot.googleplex.com/euT65Eq4PsW
https://screenshot.googleplex.com/hkZgqZ3a6GL

ldemailly: 0.0.0.0 won't work for anyone :)

Pro tip: you can drop screenshot directly into github

_docs/setup/kubernetes/quick-start.md
```diff
@@ -87,6 +87,11 @@ run the following command to download and extract the latest release automatical
     * The `istioctl` client binary in the `bin/` directory. `istioctl` is used when manually injecting Envoy as a sidecar proxy and for creating routing rules and policies.
     * The `istio.VERSION` configuration file
 
+1. Change directory to istio package. For example, if the package is istio-0.2.7
+  ```bash
+  cd istio-0.2.7
+  ```
```
ldemailly: Use {data.version} (I can check the exact syntax from my phone sorry, need to check the footer for an example) or mention that is just an example and the version could be different

wattli: I mentioned 'For example, if the package is istio-0.2.7...'

ldemailly: sorry quick reading on my phone, I missed that - thx (still could use the version variable but it's fine)

PR #671, opened 2017-10-18
Revise expansion page

wattli:
Currently the page does not look well.

[Webpage] https://istio.io/docs/setup/kubernetes/mesh-expansion.html
[Screenshot] https://screenshot.googleplex.com/aSUpnDHY66D

### DIFF
--- a/_docs/setup/kubernetes/mesh-expansion.md
+++ b/_docs/setup/kubernetes/mesh-expansion.md
@@ -127,14 +127,15 @@ Example generated message:
 # Verify you get the same address as shown as "EXTERNAL-IP" in 'kubectl get svc -n istio-system istio-pilot-ilb'
 istio-pilot.istio-system has address 10.150.0.6
 ```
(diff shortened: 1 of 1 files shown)


ldemailly:
hey I noticed the section where the variable are exported and the install command are seperated since your last PR can you put them back in 1 block - they all are commands ?

wattli:
@ldemailly , somehow github won't allow me to reply your comment inline.

"hey I noticed the section where the variable are exported and the install command are seperated since your last PR can you put them back in 1 block - they all are commands ?"

Do you mean this one 

"export GCP_OPTS="--zone MY_ZONE --project MY_PROJECT"
"install/tools/setupMeshEx.sh generateClusterEnv MY_CLUSTER_NAME"

If so, I purposely separate them since they are two commands so the copy button on the right would work.

ldemailly:
ok - given they need to be edited it makes sense, thanks

PR #674, opened 2017-10-19
Fix formatting for ingress.html

geeknoid:


_docs/tasks/traffic-management/ingress.md
```diff
@@ -95,11 +95,9 @@ rules.
          percent: 100
          httpStatus: 403 #Forbidden for all URLs
    EOF
```
wattli: Thanks for fixing this, but there is another issue for the above command when format is fixed. The error I got is 

"Error: configuration is invalid: 1 error occurred:
name or service is mandatory for a service reference"

geeknoid: Where does this error occur? Are you saying that the instructions in the document are incorrect and lead to this error when you try to follow the instructions? If that's the case, then someone else will need to fix the content. I only handled the formatting issues here.

PR #678, opened 2017-10-19
Modify yaml to fix the error from istioctl

linsun:
https://github.com/istio/istio.github.io/issues/672#issuecomment-338002266

https://github.com/istio/istio.github.io/issues/672

### DIFF
--- a/_docs/setup/consul/install.md
+++ b/_docs/setup/consul/install.md
@@ -128,5 +128,5 @@ found in the
 > the sidecar process. 
 
 Refer to the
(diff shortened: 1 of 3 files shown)


juliaaano:
Using kubectl instead of istioctl means the yaml file won't be validated. Is this the desired outcome for this? I believe the right fix would be adding the required field to that file, so it pleases istioctl's validation.

linsun:
good suggestion!  The error msg was not helpful but I figured it out and tested it.  PTAL @rshriram @wattli

PR #680, opened 2017-10-20
Updated document for excludeNamespaces in istio initializer.

gyliu513:
/cc @ayj @linsun @rshriram


linsun:
Thanks for this, you are reading my mind :).

PR #681, opened 2017-10-20
Put `Prometheus install` first before `Collect Metrics`.

gyliu513:
As the `Collect Metrics` topics depend on `Prometheus install`, so
we need to have the `Prometheus install` topic first.


rshriram:
That’s the reason I put Prometheus install as an explicit step in collect metrics. It does not make sense for an end user to see how to query metrics before actually collecting them.

gyliu513:
Yes, fair enough, will close this one. thanks @rshriram

PR #694, opened 2017-10-29
Update kubernetes version to 1.7 or newer.

gyliu513:
All of the 1.7 kubernetes can works well with istio CRD resources,
we should update it to avoid confuse.

/cc @linsun


gyliu513:
@ldemailly what is the reason of 1.7.2 does not work, at least it works for 1.7.3 in my env.

ldemailly:
I forget but there was an issue - going to close this for now

gyliu513:
@ldemailly I think that we need to update kubectl version to 1.7.3 as 1.7.3 also works, I got some question from customers about the version who is using 1.7.3, so better update to 1.7.3

linsun:
Updating to 1.7.3 is fine with me given Guang has tested it and we didn't recall any issue but we picked 1.7.4 given it was the latest.

ldemailly:
I am pretty sure there is a bug with older than 1.7.4 but maybe it was only initializer related ? or was it about validation hooks ? @andraxylia do you remember?

gyliu513:
@ldemailly We have a wechat group for istio in China and I got some questions for the kubernetes version for istio, and some people are using 1.7.3 and they also did some test and found istio works well. I think that we should update the document so as not bring confuse, comments?

ldemailly:
sorry to keep repeating the same thing but I'm sure there was a reason (it's not that we picked the latest at the time) - maybe if you go over the release notes it may show up but I hope @andraxylia or @ayj or someone remembers

ayj:
I don't recall the reason. Maybe a GKE specific requirement since they bump up the latest minor version frequently and make previous minor versions unavailable for new clusters?

ldemailly:
I searched a bit and didn't find except that we put 1.7.4 for the first time in 0.2.2
https://github.com/istio/istio/wiki/v0.2.2
I guess go ahead and merge and we'll see if we get reports/errors

linsun:
Thanks for digging around, I'll merge it.

PR #703, opened 2017-11-03
Fix a bunch of links so that self-hosted istio.io works properly.

geeknoid:


geeknoid:
The presubmit failure is claiming broken links which are in fact completely valid in the generated site. These are references to the v-0.1 subtree. I'm not sure what to do about these.

geeknoid:
I just reverted those two links from using {{home}} to using the absolute https://istio.io form instead. The links can't work anyway when staging the site, so there's no harm.

ldemailly:
thx a lot ! quick question:

what is the difference between {{home}} and / ? also is there a problem with just relative links, ie if in say docs/setup any issue with a link like a href="kube/foo.html" ?

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (4.2.8)
+    activesupport (4.2.9)
```
ldemailly: we probably don't need to change that file but it's ok too (we should track what github pages actually uses, last time I checked it was jekyll/jekyll:3.5.2  but we probably need to track it as some day we'll diverge

geeknoid:
{{home}} gets a different value depending on where the site is hosted, so
that the links work correctly. Otherwise, when you directly have
https://istio.io for links internal to the site, then when you stage the
site and click on the link, you jump out of your staged site and end up on
istio.io instead.

Relative links work fine in general. We want only links pointing downward
into the hierarchy and no .. references.

ldemailly:
I understand that a fully qualified url (with host and scheme) is bad, but a relative to the root, or relative to current dir - ie  {{home}} is equivalent to / except when we re-root in the v-0.1 I guess which we do so probably good indeed to not use /... paths

geeknoid:
Staged GitHub sites aren't at the root (e.g.
geeknoid.github.io/istio.github.io), so just using / doesn't work for that
case.

ldemailly:
ic forked vs branches, understood ! thx

PR #704, opened 2017-11-07
docs: istioctl create now requires a namespace

rosenhouse:
without these changes, I get:
  "Error: an empty namespace may not be set during creation"


rosenhouse:
The Jenkins failed, and when I try to view details I see

    rosenhouse is missing the Overall/Read permission

ayj:
FWIW, this is already fixed on master with https://github.com/istio/old_pilot_repo/pull/1499.

rosenhouse:
Oh, that's better anyhow!  Thanks.

Issue #718, opened 2017-11-18
broken links to envoy


ldemailly:
Checking 170 external links...
Ran on 146 files!

- ./_site/blog/using-network-policy-in-concert-with-istio.html
  *  External link https://envoyproxy.github.io/envoy/operations/admin.html#operations-admin-interface failed: 404 No error
- ./_site/docs/concepts/traffic-management/pilot.html
  *  External link https://envoyproxy.github.io/envoy/configuration/cluster_manager/cds.html failed: 404 No error
  *  External link https://envoyproxy.github.io/envoy/configuration/cluster_manager/sds_api.html failed: 404 No error
  *  External link https://envoyproxy.github.io/envoy/configuration/http_conn_man/rds.html failed: 404 No error
- ./_site/docs/reference/config/traffic-rules/destination-policies.html
  *  External link https://envoyproxy.github.io/envoy/intro/arch_overview/circuit_breaking.html failed: 404 No error
  *  External link https://envoyproxy.github.io/envoy/intro/arch_overview/load_balancing.html failed: 404 No error
  *  External link https://envoyproxy.github.io/envoy/intro/arch_overview/outlier.html failed: 404 No error
- ./_site/docs/tasks/telemetry/tcp-metrics.html
  *  External link https://envoyproxy.github.io/envoy/configuration/network_filters/mongo_proxy_filter.html#statistics failed: 404 No error
rake aborted!
HTML-Proofer found 8 failures!

(found in #717)

ldemailly:
Fixed in #721 (thx @geeknoid !)

PR #719, opened 2017-11-18
Setup the html check in circleci

ldemailly:


ldemailly:
as jenkins is broken I've setup the `rake test` in circle ci (which is not passing yet because of #718 )
and removed jenkins from the required checks (we can delete it entirely)

it's super fast thanks to caching the bundle install results

cc @geeknoid fyi

PR #724, opened 2017-11-21
Add external url cache

ldemailly:
That should help avoid the spurious failures or at least make “rebuild”
work if the remote site isn’t actually down

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,7 @@ require 'html-proofer'
 task :test do
   sh "bundle exec jekyll build"
   options = { :check_html => true, # :validation => { :report_missing_names, :report_invalid_tags => true, :report_script_embeds => true },
+              :cache => { :timeframe => '2d'},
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -3,6 +3,7 @@ require 'html-proofer'
 task :test do
   sh "bundle exec jekyll build"
   options = { :check_html => true, # :validation => { :report_missing_names, :report_invalid_tags => true, :report_script_embeds => true },
+              :cache => { :timeframe => '2d'},
```
ldemailly: 2 days is a random compromise between having a cache for a couple of PRs and also noticing not too long after when some links go dead (we could make do a 1 week cache if we did a nightly check or some such)

ldemailly:
This works though strangely it seems it has 1 link not in cache each subsequent run
```
Found 166 links in the cache...
Adding 1 link to the cache...
Removing 1 link from the cache...
Checking 1 external link...
Ran on 146 files!
```

Issue #720, opened 2017-11-18
remove jenkins from istio.github.io hooks/checks


ldemailly:
I've setup circleci rake test (#719) as a replacement for the broken jenkins to check for bad html and links

it happens to be a lot faster thanks to caching too

can you remove the jenkins from the hooks ? I didn't see where that setting is ?

thanks!

sebastienvas:
jenkins is using a webhook at the org level. I deleted the jobs in there. Thanks!

ldemailly:
thanks!

PR #728, opened 2017-11-21
Dropping jenkins

ldemailly:
Presubmit has been failing for days


ldemailly:
well that doesn't seem to remove the "istio.github.io/presubmit — Jenkins queued the run (PR opened)" entry so closing this for now, will let seb fix it for real in #720

PR #730, opened 2017-11-23
fix minikube command to add missing namespace

skyao:
The command of minikube should also add namespace option, otherwise the command will fail.


hmalphettes:
Hi, this commit is duplicating the namespace argument:
```
-n istio-system -n istio-system
```
I have tested the original command before this commit and it works fine.

ldemailly:
indeed the 2nd change is a dup, I didn’t notice,
can you make a pr removing that part?

hmalphettes:
Sure. The 1st change is also a dup with `-l istio=ingress` in between the 2 identical arguments:
```
kubectl get po -n istio-system -l istio=ingress -n istio-system -o
```
So the PR will be reverting this commit.

ldemailly:
wow, indeed

PR #732, opened 2017-11-23
Add a custom 404 page so we don't get the default GitHub one.

geeknoid:


ldemailly:
see the circleci/rake test output, need an alt tag on the pict

there seems to also be unrelated timeout reaching some external links (despite the cache i setup)

geeknoid:
I added the missing attributes, so that's fixed. The timeouts persist though....

PR #739, opened 2017-11-26
Expand the glossary.

geeknoid:
- Add a few more words

- Add a bunch of cross-linking between entries

### DIFF
--- a/_glossary/adapters.md
+++ b/_glossary/adapters.md
@@ -1,0 +1,13 @@
+---
+title: Adapters
+type: markdown
+---
+{% include home.html %}
+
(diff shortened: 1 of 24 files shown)


_glossary/instance.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Instance
+type: markdown
+---
+{% include home.html %}
+
+An instance represents a chunk of Mixer data that is produced by inspecting a
```
ldemailly: maybe this should be "Mixer Instance" as instance seems quite generic (a pod could be an instance of a service for instance (ah!))

geeknoid:
Done, renamed to mixer-instance and mixer-handler

PR #741, opened 2017-11-26
Clean up dead stuff

geeknoid:
- Remove unreferenced files

- Stop using sitemap_exclude in front matter since nothing is looking at this.


js/common.js
```diff
@@ -87,7 +87,7 @@ $(document).ready(function() {
 });
 
 // Collapsible navbar menu, using https://github.com/jordnkr/collapsible
-$.getScript("{{ site.baseurl }}/js/jquery.collapsible.js", function(){
+$.getScript("{{ site.baseurl }}/js/jquery.collapsible.min.js", function(){
```
ldemailly: but you left/updated the non .min. version?

geeknoid:
Yes, we need the non-min version since that's the source code, whereas the
.min version is the "compiled" version. We need the source code to make
future edits.

On Sun, Nov 26, 2017 at 11:18 AM, Laurent Demailly <notifications@github.com

ldemailly:
ah it seemed that version was stock from the source, but we've still needed edits?

PR #753, opened 2017-12-04
Add Istio Config concept, glossary

xiaolanz:
Add config concept, glossary and update existing reference to the new terms.


geeknoid:
Do you have a doc staging web site setup? Makes doc reviews a lot nicer.

geeknoid:
This page explains how to stage on GitHub: https://istio.io/docs/reference/contribute/staging-your-changes.html

xiaolanz:
Staged website: https://xiaolanz.github.io/docs/concepts/
PTAL

_docs/concepts/configuration/model.md
```diff
@@ -0,0 +1,34 @@
...
+title: Data model
+overview: Istio configuration data model.
+
+order: 10
+
+layout: docs
+type: markdown
+---
+
+Istio config is a conglomerate of config aspects. Each config aspects is
```
geeknoid: I'm worried that this text is generally too focused on implementation strategy, as opposed to the user experience. Consider https://kubernetes.io/docs/concepts/overview/kubernetes-api/ as an example that's more focused on what the user sees rather than how it's implemented underneath. Can we adapt their content and just drop it here instead?

_docs/concepts/configuration/space.md
```diff
@@ -0,0 +1,19 @@
+---
```
geeknoid: Looking at the words in the glossary, I don't think these should be in there. They're all too specific terms to be in the general platform

geeknoid:
Note that right now, we have _docs/reference/writing-config.md. That file should probably be moved out of reference and sit side-by-side with the config-related concept docs.

_docs/concepts/configuration/space.md
xiaolanz: Do you suggest to put the glossary under config concepts? I prefer not. I'd like to have one "contains all" Index like experience for glossary. Most ppl use glossary like dictionary, they don't really care what other terms are there too.

geeknoid:
I have two issues with this stuff right now:

- We have a very complicated diagram with no legend of explanation, only a
pointer to the glossary.

- I don't believe these concepts belong in the glossary, it's too detailed.
We don't put all the command-line options, all the adapter config types,
etc, in the glossary. Look at the K8s glossary as an example (
https://kubernetes.io/docs/reference/glossary/?all=true). The stuff in
there is high-level in nature. Don't forget we have search on the site, so
users can find what they need that way too. I can guarantee that nobody is
going to go to the glossary and search for "Security config" for example.
They'll look for the name of the CRD, they'll search for "configuring
security" in the search box, they'll look in the reference material, etc.

xiaolanz:
addressed all comments. PTAL

ldemailly:
please fix the bad links (see circleci and/or run rake test on your machine)

xiaolanz:
Sure, I will fix all the broken links after the structure/text comments are all made and addressed.

_docs/concepts/configuration/model.md
```diff
@@ -0,0 +1,34 @@
...
+corresponding to an kubernetes resource. The organization of resources is "flat"
+on files representation. The Kubernetes API server stores the config data
+according to name and namespace in the hierarchy.
+
+Istio config is completely built on Kubernetes resources. Besides a small set of
```
gjwestlife: Saying "Kubernetes resource" here could confuse with k8s core resources. Let's change to k8s CRD?

_docs/concepts/configuration/model.md
```diff
@@ -0,0 +1,34 @@
...
+installation scripts or docker yaml for other environments), Istio config are
+only three kinds:
+
+## Types
+A set of configs are for custom defined resources types (CRD). Custom resource spec proto is defined in https://github.com/istio/api.
+
+## Instances
+The rest are config instances for a certain type. There is no inheritance or abstraction relationships.
+
+## ConfigMap
```
gjwestlife: Why is ConfigMap a kind of config? To me, it seems an implementation details.

_docs/concepts/configuration/taxonomy.md
```diff
@@ -0,0 +1,154 @@
...
+
+layout: docs
+type: markdown
+---
+{% include home.html %}
+
+Istio configuration is component oriented. Each component exposes independently
+evolved configuration resources to the Istio end user, such as Mixer
+configuration.  Another way to understand the configuration is through their
+functionality and the entities where they apply on.  Three entities in Istio
```
gjwestlife: s/Three/These

_docs/concepts/configuration/taxonomy.md
```diff
@@ -0,0 +1,154 @@
...
+Service interface definition, such as API IDL, protocol.
+
+## Istio configuration
+
+Joint space of [core configuration](#core-configuration) and [service configuration](#service-configuration).
+Mesh operator can read/write these configurations through [istioctl]({{home}}/docs/reference/commands/istioctl.html).
+
+## Logging configuration
+
+Define log entries, format, and input data.
```
gjwestlife: What is input data?

_docs/concepts/configuration/taxonomy.md
```diff
@@ -0,0 +1,154 @@
...
+## Management configuration
+
+The configuration that defines and manages the traffic sent to a service. This
+includes [traffic configuration](#traffic-configuration), [catalog configuration](#catalog-configuration),
+[IDL configuration](#idl-configuration), [authentication configuration](#authentication-configuration),
+[adapter configuration](#adapter-configuration), and [policy configuration](#policy-configuration).
+
+## Mesh configuration
+
+Mesh configuration refers to a particular scoped [service configuration](#service-configuration)
```
gjwestlife: Is it just a Mesh level config or service config scoped at the mesh level?

_docs/concepts/configuration/taxonomy.md
```diff
@@ -0,0 +1,154 @@
+---
+title: Taxonomy
```
smawson: I don't think this document needs to be on the website. Maybe in github, but maybe just keep as a google doc.

I'm worried it will be more confusing than helpful.

_docs/concepts/configuration/space.md
smawson: I agree with Martin, I don't think we need all of this in the glossary.

I'm not sure we need it on the website at all.

What are you trying to accomplish by adding all of this to the website?  What is the goal?

For example, having "X config" entries for every type of configuration in the glossary is a waste. People know what configuration means.

So for example, rather than defining "API Management config" as "Configuration for API Management", we should have an entry for "API Management" that defines what we mean by API Management.

The config entries are just redundant.

_docs/concepts/configuration/model.md
smawson: +1, fully agree. Anything on istio.io needs to be for the end user of Istio, and needs to be kept as simple as possible.

smawson:
General question from my review: What problem is this PR trying to solve?

I'm not sure why we need any of this on the website.

wora:
I have the same feeling as @smawson. The purpose seems very unclear and the scope seems too broad. IMHO, there is little chance to design a platform feature without prototyping. For example, we could introduce the monitoring aspect in an experimental branch and make sure it works well before we submit it into main branch. It is too early to claim such a big design scope actually works in practice.

_docs/concepts/configuration/model.md
```diff
@@ -0,0 +1,34 @@
...
+type: markdown
+---
+
+Istio config is a conglomerate of config aspects. Each config aspects is
+corresponding to an kubernetes resource. The organization of resources is "flat"
+on files representation. The Kubernetes API server stores the config data
+according to name and namespace in the hierarchy.
+
+Istio config is completely built on Kubernetes resources. Besides a small set of
+installation scripts or docker yaml for other environments), Istio config are
```
cmluciano: Can you split this to a new sentence?

`Istio config formats:`

_docs/concepts/configuration/model.md
```diff
@@ -0,0 +1,34 @@
...
+only three kinds:
+
+## Types
+A set of configs are for custom defined resources types (CRD). Custom resource spec proto is defined in https://github.com/istio/api.
+
+## Instances
+The rest are config instances for a certain type. There is no inheritance or abstraction relationships.
```
cmluciano: What is "the rest" ?

_docs/concepts/configuration/model.md
```diff
@@ -0,0 +1,34 @@
...
+
+## ConfigMap
+ConfigMap is the original Istio config solution. The config is read by Pilot. It
```
cmluciano: You can probably combine the first two sentences.

_docs/concepts/configuration/model.md
```diff
@@ -0,0 +1,34 @@
...
+currently contains mesh wide envoy configs and security configs. We try to keep
```
cmluciano: Capitalize Envoy.

_glossary/core-config.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Core config
+type: markdown
+---
+Istio core component configs. Typically the configs to install Pilot, Proxy,
```
cmluciano: Core config and Deployment config descriptions feel very similar. Can we add specific examples to each?

_glossary/mesh-config.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Mesh config
+type: markdown
+---
+Mesh config refers to a particular scoped [service config](#service-config)
```
cmluciano: Can we call out that MeshConfig is essentially Envoy runtime flags?

_docs/concepts/configuration/taxonomy.md
xiaolanz: Currently, we don't have a  functional definition of istio configures. It poses challenge when we communicate with our users, basically, ppl use the same term for different things. This tends to clearly define configuration terms for Istio users, also make sure references in other parts of the doc is consistent.

linsun:
Can we talk about configuration benefit for the end user before we even introduce any details?  I feel that is missing for me.

smawson:
+1 Lin, this is diving into a bunch of detail without setting the context, and I fear it will lead to more confusion rather than clearing up confusion.

xiaolanz:
Let's keep them in the doc for now.

PR #755, opened 2017-12-05
Document RBAC set up for ingress key cert protection.

JimmyCYJ:
Which issue this PR fixes: fixes https://github.com/istio/old_auth_repo/issues/313


_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,344 @@ rules.
    curl -I -k https://$INGRESS_HOST/status/200
    ```
 
+## Configuring RBAC for ingress key/cert
```
wattli: The key/cert is only used in above section, would it make more sense to put the new content in that section?

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,344 @@ rules.
...
+There are service accounts which can access this ingress key/cert, and the ingress key/cert is not 
+encrypted. This leads to risks of leaking key/cert. We can set up Role-Based Access Control ("RBAC")
```
wattli: This is not accurate. K8s supports encrypted secret already, see here: https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/ . It would be great if we could update this as well. But maybe in a follow up PR.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,344 @@ rules.
...
+defines some ClusterRoles which grant access to all secret resources. For example, 
+istio-pilot-istio-system and istio-mixer-istio-system allow reading secret resources, and 
+istio-ca-istio-system allows reading, creating and updating secret resources.
+[istio/install/kubernetes/istio.yaml](https://github.com/istio/istio/blob/master/install/kubernetes/istio.yaml) 
+also defines ClusterRoleBindings which bind these ClusterRoles to service accounts, such as 
+istio-pilot-service-account, istio-mixer-service-account, istio-ca-service-account and 
+istio-ingress-service-account. We need to update or replace these RBAC set up to only allow 
+istio-ingress-service-account to access ingress key/cert.       
+
+1. Update RBAC set up for istio-pilot-service-account and istio-mixer-istio-service-account.
```
wattli: Please help to keep the new content succinct and only display things of interest.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,344 @@ rules.
...
+   kubectl delete ClusterRole istio-mixer-istio-system
+   ```
+   List all secrets in namespace istio-system that we need to protect using RBAC.
+   
+   ```bash
+   kubectl get secrets -n istio-system
+   ```
+   This produces the following output:
+   
+   ```bash
```
wattli: We don't have to display everything, please focus on ingress related one.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,344 @@ rules.
...
+   bind this new ClusterRole to service accounts istio-pilot-service-account and 
+   istio-mixer-istio-service-account.
+   
+   ```bash
+   kind: ClusterRole
+   apiVersion: rbac.authorization.k8s.io/v1beta1
+   metadata:
+     name: istio-pilot-mixer-istio-system
+   rules:
+   - apiGroups: [""] # "" indicates the core API group
```
wattli: Same here and below. Please avoid showing unrelated content.

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: Thanks for pointing that out. I didn't know that we can encrypt secret. I will document encrypting secrets in a follow up PR.

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: Done. Thanks!

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: Done.

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: I have made the content shorter now.

JimmyCYJ:
I have revised the document, please take a look. Thanks.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,204 @@ rules.
...
+   ```
+   
+1. Update RBAC set up for istio-pilot-service-account and istio-mixer-istio-service-account.
+   
+   Delete existing ClusterRoleBindings and ClusterRole.  
+
+   ```bash
+   kubectl delete ClusterRoleBinding istio-pilot-admin-role-binding-istio-system
+   kubectl delete ClusterRoleBinding istio-mixer-admin-role-binding-istio-system
+   kubectl delete ClusterRole istio-mixer-istio-system
```
wattli: Is there a similar role for istio-pilot-istio-system?

wattli: Is there any way to recover the deletion here at the end of this file? The idea is to clean up the change in this file after everything is done.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,204 @@ rules.
...
+   ```bash
+   kubectl get secrets -n istio-system
+   ```
+   This produces the following output:
+   
+   ```bash
+   NAME                                        TYPE                                  DATA      AGE
+   istio-ca-secret                             istio.io/ca-root                      2         7d
+   istio-ingress-certs                         kubernetes.io/tls                     2         7d
+   istio-ingress-service-account-token-kwr85   kubernetes.io/service-account-token   3         7d
```
wattli: Can we remove this one as well? And istio-ca-secret?

_docs/tasks/traffic-management/ingress.md
wattli: And what privileges will pilot/mixer lose after this change? I am afraid this will break existing system. Let me add some pilot and mixer guy.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,204 @@ rules.
...
+   
+   ```bash
+   kind: ClusterRole
+   apiVersion: rbac.authorization.k8s.io/v1beta1
+   metadata:
+     name: istio-pilot-mixer-istio-system
+   rules:
+   - apiGroups: [""] # "" indicates the core API group
+     resources: ["secrets"]
+     resourceNames: ["istio-ca-service-account-token-rl4xm"]
```
lookuptable: It seems quite dangerous to hard-code names with randomized tokens (`rl4xm`).

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,204 @@ rules.
...
+   apiVersion: rbac.authorization.k8s.io/v1beta1
+   metadata:
+     name: istio-pilot-mixer-admin-role-binding-istio-system
+   subjects:
+   - kind: ServiceAccount
+     name: istio-pilot-service-account
+     namespace: istio-system
+   - kind: ServiceAccount
+     name: istio-mixer-service-account
+     namespace: istio-system
```
lookuptable: So we're binding the same role to both mixer SA and pilot SA. Is this desirable? IRCC mixer and pilot need different permissions.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,6 +269,204 @@ rules.
...
+   Delete existing ClusterRoleBindings and ClusterRole.  
+
+   ```bash
+   kubectl delete ClusterRoleBinding istio-pilot-admin-role-binding-istio-system
+   kubectl delete ClusterRoleBinding istio-mixer-admin-role-binding-istio-system
+   kubectl delete ClusterRole istio-mixer-istio-system
+   ```
+   
+   Create istio-pilot-mixer-istio-system.yaml, which allows istio-pilot-service-account and 
+   istio-mixer-service-account to read all kubernetes.io/service-account-token,
```
lookuptable: I don't think anyone needs to read `kubernetes.io/service-account-token`, no?

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: Done.

JimmyCYJ:
Sure, I will update this PR.

ldemailly:
ping ?

JimmyCYJ:
Sorry I was working on other stuff, will update this PR soon.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -214,11 +214,6 @@ The following are the known limitations of Istio ingress:
      kubectl get ingress secure-ingress -o wide
      ```
 
-     ```bash
-     NAME             HOSTS     ADDRESS                 PORTS     AGE
-     secure-ingress   *         130.211.10.121          80        1d
```
ldemailly: why is this sample output removed?

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: istio-pilot-istio-system is also bound to account istio-ingress-service-account by istio-ingress-admin-role-binding-istio-system. We will delete istio-pilot-istio-system in next step, after deleting that ClusterRoleBinding object.

According to https://github.com/istio/istio/blob/11dc2fb937a23157215fffa06921cf507d27f757/install/kubernetes/istio.yaml#L22
istio-pilot-istio-system grants access permission to many resources. We need to bind those rules to istio-pilot-service-account. We should apply similar role binding process to account istio-mixer-service-account and  istio-ca-service-account

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: Agree. That line is removed.

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: This is a kubernetes.io/service-account-token, and it is removed.

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: Yes, they need different permissions. I have create separate ClusterRoles for each of them.

_docs/tasks/traffic-management/ingress.md
JimmyCYJ: That part is added back. Thanks.

JimmyCYJ: This PR is for the ingress controller task only. Maybe we can update the installation with a more general solution in future work?

wattli:
Please validate it before submission. Thanks for working on this, Jimmy.

PR #757, opened 2017-12-05
Updated artwork for home page + cleaned up logo + new logo label font

geeknoid:


ldemailly:
hmmm seems a bit hasty change
I can’t say I am a fan of that font, and maybe this could have been discussed more

PR #762, opened 2017-12-06
Add tutorial for per service mTLS enablement.

diemtvu:
This taks demos how to:
- use annotation to enable/disable mTLS
- Istio config to exclude mTLS for control services.

https://github.com/istio/istio.github.io/pull/752


_docs/tasks/security/per-service-mtls.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Before you begin
+
+* Understand Isio [mutual TLS authentication]({{home}}/docs/concepts/security/mutual-tls.html) concepts.
+
+* Familiar with [testing Istio mutual TLS authentication]({{home}}/docs/tasks/security/mutual-tls.html).
+
+* Install Istio with mutual TLS authentication by following the instructions in the [Installation guide]({{home}}/docs/setup/kubernetes/).
+
+* Start [httpbin demo](https://github.com/istio/istio/tree/master/samples/httpbin) with Istio sidecar. Also, for testing purpose, run two instances of [sleep](https://github.com/istio/istio/tree/master/samples/sleep), one with sidecar and one without (in different namespace)
```
wattli: Can we adding something like 'Following the commands below to start httpbin demo with Istio sidecar'. Just want to make sure the audiences know the relation between this paragraph and the commands below.

_docs/tasks/security/per-service-mtls.md
```diff
@@ -0,0 +1,122 @@
...
+
+```bash
+annotations:
+  auth.istio.io/8000: NONE
+```
+
+For a quick test, run `kubectl edit svc httpbin` and add the annotations above (
+or you can edit the original httpbin.yaml file and re-apply it). After the change is applied, request from sleep.legacy should now success, as the result of mTLS was dropped.
+
+Note: the annotations can be used in the opposite direction, i.e enable mTLS for a services, simply by using annotation value  `MUTUAL_TLS`, instead of `NONE`.
```
wattli: s/a services/a service

_docs/tasks/security/per-service-mtls.md
```diff
@@ -0,0 +1,122 @@
...
+
+The same test request above now fail with code 35, as sleep's sidecar starts using mTLS again:
+
+```bash
+kubectl exec $(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name}) -c sleep -- curl https://kubernetes.default:443/api/ -k -s
+```
+
+```bash
+command terminated with exit code 35
+```
```
wattli: Should we add another section about the other direction communication, i.e., from control service to sidecar?

diemtvu: If Istio service is as-server, then it falls in the first scenario, i.e we can disable TLS using annotation.

_docs/tasks/security/per-service-mtls.md
diemtvu: Done

_docs/tasks/security/per-service-mtls.md
diemtvu: Done

_docs/tasks/security/per-service-mtls.md
wattli: After a second thought, what about rename the two sections to something like 

Disable mutual TLS for Envoy as a server
Disable mutual TLS for Envoy as a client

That will make them more clear. And maybe we can explicitly mention that health check belongs to the first case, as health check is something customers/developers care about a lot.

wattli: Discussed offline, health check story is not complete yet. And we are waiting for multiplex port/alias port support.

diemtvu: Recap from offline discussion: I tried to make this as a tutorial to use the new features (annotation, exclude mtls list) rather than a solution for a specific problem, as while it helps to walk around the issue, it's not the solution we want to provide at the end. We can have another doc in future to address health check solution for example.

diemtvu:
@ldemailly Can you approve? Thanks.

ldemailly:
merged, thank you !

PR #775, opened 2017-12-07
change request_count metric to istio_request_count

frankbu:


geoand:
Hi, 

If I am not mistaken, the screenshot need to be updated as well to reflect the change.
Is this correct?

frankbu:
You are right @geoand. Good catch.

PR #776, opened 2017-12-07
Make the I in istio readable

ldemailly:
Verdana is a sans serif font - except it has a readable upper case I - see screenshots:


ldemailly:
After/Before:

screen shot 2017-12-07 at 11 50 54 am

screen shot 2017-12-07 at 11 50 19 am

ldemailly:
Survey:
https://docs.google.com/forms/d/e/1FAIpQLScfsO_ATK3rRj92mHaCsqTbfC2AU2wypJO5anLswjn8lG_gzw/viewanalytics

linsun:
Agreed, there are lots of places that this logo and possibly font is used...

ldemailly:
@zinuga from steering committee notes: "Punting discussion of Istio “I” to PM/product folks."
can you make a decision ?

zinuga:
Main goal here should be consistency. cant have different fonts, colors, brand colors in different places for istio. I suggest we make readable upper case I as our default font and add a branding guidelines covering logo, text in our repo. Similar to this one.

https://github.com/kubernetes/kubernetes/blob/master/logo/usage_guidelines.md
I can get you some help from marketing to get this published.

geeknoid:
So:

- I think our Twitter account already has a different font, which is
unfortunate.

- To get serifs on the I, we need a different font, here and elsewhere.

- We have an alternate brand logo available if we want to use it.

- Finally, I strongly suspect that this is not an actual problem. I haven't
heard of anyone actually getting confused by this. Heck, the web site is
istio.io. I think this is merely a theoretical concern.

What I'd like us to do at this point is make the Twitter font look like
what's on istio.io today and then leave things as-is. If we want, I can add
a popover so that if you hover the mouse over the name on istio.io, it
would show it in the popup using a serif font.

ldemailly:
> I suggest we make readable upper case I as our default font

this is what this PR does and it's already readable/similar on twitter

so I think that means merge this rather than not

zinuga:
lets merge this

PR #787, opened 2017-12-11
Cleanups to feature stages tables

rshriram:
Reordering fields, for clarity and some trimming.

### DIFF
--- a/_docs/welcome/feature-stages.md
+++ b/_docs/welcome/feature-stages.md
@@ -30,28 +30,18 @@ within the product, not to the product as a whole. Here is a high level descript
 
 Below is our list of existing features and their current phases. This information will be updated after every monthly release.
 
-### Config / APIs
-
-
-| Feature           | Phase      
-|-------------------|-------------------
-| [Mesh-wide Settings](https://istio.io/docs/reference/config/service-mesh.html) | Alpha
-| [Basic config validation](https://github.com/istio/istio/issues/1894)    | Alpha
-
-
 ### Networking
 
 | Feature           | Phase      
 |-------------------|-------------------
-| [Ingress Controller](https://istio.io/docs/tasks/traffic-management/ingress.html)         | Alpha
-| [Basic Egress](https://istio.io/docs/tasks/traffic-management/egress.html)               | Alpha
-| [Protocols HTTP 1.1](https://github.com/istio/istio/blob/master/pilot/model/service.go#L104)         | Beta
-| [Protocols HTTP 2.0](https://github.com/istio/istio/blob/master/pilot/model/service.go#L101)        | Alpha
-| [Protocols gRPC](https://github.com/istio/istio/blob/master/pilot/model/service.go#L97)     		   | Alpha
-| [Protocols TCP](https://github.com/istio/istio/blob/master/pilot/model/service.go#L107)         	   | Alpha
+| [Kubernetes Ingress](https://istio.io/docs/tasks/traffic-management/ingress.html)         | Alpha
+| [Egress Rules](https://istio.io/docs/tasks/traffic-management/egress.html)               | Alpha
+| [Protocols HTTP/1.1 & HTTP2](https://github.com/istio/istio/blob/master/pilot/model/service.go#L104)         | Beta
+| [Protocols gRPC](https://github.com/istio/istio/blob/master/pilot/model/service.go#L97)     		   | Beta
+| [Protocols TCP](https://github.com/istio/istio/blob/master/pilot/model/service.go#L107)         	   | Beta
 | [Protocols: WebSocket](https://github.com/istio/istio/blob/master/pilot/proxy/envoy/testdata/websocket-route.yaml.golden)       | Alpha
 | [Protocols: MongoDB](https://github.com/istio/istio/blob/master/pilot/model/service.go#L134)         | Alpha
-| [Basic Routing Rules](https://istio.io/docs/tasks/traffic-management/request-routing.html)      | Alpha
+| [Traffic Routing Rules](https://istio.io/docs/tasks/traffic-management/request-routing.html)      | Alpha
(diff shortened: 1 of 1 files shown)


_docs/welcome/feature-stages.md
```diff
@@ -98,22 +84,4 @@ Below is our list of existing features and their current phases. This informatio
 | [Istio mTLS](https://istio.io/docs/concepts/security/mutual-tls.html)         | Beta
 | [Pluggable Key/Cert Support for Istio CA](https://istio.io/docs/tasks/security/plugin-ca-cert.html)        | Beta
 
-### User Interface
```
jasminejaksic: We should keep this section

mandarjog:
@rshriram Can you mention the rationale behind trimming specific items.
I can see a potential rationale, but would be good to clarify it.

_docs/welcome/feature-stages.md
```diff
@@ -30,28 +30,18 @@ within the product, not to the product as a whole. Here is a high level descript
 
 Below is our list of existing features and their current phases. This information will be updated after every monthly release.
 
-### Config / APIs
```
smawson: Seems fine to remove these, not really well specified enough as features.

_docs/welcome/feature-stages.md
smawson: +1, we want to keep these.

_docs/welcome/feature-stages.md
```diff
@@ -98,22 +84,4 @@ Below is our list of existing features and their current phases. This informatio
...
-### User Interface
-
-
-| Feature           | Phase       
-|-------------------|-------------------
-| [Grafana Dashboard](https://istio.io/docs/tasks/telemetry/using-istio-dashboard.html) | Alpha
-| [Service Graph](https://istio.io/docs/tasks/telemetry/servicegraph.html)    | Alpha
-
-
-### Performance and Scalability
```
smawson: I think we want to clean this up rather than remove, should figure out how to word these.

jasminejaksic:
Sriram, I'm fine with removing config and reordering/combining network and env features. That said, I think we should keep UI and performance sections.

_docs/welcome/feature-stages.md
linsun: +1

linsun: agreed.

_docs/welcome/feature-stages.md
linsun: For UI, i think it is better to call it Observability and have Zipkin and Jaeger in there too.

_docs/welcome/feature-stages.md
linsun: given we are alpha and we said performance is random for alpha and beta... it may generate more confusion at this point.

rshriram:
if we classify alpha/beta/stable as stages, and associate performance metrics with them, what does it mean to say performance and scalability and have alpha/beta/stable features under it? Given that everything else also has alpha/beta/stable.

Reproducible oss characterization tool: I don't think we should be endorsing reproducible open source software benchmarking as a feature of a service mesh. That is not istio's charter. There are so many benchmarking tools that it is unclear why this was created or ended up in Istio's feature set. It certainly has not gone through any design review or any discussion. These are our internal tools. They should not end up in a user visible page.

Initial scenarios, manual baseline: What does an alpha categorization for initial scenarios with manual baseline mean to a retail customer evaluating Istio? is there an advanced scenarios with automated baseline that would become beta/stable and how would this be useful to them? 

Mixer bottom up perf - do we have a mixer top down perf? (first is the use of informal language "perf" on a public facing site), and what does it mean to have mixer performance called out in a section, when there are a ton of mixer features called out? 

It almost seems pointless to define alpha/beta/stable characterization and arbitrarily assign them to aspects of development that don't fit into this characterization. It might be better to simply state it as "Unsure, might change", or something to that effect.

The UI is not something we sell. Its a telemetry dashboard and as such, it falls under telemetry. Keep in mind that its also not our contribution. Someone else did all the grafana javascript, zipkin/jaeger design etc. We are merely populating it and defining the queries. Its a part of telemetry and tracing in anyone's mind, as there is no other sensible way to see if these are working (except for tracing queries).

I would have trimmed this down further, but decided to keep the changes minimal. Our internal feature tracking boards from the google sheets are not easy for public consumption. This is a user facing site and people are evaluating whether or not to use Istio. What should be here are APIs and their versions, high level features that are easy to grasp by folks from enterprise (not meant for PMs/other istio developers/managers as they have more context into development compared to a casual onlooker). The classification is also inconsistent with the diagrams we present to people (mixer, pilot, security) - while this talks in terms of a subset of working groups.

geeknoid:
The issue is that "perf" is not a feature, it's a characteristic. Imagine
if one of our goals was to make our binaries as small as possible. We
certainly wouldn't have a row in a table that says "size is Alpha", "size
is Beta", etc. Similarly, perf is not something we can rank like that.

We can talk about perf tooling or perf infrastructure. But that's
irrelevant to our customers (people reading istio.io) and is an internal
implementation detail we shouldn't be discussing. We really should be
removing the perf section as it exists, it's not relevant to the audience
of the web site.

smawson:
Give me a few days to put together a strawman for a better way to present the information we want to present. I agree we need to clean this up but think it warrants a broader discussion.

For example we need to decide if we arrange by task/use case or by working group.

I think I actually like use-case driven:
- Traffic Management
- Security
- Observability
- Extensibility
- Usability

Not sure if those are the right categories though.

linsun:
I'd vote for user scenarios driven.  I think Traffic Management, Security and Observability are pretty clear.  Where do you see policy enforcement go?  I would like one for HA / Scalability.

geeknoid:
Given PR #825, should we close this PR?

Issue #363, opened 2017-06-05
Add a presubmit check to prevent hardcoded http(s)://istio.io/...


ldemailly:
```
ldemailly-macbookpro:istio.github.io ldemailly$ git grep  "http://istio.io"
_docs/reference/commands/istioctl.md:See http://istio.io/docs/reference for an overview of routing rules
scripts/generate-cli-docs.sh:        -e "s,http://istio.io/\(.*/\)\(.*\).html,[\2]({{home}}/\1\2.html),g"
ldemailly-macbookpro:istio.github.io ldemailly$ git grep  "https://istio.io" | grep -v README.md
_docs/concepts/network-and-auth/auth.md:1.  [Istio Manager](https://istio.io/docs/concepts/traffic-management/manager.html) generates the config with proper key and certificate and secure naming information, 
_docs/concepts/network-and-auth/auth.md:in [Istio Mixer](https://istio.io/docs/concepts/policy-and-control/mixer.html) such that photo-frontend cannot access datastore.
_docs/reference/config/traffic-rules/manager-swagger.json:          "description": "Route-Rule or Destination Policy spec schemas can be found here: https://istio.io/docs/reference/api/traffic-rules/"
_docs/reference/contribute/writing-a-new-topic.md:https://istio.io/docs/tasks/tbd.html
_docs/tasks/istio-auth.md:(refer [here](https://istio.io/docs/concepts/network-and-auth/auth.html#identity) for more information). 
_docs/tasks/istio-auth.md:Please check secure naming [here](https://istio.io/docs/concepts/network-and-auth/auth.html#workflow) for more information
_posts/2017-05-24-0.1-announcement.md:Google, IBM, and Lyft are proud to announce the first public release of [Istio](https://istio.io/): an open source project that provides a uniform way to connect, secure, manage and monitor microservices. Our current release is targeted at the [Kubernetes](https://kubernetes.io/) environment; we intend to add support for other environments such as virtual machines and Cloud Foundry in the coming months.
_posts/2017-05-24-0.1-announcement.md:Istio is a completely open development project. Today we are releasing version 0.1, which works in a Kubernetes cluster, and we plan to have major new releases every 3 months, including support for additional environments. Our goal is to enable developers and operators to rollout and operate microservices with agility, complete visibility of the underlying network, and uniform control and security in all environments. We look forward to working with the Istio community and our partners towards these goals, following our [roadmap](https://istio.io/docs/reference/release-roadmap.html). 
[...]

PR #789, opened 2017-12-12
Some https manual fixes

ldemailly:
Related to #363

Source fix is in
https://github.com/istio/istio/pull/2093/files

### DIFF
--- a/_docs/reference/commands/istioctl.md
+++ b/_docs/reference/commands/istioctl.md
@@ -24,8 +24,8 @@ Available routing and traffic management configuration types:
 
 	[route-rule ingress-rule egress-rule destination-policy]
 
-See http://istio.io/docs/reference for an overview of routing rules
-and destination policies.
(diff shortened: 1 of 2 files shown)


_docs/reference/commands/istioctl.md
```diff
@@ -24,8 +24,8 @@ Available routing and traffic management configuration types:
 
 	[route-rule ingress-rule egress-rule destination-policy]
 
-See http://istio.io/docs/reference for an overview of routing rules
-and destination policies.
+See [https://istio.io/docs/reference/](https://istio.io/docs/reference/) for an
```
geeknoid: Should be {{home}} instead of https://istio.io

ldemailly: The text is verbatim from istioctl output with the hyperlink a bonus. but I will change it if needed.

geeknoid:
I know, it's part of the transformation we need to perform when we generate
the markdown files from the original generated docs.

PR #794, opened 2017-12-13
Remove istio-egress from kubernetes quick start guide.

gyliu513:
Fixed https://github.com/istio/istio.github.io/issues/792

/cc @rshriram


linsun:
CLAs are in place.

PR #795, opened 2017-12-13
Add information about grafana not displaying metrics

sdake:
If the server time is behind the client time at minimum 6 minutes,
the Grafana dashboard will not display metrics to the web client (e.g. Chrome).

This is clearly some type of bug - the time of the client should have
zero bearing on the metrics displayed by the server.  Note the metrics
are recorded.  I am unclear if the metric timestamps are recorded correctly
in the server.

See: https://github.com/istio/istio/issues/1954 for more details.

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -57,6 +57,17 @@ server {
 }
 ```
 
+## No grafana output when connecting from a local web client to Istio remotely hosted
+
+Validate the client and server date and time match.
+
+The time of the web client (e.g. Chrome) affects the output from Grafana. A simple solution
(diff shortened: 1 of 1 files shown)


linsun:
I think this is a common problem for viewing data from grafana (e.g. nothing istio unique), as you as essentially query a time window that doesn't have data from the server perspective. :)

sdake:
Possibly.  What tipped me off was the mild warning in prometheus that my client time and server time were skewed, resulting in possibly incorrect data.  Prometheus however seemed to handle it when I used the prom graphing feature.

Issue #702, opened 2017-11-03
Ingress tutorial - route rules are not working 


wstrange:
Following the ingress tutorial at . https://istio.io/docs/tasks/traffic-management/ingress.html 

minikube 0.22.3, running Kube 1.8.0 , istio-0.2.10/

Kube started with: 

minikube start \
--extra-config=apiserver.Authorization.Mode=RBAC \
--extra-config=apiserver.Admission.PluginNames="Initializers,NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,GenericAdmissionWebhook,ResourceQuota"

1) minor problem with namespaces in the example. I had to add -n default to the istioctl commands. 

2) The ingress works OK (I get routed to the page), but the deny rule does not work.. All pages can be accessed, even non /status ones

The side car looks like it was injected OK into the httpbin pod.

wstrange:
Do you want me to refile this under  https://github.com/istio/issues?

rshriram:
Fixed by #796. Fault injection does not work in ingress.

Issue #731, opened 2017-11-23
Istio fault injection routing rule from Ingress


ashth231:
I am trying to get the fault injection abort rule to work from Ingress ( I use traefik on my local minikube)  with my local application. For instance, say Grafana, 

**Service definition:** 
```
apiVersion: v1
kind: Service
metadata:
  name: myhost-grafana
  labels:
    app: grafana
    chart: grafana-0.1.0
    release: foiled-panda
    heritage: Tiller
spec:
  type: ClusterIP
  # type: NodePort
  ports:
    - port: 3000
      targetPort: 3000
      protocol: TCP
      name: grafana
  selector:
    app: grafana
```

**Ingress:**
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: myhost-grafana
  labels:
    app: grafana
    chart: grafana-0.1.0
    heritage: Tiller
  annotations:
      0: "map[kubernetes.io/ingress.class:istio]"
      1: "map[kubernetes.io/ingress.class:traefik]"
spec:
  rules:
    - host: "grafana.local"
      http:
        paths:
          - path: /
            backend:
              serviceName: myhost-grafana
              servicePort: 3000
```

**Istio abort rule:**
```
apiVersion: config.istio.io/v1alpha2
kind: RouteRule
metadata:
  name: deny-route
spec:
  destination:
    name: myhost-grafana
  match:
    source:
      labels:
        app: grafana
  precedence: 1
  route:
  - weight: 100
  httpFault:
    abort:
      percent: 100
      httpStatus: 403
```
But all I see when I curl is ,
```
$ curl -I http://grafana.local
HTTP/1.1 200 OK
Content-Type: text/html; charset=UTF-8
Date: Thu, 23 Nov 2017 12:05:47 GMT
Set-Cookie: grafana_sess=195f2cc07373ac95; Path=/; HttpOnly
```

I also tried curl from within the minikube vm with the  endpointIP:port . The output is same. 
May be am missing some configuration. Has someone got this working for your application?  Any help/suggestions would be greatly appreciated.

rshriram:
fixed by #796 .. Fault injection is not available in ingress.

PR #796, opened 2017-12-13
Fix ingress docs

rshriram:


rshriram:
Fixes #702, #731 
Also, @rfevang , we support dots in the paths now. So I removed the gRPC specific workarounds you added. Please let me know if I missed anything.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -58,79 +60,16 @@ rules.
...
-     - weight: 100
-   EOF
-   ```
-  You can use other features of the route rules such as redirects,
-  rewrites, regular expression based match in HTTP headers, websocket
-  upgrades, timeouts, retries, and so on. Please refer to the
-  [routing rules]({{home}}/docs/reference/config/traffic-rules/routing-rules.html)
-  for more details.
-
+   match, specifically a configuration of the form (`prefix: /`).
```
frankbu: specifically a [rule match configuration]({{home}}/docs/reference/config/traffic-rules/routing-rules.html#matchrequest) of the form

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,107 +210,27 @@ rules.
...
-Ingresses provide gateways for external traffic to enter the Istio service mesh
-and make the traffic management and policy features of Istio available for edge services.
+Ingresses provide gateways for external traffic to enter the Istio service
+mesh and make the traffic management and policy features of Istio available
+for edge services.
+
+The servicePort field in the Ingress specification can take a port number
+(integer) or a name. The port name must follow the Istio port naming
+conventions (e.g., `grpc-*`, `http2-*`, `http-*`, etc.) in order to
+function properly. The name used must match with the port name in the
```
frankbu: must match the port name

_docs/tasks/traffic-management/ingress.md
```diff
@@ -58,79 +60,16 @@ rules.
...
-     route:
-     - weight: 100
-   EOF
-   ```
-  You can use other features of the route rules such as redirects,
-  rewrites, regular expression based match in HTTP headers, websocket
-  upgrades, timeouts, retries, and so on. Please refer to the
-  [routing rules]({{home}}/docs/reference/config/traffic-rules/routing-rules.html)
-  for more details.
-
```
frankbu: When you added the above, you deleted the section from the 0.1 version of the task that showed how routerules work in conjunction with ingress: https://github.com/istio/istio.github.io/blob/release-0.1/_docs/tasks/ingress.md#setting-istio-rules-on-an-edge-service

Maybe now that you're removing it, we should put that part back?

rshriram: Yeah I saw that. I am trying to intentionally not do it, since different aspects of routing rules do not work properly with Ingress.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,107 +210,75 @@ rules.
...
+             prefix: /status/ #must match the path specified in ingress spec
+                 # if using prefix paths (/status/.*), omit the .*.
+                 # if using exact match, use exact: /status
      route:
-     - weight: 100
-       destination:
-         name: bar-service
-   EOF      
+     - labels:
+         version: v1
```
frankbu: This won't work because the `httpbin` service has no version labels. Previously I used a timeout rule as the example.

rfevang:
Awesome! Does dots work with a released version (i.e. 0.3.0), or will it be in an upcoming version?

rshriram:
On Thu, Dec 14, 2017 at 9:14 AM Rune Fevang <notifications@github.com>
wrote:

> Awesome! Does dots work with a released version (i.e. 0.3.0), or will it
> be in an upcoming version?
>
i think it should work with 0.3. If not, in next, as we basically decided
to dump regex and just do path and prefix. That allows dots. The code is in
master.

_docs/tasks/traffic-management/ingress.md
rshriram: That was just an example though.. Do you want it to be accurate?

_docs/tasks/traffic-management/ingress.md
```diff
@@ -269,107 +214,78 @@ rules.
...
+conventions (e.g., `grpc-*`, `http2-*`, `http-*`, etc.) in order to
+function properly. The name used must match with the port name in the
+backend service declaration.
 
-In the preceding steps we created a service inside the Istio service mesh and showed how
-to expose both HTTP and HTTPS endpoints of the service to external traffic.
-We also showed how to control the ingress traffic using an Istio route rule.
+In the preceding steps we created a service inside the Istio service mesh
+and showed how to expose both HTTP and HTTPS endpoints of the service to
+external traffic.
```
frankbu: put back: 
We also showed how to control the ingress traffic using an Istio route rule.

PR #809, opened 2017-12-19
Update feature-stages.md

oaktowner:
Improving some phrasing, reordered the tables to have all beta features at the top.


jasminejaksic:
The alpha/beta content needs to be grouped by use-cases and needs to reflect 0.4 release. I will submit a different PR for that. 
Approving and merging this PR because of the changes Dan made to the phase definition.

Issue #802, opened 2017-12-15
Update metrics docs to account for "istio" namespacing


douglas-reid:
As reported on mailing-lists, our docs are now out-of-date with respect to metrics naming in Prometheus. Need to update.

PR #811, opened 2017-12-19
Update Metrics docs to reflect istio_ namespace.

jeffmendoza:
Update to reflect istio/istio#1325

Fixes: #802


jeffmendoza:
cc @douglas-reid

PR #816, opened 2017-12-21
Update attribute description for request.headers for gRPC.

qiwzhang:


qiwzhang:
@geeknoid

wora:
HTTP/2 headers can be binary data while attributes are string map. Do we
have a problem?

qiwzhang:
@wora It is a good point.  We may have to base64 encode the binary header.  Here is the issue to track it.
https://github.com/istio/proxy/issues/749

PR #818, opened 2017-12-29
Switch uses of <figure> to {% include figure.html %}

geeknoid:
figure.html now contains a bit of magic to substantially improve
the load time and visual appearance of pages that contain imagery.
The magic pre-allocates space on the page of the exact size needed
for the image, even though the image isn't loaded yet. The result
is that the page can be layed out once, layout doesn't need to
happen again for each image that gets loaded.

This is part of a bigger PR which I'm pushing in pieces to
make it easier to deal with.


_blog/posts/2017/0.1-announcement.md
```diff
@@ -36,18 +36,19 @@ Google, IBM and Lyft joined forces to create Istio from a desire to provide a re
...
-    </a>
-</figure>
+{% include figure.html width='100%' ratio='55.42%'
+    img='./img/istio_grafana_dashboard-new.png'
+    alt='Grafana Dashboard with Response Size'
+    title='Grafana Dashboard with Response Size'
+    caption='Grafana Dashboard with Response Size'
+    %}
+
+{% include figure.html width='100%' ratio='29.91%'
```
ldemailly: those numbers, 29.91% - where are they coming from ?

geeknoid: It's (image height / image width) * 100. I mentioned this in the updated instructions on how to insert an image.

With figure.html, the author specifies the relative width of the image as compared to the surrounding content. The author also specifies the aspect ratio of the image. This is enough info to be able to calculate the actual width & height of the image during layout.

geeknoid:
I didn't think links on images was a good thing. They don't bring you to a new page, just point the browser to the same image. You can right-click on the image and select "Open Image in New Tab" to get the same effect.

Do you think those links are useful?

_blog/posts/2017/0.1-announcement.md
ldemailly: can’t a computer calculate that ? preserving the aspect ratio should be the default, this seems error prone (and cumbersome)?

ldemailly:
I thought the clickability helped see the images that are zoomed out to fit/in full resolution- maybe a js pop up / overlay would be even fancier way to get that. it doesn’t harm to have the link for now ?

geeknoid:
Yes, a computer could calc the aspect ratio. Unfortunately, the tool chain
we're using doesn't provide a good place to hook in.

If we don't specify the aspect ratio statically, then the browser needs to
figure it out dynamically. To figure it out, it needs to load the SVG file.
This means that the browser will do the layout, and then when the SVG is
done loading, the browser will redo the layout, causing to flickering and
increased overall load time.

Ideally, there would be a plugin in Jekyll that would extract this
information automatically. Alas, this doesn't exist. And even if I wrote
one, we can't load custom plugins to GitHub Pages :-(

It's a tradeoff between having a nicer looking web site (especially on
mobile) and a few seconds of extra work when someone adds an image to the
site. If the aspect ratio is incorrect, the worst thing that happens is
that the image will be squashed.

geeknoid:
I just looked at a few web sites and see that clicking on images is done by some sites, while others don't. Kubernetes.io doesn't, but the Envoy site does.

Anyway, since you like the links, I'll add 'em back. No biggie.

geeknoid:
Turns out that only a few of the figures on istio.io are clickable. Most
are not...

geeknoid:
PTAL.

All images on the site are now clickable.

PR #823, opened 2017-12-29
Fixing a broken link on istio.io

jasminejaksic:
Adding redirect from https://istio.io/docs/reference/feature-stages.html. 
I understand the content needs to be updated but that's orthogonal to fixing a broken link.

### DIFF
--- a/_docs/welcome/feature-stages.md
+++ b/_docs/welcome/feature-stages.md
@@ -6,8 +6,9 @@ order: 10
 
 layout: docs
 type: markdown
-redirect_from: "/docs/reference/release-roadmap.html"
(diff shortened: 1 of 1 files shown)


_docs/welcome/feature-stages.md
```diff
@@ -7,6 +7,7 @@ order: 10
 layout: docs
 type: markdown
 redirect_from: "/docs/reference/release-roadmap.html"
+redirect_from: "/docs/reference/feature-stages.html"
```
ldemailly: try:
```
redirect_from:
  - "/docs/reference/release-roadmap.html"
  - "/docs/reference/feature-stages.html"
```
because currently, it seems it's last one wins and it loses the current redirect

PR #826, opened 2018-01-03
Major update of the web site's infrastructure.

geeknoid:
Four major themes to this work:

* Performance. This new site is fast. It loads considerably faster, especially on mobile.
Every pixel is drawn exactly once. Clicking around doesn't result in things jiggling everywhere.

* Navigation. I wanted to make it easier to get around the site, especially on mobile. To that end, we now have:

  - Sticky navigation elements. The top nav bar, the left sidebar and the right TOC are now sticky and remain
on screen even as page content is scrolled.

  - Tabs in the Docs section have been removed and integrated as accordion items in the left sidebar.

* Layout. I moved a bit of content around to make it easier to find and give us some room to grow:

  - The Help page is now replaced with a full Help section. This provides a home for the FAQ, glossary,
bug reporting, and troubleshooting pages.

  - The About page is now replaced with a full About section. This provides a home for the release notes, roadmap, upcoming "in the news" section, etc.

* Design. I tried to improve consistency across the site and provide a slightly lighter feel by minimizing the impact of some design elements. Specifics:

  - Font sizes are now consistent across the site.

  - The site does a much better job at adjusting to different screen sizes which makes the mobile experience better.

  - Accessibility is improved by adding many attributes for screen readers, and by ensuring the site scales properly based on user-selected fonts.


linsun:
The staging site looks great on my laptop and mobile, thank you Martin, great work!

PR #825, opened 2018-01-02
Updates to features page

jasminejaksic:
Grouped features by use-cases/tracks and modified them to reflect 0.4 release

PR #829, opened 2018-01-04
Updates to features page

jasminejaksic:
redoing #825


smawson:
LG other than the change to have platform support pulled out.

jasminejaksic:
Per our discussion, we will make additional changes (including separating platform support) in the next iteration.

PR #833, opened 2018-01-04
Add support for archiving istio.io contents.

geeknoid:
This change has no visible effect on anything, new behavior is controlled by the _data/istio.yml archive variable, and by removing the draft: front-matter entry in archive.md.


geeknoid:
Which errors? The change as it is shouldn't have any visible effect on
istio.io, no new errors should be there. So what are you thinking about
then?

ldemailly:
click the link ? draft are being rendered and participate in link checking, it only hides from nav

ldemailly:
also where is the dropdown of versions - to create the v0.1, v0.2.... I'd like to merge the menu/navigation first into those branches

geeknoid:
Oh, those errors :-)

I'll just remove the archive entries in _data/site.yml. Once the archives
exist, then we add these entries.

The way I implemented this is not via a dropdown. Instead, you navigate to
the About page, and there's a Site Archive page where you select the
version you want. It's simple and doesn't waste screen space in the header
for something which is not the prime use case. Of course, if we find that
customers wants to look at old versions all the time, we could add an
explicit dropdown like k8s has.

ldemailly:
i meant for people who end up on an archived version, to know to switch back to latest or another version

geeknoid:
​The model is like this:

   - istio.io has a Site Archive page which lists all available archived
   versions.


   - ​When the user clicks on an archive, they are sent to the archived
   site on archive.istio.io.


   - Archive sites use different coloring and specific markings to tell the
   user that they are archived sites. I used some gold coloring, as in "golden
   oldies".


   - If the user performs a search in an archived site, results will be
   limited to the archive site.


   - If the user navigates to the Site Archive page of an archived site,
   there will be a link pointing back to the Site Archive page of the main
   istio.io site. So individual archive sites don't reference other archive
   sites.

You can see a sample archive site at https://geeknoid.github.io/
istio.github.io/

ldemailly:
cool, I guess what I am suggesting is:
screen shot 2018-01-04 at 1 38 54 pm
would be dropdown instead of a link to the top of the archive
so people would go from 0.4 (should not be 0.4.0 btw) to latest or 0.5 directly from each page

ldemailly:
ditto for
screen shot 2018-01-04 at 1 40 26 pm

ldemailly:
by the way an interesting challenge is going to be how to redo that in the old framework for each branch pre your big change

geeknoid:
​That would be doable, but:

- Everytime a new archive is produced, we would have to build previous
archives such that they contain a new entry in their respective dropdowns.
Either that or we dynamically fetch the list of archives from somewhere in
JavaScript and render the dropdown on the fly accordingly.

- I just don't think browsing between archived versions is a key scenario.
You can access all the archived versions from the main Site Archive page,
so that seems to hit the 99% case for this stuff (which is I believe the
"I'm running 0.2 in my setup and I want to see the docs for that")
​

ldemailly:
well at least there should be a link to istio.io from all the archive headers

but yes it could get the list from javascript from 1 file at the top of archive.istio.io

geeknoid:
All right, there is now a dynamically updated dropdown. Instead of listing available archive versions in _data/istio.yml you need to have https://archive.istio.io/versions.txt.

_includes/archives.html
```diff
@@ -0,0 +1,22 @@
+
+<script>
+    fetch('https://archive.istio.io/versions.txt')
```
ldemailly: awesome !

ldemailly:
I have a ticket open with firebase, it's taking longer than expected to get the cert setup for archive.istio.io - but do you want to start applying this to 

https://github.com/istio/istio.github.io/tree/release-0.1

if you get that one working, I'll create 0.2, 0.3 and 0.4 and cherrypick there
(though I guess 0.4 can probably be == master as I don't think we have much 0.5 content yet ?)

ldemailly:
I have a first version:
https://github.com/istio/admin-sites/pull/34
https://archive.istio.io/

ldemailly:
The menu on the right is awesome, but I wouldn't change the color scheme - or at least not blue to yellow 

I think https://archive.istio.io/v0.2/docs/
screen shot 2018-01-05 at 3 44 02 pm

(+ the dropdown) is clear enough that it is not the latest

though "archive" is a bit long - maybe that part can be smaller

ldemailly:
So I added versions.txt generation/support and had to add CORS header too so it's downloadable from JS from anywhere/staging site

It doesn't render quite right:

https://geeknoid.github.io/istio.github.io/

screen shot 2018-01-05 at 5 48 15 pm

geeknoid:
Refreshed colors for archived sites: gold be gone.

ldemailly:
Thanks, oldie is still there though.
Also can you pass « rake test »/fix so circleci is happy?

geeknoid:
I'm not sure what to do about the rake test. I'm using an href of # which
is what I'm supposed to use for dropdowns, according to the Bootstrap docs.
But the test is claiming that # is not valid.

geeknoid:
Turns out I can just remove the # and everything seems to work.

ldemailly:
^

geeknoid:
Ready to commit.

_includes/header.html
```diff
@@ -12,7 +12,11 @@
 
         <a class="navbar-brand d-flex w-50 mr-auto" href="{{home}}/" style="visibility: {{visibility}}">
             <img class="logo" src="{{home}}/img/istio-logo.svg" alt="Istio Logo"/>
-            <span class="brand-name">Istio</span>
+            {% if site.data.istio.archive %}
+                <span class="brand-name">Istioldie {{site.data.istio.short_version}}</span>
```
ldemailly: Istioldie -> Istio Archive

_layouts/base.html
```diff
@@ -22,10 +22,18 @@
             <meta name="og:description" content="{{page.overview}}">
         {% endif %}
 
-        {% if page.title == 'Istio' %}
-            <title>Istio</title>
+        {% if site.data.istio.archive %}
+            {% if page.title == 'Istio' %}
+                <title>Istioldie {{site.data.istio.short_version}}</title>
+            {% else %}
+                <title>Istioldie {{site.data.istio.short_version}} / {{ page.title }}</title>
```
ldemailly: same

_sass/base/_archive_colors.scss
```diff
@@ -0,0 +1,4 @@
+// Colors for the archive version of the site
+$mainBrandColor: #466BB0;
+$secondBrandColor: #041A77;
```
ldemailly: can we not change the colors at all ?

ldemailly:
ping

geeknoid:
I am ready to check in. I disagree with your PR review comments and have no
intention to make those changes.

ldemailly:
> I disagree with your PR review comments and have no intention to make those changes.

let's recap what I think are on the objectives we can agree on:

a) provide navigation from versions to version
b) make it clear for visitor being on a specific version that indeed they are on that version and not the latest
c) respect our branding
d) provide the functionality to our users sooner than later

This PR's infrastructure is great (for a) and b)) and just needs a few characters of change to achieve c)

@jasminejaksic and I already mentioned `Istioldie` while cute is probably not right, and ditto about changing the colors. I'm not attached to `Istio Archive` but it's clearly (to me) a better, professional wording.

@zinuga what do you think ?

rshriram:
fwiw, I would be happy with the stuff as is. Its bad enough that we don't have a dedicated web dev to all this UX stuff. I am really grateful to @geeknoid for doing all this webdev work _on top of_ development/design/architecture etc. that he is doing related to Istio. If I was knowledgeable enough and resourceful enough to send my own PRs for site wide UX changes, I might argue for the wording (Istioldie -> Istio Archive or what not). But since this is volunteer contribution as a hobbyist, I will take what I can get. 

May be as a compromise, Isti'oldie ? It took me a couple of seconds to recognize the "oldie" in Istioldie (my eyes fell on Istio, then l, then "die"). 

A small amount of informalism is fine IMO, as its not really offending anyone nor indicating any form of inconsistency in messaging/docs.

jasminejaksic:
There is no right or wrong answer here, these are just design preferences. While I provided my opinion, it's just that, "an opinion", so it's up to geeknoid@ to take it or leave it.

smawson:
I agree with Shriram, while I like Istio Archive slightly better than Istio Oldie, I'm not the one putting in all this work so I'm fine with it as is.

We should just get this checked in and then iterate on a better name and/or better colors after. Launch and iterate.

PR #835, opened 2018-01-04
Add troubleshooting headless TCP service disconnects.

hollinwilkins:
This PR adds documentation on how to prevent Istio from disconnecting connections to headless TCP services every 15 minutes or so.

See conversation here for reference:
https://github.com/istio/istio/issues/506

### DIFF
--- a/_help/troubleshooting.md
+++ b/_help/troubleshooting.md
@@ -72,7 +72,7 @@ properly to point at the lab-based NTP services.
 The expected flow of metrics is:
 
 1. Envoy reports attributes to Mixer in batch (asynchronously from requests)
-1. Mixer translates the attributes from Mixer into instances based on 
+1. Mixer translates the attributes from Mixer into instances based on
    operator-provided configuration.
 1. The instances are handed to Mixer adapters for processing and backend storage.
 1. The backend storage systems record metrics data.
@@ -95,7 +95,7 @@ Mixer generates metrics for monitoring the behavior of Mixer itself.
 Check these metrics.
 
 1. Establish a connection to the Mixer self-monitoring endpoint.
-   
+
    In Kubernetes environments, execute the following command:
 
    ```bash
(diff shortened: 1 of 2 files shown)


_help/troubleshooting.md
```diff
@@ -325,3 +325,29 @@ Check your `ulimit -a`. Many systems have a 1024 open file descriptor limit by d
 ```
 
 Make sure to raise your ulimit. Example: `ulimit -n 16384`
+
+## Headless TCP Services Losing Connection from Istiofied Containers
+
+Envoy is reset every 15 minutes by the `istio-ca` deployment, as a part
```
andraxylia: If `istio-ca` is deployed, Envoy is restarted every 15 minutes to refresh the certificates, causing the disconnection of TCP streams or long-running connections.

hollinwilkins: @andraxylia Do you want me to make these changes?

ldemailly: yes please do

_help/troubleshooting.md
```diff
@@ -325,3 +325,29 @@ Check your `ulimit -a`. Many systems have a 1024 open file descriptor limit by d
...
+of this process it will restart Envoy sidecars causing them to
+disconnect TCP streams. In order to stop this from happening, you will
```
ldemailly: I would insert something like:
"While your TCP services should be resilient to rare connection close events (as your pods may move for a number of reasons instance), if you really need to keep long running connection, and until we release the version of istio that doesn't require envoy restart for certificate reload, you will have to disable mTLS [...]"
(though that's kind of a loooong sentence)

_help/troubleshooting.md
andraxylia: Yes, please rephrase, sorry if that was not clear. Envoy is not restarted by the istio-ca, but by the pilot agent.

_help/troubleshooting.md
hollinwilkins: @ldemailly 30 seconds isn't a long TCP connection. DB connections often stay open for hours by default before being recycled. Will add something about connection resilience as well.

hollinwilkins:
Hey, I made the requested changes, had to open a new PR because the old repository was deleted:

https://github.com/istio/istio.github.io/pull/870

ldemailly:
closing -> #870

PR #843, opened 2018-01-06
extend copyright to 2018

ldemailly:
preview:

screen shot 2018-01-05 at 6 21 51 pm

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2017 Istio Authors
+   Copyright 2017-2018 Istio Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


geeknoid:
I don't believe this is necessary. The copyright date simply needs to reflect the date of creation and doesn't need to include date of mutation. We're not going to patch all our source code copyrights in this way for example.

ldemailly:
it looks odd/a bit sloppy to have a site with an old copyright, usually it's at least current with the year

for instance:
cnn:
screen shot 2018-01-06 at 10 57 32 am

Facebook:
screen shot 2018-01-06 at 11 00 07 am

etc

geeknoid:
How about we just go with the latest year than, like CNN and Facebook and
K8s?

© 2018 The Kubernetes Authors

We could in fact make this automatic based on the site build time if we'd
like.

ldemailly:
changed to just 2018, yes extracting the year is I guess what the next line does

PR #844, opened 2018-01-06
Update for archiving

geeknoid:


_includes/archives.html
```diff
@@ -0,0 +1,19 @@
...
+            var e = document.getElementById('{{include.id}}');
+            var lines = data.split("\n");
+            for (var i in lines) {
+                if (lines[i] == "") {
+                    continue;
+                }
+
+                var li = e.appendChild(document.createElement("li"));
+                var a = li.appendChild(document.createElement("a"));
+                a.setAttribute("href", "https://archive.istio.io/" + lines[i]);
```
ldemailly: needs the trailing "/" here

_includes/nav.html
```diff
@@ -24,7 +24,11 @@
             <a class="navbar-brand" href="{{home}}/">
               <div>
                 <img src="{{home}}/img/logo.png" alt="Istio" width="36px" height="54px"/>
-                <span class="brand-name">Istio</span>
+                {% if site.data.istio.archive %}
+                  <span class="brand-name">Istioldie {{site.data.istio.short_version}}</span>
```
ldemailly: `Istioldie` was fine/fun for the top level of https://archive.istio.io/ but can we just have
`Istio archive v{{site.data.istio.short_version}}` here
  
Also I made it smaller than Istio in https://archive.istio.io/v0.2/docs/ for instance, which I think is useful
(add another span/class for the archive tag)

_layouts/base.html
```diff
@@ -21,13 +21,21 @@
     <meta name="og:description" content="{{page.overview}}">
     {% endif %}
 
-    {% if page.title == 'Istio' %}
-    <title>Istio</title>
+    {% if site.data.istio.archive %}
+        {% if page.title == 'Istio' %}
+            <title>Istioldie {{site.data.istio.short_version}}</title>
```
ldemailly: ditto

_sass/base/_archive_colors.scss
```diff
@@ -0,0 +1,6 @@
+// Colors for the archive version of the site
+$mainBrandColor: #466BB0;
+$secondBrandColor: #CEAF0C;
```
ldemailly: let's not change the colors it's confusing for users (and not good for the brand)
as long as each page has that "archive" marking it's good

geeknoid:
I fixed the missing /

I also removed the need to run JavaScript to display the dropdown. As a result, I'm happy to put the dropdown on every page, so I removed the special "site archives" page.

I'd like to keep the visuals as-is and we'll see what others think. Easily changed later.

ldemailly:
can you add a link or screenshots so "others" can decide on the look & feel

jasminejaksic:
+100 to no "Istioldies" and gradient (leave alone blue to yellow) - it reminds me of the 90s Flash-based-websites :)

geeknoid:
Refreshed colors for archived sites: gold be gone.

Fixed search in archived sites.

ldemailly:
thanks, oldie is still there

PR #850, opened 2018-01-07
Improve consistency of the coloring of a few elements with the site t…

geeknoid:
…heme.


ldemailly:
Do you have a before/after screenshot of some example of what this is changing ?

geeknoid:
It's changing:

- The highlight of buttons, like the GET STARTED button and the Copy button
on code blocks
- It changes the color of table headers from some off-theme blue to an
on-theme blue
- It changes the highlight color when you click on a link in the header
from red to our themed orange.

ldemailly:
now #851

Issue #852, opened 2018-01-08
broken nav


ldemailly:
not sure why/how/what happened but :

https://istio.io/docs/

shows "2017 Posts" (ie the blog TOC)

<img width="1133" alt="screen shot 2018-01-07 at 8 22 22 pm" src="https://user-images.githubusercontent.com/3664595/34658978-9d3172da-f3e8-11e7-8c3d-dccaea32f1f3.png">

ldemailly:
was fixed

PR #858, opened 2018-01-08
Improve color consistency of some elements.

geeknoid:
- This affects table headers, button highlighting, and nav bar link highlighting.
They use more consistent coloring now.

Same change as before, minus the change to _layouts/docs.html which damaged the sidebar in the docs section. This is a file I change locally when working on the site since it reduces site build time from ~40 seconds to ~12 seconds. I accidentally checked in that change with my previous PR.


ldemailly:
please see #852

PR #863, opened 2018-01-09
fix the definition of sleepWindow: …

vadimeisenbergibm:
minimum time the circuit will be **open** instead of **closed** as it was previously. The circuit is open = "no traffic".


vadimeisenbergibm:
Will be fixed by https://github.com/istio/api/pull/292

rshriram:
Sorry, opening this as well.. Until the new api docs are published, we need this fix in the webpage

vadimeisenbergibm:
@rshriram Fine, thank you.

PR #867, opened 2018-01-11
Add Envoy high CPU troubleshooting section.

hollinwilkins:
I was noticing high CPU usage of the Envoy proxy while idling. This was due to a high refresh rate of the configuration files. Added a troubleshooting section for anyone else experiencing this issue.


_help/troubleshooting.md
```diff
@@ -325,3 +325,27 @@ Check your `ulimit -a`. Many systems have a 1024 open file descriptor limit by d
...
+```
+# increase the field rdsRefreshDelay in the mesh and defaultConfig section
+# set the refresh interval to 30s
+kubectl edit configmap -n istio-system istio
+
+# restart pilot and wait a few minutes
+kubectl delete pods -n istio-system -l istio=pilot
+```
+
+Also make sure to reinject the sidecar into all of your pods, as
```
ldemailly: That shouldn't be necessary (the refresh rate after the first connection is controlled by pilot afaik)

kyessenov: No, we need to re-inject here since some refresh rates are only defined at startup (limitation of v1 APIs).

rshriram:
This is great. But do add a warning note that this will also delay the propagation of routing rule based changes and load balancer pool updates, for the specified time interval

PR #871, opened 2018-01-13
additional link to doc archives

ldemailly:


_docs/index.md
```diff
@@ -34,3 +34,5 @@ We're always looking for help improving our documentation, so please don't hesit
 [file an issue](https://github.com/istio/istio.github.io/issues/new) if you see some problem.
 Or better yet, submit your own [contributions]({{home}}/about/contribute/editing.html) to help
 make our docs better.
+
+Follow this link for the archive of the [earlier releases' documentation](https://archive.istio.io/).
```
ldemailly: this used to point to v0.1 as can be seen on https://archive.istio.io/v0.4/docs/welcome

geeknoid:
We should not be doing this. We have perfectly usable PRs that enable the full functionality. Sven, Shriram, Louis and myself are fine with the full-featured PRs and would like to see those merged in.

ldemailly:
the two can proceed in parallel - even with the menu/navigation an explicit link to the archive on the main landing page is useful

PR #873, opened 2018-01-16
add circuit breaking task for https://github.com/istio/istio.github.i…

christian-posta:
…o/issues/872

### DIFF
--- a/_docs/tasks/traffic-management/circuit-breaking.md
+++ b/_docs/tasks/traffic-management/circuit-breaking.md
@@ -1,0 +1,239 @@
+---
+title: Circuit Breaking
+overview: This task demonstrates the circuit-breaking capability for resilient applications
+
+order: 50
+
+layout: docs
+type: markdown
+---
+
+This task demonstrates the circuit-breaking capability for resilient applications. Circuit breaking allows developers to write applications that limit the impact of failures, latency spikes, and other undesirable effects of network peculiarities. This task will show how to configure circuit breaking for connections, requests, and outlier detection.
+
+## Before you begin
+
+* Setup Istio by following the instructions in the
+  [Installation guide]({{home}}/docs/setup/).
+  
+* Start the [httpbin](https://github.com/istio/istio/tree/master/samples/httpbin) sample
+  which will be used as the backend service for our task
+  
+  ```bash
+  kubectl apply -f <(istioctl kube-inject -f samples/httpbin/httpbin.yaml)
+  ```  
+    
+
+
+## Circuit breaker
+
+Let's set up a scenario to demonstrate the circuit-breaking capabilities of Istio. We should have the `httpbin` service running from the previous section. We'll want to make sure we set some route rules to route all traffic to `httpbin` with `version=v1`. We'll be setting the circuit-breaking settings by creating an Istio [destination policy]({{home}}/docs/reference/config/traffic-rules/destination-policies.html) but for destination policies to take effect, there needs to be a route rule for the destination first. Let's create the route rule:
+
+### Creating circuit breaking policies
+
+1. Create a default route rule to route all traffic to `v1` of our `httpbin` service:
+
+```bash
+istioctl create -f samples/httpbin/routerules/httpbin-v1.yaml
+```
+
+2. Create a [destination policy]() to specify our circuit breaking settings when calling `httpbin` service:
+
+   ```bash
+   cat <<EOF | istioctl create -f -
+   apiVersion: config.istio.io/v1beta1
+   kind: DestinationPolicy
+   metadata:
+     name: httpbin-circuit-breaker
+   spec:
+     destination:
+       name: httpbin
+       labels:
+         version: v1
+     circuitBreaker:
+       simpleCb:
+         maxConnections: 1
+         httpMaxPendingRequests: 1
+         sleepWindow: 3m
+         httpDetectionInterval: 1s
+         httpMaxEjectionPercent: 100
+         httpConsecutiveErrors: 1
+         httpMaxRequestsPerConnection: 1
+   EOF
+   ```
+   
+3. Verify our destination policy was created correctly:
+
+    ```bash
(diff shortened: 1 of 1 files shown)


_docs/tasks/traffic-management/circuit-breaking.md
```diff
@@ -0,0 +1,172 @@
+---
+title: Circuit Breaking
+overview: This task demonstrates the circuit-breaking capability for resilient applications
```
rshriram: one thought: Envoy's notion of circuit breaker is pretty confusing for folks from the netflix land. In fact, envoy's circuit breaker is close to bulk head/connection pooling, while the outlier detection is close to the circuit breaker that people have become familiar with.

In the new v1alpha2 APIs, I have renamed the circuit breaker to ConnectionPools and the outlier detection is called outlier detection. They are two separate things. Do you think it makes sense to rename connection pool to bullhead, and outlier detection to circuit breaker? And if we do so, this doc would be bulk heads and circuit breakers. Though in strict sense, its not really a bulk head, because Envoy's connection pool is per upstream host in the cluster (not per cluster). Also, the concept of a bulk head in an event driven sidecar does not really click, as bulk heads are typically useful when you do synchronous calls, with one thread per request. When you have a shared thread pool serving an  event stream, its something else.

Let me know what you think. I am okay just calling this circuit breaker if it simplifies the message for the audience.

christian-posta: These are great points. One thing I did like about the current config of "circuit breaker" is that it kinda abstracts/hides away the impedance mismatch between hystrix and envoy style circuit breaking -- though the abstraction is a bit leaky. I do like the idea to make "Bulkhead" and "Circuit breaking" more distinct in the configs vs connection throttling/pooling/outlier detection, etc. Guess for reactive calls, a bulkhead would just be a semaphore to count # of outstanding requests, etc. 

What if we have something like "circuit breaker" which is really outlier detection and then "connection limits" which is really envoy's circuit breaker?

rshriram: Thats basically what I have in v1alpha2. ConnectionPool (which is circuit breaker), and Outlier Detection (which is outlier detection). I could rename Outlier Detection to Circuit Breaker if it helps the messaging. Let me know and I will fix the istio/api

_docs/tasks/traffic-management/circuit-breaking.md
```diff
@@ -0,0 +1,172 @@
...
+    NAME                    KIND                                            NAMESPACE
+    httpbin-circuit-breaker DestinationPolicy.v1alpha2.config.istio.io      istio-samples
+    ```   
+    
+### Setting up our client
+
+Now that we've set up rules for calling the `httpbin` service, let's create a client we can use to send traffic to our service and see whether we can trip the circuit breaking policies. We're going to use a simple go client called [gobench](https://github.com/cmpxchg16/gobench). With this client we can control the number of connections, concurrency, and delays of outgoing HTTP calls. In this step, we'll set up a client that is injected with the istio sidecar proxy so our network interactions are governed by Istio:
+
+    ```bash
+    kubectl apply -f <(istioctl kube-inject -f samples/httpbin/sample-client/gobench-deploy.yaml)
```
ldemailly: As we chatted, let's use fortio instead (could also change httpbin btw but at least the driver)

Let me know if I can help with the setup/doc/...

christian-posta:
@ldemailly i've updated the docs to reflect changing to use fortio. thanks!

_docs/tasks/traffic-management/circuit-breaking.md
christian-posta: Yah @rshriram i think Outlier Detection to Circuit breaker would be a good move.

_docs/tasks/traffic-management/circuit-breaking.md
```diff
@@ -0,0 +1,237 @@
...
+    > 0.011 <= 0.012 , 0.0115 , 75.00, 1
+    > 0.012 <= 0.014 , 0.013 , 90.00, 3
+    > 0.016 <= 0.018 , 0.017 , 95.00, 1
+    > 0.018 <= 0.0194349 , 0.0187174 , 100.00, 1
+    # target 50% 0.0095
+    # target 75% 0.012
+    # target 99% 0.0191479
+    # target 99.9% 0.0194062
+    Code 200 : 19 (95.0 %)
+    Code 503 : 1 (5.0 %)
```
ldemailly: I guess this is the interesting part of the output, 5% of 503s

you could maybe call fortio with `-loglevel warning` to reduce the flood of output

christian-posta: I think that's a useful thing -- would love to see how much gets trimmed -- but i don't see any effect when i change the log level. may take a look at https://github.com/istio/fortio/issues/85 ?

_docs/tasks/traffic-management/circuit-breaking.md
```diff
@@ -0,0 +1,237 @@
...
+    Response Header Sizes : count 20 avg 218.85 +/- 50.21 min 0 max 231 sum 4377
+    Response Body/Total Sizes : count 20 avg 652.45 +/- 99.9 min 217 max 676 sum 13049
+    All done 20 calls (plus 0 warmup) 10.215 ms avg, 187.8 qps
+    ```
+    
+We see almost all requests made it through! The istio-proxy does allow for some leeway. Let's bring the number of concurrent connectiosn up to 3:
```
frankbu: s/connectiosn/connections/

_docs/tasks/traffic-management/circuit-breaking.md
ldemailly: loglevel only trims logger output (lines with timestamp an level letter `I` `W`...)
I guess you can just delete the lines that feel noisy and put [...]

christian-posta:
Fixed spelling @frankbu thanks for catching that

PR #876, opened 2018-01-18
Update ServiceModel_Versions.svg

### LINKED ISSUE #869: Issue with _docs/concepts/traffic-management/img/pilot/ServiceModel_Versions.svg
Looking at the image https://istio.io/docs/concepts/traffic-management/img/pilot/ServiceModel_Versions.svg, I think that the box at the bottom right currently labelled << SvcB >> should be labelled << SvcB' >> (B prime) as it should be a different variant of the service << serviceB.example.com >>.

This B prime correction would be in line with the drawing https://istio.io/docs/concepts/traffic-management/img/pilot/TrafficManagementOverview.svg.

patricekrakow:
Rename the box at the bottom right from << SvcB >> to << SvcB' >> (B prime) as it should be a different variant of the service << serviceB.example.com >>.

Fixes #869


ldemailly:
edited your PR to say "Fixes #869" so it will auto close once merged

PR #881, opened 2018-01-25
Blog Post about HTTPS Egress Traffic in Istio

vadimeisenbergibm:
The blog post in HTML: https://vadimeisenbergibm.github.io/blog/2018/egress-https.html.

Any comments are welcome, including comments about English wording and grammar.


ldemailly:
just curious why the "0.5.0" - afaik it already works like that in 0.4 (and maybe 0.3?)
I have it live in https://github.com/istio/admin-sites/pull/32 for instance using 0.4.0

vadimeisenbergibm:
@ldemailly The functionality is there already in 0.2, in 0.5.0 `samples/bookinfo/kube/bookinfo-details-v2.yaml` used in the blog post, will be added.

ldemailly:
OT do you know if we have an issue tracking the awkwardness of having to change the app to make https work ?

vadimeisenbergibm:
@ldemailly 
> OT do you know if we have an issue tracking the awkwardness of having to change the app to make https work ?

No, I am not aware of an issue to track this. Note that the issue is similar to the in-mesh micro services that use https. Once https is used, Istio has a problem.

_blog/2018/egress-https.md
```diff
@@ -0,0 +1,173 @@
...
+
+The diagram below shows how the HTTPS traffic to external services is performed. On the top, a microservice outside an Istio Service Mesh
+sends regular HTTPS requests, encrypted end-to-end. On the bottom, the same microservice inside an Istio Service Mesh must send unencrypted HTTP requests inside a pod, which are intercepted by the sidecar Envoy proxy. The sidecar proxy performs TLS origination, so the traffic between the pod and the external service is encrypted.
+
+<figure><img src="img/https_from_the_app.svg" alt="HTTPS traffic to external services, from outside vs. from inside an Istio Service Mesh" title="HTTPS traffic to external services, from outside vs. from inside an Istio Service Mesh" />
+<figcaption>HTTPS traffic to external services, from outside vs. from inside an Istio Service Mesh</figcaption></figure>
+
+Here is how we code this behavior in the [the Bookinfo details microservice code](https://github.com/istio/istio/blob/master/samples/bookinfo/src/details/details.rb), using Ruby [net/http module](https://docs.ruby-lang.org/en/2.0.0/Net/HTTP.html):
+```ruby
```
mandarjog: The ruby details take a lot of explaining...
I think we should use the curl example here of accessing external site.
`curl http://myservice.com:443/`  because it is more accessible.

On a related note, I use debian egress rule so pods can get new software ...
```
apiVersion: "config.istio.io/v1alpha2"
kind: EgressRule
metadata:
  name: debian-egress-rule
spec:
  destination:
    service: "*.debian.org"
  ports:
    - port: 80
      protocol: http
    - port: 443
      protocol: https
```

vadimeisenbergibm: @mandarjog I wanted to show specifically how this can be done in code: you add some environment variable, e.g. `WITH_ISTIO` and depending on it you set schema to be `http` or `https`, or set some flag as in case of `net/http` in Ruby. Then you set this environment variable in kubernetes Deployment.

_blog/2018/egress-https.md
```diff
@@ -0,0 +1,173 @@
...
+<figure><img src="img/errorFetchingBookDetails.png" alt="Error Fetching Product Details" title="Error Fetching Product Details" />
+<figcaption>Error Fetching Product Details Displayed</figcaption></figure>
+
+
+Good news is that our application did not crash. With a good microservice design we do not have **failure propagation**. In our case the failing _details_ microservice does not cause the _productpage_ microservice to fail. Most of the functionality of the application is still provided despite the failure in the _details_ microservice. We have **graceful service degradation**: as you can see, the reviews and the ratings are displayed correctly, and the application is still useful.
+
+Hmm, so what might have gone wrong? Ah... The answer is that I forgot to enable traffic from inside the mesh to an external service, in this case to the Google Books web service. By default, the Istio sidecar proxies ([Envoy proxies](https://www.envoyproxy.io)) block all the traffic to the outside the cluster. To enable such traffic, we must define an [Egress Rule]({{home}}/docs/reference/config/traffic-rules/egress-rules.html).
+
```
linsun: I think it is worthwhile to highlight the default behavior of istio here.

_blog/2018/egress-https.md
```diff
@@ -0,0 +1,173 @@
...
+  name: googleapis
+  namespace: default
+spec:
+  destination:
+      service: "*.googleapis.com"
+  ports:
+      - port: 443
+        protocol: https
+EOF
+```
```
linsun: is it possible to have the yaml in git and create some e2e for this?  that can allow us to ensure the step in the blog is always current.

_blog/2018/egress-https.md
```diff
@@ -0,0 +1,173 @@
...
+
+Now accessing the web page of the application displays the book details without error:
+
+<figure><img src="img/externalBookDetails.png" alt="Book Details Displayed Correctly" title="Book Details Displayed Correctly" />
+<figcaption>Book Details Displayed Correctly</figcaption></figure>
+
+Note that our Egress Rule allows traffic to any domain matching _*.googleapis.com_, on port 443, using the HTTPS protocol. Let's assume for the sake of the example, that we must access multiple domains matching _*.gooogleapis.com_, not only _www.googleapis.com_ that is accessed by our code in reality. This **wildcard** feature allows us to enable traffic to multiple domains by a single Egress Rule.
+
```
linsun: could we explain if I need multiple domain matching, what do I need to do?

linsun: like googleapi.com and ibmapi.com

_blog/2018/egress-https.md
```diff
@@ -0,0 +1,173 @@
...
+```bash
+istioctl delete egressrule googleapis -n default
+```
+
+and see in the output of _istioctl delete_ that the Egress Rule is deleted:
+```
+Deleted config: egressrule googleapis
+```
+
+Accessing the web page now will produce the same error that we experienced before, namely _Error fetching product details_. As we can see, the Egress Rules are defined **dynamically**, as many other Istio configuration artifacts. The Istio Operators can decide dynamically which domains they allow the microservices to access. They can enable and disable traffic to the external domains on the fly, without redeploying the microservices.
```
linsun: why are we still getting the Error fetching product details??

_blog/2018/egress-https.md
```diff
@@ -0,0 +1,173 @@
...
+We set `WITH_ISTIO` environment variable to _"true"_ in the [kubernetes deployment of _details v2_](https://github.com/istio/istio/blob/master/samples/bookinfo/kube/bookinfo-details-v2.yaml), the `container` spec:
+```yaml
+env:
+- name: WITH_ISTIO
+  value: "true"
+```
+
+### Malicious Microservices Threat
+Another issue is that the Egress Rules are currently **not a security feature**, they only **enable** traffic to external services. For HTTP based protocols, the rules are based on domains. Istio does not check that the destination IP of the request matches the _Host_ header. It means that a malicious microservice inside a Service Mesh could trick Istio to allow traffic to a malicious IP. The trick is to set one of the domains allowed by some existing Egress Rule as the _Host_ header of the malicious request.
+
```
linsun: do we really want to disclosure this security issue here?

linsun:
I forgot to say, this is a really good start!  I enjoyed reading it and entered some comments while reading, thank you @vadimeisenbergibm for sharing your technical expertise in this area!

_blog/2018/egress-https.md
vadimeisenbergibm: Good idea. I will submit a PR to istio/istio, and then fix the blog. It can be done after the blog is published.

_blog/2018/egress-https.md
vadimeisenbergibm: I think we better should disclose everything. We explain in bold that egress rule is not a security feature and explain the possible attack. And it is already disclosed in this PR anyway (the PR itself is already a disclosure).

vadimeisenbergibm:
@linsun Regarding your changes request https://github.com/istio/istio.github.io/pull/881#pullrequestreview-91695138
> can we mention how this work when mTLS is enabled? also would be good mention briefly how this works when service A in the mesh accessing service B in the k8s cluster, e.g. is the current egress only scoped for traffic going out of the cluster or not.

I propose that we agree on the purpose of blog posts. In my opinion, there are two different kinds of blog posts:

1. Survey/article - a long post that covers a large topic, all the aspects.
2. Short post, similar to a twit - "here is my egress rule". 

In a short post, some narrow aspect should be explored and the simplest example should be given, to make the blog post as short and easy-to-understand as possible. I planned to make this blog post of the short kind: _here is a version of Bookinfo that must access an external web service and here is an Egress Rule that enables it_. However, since it is the first blog post on this topic, I have to explain all the relevant aspects to this simple scenario. This blog post got already too long, and I would like not to add to it any additional information, unless totally necessary, it makes it harder to read. Once people understand the simplest scenario, a series of blogs can be written on top of it, referencing the original blog post. Each small blog post can examine a different aspect: Istio and Kubernetes ExternalName services, Controlling TCP traffic, mTLS and egress traffic etc. Basically, any answer to a question on istio-users or stack overflow about Istio can be followed by a short blog post that would repeat the answer, maybe in a more elaborate way, with some example and a diagram.

vadimeisenbergibm:
@linsun Thank you for your comments! I think I addressed them all. More comments are welcome.

linsun:
@vadimeisenbergibm thank you for the replies.  I'm okay with them, except I do think we need a simple line or so to indicate whether this works with mTLS because that is the default istio install and a lot of users are interested in that.

_blog/2018/egress-https.md
vadimeisenbergibm: Added yamls and a smoke e2e test for testing the scenario of the blog:
https://github.com/istio/istio/pull/2923.

vadimeisenbergibm:
@linsun I have addressed the mTLS issue in the commits https://github.com/istio/istio.github.io/pull/881/commits/06751264ce5fa078cd524abafba67cd324913790, https://github.com/istio/istio.github.io/pull/881/commits/1d19fae7900763b2b7e84d1b68965a1bf9fb7208 and https://github.com/istio/istio.github.io/pull/881/commits/c809f91b523eaf4bf0f1b8904062ac31c28b43e8.

PR #884, opened 2018-01-26
Fixes https://github.com/istio/istio.github.io/issues/883 for incorre…

christian-posta:
…ct link to Envoy docs for mirroring


christian-posta:
cc @ldemailly

PR #889, opened 2018-01-27
Prepare for new generated reference content.

geeknoid:
- Stop using the prism functionality for syntax highlighting since
it doesn't handle embedded links in <pre> blocks, which prevents
cross-linking to type names properly. We now roll our own PRE block
handling.

- Stop drawing a box around inline code elements which substantially improves
the visuals, especially in tables.

- Improve appearance of tables to be a bit cleaner.


rshriram:
quick comments:
is it possible to shrink the font rather than wrap it ? I am assuming that most folks will read the reference docs on their big displays rather than over mobile. It feels a bit easier to comprehend "condition" in a single line compared to "cond" and "ition" wrapped..

Essentially, the explanation field (third column) is too long.

geeknoid:
Yes, I noticed the same issue with some of the tables. I think I can fix that, it'll require some experimentation.

PR #895, opened 2018-01-29
Update the ingress controller task.

myidpt:
Added the secure ingress verification steps.
Updated the response code (403 -> 404) in the verification.

### DIFF
--- a/_docs/tasks/traffic-management/ingress.md
+++ b/_docs/tasks/traffic-management/ingress.md
@@ -76,7 +76,7 @@ The following are the known limitations of Istio ingress:
    [rule match configuration]({{home}}/docs/reference/config/traffic-rules/routing-rules.html#matchcondition)
    of the form (`prefix: /`).
    
-## Verifying ingress
+### Verifying ingress
 
 1. Determine the ingress URL:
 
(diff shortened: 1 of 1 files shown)


myidpt:
@wattli @ldemailly Can you please take a look? Thanks.

_docs/tasks/traffic-management/ingress.md
```diff
@@ -205,21 +198,85 @@ The following are the known limitations of Istio ingress:
...
    ```
 
-   Set the INGRESS_HOST to point to the ip address and the
-   port number of the ingress service as shown earlier.
+   > Note: Because SNI is not yet supported, Envoy currently only allows a single TLS secret in the ingress.
+   > That means the secretName field in ingress resource is not used.
 
-   > Note: Envoy currently only allows a single TLS secret in the ingress
-   > since SNI is not yet supported. That means that the secret name field
-   > in ingress resource is not used, and the secret must be called
```
ldemailly: i believe you need this part, the cert must be called istio-ingress-certs

myidpt: OK. I can add it back to make it more explicit.

myidpt: Added to Line 170. Thanks.

PR #904, opened 2018-01-31
added "Relation to Security" section for egress rules task

vadimeisenbergibm:
explain that Egress Rules cannot be a security measure for the traffic to
the external services


_docs/tasks/traffic-management/egress.md
```diff
@@ -267,6 +267,9 @@ cloud provider specific knowledge and configuration.
    kubectl delete -f samples/sleep/sleep.yaml
    ```
 
+## Relation to Security
+Note that Egress Rules is **not a security feature**. Egress Rules enable access to external (out of the service mesh) services. However, they cannot prevent from malicious applications inside the service mesh to access forbidden external services. Egress Rules cannot apply access policies to the traffic to external services. For securing the traffic to external services other security mechanisms must be applied, for example a firewall.
+
```
rshriram: Some rephrasing:

## Egress Rules and Access control
Note that Egress Rules is **not a security feature**.  They enable access to external (out of the service mesh) services. It is upto the user to ensure that only trusted external services are represented via Egress rules, and deploy appropriate security instruments such as firewalls.

vadimeisenbergibm: OK, the final wording is:
## Egress Rules and Access Control
 Note that Istio Egress Rules are **not a security feature**. They enable access to external (out of the service mesh) services. It is up to the user to deploy appropriate security mechanisms such as firewalls to prevent unauthorized access to the external services. We are working on adding access control support for the external services.

ldemailly:
why not btw ? (why couldn't someone put mixer check rules on the egress for instance ?)
how is egress different from other rules?

vadimeisenbergibm:
@ldemailly Regarding your question:
> why couldn't someone put mixer check rules on the egress for instance ?

Please see https://goo.gl/s7fSqM, your comments are welcome .Currently, Mixer checks are performed on the inbound into-the-pod traffic by sidecar proxies. The egress traffic is outbound out-of-the-pod.

PR #902, opened 2018-01-31
0.5 release notes.

geeknoid:
These are the basic notes. Please:

- Make sure I didn't forget anything significant, or overstate trivial things.

- Look at the TODO items. Can you supply any missing doc links? If you own one of these features, considering adding a doc page to istio.io so we can link from it here.

Unfortunately, istio.io doesn't yet have updated adapter, template, and command-line reference documentation. Otherwise, I could point to these pages for several of the items. Hopefully, we'll have updated versions of these docs soon and I can come back and add links for these.

Thanks.

PR #912, opened 2018-01-31
Added release status / latest vs stable information

### LINKED ISSUE #910: add release status page
document the state:

latest release : 0.5.0

stable release: 0.2.12

possibly mentioning https://archive.istio.io/v0.2/docs/
and
`curl -L https://git.io/getIstio` (pointing to 0.2.12)
vs
`curl -L https://git.io/getLatestIstio `

cc @costinm per slack chat, @hklai (about when we want 0.5.0 to be mentioned)

ldemailly:
Fixes #910

### DIFF
--- a/_about/notes/index.md
+++ b/_about/notes/index.md
@@ -1,7 +1,5 @@
 ---
 title: Release Notes
-overview: What's been happening with Istio.
-
 order: 5
 
 layout: about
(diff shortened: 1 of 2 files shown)


ldemailly:
it looks like this:

screen shot 2018-01-31 at 2 40 23 pm

_data/istio.yml
```diff
@@ -1 +1 @@
-version: 0.4.0
+version: 0.5.0
```
ldemailly: this pr should go in with #902

ldemailly:
@geeknoid PTAL

_about/notes/index.md
```diff
@@ -1,7 +1,5 @@
 ---
 title: Release Notes
-overview: What's been happening with Istio.
```
geeknoid: Why remove the overview? That string is used for the hover text in the sidenav, and as the description test if you go to about/index.html

ldemailly: the text showed up on the page and looked strange with the new content - do you have a proposed new/better text or you want me to keep this one ?

ldemailly: updated. ptal

ldemailly:
updated preview:
screen shot 2018-01-31 at 3 59 59 pm

geeknoid:
Oh, I see. Things aren't quite right in general if you try to put actual
content on a section index page. It's fine to remove the overview then.

ldemailly:
our story on stable vs not etc isn't super solid so I didn't want to create a new proeminent page yet I think we do need to document the 2 vanity getIstio*.io urls as well as what was discussed on slack with respect to baking release so this existing page seems like a good compromise: not too prominent yet discoverable

PR #916, opened 2018-02-01
Update CLI docs and create reproducible process to update these docs.

geeknoid:
Staging: https://geeknoid.github.io/istio.github.io/docs/reference/commands/


scripts/grab_reference_docs.sh
```diff
@@ -39,12 +41,37 @@ locate_file() {
...
+# runs it to extract its command-line docs
+#
+# TODO: Even though this CDs into the source tree we've extracted, it's not actually
+# using that as input sources since imports are resolved through $GOPATH and such.
+# I'm not clear what voodoo is needed so I'm leaving this as-is for the time being
+get_command_doc() {
+    COMMAND_PATH=$1
+    COMMAND=$2
+
+    cd $COMMAND_PATH
```
ZackButcher: pushd/popd would be better so we can exit this function back in the directory it was invoked from.

Not blocking.

ldemailly:
istioctl gen command table isn’t rendering on the preview

also is there a way to extract the common flags and list them once rather than for each subcommand ?

there also seems to be leftover flag from glog (-v, -logmodule etc)

geeknoid:
I see istioctl gen-deploy listed, which is what I see when doing "istioctl --help".

Yes, there's a lot of glog crap left behind. I've got a plan to get rid of those (using the istio/glog repo). But for those, those are part of the command-line interfaces of our commands.

As for the layout, we can definitely update the layout. The logic that controls the layout is in istio/pkg/collateral/control.go. We can tackle that separately.

ldemailly:
as mentioned above
https://istio.io/docs/reference/commands/istioctl.html

ldemailly:
gen-deploy section is broken

geeknoid:
I pushed a quick fix for that. The real fix will involve updates to the
markdown production logic in istio/pkg/collateral.

ldemailly:
also all the argument types are missing ( < > needs escaping to &lt; I am guessing)

geeknoid:
Huh, I was expecting this to be taken of by markdown conversion. I guess
I'll need to emit entities...

ldemailly:
maybe there is an option to the table for md to do the escaping for us, not sure

ldemailly:
the upside I guess is you could use <i>italic</i> etc... to make it more lively

PR #918, opened 2018-02-01
remove v1alpha2 routing docs

rshriram:
We have not enabled v1alpha2 yet.


rshriram:
@geeknoid I am not sure if something else needs to be updated other than simply removing the files. LMK

geeknoid:
You should be instead marking the proto with the $hide_from_docs marker. Otherwise, this particular doc file will automatically come back.

PR #919, opened 2018-02-01
fix v1alpha2 docs

rshriram:


SerialVelocity:
@rshriram Why were the v1alpha2 docs removed? Are they not actually part of 0.5.0?

Issue #913, opened 2018-02-01
(cosmetic) new `command line example` is pink, not matching blockquote


ldemailly:
the new `command line example` style, without border also appears lighter/pink color and is a bit hard to make out on a white background:

![screen shot 2018-01-31 at 4 08 28 pm](https://user-images.githubusercontent.com/3664595/35654248-1c2dfe8a-06a1-11e8-9f96-08abdc2da669.png)

it should be the same color as in a blocquote (```):

![screen shot 2018-01-31 at 4 10 58 pm](https://user-images.githubusercontent.com/3664595/35654321-65f1af8a-06a1-11e8-918d-65d62c7f0c42.png)

`cluster-name` vs `export KUBECONFIG` above

ldemailly:
if you point me to the css I can make the change if you want

geeknoid:
I think that comes from syntax.css. But I'd have to track it down...

ldemailly:
fixed in #924

PR #923, opened 2018-02-01
Refresh CLI docs given updated generator.

geeknoid:
Staging: https://geeknoid.github.io/istio.github.io/docs/reference/commands/


ldemailly:
much nicer, thanks ! (modulo the pre existing #913)

geeknoid:
Ordering the different entries would require us to be able to introduce some form of metadata in the Cobra state. We could introduce an IstioCmd which is an extended version of a cobra.Command which holds this kind of info.

PR #924, opened 2018-02-02
Improve styling for <code> blocks.

geeknoid:
- Made <code> a bit smaller, a bit darker red, and substantially improve word break appearance in
tables.

### DIFF
--- a/_sass/base/_common.scss
+++ b/_sass/base/_common.scss
@@ -286,6 +286,12 @@ blockquote {
     background-color: $light-gray;
 }
 
+code {
+    color: #d14;
+    font-size: 75%;
+    word-break: normal;
+}
+
 .spacer {
     height: .1rem;
 


ldemailly:
screen shot 2018-02-01 at 5 50 49 pm

can we use #d14; like in the code blocks (or change the code blocks to be red but I think #d14 and in general any color is better when not pure red/green/blue)

ldemailly:
screen shot 2018-02-01 at 5 52 46 pm

_sass/base/_common.scss
```diff
@@ -286,6 +286,12 @@ blockquote {
     background-color: $light-gray;
 }
 
+code {
+    color: red;
```
ldemailly: can we change to
```
  color: #d14;
```
for consistency with highlight sd2 (see sshots)

ldemailly:
it looks great now thanks!

PR #928, opened 2018-02-02
add Control Egress TCP Traffic task

vadimeisenbergibm:
See it generated: https://vadimeisenbergibm.github.io/docs/tasks/traffic-management/egress-tcp.html

### DIFF
--- a/_docs/tasks/traffic-management/egress-tcp.md
+++ b/_docs/tasks/traffic-management/egress-tcp.md
@@ -1,0 +1,147 @@
+---
+title: Control Egress TCP Traffic
+overview: Describes how to configure Istio to route TCP traffic from services in the mesh to external services.
+
+order: 41
+
+layout: docs
+type: markdown
+---
+{% include home.html %}
+
+The [Control Egress Traffic]({{home}}/docs/tasks/traffic-management/egress.html) task demonstrated how external (out of the mesh) HTTP and HTTPS services can be accessed from applications inside the mesh. To remind: by default, Istio-enabled applications are unable to access URLs outside of the mesh. To enable such access, [egress rules]({{home}}/docs/reference/config/istio.routing.v1alpha1.html#EgressRule) must be defined, or, alternatively, [direct access to external services]({{home}}/docs/tasks/traffic-management/egress.html#calling-external-services-directly) must be configured.
+
+This task describes how to configure Istio to expose external TCP services to applications inside the Istio service mesh.
+
+## Before you begin
+
+* Setup Istio by following the instructions in the
+  [Installation guide]({{home}}/docs/setup/).
+
+* Start the [sleep](https://github.com/istio/istio/tree/master/samples/sleep) sample application which will be used as a test source for external calls.
+
+  ```bash
+  kubectl apply -f <(istioctl kube-inject -f samples/sleep/sleep.yaml)
+  ```
+
+  Note that any pod that you can execute `curl` from, would do.
+
+## Using Istio egress rules for external TCP traffic
+In this task we access `wikipedia.org` by HTTPS originated by the application. This task demonstrates the use case when the application cannot use HTTP with TLS origination by the sidecar proxy. Using HTTP with TLS origination by the sidecar proxy is described in the [Control Egress Traffic]({{home}}/docs/tasks/traffic-management/egress.html) task. In that task, `https://google.com` was accessed by issuing HTTP requests to  `http://www.google.com:443`.
+
+The HTTPS traffic  originated by the application will be treated by Istio as _opaque_ TCP. To enable such traffic, we define a TCP egress rule on port 443.
+
+In TCP egress rules as opposed to HTTP-based egress rules, the destinations are specified by IPs or by blocks of IPs in [CIDR notation](https://tools.ietf.org/html/rfc2317).
+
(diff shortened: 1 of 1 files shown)


_docs/tasks/traffic-management/egress-tcp.md
```diff
@@ -0,0 +1,147 @@
...
+overview: Describes how to configure Istio to route TCP traffic from services in the mesh to external services.
+
+order: 41
+
+layout: docs
+type: markdown
+---
+{% include home.html %}
+
+The [Control Egress Traffic]({{home}}/docs/tasks/traffic-management/egress.html) task demonstrated how external (out of the mesh) HTTP and HTTPS services can be accessed from applications inside the mesh. To remind: by default, Istio-enabled applications are unable to access URLs outside of the mesh. To enable such access, [egress rules]({{home}}/docs/reference/config/istio.routing.v1alpha1.html#EgressRule) must be defined, or, alternatively, [direct access to external services]({{home}}/docs/tasks/traffic-management/egress.html#calling-external-services-directly) must be configured.
```
ZackButcher: > To remind: by default, Istio-enabled applications are unable to access URLs outside of the mesh.

This is only true when mTLS is enabled, right?

_docs/tasks/traffic-management/egress-tcp.md
```diff
@@ -0,0 +1,147 @@
...
+  ```bash
+  kubectl apply -f <(istioctl kube-inject -f samples/sleep/sleep.yaml)
+  ```
+
+  Note that any pod that you can execute `curl` from, would do.
+
+## Using Istio egress rules for external TCP traffic
+In this task we access `wikipedia.org` by HTTPS originated by the application. This task demonstrates the use case when the application cannot use HTTP with TLS origination by the sidecar proxy. Using HTTP with TLS origination by the sidecar proxy is described in the [Control Egress Traffic]({{home}}/docs/tasks/traffic-management/egress.html) task. In that task, `https://google.com` was accessed by issuing HTTP requests to  `http://www.google.com:443`.
+
+The HTTPS traffic  originated by the application will be treated by Istio as _opaque_ TCP. To enable such traffic, we define a TCP egress rule on port 443.
```
ZackButcher: There are a few double spaces, `sed s/  / /g` please.

ZackButcher: Wow, Github renders `s/<space><space>/<space>/g` as `s/<space>/<space/g` :(

_docs/tasks/traffic-management/egress-tcp.md
```diff
@@ -0,0 +1,147 @@
...
+
+In TCP egress rules as opposed to HTTP-based egress rules, the destinations are specified by IPs or by blocks of IPs in [CIDR notation](https://tools.ietf.org/html/rfc2317).
```
ZackButcher: Do we support using SNI hosts for egressing TCP connections if TLS is enabled?

_docs/tasks/traffic-management/egress-tcp.md
```diff
@@ -0,0 +1,147 @@
...
+
+Let's assume for the sake of the example that we want to access `wikipedia.org` by the domain name. This means that we have to specify all the IPs of `wikipedia.org` in our TCP egress rules. Fortunately, the IPs of `wikipedia.org` are published [here]( https://www.mediawiki.org/wiki/Wikipedia_Zero/IP_Addresses). It is a list of IP blocks in [CIDR notation](https://tools.ietf.org/html/rfc2317): `91.198.174.192/27`, `103.102.166.224/27`, and more. We must define an egress rule per each IP block.
```
ZackButcher: Again, not totally related to the topic of this Task, but don't we support egressing with external DNS resolution?

_docs/tasks/traffic-management/egress-tcp.md
vadimeisenbergibm: Good point.  Actually, pods with sidecars cannot access URLs outside of the Kubernetes cluster. The question is if we should be so specific here.

_docs/tasks/traffic-management/egress-tcp.md
vadimeisenbergibm: No, it is not yet supported by envoy https://github.com/envoyproxy/envoy/issues/1843

_docs/tasks/traffic-management/egress-tcp.md
vadimeisenbergibm: We are waiting for this issue https://github.com/envoyproxy/envoy/issues/1843, we do not support SNI-based routing currently.

ZackButcher: Resolving DNS names outside of the mesh should be a different issue, though IIRC Envoy hasn't implemented it yet either (@rshriram you'd pointed me to an issue for this before, I think?).

vadimeisenbergibm: As far as I understand, resolving DNS names happens before the request arrives to envoy, inside the application container. The request arrives to envoy with IP and SNI. Currently SNI cannot be used for routing, only IP.

_docs/tasks/traffic-management/egress-tcp.md
vadimeisenbergibm: Replaced "outside the mesh" with "outside the cluster".

_docs/tasks/traffic-management/egress-tcp.md
vadimeisenbergibm: removed the whitespaces. `s/<space><space>/<space>/g` would remove yaml and markdown indentation, so I did not use it.

ZackButcher:
LGTM

PR #936, opened 2018-02-06
updated the egress rules reference with regard to TCP traffic

vadimeisenbergibm:
picking only the relevant parts from the the generated HTML
the TCP parts are required now, unrelated to other changes.


vadimeisenbergibm:
Made redundant by updating the reference by another PR.

PR #943, opened 2018-02-07
Fix figure references to use the correct approach for istio.io

geeknoid:


geeknoid:
Check out https://istio.io/about/contribute/writing-a-new-topic.html#adding-images for the info.

PR #945, opened 2018-02-07
added traffic mirroring task

christian-posta:


_docs/tasks/traffic-management/mirroring.md
```diff
@@ -0,0 +1,228 @@
...
+  destination:
+    name: httpbin
+  precedence: 11
+  route:
+  - labels:
+      version: v1
+    weight: 100
+  - labels: 
+      version: v2
+    weight: 0
```
rshriram: You can omit this piece - the labels: version:v2, weight 0.

_docs/tasks/traffic-management/mirroring.md
```diff
@@ -0,0 +1,228 @@
...
+  mirror:
+    name: httpbin
+    labels:
+      version: v2
+EOF
+```
+
+This route rule specifies we route 100% of the traffic to v1 and 0% to v2. At the moment, it's necesary to call out the v2 service explicitly because this is what creates the cluster definitions in the background. The last stanza specifies we want to mirror to the `httpbin v2` service.
+
```
rshriram: Wait, are you saying that without that dummy route (Weight 0), the envoy clusters are not being formed, even if I specify just `mirror:...` ? That seems more like a bug that we should fix.

Also, i would qualify cluster as "envoy cluster"

christian-posta: Yep, a bug then. https://github.com/istio/istio/issues/3270

Can clarify "envoy cluster"

thanks!

_docs/tasks/traffic-management/mirroring.md
```diff
@@ -0,0 +1,228 @@
...
+
+Now if we send in traffic:
+
+```bash
+kubectl exec -it $SLEEP_POD -c sleep -- sh -c 'curl  http://httpbin:8080/headers'
+```
+
+We should see access logging for both `v1` and `v2`. The access logs created in `v2` is the mirrored requests that are actually going to `v1`.
+
```
frankbu: Maybe it's obvious, but it might we worth mentioning that the response from the mirror is discarded.

_docs/tasks/traffic-management/mirroring.md
```diff
@@ -0,0 +1,228 @@
...
+      version: v2
+    weight: 0
+  mirror:
+    name: httpbin
+    labels:
+      version: v2
+EOF
+```
+
+This route rule specifies we route 100% of the traffic to v1 and 0% to v2. At the moment, it's necesary to call out the v2 service explicitly because this is what creates the envoy-cluster definitions in the background. The last stanza specifies we want to mirror to the `httpbin v2` service.
```
frankbu: "it's necesary to call out the v2 service explicitly"
Is this a "feature" or a bug?

_docs/tasks/traffic-management/mirroring.md
christian-posta: Good point. will add that.

_docs/tasks/traffic-management/mirroring.md
christian-posta: Well... I'd classify it as a bug for long term, but for 0.5.0, that's how to get it working.

https://github.com/istio/istio/issues/3270

rshriram:
Please remove the .DS_Store file from mac. :). Once done, its good to go.

PR #950, opened 2018-02-08
added .DS_Store file to .gitignore

vadimeisenbergibm:
https://en.wikipedia.org/wiki/.DS_Store


vadimeisenbergibm:
@ldemailly yes, tested it in a sub dir too.

PR #957, opened 2018-02-08
Synch examples with attribute vocabulary.

metacosm:


metacosm:
@geeknoid please check

PR #964, opened 2018-02-09
rfc: update automatic proxy injection clean-up steps

ayj:
@linsun - As discussed on slack, here are the additional steps to clean-up resources created during the task. They're optionally in the sense that keeping them around won't break the cluster.


linsun:
@ayj thanks you much!  Sorry I missed the review req earlier, LGTM 👍

PR #966, opened 2018-02-10
Add support for a dark color theme.

geeknoid:
- The version dropdown is now replaced with a gear dropdown. This dropdown lets you
pick the version of the site to look at, and lets you pick the light or dark color
themes for the site.

- Fixed a rendering glitch on the community page where a stray right and bottom border
was rendered.


ldemailly:
circle (`rake test`) isn't happy

ldemailly:
also, even though not exactly related, but can you make the archive links go to /docs/ instead of / by default (ideally it'd go to the exact same spot but given we move things around all the time, let's use /docs/ as it's more likely to be useful to someone switching version than the home page)

or maybe even: `archive.istio.io/{{version}}/docs/reference/` ? wdyt ?

ldemailly:
screen shot 2018-02-10 at 7 26 24 pm

lastly on desktop the checkbox seems a bit too bold/strong ^

geeknoid:
PTAL.

I fixed the rake tests and slightly improved the visuals of the checkmark. Unfortunately, I'm constrained by the font package we're using in terms of checkmark stroke width. So I made the checkmark smaller and improved the coloring, which I think helps.

As for the idea of having the version selector send you to somewhere specialized, I have been looking at doing something there. As you point out, we can switch page to page, since each site version is different. But we should be able to switch section-by-section at least. So if you're in the reference section, it'll throw you at the root of the reference section in the other version, same for about & blog. Let's tackle this in a separate PR.

PR #968, opened 2018-02-12
Add "smart" version switching, where we try to switch you to the same general area of the web site across  versions.

geeknoid:
Also, upgrade to FontAwesome 5.0

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -35,23 +35,62 @@
             <div class="dropdown">
                 <a href="" class="nav-link nav-item dropdown-toggle" data-toggle="dropdown" aria-haspopup="true"
                          aria-expanded="false">
-                    <i class='fa fa-lg fa-gear'></i>
+                    <i class='fa fa-lg fa-cog'></i>
                 </a>
 
                 <div class="dropdown-menu">
+                    {% assign target = "" %}
+                    {% if current[1] == 'about' %}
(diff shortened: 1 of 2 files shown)


_includes/header.html
```diff
@@ -35,23 +35,62 @@
             <div class="dropdown">
                 <a href="" class="nav-link nav-item dropdown-toggle" data-toggle="dropdown" aria-haspopup="true"
                          aria-expanded="false">
-                    <i class='fa fa-lg fa-gear'></i>
+                    <i class='fa fa-lg fa-cog'></i>
                 </a>
 
                 <div class="dropdown-menu">
+                    {% assign target = "" %}
```
ldemailly: let's make the default /docs/reference/ ?

geeknoid:
Luckily, the way this is done, we end up with links to archive.istio.io in the header of all documents. If one of the links doesn't work, circle ci complains. So we get reasonable automated tests of the current site. As we archive different releases, it's conceivable that the archived site's magic links will get stale.

We already have some dead links in the archived sites as it is. Maybe we should be running rake test on the archive and fix breaks when they emerge?

ldemailly:
I guess it depends how much traffic we will get. I think it’s ok archives have a few errors. chasing down changes to envoy docs for instance would be a lot to apply in all branches (and some of those changes only apply to new code/feature)

PR #976, opened 2018-02-12
Add note about using short names for istioctl get

jwendell:
This will be valid once https://github.com/istio/istio/pull/3394 gets merged.


ldemailly:
those are auto generated so I think will get lost unless the code also exports the short names somewhere else?

Issue #955, opened 2018-02-08
make all external links open in a new window/tab


ldemailly:
follow up on #951

geeknoid:
Done.

PR #980, opened 2018-02-13
Link improvements

### LINKED ISSUE #977: Need to refer to subsection in any of our doc easily as a url
i noticed on istio.io site, i can’t click on a subsection to point to a url, for example, installation steps in k8s quick start:  https://istio.io/docs/setup/kubernetes/quick-start.html

 i would like to click on “installation steps” section in this doc, and give people the link pointing to that section.

Per @geeknoid - this can only be done using the TOC on the right side but not from main text today.

geeknoid:
- When you hover over an HTML header, we now pop up a link icon holding a link to the
header. Useful to support deep linking to a header.

- When you click on a link that points outside of istio.io, the target is now
opened in a separate tab.

- Improved the coloring of links in the light theme so they're more visible. I was almost not able to
see the difference between normal text and links, now the color delta is greater. Additionally, when you hover on a link, we now draw an underline.

- Change the color of H1 headers to be black instead of being the same color
as links. They're not links, they shouldn't be that color.

Fixes #977 and #955


linsun:
Looks good!  I love the istio blue on the link :). thank you Martin so much!!

linsun:
merging, as color comment may be addressed later if Martin wants to. :)

geeknoid:
We can tweak the link color, but the previous color wasn't providing enough
contrast between the link color and normal text color. I often couldn't
tell what was a link and what wasn't without hovering the mouse over it. I
tried toning the color down a bit last night, but then it once again became
hard to tell the difference between links and normal text.

ldemailly:
having non link being black is a good idea. but having links be electric blue isn't: #985

ldemailly:
maybe you want to add a high contrast theme, but leave the default to the standard istio colors ?
(which shouldn't include any cyan or other too close to 100% in 1 component colors)

PR #982, opened 2018-02-13
0.5.1

ldemailly:


ldemailly:
there is more to be changed but let’s start with that

PR #983, opened 2018-02-13
update docs for the fact the webhook files are not included in 0.5.1

ayj:


ldemailly:
sad we missed this :( but thx for the update

PR #995, opened 2018-02-15
Add info on creating a site archive.

geeknoid:


README.md
```diff
@@ -48,3 +55,48 @@ bundle install
 bundle exec rake test
 bundle exec jekyll serve
 ```
+
+## Archiving the site
+
+We archive snapshots of the whole site for every release, which allows our customers to time travel and review
+older docs. This is very valuable if they're running older versions and need to look something up.
+Archives are maintained at https://archive.istio.io.
```
ldemailly: hyperlink?

README.md
```diff
@@ -48,3 +55,48 @@ bundle install
...
+
+To create an archive version, you need to follow these steps:
```
ldemailly: I'm wondering if this is really something consumable by the typical istio contributor ?
should it go to https://github.com/istio/admin-sites/tree/master/archive.istio.io instead ?

README.md
```diff
@@ -48,3 +55,48 @@ bundle install
...
+
+- Go to https://google.com/cse and create a new search engine for the archive site. After you've gone through the hoops necessary
+to get access to this site, you'll see there are many search engines already created. We create one search engine
+per site archive, which makes it possible for customers to search within this specific version of the site. You will need
+to create a new search engine. Just create a new engine and specify that you'd like to search the url https://archive.istio.io/vXX
```
ldemailly: I don't think we should make 1 engine per version, seems wasteful and unnecessary (not counting an extra step/barrier every month)

README.md
```diff
@@ -48,3 +55,48 @@ bundle install
...
+web site.
+
+- Once the search engine is created, hunt around in the UI to find the search engine ID and copy that.
+
+- Get the latest version of the site source code and create a new branch for the archive:
+
+   ```bash
+   git fetch upstream
+   git rebase upstream/master
+   git checkout -b release-XX
```
ldemailly:
```
git checkout master
git pull
git checkout -b release-XX
```

PR #999, opened 2018-02-15
Version bump.

geeknoid:
- Bump site version to 0.6pre and include archive of 0.5

- Refresh the reference docs.


ldemailly:
aren’t those changes in 0.5.1 or is this from new commits in the lqst few days?

PR #1001, opened 2018-02-15
Update key/cert path

wattli:


wattli:
Corresponding change is here: https://github.com/istio/istio/pull/3520

PR #1002, opened 2018-02-15
Add secure Ingress instructions.

myidpt:
Fixes https://github.com/istio/istio.github.io/issues/1000


myidpt:
@andraxylia @wattli PTAL

PR #1003, opened 2018-02-15
Fix telemetry docs for 0.5.1 Prometheus

salrashid123:
```0.5.1``` config for Prometheus is missing the namespace [#3139](https://github.com/istio/istio/issues/3149)

Added in instructions to modify ```prometheus.yaml``` file.

I'll revert this change once ```0.6.0``` is out


ldemailly:
so this should go in the release-0.5 branch then?
cc @geeknoid

salrashid123:
@geeknoid   Is this ok to merge in or if  ```0.6.0``` is coming out next week, i can abandon this PR. LMK

geeknoid:
I think might as well abandon at this point. 0.6 is actually done at this point, we're just waiting for the release notes to be finished before sending out the announcement.

salrashid123:
SGTM.  i'll review the GKE install istio docs next week and see if there is a way to provide a self-contained 'copy-paste' command set like [this](https://github.com/salrashid123/istio_helloworld#create-a-19-gke-cluster-and-bootstrap-istio)

PR #1007, opened 2018-02-16
A few fixes

geeknoid:
- Add support for showing the archive date in the footer.

- Fix invalid HTML in the header which improves the responsive layout. In particular, the down arrow
for the dropdown menu tends to stay on the same line as the cog more than before.

- Tone down the blue color of links so it's not so intense. There's still good contrast
between links and normal text.

- Remove broken og:image meta tag from each page. It used to point to a logo.png
file, which we deleted months ago. Turns out it's not intended for web site logos
anyway.

- Add og:url and og:site_name per best practices.

- Remove one more stray border on the community page. I thought I got rid of 'em last time,
but I missed one.

### DIFF
--- a/_sass/themes/_light.scss
+++ b/_sass/themes/_light.scss
@@ -9,7 +9,7 @@ $backgroundColor: $white;
 $textColor: #535f61;
 $textCodeColor: #d14;
 
-$linkColor: #469BFF;
+$linkColor: #469Bdd;
 $linkHoverColor: darken($linkColor, 20%);
 $linkDisabledColor: #CCCCCC;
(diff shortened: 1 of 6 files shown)


_sass/themes/_light.scss
```diff
@@ -9,7 +9,7 @@ $backgroundColor: $white;
 $textColor: #535f61;
 $textCodeColor: #d14;
 
-$linkColor: #469BFF;
+$linkColor: #469Bdd;
```
ldemailly: I'd go a bit darker/further to what it used to be but thanks for going in that direction already !

PR #1010, opened 2018-02-16
Change the footer to show the archive date.

geeknoid:


ldemailly:
did you get those dates from git log ?

_includes/footer.html
```diff
@@ -48,7 +48,7 @@
         <div class="row">
             <p class="description small text-center">
                 Istio Version {{site.data.istio.version}}, Copyright &copy; 2017 Istio Authors<br>
-                This site was built on {{site.time | date_to_string | split: ' ' | join: '-'}}
+                Archived on {{site.data.istio.archive_date}}
```
ldemailly: maybe this can be "Archive as of " vs "on" as it wasn't archived on / there are all kind of cosmetic changes on top / since ? or is that nit picking ?

the built date is still somewhat interesting but maybe that one shows at the toplevel of archive ?

geeknoid:
Yes, the dates represent the time at which the archive diverged from the mainline. I don't think cosmetic fixes and such are relevant to the consumer of the site, they care about what the site is documenting.

I'm not sure what the build is used for by a customer. Why do they care that the 0.1 archive site was rebuilt on feb-14? And everytime we rebuild archive.istio.io, the build date would change. Who cares?

PR #1027, opened 2018-02-24
Fix broken SVGs

dankohn:
Signed-off-by: Dan Kohn <dan@dankohn.com>

All of these SVGs were corrupted.

### DIFF
--- a/img/bulb.svg
+++ b/img/bulb.svg
@@ -1,6 +1,3 @@
----
-layout: compress
----
 <?xml version="1.0" encoding="iso-8859-1"?>
 <svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
 	 viewBox="0 0 512 512" xml:space="preserve">
--- a/img/exclamation-mark.svg
+++ b/img/exclamation-mark.svg
@@ -1,6 +1,3 @@
----
-layout: compress
----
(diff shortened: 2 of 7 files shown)


ldemailly:
cc @geeknoid

geeknoid:
In what way were the SVG files corrupted?

The front-matter was intentionally added so that the SVG gets compressed before being emitted in the main site. In usage, as far as I can tell, these SVGs were fine and being loaded by browsers.

Did the compression mess something up that browsers just overlook?

ldemailly:
ah glad it didn't break on the site, but that header makes them uneditable 
also I don't think compression for a 3k long cached file is really that important ? it should be done by the webserver config if we needed

geeknoid:
It's good hygiene to statically compressed everything. I agree that it
doesn't make much of a difference for those small files, but I rather just
do it systematically for all our site content. All .html files in the site
are also 'broken' in that respect since they're not valid HTML file until
the site is published.

ldemailly:
most (all?) the user editable content is in MD not html

but more relevant: you can't exactly draw in text mode when opening the files; if one needed to edit the icons (yes I have a couple hand coded svg too but that's not quite the common case)

dankohn:
It breaks them in the SVG viewer on GitHub. 
--
Dan Kohn <dan@linuxfoundation.org>
Executive Director, Cloud Native Computing Foundation  https://www.cncf.io
+1-415-233-1000  https://www.dankohn.com

PR #1038, opened 2018-02-28
Add task for Istio CA health check.

myidpt:


myidpt:
@wattli @ldemailly PTAL. Thanks!

_docs/tasks/security/health-check.md
```diff
@@ -0,0 +1,143 @@
...
+order: 70
+
+layout: docs
+type: markdown
+---
+{% include home.html %}
+
+This task shows how to enable Istio CA health check. Note this is an alpha feature after Istio V0.6.
+
+After Istio V0.6, Istio CA has a health check feature that can be optionally enabled.
```
wattli: So this feature is disabled by default? If so, please mentioned it here.

ldemailly: ^After^Since

_docs/tasks/security/health-check.md
```diff
@@ -0,0 +1,143 @@
...
+Currently, the health check feature is able to detect the failures of the CA CSR signing service,
+by periodically sending CSRs to the service. More health check features are coming shortly.
+
+The Istio CA contains a _prober client_ module that periodically checks the CA's status (currently only the health
+status of the gRPC server).
+If the Istio CA is healthy, the _prober client_ updates the _modificate time_ of the health status file
```
ldemailly: sort of out of scope for this doc PR but why aren't we using grpc health check instead of this complication/file based approach ?

_docs/tasks/security/health-check.md
myidpt: Sure.

_docs/tasks/security/health-check.md
myidpt: This is mostly because the command line offers better security. K8s probe does not support TLS so we have to open plain text port, which makes the health information public, and makes the CA vulnerable to DoS attacks. I think when K8s probe supports TLS, we can use gRPC health check.

myidpt:
@ldemailly I think this PR is ready to be merged :)

_docs/tasks/security/health-check.md
ldemailly: afaik we are allowing grpc calls from the mesh and vms, without mtls so not sure I get your point with the benefits

ldemailly:
feel free to merge it and we can continue the discussion on why not use the open port for health check elsewhere

my last comment:

> afaik we are allowing grpc calls from the mesh and vms, without mtls so not sure I get your point with the benefits/requirements of not use the open port for health checks

myidpt:
Sure. Could you merge it for me? I don't have write access to this repo.
The design doc is here:
https://docs.google.com/document/d/1ig1ieuaPYr-EB7gByt4BgCQAXudnmZrqXV294K_oU1M/edit#
Please feel free to comment.

PR #1044, opened 2018-02-28
Sidecar proxy help

john-a-joyce:
Resolves https://github.com/istio/istio.github.io/issues/1043


_help/troubleshooting.md
```diff
@@ -384,3 +384,17 @@ of HTTP 404s (upto 2x the refresh interval) until the Envoy sidecars get all rel
 
 NOTE: The 0.5.0 and 0.5.1 releases are missing scripts to provision webhook certificates. Download the missing files from [here](https://raw.githubusercontent.com/istio/istio/master/install/kubernetes/webhook-create-signed-cert.sh) and [here](https://raw.githubusercontent.com/istio/istio/master/install/kubernetes/webhook-patch-ca-bundle.sh). Subsqeuent releases (> 0.5.1) should include these missing files.
 
+## Automatic side car injection will fail if the kube-apiserver has proxy settings
```
geeknoid: side car -> sidecar

_help/troubleshooting.md
```diff
@@ -384,3 +384,17 @@ of HTTP 404s (upto 2x the refresh interval) until the Envoy sidecars get all rel
...
+
+This was tested on 0.5.0 with the additional files required as referenced in the above issue.   When the Kube-apiserver included 
+proxy settings such as:
+      env:
```
geeknoid: Please enclose YAML content in
<pre>
```yaml
... 
```
</pre>

_help/troubleshooting.md
```diff
@@ -384,3 +384,17 @@ of HTTP 404s (upto 2x the refresh interval) until the Envoy sidecars get all rel
...
+    - name: http_proxy
+      value: http://proxy-wsa.esl.foo.com:80
+    - name: https_proxy
+      value: http://proxy-wsa.esl.foo.com:80
+    - name: no_proxy
+      value: 127.0.0.1,localhost,dockerhub.cisco.com,devhub-docker.foo.com,10.84.100.125,10.84.100.126,10.84.100.127
+
+The sidecar injection would fail.   The only related failure logs was in the kube-apiserver log:
+W0227 21:51:03.156818       1 admission.go:257] Failed calling webhook, failing open sidecar-injector.istio.io: failed calling admission webhook "sidecar-injector.istio.io": Post https://istio-sidecar-injector.istio-system.svc:443/inject: Service Unavailable
```
geeknoid: Please enclose this in

<pre>
```bash
...
```
</pre>

rcurran1: Consider adding a sentence on how to fix the issue, ex. "To fix this issue, update the proxy settings to match the deployment server or, if not needed, remove the proxy settings from the kube-apiserver.yaml manifest."

ayj: Reference to a similar problem: https://github.com/kubernetes/kubeadm/issues/666
Possible solution: https://github.com/kubernetes/kubernetes/pull/58698#discussion_r163879443

john-a-joyce:
Resolved all the comments I believe.

geeknoid:
Jason, are you cool to check this in?

_help/troubleshooting.md
```diff
@@ -416,3 +416,24 @@ of HTTP 404s (upto 2x the refresh interval) until the Envoy sidecars get all rel
...
+  - name: no_proxy
+  value: 127.0.0.1,localhost,dockerhub.foo.com,devhub-docker.foo.com,10.84.100.125,10.84.100.126,10.84.100.127
+```
+The sidecar injection would fail.   The only related failure logs was in the kube-apiserver log:
+```bash
+W0227 21:51:03.156818       1 admission.go:257] Failed calling webhook, failing open sidecar-injector.istio.io: failed calling admission webhook "sidecar-injector.istio.io": Post https://istio-sidecar-injector.istio-system.svc:443/inject: Service Unavailable
+```
+A workaround is to remove the proxy settings from the kube-apiserver manifest and restart the server.   
+
+An issue has been filed in kubernetes related to this.   [https://github.com/kubernetes/kubeadm/issues/666](https://github.com/kubernetes/kubeadm/issues/666)
```
ayj: (copying my comment from https://github.com/istio/istio.github.io/issues/1043)

> Possibly related to kubernetes/kubeadm#666. According to kubernetes/kubernetes#58698 (comment),
>
> kubeadm gives warnings during init stage if access to any IP address in pod or service CIDRs will be going via proxies, according to values of *_proxy variables.
>
> Do you see anything in the kubeadm logs for the when this problem occurs? You may need to add pod-cidr / service-cidr ranges to your no_proxy settings.

https://github.com/kubernetes/kubernetes/pull/58698 isn't in progress anymore (it's closed). Assuming a functional kubeadm, you should see a warning/error about this sort of issue in the install logs. If so, we should highlight that here. If not, we should file a follow-up bug with kubeadm.

I think the text here should read something along the lines of "make sure pod or service CIDR are not proxied according to *_proxy variables. Check such-and-such files and logs to verify".

cc @sdake @john-a-joyce

john-a-joyce: I can update the text as you suggested "make sure pod or service CIDR are not proxied according to *_proxy variables'  but regarding the first point about the kubernetes issue being fixed.  This is a troubleshooting guide for Istio.  IMO it should avoid as much as possible getting into details on the Kubernetes issues and especially the timing of the fixes.   Even the scant info I put in got me in trouble with the state of the defect.  As a case in point we made sure we used the minimal Kubernetes version Istio required when we saw this.   We can certainly try to retest this at some point,  but I don't see the benefit of waiting to add this to the guide until that is done.

ayj: It looks like kubeadm already prints a warning about this issue (see [here](https://github.com/kubernetes/kubernetes/pull/35972/files#diff-a466e16a4b5d497feb0dc3539119ed25R211) and error message snippet below).

```golang
return []error{fmt.Errorf("Connection to %q uses proxy %q. If that is not intended, adjust your proxy settings", url, proxy)}, nil
```

As far as I can tell, based on the linked PRs, this is working as intended and there is nothing to fix in k8s, other than perhaps elevate the warning to an error (see https://github.com/kubernetes/kubernetes/pull/58698#issuecomment-365335555). The current wording with use of "workaround" and "possible fix" suggests otherwise.

john-a-joyce: I am not sure why what kubernetes is doing correctly or incorrectly matters.  This is for the Istio troubleshooting guide.  This is to inform users about a possible mis-configuration.  A group we work with spent well over a day trying to figure out what was wrong and I thought this would help others more quickly identify what might be wrong in their system.   I am not sure why you seem resistant to adding it.  If I mis-understood what type of content was supposed to go into the guide I apologize.  We can just reject the PR and close the issue.  The system this was occurring in was never under my direct control and I have no ability to check on prior state.   Spending any more time on this at this point is not worth the effort IMO.  Please advise on how you want to proceed on the PR.

ayj: The latest version is better - thanks.

PR #1050, opened 2018-03-02
Fix misnamed link

embs:
Link is named "below" but adapters section is above it.


_docs/concepts/policy-and-control/mixer.md
```diff
@@ -88,7 +88,7 @@ operator is responsible for:
 
 - Configuring a set of *handlers* for Mixer-generated data. Handlers are
   configured adapters (adapters being binary plugins as described
-  [below](#adapters)). Providing a `statsd` adapter with the IP address for a
+  [here](#adapters)). Providing a `statsd` adapter with the IP address for a
```
embs: 1. Used "here" instead of "above" to avoid future new occurrences of similar semantics trickies
2. Tests for this would be an overkill, right?

PR #1051, opened 2018-03-02
Use same instance name in Mixer config example

embs:


_docs/concepts/policy-and-control/mixer.md
```diff
@@ -102,7 +102,7 @@ operator is responsible for:
   when Mixer will execute the specified actions. Actions specify the set of
   instances to generate and the handlers that should process the generated
   instances. For example, a rule might tell Mixer to send generated
-  `requestcount` instances to a `statsd` handler for all `Report` calls.
+  `request_count` instances to a `statsd` handler for all `Report` calls.
```
embs: Exact match with https://github.com/istio/istio.github.io/blob/db6b6d993b44bb104454dc60dd01108928edd581/_docs/concepts/policy-and-control/mixer.md#L96

douglas-reid: suggestion: let's change the previous mention to be `requestcount`, so that it matches the instance names used in actual config.

See: https://github.com/istio/istio/blob/master/install/kubernetes/templates/istio-mixer.yaml.tmpl#L785

embs: Thanks, @douglas-reid. It's done.

cc @geeknoid

PR #1053, opened 2018-03-04
Make the rake test ensure that we use {{home}} consistently.

geeknoid:
We now generate the test site into a subdirectory such that we can ensure all
links are correctly using {{home}}, which makes the site work correctly once
archived.

Fixed a bunch of broken cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 _site
+_rakesite
 _static_site
 .bundle
 config_override.yml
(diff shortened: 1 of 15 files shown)


ldemailly:
does it fail circle if you remove one of the {{home}} and put / or https://istio.io/ instead ?
we may need a legal variable for {{istio_io_root}} when one wants to link there ? (like in theory all blog links should go there directly)

geeknoid:
It won't fail if a link is using https://istio.io explicitly, so long as the link target already exists.

PR #1054, opened 2018-03-04
Reduce the visual weight of code blocks so they don't break up the page so much

geeknoid:


ldemailly:
before/after screenshot ?

geeknoid:
Before:

screen shot 2018-03-05 at 06 22 30

After, showing the less stark and visual obtrusive code blocks.

screen shot 2018-03-05 at 06 23 15

PR #1057, opened 2018-03-06
Update document to use --debug option when deploy BookInfo app

incfly:
This is required for mutual TLS verification, where we asks user to login into envoy container and `curl`. program curl is only installed in docker.io/istio/proxy_debug:.*.

See https://github.com/istio/istio/issues/3399 for background.

### DIFF
--- a/_docs/guides/bookinfo.md
+++ b/_docs/guides/bookinfo.md
@@ -85,7 +85,7 @@ To start the application, follow the instructions below corresponding to your Is
    use the following command instead:
 
    ```bash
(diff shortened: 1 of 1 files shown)


incfly:
I verified in the lastest istio 0.6, this change makes mTLS tutorial works again.

wattli:
@ldemailly , should we use debug version in bookinfo demo? Some of the demo requires libs that only exist in debug version, like 'curl'

wattli:
https://github.com/istio/istio/issues/3399 is the issue

ldemailly:
yes sounds good; for bookinfo performance should not matter

PR #1063, opened 2018-03-08
Small variable name fix

RcrsvSquid:
Edited the variable name to match the previous example in order to avoid confusion


geeknoid:
#1051 went in the opposite direction earlier this week, so I don't think we want this.

Thanks.

PR #1065, opened 2018-03-09
Update solarwinds.html

leecalcote:
spelling correction - "backendsu" changed to "backends"


geeknoid:
This is auto-generated from the source material in istio/istio/mixer/adapter/... So please fix it there, and the fix will be picked up when we regenerate the reference docs.

Thanks.

leecalcote:
Thanks @geeknoid. Will do.

PR #1067, opened 2018-03-09
[WIP] Update traffic routing tasks to use v1alpha3 config

frankbu:
This is a WIP PR to convert all the traffic routing tasks to use the new routing model.
The task converted so far (mostly) work, but there are some bugs.

Rather than replacing the `v1alpha1` tasks, these new versions are added as an alternative, i.e., keeping both the old and new versions around for a now, since `v1alpha3` is not yet a fully functional replacement. They appear in a new Tasks subfolder with title: `Traffic Management (v1alpha3)`.


_docs/tasks/traffic-management-v1alpha3/fault-injection.md
```diff
@@ -0,0 +1,172 @@
...
+     http:
+     - fault:
+         delay:
+           fixedDelay: 7s
+           percent: 100
+       match:
+       - headers:
+           cookie:
+             regex: ^(.*?;)?(user=jason)(;.*)?$
+     - route:
```
rshriram: this is incorrect. route and fault are in the same object

_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -0,0 +1,621 @@
...
+
+   ```bash
+   kubectl delete -f samples/httpbin/httpbin.yaml
+   ```
+
+
+## What's next
+
+* Learn more about [Ingress Resources](https://kubernetes.io/docs/concepts/services-networking/ingress/).
+
```
rshriram: we can skip this doc and wait for gateway to land

kyessenov: I think it's OK to keep the part of ingress that does not require combining route rules. We _do_ want basic ingress to continue to work, since it's crucial part of UX for the majority of k8s customers.

kyessenov:
Do not remove ingress. Most of our customers use ingress, and do not always
require combining routing rules with the ingress.

rshriram:
Ingress docs already exist in v1alpha1. This is just to get v1alpha3 docs a place, so that we can clean it up. It will most probably not be visible for a month.

_docs/tasks/traffic-management-v1alpha3/fault-injection.md
frankbu: good catch. I didn't notice since it works anyway, because there's only one version of ratings running.

That said, I guess the route actually needs to be in both rules, in the first rule and also in the non-matched lower priority rule.

_docs/tasks/traffic-management-v1alpha3/ingress.md
frankbu: I was thinking that we could add another section in this task to talk about gateway (when it's working), so the "Ingress Task" will show how to do ingress both ways. For now, I guess it's best to just leave it as is, and enhance it in a later PR.

frankbu: I'm going to remove the ingress task from this PR for now, and then add a new combined ingress/gateway task in a followup PR.

PR #1068, opened 2018-03-11
Eliminate flickering on page load.

geeknoid:
- Fix another issue with my arch-nemesis, the Copy button. My last fix for Copy button issues
resulted in screen flickering upon page loading. This is now fixed.

- Pin the size of the gear and magnifying glass icons in the header to avoid flicker as the
fonts for those renders a few ms too late and lead to flickering on page load.

- Cleaned up the site's JavaScript for clarity, and include minimized versions in the
site for improved perf.


geeknoid:
Nope, sadly. The whole DOM patching stuff should be done statically
offline, but alas the tool chain currently has limits. I'm hoping to
eventually change how the web site is produced in order to give us more
control. Basically, if we run Jekyll locally and use GitHub Pages strictly
as a web server, we could do more statically.​

Issue #887, opened 2018-01-26
Update automatic sidecar injection prerequisites for k8s 1.9 (IBM CCS) 


ayj:
https://istio.io/docs/setup/kubernetes/sidecar-injection.html#prerequisites for IBM CCS need to be updated for k8s 1.9 to enable support for mutating webhooks.

```bash
$ kubectl api-versions | grep admissionregistration.k8s.io/v1beta1
admissionregistration.k8s.io/v1beta1
```

See https://archive.istio.io/v0.4/docs/setup/kubernetes/sidecar-injection#prerequisites for previous 0.4.0 doc version.

linsun:
Think this can be closed via #1072

PR #1072, opened 2018-03-12
Update sidecar-injection.md to add IBM Cloud Container service

linsun:
#887 cc @ayj

### DIFF
--- a/_docs/setup/kubernetes/sidecar-injection.md
+++ b/_docs/setup/kubernetes/sidecar-injection.md
@@ -155,6 +155,20 @@ kubectl create clusterrolebinding cluster-admin-binding \
     --user=$(gcloud config get-value core/account)
 ```
 
+#### IBM Cloud Container Service (IKS)
+
+Kubernetes 1.9 is generallly available on IBM Cloud Container Service (IKS).  At the time of writing it is not the default version, so to create a new lite cluster:
+
+```bash
+bx cs cluster-create --name my_cluster --kube-version 1.9.3
+``` 
+
+or a new paid cluster:
+
+```bash
+bx cs cluster-create --location location --machine-type u2c.2x4 --name my_cluster --kube-version 1.9.3
+```
+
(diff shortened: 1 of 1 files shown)


_docs/setup/kubernetes/sidecar-injection.md
```diff
@@ -155,6 +155,20 @@ kubectl create clusterrolebinding cluster-admin-binding \
     --user=$(gcloud config get-value core/account)
 ```
 
+#### IBM Cloud Container Service (IKS)
+
+Kubernetes 1.9 is generallly available on IBM Cloud Container Service (IKS).  At the time of writing it is not the default version, so to create a new lite cluster:
+
+```bash
+bx cs cluster-create --name my_cluster --kube-version 1.9.3
```
gyliu513: @linsun just double confirm, so here the command will create a cluster with admission controller plugin as `MutatingAdmissionWebhook,ValidatingAdmissionWebhook`?

linsun: @gyliu513 yes, they are turned on by default in our 1.9.x clusters in IKS.

ayj: Is the certificate manager controller also enabled by default? See the minikube setup steps for specific flags. This will be needed if/when we use k8s CA for cert provisioning.

geeknoid:
Should we go ahead and merge this?

ayj:
> Should we go ahead and merge this?

The k8s setup instructions have been consolidated under `_docs/setup/kubernetes/quick-start.md`. I ported this PR's changes to that location in https://github.com/istio/istio.github.io/pull/1136. Closing this PR in favor of that one.

_docs/setup/kubernetes/sidecar-injection.md
linsun: yes - i think so as I didn't have to enable it.

PR #1078, opened 2018-03-17
Fix broken links due to missing .html in URLs.

myidpt:
Links in https://istio.io/help/faq/ page are broken due to missing ".html" in the URL.
Also 2 links to page community.html are missing ".html".


geeknoid:
Addressed by #1080

PR #1082, opened 2018-03-21
Fix some newly broken links.

geeknoid:
We deleted some files in istio/api which broke these links. These broken links didn't show up in local "rake test" runs due to the 2 days cache setting we have in Rakefile. Once I set the cache duration to 0 days, the broken links emerged.


ldemailly:
one more ?

  *  External link https://github.com/istio/api/blob/master/mixer/v1/config/descriptor/value_type.proto failed: 404 No error

PR #1083, opened 2018-03-22
fix service account names in the instructions for OpenShift

tvieira:
This commit replaces the service account names for grafana and
prometheus in the instructions to set the security context
constraints for OpenShift.

This commit also eliminates extra spaces from few other lines.


geeknoid:
Would you please update your PR to resolve conflicts? Then we can merge your changes.

Thanks.

tvieira:
@geeknoid Thanks. Just pushed a re-based PR. Previous one also had some Markdown fixes because of my editor, and I eliminated that to reflect only on the service names.

PR #1091, opened 2018-03-23
Update circuit break use existing file.

gyliu513:


frankbu:
There is a new version of this task: https://github.com/istio/istio.github.io/blob/master/_docs/tasks/traffic-management-v1alpha3/circuit-breaking.md

geeknoid:
Should we go ahead and merge this?

frankbu:
@geeknoid, I'm not sure it makes sense to make this change in the soon to be deprecated version of the task. The corresponding change to the v1apha3 version of the task, however, prereqs creating the yaml file in istio/istio. Personally, I would just leave it as is for now.

geeknoid:
@gyliu513 should we close this per Frank's comments?

gyliu513:
As we need to maintain different istio versions, so I think that we may still need this to be merged. @frankbu @geeknoid

geeknoid:
@frankbu Frank, what do you think?

frankbu:
I don't have strong opinion, but I wouldn't do it, since this is going to get tossed soon anyway. Starting with 0.8, we are going to point people at v1alpha3.

geeknoid:
All right, let's just put it in. It doesn't hurt.

PR #1092, opened 2018-03-23
Update Readme

jsenon:
Change readme for better understanding of zipkin and jaeger deployment yaml


jsenon:
I received error with https://docs.ansible.com/ansible/latest/intro_installation.html, could you have a look?

PR #1094, opened 2018-03-23
WIP - Combined ingress/gateway task for v1alpha3

frankbu:
This doesn't work yet, but I'm trying to put together something that describes how it "will work"
once the Gateway implementation is ready.


_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -0,0 +1,579 @@
...
+An Istio Ingress specification is based on the standard [Kubernetes Ingress Resource](https://kubernetes.io/docs/concepts/services-networking/ingress/)
+specification, with the following differences:
+
+1. Istio Ingress specification contains a `kubernetes.io/ingress.class: istio` annotation.
+
+2. All other annotations are ignored.
+
+The following are known limitations of Istio Ingress:
+
+1. Regular expressions in paths are not supported.
```
rshriram: We can tackle this in a better way by setting all path matches to be regex matches.. So, instead, we should tell people that the Istio regex format is c++11 regex format (http://en.cppreference.com/w/cpp/regex/ecmascript).

The main limitation of ingress is that virtual services cannot be combined with Ingress specs. Essentially, whatever is in ingress will just work as any normal ingress (path/prefix match to one backend service). DestinationRules will work as they apply to the service. However, virtual services (Traffic splitting, mirroring, header match, etc.) will not work, nor will fault injection. 

In other words, the ingress user will get the same feature that the ingress spec supports. And nothing more than that.

_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -0,0 +1,579 @@
...
+             serviceName: httpbin
+             servicePort: 8000
+         - path: /delay/.*
+           backend:
+             serviceName: httpbin
+             servicePort: 8000
+   EOF
+   ```
+ 
+   `/.*` is a special Istio notation that is used to indicate a prefix
```
rshriram: We don't need any of this. Lets just say that the path must follow the ecmascript format.

_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -0,0 +1,579 @@
...
+
+   ```
+   HTTP/1.1 404 Not Found
+   date: Mon, 29 Jan 2018 04:45:49 GMT
+   server: envoy
+   content-length: 0
+   ```
+
+## Using Istio routing rules to control ingress traffic
+
```
rshriram: Lets remove all this. No route rule business.

PR #1098, opened 2018-03-26
Revise note to install curl

wattli:


_docs/tasks/security/mutual-tls.md
```diff
@@ -82,21 +75,39 @@ There are several steps:
...
    ```
    ```bash
    cert-chain.pem   key.pem   root-cert.pem
-   ``` 
-   
-   Note that cert-chain.pem is envoy's cert that needs to present to the other side. key.pem is envoy's private key paired with cert-chain.pem. root-cert.pem is the root cert to verify the other side's cert. Currently we only have one CA, so all envoys have the same root-cert.pem.  
-   
+   ```
+
+   Note that cert-chain.pem is envoy's cert that needs to present to the other side. key.pem is envoy's private key paired with cert-chain.pem. root-cert.pem is the root cert to verify the other side's cert. Currently we only have one CA, so all envoys have the same root-cert.pem.
```
geeknoid: Here and elsewhere, please capitalize Envoy.

_docs/tasks/security/mutual-tls.md
```diff
@@ -82,21 +75,39 @@ There are several steps:
...
+
+1. make sure 'curl' is installed by
+   ```bash
+   curl
```
ldemailly: I guess you could put
```
curl localhost:15000
```
or some such (the admin port for the proxy)

wattli: I am afraid that would be misleading, people may wonder what is this 15000? And sometimes if 15000 is down for some reason, it will mislead people.

_docs/tasks/security/mutual-tls.md
wattli: done.

PR #1115, opened 2018-03-28
0.2.12 typo fix + doc link should be to docs/ directly + ...

ldemailly:
0.2.12 doc link should be to docs/ directly + note about shell security + typo fix (for for) + seperate the monthly from the stable

### DIFF
--- a/_about/notes/index.md
+++ b/_about/notes/index.md
@@ -16,20 +16,20 @@ toc: false
 
 {% include section-index.html docs=site.about %}
 
-The [latest Istio monthly release](https://github.com/istio/istio/releases) is {{site.data.istio.version}} which
-you can download with(*):
+- The [latest Istio monthly release](https://github.com/istio/istio/releases) is {{site.data.istio.version}} which
+  you can download with(*):
 
-```
-curl -L https://git.io/getLatestIstio | sh -
-```
+  ```
+  curl -L https://git.io/getLatestIstio | sh -
+  ```
 
-The most recent stable release is [0.2.12](https://github.com/istio/istio/releases/tag/0.2.12) which you
-can download with:
+- The most recent stable release is [0.2.12](https://github.com/istio/istio/releases/tag/0.2.12) which you
+  can download with:
 
-```
-curl -L https://git.io/getIstio | sh -
-```
+  ```
+  curl -L https://git.io/getIstio | sh -
+  ```
 
-[Archived documentation for for 0.2.12 release](https://archive.istio.io/v0.2/docs/).
+  [Archived documentation for the 0.2.12 release](https://archive.istio.io/v0.2/docs/).
 
 > (*) Note: As we don't control the git.io domain, please examine the output of the curl command before piping it to a shell if running in any sensitive or non sandboxed environment.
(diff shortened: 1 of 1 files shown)


_about/notes/index.md
```diff
@@ -16,18 +16,20 @@ toc: false
 
 {% include section-index.html docs=site.about %}
 
-The [latest Istio monthly release](https://github.com/istio/istio/releases) is {{site.data.istio.version}} which
-you can download with:
+- The [latest Istio monthly release](https://github.com/istio/istio/releases) is {{site.data.istio.version}} which
```
geeknoid: There shouldn't be a - here and in the next item, it's not necessary and doesn't mesh well with the bulleted list of releases that are displayed right above.

_about/notes/index.md
```diff
@@ -16,18 +16,20 @@ toc: false
...
 
-```
-curl -L https://git.io/getIstio | sh -
-```
+  ```
+  curl -L https://git.io/getIstio | sh -
+  ```
 
-[Archived documentation for for 0.2.12 release](https://archive.istio.io/v0.2/).
+  [Archived documentation for the 0.2.12 release](https://archive.istio.io/v0.2/docs/).
```
geeknoid: Why is /docs needed?

ldemailly: because who cares about the front page when looking for docs for a given release ?
it saves (at least) one unnecessary click for our users

_about/notes/index.md
ldemailly: k I didn't notice the inlined TOC there, let me update locally and fix it up
(the issue is the doc link and the curl examples look detached from their respective release)

ldemailly: @geeknoid why do we need to list the releases twice - it's already on the left bar ?

also the period at the end of "Istio releases information." looks odd, would be a ":" instead for that toc to work

how about:
screen shot 2018-03-28 at 3 40 09 pm

ldemailly:
Preview:
screen shot 2018-03-28 at 3 40 09 pm

geeknoid:
This is the common pattern we follow on all navigation pages. You can fully
navigate the site without ever using the left-hand sidebar, which is
essential for a good experience on mobile, where the left sidebar is hidden
and inconvenient to use.

_about/notes/index.md
```diff
@@ -14,22 +14,19 @@ redirect_from:
 toc: false  
 ---
 
-{% include section-index.html docs=site.about %}
```
geeknoid: This needs to be present like in all other index pages in the site, so as to enable good mobile navigation. Please just let that there and remove the - and indent in the subsequent sections.

ldemailly:
you can't actually reach the release notes through navigation:

the about link goes to https://istio.io/about/intro.html
which if you don't have the nav bar is a dead end

should that be fixed (in a separate PR)

ldemailly:
same for blog

ps: I do agree the about should go to an actual page/content (blog is more arguable) so that's not what should be fixed, I think mobile can pop out the left bar when needed

ldemailly:
lastly https://istio.io/about/contribute/
doesn't have a double TOC either

ldemailly:
Merging this for now as it's incrementally better before the 0.7 branch gets cut. we can further refine navigation later

geeknoid:
Man, you are sure stubborn.

The contribute link does have a TOC in it, I fixed that a little while ago and its on preliminary.istio.io. And the design intent is that navigation is possible without the left side nav. Even though I mention this, you proceed with checking this in even though it is not a necessary change in any way and is just your whim, ignoring requests and bypassing the approval process.

PR #1121, opened 2018-03-29
Add Istio mTLS support for https service demo

wattli:


_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+
+layout: docs
+type: markdown
+---
+{% include home.html %}
+
+This task shows how Istio Mutual TLS works with https services. It includes: 1)
+Deploy a https service without Istio sidecar; 2) Deploy a https service with
+Istio with mTLS disabled; 3) Deploy a https service with mTLS enabled. For each
+deployment, connect to this service and verify it works.
```
ldemailly: this is a bit confusing to me

for 3) don't you need to use tcp proxying ? and for 2) it also means tcp proxying and metrics

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+
+order: 80
+
+layout: docs
+type: markdown
+---
+{% include home.html %}
+
+This task shows how Istio Mutual TLS works with https services. It includes: 1)
+Deploy a https service without Istio sidecar; 2) Deploy a https service with
```
myidpt: a https -> an https

_docs/tasks/security/https-overlay.md
wattli: Yes, it's true and explained below.

_docs/tasks/security/https-overlay.md
wattli: done

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+
+```bash
+$ kubectl create configmap nginxconfigmap --from-file=samples/https/default.conf
+configmap "nginxconfigmap" created
+```
+
+Then deploy the service:
+
+```bash
+$ kubectl apply -f samples/https/nginx-app.yaml
```
diemtvu: same here.

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+```bash
+$ kubectl apply -f <(bin/istioctl kube-inject --debug -f samples/sleep/sleep.yaml)
+```
+
+Get the pods
+
+```bash
+$ kubectl get pod
+NAME                              READY     STATUS    RESTARTS   AGE
+my-nginx-jwwck                    2/2       Running   0          1h
```
diemtvu: Can you change the deployment yaml to have the pod name and svc name similar. Just to be consistent with other demo.

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+sleep-847544bbfc-d27jg            2/2       Running   0          18h
+```
+
+Ssh into the istio-proxy container of sleep pod.
+```bash
+$ kubectl exec -it sleep-847544bbfc-d27jg -c istio-proxy /bin/bash
```
diemtvu: To be fancy, this one line command would work as is

kubectl exec $(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name}) -c istio-proxy -- curl https://nginxsvc -k

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+Then deploy the service:
+
+```bash
+$ kubectl apply -f samples/https/nginx-app.yaml
+...
+service "nginxsvc" created
+replicationcontroller "my-nginx" created
+```
+
+Then, create another pod to call this service.
```
diemtvu: May be move this part to "Before you begin", or (new) "common setup" section (which include both this and certificate generation/installation), as it's shared across all demos below.

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+Make sure the pod is up and running
+
+```bash
+$ kubectl get pod
+NAME                              READY     STATUS    RESTARTS   AGE
+my-nginx-6svcc                    2/2       Running   0          1h
+sleep-847544bbfc-d27jg            2/2       Running   0          18h
+```
+
+And run
```
diemtvu: Similar, change to single command to make it easy to follow.

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+...
+<h1>Welcome to nginx!</h1>
+...
+```
+
+Note: this example is borrowed from [kubernetes examples](https://github.com/kubernetes/examples/blob/master/staging/https-nginx/README.md).
+
+### Create an https service with Istio sidecar with mTLS enabled
+
+You need to redeploy the Istio control plane with mTLS enabled.
```
diemtvu: Or you can also refer to per-service enable/disable mTLS option :). I'll add using authn policy for that soon, and it would be easier.

diemtvu: Alternatively, you can assume/state the instruction start from scratch (istio is not install, apps are not install), and give a hint what to do if they want to switch. It would be easier for user to jump to this demo right away, instead of following the previous ones.

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+type: markdown
+---
+{% include home.html %}
+
+This task shows how Istio Mutual TLS works with https services. It includes: 1)
+Deploy an https service without Istio sidecar; 2) Deploy an https service with
+Istio with mTLS disabled; 3) Deploy an https service with mTLS enabled. For each
+deployment, connect to this service and verify it works.
+
+When Istio sidecar is deployed with an https service, the proxy automatically downgrades
```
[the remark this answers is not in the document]
wattli: Yes, this is confirmed with Laurent/Costin

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+### Create an https service without Istio sidecar
+
+This subsection creates a nginx-based https service.
+
+First we need to create a configmap:
+
+```bash
+$ kubectl create configmap nginxconfigmap --from-file=samples/https/default.conf
```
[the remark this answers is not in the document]
wattli: The file will be checked in later. Currently istio.io website if for release-0.6 branch, and this PR (commit to master) will not be shown in current istio.io

_docs/tasks/security/https-overlay.md
wattli: Same as above, i will have a separate PR to add the files. And this PR won't be in istio.io for a while

_docs/tasks/security/https-overlay.md
wattli: Not really, in step 2 and 3, we need to redeploy 'sleep' to trun on/off mTLS.

_docs/tasks/security/https-overlay.md
wattli: I do have the one line command listed as an alternative. Some users (including me) may feel one line is harder to debug (e.g., you can't ssh into the pod).

_docs/tasks/security/https-overlay.md
wattli: Good point, updated.

_docs/tasks/security/https-overlay.md
wattli: I do have the one line command listed as an alternative (right below). Some users (including me) may feel one line is harder to debug (e.g., you can't ssh into the pod) if the command does not work.

_docs/tasks/security/https-overlay.md
wattli: Good point, done.

_docs/tasks/security/https-overlay.md
```diff
@@ -0,0 +1,216 @@
...
```
diemtvu: Does this one need to be repeated for each demo? Maybe move the previous section

wattli: Done.

wattli:
Test passed and there is a lgtm label.

PR #1126, opened 2018-03-30
Fix more headers.

geeknoid:
Attempt 2.

Staging: https://geeknoid.github.io/istio.github.io/docs/setup/kubernetes/sidecar-injection.html

### DIFF
--- a/_docs/setup/kubernetes/sidecar-injection.md
+++ b/_docs/setup/kubernetes/sidecar-injection.md
@@ -45,7 +45,7 @@ cluster must satisfy the following requirements:
   CLI tool or automatically using the Istio Initializer.  Note that the
   sidecar is not involved in traffic between containers in the same pod.
   
-# Injection
+## Injection
 
 Manual injection modifies the controller configuration, e.g. deployment. It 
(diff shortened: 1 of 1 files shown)


_docs/setup/kubernetes/sidecar-injection.md
```diff
@@ -132,7 +132,7 @@ See the Kubernetes [quick start]({{home}}/docs/setup/kubernetes/quick-start.html
 
 Note that unlike manual injection, automatic injection occurs at the pod-level. You won't see any change to the deployment itself. Instead you'll want to check individual pods (via `kubectl describe`) to see the injected proxy.
 
-#### Installing the webhook
+## Installing the webhook
```
ldemailly: I'm pretty sure installing the webhook applies only to automatic injection and thus should be nested under automatic injection cc @ayj

ayj: `Installing the webhook`, `Deploying an app`,  and ` Understanding what happened` should be nested under `Automatic sidecar injection`.

cc @yusuoh

geeknoid:
Updated. PTAL.

PR #1129, opened 2018-03-31
Revamped nav header for better mobile experience.

geeknoid:
- We now only use the skinny version of the navbar instead of dynamically switching
based on viewport size. This looks cleaner, giving more screen space to the content rather than
our chrome.

- The search textbox is replaced with a search button. Clicking the button brings up the
search textbox. This looks less cluttered and works considerably better on smaller screens.

- When on a phone and the nav links are collapsed into a hamburger menu, cleanly show the
search box in the menu that comes up when you click the hamburger.

- Remove the down arrow next to the cog, it's superfluous and things look cleaner without
it.


geeknoid:
That's the gradient we use for preliminary builds.

preliminary.istio.io contains the light blue
istio.io has the normal blue
archive.istio.io has the darker blue

Thus, obviously, Istio darkens with age...

PR #1139, opened 2018-04-02
Add multicluster deployment documentation to Istio

sdake:


sdake:
@sebastienvas note clicking the link to the `OWNERS` file results in a 404.

sebastienvas:
@sdake that s because the file does not exist. @geeknoid  What should be the initial file ?

geeknoid:
For now, please list me and Lin as approvers.

sdake:
@sebastienvas right - ok makes sense :)  I guess in the bot convergence across repos, this OWNERS file was missed.

Cheers
-steve

sebastienvas:
Added the OWNERS file

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+
+* Istio secret `istio.default` replicated to all remote Kubernetes clusters
+from the Istio control plane per application namespace.
+
+* The local cluster and all remote Kubernetes cluster credentials (typically
+stored in `$HOME/.kube/config`) copied to
+`$HOME/multicluster/${UNIQUE_IDENTIFIER}.kube.conf`.
+
+* A [deployed Istio control plane]({{home}}/docs/setup/kubernetes/quick-start.html)
+on **one** Kubernetes cluster.  Note that Iatio must be [deployed with Helm]({{home}}/docs/setup/kubernetes/helm-install.html) using the `--set global.multicluster.enabled=true` flag.  This is currently mandatory.
```
linsun: typo Iatio -> Istio

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+type: markdown
+---
+
+{% include home.html %}
+
+Instructions for the installation of Istio multicluster.
+
+## Prerequisites
+
+* Individual cluster Pod CIDR ranges and service CIDR ranges must be unique
```
linsun: Is this absolutely a hard requirement?  I don't think this works for IBM Cloud.

sdake: I believe this is the case unless there is some advanced functionality added to Kubernetes in the future.  The main reason is when a multicluster mesh is instantiated, in order for deployments and their services not to collide (as one big mesh) the service IPs must not be allocated twice.  Something must do service CIDR allocation.  I could imagine an extra Kubernetes service in the future managing the service cidr allocations so multiple clusters could co-exist and use the same service CiDR.  The simple answer is simply to just force pod and service cidr allocations to always be unique.

Since routing is required between all nodes in the cluster for the PODs IP CIDRs to communicate with one another, a similar situation exists where no collision is a hard requirement.  Again something more advanced and centralized in Kubernetes could handle this.

I could also see a future where some type of egress/ingress gateway between multicluster isolated the pod and service CIDRs providing a true multicloud experience.

This document describes  the current merged state of the implementation.  I think it is important to get feedback from Operators (such as IBM Cloud) to determine the pain points around multicluster/multicloud, so those pain points can be prioritized, and those pain points can be resolved.

Without documentation/CI, the current state of the implementation cannot be understood nor used by anyone, so there would be no feedback loop as per Agile methods.

_docs/setup/kubernetes/multicluster-install.md
sdake: Not sure how I missed that, pushing an update now :)

_docs/setup/kubernetes/multicluster-install.md
sdake: @tiswanso can you validate the above statement I made please.

john-a-joyce: This is the simplest way to do it,  although there are other possibilities.  AFAIK no one has attempted any other networking setups so we definitely don't want to be recommending something else at the current time.  A couple things to add to the discussion.  The current solution requires a VPN (or a single RFC-1918 compliant network).   Note when the Zero VPN work is available some of it will apply to this solution as well.  
Considering that this is a single RFC-1918 compliant network implies that the CIDRs must not over-lap.  It also implies that there needs to be Ip reachability across the clusters.  Now it would be possible to set up your app deployment and Istio deployment so service IPs that are externally reachable are used for all endpoints and services that have to reach each other across the clusters.   Using the recently added gateway functionality may also make this easier to avoid exposing the Pod IPs across the clusters.  As I noted this hasn't been tested so in may not work in practice.  Additionally it is likely too cumbersome for anyone to want to deply it that way.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+
+   ```bash
+   helm install install/kubernetes/helm/istio-remote --name istio-remote --set pilotEndpoint=pod_ip_of_pilot_master --set mixerEndpoint=pod_ip_of_mixer_master`
+   ```
+
+### Mandatory Helm configuration parameters
+
+The isito-remote Helm chart requires the configuration of two specific variables defined in the following table:
+
+**Note** The `pilotEndpoint` and `mixerEndpoint` can also use DNS resolution, assuming DNS is setup to resolve
```
john-a-joyce: I am not sure what you are trying to say here.   The remote cluster will use kube-DNS to try to resolve the remote pilot (istio-pilot) and mixer (istio-mixer) endpoints.  It doesn't work any other way with the current code.   So each cluster must have a kube-dns entry for that service name.  how the cluster running pilot and mixer map those to an IP address is the part that has options.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+these IPs.
+
+| Helm Variable | Accepted Values | Default | Purpose of Value |
+| --- | --- | --- | --- |
+| `global.pilotEndpoint` | A valid IPv4 address | none | Specifies the Istio control plane's pilot Pod IP address |
+| `global.mixerEndpoint` | A valid IPv4 address | none | Specifies the Istio control plane's mixer Pod IP address |
+
+## Uninstalling
+
+** Note the uninstallation method must match the installation method (Helm or kubectl based) **
```
john-a-joyce: All the prior instructions are only for Helm.  So this should be reworked,  but per my comment above I think it would be great to also have the kubectl installation described.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+
+  ```bash
+  kubectl delete -f $HOME/istio-remote.yaml
+  ```
+
+### Using tiller to uninstall the istio-remote
+
+  ```bash
+  helm delete --purge istio-remote
+  ```
```
john-a-joyce: I think we need some way to point out limitations and issues.   Especially the mixer limitation on learning of the remote service naming.   Also istio-ca isn't multicluster aware.   I would prefer a section here,  but at the least we should create issues in istio that we can reference here.

_docs/setup/kubernetes/multicluster-install.md
linsun: I agree, we should explain why these pre-requisites exist and probably update the docs when the gateway functionality exists as that provides a way to easily expose services across clusters.

> Now it would be possible to set up your app deployment and Istio deployment so service IPs that are externally reachable are used for all endpoints and services that have to reach each other across the clusters.

I think this is also a good option too as users could expose all the services to be reachable within user's private network.  We have such an option in ibm cloud and could be leveraged.

sdake: @linsun I am not opposed to explaining the why of the prereqs, although it would be a strong delta from patterns in all of our other docs which don't explain why.  I am not keen to rework all the documentation in the repo to answer why questions (many that I don't have the answer to).

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+* Currently only [manual sidecar injection]({{home}}/docs/setup/kubernetes/sidecar-injection.html)
+has been validated with multicluster.
+
+## Install a multicluster configmap on the Istio control plane
+
+**Important**: Pilot will not start until the configmap in these instructions
+is created.  This is normal behavior but different from what is seen without
+multicluster enabled.
+
+Create a clusterregistry blob per Kubernetes cluster in $HOME/multicluster:
```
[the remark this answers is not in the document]
sdake: Perhaps this is poorly worded, however the clusterregistry blob is an example of one cluster node.  A descriptor is needed for each Kubernetes cluster and then they are merged into one configmap.  Any suggested wording changes?

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+## Deploy on the remote cluster
+
+Multicluster makes **one** Istio service mesh from all Kubernetes clusters
+that participate in multicluster.  Multicluster uses **one** Kubernetes
+cluster to run the Istio control plane.  **Any reasonable** number of remote
+Kubernetes clusters may be attached to the **one** Istio control plane.
+
+### Use kubectl with Helm to connect the remote cluster
```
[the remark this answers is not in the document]
sdake: That seems like follow-on work as currently this cannot be done without`helm template` which is an acceptable installation method as per community standards.

_docs/setup/kubernetes/multicluster-install.md
sdake: see review comment: https://github.com/istio/istio/pull/4655#discussion_r178597476 for rationale on this Note.  The **Note** can be removed, since nothing actually validates it works and has not been validated by a human.

_docs/setup/kubernetes/multicluster-install.md
sdake: There is no mechanism in the repository to customize templates except for `helm template`.  If Pod and Mixer IPs were discoverable, your proposal would be viable.  Note `tiller` is not a hard requirement, just `helm` and only to build the manifests.

_docs/setup/kubernetes/multicluster-install.md
sdake: An issue sounds good.  Please file one in the istio/istio repo and this repo with what you have in mind for improvements.  This simply documents the current state of the implementation and I don't have a strong desire to create sea salt with this PR.

_docs/setup/kubernetes/multicluster-install.md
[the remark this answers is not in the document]
john-a-joyce: Your wording in the comment might be a start.  " A description or blob is needed for each Kubernetes cluster (in the clusterregistry format) and then they are merged into one configmap. ".  The original wording suggests you want a configmap on each cluster.  Which is not the case.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+    config.istio.io/pilotCfgStore: True
+    config.istio.io/accessConfigFile: "falkor07.kube.conf"
+  spec:
+    kubernetesApiEndpoints:
+      serverEndpoints:
+        - clientCIDR: "10.23.230.0/28"
+          serverAddress: "10.23.230.11"
+```
+
+A unique configmap is required for each cluster.  In the above example,
```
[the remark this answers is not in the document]
sdake: your correct, configmap is the wrong here, and is confusing.  It is a single configmap with multiple stanzas.

Section is confusing, I'll rework.

_docs/setup/kubernetes/multicluster-install.md
sdake: you can deploy with kubectl if you so choose, if you render with Helm first which is what Istio does with other manifests.  The key here is to avoid a hard dependency on tiller (but things should work with tiller).

_docs/setup/kubernetes/multicluster-install.md
sdake: sounds like a good approach.  Thanks for the suggestion.

_docs/setup/kubernetes/multicluster-install.md
sdake: @linsun I've thought this over a bit, and will followup with a diagram that takes a look at explaining why/how for many of these prereqs.  I'd prefer to do that as a separate PR, as it seems tricky and I don't want to set a standard that all documentation must answer why/how.  At present, the docs generally answer what, and I think that is a pretty good approach to managing documentation.

Related to the request to updating docs with the gateway functionality, I've had a look at some of the work taking place there.  I think it looks very promising for providing a true multicloud experience and will of course commit to updating the docs when that work is integrated with multicluster/multicloud.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+
+layout: docs
+type: markdown
+---
+
+{% include home.html %}
+
+Instructions for the installation of Istio multicluster.
+
+## Prerequisites
```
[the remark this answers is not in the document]
sdake: done

_docs/setup/kubernetes/multicluster-install.md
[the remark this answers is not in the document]
sdake: reworked as per suggestions.

_docs/setup/kubernetes/multicluster-install.md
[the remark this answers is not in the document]
sdake: section reworked.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+        - clientCIDR: "10.23.230.0/28"
+          serverAddress: "10.23.230.11"
+```
+
+A unique configmap is required for each cluster.  In the above example,
+`falkor07.kube.conf` is the Istio control plane credential file. The
+config.istio.io/pilotCfgStore = `True` since this is the Istio control plane.
+For remotes, pilotCfgStore should = `False`.
+
+**Important**: The impelementation only uses the
```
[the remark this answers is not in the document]
sdake: thankfully the gate will find spelling errors in the future ;)

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,161 @@
...
+`config.istio.io/pilotCfgStore` and `config.istio.io/accessConfigFile`
+annotations, although every other annotation and spec is validated for
+correct syntax.  They may be set to dummy values, as long as they can
+be validated correctly.
```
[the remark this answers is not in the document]
sdake: accepted as per suggestion.

_docs/setup/kubernetes/multicluster-install.md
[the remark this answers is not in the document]
sdake: clarified.

_docs/setup/kubernetes/multicluster-install.md
sdake: clairified.

sdake:
rebased to pull in spell checking and automated linting. (plus branch was in merge conflict anyway).

sdake:
@linsun @geeknoid PTAL especially the dictionary expansion.

.spelling
```diff
@@ -435,6 +439,7 @@ roadmap
 roleRef
 rollout
 rollouts
+routeable
```
geeknoid: Please just use 'routable' instead.

.spelling
```diff
@@ -349,6 +351,8 @@ mixer-spof-myth-1
 mixer-spof-myth-2
 mongodb
 mtls_excluded_services
+multicloud
+multicluster
```
geeknoid: The 0.2 annoucement blog post uses multi-cluster, could you change the announcement to use multicluster please?

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+
+{% include home.html %}
+
+Instructions for the installation of Istio multicluster.
+
+## Prerequisites
+
+* Two or more Kubernetes clusters.
+
+* The usage of a RFC1918 network, VPN, or other more advanced network techniques
```
geeknoid: of an

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+
+    * All nodes' pod CIDR in every cluster must be routeable to every other nodes'
+pod CIDR.
+
+    * All Kubernetes control plane API servers must be routeable to each other.
+
+* Helm **2.7.2 or newer**.  The use of Tiller is optional.
+
+* If using Tiller, the ability to modify RBAC rules required to install Helm
+or alternatively Helm should already installed.
```
geeknoid: already be

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+from the Istio control plane per application namespace.
+
+* The local cluster and all remote Kubernetes cluster credentials (typically
+stored in `$HOME/.kube/config`) copied to
+`$HOME/multicluster/${UNIQUE_IDENTIFIER}.kube.conf`.
+
+* A [deployed Istio control plane]({{home}}/docs/setup/kubernetes/quick-start.html)
+on **one** Kubernetes cluster.  Note that Istio must be [deployed with Helm]({{home}}/docs/setup/kubernetes/helm-install.html) using the `--set global.multicluster.enabled=true` flag.  This is currently mandatory.
+
+* Currently only [manual sidecar injection]({{home}}/docs/setup/kubernetes/sidecar-injection.html)
```
geeknoid: Add a # link to the manual injection topic on the page.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+has been validated with multicluster.
+
+## Install a multicluster configmap on the Istio control plane
+
+**Important**: Pilot will not start until the configmap in these instructions
```
geeknoid: Remove the **Important**. and replace with ">" to create a block note. If this is really really important, you can also add the exclamation mark imagery (search for exclamation in the docs for example uses)

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+A unique security context is required to describe how to securely access
+each Kubernetes cluster in the system.  In the above example,
+`falkor07.kube.conf` is the Istio control plane credential file. The
+`config.istio.io/pilotCfgStore: True` flag is set in the condition this
+`clusterregistry.k8s.io` descriptor describes the Istio control plane Kubernetes
+cluster.  For remotes, please set `pilotCfgStore: False`.
+
+**Important**: The implementation only uses the
```
geeknoid: Use ">" instead of **Important**

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+
+   ```bash
+   helm install install/kubernetes/helm/istio-remote --name istio-remote --set pilotEndpoint=pod_ip_of_pilot_master --set mixerEndpoint=pod_ip_of_mixer_master`
+   ```
+
+### Mandatory Helm configuration parameters
+
+The `isito-remote` Helm chart requires the configuration of two specific variables defined in the following table:
+
+**Note** The `pilotEndpoint` and `mixerEndpoint` need to be resolvable via
```
geeknoid: Use ">"

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+Kubernetes.
+
+| Helm Variable | Accepted Values | Default | Purpose of Value |
+| --- | --- | --- | --- |
+| `global.pilotEndpoint` | A valid IPv4 address | none | Specifies the Istio control plane's pilot Pod IP address |
+| `global.mixerEndpoint` | A valid IPv4 address | none | Specifies the Istio control plane's mixer Pod IP address |
+
+## Uninstalling
+
+** Note the uninstall method must match the installation method (`kubectl` or `tiller` based) **
```
geeknoid: Use ">"

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+
+* Uninstall an Istio remote:
+
+  ```bash
+  kubectl delete -f $HOME/istio-remote.yaml
+  ```
+
+### Using tiller to uninstall the istio-remote
+
+  ```bash
```
geeknoid: Remove leaving spaces on this code block.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,172 @@
...
+### Mandatory Helm configuration parameters
+
+The `isito-remote` Helm chart requires the configuration of two specific variables defined in the following table:
+
+**Note** The `pilotEndpoint` and `mixerEndpoint` need to be resolvable via
+Kubernetes.
+
+| Helm Variable | Accepted Values | Default | Purpose of Value |
+| --- | --- | --- | --- |
+| `global.pilotEndpoint` | A valid IPv4 address | none | Specifies the Istio control plane's pilot Pod IP address |
```
geeknoid: Is this really only ipv4?

.spelling
sdake: will dio.

.spelling
sdake: right - my spelling is off.

_docs/setup/kubernetes/multicluster-install.md
sdake: To answer your question, I think there is not really a good reason for that to be the case other than IPV6 just became a reality in 1.10 of Kubernetes (iiuc).  @pmichali is taking a look at adding an IPv6 implementation to Istio; detailed here are the steps he  went through to get that running: http://blog.michali.net/2018/03/06/istio-on-ipv6-kubernetes-undiscovered-country/.  I have not tested IPv6 and dualstack (ipv4+ipv6 in same Kubernetes system) is non-existent at the moment.  So its either IPv4, or IPv6, depending on how you set up your system.

To clarify, I will change this to be "A valid IP address".

_docs/setup/kubernetes/multicluster-install.md
sdake: done.

_docs/setup/kubernetes/multicluster-install.md
sdake: done.

_docs/setup/kubernetes/multicluster-install.md
sdake: done.

_docs/setup/kubernetes/multicluster-install.md
sdake: done.

_docs/setup/kubernetes/multicluster-install.md
sdake: done.

_docs/setup/kubernetes/multicluster-install.md
sdake: done

_docs/setup/kubernetes/multicluster-install.md
sdake: done.

_docs/setup/kubernetes/multicluster-install.md
sdake: nice catch - done.

_docs/setup/kubernetes/multicluster-install.md
john-a-joyce: Istio - is not IPv6 capable at the moment so having Ipv4 here is factually correct,  but probably better to change this so there is one less place to fix to get IPv6 fully functional

_docs/setup/kubernetes/multicluster-install.md
john-a-joyce: What is your plan with Caveats?   Mixer is not multicluster capable and neither is istio-ca.  Should we just create issue to not those?

sdake:
apologies on the force push - git head in detached state.

geeknoid:
Do you have a staging site I can look at for the final rendered form?

sdake:
I render locally on my lab gear, but will spin up a GCP instance to render the docs.  Will need about 20 mins or so - and its dinner time, so t-2 hrs?

sdake:
@geeknoid give this a go: http://35.184.249.178:4000/docs/setup/kubernetes/

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,169 @@
...
+across the multicluster environment and may not overlap.
+
+    * All nodes' pod CIDR in every cluster must be routable to every other nodes'
+pod CIDR.
+
+    * All Kubernetes control plane API servers must be routable to each other.
+
+* Helm **2.7.2 or newer**.  The use of Tiller is optional.
+
+* If using Tiller, the ability to modify RBAC rules required to install Helm
```
linsun: I feel we are missing a verb here... If using Tiller, the ability to modify RBAC rules required -> requires? to install Helm

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,169 @@
...
+
+| Helm Variable | Accepted Values | Default | Purpose of Value |
+| --- | --- | --- | --- |
+| `global.pilotEndpoint` | A valid IP address | none | Specifies the Istio control plane's pilot Pod IP address |
+| `global.mixerEndpoint` | A valid IP address | none | Specifies the Istio control plane's mixer Pod IP address |
+
+## Uninstalling
+
+> The uninstall method must match the installation method (`kubectl` or `tiller` based)
+
```
linsun: replace `tiller` with `helm`?

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,169 @@
...
+## Deploy on the remote cluster
+
+Multicluster makes **one** Istio service mesh from all Kubernetes clusters
+that participate in multicluster.  Multicluster uses **one** Kubernetes
+cluster to run the Istio control plane.  **Any reasonable** number of remote
+Kubernetes clusters may be attached to the **one** Istio control plane.
+
+### Use kubectl with Helm to connect the remote cluster
+
+1. Use the helm template command on a remote to specify the Istio control plane:
```
linsun: for istio-remote, does user need to deploy it to each of the remote k8s clusters?  This was not clear to me.

linsun:
minor comments, otherwise looks good 

/lgtm

_docs/setup/kubernetes/multicluster-install.md
sdake: done.

_docs/setup/kubernetes/multicluster-install.md
sdake: Helm is also used in the kubectl install method.  What I did was clarify this "Helm + kubectl" or "Helm + Tiller".

It is not satisfying to me to have to require Helm just for templating the two mandatory variables.  This is something I want to resolve in the future.

_docs/setup/kubernetes/multicluster-install.md
sdake: clarified.

sdake:
@linsun thanks for picking up those problems.  There should be more clarity now.

sdake:
@sebastienvas You had mentioned on the ML that tide is being used for this repo.  I see that is the case after I just pushed an update.  One thing that is odd is the PR had  lgtm/approve/cla/passing checks/no merge conflicts flags for 16 hours, yet did not automatically merge.  In this ok, thats fine - as I needed to push 1 more update to fix reviewer comments.  Are manual merges currently required for this repo?  Looking over the closed PRs, it looks as though individual devs have been approving pulls (rather than a bot).

Is this inconsistent from istio/istio (which automatically merges), just a misunderstanding about how the automatic merging bot should work for this repo, a bug, or some other condition?

Cheers
-steve

sdake:
@hklai The PR looks good for merge yet tide hasn't merged it.  Can you take a look?

Cheers
-steve

hklai:
@sebastienvas 
The PR is in the MERGE state (https://prow.istio.io/tide.html) but is not merged. I don't seem to see why, do you?

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,178 @@
...
+
+If the prerequisites are met, the credentials for each Kubernetes cluster
+will also be present in `$HOME/multicluster`.
+
+Assemble and create the configmap from these files in the Kubernetes cluster
+where the Istio control plane is operational:
+
+```bash
+kubectl create configmap -n istio-system multicluster -f $HOME/multicluster
+```
```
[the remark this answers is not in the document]
baodongli: oh missed that. Didn't think it's a prerequisite.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Caveats and known problems
+
+<img src="{{home}}/img/exclamation-mark.svg" alt="Warning" title="Warning" style="width: 32px; display:inline" />
+All known caveats and known problems with multicluster for the 0.8 release are [tracked here](https://github.com/istio/istio/issues/4822).
+
+## Install a multicluster configmap on the Istio control plane
+
+<img src="{{home}}/img/exclamation-mark.svg" alt="Important" title="Important" style="width: 32px; display:inline" />
+Pilot will not start until the configmap in these instructions
```
costinm: This seems to be a short-term limitation - I think the plan (with the code using programmatic loading of configmap) is to watch for the config map to show up and change. But it's probably ok for now, dynamic adding of clusters is low priority

( this is mainly due to primary cluster loaded from configmap - will be fixed in 0.8 with the secrets change)

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,178 @@
...
+where the Istio control plane is operational:
+
+```bash
+kubectl create configmap -n istio-system multicluster -f $HOME/multicluster
+```
+
+## Deploy on the remote cluster
+
+Multicluster makes **one** Istio service mesh from all Kubernetes clusters
+that participate in multicluster.  Multicluster uses **one** Kubernetes
```
costinm: That's debatable: mixer can be in multiple clusters, CA will likely have to be in all clusters for MTLS on.
Pilot could be run in multiple clusters - with additional config for the secondary clusters (to read configs from the primary cluster, or CI/CD).

Maybe we can say 'in simple/basic installs, one kubernetes cluster is used to hold the istio configs, and pull endpoint/service from all clusters'

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,178 @@
...
+cluster to run the Istio control plane.  **Any reasonable** number of remote
+Kubernetes clusters may be attached to the **one** Istio control plane.
+
+<img src="{{home}}/img/exclamation-mark.svg" alt="Important" title="Important" style="width: 32px; display:inline" />
+The istio-remote component must be deployed to each remote Kubernetes cluster.
+
+### Use kubectl with Helm to connect the remote cluster
+
+1. Use the helm template command on a remote to specify the Istio control plane:
+
```
costinm: Add some note on what this is doing ? (create a pilot/mixer reference in each cluster, pointing to the main pilot )

I assume there is a missing step here - the pilot must be exposed as an internal load balancer ( to get the IP ) ?

IMPORTANT: mixer has changed into 2 components, I think the remote template was not updated.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,178 @@
...
+   ```bash
+   helm template install/kubernetes/helm/istio-remote --name istio-remote --set pilotEndpoint=`pod_ip_of_pilot_master` --set mixerEndpoint=`pod_ip_of_mixer_master` > $HOME/istio-remote.yaml
+    ```
+
+1. Instantiate the remote cluster's connection to the Istio control plane:
```
costinm: Maybe mention that another option is to define an internal DNS entry for pilot ( and possibly the 2 mixer components), and just use that as 'discovery address' when injecting in all remote clusters.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,178 @@
...
+   helm install install/kubernetes/helm/istio-remote --name istio-remote --set pilotEndpoint=pod_ip_of_pilot_master --set mixerEndpoint=pod_ip_of_mixer_master`
+   ```
+
+### Mandatory Helm configuration parameters
+
+The `isito-remote` Helm chart requires the configuration of two specific variables defined in the following table:
+
+| Helm Variable | Accepted Values | Default | Purpose of Value |
+| --- | --- | --- | --- |
+| `global.pilotEndpoint` | A valid IP address | none | Specifies the Istio control plane's pilot Pod IP address |
```
costinm: ILB address - the pod address will change too often.

_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -0,0 +1,178 @@
...
+
+* If using Tiller, the ability to modify RBAC rules is required to install Helm
+or alternatively Helm should already be installed.
+
+* Kubernetes **1.7.3 or newer**.
+
+* Istio secret `istio.default` replicated to all remote Kubernetes clusters
+from the Istio control plane per application namespace.
+
+* The local cluster and all remote Kubernetes cluster credentials (typically
```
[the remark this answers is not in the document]
sdake: agreed.

sdake:
@costinm thanks for the great comments.  I will address existing issues today in a new PR.  An ILB makes good sense for Pilot/Mixer.  I'll submit a PR that makes that happen.  I am aware the Pod IPs change too often - and think it is a big limitation of the current implementation.  Thanks for making the ILB suggestion.

RE forward looking comments I am doing my best to make this documentation match the reality of the implementation, not what it should be or we wish it to be.  I have a separate PR in the works to be ready to go with the secrets changes in @sbezverk 'z work once his work merges.

Cheers
-steve

_docs/setup/kubernetes/multicluster-install.md
sdake: The step that is missing is the helm template is missing any ILB logic.  It only has Pod IPs, so the documentation matches the implementation as is today.  Thanks for the suggestion though - will work on a PR to solve this problem..

yup  I caught that change - will update the remote.  Longer term the remote can easily merge into the main template, just wasn't keen to overload that functionality until the remote definition was finalized.

PR #1142, opened 2018-04-03
Add labels over community icons in the footer.

geeknoid:

Hover your mouse over the footer community icons to see the labels.


Gemfile.lock
```diff
@@ -28,38 +28,42 @@ GEM
...
+    ffi (1.9.23)
     forwardable-extended (2.6.0)
     gemoji (3.0.0)
-    github-pages (172)
+    github-pages (180)
       activesupport (= 4.2.9)
-      github-pages-health-check (= 1.3.5)
-      jekyll (= 3.6.2)
+      github-pages-health-check (= 1.4.0)
+      jekyll (= 3.7.3)
```
ldemailly: @geeknoid do we require those changes ? if can you update https://github.com/istio/istio.github.io#working-with-the-site

ldemailly: for me bundle install now hangs at
Installing commonmarker 0.17.9 with native extensions

ldemailly: it was hanging because I had a go binary called "echo" in my path (!)

PR #1138, opened 2018-04-02
Simplify the demo of plugin ca cert.

wattli:

PR #1145, opened 2018-04-03
Fix bug caused by #1138

wattli:


geeknoid:
Consider cherry-picking to the release-0.7 branch.

PR #1146, opened 2018-04-04
Switch back to normal html-proofer

### LINKED ISSUE #849: html-proofer 3.8.0 contains the patch we needed about trailing / - upgrade from custom
https://github.com/istio/istio.github.io/blob/master/Gemfile#L6

@ldemailly ^^

I have an upcoming PR to fix broken links and i've updated to latest html proofer 3.7.0. That should be sufficient, right?

ldemailly:
As my pr was merged

Fixes #849


ldemailly:
do we really need the bot in this repo ? and if we do, can it work/do something ?
(it didn't merge the PR)

geeknoid:
​I think the bot is still being setup, so it's not fully functional.
​

PR #1154, opened 2018-04-05
Make spell checking and style checking part of our doc checkin gate.

geeknoid:


scripts/linters.sh
```diff
@@ -1,3 +1,3 @@
 mdspell --en-us --ignore-acronyms --ignore-numbers --no-suggestions --report *.md */*.md */*/*.md */*/*/*.md */*/*/*/*.md
```
ldemailly: there should be a
```
#! /bin/sh
# This runs 3 checks.... [some doc/header]
set -x
set -e
```
the set -e is to stop on first error, set -x to see what it is doing

ldemailly: edit: but it's good indeed to separate the stages in circle so it's more clear what it is doing and where it may fail/stop

.circleci/config.yml
```diff
@@ -17,32 +17,45 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-          - v3-dependencies-{{ checksum "Gemfile.lock" }}
+          - v4-dependencies-{{ checksum "Gemfile.lock" }}
           # fallback to using the latest cache if no exact match is found
-          - v3-dependencies-
+          - v4-dependencies-
```
ldemailly: that's going to be a bit of an issue, you'd have to manually change - can you keep at least the {{ checksum "Gemfile.lock" }} so if a new gem change is needed it'll just work and would only be manual change v4->v5 if changing markdown etc... there is also probably a "proper" lock file of sort for npm too

.circleci/config.yml
```diff
@@ -17,28 +17,45 @@ jobs:
...
+
+      - run:
+          name: Checking markdown spelling
+          command:
+            .vendor/node/node_modules/markdown-spellcheck/bin/mdspell --en-us --ignore-acronyms --ignore-numbers --no-suggestions --report *.md */*.md */*/*.md */*/*/*.md */*/*/*/*.md
+
+      - run:
+          name: Checking markdown style
+          command:
+            mdl --ignore-front-matter --style mdl_style.rb .
```
ldemailly: I think in theory this should be "bundle exec mdl ..." but if it works as is it's fine

ldemailly: also in theory I think it could/should go inside
https://github.com/istio/istio.github.io/blob/master/Rakefile

geeknoid:
Most of the build time in Jekyll is spent sorting and resorting the site pages in order to produce the left hand nav bar. The logic for that is here: https://github.com/istio/istio.github.io/blob/master/_includes/sort-hierarchy.html. What makes this code painful is that the Liquid language has no functions, no data structures, no recursions, and is incredibly slow. Second, the sorting logic has to run on every page of the site. When working on the site, I go to _layouts/docs.html and change docs=site.docs to docs=site.blog which cuts the total build time by 80% by using the small blog data set to produce the main doc sidebar. Its incorrect, but it is fast. Anyway, if you can think of a better way to sort the site properly, I'd love to see it :-)

I'll do the follow on tests. I did already see that the style checker doesn't actually report an error code when it finds errors. So for now, it'll dump stuff out on the console, but won't prevent the presubmit from failing. I'll look for a solution for that.

PR #1157, opened 2018-04-05
Remove https-overlay task since the files are not in 0.7.1 release

wattli:


geeknoid:
An FAQ page has a pointer to the http-overlay page it appears...

wattli:
Good point, removed as well.

ldemailly:
why not just add a pointer to whichever file is missing ?

also which file is missing/why ?

Issue #1156, opened 2018-04-05
Quick Start Kubernetes 


jsenon:
Hi,

In Quick start Kubernetes documentation path in `Installation steps` section 3 is not correct, I would like to modify it but it's automatically replace by git branch name. So we have `cd istio-0.7` but Its `cd istio-0.7.1`. Appreciate If somebody explain me how to correct this (create branch release-0.7.1 seems not the correct way to do)

sdake:
@jsenon good catch with this problem.  I believe the documentation is suitable as is, but it is a little confusing and will need a resolution prior to 1.0.z.  I'm not really sure the best way to solve this problem.

A branch from the 0.7 branched hash would solve this problem, however, I'm not sure we want to get into branching z streams for documentation prior to 1.0.z.

One solution (that may be a little over-the-top):

* submit a hardcode of the version to master
* cherrypick patch to 0.7 branch
* revert the hardcode of the version to master

This keeps a "master-first" approach - which is a best practice without the problems of branching a 0.7.1.  This may also be too much.

Given that the instructions state:

https://istio.io/docs/setup/kubernetes/quick-start.html#installation-steps
(step 3)
```
Change directory to istio package. For example, if the package is istio-0.7
cd istio-0.7
```

A third option would be to just ignore the problem and expect the operator to read the conditional "for example, if"

Perhaps @geeknoid has a better solution in mind than the 3 choices above.

Cheers
-steve

geeknoid:
We definitely don't want to branch docs on patch versions, that'd be too much.

I think the text is clear enough as-is, it's just stating an example.

geeknoid:
As I said, I think it's OK.

PR #1161, opened 2018-04-06
Force change directory in the Quick Start doc

jsenon:
As discuss in #1156, I propose to force the change directory with static value, instead of create a new branch (due to automatic parameter {{site.data.istio.version}} )

### DIFF
--- a/_docs/setup/kubernetes/quick-start.md
+++ b/_docs/setup/kubernetes/quick-start.md
@@ -187,7 +187,7 @@ curl -L https://git.io/getLatestIstio | sh -
 
 1. Change directory to istio package. For example, if the package is istio-{{site.data.istio.version}}
 ```bash
-cd istio-{{site.data.istio.version}}
+cd istio-0.7.1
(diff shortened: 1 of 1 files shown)


_docs/setup/kubernetes/quick-start.md
```diff
@@ -187,7 +187,7 @@ curl -L https://git.io/getLatestIstio | sh -
 
 1. Change directory to istio package. For example, if the package is istio-{{site.data.istio.version}}
 ```bash
-cd istio-{{site.data.istio.version}}
+cd istio-0.7.1
```
sdake: I forgot to add my comment here:
This would require maintenance prior to branching.  It is also inconsistent with line 188.  I know there are other places in the docs where we want to remove maintenance burden on version changes and we don't want to incrementally make the problem worse :).  @geeknoid had indicated in the issue tracker for this issue that the wording used "For example" was good enough (and I think was meant to solve this particular problem).

PR #1163, opened 2018-04-07
Make istio.io support quality print output.

geeknoid:
- Get rid of all the chrome when printing a page. So no headers, sidebars, etc.

- Generate endnotes for each page printed which lists the URLs of the various links
on the page. Each link site is annotated with a superscript number referencing this
table.


sdake:
Gradients look more visually pleasing.  I don't see any delta in print mode though.  I don't know much about HTML (my last HTML work was with vi in 1999 making a website for my Father's business).  I don't have anything to offer on the technical review of the work.

Cheers
-steve

geeknoid:
@sdake Can you clarify? Do you want gradient somewhere?

My goal with this first print-friendly effort was to get rid of the chrome (nav bars, sidebars, etc) so that the printed page looks good.

After that, I'll improve the CSS to prevent some elements from being broken up across page boundaries, I'll add support for footnotes that show the full URLs of the different links, will improve colors a bit, and will do other minor tweaks. 

But gradients weren't on my list :-)

sdake:
Maybe this wasn't your change, but there is a gradient in the change (and it looks good).  I'll check out the other elements you mentioned.

Screenshot of your rendered page:
screen shot 2018-04-08 at 7 46 53 am

vs. Screenshot of istio.io:
screen shot 2018-04-08 at 7 47 29 am

sdake:
I see the delta now in print mode - should print much better now.

Cheers
-steve

sdake:
btw I'm the last person you want judging web page user interfaces :)

geeknoid:
Ah, that gradient. We have three color schemes for Istio:

archive.istio.io - old stuff, in a darker blue
istio.io - current stuff, in "Istio" blue
preliminary.istio - future stuff, in a lighter blue

They all three have a gradient on the landing page and in the
header/footer, it's just that it's more apparent with the
preliminary.istio.io colors (which is what staging sites build as
nowadays).

Of course, printing web pages is really intended for content pages, I don't
expect many people to print the home page :-)

sdake:
@geeknoid thanks for the explanation.  My eyesight isn't all that fantastic so have trouble seeing the other gradients, but if you say they are there, they are ;)  Agree printing is for content pages not headers.

I went through several and took a look at print mode, and IMO this PR works as intended.

Cheers
-steve

ZackButcher:
This is a nice improvement over today, but there's still a few problems:
- Some of the context overlaps the print headers
- The code boxes aren't expanded, which means they're still scroll areas and don't actually render when printing.

Here's a screenshot with both issues:

image

IMO we should also drop the styling on the boxes in print mode, and just outline them (get rid of the bar on the left of the code blocks, basically)

geeknoid:
I fixed the scroll bars on PRE blocks, those are gone. I kept the thick border on the left though for the sake of consistency with the web view.

As for the content overlapping the header/footer, that's really out of my control. The header/footer are things inserted by the browser around the rendered content. The only way I was able to repro your case in Chrome was to set a custom margin in the print dialog and to explicitly make the margin too skinny for the header/footer to fit in. Is this your case per chance? Otherwise, using Chrome's Minimum or Default margins, there was plenty of room for the header/footer.

Thanks.

PR #1173, opened 2018-04-12
Improve sidenav behavior on mobile.

geeknoid:
The sidenav now hovers over the main text instead of pushing the main
text sideways which both looks cooler and is less disruptive to the visuals.

The rendering of the sidenav toggler button now matches the "back to top"
button I added last week.


geeknoid:
Ping.

PR #1181, opened 2018-04-14
Bunch of improvements

geeknoid:
- New visuals for the sailboat in the header. It now overflows the header.

- The TOC now highlights the currently displayed portion of the current page.
As you scroll through the doc, the selected entry updates accordingly.

- Add previous/next page links in every doc page. These used to be present only in
blog posts, but they're useful everywhere.

- Fix a few off-by-one formatting errors that stemed from using a mixture of
min-width and max-width throughout the stylesheet. This caused some strange
formatting to happen at specific window widths. Now, we're consistently using
min-width and everything lines up properly.

- Improved footer formatting so it looks better on mobile.

- Only display the TOC on XL screens, otherwise it wraps too much.
Screens smaller than XL now all get the inlined TOC instead.

- Add support for pages to request that the TOC be generated inline instead of in a sidebar.
This is useful for pages that have headings which cause too much wrapping in the TOC,
such as the Troubleshooting Guide.

- Add some blank space between an inlined TOC and the main text so that things don't look
so crowded, especially when printing.

- Inline the sailboat SVG into each page. This avoids a network roundtrip and allows the
SVG to be controlled with the same CSS as everything else.

- Eliminate a huge amount of redundancy in the four main layout files for the site.
They now share a single primary.html include file which carries most of the weight. This
will avoid having to constantly make the same change in four different files.

- Improve the generated HTML for FIGURE elements which makes
things better for screen readers.

- Simplify the HTML & CSS for the footer.


ZackButcher:
It all looks good except for the animation on the sailboat on page load: IMO it's fairly annoying. I really dislike anything shifting on the page after load, and I don't think that the animation is highlighting anything important (i.e. by making it move we're visually signalling that it's important, but really it's just the logo).

geeknoid:
Animation nuked...

Thanks.

PR #1184, opened 2018-04-17
Switch to PrismJS for syntax highlighting.

geeknoid:
Instead of doing syntax highlighting statically in Jekyll, we now
go back to the PrimsJS library we used in the 0.2-0.4 timeframe.
It used to be problematic, but the cause for the problems has
been addressed a while ago.

This gives us highlighting for non-markdown content,
such as dynamically loaded PRE blocks and PRE blocks that
come from HTML generated from protos.


sdake:
I see preliminary site, but would still like to render (unless it is static?).

sdake:
@geeknoid site looks fantastic!  Especially like the TOC on the right side.  I clicked around through the site, and everything seems to work well enough.

Note this page doesn't seem to be consistent with the rest of the site:
https://geeknoid.github.io/istio.github.io/help/troubleshooting.html

robcresswell:
👋 Hi there,

Just weighing in because Steve asked me to take a look, so please appreciate there may be an aspect of naivety in my comments in regards to the overall impact. In terms of what I see here, Prism is a solid solution (it's lightweight and dynamic) to the issue specified (dynamically injected `<pre>` blocks, etc.) and has been used before, so it's not a radically new solution.

There are a couple of things that would be nice to see in the commit message:

- Could you elaborate a little on "It used to be problematic, but the cause for the problems has
been addressed a while ago." ? It would be nice for those of us without historic knowledge to be able to better grok these changes and understand the context.
- Is there a long term plan to address what may be a deeper issue? I may be missing a point, but it seems like all of the content in this documentation should be available at compile / build time, so it seems perhaps that a gap in the toolchain is then needing to be patched by this client side syntax highlighting. That said, perhaps it is not worth the engineering hours to solve (again, I lack wider context here). If there is a desire to rearchitect this, I'd suggest looking at Hugo instead of Jekyll; it's in Go rather than Ruby, has significantly faster build times, and has a type of pluggable content called "shortcodes" that may let you insert items like the "copy" button without needing to do it dynamically at runtime.

Hope this helps a little, cheers.

sdake:
thanks @robcresswell 
makes sense @geeknoid 

Cheers
-steve

PR #1186, opened 2018-04-17
Adding info about new expression language methods.

ozevren:
Adding info about dnsName, email, and uri functions.


geeknoid:
Regarding the CircleCi failure: replace 'dns' with 'DNS'. Is that still doesn't fly, then add DNS to the .spelling file at the root of the doc repo.

PR #1190, opened 2018-04-19
v1alpha3 routing blog

frankbu:


rshriram:
client->[virtualService]->[DestinationRule] --> server : [] indicates optional components
client->gateway->virtualService->[DestinationRule] ->server
client->ExternalService->[virtualService]->[DestinationRule]->server

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,334 @@
...
+{% include home.html %}
+
+Up until now, Istio has provided a simple API for traffic management using four configuration resources:
+`RouteRule`, `DestinationPolicy`, `EgressRule`, and (Kubernetes) `Ingress`.
+With this API, users have been able to easily manage the flow of traffic in an Istio service mesh.
+The API has allowed users to route requests to specific versions of services, inject delays and failures for resilience
+testing, add timeouts and circuit breakers, and more, all without changing the application code itself.
+
+While this functionality has proven to be a very compelling part of Istio, user feedback has also shown that this API does
+have some shortcoming, specifically when using it manage very large applications containing thousands of services, and
```
geeknoid: it to manage

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,334 @@
...
+when working with protocols other than HTTP. Furthermore, the use of Kubernetes `Ingress` resources to configure external
+traffic has proven to be woefully insufficient for our needs.
+
+To address these, and other concerns, a new traffic management API, a.k.a. `v1alpha3`, is being introduced, which will
+completely replace the previous API going forward. Although the `v1alpha3` model is fundamentally the same, it is not
+backward compatible and will require manual conversion from the old API. A
+[conversion tool]({{home}}/docs/reference/commands/istioctl.html#istioctl%20experimental%20convert-networking-config)
+is included in the the next few releases of Istio to help with the transition.
```
geeknoid: the the

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,334 @@
...
+
+To justify this disruption, the `v1alpha3` API has gone through a long and painstaking community
+review process that has hopefully resulted in a greatly improved API that will stand the test of time. In this article,
+we will introduce the new configuration model and attempt to explain some of the motivation and design principles that
+influenced it.
+
+## Design Principles
```
geeknoid: Only capitalize the first letter of headings, here and elsewhere

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,334 @@
...
+
+As you can see, both of the rules for the `reviews` service are consolidated in one place, which at first may or may not
+seem preferable. However, if you look closer at this new model, you’ll see there are fundamental differences that make
+`v1alpha3` vastly more functional.
+
+First of all, notice that the destination service for the `VirtualService` is specified using a `hosts` field (repeated field,
+in fact) and is then again specified in a `destination` field of each of the route specifications. This is a very important
+difference from the previous model.
+
+A `VirtualService` describes the mapping between one, or more, user-addressable destinations to the actual destination services (workloads) inside the mesh. In our example, they are the same, however, the user-addressed hosts can be any DNS
```
geeknoid: one, or more -> one or more

geeknoid: I think it should just say "actual destination workloads".

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,334 @@
...
+
+`ExternalService` is used to add additional entries into the service registry that Istio maintains internally. This allows
+one to model traffic to external dependencies of the mesh such as APIs consumed from the web or traffic to services in
+legacy infrastructure.
+
+Replacing `EgressRule` from the previous API, `ExternalService` provides several significant improvements, such as:
+
+1. Individual service ports and service discovery mode are now configurable.
+2. One or more (different) external endpoints can be configured to implement a “virtual” external service.
+3. Secure HTTP services (automatic TLS upgrade) can now be accessed using standard https (e.g., https://secureservice.com/
```
geeknoid: Put backticks around URLs here and below.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,334 @@
...
+   instead of http://secureservice.com:443/.
+4. Multiple CIDR subsets can now be included in a single `ExternalService` configuration.
+
+Because an `ExternalService` configuration simply adds an external destination to the internal service registry, it can be
+used in conjunction with a `VirtualService` and/or `DestinationRule`, just like any other service in the registry.
+
+### Gateway
+
+It all started with ingress. The Istio ingress feature inherited the Kubernetes `Ingress` resource model for expediency, but
+unfortunately it is not able to express all of the routing capabilities of Istio. The `Ingress` APIs are inadequate to model
```
geeknoid: all of Istio's routing capabilities

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,334 @@
...
+    route:
+    ...
+```
+
+Although primarily used as an ingress allowing external traffic into the mesh, a `Gateway` can also act as an egress to
+allow traffic in the mesh to exit. A `Gateway` can also model a proxy that is entirely internal to the mesh and implemented
+either by sidecars or by a middle proxy.  Ingress and egress gateways can also expose non HTTP services with the same
+ease.
+
+A `Gateway` simply configures a loadbalancer, regardless of where it will be running. Any number of gateways can exist
```
geeknoid: load balancer

geeknoid:
I agree with @ermik , this could use a diagram or two...

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,334 @@
...
+
+1. [VirtualService]({{home}}/docs/reference/config/istio.networking.v1alpha3.html#VirtualService) replaces `RouteRule`
+2. [DestinationRule]({{home}}/docs/reference/config/istio.networking.v1alpha3.html#DestinationRule) replaces
+   `DestinationPolicy`
+3. [ExternalService]({{home}}/docs/reference/config/istio.networking.v1alpha3.html#ExternalService) replaces `EgressRule`
+4. [Gateway]({{home}}/docs/reference/config/istio.networking.v1alpha3.html#Gateway) is the recommended replacement
+   for (Kubernetes) `Ingress`
+
+The old `RouteRule`, `DestinationPolicy`, and `EgressRule` configuration resources will be completely removed and no longer
+available in future releases of Istio. Istio `Ingress`, however, will continue to be available but using a `Gateway` provides
```
linsun: it may be good to clarify in future releases of Istio after v0.8.

_blog/2018/v1alpha3-routing.md
```diff
@@ -21,22 +21,22 @@ The API has allowed users to route requests to specific versions of services, in
 testing, add timeouts and circuit breakers, and more, all without changing the application code itself.
 
 While this functionality has proven to be a very compelling part of Istio, user feedback has also shown that this API does
-have some shortcoming, specifically when using it manage very large applications containing thousands of services, and
+have some shortcoming, specifically when using it to manage very large applications containing thousands of services, and
```
geeknoid: shortcoming -> shortcomings

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,347 @@
...
+## Design principles
+
+A few key design principles played a role in the routing model redesign:
+
+* Explicitly model infrastructure as well as intent. For example, in addition to configuring an ingress gateway, the
+  component (controller) implementing it can also be specified.
+* The authoring model should be "producer oriented" and "host-centric" as opposed to compositional. For example, all
+  rules associated with a particular host are configured together, instead of individually.
+* Clear separation of routing from post-routing behaviors.
+
```
linsun: might be good to clarify "post-routing behaviors" -> "post-routing behaviors as defined in destination rules".  It took me a while to realize that is what you are referring to here.

_blog/2018/v1alpha3-routing.md
```diff
@@ -324,6 +324,24 @@ to a particular workload by specifying the set of workload (pod) labels as part
...
 
+## Creating and deleting v1alpha3 route rules
+
+Because all route rules for a given destination are now stored together as an ordered
+list in a single `VirtualService` resource, adding subsequent rules for a particular destination
+is no longer done by creating a new (`RouteRule`) resource, but instead by updating the one-and-only `VirtualService`
+resource for the destination.
+
+old routing rules:
+```
```
geeknoid: Need to add "bash"  here and below.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,347 @@
...
+
+A `Gateway` simply configures a load balancer, regardless of where it will be running. Any number of gateways can exist
+within the mesh and multiple different gateway implementations can co-exist. In fact, a gateway configuration can be bound
+to a particular workload by specifying the set of workload (pod) labels as part of the configuration, allowing users to
+reuse off the shelf network appliances by writing a simple gateway controller. `Gateway` is much more general purpose than
+simply an ingress replacement.
+
+## Summary
+
+The Istio `v1alpha3` routing API is significantly more functional than its predecessor, but unfortunately not backwards
```
ijsnellf: tone nit: instead of emphasizing `functional`, emphasize `functionality`. v1alpha1 was functional, it just didn't have as many features.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,347 @@
...
+    route:
+    - destination:
+        name: reviews
+  - match:
+    - uri:
+        prefix: /ratings
+    route:
+    - destination:
+        name: ratings
+  ...
```
ermik: I suggest removing k8s-style-documentation dots. Usually they signify that there is more to yaml-formatted command output than shown in example e.g. case `kubectl describe`. This will remedy a plethora of copy-paste issues.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,347 @@
...
+```
+
+The hosts of a `VirtualService` do not actually have to be part of the service registry, they are simply virtual
+destinations. This allows users to model traffic for virtual hosts that do not have routable entries inside the mesh.
+These hosts can be exposed outside the mesh by binding the `VirtualService` to a `Gateway` configuration for the same host
+(see [Gateway](#gateway), below).
+
+In addition to this fundamental restructuring, `VirtualService` includes several other important changes:
```
ermik: I suggest moving this paragraph higher up to provide further context to the changes before showing configuration examples. Otherwise, the reader will be repeatedly perplexed while trying to read the configuration (e.g. "What is this `subset` thing? -> Cmd-F 'subset' enter-enter-enter...").  Overall, IMHO, the narrative in each section should be: 

- Describe problems with v1alpha1(2): now it is well described, redundant route rules and all
- Introduce new concepts for v1alpha3: that's where short descriptions of terms such as 'subset' should be mentioned
- Provide examples of the configuration changes (code excerpts with commentary)
- Close section by delving deeper into the reasons behind the change and expanding on its ontology

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,347 @@
...
+* Define resilience features such as retry & circuit breaking behavior
+* Load balance TCP traffic
+
+Another problem with the `Ingress` specification is that it tries to draw a common denominator across different L7 proxies.
+This results in a least common denominator that uses the most basic HTTP routing and discards every other feature of the
+proxy, which end up in annotations that are not portable.
+
+So the `Gateway` effort started. Istio `Gateway` overcomes the `Ingress` shortcomings by separating the L4-L6 spec from L7.
+It only configures the L4-L6 functions (e.g., ports to expose, TLS configuration) that are uniformly implemented by all good
+L7 proxies. Users can then use standard Istio route rules to control HTTP requests & TCP traffic entering a `Gateway` by
```
ermik: Shouldn't this be referring to `DestinationRule`? If I understand correctly `RouteRule` is on deprecation track. It currently is a well-adopted descriptor and to avoid confusion should not be used when referring to the combination of `Gateway`, `DestinationRule`, and `VirtualService`. In the future, I am sure, the community will call it that all on _our_ own.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,347 @@
...
+    hosts:
+    - bookinfo.com
+    tls:
+      mode: SIMPLE
+      serverCertificate: /tmp/tls.crt
+      privateKey: /tmp/tls.key
+```
+
+In this example, the `hosts` field specifies bookinfo.com concretely, but in general the binding can be much more flexible.
+A single `Gateway` can bind to multiple `VirtualServices` or a single `VirtualService` can be exposed on more than one
```
ermik: Please gently highlight this.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,347 @@
...
+```yaml
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: bookinfo
+spec:
+  hosts:
+    - bookinfo.com
+  gateways:
+  - bookinfo-gateway # <---- bind to gateway
```
ermik: I want to use this opportunity to suggest Pep8-style inline comments throughout core examples, such as:

```yaml
apiVersion: networking.istio.io/v1alpha3        # New API version
kind: VirtualService                            # replaces RouteRule
metadata:
  name: reviews
spec:                                           # new spec defines:
  hosts:                                        # destinations for this route
    - reviews                                   # in this case 'reviews' service
```
I would gladly take on this additional work if I could vouch for my own technical fluency.  Don't shoot the messenger.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,365 @@
...
+    route:
+    ...
+```
+
+Although primarily used as an ingress allowing external traffic into the mesh, a `Gateway` can also act as an egress to
+allow traffic in the mesh to exit. A `Gateway` can also model a proxy that is entirely internal to the mesh and implemented
+either by sidecars or by a middle proxy.  Ingress and egress gateways can also expose non HTTP services with the same
+ease.
+
+A `Gateway` simply configures a load balancer, regardless of where it will be running. Any number of gateways can exist
```
costinm: Unfortunately a gateway requires additional setup (at least for now): to get the internal or external IP and to open firewall/forwarding rules it needs associated K8S Service.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,365 @@
...
+
+Replacing `EgressRule` from the previous API, `ExternalService` provides several significant improvements, such as:
+
+1. Individual service ports and service discovery mode are now configurable.
+2. One or more (different) external endpoints can be configured to implement a “virtual” external service.
+3. Secure HTTP services (automatic TLS upgrade) can now be accessed using standard https (e.g., `https://secureservice.com/`
+   instead of `http://secureservice.com:443/`.
+4. Multiple CIDR subsets can now be included in a single `ExternalService` configuration.
+
+Because an `ExternalService` configuration simply adds an external destination to the internal service registry, it can be
```
costinm: Maybe we should mention ExternalService is the new way to represent 'mesh expansion', i.e. VMs running istio outside a cluster but still as part of the mesh.

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,365 @@
...
+intended to be authored by service owners, describing the circuit breakers, load balancer settings, TLS settings, etc.. 
+`DestinationRule` is more or less the same as its predecessor, `DestinationPolicy`, with the following exceptions:
+
+1. The `host` of a `DestinationRule` can include wildcard prefixes, allowing a single rule to be specified for many actual
+   services.
+2. A `DestinationRule` defines addressable `subsets` (i.e., named versions) of the corresponding destination host. These
+   subsets are used in `VirtualService` route specifications when sending traffic to specific versions of the service.
+   Naming versions this way allows us to cleanly refer to them across different virtual services, simplify the stats that
+   Istio proxies emit, and to encode subsets in SNI headers.
+
```
rshriram: Another big thing to add here: wildcard destination rules to address multiple services using a single rule. And per-port customization

_blog/2018/v1alpha3-routing.md
rshriram: Are you sure you want to go down this path? The code to treat external services is different. It does not set up istio auth. You have to configure tls using destination rules. Mesh expansion assumes that services are part of the mesh and honor auth, mixer, etc. 

External service assumes that traffic is going to potentially untrusted/not in istio control, entities. The mixer configuration is different as well.

I would be skeptical of overriding this behavior with dual semantics. Put another way, being part of service registry is not equivalent to being part of the mesh. 

Why not add a simple MeshService resource that allows you to statically add services to the mesh ? Or add label annotations to endpoints that allow you to supply label info for vm endpoints ( that’s all you seem to want out of external services).

Cc @louiscryan wdyt of adding label annotations to vm endpoints ? If you think you need a cross platform solution, then let’s add an explicit resource that models the intent of an in-mesh entity

_blog/2018/v1alpha3-routing.md
```diff
@@ -0,0 +1,365 @@
...
+2. One or more (different) external endpoints can be configured to implement a “virtual” external service.
+3. Secure HTTP services (automatic TLS upgrade) can now be accessed using standard https (e.g., `https://secureservice.com/`
+   instead of `http://secureservice.com:443/`.
+4. Multiple CIDR subsets can now be included in a single `ExternalService` configuration.
+
+Because an `ExternalService` configuration simply adds an external destination to the internal service registry, it can be
+used in conjunction with a `VirtualService` and/or `DestinationRule`, just like any other service in the registry.
+
+### Gateway
+
```
rshriram: This needs to move up

_blog/2018/v1alpha3-routing.md
```diff
@@ -326,24 +326,66 @@ spec:
...
+  - foo.com
+  ports:
+  - number: 80
+    name: http
+    protocol: HTTP
+```
+
+That said, `ServiceEntry` has significantly more functionality than its predecessor.
+First of all, `ServiceEntry` is not limited to external service configuration.
+It can also be used to explicitly add services as part of expanding the service mesh with unmanaged infrastructure
```
rshriram: to include unmanaged infrastructure

_blog/2018/v1alpha3-routing.md
```diff
@@ -326,24 +326,66 @@ spec:
...
+(e.g., VMs added to a Kubernetes based service mesh). Such entries are treated just like all other internal services,
```
rshriram: Kubernetes-based

_blog/2018/v1alpha3-routing.md
```diff
@@ -326,24 +326,66 @@ spec:
...
+      caCertificates: /etc/certs/rootcacerts.pem
+```
+
+In addition to its expanded generality, `ServiceEntry` includes several other improvements over `EgressRule`
+including the following:
+
+1. The resolution mode for a service is now configurable (`PASSTHROUGH`, `STATIC`, or `DNS`).
 2. One or more (different) external endpoints can be configured to implement a “virtual” external service.
 3. Secure HTTP services (automatic TLS upgrade) can now be accessed using standard https (e.g., `https://secureservice.com/`
    instead of `http://secureservice.com:443/`.
```
rshriram: @ZackButcher this is implemented right?

_blog/2018/v1alpha3-routing.md
```diff
@@ -326,24 +326,66 @@ spec:
...
 4. Multiple CIDR subsets can now be included in a single `ExternalService` configuration.
```
rshriram: stray line? Also, this does not compute in my brain, as I am not fully familiar with pains of egress rules.

_blog/2018/v1alpha3-routing.md
```diff
@@ -326,24 +326,66 @@ spec:
...
 
-Because an `ExternalService` configuration simply adds an external destination to the internal service registry, it can be
-used in conjunction with a `VirtualService` and/or `DestinationRule`, just like any other service in the registry.
-
```
rshriram: I wonder if we can completely get rid of this "In addition to its expanded generality, `ServiceEntry` includes several other improvements over `EgressRule` including the following", and use the space to showcase an example of integrating a VM into the registry. These improvements are kind of hard to empathize with, given that egressrules did not see that much use compared to route rules/ingress, and these points are talking about finer details that a casual reader may not be familiar with.
Thoughts?

frankbu: I think that might be too much detail for this blog, which is just meant to highlight the changes. Maybe that would better suited to the new v1alpha3 version of the traffic managment concept doc (which is still TBD): https://istio.io/docs/concepts/traffic-management/rules-configuration.html#egress-rules.

I condensed the list and clarified what the additional improvements are. PTAL

PR #1191, opened 2018-04-19
v1alpha3 — "greatly improved" caution message

ermik:
Clarified the warning immediately following the "Recommended" approach
- added information about what "WIP" means, i.e. it doesn't work at all in 0.7.1. 
- added a #header link to help people skip to the Kubernetes Ingress section (needs to be verified on a live template)

I feel like it wasn't just me who jumped on the opportunity to test drive the v1alpha3 API, and what is more confusing I have succeeded — `VirtualService` is a bit quirky but it works and the lack of clarity in the API reference has not deterred my love for decoupling k8s and Istio. But I have glanced over the note @geeknoid left in the docs, taking the success of one Istio extension as a green light to use the remaining ones — `DestinationRule` and `Gateway`. It was only when I tried to use a good'ol `Ingress` resource instead that I found everything actually works in my toy cluster. Moreover, this experimentation led me to [issues#294](https://github.com/istio/issues/issues/294) which was completely baffling until just a couple of hours ago that @ZackButcher has said the magic words: "it's not there yet."

This PR is meant to patch the main portion of the website for a remaining couple of weeks before 0.8 drops — clearly deterring usage of the practically non-existent `Gateway`. Please consider merging it regardless of future plans.

### DIFF
--- a/_docs/tasks/traffic-management-v1alpha3/ingress.md
+++ b/_docs/tasks/traffic-management-v1alpha3/ingress.md
@@ -54,9 +54,9 @@ This task describes how to configure Istio to expose a service outside of the se
    ```
 
 ## Configuring ingress using an Istio Gateway resource (recommended)
-
-> This is still a work in progress and is not yet functional.
-
+----
+> This is still a work in progress and is **not available** in current Istio release (0.7.1). Current way of enabling inbound cluster traffic via Kubernetes Ingress resource is outlined [below](#configuring-ingress-using-a-kubernetes-ingress-resource).
+----
 An [Istio Gateway]({{home}}/docs/reference/config/istio.networking.v1alpha3.html#Gateway) is the preferred
 model for configuring ingress traffic in Istio.
 An ingress `Gateway` describes a load balancer operating at the edge of the mesh receiving incoming
--- a/_docs/tasks/traffic-management-v1alpha3/ingress.md
+++ b/_docs/tasks/traffic-management-v1alpha3/ingress.md
@@ -8,6 +8,7 @@ layout: docs
 type: markdown
 redirect_from: /docs/tasks/ingress.html
 ---
+
 {% include home.html %}
 
 In a Kubernetes environment, the [Kubernetes Ingress Resource](https://kubernetes.io/docs/concepts/services-networking/ingress/)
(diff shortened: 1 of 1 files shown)


_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -54,9 +54,9 @@ This task describes how to configure Istio to expose a service outside of the se
    ```
 
 ## Configuring ingress using an Istio Gateway resource (recommended)
-
-> This is still a work in progress and is not yet functional.
-
+----
+> This is still a work in progress and is **not available** in current Istio release (0.7.1). Current way of enabling inbound cluster traffic via Kubernetes Ingress resource is outlined [below](#configuring-ingress-using-a-kubernetes-ingress-resource).
```
ZackButcher: Slight edit:

> in **the** current Istio release (0.7.1). **The** current way of enabling...

_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -47,15 +48,19 @@ This task describes how to configure Istio to expose a service outside of the se
...
+  ```bash
+  openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /tmp/tls.key -out /tmp/tls.crt -subj "/CN=foo.bar.com"
+  ```
 
 ## Configuring ingress using an Istio Gateway resource (recommended)
 
-> This is still a work in progress and is not yet functional.
+---
+
+> This is still a work in progress and is **not available** in the most recent Istio release (0.7.1). The current way of enabling inbound cluster traffic via Kubernetes Ingress resource is outlined [below](#configuring-ingress-using-a-kubernetes-ingress-resource).
```
geeknoid: If you push this doc change in, it will only show up as part of the 0.8 docs. So is this warning still relevant in that context? 0.8 should be published next week.

We could apply this warning to the release-0.7 branch and push that out immediately to istio.io as well, if you'd like to submit a PR for that.

_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -47,15 +48,19 @@ This task describes how to configure Istio to expose a service outside of the se
 
 * Generate a certificate and key that will be used to demonstrate a TLS-secured gateway
 
-   A private key and certificate can be created for testing using [OpenSSL](https://www.openssl.org/).
```
geeknoid: Please undo all the spacing/indent changes in this PR. This is unnecessary and is causing the presubmit test to fail.

ermik: Holy hell, VS Code or something. Apologies.

_docs/tasks/traffic-management-v1alpha3/ingress.md
ermik: Then I'm gonna close this and push to 0.7 instead — then, it's up to you guys, I just think that if 0.8 lags behind due to the fact you're dotting the *i*'s then this change would be a good patch in the meantime.

So sorry I haven't noticed the drastic spacing change!

PR #1199, opened 2018-04-24
remove installation instructions for prometheus

rkpagadala:
prometheus is installed by default, so the instructions are not necessary and are wrong

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ To run the site locally with Docker, use the following command from the top leve
 
 ```bash
 # First time: (slow)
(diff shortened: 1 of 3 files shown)


_docs/tasks/telemetry/querying-metrics.md
```diff
@@ -24,15 +23,7 @@ application.
...
-1. To query the metrics provided by Mixer, first install the Prometheus add-on.
-
-   In Kubernetes environments, execute the following command:
-
-   ```bash
-   kubectl apply -f install/kubernetes/addons/prometheus.yaml
-   ```
-
-1. Verify that the service is running in your cluster.
+1. Verify that the prometheus service is running in your cluster.
```
ldemailly: add 
Verify that the prometheus service is running in your cluster (since 0.8 the prometheus setup is included in istio.yaml and istio-auth.yaml by default)

rkpagadala:
@ldemailly PTAL

PR #1201, opened 2018-04-25
Fix docker run command

deeglaze:
The command as it stands will fail with "Gemfile not found". The working directory should be set to $(pwd) as well to start execution in the istio.github.io directory and find the Gemfile.


deeglaze:
:( neither of these commands work for me actually. Without -w $(pwd) it
says "Gemfile not file", with it, it says Rakefile not found. Without rake
test, it sits there for a second and then quits quietly.

PR #1204, opened 2018-04-25
Minor format fix for istio.rbac.v1alpha1.html.

yangminzhu:
- Changed to use list instead of plain text '*' symbol 
- Changed to use hyperlink instead of plain text link


geeknoid:
This file is generated from the proto. Any manual edits will be lost the next time we scrape the protos.

So I'm closing this. Please go apply related changes to the proto file in istio/api

yangminzhu:
Oops, didn't realized it's auto generated, thanks for the remind.

PR #1206, opened 2018-04-26
Fix command to build & serve site locally using docker (bad workdir)

zbbfufu:
Workdir option in `docker run ... jekyll` is wrong. Workdir should be `/srv/jekyll`, not `$(pwd)`


zbbfufu:
I signed it!
/assign @geeknoid

PR #1211, opened 2018-04-27
fix manual sidecar injection docs for helm template changes

ayj:
Addresses https://github.com/istio/istio.github.io/issues/1210


ayj:
cc @sdake @tiswanso @yusuoh 

The sidecar injector is not enabled by default in any of the install YAML files generated by `make generate_yaml`. There isn't a non-templated version of the injection configmap available. This PR uses istioctl's internal template to workaround the problem.

PR #1212, opened 2018-04-28
clarify we support more than just k8s

linsun:


_docs/concepts/what-is-istio/overview.md
```diff
@@ -11,7 +11,7 @@ type: markdown
 
 This document introduces Istio: an open platform to connect, manage, and secure microservices. Istio provides an easy way to create a network of deployed services with load balancing, service-to-service authentication, monitoring, and more, without requiring any changes in service code. You add Istio support to services by deploying a special sidecar proxy throughout your environment that intercepts all network communication between microservices, configured and managed using Istio's control plane functionality.
 
-Istio currently only supports service deployment on Kubernetes, though other environments will be supported in future versions.
+Istio currently supports service deployment on Kubernetes, as well as services registered with Consul or Eureka and services running on individual VMs.
```
gyliu513: How about adding some links for deployment with `Consul`, `Eureka` and `individual VMs`?

rshriram:
I would probably state that we support k8s and CloudFoundry. I don't see any participation from Consul people and no one is using Eureka. Also, Consul is thoroughly untested (even for release). Someone or other keeps breaking it every release by hardcoding kubernetes specific stuff.

linsun:
@rshriram fair point.  I saw Martin already merged the PR.  We need to decide if we should remove consul or eureka off our docs in 0.8.  I don't see any cloud foundry support in our docs, thus I didn't mention it.

PR #1214, opened 2018-04-29
Needless redirect causes timeout on linkchecking

sdake:


sdake:
this is an incorrect solution.

PR #1218, opened 2018-04-30
adding the recommended namespace

linsun:
https://github.com/istio/issues/issues/312


frankbu:
The `helm template` instructions require the istio-system ns to be created as well. Please replace line line 51 with the following:
```
   kubectl create ns istio-system
   kubectl create -f $HOME/istio.yaml
```

linsun:
Thanks @frankbu good catch.  PTAL

PR #1224, opened 2018-04-30
Introduce support for new fangled PRE blocks.

geeknoid:
Instead of having to have two PRE blocks, one for commands and one for the output,
we can now have a single PRE block and we take care of rendering things to show the
command vs. the output. The Copy button on such a thing only copy the command, and not
the output.

We now also show a $ on command-lines, but the Copy button doesn't copy that and knows to just
copy the usable part of the command-line.

For now, it's only enabled on a single page in the site: [https://geeknoid.github.io/istio.github.io/docs/setup/kubernetes/sidecar-injection.html](https://geeknoid.github.io/istio.github.io/docs/setup/kubernetes/sidecar-injection.html). If this looks good, I'll do a pass through the entire site and adjust all the PRE blocks accordingly.


rshriram:
This looks cool.

PR #1228, opened 2018-05-01
Fix doc

wattli:


_docs/tasks/security/per-service-mtls.md
```diff
@@ -9,7 +9,7 @@ type: markdown
 ---
 {% include home.html %}
 
-> If you are using Istio 0.7 or later, please refer to [authentication policy task]({{home}}/docs/tasks/security/authn-policy.html) for alternative (recommended) solution using authentication policy.
```
geeknoid: Can we just remove the page for 0.8? People using 0.6 or earlier can go to archive.istio.io to find the version-appropriate docs.

wattli: I am afraid this will cause confusion to people. They will wonder whether this is a mistake on website or it's not supported any more. What about keeping this page for a while and mentioning this will be deprecated, and then delete this page? I feel this will give people (who care about this feature) some time to switch.

geeknoid: All righty....

_docs/tasks/security/per-service-mtls.md
```diff
@@ -9,7 +9,7 @@ type: markdown
...
+> This page will be deprecated soon, if you are using Istio 0.7 or later, please refer to [authentication policy task]({{home}}/docs/tasks/security/authn-policy.html) for alternative (recommended) solution using authentication policy.
```
geeknoid: Should we say the feature will be deprecated, instead of talking about the page? e.g

This feature will soon be deprecated. If you are using Istio 0.7 or later, please refer to [authentication policy]({{home}}/docs/tasks/security/authn-policy.html) for the recommended approach moving forward.

wattli: fixed, thanks!

PR #1234, opened 2018-05-03
Added namespace when create helm template.

gyliu513:
/cc @sdake


gyliu513:
@geeknoid @linsun can we have this merged, it is really not convenient that I have to copy the command from document and update the command again when doing the test.

sdake:
@gyliu513 this should go into release 0.8.  Can you submit a PR against that branch?

gyliu513:
@sdake I think we do not have 0.8 branch cut yet for document, seems we still need merge this to master?

@linsun @geeknoid please help confirm, thanks.

PR #1242, opened 2018-05-06
Switch most uses of ```bash to ```command.

geeknoid:
This takes advantage of the new rendering for command-lines and their outputs.


geeknoid:
Ignoring the bad link reported by CircleCi since it seems to be an outage/broken site on kubernetes.io.

PR #1252, opened 2018-05-08
Update gateway docs to account for environments with loadbalancers

cmluciano:


_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -143,14 +143,28 @@ In our case, we have set the selector value to `istio: ingressgateway` to use th
...
+   ```command
+   $ export INGRESS_HOST=130.211.10.121
+   ```
+
+   * If load balancers are not supported, use the ingressgateway controller pod's hostIP:
 
    ```command
    $ kubectl -n istio-system get po -l istio=ingressgateway -o jsonpath='{.items[0].status.hostIP}'
    169.47.243.100
    ```
```
linsun: This instruction doesn't work for IBM Cloud container service, could you add steps for ICS?

cmluciano: Strange, I was able to get this working using the LB steps in IKS. Is their some piece that is failing in your cluster?

frankbu: I think the problem is IKS free tier. There a special instructions for it in the bookinfo guide. I guess we need to repeat all those same instructions in here too?
https://github.com/istio/istio.github.io/blob/master/_docs/guides/bookinfo.md#determining-the-ingress-ip-and-port

linsun: Yes, the host ip needs to be obtained from `bx cs workers {cluster-name}` for node port.  I'm open to suggestion on the best way to refer to it.

linsun:
@cmluciano can you discuss with @frankbu on the best way to reflect instructions for node port on ICCS?

_docs/tasks/traffic-management-v1alpha3/ingress.md
```diff
@@ -143,21 +143,43 @@ In our case, we have set the selector value to `istio: ingressgateway` to use th
...
    ```command
-   $ kubectl -n istio-system get svc istio-ingressgateway
-   NAME                   CLUSTER-IP     EXTERNAL-IP   PORT(S)                      AGE
-   istio-ingressgateway   10.10.10.155   <pending>     80:31486/TCP,443:32254/TCP   32m
+   $ export INGRESS_HOST=130.211.10.121
    ```
 
+   * If load balancers are not supported, use the ingressgateway controller pod's hostIP:
+
+        *  _GKE:_
```
ZackButcher: You don't need this: if you just deploy the ingressgateway service it will be assigned a public static IP on GKE (we ship config to make that service a LoadBalancer service in k8s by default).

linsun:
@cmluciano could you fix the circleci errors?

cmluciano:
Closing in favor of https://github.com/istio/istio.github.io/pull/1341

PR #1254, opened 2018-05-08
update samples to align with latest destination rule proto definition

quanjielin:


andraxylia:
/assign @geeknoid
 /assign @linsun 

I think you need to add more people in the OWNERS.

quanjielin:
The test failure should not be caused by this PR.

PR #1255, opened 2018-05-09
indentation issue in example

ipedrazas:
There's a little mistake with the indentation in the `HTTPRedirect`. The `redirect:` entry should be at the same level as `- uri:`


frankbu:
Thanks for spotting the error, but this doc is generated from comments in the proto files in istio/api, so it needs to be changed there. https://github.com/istio/api/blob/28cb7ff671111dbcc1f4ba86046d8dc46d437400/networking/v1alpha3/virtual_service.proto#L651

frankbu:
Fixed as part of https://github.com/istio/api/pull/474

PR #1268, opened 2018-05-11
Add a release cadence page.

geeknoid:
Staging site: https://geeknoid.github.io/istio.github.io/about/release-cadence.html

### DIFF
--- a/_about/release-cadence.md
+++ b/_about/release-cadence.md
@@ -53,6 +53,5 @@ next LTS.
 In the unlikely event we need to issue a patch to a snapshot, it is numbered as:
 
 ```plain
-<major>.<minor>.<LTS patch level>-snapshot.<snaoshot count>.<snapshot patch level>
+<major>.<minor>.<LTS patch level>-snapshot.<snapshot count>.<snapshot patch level>
 ```
-S
--- a/_about/release-cadence.md
+++ b/_about/release-cadence.md
@@ -15,16 +15,16 @@ The different types (Daily, Snapshot, LTS) represent different product quality l
 from the Istio team. In this context, *support* means that we will produce patch releases for critical issues and
 offer technical assistance. Separately, 3rd parties and partners may offer longer-term support solutions.
 
-|Type             | Support Level                                         | Quality and Recommended Use
-|-----------------+-------------------------------------------------------+----------------------------
-|Daily Build      | No support                                            | Dangerous, may not be fully reliable. Useful to experiment with.
+|Type             | Support Level                                            | Quality and Recommended Use
+|-----------------+----------------------------------------------------------+----------------------------
+|Daily Build      | No support                                               | Dangerous, may not be fully reliable. Useful to experiment with.
 |Snapshot Release | Support is provided for the latest snapshot release only | Expected to be quite stable, but use in production should be limited to an as-needed basis. Usually only adopted by bleeding edge users, or users seeking specific features.
-|LTS Release      | Support is provided until 3 months after the next LTS | Safe to deploy in production. Users are encouraged to upgrade to these releases as soon as possible.
-|Patches          | Same as corresponding Snapshot/LTS release            | Users are encouraged to adopt patch releases as soon as they are available for a given release.
+|LTS Release      | Support is provided until 3 months after the next LTS    | Safe to deploy in production. Users are encouraged to upgrade to these releases as soon as possible.
+|Patches          | Same as the corresponding Snapshot/LTS release           | Users are encouraged to adopt patch releases as soon as they are available for a given release.
 
 You can find available releases on the [releases page](https://github.com/istio/istio/releases),
 and if you're the adventurous type, you can get learn about our daily build on the [daily builds wiki](https://github.com/istio/istio/wiki/Daily-builds).
-You can find high-level releases notes for each LTS release [here]({{nome}}/about/notes).
+You can find high-level releases notes for each LTS release [here]({{home}}/about/notes).
 
 ## Naming Scheme
 
@@ -43,7 +43,7 @@ current LTS release.
 For snapshot releases, our numbering scheme is:
 
 ```plain
-<major>.<minor>.<LTS patch level>-snapshot.<snapshot count>
+<major>.<minor>.0-snapshot.<snapshot count>
 ```
 
 where `major`, `minor` and `LTS patch level` represent the next LTS, and
@@ -53,5 +53,5 @@ next LTS.
 In the unlikely event we need to issue a patch to a snapshot, it is numbered as:
 
 ```plain
-<major>.<minor>.<LTS patch level>-snapshot.<snapshot count>.<snapshot patch level>
+<major>.<minor>.0-snapshot.<snapshot count>.<snapshot patch level>
 ```
--- a/_about/release-cadence.md
+++ b/_about/release-cadence.md
@@ -5,24 +5,25 @@ weight: 5
 ---
 
 We produce new builds of Istio on a daily basis. Around once a month or so, we take one of these daily
-builds and run it through a number of additional qualification tests and tag the build as an Alpha release.
-Around once a quarter or so, we take one of these Alpha releases, run through a bunch more tests
-and tag the build as a Long Term Support (LTS) release. Finally, if we find something wrong with an Alpha
-or LTS release, we issue patches.
+builds and run it through a number of additional qualification tests and tag the build as a Snapshot release.
+Around once a quarter or so, we take one of these Snapshot releases, run through a bunch more tests
+and tag the build as a Long Term Support (LTS) release. Finally, if we find something wrong with an
+LTS release, we issue patches.
 
-The different types (Daily, Alpha, LTS) represent different product quality levels and different levels of support
+The different types (Daily, Snapshot, LTS) represent different product quality levels and different levels of support
 from the Istio team. In this context, *support* means that we will produce patch releases for critical issues and
 offer technical assistance. Separately, 3rd parties and partners may offer longer-term support solutions.
 
-|Type          | Support Level                                         | Quality and Recommended Use
-|--------------+-------------------------------------------------------+----------------------------
-|Daily Build   | No support                                            | Dangerous, may not be fully reliable. Useful to experiment with.
-|Alpha Release | Support is provided for the latest Alpha release only | Expected to be quite stable, but use in production should be limited to an as-needed basis. Usually only adopted by bleeding edge users, or users seeking specific features.
-|LTS Release   | Support is provided until 3 months after the next LTS | Safe to deploy in production. Users are encouraged to upgrade to these releases as soon as possible.
-|Patches       | Same as corresponding Alpha/LTS release               | Users are encouraged to adopt patch releases as soon as they are available for a given release
+|Type             | Support Level                                         | Quality and Recommended Use
+|-----------------+-------------------------------------------------------+----------------------------
+|Daily Build      | No support                                            | Dangerous, may not be fully reliable. Useful to experiment with.
+|Snapshot Release | Support is provided for the latest Alpha release only | Expected to be quite stable, but use in production should be limited to an as-needed basis. Usually only adopted by bleeding edge users, or users seeking specific features.
+|LTS Release      | Support is provided until 3 months after the next LTS | Safe to deploy in production. Users are encouraged to upgrade to these releases as soon as possible.
+|Patches          | Same as corresponding Snaoshot/LTS release            | Users are encouraged to adopt patch releases as soon as they are available for a given release.
 
 You can find available releases on the [releases page](https://github.com/istio/istio/releases),
 and if you're the adventurous type, you can get learn about our daily build on the [daily builds wiki](https://github.com/istio/istio/wiki/Daily-builds).
+You can find high-level releases notes for each LTS release [here]({{nome}}/about/notes).
 
 ## Numbering Scheme
 
@@ -38,12 +39,18 @@ Our naming scheme for LTS releases is:
 where `minor` is increased for every LTS release, and `patch level` counts the number of patches for the
 current LTS release.
 
-For Alpha releases, our numbering scheme is:
+For Snapshot releases, our numbering scheme is:
 
 ```plain
-<major>.<minor>.<patch level>-alpha.<alpha count>
+<major>.<minor>.<patch level>-snapshot.<snaoshot count>
 ```
 
 where `major` and `minor` represent the number of the next LTS release, `patch level` counts the number of
-patches for the Alpha release, and `alpha count` starts at 0 and increases for every Alpha release leading up
+patches for the Snapshot release, and `snapshot count` starts at 0 and increases for every Snapshot release leading up
 to the next LTS.
+
+In the unlikely event we need to issue a patch to a snapshot release, it is numbered as:
+
+```plain
+<major>.<minor>.<patch level>-snapshot.<snaoshot count>.<snapshot patch level>
+```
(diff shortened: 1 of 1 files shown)


_about/release-cadence.md
```diff
@@ -0,0 +1,28 @@
...
+title: Build & Release Cadence
+description: How we manage, number, and support Istio releases.
+weight: 5
+---
+
+We produce new builds of Istio on a daily basis.
+Once a week, one of these daily build undergoes additional
+qualification and gets anointed the weekly release. Once a month, one of the weekly releases undergo
+additional qualifications and gets anointed as the monthly release. Once a quarter or so, one of the monthly
+releases undergoes additional qualifications and gets anointed as a long term supported (LTS) release.
```
linsun: I'm not familiar with the additional qualification for weekly release.  Is this doc-ed anywhere?

_about/release-cadence.md
```diff
@@ -0,0 +1,28 @@
...
+
+Each release type implies different levels of quality and support:
+
+|Release Type | Support Level | Typical Use
+|-------------+---------------+------------
+|Daily   | No support           | Dangerous, may not be fully reliable. Useful to experiment with.
+|Weekly  | No support           | Dangerous, may not be fully reliable. Useful to experiment with.
+|Monthly | Support is provided for the current monthly release only | Safe to deploy in production. Usually only adopted by bleeding edge users, or users seeking specific features.
```
linsun: Do we really want to say monthly is safe to deploy in prod?

geeknoid: That came from this doc: https://docs.google.com/document/d/1scC_f7nObwBut_xxE6lUfvcHBybmttUsJDMwXD3gcO4/edit#heading=h.ob68p14f45rs

Let's talk about it during the TOC meeting.

_about/release-cadence.md
geeknoid: That came from https://github.com/istio/istio/blob/master/release/README.md and from https://docs.google.com/document/d/1_SNEDOSGyw69UnLC9kYY7bTfxVszTE-W-mLdtPVJuOA/edit

hklai: Weekly release has not been implemented.
And I will send a PR to update the ETA in https://github.com/istio/istio/blob/master/release/README.md

_about/release-cadence.md
```diff
@@ -0,0 +1,28 @@
...
+|LTS     | Support is provided until 3 months after the next LTS                | Safe to deploy in production. Users are encouraged to upgrade to these releases as soon as possible.
+
+In this context, *support* means that we will produce patch releases for critical issues and offer technical assistance.
+In addition, longer-term support may be available from 3rd parties and partners.
+
+Prior to Istio 0.8, we increased the product's version number for every monthly release. Effective with 0.8,
+we will increase the product's version number only for LTS releases.
+
+You can find all available releases on our [releases page](https://github.com/istio/istio/releases).
```
hklai: daily releases can be found in https://gcsweb.istio.io/gcs/istio-prerelease/daily-build/, not on the github release page.

_about/release-cadence.md
hklai: Weekly release pipeline has not been implemented.

I will send a PR to update the ETA in https://github.com/istio/istio/blob/master/release/README.md.

_about/release-cadence.md
ldemailly: or with 1 level of indirection so this page doesn't need to be updated when something changes, I would link to https://github.com/istio/istio/wiki/Daily-builds instead which has additional information

geeknoid:
Updated following the TOC meeting. PTAL.

Thanks.

_about/release-cadence.md
```diff
@@ -0,0 +1,49 @@
...
+title: Build & Release Cadence
+description: How we manage, number, and support Istio releases.
+weight: 5
+---
+
+We produce new builds of Istio on a daily basis. Around once a month or so, we take one of these daily
+builds and run it through a number of additional qualification tests and tag the build as an Alpha release.
+Around once a quarter or so, we take one of these Alpha releases, run through a bunch more tests
+and tag the build as a Long Term Support (LTS) release. Finally, if we find something wrong with an Alpha
+or LTS release, we issue patches.
```
ldemailly: historically we have not patched the monthly release after the first qualified one. not sure if we want to commit to fixing critical issues in monthlies vs ask people to wait for the next one

_about/release-cadence.md
```diff
@@ -0,0 +1,49 @@
...
+Prior to Istio 0.8, we increased the product's version number on a monthly basis. Effective with 0.8,
+we will increase the product's version number only for LTS releases.
+
+Our naming scheme for LTS releases is:
+
+```plain
+<major>.<minor>.<patch level>
+```
+
+where `minor` is increased for every LTS release, and `patch level` counts the number of patches for the
```
ldemailly: numbe of patch releases

(there will be more than 1 bug fix per patch release)

_about/release-cadence.md
```diff
@@ -0,0 +1,49 @@
...
+current LTS release.
+
+For Alpha releases, our numbering scheme is:
+
+```plain
+<major>.<minor>.<patch level>-alpha<alpha count>
+```
+
+where `minor` represents the number of the next LTS release, `patch level` counts the number of
+patches for the Alpha release, and `alpha count` increases by one for every Alpha release leading up
```
ldemailly: that doesn’t follow sem ver
(or forces to make the first lts not be .0 )

_about/release-cadence.md
```diff
@@ -0,0 +1,49 @@
...
+or LTS release, we issue patches.
+
+The different types (Daily, Alpha, LTS) represent different product quality levels and different levels of support
+from the Istio team. In this context, *support* means that we will produce patch releases for critical issues and
+offer technical assistance. Separately, 3rd parties and partners may offer longer-term support solutions.
+
+|Type          | Support Level                                         | Quality and Recommended Use
+|--------------+-------------------------------------------------------+----------------------------
+|Daily Build   | No support                                            | Dangerous, may not be fully reliable. Useful to experiment with.
+|Alpha Release | Support is provided for the latest Alpha release only | Expected to be quite stable, but use in production should be limited to an as-needed basis. Usually only adopted by bleeding edge users, or users seeking specific features.
```
hklai: Are we going to produce release notes for alpha releases going forward?

_about/release-cadence.md
```diff
@@ -0,0 +1,49 @@
...
+where `minor` is increased for every LTS release, and `patch level` counts the number of patches for the
+current LTS release.
+
+For Alpha releases, our numbering scheme is:
+
+```plain
+<major>.<minor>.<patch level>-alpha<alpha count>
+```
+
+where `minor` represents the number of the next LTS release, `patch level` counts the number of
```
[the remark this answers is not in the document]
ldemailly: and patch (again you can't have 0.9.3-alpha2 be before 0.9.0

also is the first one 0.9.0-alpha0 or alpha1 ?

what is 0.9.0-alpha2 vs 0.9.2-alpha1 for instance ?

_about/release-cadence.md
geeknoid: I don't think so, only full notes for LTS releases

_about/release-cadence.md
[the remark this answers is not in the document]
geeknoid: How does this model conflict with what's defined in https://semver.org/?

First Alpha release for 0.9 would be 0.9.0-alpha.0

0.9.0-alpha.2 means "the third alpha release for 0.9.0"
0.9.2-alpha.1 means "the second alpha release for 0.9.2"

So as per [this](https://semver.org/#spec-item-9), the stuff to the right of the - indicates these are pre-release versions.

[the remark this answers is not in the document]
hklai: Say we release the following over the course of few months. They are all cut at head in chronological order.

0.9.0-alpha.0, 0.9.0-alpha.1, 0.9.0-alpha.2, and 0.9.0.

When we patch 0.9.0, we create 0.9.1. *If* we apply the same patch to the alpha releases, the patched 0.9.0-alpha.2 will become 0.9.1-alpha.2, right? Strictly speaking, 0.9.1-alpha.2 is not really the third alpha release of 0.9.1.

I think this scheme works best if we only patch LTS releases but not alpha releases, so we won't end up in a confusing situation like above.

geeknoid:
OK, that's right. Patching of alpha releases is a different beast than
patching of LTS releases. We can't overload the number like that. So
following semver, we would need to add another number:

<major>.<minor>.<patch level>-alpha.<alpha count>.<alpha patch level>

_about/release-cadence.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Build & Release Cadence
+description: How we manage, number, and support Istio releases.
+weight: 5
+---
+
+We produce new builds of Istio on a daily basis. Around once a month or so, we take one of these daily
+builds and run it through a number of additional qualification tests and tag the build as an Alpha release.
```
smawson: I feel weird calling these alpha releases, since alpha has a specific meaning in terms of features, and this is unrelated to that. Can we just call them monthlies?  Prerelease? Anything other than alpha/beta/stable would be useful here.

_about/release-cadence.md
smawson: I'd almost say we don't bother mentioning patching of the monthlies. Over time we don't want to spend time doing that, those patches should just roll into a new monthly RC.

_about/release-cadence.md
smawson: We should at a minimum point at an automated set of changes between monthlies, but not do a curated set of notes.

_about/release-cadence.md
geeknoid: Oh, yeah, the use of alpha here will be confusing with what we use for individual features. I'll change to 'prerelease'.

ldemailly: standard is "rc" for the short form in the semver suffix

_about/release-cadence.md
```diff
@@ -0,0 +1,56 @@
...
+builds and run it through a number of additional qualification tests and tag the build as a Snapshot release.
+Around once a quarter or so, we take one of these Snapshot releases, run through a bunch more tests
+and tag the build as a Long Term Support (LTS) release. Finally, if we find something wrong with an
+LTS release, we issue patches.
+
+The different types (Daily, Snapshot, LTS) represent different product quality levels and different levels of support
+from the Istio team. In this context, *support* means that we will produce patch releases for critical issues and
+offer technical assistance. Separately, 3rd parties and partners may offer longer-term support solutions.
+
+|Type             | Support Level                                         | Quality and Recommended Use
```
ldemailly: table doesn't render on github md's renderer: https://github.com/geeknoid/istio.github.io/blob/64a98d66f5ccb53f698ac72d9be86ac28cab3592/_about/release-cadence.md

_about/release-cadence.md
```diff
@@ -0,0 +1,56 @@
...
+<major>.<minor>.<patch level>
+```
+
+where `minor` is increased for every LTS release, and `patch level` counts the number of patches for the
+current LTS release.
+
+For Snapshot releases, our numbering scheme is:
+
+```plain
+<major>.<minor>.<patch level>-snapshot.<snaoshot count>
```
ldemailly: snaoshot-> snapshot

_about/release-cadence.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+where `major` and `minor` represent the number of the next LTS release, `patch level` counts the number of
+patches for the Snapshot release, and `snapshot count` starts at 0 and increases for every Snapshot release leading up
+to the next LTS.
```
ldemailly: this is confusing... what is patchlevel vs snapshot count - when do each change/seems backward

it seems to read like we would do 1.0.0-snapshot.0 followed by 1.0.1-snapshot.0 if making a patch, followed by 1.0.0-snapshot.1 for the next month... which is backward ?

ldemailly: it should all be 1.0.0-rc1.0 then 1.0.0.rc1.1 then 1.0.0.rc2.0 etc... the first 3 digits are frozen to "next lts" by definition

geeknoid:
The model follows quite closely with the semver approach,

* major.minor.patch represents the labeling for the next LTS release.

* The -snapshot-count suffix reflects monthly-ish releases we cut on the
way to an LTS.

* Snapshot patch level is about us issuing an (unlikely) patch to an
existing snapshot. More likely, we would probably not release a patch and
just say "wait til the next snapshot to get this fix". But you never know...

_about/release-cadence.md
geeknoid: See https://geeknoid.github.io/istio.github.io/about/release-cadence.html

_about/release-cadence.md
```diff
@@ -33,24 +34,25 @@ we will increase the product's version number only for LTS releases.
...
 
 ```plain
-<major>.<minor>.<patch level>-snapshot.<snaoshot count>
+<major>.<minor>.<LTS patch level>-snapshot.<snapshot count>
 ```
 
-where `major` and `minor` represent the number of the next LTS release, `patch level` counts the number of
-patches for the Snapshot release, and `snapshot count` starts at 0 and increases for every Snapshot release leading up
-to the next LTS.
+where `major` and `minor` and `LTS patch level` represent the next LTS, and
```
ldemailly: that's clearer thank you, but you might as well replace <LTS patch level> by 0 as before an LTS release happens the patch level will always be 0
(though we may do -rc for patch levels I suppose but those wouldn't be monthlies anyway)

I would recommend to replace "snapshot" with "rc" in the string for both familiarity and brievety

_about/release-cadence.md
```diff
@@ -33,24 +34,25 @@ we will increase the product's version number only for LTS releases.
...
+`snapshot count` starts at 0 and increases for every Snapshot leading up to the
+next LTS.
 
-In the unlikely event we need to issue a patch to a snapshot release, it is numbered as:
+In the unlikely event we need to issue a patch to a snapshot, it is numbered as:
 
 ```plain
-<major>.<minor>.<patch level>-snapshot.<snaoshot count>.<snapshot patch level>
+<major>.<minor>.<LTS patch level>-snapshot.<snaoshot count>.<snapshot patch level>
```
ldemailly: snaoshot -> snapshot here too (isn't the build supposed to find typos?)

_about/release-cadence.md
```diff
@@ -0,0 +1,57 @@
...
+<major>.<minor>.<LTS patch level>
+```
+
+where `minor` is increased for every LTS release, and `LTS patch level` counts the number of patches for the
+current LTS release.
+
+For snapshot releases, our numbering scheme is:
+
+```plain
+<major>.<minor>.<LTS patch level>-snapshot.<snapshot count>
```
ldemailly: can you replace _LTS patch level_ by 0 for clarity here and the below example ?

ldemailly:
lgtm thx!

_about/release-cadence.md
```diff
@@ -0,0 +1,57 @@
...
+Prior to Istio 0.8, we increased the product's version number on a monthly basis. Effective with 0.8,
+we will increase the product's version number only for LTS releases.
+
+Our naming scheme for LTS releases is:
+
+```plain
+<major>.<minor>.<LTS patch level>
+```
+
+where `minor` is increased for every LTS release, and `LTS patch level` counts the number of patches for the
```
hklai: It may be a good idea to mention that a patch is always created on top of the corresponding LTS release, and not from head.

_about/release-cadence.md
```diff
@@ -0,0 +1,57 @@
...
+current LTS release.
+
+For snapshot releases, our numbering scheme is:
+
+```plain
+<major>.<minor>.0-snapshot.<snapshot count>
+```
+
+where `major`, `minor` and `LTS patch level` represent the next LTS, and
```
hklai: Remove `LTS patch level`

hklai: Also, I wonder if we should use a different character between "snapshot" and <snapshot count>? Say '_' instead?

geeknoid: The . is what the semver spec says to use...

geeknoid:
All righty. hopefully final call for comments.

smawson:
Just took a look over the staging version and looks great, approved from me!

PR #1270, opened 2018-05-11
clarify the function of pilot agent

linsun:


linsun:
@frankbu is this file auto generated.... realized that after creating the PR :(  If so, I'll update the code first.

frankbu:
@linsun yes, this is generated from the code.

linsun:
Thanks @frankbu

PR #1285, opened 2018-05-16
Updated mc document for secret register.

gyliu513:
https://github.com/istio/istio/issues/5534 has been fixed and we can add secrets dynamically now.

/cc @sdake @sbezverk

### DIFF
--- a/_docs/setup/kubernetes/multicluster-install.md
+++ b/_docs/setup/kubernetes/multicluster-install.md
@@ -90,10 +90,8 @@ Create a namespace for instantiating the secrets:
 $ kubectl create ns istio-system
 ```
 
-> Ordering currently matters.  Secrets must be created prior to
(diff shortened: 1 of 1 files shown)


_docs/setup/kubernetes/multicluster-install.md
```diff
@@ -90,10 +90,8 @@ Create a namespace for instantiating the secrets:
 $ kubectl create ns istio-system
 ```
 
-> Ordering currently matters.  Secrets must be created prior to
-the deployment of the Istio control plane.  Creating secrets
-after Istio is started will not register the secrets with Istio
-properly.
+> You can create the secrets either before or after deloy the istio control
```
geeknoid: deloy -> deploying
istio -> Istio

gyliu513: Done

PR #1297, opened 2018-05-17
Remove old routing api version of tasks

frankbu:
I added a note at the top of every task like this:
> Note: This task uses the new [v1alpha3 traffic management api]({{home}}/blog/2018/v1alpha3-routing.html). The old api has been deprecated and will be removed in the next Istio release. If you need to use the old version, follow the docs [here](https://archive.istio.io/v0.6/docs/tasks/).

It should actually point to archived 0.7 docs, but since they don't exist yet I needed to use 0.6 (for now) to satisfy the HTML-Proofer


_docs/tasks/traffic-management/ingress.md
```diff
@@ -1,30 +1,27 @@
...
 
-This task describes how to configure Istio to expose a service outside of the service mesh cluster.
-In a Kubernetes environment, the [Kubernetes Ingress Resource](https://kubernetes.io/docs/concepts/services-networking/ingress/)
-allows users to specify services that should be exposed outside the
-cluster. It allows one to define a backend service per virtual host and path.
-
-Once the Istio Ingress specification is defined, traffic entering the cluster is directed through the `istio-ingress` service. As a result, Istio features, for example, monitoring and route rules, can be applied to the traffic entering the cluster.
-
-The Istio Ingress specification is based on the standard [Kubernetes Ingress Resource](https://kubernetes.io/docs/concepts/services-networking/ingress/) specification, with the following differences:
+> Note: This task uses the new [v1alpha3 traffic management API]({{home}}/blog/2018/v1alpha3-routing.html). The old api has been deprecated and will be removed in the next Istio release. If you need to use the old version, follow the docs [here](https://archive.istio.io/v0.6/docs/tasks/).
```
rshriram: Should we mention the use of istioctl convert command to help transition ?

PR #1301, opened 2018-05-17
fix command in docs so it can be copied

rkpagadala:


rkpagadala:
Verify installation section copies the entire command and output which is incorrect.
https://preliminary.istio.io/docs/setup/kubernetes/quick-start-gke-dm.html#verify-installation

PR #1312, opened 2018-05-18
Major update to the Mixer docs.

geeknoid:
- Revamp the docs to remove programmer-level info and present an
operator-centric view of the design.

- Deemphasize "Mixer" in favor of Policies and Telemetry throughout. We now
focus on the "What" rather than the "How".

- Increase consistency in the terms we use to avoid confusion for the reader.


_docs/concepts/policies-and-telemetry/config.md
```diff
@@ -0,0 +1,221 @@
...
+
+```plain
+request.path: xyz/abc
+request.size: 234
+request.time: 12:34:56.789 04/17/2017
+source.ip: 192.168.0.1
+destination.service: example
+```
+
+Mixer is the Istio component that implements policy and telemetry functionality.
```
douglas-reid: Do we want to mention the two services here (istio-policy and istio-telemetry) ?

PR #1313, opened 2018-05-18
Some cleanup for kubernetes setup.

gyliu513:
1) `kubectl get svc -n istio-system` should also include new services,
such as `istio-egressgateway` etc.
2) `kubectl get pods -n istio-system` should incldue new pods.
3) Updated uninstall part.

/cc @linsun @geeknoid


gyliu513:
@linsun

1) Added ilb ip address for ingress gateway.
2) `istio-mixer-create-cr-jwtrz 0/1 Completed 0 11h ` is a job which is used to create some CRD for mixer. I think you are not using Kubernetes 1.10, if you run `kubectl get pods -a -n istio-system`, you will see the completed pod which is generated by a job. In kuberntes 1.10 which I was using, `kubect get pods` will list all pods including `Completed` as well by default.

PR #1314, opened 2018-05-18
more update to helm instruction

linsun:


_docs/setup/kubernetes/helm-install.md
```diff
@@ -22,7 +22,7 @@ There are two options for using Helm to deploy Istio.
 1. Use `helm template` to render a manifest and use `kubectl`
 to create it.
 
-1. Use Helm's Tiller service to manage the lifecycle
+1. Use Helm's [Tiller](https://github.com/kubernetes/helm#helm-in-a-handbasket) service to manage the lifecycle
```
gyliu513: how about use this link https://github.com/kubernetes/helm/blob/master/docs/architecture.md#components ?

PR #1315, opened 2018-05-18
Update helm-install.md

douglas-reid:
To match the generated `istio.yaml` that is packaged with the release, this updates the default installation via helm to include prometheus.


linsun:
sorry this is needed for RC generated on 05/15 but @costinm has fixed this so that prometheus is enabled by default which is why I removed that flag.

Tahler:
Looks like Grafana work fine fine following [its doc](https://preliminary.istio.io/docs/tasks/telemetry/using-istio-dashboard.html), but only if Prometheus is pre-installed (installing the Grafana add-on does not also install Prometheus like it says). So I think these should be changed:
- Prometheus doc should mention how to install it if it's not there.
- Grafana doc should mention Prometheus service as a prerequisite.

linsun:
ok to close this?  pls reopen if needed.

PR #1329, opened 2018-05-21
Use envoryv2 for bookinfo example.

gyliu513:
Fixed issue https://github.com/istio/istio/issues/5701

/cc @linsun @ymesika @morvencao  @geeknoid


linsun:
I think this is not needed for new install if you follow the new build that uses proxyv2 as the global image by default.

This should be only needed if users are migrating to 0.8.

linsun:
/lgtm
it doesn't hurt to add it

Issue #1299, opened 2018-05-17
Friction log when using helm to install Istio


andraxylia:
Followed this:

https://preliminary.istio.io/docs/setup/kubernetes/helm-install.html

- missing instructions on how to install helm

- Not sure what this is, followed the link but not clear what is the action

	•	"If you want to manage Istio releases with Tiller, the capability to install service accounts is required."

- very misleading, it indicates sidecar injector must be installed before anything
Prerequisites: sidecar injection 
This should be moved after the main istio.install, but I get an error when I run it:
$helm install --namespace=istio-system --set sidecar-injector.enabled=true install/kubernetes/helm/istio
Error: release loopy-stoat failed: configmaps "istio-mixer-custom-resources" already exists

- missing instruction to download the latest release before running the helm template command

- missing what is next the link to bookinfo demo with alpha3
What’s next
See the sample Bookinfo application.

sdake:
@andraxylia perhaps the docs have been updated, but the docs state:

helm install install/kubernetes/helm/istio --name istio --namespace istio-system --set sidecar-injector.enabled=true

the name of the deployment is required.  This works for with the 5-18 build.

sdake:
RE: ```If you want to manage Istio releases with Tiller, the capability to install service accounts is required.``` - the action is to enable the capability to install service accounts for this step:
```kubectl create -f install/kubernetes/helm/helm-service-account.yaml
```

sdake:
Note working from these docs: https://preliminary.istio.io/docs/setup/kubernetes/helm-install.html

sdake:
@andraxylia - the docs have undergone significant revision by the community and the implementation is significantly different from when this issue was filed.  Do these problems still persist?  If so, could clarify what problems exist with the documentation?

Cheers
-steve

sdake:
@andraxylia closing the helm friction log as I believe all items have been addressed or are recorded in other larger works (such as the SRE guide).  If you believe this was closed in error, feel free to reopen this issue or open a new issue with a new set of requests.

Cheers
-steve

PR #1338, opened 2018-05-22
Split "Quick Start" docs into their own section.

Tahler:
Should address some confusion and part of #1299.

https://docs.google.com/document/d/1ks8lpbK1lOh1HKLqcvo9QOcyeFwWcx5o7MB6A9-ycr0/ for more info.


geeknoid:
Could you provide a staging site so we can look at the rendered content? You can do this by pointing to your GItHub fork via GitHub pages.

Thanks.

Tahler:
[There](https://tahler.github.io/istio.github.io/docs/). Had to merge it to my master branch and took a bit to actually publish.

linsun:
Agreed, I'd appreciate a staging site :)  I think it is a good idea to move up Quick Start up to one level @geeknoid

Tahler:
@linsun You can see the staging site here:
image

@geeknoid Thanks for the feedback. I definitely see what you're talking about. I still think there's a problem with the current layout, however. The setup section is rather disorganized and several people were frustrated with that. Even in the description, 

    Setup contains instructions for installing the Istio control plane in various 
    environments, as well as instructions for installing the sidecar in the application
    deployment. Quick start instructions are available for Kubernetes and Docker 
    Compose w/ Consul.

it seems to me the setup section has too much going on. It mixes quick start / experimental installation instructions with production installation instructions. The sidecar and mesh expansion docs depend on other docs.

The solution I came up was to make the layout more hierarchical and decision-tree-like. I'm curious to hear your thoughts on that tree in the [doc](https://docs.google.com/document/d/1ks8lpbK1lOh1HKLqcvo9QOcyeFwWcx5o7MB6A9-ycr0/edit).

At the very least, I think there should be more clear focus in the docs what user is going to be reading them (e.g. Is this doc only for users who are wanting to experiment? Is this doc only for people who already have Istio installed on their cluster?) and it should be quick for a user to find the installation doc for their use case. In its current state, a new user basically has to linear search through the documents or restart their progress when they realize they didn't want to install a certain way.

tl;dr I think the setup needs more hierarchy and focus.

linsun:
Thanks, preview was neat.

Should we at least start with Kubernetes as the first item in quick start?  Also, it is probably good to have each provider have its own quick start so the quick start could be short and focus on docker for k8s or minikube (if works).

andraxylia:
@Tahler where is the staging site? I do not see a link above.

andraxylia:
Staging site is here: https://deploy-preview-1338--preliminary-istio.netlify.com/

linsun:
I think this layout is good.  My only comment is can we have Kubernetes as the first item in quick start?

linsun:
also, pls resolve merge conflict so we could get this in :)

Tahler:
Are we sure we want to get this in? I don't think we resolved @geeknoid's thoughts on it so we don't have consensus on this.

geeknoid:
Since the site has changed so much, I'm closing this PR. Please resubmit with updates if you think this is valuable.

Thanks.

PR #1342, opened 2018-05-23
Renamed the VirtualService's destination name to host

ymesika:
Otherwise users would experience an `istioctl` error when trying to create the sample VirtualService snippet.


ymesika:
/hold

Will update the name in `DestinationRule` too..

PR #1343, opened 2018-05-23
Fix markdown code blocks using "```command"

Tahler:
HEAD - no top padding and buggy indented code block:
image

With changes:
image


geeknoid:
Trimming is a good idea.

I don't think we should change the top padding though. The idea is that the vertical layout of PRE blocks needs to be the same before & after the manipulations done in misc.js. Otherwise, the pages flicker when they load. That is, stuff loads and displays in one way, then JavaScript comes around and changes the DOM, causing everything to flicker and redraw. I've worked hard on the site to avoid all flickering, shifting, etc, so that our pages are rock solid from that standpoint.

Once we switch to Hugo and have our new end-to-end pipeline, I'm hoping that the DOM patching we do in misc.js can be done statically in the original HTML. This would give us additional freedom and would further improve loading speed. There's this thing called PhantomJS where we could load the pages in memory, run the DOM patching logic, and then save the pages back out. Run this in a loop for every page and upload to the web server....

So mind if we just add the trim functionality, but leave the padding alone?

Tahler:
Hmm. I still think the output touching the top of its div is rather ugly, but I see your point about the cascading flickering.

I removed the padding change.

geeknoid:
This change messes up the formatting of multi-line command-lines by removing all the leading space on the lines. You can see the effect here: https://preliminary.istio.io/docs/setup/kubernetes/quick-start.html#minikube

Compare that to: https://istio.io/docs/setup/kubernetes/quick-start.html#minikube

Tahler:
Wasn't sure if I should revert or not but there's a fix at #1358.

PR #1344, opened 2018-05-23
Added route rule to virtual service in Egress task

nrjpoddar:
The example virtual service in Egress task was missing route rules which was
causing istioctl to reject the configuration.


nrjpoddar:
Addresses https://github.com/istio/istio.github.io/issues/1345

PR #1353, opened 2018-05-24
Rename firewall-rules in "Control Ingress Traffic"

### LINKED ISSUE #1352: Control Ingress Traffic - duplicate firewall names
The [instructions](https://preliminary.istio.io/docs/tasks/traffic-management/ingress.html#determining-the-ingress-ip-and-ports-for-a-nodeport-ingress-gateway) creates two firewalls, but they both have the same name which is disallowed by GCP. 

Also seems like the name, "allow-book" is a relic of bookinfo but correct me if I'm wrong.

Easy fix: call them "allow-httpbin-http" and "allow-httpbin-https"

This was the only issue I had while following these instructions.

Tahler:
Resolves #1352.


Tahler:
Done.

PR #1356, opened 2018-05-25
Some cleanup for mesh expansion.

gyliu513:
1) Added an example ILB
2) Port 27017 is not needed.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -88,6 +88,7 @@ Istiofied
 JSON-formatted
 JWT
 JWTs
+Keepalived
 Kibana
 Kops
 Kuat
--- a/content/docs/setup/kubernetes/mesh-expansion.md
+++ b/content/docs/setup/kubernetes/mesh-expansion.md
@@ -36,7 +36,9 @@ You should customize it based on your provisioning tools and DNS requirements.
 
 *   Setup Internal Load Balancers (ILBs) for Kube DNS, Pilot, Mixer and Citadel. This step is specific to
 each cloud provider, so you may need to edit annotations. You can use an ILB based on keepalived at
-[here](https://github.com/gyliu513/work/tree/master/k8s/charts/keepalived) for test.
+[here](https://github.com/gyliu513/work/tree/master/k8s/charts/keepalived) for demo or test in case where
+the cloud provider or private cloud (for example IBM Cloud Private) doesn't have load balancer service
+support out of box.
 
     ```command
     $ kubectl apply -f install/kubernetes/mesh-expansion.yaml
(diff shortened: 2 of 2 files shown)


_docs/setup/kubernetes/mesh-expansion.md
```diff
@@ -38,13 +38,14 @@ You should customize it based on your provisioning tools and DNS requirements.
 ### Preparing the Kubernetes cluster for expansion
 
 *   Setup Internal Load Balancers (ILBs) for Kube DNS, Pilot, Mixer and Citadel. This step is specific to
-each cloud provider, so you may need to edit annotations.
+each cloud provider, so you may need to edit annotations. You can use an ILB based on keeplived at
```
geeknoid: keeplived -> keepalived

gyliu513: Done

gyliu513:
ping @geeknoid , can this get merged?

geeknoid:
Presubmit test isn't passing. You need to add keepalived to the dictionary (.settings file at the root)

gyliu513:
@geeknoid did not find the `.settings` file, can you please show me a link for this?

linsun:
I'm interested in understanding how to add a word to bypass spelling check too.  cc @geeknoid 

@gyliu513 a little concerned with the keepalived off your own repo in master.  I think it'd be helpful to explain why people wants to use the test approach based on keepalived.  Is it when cloud provider doesn't provide service type load balancer for private cloud IPs?

gyliu513:
@linsun currently, only public cloud can provide loadbalancer services, such as AWS etc, but there does not seems any solutions can be used by on-prem clusters, so here I was providing an example keepalived chart, this can help the customer to do some test and demo if they do not have a loadbalancer, hope it is OK.

linsun:
@gyliu513 I see, thanks, would be good to mention why users would use it, e.g. in the case where the cloud provider or private cloud (for example ICP) doesn't support k8s service type load balancer service.

geeknoid:

Built with commit c549144176ff348b9307adddc7dd20f22f4725b8

https://app.netlify.com/sites/preliminary-istio/deploys/5b0eb0d1dd6a546397ec028f

geeknoid:
Sorry, I didn't see these comments earlier.

The dictionary of exceptions is located at the root of the istio.github.io in .settings. Just add a word in there when needed. Note that often, words like keepalived should be in backticks `. When you put something in backticks, it is automatically excluded and not spell checked.

https://github.com/istio/istio.github.io/blob/master/.spelling

gyliu513:
ping @geeknoid @linsun , all comments are now addressed.

PR #1366, opened 2018-05-28
A number of visual improvements.

geeknoid:
- The privacy link in the footer is now actually visible and uses the same coloring as the rest of the
text in the footer.

- The gear menu in the top right now actually looks like a real menu instead of a panel of links. Highlights behave as expected and
visuals are cleaner.

- The :before hack to cause proper scrolling in the face of our header is now more selective. It's only applied to header items instead of all items that have an id attribute. This is more surgical in nature and avoids potential issues such as was worked around by a counter-hack for tables. The counter hack is no longer needed.

- Don't show the Edit This Pages on GitHub menu items for pages that are generated outside of the main doc process (reference pages mainly). This won't kick in until I regenerate the reference docs, at which point they'll gain the 'generated' front-matter entry, which will trigger the removal of the menu item.


geeknoid:

Built with commit 400228e352308fbea7c7e5bdac253b94761396cc

https://app.netlify.com/sites/preliminary-istio/deploys/5b0cbb061f12b76ed13c66c5

linsun:
1. for the privacy link, could we link to a privacy link in istio.io or istio github since this is a community project?

2/3. LGTM

4. I tested a few pages, seems we still show edit this page?

https://deploy-preview-1366--preliminary-istio.netlify.com/docs/reference/config/istio.networking.v1alpha3/
https://deploy-preview-1366--preliminary-istio.netlify.com/docs/reference/commands/istioctl/

Perhaps I'm missing something?

linsun:
SGTM, i appreciate you follow up on the privacy link in a separate PR.

PR #1367, opened 2018-05-28
Change `details` to container but not VM for mesh expansion.

gyliu513:
The example in https://preliminary.istio.io/docs/guides/integrating-vms/
actually do not have `details` running in VM.

/cc @costinm @geeknoid @linsun @sdake


linsun:
Could we add details service in the diagram to be consistent with the other bookinfo diagram?   I assume it is in k8s?

gyliu513:
@linsun done, thanks.

PR #1369, opened 2018-05-28
Use `command` istead of `bash` for YAML template.

gyliu513:
/cc @geeknoid @frankbu


geeknoid:
All these types of uses need to stay as bash. Otherwise, using command yields incorrect rendering.

geeknoid:
Shoot. I found this on the troubleshooting page yesterday too. The markdown parser used by the new Hugo toolchain, called BlackFriday, is getting confused and not processing the code block as a code block. It's happening only with indented blocks, and only when the block has lines that start with specific sequences (like with -). I filed a bug with the Hugo team for this.

In the meantime, let me see if I can improve things.

PR #1372, opened 2018-05-28
Workaround Hugo formatting bug with indented code blocks that have lines that start with -

geeknoid:


gyliu513:
@geeknoid I found issues still there at https://preliminary.istio.io/docs/tasks/traffic-management/ingress/#configuring-a-gateway-for-http

image

geeknoid:
That's been definitely fixed, it doesn't show up when I generate the site locally. THis means it's probably something to do with netlify not having rebuilt the site...

geeknoid:
Yep, publishing failed because Netlify wasn't able to complete a build of the site because the link checker got a timeout when contacting a particular URL. I'm rerunning the site build now.

PR #1376, opened 2018-05-30
Fixes to the upgrade and sidecar docs

costinm:


geeknoid:

Built with commit c75f303e896099350f3a5bd42954fbdf02cc00cb

https://app.netlify.com/sites/preliminary-istio/deploys/5b0df075b31274114c59db46

content/docs/setup/kubernetes/upgrading-istio.md
```diff
@@ -55,20 +53,19 @@ or
...
+The applications already running istio will still using the sidecar from 0.7.1 and will continue to work. After the control plane is upgraded, you will need to re-inject so they run with the new version of sidecar proxy. There are two cases: Manual injection and Automatic injection.
 
 1.  Manual injection:
 
     If automatic sidecar injection is not enabled, you can upgrade the
     sidecar manually by running the following command:
 
     ```command
-    $ kubectl apply -f <(istioctl kube-inject -i $ISTIO_NAMESPACE -f $ORIGINAL_DEPLOYMENT_YAML)
+    $ kubectl apply -f <(istioctl kube-inject -f $ORIGINAL_DEPLOYMENT_YAML)
```
linsun: shouldn't the user specify proxyv2 on the inject cmd because the default is proxyv1 in this case?

content/docs/setup/kubernetes/upgrading-istio.md
```diff
@@ -89,3 +86,20 @@ of sidecar proxy. There are two cases: Manual injection and Automatic injection.
 
     There are some tricks to reload all pods. E.g. There is a [bash script](https://gist.github.com/jmound/ff6fa539385d1a057c82fa9fa739492e)
     which triggers the rolling update by patching the grace termination period.
+
+### Migrating to the new networking APIs
+
```
linsun: should we move this section earlier under ### sidecar upgrade? the proxyimage annotation is useful for auto injection and manual injection (I think).

content/docs/setup/kubernetes/upgrading-istio.md
costinm: That would be for migration from alpha1 to alpha3 - I think the doc at this point is about users who want to keep using alpha1/v1 and upgrade to 0.8. They will still use proxyv1.

content/docs/setup/kubernetes/upgrading-istio.md
costinm: I think there are different things - first you upgrade your existing app and install to 0.8, including injecting to 0.8 v1 sidecar.

We do need more content on how to use the conversion tool - I just mentioned how to upgrade an individual pod to proxyv2, this should link or include the info on how to convert the config.

costinm:
Not very familiar with lint - it says 2 spelling errors but not clear on which words.

content/docs/setup/kubernetes/upgrading-istio.md
mandarjog: With auth you get the following error about nodePort collision, this doc already improves the state of the art so I will work on fixing the other part.

```
Error from server (Invalid): error when applying patch:
{"metadata":{"annotations":{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"v1\",\"kind\":\"Service\",\"metadata\":{\"annotations\":{},\"labels\":{\"chart\":\"ingress-0.8.0\",\"heritage\":\"Tiller\",\"istio\":\"ingress\",\"release\":\"RELEASE-NAME\"},\"name\":\"istio-ingress\",\"namespace\":\"istio-system\"},\"spec\":{\"ports\":[{\"name\":\"http\",\"nodePort\":32000,\"port\":80},{\"name\":\"https\",\"port\":443}],\"selector\":{\"istio\":\"ingress\"},\"type\":\"LoadBalancer\"}}\n"},"labels":{"chart":"ingress-0.8.0","heritage":"Tiller","release":"RELEASE-NAME"}},"spec":{"$setElementOrder/ports":[{"port":80},{"port":443}],"ports":[{"nodePort":32000,"port":80}]}}
to:
&{0xc420abd140 0xc420158620 istio-system istio-ingress install/kubernetes/istio-auth.yaml 0xc42149ec60 0xc42149ee50 9964381 false}
for: "install/kubernetes/istio-auth.yaml": Service "istio-ingress" is invalid: spec.ports[0].nodePort: Invalid value: 32000: provided port is already allocated
Error from server (Invalid): error when creating "install/kubernetes/istio-auth.yaml": Service "istio-ingressgateway" is invalid: spec.ports[0].nodePort: Invalid value: 31380: provided port is already allocated
```

content/docs/setup/kubernetes/sidecar-injection.md
```diff
@@ -135,7 +135,7 @@ with the option `sidecarInjectorWebhook.enabled` set to `false`. E.g.
 ```command
 $ helm template --namespace=istio-system --set sidecarInjectorWebhook.enabled=false install/kubernetes/helm/istio > istio.yaml
 $ kubectl create ns istio-system
-$ kubectl apply -f istio.yaml
+$ kubectl apply -n istio-system -f istio.yaml
```
gyliu513: I think we do not need to specify `namespace` here as we already specified `namespace` when creating the template.

content/docs/setup/kubernetes/upgrading-istio.md
```diff
@@ -89,3 +86,20 @@ of sidecar proxy. There are two cases: Manual injection and Automatic injection.
...
+kind: Deployment
+...
+spec:
+  template:
+    annotations:
+       sidecar.istio.io/proxyImage: istio/proxyv2:0.8.latest
+
+```
+
+When all your applications have been migrated and tested, you can repeat the istio upgrade process, removing the '--set global.proxy.image=proxy' option.
```
geeknoid: replace ' with `

content/docs/setup/kubernetes/upgrading-istio.md
costinm: Node port should be off by default - only used for testing (or special cases)

content/docs/setup/kubernetes/sidecar-injection.md
costinm: Let's keep it - there is a bit of confusion about where it goes, this makes it a bit more consistent.

PR #1382, opened 2018-05-30
Update authn doc

jsenon:
Change fqdn local.cluster to cluster.local


linsun:
Note: file changes tab showed 0 changes for me.

jsenon:
@linsun indeed, it seems that modification has been done before merge my pr

PR #1391, opened 2018-05-31
Delete stale files.

geeknoid:


ldemailly:
? how did that show up? did I somehow commit both old and new locations?

PR #1396, opened 2018-05-31
Fix the table format in expression-language.md

yangminzhu:
Add `|` to make the table correct and use `\` to escape the `|` character inside the table. Not sure if this is the right way to fix but the preview page looks good.


geeknoid:

Built with commit 2f6dbba1585ce1e899f132ed19c2cdce16a0cb9e

https://app.netlify.com/sites/preliminary-istio/deploys/5b1042d411b73b07722cf03b

yangminzhu:
@geeknoid 
I don't see how would this PR cause it? The log says it failed because unable (404) to find a lot of .png files.

geeknoid:
Sorry, I've been playing with stuff and caused these 404. Once those are fixed, I'll rerun your changes and check them in after that.

Thanks for fixing this.

PR #1401, opened 2018-06-01
update limitation of converter tool

linsun:


content/about/notes/0.8.md
```diff
@@ -60,3 +60,5 @@ and about [`istioctl gen-deploy`](/docs/reference/commands/istioctl/#istioctl%20
 
 - There are [problems with Kubernetes 1.10.2](https://github.com/istio/istio/issues/5723).
 The workaround is to use Kubernetes 1.9 or switch the node image to Ubuntu.
+
+- There is a known namespace issue with `istioctl experimental convert-networking-config` tool, please manually adjust to use the desired namespace after running the conversation tool.   [Learn more] (https://github.com/istio/istio/issues/5817)
```
mandarjog: Can you specifically mention that it always uses `istio-system` namespace regardless of the namespace of the original resource.

linsun: well, the prob is some virtual service has correct namespace, some doesn't.  also my destination rule has the correct namespace.

content/about/notes/0.8.md
```diff
@@ -60,3 +60,5 @@ and about [`istioctl gen-deploy`](/docs/reference/commands/istioctl/#istioctl%20
 
 - There are [problems with Kubernetes 1.10.2](https://github.com/istio/istio/issues/5723).
 The workaround is to use Kubernetes 1.9 or switch the node image to Ubuntu.
+
+- There is a known namespace issue with `istioctl experimental convert-networking-config` tool where the desired namespace may be changed to the istio-system namespace, please manually adjust to use the desired namespace after running the conversation tool.   [Learn more] (https://github.com/istio/istio/issues/5817)
```
gyliu513: Should be 

[Learn more](https://github.com/istio/istio/issues/5817)

Remove the space between `[Learn more]` and `(https://github.com/istio/istio/issues/5817)`

PR #1404, opened 2018-06-01
#1324 fix: istio-telemetry is third item in the list after istio-mixe…

olegchorny:
…r and istio-policy


olegchorny:
I've found better fix, let's close this one

PR #1421, opened 2018-06-03
use HTTPS egress instead of HTTP + TLS origination

vadimeisenbergibm:
is possible due to the SNI routing feature

### DIFF
--- a/content/docs/tasks/traffic-management/egress.md
+++ b/content/docs/tasks/traffic-management/egress.md
@@ -69,24 +69,9 @@ from within your Istio cluster. In this task we will use
           ports:
           - number: 443
             name: https
-            protocol: HTTP
-        ---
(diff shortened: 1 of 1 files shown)


rshriram:
How about we mention both ways? Alternatively, we could mention https direct from sidecar and routing via egress gateway with https origination from the gateway.

vadimeisenbergibm:
@rshriram @geeknoid I propose that we add a section "Advanced Tasks" that will cover more advanced/esoteric use cases. I thought about the following esoteric tasks:
1. mTLS with ingress
2. TLS origination for egress
3. Directing egress traffic thru egress gateway
4. mTLS from egress gateway

Otherwise, this task will become too complex. Also the Tasks section may become to large and complex to go over. So I propose to have short tasks that cover the basic commonly used functionality, and putting esoteric use cases into Advanced Tasks section.

PR #1427, opened 2018-06-04
remove --debug as it is deprecated

linsun:


ymesika:
There are several more docs with the `--debug` flag. I think they should be updated too in this PR.
It is both deprecated **and** there is no proxyv2 debug image at the moment so it's just useless flag in 0.8.

geeknoid:
This change was already applied by another PR.

PR #1428, opened 2018-06-04
return the instructions for deploying httpbin without Istio initializer

vadimeisenbergibm:
it seems it fell thru during some merge

### DIFF
--- a/content/docs/tasks/traffic-management/ingress.md
+++ b/content/docs/tasks/traffic-management/ingress.md
@@ -31,6 +31,12 @@ This task describes how to configure Istio to expose a service outside of the se
     $ kubectl apply -f samples/httpbin/httpbin.yaml
     ```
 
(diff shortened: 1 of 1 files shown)


content/docs/tasks/traffic-management/ingress.md
```diff
@@ -31,6 +31,12 @@ This task describes how to configure Istio to expose a service outside of the se
     $ kubectl apply -f samples/httpbin/httpbin.yaml
     ```
 
+    Without _Istio-Initializer_:
```
morvencao: Actually we have replace `Initializer` with `webhook`

gyliu513: This is already merged at https://github.com/istio/istio.github.io/pull/1435

frankbu:
This has been fixed already.
image

PR #1429, opened 2018-06-04
Capitalize NodePort and make a link to Kubernetes documentation

vadimeisenbergibm:


content/docs/tasks/traffic-management/ingress.md
```diff
@@ -50,7 +50,7 @@ istio-ingressgateway   LoadBalancer   172.21.109.129   130.211.10.121  80:31380/
 
 If the `EXTERNAL-IP` value is set, your environment has an external load balancer that you can use for the ingress gateway.
 If the `EXTERNAL-IP` value is `<none>` (or perpetually `<pending>`), your environment does not provide an external load balancer for the ingress gateway.
-In this case, you can access the gateway using the service `nodePort`.
+In this case, you can access the gateway using the service's [node port](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport).
```
gyliu513: How about 

using the service with type as [NodePort](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport)

vadimeisenbergibm: Here I actually mean the (node) port of the service, not the service with type NodePort. @geeknoid what do you think? Is "accessing a service's node port" clear sentence in Kubernetes parlance?

PR #1432, opened 2018-06-04
Highlight cidr.

gyliu513:
We need to highlight cidr, otherwise, cidr will be as follows.

image

/cc @geeknoid @linsun


vadimeisenbergibm:
@gyliu513 Note that this problem has been fixed by https://github.com/istio/istio.github.io/pull/1422.

PR #1443, opened 2018-06-05
Introduce better linking support to github.com/istio/istio

### LINKED ISSUE #567: yaml references to distributed files should be clickable
for instance when looking at
https://istio.io/docs/tasks/traffic-management/traffic-shifting.html

telling the user to
istioctl create -f samples/bookinfo/kube/route-rule-all-v1.yaml
yet having no easy way to just peek at what is in that file is not as helpful as it could be

geeknoid:
- Within a code block, you can now surround a relative file path with @@. This will
cause the path to be rendered as a link to raw.githubusercontent.com/istio/istio/<path>.
This lets the user click on the link to see the content of the file, which is mighty
handy.

- Updated all code blocks to take advantage of the above.

- Introduce support for {{< branch_name >}} which returns the source code branch
name associated with the current doc site.

- Use {{< branch_name >}} in all our references to content in istio/istio on GitHub. This thus
pins our references to the correct version of the content in GitHub. This prevents errors from
gradually appearing in our doc set as content in GitHub starts to diverge from the expectation
in the site content.

Fixes #567


src/js/misc.js
```diff
@@ -283,6 +285,8 @@ function handleDOMLoaded() {
                     }
 
                     if (cmd !== "") {
+                        cmd = cmd.replace(githubRef, "<a href='http://github.com/istio/istio/blob/master/$1'>$1</a>");
```
rshriram: woa. You really became a JS expert!.. 
Can I suggest pointing to the release branch instead of master? Ideally, there would be some variable that would help us point to the right code branch.

geeknoid:

Built with commit 55bff7e701ea6e10a143e95016012a785bf15a54

https://app.netlify.com/sites/preliminary-istio/deploys/5b1767683672df60d23d27d7

geeknoid:
This expands considerably on the original PR. PTAL.

geeknoid:

Built with commit 3b125bf6dbdd32f1ca0fee0c377367d628b045f8

https://app.netlify.com/sites/preliminary-istio/deploys/5b1767898df8941d22afcb96

geeknoid:

Built with commit 415049dfafb40e9f22a99f2cdf14ecea57e97db0

https://app.netlify.com/sites/preliminary-istio/deploys/5b1767d1b31274436ba116e9

PR #1451, opened 2018-06-05
Adding details to "What happened"

cfe84:
Adding in particular that the rules are applied only on "http" ports, not "TCP" - I've been blocked for several hours on having istio work with my app because the ports were TCP.

### DIFF
--- a/content/docs/tasks/traffic-management/request-routing.md
+++ b/content/docs/tasks/traffic-management/request-routing.md
@@ -162,6 +162,22 @@ on a header (i.e., a user cookie) in a request.
 Once the v2 version has been tested to our satisfaction, we could use Istio to send traffic from
 all users to v2, optionally in a gradual fashion. We'll explore this in a separate task.
 
+The `productpage` service is querying for the `reviews` service using a regular HTTP call (i.e.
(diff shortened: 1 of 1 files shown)


cfe84:
I signed it!

content/docs/tasks/traffic-management/request-routing.md
```diff
@@ -162,6 +162,22 @@ on a header (i.e., a user cookie) in a request.
...
+```yaml
+kind: Service
+apiVersion: v1
+...
+spec:
+  ports:
+  - name: http
+    port: 8080
+```
+
```
frankbu: Thanks for pointing out this issue but this is not the right way to explain it. Instead of adding the above, I would suggest adding something like this:

Note that Kubernetes service, like the Bookinfo ones used in this task, must adhere to certain restrictions in order to take advantage of Istio's L7 routing features. Refer to the [sidecar injection documentation](https://istio.io/docs/setup/kubernetes/sidecar-injection/#pod-spec-requirements) for details.

geeknoid:
Closing this PR due to inactivity.

PR #1457, opened 2018-06-06
Remove the deprecated --debug in istioctl

wattli:
Currently the default proxy image has curl already, which means --debug option is not needed, and it's deprecated as well.


ldemailly:
I wish we'd kept the feature, or clearly document the alternative, instead of deleting it but I guess it's good to not have misleading docs

ldemailly:
/lgtm
(with caveat above)

ldemailly:
https://github.com/istio/istio.github.io/pull/1458 for owners issue

PR #1458, opened 2018-06-06
use same owners than https://github.com/istio/istio/blob/master/OWNERS

ldemailly:
and at least more than 2

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,4 +1,13 @@
 
 approvers:
+  - costinm
+  - ldemailly
+  - douglas-reid
   - geeknoid
+  - hklai
   - linsun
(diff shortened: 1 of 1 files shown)


geeknoid:
What's the motivation for wanting this change? Has there been latency to the current set of owners?

I like seeing all the changes going in in order to ensure increased consistency and to avoid common formatting problems.

andraxylia:
@geeknoid we need more Owners, all networking folks should for instance be able to approve changes in the nw related content.

I understand the need to keep a grip on the overall structure, but not on the content.

geeknoid:
Not for now...

PR #1470, opened 2018-06-07
Maxconn

gyliu513:
The example in https://preliminary.istio.io/docs/tasks/traffic-management/circuit-breaking/ want to use  `maxConnections: 1`, but the destinationrules is setting `maxConnections: 100`, we should update document to set it as 1.

/cc @geeknoid @frankbu


geeknoid:

Built with commit 44c8f0d0522c6432b80722f5cead87466a3f45e0

https://app.netlify.com/sites/preliminary-istio/deploys/5b18ed9211b73b74f0997acc

PR #1474, opened 2018-06-07
use html encoding for the forward slash.

gyliu513:
/cc @geeknoid


gyliu513:
@ldemailly 

image

ldemailly:
oh that's odd, must be a magic feature of hugo to do that

PR #1475, opened 2018-06-07
Use virtualservices for basic access control.

gyliu513:
/cc @geeknoid @linsun

### DIFF
--- a/content/docs/tasks/security/basic-access-control.md
+++ b/content/docs/tasks/security/basic-access-control.md
@@ -22,7 +22,7 @@ This task shows how to control access to a service using the Kubernetes labels.
     $ istioctl create -f @samples/bookinfo/routing/route-rule-all-v1.yaml@
     ```
 
-    Save the following YAML snippet as `route-rule-reviews-jason-v2-v3.yaml.yaml`:
+    Save the following YAML snippet as `route-rule-reviews-jason-v2-v3.yaml`:
 
     ```yaml
     apiVersion: networking.istio.io/v1alpha3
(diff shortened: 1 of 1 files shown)


content/docs/tasks/security/basic-access-control.md
```diff
@@ -19,8 +19,38 @@ This task shows how to control access to a service using the Kubernetes labels.
     test user "jason" to version v2 and requests from any other user to v3.
 
     ```command
-    $ istioctl create -f @samples/bookinfo/kube/route-rule-reviews-test-v2.yaml@
-    $ istioctl create -f @samples/bookinfo/kube/route-rule-reviews-v3.yaml@
+    $ istioctl create -f @samples/bookinfo/routing/route-rule-all-v1.yaml@
+    ```
+
+    Save the following YAML snippet as `route-rule-reviews-jason-v2-v3.yaml.yaml`:
```
geeknoid: -v2-v3.yaml.yaml? That's a bit odd isn't it?

gyliu513: I want to highlight that this virutalservice can help direct request to both v2 and v3, while v2 request user jason login. Comments for a new name?

frankbu: I think @geeknoid is just commenting on the extra `.yaml` in the name (i.e., `.yaml.yaml`) ... I assume that's just a typo?

My only question is can we change this task to work with one of the existing yaml files? The fewer files we have the better. Why can't this task be changed to work with v1 and v2, instead of v2 and v3? Seems to me that the only requirement for this task is that we have 2 versions, any 2 will do.

gyliu513: Ah, I see, thanks @geeknoid and @frankbu 

The reason I want to keep `v3` because all of the following content are mentioning red star and I do not make much change for the content, so that's why I post a PR at https://github.com/istio/istio/pull/6102 want to add a new file.

frankbu: Yes, I understand, but I think it's basically just a matter of replacing "v3" with "v1", and replacing "red" with "no" in the document, and then it will work with the existing yaml file.

PR #1478, opened 2018-06-08
Remove global.mtls.mtlsExcludedServices.

gyliu513:
This is a follow up PR for https://github.com/istio/istio/pull/6056

### DIFF
--- a/content/help/faq/security/accessing-control-services.md
+++ b/content/help/faq/security/accessing-control-services.md
@@ -11,6 +11,8 @@ control services for that matter). The default value of `mtlsExcludedServices`
 already contains `kubernetes.default.svc.cluster.local`, which is the default
 service name of the Kubernetes API server.
 
+This parameter was removed in release 1.0.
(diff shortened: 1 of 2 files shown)


content/help/faq/security/accessing-control-services.md
```diff
@@ -11,6 +11,8 @@ control services for that matter). The default value of `mtlsExcludedServices`
 already contains `kubernetes.default.svc.cluster.local`, which is the default
 service name of the Kubernetes API server.
 
+This parameter was removed in release 1.0.
```
geeknoid: In the 1.0 docs, we should not leave breadcrumbs that document previous versions (we have the archive.istio.io snapshots for that purpose). So can we rewrite this entire document focused on what the modern solution is and delete the old stuff completely?

gyliu513: OK, then I think that we can remove this sentence.

PR #1481, opened 2018-06-08
Update GCP-Quickstart DM link to specify 0.8 directly

salrashid123:
@geeknoid 

Currently the DM link from the ```0.8``` docs istio points to ```master``` when instead it should be to the DM template for the release.  (it seems the ```{{<branch_name}}``` variable translates to ```master```):

so currently for ```0.8```, we have 
- [https://istio.io/docs/setup/kubernetes/quick-start-gke-dm/#launch-deployment-manager](https://istio.io/docs/setup/kubernetes/quick-start-gke-dm/#launch-deployment-manager)

eg. if you hover over or inspect the link to for ```Istio GKE Deployment Manager```, it goes master.  THis change alters it to go to the release template explictly.

### DIFF
--- a/content/docs/setup/kubernetes/quick-start-gke-dm.md
+++ b/content/docs/setup/kubernetes/quick-start-gke-dm.md
@@ -42,7 +42,7 @@ caption="GKE-IAM Role"
 
 1.  Once you have an account and project enabled, click the following link to open the Deployment Manager.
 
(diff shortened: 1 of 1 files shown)


salrashid123:
SGTM, i'll cancel ths out.  WHen can i expect the new value to pick up?

salrashid123:
yep, its live

PR #1490, opened 2018-06-09
Fixed default values

fleeto:
Fixed default values, from [values.yaml in 0.8.0](https://github.com/istio/istio/blob/0.8.0/install/kubernetes/helm/istio/values.yaml)


geeknoid:
@sdake?

PR #1495, opened 2018-06-11
Fix Gateway example in Route Rules v1alpha3

x15zhang:
Correct destination field from "name" to "host" in VirtualService for Gateway example


x15zhang:
I signed it!

geeknoid:
Please go change the source proto in the istio/api repo, since this file is generated from that proto.

Thanks.

PR #1497, opened 2018-06-11
Converts a documented values.yaml to MD format

sdake:
Co-Authored-By: Guang Ya Liu <gyliu513@gmail.com>


sdake:
[values.tar.gz](https://github.com/istio/istio.github.io/files/2090858/values.tar.gz)

An example values.yaml file to which this code processes.

sdake:
@gyliu513 can you confirm your OK with merging this PR as per requirements in:https://github.com/istio/istio.github.io/pull/1497#issuecomment-396301038

sdake:
@geeknoid - would like to integrate this somehow into the build system.

The workflow would be something like
1. Pull istio/istio
2. Run MD script on istio/istio/install/kubernetes/helm/values.yaml
3. generate MD files from that

barring that as an option, we can manually update.  If this approach is taken, could you offer up some idea of how to get the md file included in the helm.md file without vi :r?

Cheers
-steve

geeknoid:

Built with commit 9de3b45bafbda9b87e7eb6a7c9fd16b8d506b488

https://app.netlify.com/sites/preliminary-istio/deploys/5b278ea911b73b6e0e26c655

sdake:
Closer - still incomplete as values.yaml hasn't been doc-stringified.

sdake:
@mandarjog can you provide a review of the python code?

TIA
-steve

scripts/tablegen.py
```diff
@@ -0,0 +1,146 @@
...
+            value = nextLine.replace(' ', '')
+            valueList.append(value.lstrip('-').strip())
+            flag += 1;
+        nextLineNum += 1
+
+    if lastLineNum == totalNum - 1 and len(currentLine.lstrip()) != 0 and '#' != currentLine.lstrip()[0]:
+        valueList.append(currentLine.split(':', 1)[1].strip())
+
+    return True, valueList
+
```
gyliu513: kill two lines here

content/docs/setup/kubernetes/helm-install.md
```diff
@@ -91,21 +91,175 @@ may be used in the same Helm operation.
 Helm charts expose configuration options which are currently in alpha.  The currently exposed options are explained in the
 following table:
 
-| Parameter | Description | Values | Default |
+<!-- AUTO-GENERATED-START -->
```
gyliu513: @sdake do we need to add some comments for how does this table is generated with the python script at here?

sdake: Yes we can add comments on line 93 - think it would make sense.

scripts/tablegen.py
sdake: ack

scripts/tablegen.py
```diff
@@ -0,0 +1,146 @@
...
+    if lineNum != lastLineNum:
+        return False, valueList
+
+    for nextLineNum in range(lineNum + 1, totalNum):
+        nextLine = context[nextLineNum]
+        if  len(nextLine.lstrip()) != 0 and '#' != nextLine.lstrip()[0] and ':' in nextLine:
+            if whitespaces >= (len(nextLine) - len(nextLine.lstrip())) / 2:
+                if flag == 0:
+                    valueList.append(currentLine.split(':', 1)[1].strip())
+                    return True, valueList
```
sdake: this is odd logic - why not just return True, valueList at the scope of the if whitespace conditional?

gyliu513: Yes, we can use one `return True, valueList`

```
if flag == 0:
    valueList.append(currentLine.split(':', 1)[1].strip())

return True, valueList
```

scripts/tablegen.py
sdake: done.

scripts/tablegen.py
sdake: done

sdake:
G R O A N

PR #1505, opened 2018-06-13
Removing tracing since it's not deployed

carlioth:
Jaeger isn't deployed when running the deployment manager: 
https://github.com/istio/istio/blob/master/install/gcp/deployment_manager/istio-cluster.jinja
Hence the commands will not work, and the link is broken to "understand what happenede"

### DIFF
--- a/content/docs/setup/kubernetes/quick-start-gke-dm.md
+++ b/content/docs/setup/kubernetes/quick-start-gke-dm.md
@@ -226,23 +226,6 @@ http://localhost:8088/dotviz
 
 For more details, see [About the ServiceGraph Add-on](/docs/tasks/telemetry/servicegraph/#about-the-servicegraph-add-on).
 
-## Tracing
-
-Set up a tunnel to the tracing dashboard:
-
-```command
-$ kubectl port-forward -n istio-system $(kubectl get pod -n istio-system -l app=jaeger -o jsonpath='{.items[0].metadata.name}') 16686:16686 &
-```
-
-You should see the trace statistics sent earlier on [http://localhost:16686](http://localhost:16686)
-
-{{< image width="100%" ratio="42.35%"
-    link="../img/dm-tracing.png"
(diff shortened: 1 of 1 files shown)


content/docs/setup/kubernetes/quick-start-gke-dm.md
```diff
@@ -226,23 +226,6 @@ http://localhost:8088/dotviz
 
 For more details, see [About the ServiceGraph Add-on](/docs/tasks/telemetry/servicegraph/#about-the-servicegraph-add-on).
 
-## Tracing
```
gyliu513: Tracing should already be included, why remove this? both istio-demo.yaml and istio-demo-auth.yaml should already include tracing.

carlioth: Perhaps I'm missing something in this guide: https://istio.io/docs/setup/kubernetes/quick-start-gke-dm/ 
But following the guide, I'll click this link: https://accounts.google.com/signin/v2/identifier?service=cloudconsole&continue=https://console.cloud.google.com/launcher/config?templateurl=https://raw.githubusercontent.com/istio/istio/release-0.8/install/gcp/deployment_manager/istio-cluster.jinja&followup=https://console.cloud.google.com/launcher/config?templateurl=https://raw.githubusercontent.com/istio/istio/master/install/gcp/deployment_manager/istio-cluster.jinja&flowName=GlifWebSignIn&flowEntry=ServiceLogin

When deploying that config Jaeger isn't deployed, and hence the information in the Tracing chapter will not work. It would be good to include the tracing in this example, perhaps an update to the jinja file to include tracing (Jaeger) should be done? 
https://github.com/istio/istio/blob/master/install/gcp/deployment_manager/istio-cluster.jinja

geeknoid:
@gyliu513 What do you think we should do about this PR?

geeknoid:
OK, I opened an issue.

PR #1518, opened 2018-06-14
docker image to run website linters

ayj:
docker build -t gcr.io/istio-io/website-builder .
gcloud docker -- push gcr.io/istio-io/website-builder


geeknoid:
Wow. You make it seem so simple!

PR #1524, opened 2018-06-15
fix command to get INGRESS_POD_NAME

tomkukral:
Pod name is `istio-ingressgateway-...` so grep command should be
adjusted to match anything. Printing variable to provide way to check if
anything was matched.


tomkukral:
I signed it!

PR #1526, opened 2018-06-15
[Blog post] Proxyless Istio

subhraveti:
This blog discusses optimizations to reduce the end-to-end latency
of Istio.


subhraveti:
It may be easier to review and comment on a google doc than the PR -- https://docs.google.com/document/d/1p4Iy-0uBibSryMqDnVpKla91ewHN-9gzvgzF3YxQRBY/edit#

geeknoid:
The checks report a number of spelling errors. You can fix those in the markdown file, and if there are words that are legitimate, you can add them to the .spelling file at the root of the repo. Another way to fix "typos" is to put them in backticks so they're treated as code blocks, which is often appropriate.

subhraveti:
Thanks @geeknoid, I have seen to those errors.

geeknoid:
To get the deploy/netlify check to pass, could you please rebase your change and push it again? That'll grab the latest makefile version which will make netlify happy.

Thanks.

rshriram:
High level feedback after skimming through the draft: Please clearly state the drawbacks of using the approach you describe. The article reads like an academic paper in that it seems to imply the sys call detour/handoff greatly improves latency, but it hides a myriad of issues underneath - in terms of protocol decoding/mutation, policy enforcement per source, failure model and the threat model.

subhraveti:
Thanks for the review @rshriram.  Just a couple points to note:

Proposed mechanisms are meant to be strictly complementary and only address existing shortcomings.  For example, proxyless mechanism would *only* be used for the cases where latency is a bigger concern or where protocol decoding is not possible / necessary.  Existing dataplane remains as-is for the rest of the cases.  In fact the socket delegation mechanism is meant to optimize exactly that case.

The goal of this blog is to build awareness in the community around how some of these alternate approaches could help with current performance concerns.  What is discussed in this early proposal is just a starting point for the discussion.  Obviously the actual code may take a very different form.

rshriram:
Yes, I understand. and I know what this is doing :). All I am saying is to clearly enumerate the shortcomings of this approach, so that the reader is aware of all the tradeoffs. Bypassing a local proxy and using a host-level shared proxy has a host of security concerns with respect to using the identity, etc. It also has some availability issues (daemon dies, all pods affected).

subhraveti:
Actually it is neither shared nor proxy.  The daemon could be run per-pod or per-host.  When it is run per-pod, there is no DaemonSet.  In either case, it is not a proxy.

rshriram:
Proxy being a generic term to indicate anything that forwards request on behalf of the application. Doesn't matter if its envoy or some other daemon that traps syscalls and makes the calls on behalf of the app.

frankbu:
@subhraveti cool stuff, but I also think the article currently reads too much like an academic paper. I think it would come off more practical if you mention AppSwitch up front (maybe even in the title), and focus on how AppSwitch can be used to reduce latency (replace proxy and/or iptables mapping) in an Istio mesh, maybe even tying in bookinfo services as example. Sort of more like your AppSwitch blog :-) Also, clarifying what you said to @rshriram above, that it would be used in conjunction with Envoy, it's just used to optimize the cases where it's needed to reduce latency, would be good.

subhraveti:
Thanks for the review @frankbu.  That makes sense.  Let me update the PR with the changes you suggest.

geeknoid:
Hi, I did some refactoring and the file you edited has moved in the repo. Please rebase and push again. Sorry for the inconvenience.

subhraveti:
Thanks @geeknoid.  Just as a quick update, I am working through the surgical changes suggested by @frankbu.  It is shaping up well.  I will rebase and update shortly.

subhraveti:
@frankbu @rshriram I have taken a pass at the draft based on your suggestions.  It should sound less academic now ; ) PTAL.  Thanks,

frankbu:
Much better, however the one thing that is still lacking is practical relevance to the Istio project. The article leaves the reader thinking that AppSwitch can theoretically improve Istio in a number of ways, but how it might actually be integrated is still pretty much a mystery.

IMO the article will gain much more interest in the Istio community if it describes specific integration ideas. Could Istio's sidecar injection be changed to inject AppSwitch instead of, or in addition to Envoy, into pods? Would this be controlled by user config? Would using AppSwitch instead of a proxy need to be decided on an app by app basis? How would users know whether AppSwitch will work for all the features needed in their use case? What would be a reasonable approach to start integrating AppSwitch with Istio? Would using AppSwitch, instead of iptables, to redirect traffic to a sidecar be a good first step? This is the kind of thing I would have expected to read about in the final net net section.

As it is, it's interesting, but as far as the Istio project is concerned, I think it leaves the reader with no answer to "so, where do we go from here?".

subhraveti:
Thanks @frankbu, these are some very excellent questions.  I wanted to cover at least some of them but wasn't sure if that might be an overload.  I see that not addressing them at all leaves an unsettling feeling.  Let me get back with another update.

subhraveti:
Hi @frankbu, this took a little long due to some travel (more of which is coming up this week).  Look forward to your feedback.  Thanks much.

frankbu:
I think this blog looks good now.
/lgtm

geeknoid:
See the README.md at the root of the repo for info on fixing the linting errors.

subhraveti:
Thanks @geeknoid, I took care of them, except there seems to be an error in a different blog.

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+3.99ms
+72.96ms
+18x
+8x
+appswitch
+appswitch-0
+AppSwitch
+bitpipe
+DaemonSet
+datapath
```
rshriram: data path

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+decap
```
rshriram: ??

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+decapsulated
```
rshriram: unencapsulated

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+delayerification
```
rshriram: layerification is not a proper word either. I would remove all uses of this and use a proper expansion

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+Delayerification
+Devirtualization
```
rshriram: this is a compiler concept. Using it in context of networking will throw people off. Also suggest using proper phrase instead of new de-* terms

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+encap
+FDs
+fine-grain
```
rshriram: fine grained. and if you do it, this exception shouldn't exist

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+AppSwitch
+bitpipe
+DaemonSet
+datapath
+decap
+decapsulated
+delayerification
+Delayerification
+Devirtualization
+encap
```
rshriram: ??

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+FDs
+fine-grain
+goto
+hardcoded
+intra-host
```
rshriram: remove. no hyphens

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
+perf.png
+pre-connected
+proxyless
+Proxyless
+quo
+rethink
+reimplement
+sidecar
+subnets
+touchpoints
```
rshriram: two words

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
+---
+title: Delayering Istio with AppSwitch
+description: Automatic application onboarding and latency optimizations by removing unnecessary layers with AppSwitch
+publishdate: 2018-06-28
+subtitle:
+attribution: Dinesh Subhraveti
+weight: 90
+keywords: [performance, traffic-management, appswitch]
```
rshriram: Performance, appswitch. This is not just traffic management

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
+---
+title: Delayering Istio with AppSwitch
```
rshriram: Improving Istio Performance using AppSwitch

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+description: Automatic application onboarding and latency optimizations by removing unnecessary layers with AppSwitch
```
rshriram: I would not call these layers unnecessary until AppSwitch can be used as a 100% replacement for Envoy for all language runtimes.
Please rephrase as "Enhancing performance of certain latency-sensitive applications using AppSwitch in-process load-balancing

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+subtitle:
+attribution: Dinesh Subhraveti
+weight: 90
+keywords: [performance, traffic-management, appswitch]
+---
+
+>_All problems in computer science can be solved with another layer, except of course the problem of too many layers._ -- David Wheeler
+
+The sidecar proxy approach enables a lot of awesomeness.  Squarely in the datapath between microservices, the sidecar can precisely tell what the application is trying to do.  It can monitor and instrument protocol traffic, not in the bowels of the networking layers but at the application level, to enable deep visibility, access controls and traffic management.
+
```
rshriram: please rephrase. its a bit crude to say "bowels of the networking layers".

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+If we look closely however, there are many intermediate layers that the data has to pass through before the high-value analysis of application-traffic can be performed.  Most of those layers are part of the base plumbing infrastructure that are there just to push the data along.  In so doing, they add latency to communication and complexity to the overall system.
+
```
rshriram: In doing so

content/blog/2018/delayering-istio/delayering-istio.md
rshriram: actually ignore.. delayering seems fine

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
+---
+title: Delayering Istio with AppSwitch
+description: Automatic application onboarding and latency optimizations by removing unnecessary layers with AppSwitch
+publishdate: 2018-06-28
+subtitle:
+attribution: Dinesh Subhraveti
```
rshriram: Please add affiliation. I would add AppOrbit and Columbia University

content/blog/2018/delayering-istio/delayering-istio.md
rshriram: ignore

content/blog/2018/delayering-istio/delayering-istio.md
rshriram: In other words, this could in theory handle policies, etc. So better to keep this focused on perf and appswitch

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Don’t optimize layers, remove them
+
+In my belief, optimizing something is a poor fallback to removing its requirement altogether.  That was the goal of my [initial work](https://apporbit.com/a-brief-history-of-containers-from-reality-to-hype) on OS-level virtualization that led to Linux containers which effectively [removed virtual machines](https://www.oreilly.com/ideas/the-unwelcome-guest-why-vms-arent-the-solution-for-next-gen-applications) by running applications directly on the host operating system without requiring an intermediate guest.  For a long time the industry was fighting the wrong battle distracted by optimizing VMs rather than removing the additional layer altogether.
+
+I see the same pattern repeat itself with the connectivity of microservices, and networking in general.  Network has been going through the changes that physical servers have gone through a decade earlier.  New set of layers and constructs are being introduced.  They are being baked deep into the protocol stack and even silicon without adequately considering low-touch alternatives.  Perhaps there is a way to remove those additional layers altogether.
```
rshriram: Networking has.. or The network has

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+
+{{< image width="25%" ratio="175%"
+    link="burger.jpg"
+    alt="What current network stack has become"
+    >}}
```
rshriram: while this is an opinion/tech article, this picture is more suitable to a presentation. I am not sure this will fit the rest of the blogs in the Istio site, that are more matter of fact figures about the architecture, etc.
Thoughts @geeknoid ?

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Delayerification
+
+Now that we have an idea about what AppSwitch does, let’s look at the layers that it optimizes away from a standard service mesh.
+
+### Network Devirtualization
+
+Kubernetes offers simple and well-defined network constructs to the microservice applications it runs.  In order to support them however, it imposes specific [requirements](https://kubernetes.io/docs/concepts/cluster-administration/networking/) on the underlying network.  Meeting those requirements is often not easy.  The goto solution of adding another layer is typically adopted to satisfy the requirements.  In most cases the additional layer consists of a network overlay that sits between Kubernetes and underlying network.  Traffic produced by the applications is encapsulated at the source and decapsulated at the target which not only costs network resources but also takes up compute cores due to the additional processing.
```
rshriram: s/goto/go to/

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+
+Because AppSwitch arbitrates what the application sees through its touchpoints with the platform, it projects a consistent virtual view of the underlying network to the application similar to an overlay but without introducing an additional layer of processing along the datapath.  Just to draw a parallel to containers, the inside of a container looks and feels like a VM.  However the underlying implementation does not intervene along the high-incidence control paths of low-level interrupts etc.
+
+AppSwitch can be injected into a standard Kubernetes manifest (similar to Istio injection) such that the application’s network is directly handled by AppSwitch bypassing any network overlay underneath.  More details to follow in just a bit.
```
rshriram: application's network traffic are

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Skip TCP/IP for colocated endpoints
+
+TCP/IP is the universal protocol medium over which pretty much all communication occurs.  But if application endpoints happen to be on the same host, is TCP/IP really required?  After all, it does do quite a bit of work and it is quite complex.  Unix sockets are explicitly designed for intra-host communication and AppSwitch can transparently switch the communication to occur over a Unix socket for colocated endpoints.
+
+For each listening socket of an application, AppSwitch maintains two listening sockets, one each for TCP and Unix.  When a client tries to connect to a server that happens to be colocated, AppSwitch daemon would choose to connect to the Unix listening socket of the server.  The resulting Unix sockets on each end are passed into respective applications.  Once a fully connected FD is returned, the application would simply treat it as a bitpipe.  The protocol doesn’t really matter.  The application may occasionally make protocol specific calls such as `getsockname(2)` and AppSwitch would handle them in kind.  It would present consistent responses such that the application would continue to run on.
+
```
rshriram: It would be nice to compare/contrast with the recent work in the kernel that eliminated the need to go through the TCP/IP protocol stack for loopback socket communication. https://www.slideshare.net/ThomasGraf5/accelerating-envoy-and-istio-with-cilium-and-the-linux-kernel

subhraveti:
Thanks for taking the time @rshriram.  Will address them within a day or so.

linsun:
@subhraveti thanks for the blog - interesting thoughts.  Could you clarify if the proxyless approach is only targeted for the network only profile/configuration of istio?  e.g. https://preliminary.istio.io/docs/setup/kubernetes/helm-install/#customization-example-traffic-management-minimal-set

Seems a good summary would help:  what functions this can optimize for istio, and what other functions this proxyless approach can't help thus a user still needs to use the proxy to achieve it.

rshriram:
FYI the blog is not visible in the netlify staging site. I think you are missing some header.

subhraveti:
Oh that's a very old draft @linsun.  Most of those points are covered in the latest -- https://deploy-preview-1526--preliminary-istio.netlify.com/blog/2018/delayering-istio/delayering-istio/

.spelling
```diff
@@ -479,6 +479,54 @@ embeddable
...
 vCPU
 AES-NI
+3.99ms
+72.96ms
+18x
+8x
+appswitch
+appswitch-0
+AppSwitch
+bitpipe
```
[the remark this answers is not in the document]
subhraveti: Okay these are very tricky...  Looks like the general rule is to use the "closed" form for nouns, "open" form for verbs and hyphenated for adjectives: https://www.merriam-webster.com/help/faq-compound-words

.spelling
subhraveti: I guess the distinction is decapsulated is verb whereas unencapsulated is an adjective.  Also encap / decap are accepted networking terms.

.spelling
subhraveti: It's not supposed to be a proper word but I am not happy with its sound either.  I'll remove it.

content/blog/2018/delayering-istio/delayering-istio.md
subhraveti: The layers are unnecessary for the cases / reasons discussed but I'll remove reference to it.

content/blog/2018/delayering-istio/delayering-istio.md
```diff
@@ -0,0 +1,216 @@
...
+Before going into the details of how AppSwitch promises to remove unnecessary layers from the Istio stack, let me give a very brief introduction to its architecture.  Further details are available at the [documentation](https://appswitch.readthedocs.io/en/latest/) page.
+
+## AppSwitch
+
+Not unlike the container runtime, AppSwitch consists of a client and a daemon that speak REST with each other.  Both client and the daemon are built as one self-contained binary, `ax`.  Client transparently plugs into the application and tracks its system calls related to network connectivity and notifies the daemon about their occurrences.  As an example, let’s say an application makes the `connect(2)` system call to the service IP of a Kubernetes service.  AppSwitch client intercepts the connect, nullifies it and notifies the daemon about its occurrence along with some context that includes the system call arguments.  Daemon would then handle the system call, perhaps by directly connecting to the Pod IP of the upstream server on behalf of the application.
```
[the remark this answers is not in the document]
subhraveti: The details are a bit tedious but important in this particular case.  Otherwise, it's easy to assume that there is a connection between client and daemon.

content/blog/2018/delayering-istio/delayering-istio.md
subhraveti: Again, trying to avoid giving an impression that AppSwitch handles network traffic.

content/blog/2018/delayering-istio/delayering-istio.md
subhraveti: Yes that would be a good, but somewhat out of scope for this blog.  Perhaps a follow up...

linsun:
@subhraveti thanks - i could not find the blog earlier from the deploy site.   Will be reading it.

I don't quite like the burger picture -- it has shown istio too much layers than it really is. :(

subhraveti:
@linsun Yup, removed the picture.

linsun:
Reading through it, I am still looking for answers for this.

 Could you clarify if the proxyless approach is only targeted for the network only profile/configuration of istio? e.g. https://preliminary.istio.io/docs/setup/kubernetes/helm-install/#customization-example-traffic-management-minimal-set

> I think the answer is no, correct?   When you need other functions from Istio, you'll still be using the envoy proxy sidecar, in addition to the proxyless new layer added by AppSwitch.   It seems AppSwitch is mainly to be used to 1)reduce sidecar overhead, 2) support protocols not supported by Istio and 3)also preserve source headers or ips.  Sorry I'd be expecting this summarized somewhere in the net net or summary section.

In your last diagram (the performance one), is envoy sidecar even in the picture for the orange data?  I got a little confused as your diagram that titled "Socket delegation based connection redirection" still has envoy sidecar in the picture for both client and server.  Or perhaps envoy in the diagram means the envoy proxy per node?

subhraveti:
Thanks for the review @linsun.  You are exactly right, AppSwitch provides a "fast-path" for the cases it can handle and delegates the rest to an external proxy.  It makes sense to enumerate the value points as you did.  Let me make that change.

The graph shows the performance of the proxyless path (the first picture without proxy in the datapath).  The picture you refer to is the case where Envoy proxy *is* used.  In that case, AppSwitch makes connection redirection more efficient.

linsun:
Thank you @subhraveti for the clarification !

geeknoid:
@linsun Are you happy with the blog post now?

.spelling
```diff
@@ -466,6 +466,52 @@ embeddable
...
+DaemonSet
```
geeknoid: This should not be in the dictionary, put backticks around use in the doc.

.spelling
geeknoid: Remove this from the dictionary, not a word

.spelling
```diff
@@ -466,6 +466,52 @@ embeddable
...
+datapath
+decap
+decapsulated
+Devirtualization
+encap
+FDs
+fine-grain
```
geeknoid: Not a word, this should be fine-grained

.spelling
```diff
@@ -466,6 +466,52 @@ embeddable
...
+goto
+hardcoded
```
geeknoid: Use hard-coded instead

.spelling
```diff
@@ -466,6 +466,52 @@ embeddable
...
+intrahost
+iptables
+Jog
+kube-proxy
+Mandar
+memcached
+memcached.png
```
geeknoid: This should not be in the dictionary.

.spelling
```diff
@@ -466,6 +466,52 @@ embeddable
...
+memcached-2's
+MSG_PEEK
+onboarding
+Onboarding
+OS-level
+p50
+peek
+perf.png
```
geeknoid: This should not be in the dictionary

subhraveti:
Thanks @geeknoid, fixed .spelling and rebased.

PR #1530, opened 2018-06-17
Remove the limitation of accessing service mesh via browser.

gyliu513:
/cc @rshriram @geeknoid 

This is a follow up for https://github.com/istio/istio/issues/6319#issuecomment-397784031


vadimeisenbergibm:
@gyliu513 So you did you check it working in the browser? cc @rshriram @frankbu 

My understanding is that you cannot access istio-ingressgateway by IP or to specify a port, since the hosts match will fail if you specify hosts other than `*`.

PR #1567, opened 2018-06-21
Add support for snippets and file content insertion.

geeknoid:
- You can now use {{< file_content >}} to pull in files from the doc repo
into generated documentation. If you include the `snippet` attribute, you can
pull in only pieces of the file instead. This makes it possible to annotate
scripts and yaml files and extract those annotated bits and pieces into the
docs. This should let us have fully tested examples which are then incorporated
into the docs

- The previous {{< file_content >}} feature that lets you dynamically pull
content from a URL has been renamed to {{< fetch_content >}} instead.

### DIFF
--- a/content/about/contribute/writing-a-new-topic.md
+++ b/content/about/contribute/writing-a-new-topic.md
@@ -294,18 +294,44 @@ This will be rendered as:
 $ istioctl create -f @samples/bookinfo/kube/route-rule-reviews-v3.yaml@
 ```
 
+## Displaying file snippets
+
+It is often useful to display portions of a larger file. You can annotate a text file to create named snippets within the file by
+using the `$snippet` and `$endsnippet` annotations. For example, you could have a text file that looks like this:
+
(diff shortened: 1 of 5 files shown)


content/about/contribute/writing-a-new-topic.md
```diff
@@ -294,18 +294,44 @@ This will be rendered as:
 $ istioctl create -f @samples/bookinfo/kube/route-rule-reviews-v3.yaml@
 ```
 
+## Displaying file snippets
+
+It is often useful to display portions of a larger file. You can annotate a text file to create named snippets within the file by
+using the `$snippet` and `$endsnippet` annotations. For example, you could have a text file that looks like this:
+
+{{< file_content file="examples/snippet_example.txt" lang="plain" >}}
```
rshriram: This gets into the weeds of the file formats. May be its easier to start with just the YAML and say people have to add comments to YAMLs with 
```
# $startsnippet Name
some YAML
# $endsnippet Name
```

PR #1569, opened 2018-06-21
Ingress task clarifications

frankbu:


content/docs/tasks/traffic-management/ingress.md
```diff
@@ -267,12 +269,10 @@ In this subsection we add to our gateway the port 443 to handle the HTTPS traffi
...
 
-1. Access the _httpbin_ service by HTTPS, sending an HTTPS request by _curl_ to `SECURE_INGRESS_PORT`.
-Here we use _curl_'s `-k` option to instruct _curl_ not to check our certificate
-(since it is a fake certificate we created for testing the Gateway only, _curl_ is not aware of it).
+1. Access the _httpbin_ service with HTTPS by sending an https request using _curl_ to `SECURE_INGRESS_PORT`. Note the `--resolve` flag that we use this time. Unlike HTTP requests, the `Host` header that we pass will be encrypted this time, so the ingress gateway will not be able to use it to match the request to our configuration. The `--resolve` flag instructs curl to access the gateway IP address using the actual "httpbin.example.com" domain name, including the [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) protocol for our https request. Here we also use _curl_'s `-k` option to instruct _curl_ not to check our certificate (since it is a fake certificate we created for testing the Gateway only, so _curl_ is not aware of it).
```
rshriram: Instructs curl to supply the SNI value httpbin... when accessing the gateway ip over TLS

content/docs/tasks/traffic-management/ingress.md
```diff
@@ -317,11 +317,54 @@ If we want to only allow HTTPS traffic into our service mesh, we can remove the
...
     ```command
-    $ curl --resolve httpbin.example.com:$INGRESS_PORT:$INGRESS_HOST -HHost:httpbin.example.com -I http://httpbin.example.com:$INGRESS_PORT/status/200
+    $ curl -I -HHost:httpbin.example.com http://$INGRESS_HOST:$INGRESS_PORT/status/200
     ```
-## Accessing Istio service mesh by a browser
 
-For `NodePort` ingress gateways, access by a browser is supported partially: only for the `*` hosts in the definition of the `VirtualService`.
+## Accessing ingress services using a browser
+
+As you may have guessed, entering the httpbin service URL in a browser won't work because we don't have a way to tell the browser to pretend to be accessing "httpbin.example.com", like we did with _curl_. In a real world situation this wouldn't be a problem because the requested host would be properly configured and DNS resolvable, so we would simply be using its domain name in the URL (e.g., "https://httpbin.example.com/status/200").
```
geeknoid: Please use "you" instead of "we" throughout this doc/PR.

Thanks.

content/docs/tasks/traffic-management/ingress.md
```diff
@@ -282,6 +284,7 @@ In this subsection we add to our gateway the port 443 to handle the HTTPS traffi
     content-length: 0
     x-envoy-upstream-service-time: 6
     ```
+    Note the `--resolve` flag that we use this time. Unlike HTTP requests, the `Host` header that we pass will be encrypted this time, so the ingress gateway will not be able to use it to match the request to our configuration. The `--resolve` flag instructs _curl_ to supply the [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) value "httpbin.example.com" when accessing the gateway IP over TLS. Here we also use _curl_'s `-k` option to instruct _curl_ not to check our certificate (since it is a fake certificate we created for testing the Gateway only, so _curl_ is not aware of it).
 
     > Note that it may take time for the new gateway definition to propagate and you may get the following error: `Failed to connect to httpbin.example.com port <your secure port>: Connection refused`. Wait for a minute and retry the `curl` call again.
```
geeknoid: Remove the "Note that" part.

content/docs/tasks/traffic-management/ingress.md
geeknoid: OK.

PR #1580, opened 2018-06-23
Improve handling for code blocks

geeknoid:
- Add the {{< text >}} shortcode as a way to declare code blocks. This shortcode
is intended to hide the subtleties around code blocks. More specifically, it
provides a transparent workaround for the Hugo bug around converting indented
code blocks where lines start with dashes. It also deal with the special
formatting case we have around command-lines that start with `cat <<EOF`.

- Rename the {{< file_content >}} shortcode to {{< text_file >}} and the
{{< fetch_content >}} shortcode to {{< text_dynamic >}} for consistency in naming.


frankbu:
Does this replace the triple back-ticks or is it an alternative? Recommended to use for all formatted text or just some? For example is
```yaml 
stuff
```
still the way to add yaml?

Seems that this will screw up the preview on github, which sort of works reasonably well with the old backtick style.

geeknoid:
My intent is to use this for everything so people don't have to deal with
the underlying bugs, and don't have to think about it. Yes, it would mess
up the plain GitHub preview, but we have the full fidelity Netlify preview
instead.

One problem I still have to work around is the fact our spell checking and
style checking explicitly understand tick marks and treat the content
accordingly. So right now, these tools get pretty confused with this new
approach. I'm going to need to do some data transformation before invoking
those tools.

PR #1596, opened 2018-06-27
Adding myself as OWNERS for Chinese Content.

gyliu513:
/cc @geeknoid @linsun


rootsongjc:
We really need some owners to accelerate the Chinese translating.

fleeto:
The istio Chinese community has translated all documents except the Reference section into Chinese. If the process can be accelerated, we can quickly merge existing content into official repository.

PR #1600, opened 2018-06-27
[WIP] Update authn policy tasks with global policy

diemtvu:
- Add section to use mesh-wide policy to enable mTLS globally.
- Update examples to follow naming restriction.

### DIFF
--- a/content/docs/tasks/security/authn-policy.md
+++ b/content/docs/tasks/security/authn-policy.md
@@ -19,7 +19,7 @@ Through this task, you will learn how to:
 
 * Have a Kubernetes cluster with Istio installed, without global mutual TLS enabled (e.g use `install/kubernetes/istio-demo.yaml` as described in [installation steps](/docs/setup/kubernetes/quick-start/#installation-steps), or set `global.mtls.enabled` to false using [Helm](/docs/setup/kubernetes/helm-install/)).
 
-*   For demo, create two namespaces `foo` and `bar`, and deploy [httpbin](https://github.com/istio/istio/blob/{{<branch_name>}}/samples/httpbin) and [sleep](https://github.com/istio/istio/tree/master/samples/sleep) with sidecar on both of them. Also, run another sleep app without sidecar (to keep it separate, run it in `legacy` namespace)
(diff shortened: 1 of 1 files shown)


geeknoid:
Hi, I did some refactoring and the file you edited has moved in the repo. Please rebase and push again. Sorry for the inconvenience.

content/docs/tasks/security/authn-policy.md
```diff
@@ -44,55 +45,198 @@ Through this task, you will learn how to:
...
+    tls:
+      mode: ISTIO_MUTUAL
+EOF
+```
+
+> * Mesh-wide authentication policy name must be `default`; any policy with other name will be rejected and ignored. Also note that the CRD kind is `MeshPolicy`, which is different than the namespace-wide or service-specific policy kind (`Policy`)
+> * On the other hand, the destination rule can have any name, and in any namespace. For consistency, we use also name it `default` and keep in `default` namespace in this demo.
+> * Host value `*.local` in destination rule matches only services in the mesh, which have `local` suffix.
+> * With `ISTIO_MUTUAL` TLS mode, Istio will set the path for key and certificates (e.g `clientCertificate`, `privateKey` and `caCertificates`) according to its internal implementation.
+> * If you want to define a destination rule for a specific server, the TLS settings must be copied over to the new rule.
```
liminw: server -> service?

content/docs/tasks/security/authn-policy.md
```diff
@@ -102,41 +246,39 @@ spec:
 EOF
 ```
 
-> Note:
-* This rule is based on the assumption that there is no other destination rule in the system. If it's not the case, you need to modify traffic policy in existing rules accordingly.
-* `*.foo.svc.cluster.local` matches all services in namespace `foo`.
-* With `ISTIO_MUTUAL` TLS mode, Istio will set the path for key and certificates (e.g `clientCertificate`, `privateKey` and `caCertificates`) according to its internal implementation.
+> Host `*.foo.svc.cluster.local` limit sthe matches to services in `foo` namespace only.
```
liminw: typo: sthe

content/docs/tasks/security/authn-policy.md
```diff
@@ -156,15 +298,17 @@ spec:
 EOF
 ```
 
-Again, run the probing command. As expected, request from `sleep.legacy` to `httpbin.bar` starts failing with the same reasons.
+> In this example, we do **not** specify namespace in metadata but put it in the command line (`-n bar`). They should work the same.
+>  There is not restriction on the authentication policy and destination rule name. The example use the name of the service itself for simplicity.
```
liminw: not restriction => no restriction

diemtvu: Done

content/docs/tasks/security/authn-policy.md
diemtvu: Done

content/docs/tasks/security/authn-policy.md
diemtvu: Done

PR #1605, opened 2018-06-27
[Merge after 0.8.1] Add egress gateway instructions for Istio with mTLS

vadimeisenbergibm:


content/docs/tasks/traffic-management/egress-gateway.md
```diff
@@ -51,7 +51,35 @@ If we used the [sleep](https://github.com/istio/istio/tree/{{<branch_name>}}/sam
 
 First let's direct HTTP traffic without TLS origination
 
-1.  Create an egress `Gateway` for _edition.cnn.com_, port 80:
+1.  Create an egress `Gateway` for _edition.cnn.com_, port 80.
+
+    If you have [mutual TLS Authentication](/docs/tasks/security/mutual-tls/) enabled in Istio, you must create the
+    `Gateway` in the following way:
```
frankbu: You should explain why you must create it this way, esp. why the host is "*" vs. "edition.cnn.com" in the non mTLS case.

content/docs/tasks/traffic-management/egress-gateway.md
```diff
@@ -126,6 +154,28 @@ the traffic through the egress gateway:
         EOF
     ```
 
+    In addition, if you have [mutual TLS Authentication](/docs/tasks/security/mutual-tls/) enabled in Istio, you must
+    create a destination rule to enable mutual TLS between the sidecar proxies and the egress gateway:
```
frankbu: This could use more explanation too. What is the default tls for incoming requests to egress gateway? I would think it should be ISTIO_MUTUAL if mTLS is globally enabled.

geeknoid:
Deploy preview for *istio* failed.

Built with commit 1d1be1cb865bad0ad5367a476cc8c64dce069348

https://app.netlify.com/sites/istio/deploys/5b33a44d67610c2a2cda44d4

content/docs/tasks/traffic-management/egress-gateway.md
vadimeisenbergibm: @frankbu I think I will just change the hosts to '*' in both cases since for HTTP it does not matter anyway. We may want to rethink this configuration later.

content/docs/tasks/traffic-management/egress-gateway.md
vadimeisenbergibm: @frankbu Good point, it actually is not needed, I was confused. Removing the destination rules for the egress gateway.

vadimeisenbergibm:
No 0.8.1 is expected, will submit a corresponding PR to the master.

PR #1607, opened 2018-06-27
update URLs which have been moved to other places

rootsongjc:
update URLs which have been moved to other places


rootsongjc:
@gyliu513 CI errors fixed.

PR #1614, opened 2018-06-28
Updated image link for overview.

gyliu513:
/cc @rootsongjc @fleeto


content_zh/docs/concepts/what-is-istio/overview/index.md
```diff
@@ -38,7 +38,7 @@ Istio服务网格逻辑上分为**数据面板**和**控制面板**。
 下图显示了构成每个面板的不同组件：
 
 {{</* image width="80%" ratio="56.25%"
```
geeknoid: You should also remove the /* and */ to enable the image

gyliu513: Done

PR #1638, opened 2018-06-29
Fix for encoding errors on commands index page.

bdurette:
The apostrophes in the descriptions of the commands are incorrectly encoded on the commands index page:

istio___commands

This appears to be because HTML encoding is used in the header of the nested files. The fix is to remove the encoding in the non-HTML portion of the files.


geeknoid:
These are two auto-generated files, so the next time the auto-generation process is run these changes will be lost. The fix needs to happen in the istio/istio repo, in the pkg/collateral package.

So I'm going to close this PR and then go put the fix in the collateral package.

Thanks.

PR #1655, opened 2018-07-02
translate performance-and-scalability/overview into Chinese

xiaosuiba:
[original code](https://github.com/istio/istio.github.io/tree/master/content/docs/performance-and-scalability/overview/index.md)
[original page](https://preliminary.istio.io/docs/performance-and-scalability/overview)

### DIFF
--- a/content_zh/docs/performance-and-scalability/overview/index.md
+++ b/content_zh/docs/performance-and-scalability/overview/index.md
@@ -1,0 +1,17 @@
+---
+title: 综述
+description: 提供 Istio 性能和可伸缩性概念的介绍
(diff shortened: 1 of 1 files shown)


content_zh/docs/performance-and-scalability/overview/index.md
```diff
@@ -0,0 +1,17 @@
+---
+title: 综述
+description: 提供 Istio 性能和可伸缩性概念的介绍
```
gyliu513: 提供 Istio 性能和规模测试的介绍

其他的也一样改一下

PR #1661, opened 2018-07-02
edit reviews virtualservice yaml output

arshchimni:
**Where is the problem?**

_https://istio.io/docs/tasks/traffic-management/traffic-shifting/#weight-based-version-routing_

**What is the problem?**

As shown in step 3, the below configuration is invalid.  The extra **`route`** attribute in the review virtualservice yaml output breaks the functionality which the example represents, as shown below:

```
$ istioctl get virtualservice reviews -o yaml
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: reviews
  ...
spec:
  hosts:
  - reviews
  http:
  - route:
    - destination:
        host: reviews
        subset: v1
      weight: 50
  - route: # invalid configuration 
    - destination:
        host: reviews
        subset: v3
      weight: 50
```

The github sample yaml in the url _https://raw.githubusercontent.com/istio/istio/release-0.8/samples/bookinfo/routing/route-rule-reviews-50-v3.yaml_  in the mentioned command 

`istioctl replace -f samples/bookinfo/routing/route-rule-reviews-50-v3.yaml`

is correct and works as expected. 

**Correction made as follows:**

```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: reviews
  ...
spec:
  hosts:
  - reviews
  http:
  - route:
    - destination:
        host: reviews
        subset: v1
      weight: 50
    - destination:
        host: reviews
        subset: v3
      weight: 50
```

### DIFF
--- a/content/docs/tasks/traffic-management/traffic-shifting/index.md
+++ b/content/docs/tasks/traffic-management/traffic-shifting/index.md
@@ -59,7 +59,6 @@ two steps: 50%, 100%.
             host: reviews
             subset: v1
           weight: 50
-      - route:
         - destination:
             host: reviews
             subset: v3


geeknoid:
@frankbu Is this a case where we would use the new {{< text_file >}} feature to pull the YAML file directly from GitHub when the site is being built? Or use the @@ annotation in the command-line to just use a link to the file on GitHub instead of inserting the content inline?

See https://preliminary.istio.io/about/contribute/writing-a-new-topic/#embedding-preformatted-blocks for info on available options for preformatted blocks.

PR #1664, opened 2018-07-03
Translate performance-and-scalability/scenarios into Chinese

xiaosuiba:
[original code](https://github.com/istio/istio.github.io/tree/master/content/docs/performance-and-scalability/scenarios/index.md)
[original page](https://preliminary.istio.io/docs/performance-and-scalability/scenarios)

### DIFF
--- a/content_zh/docs/performance-and-scalability/scenarios/index.md
+++ b/content_zh/docs/performance-and-scalability/scenarios/index.md
@@ -1,0 +1,15 @@
+---
+title: 测试场景
+description: 我们正在跟踪的性能和可伸缩性的不同场景。
+weight: 25
+---
+
+{{< image width="100%" ratio="75%"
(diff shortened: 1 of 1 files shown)


content_zh/docs/performance-and-scalability/scenarios/index.md
```diff
@@ -0,0 +1,15 @@
+---
+title: 测试场景
+description: 我们正在跟踪的性能和可伸缩性的不同场景。
```
gyliu513: 我建议把”可伸缩性“改为”规模性测试”

gyliu513: after some discussion in wechat group, we agree to use "可伸缩性"

xiaosuiba: ok, I see it.

PR #1668, opened 2018-07-03
Adding rootsongjc and fleeto to OWNERS for Chinese.

gyliu513:
Thanks @rootsongjc and @fleeto for the great contribution in translating Istio from 
English to Chinese.

/cc @geeknoid


rootsongjc:
There are too many Chinese PR blocked, as I become an owner we can speed it up.

PR #1675, opened 2018-07-04
make istio installation yaml file consistent

LillyWu:


LillyWu:
For issue: https://github.com/istio/istio.github.io/issues/1672

PR #1679, opened 2018-07-04
Automate and clarify determining ingress gateway's IP and ports

vadimeisenbergibm:
Check load balancer's ip automatically, print more information, direct the user how to proceed

The current ingress gateway instructions are tricky and may cause confusion. The proposed instructions are still rather tricky, careful review is required.

### DIFF
--- a/content/docs/examples/bookinfo/index.md
+++ b/content/docs/examples/bookinfo/index.md
@@ -138,7 +138,7 @@ To start the application, follow the instructions below corresponding to your Is
 
 #### Determining the ingress IP and port
 
-1.  Follow [these instructions](/docs/tasks/traffic-management/ingress/#determining-the-ingress-ip-and-ports) to set the `INGRESS_HOST` and `INGRESS_PORT` variables.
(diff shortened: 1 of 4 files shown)


frankbu:
I don't think this is better. The current instructions seem pretty straightforward to me. We're telling the users to first check if there's a load balancer or not (if they don't already know) and then proceed based on that.

vadimeisenbergibm:
@frankbu I am OK with leaving the task AS IS. I have two concerns:
1. We are telling the users to check if the service is of the `LoadBalancer` type or `NodePort`. However, we can do it automatically, by a bash command, reducing user's involvement/preventing mistakes.
2. I am afraid some users will be confused and will not get the fact that the instructions for `LoadBalancer` and `NodePort` are exclusive, either or. The users may perform the instructions sequentially: first `LoadBalancer` then `NodePort`, which will be wrong for `LoadBalancer`.

frankbu:
For #1, I think many users will know whether their env supports loadbalancer or not, so they won't even need to do the manual step. For other users, I don't think the manual step is bad because it helps them understand what they're doing, which is a good thing for a learning exercise.

For #2, let me take a pass at wording things a little better, so it's clear that users should only do one or the other.

frankbu:
Fixed in #1835

PR #1680, opened 2018-07-04
Add download & print buttons to all PRE blocks.

geeknoid:
Also, replaced the text in the Copy button with an icon instead.


rshriram:
The download button ends up downloading yaml snippets as bash. And then chrome complains.

geeknoid:
@rshriram Any idea which yaml sample you clicked on specifically? The few I just tried downloaded as .yaml correctly.

PR #1685, opened 2018-07-05
Chineses: update content

rootsongjc:
- update policies and telemetry translation
- delete the stale files
- merge two docs together


rootsongjc:
@geeknoid all the Chinese PR are blocked, how do we deal with these?

PR #1699, opened 2018-07-07
Move Helm customization section to the references category

ymesika:
The auto-generated table for Helm customization values became too long for this page imho.
This is a suggestion to move the customization stuff to the references.

Two options I had in mind:
1. Move both the table and the minimal-set customization example to the references (current PR)
1. Move just the table and keep the example section in the main page

`scripts/tablegen.py` which updates the table from the `values.yaml` wasn't updated because it's output destination is unclear (currently `helm-install.md` which doesn't exist). cc @sdake


content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
+---
```
geeknoid: Is "helm customization" the right name for this file? This is customizing the installation options, not customizing helm, right? Perhaps "installation-options" would be a better name?

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
...
+title: Customise the Helm Installation
```
geeknoid: Similar to my comment above, this probably should be "Installation Options"

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
...
+description: Customise the Istio installation with the included Helm chart.
+weight: 30
+keywords: [kubernetes,helm]
+aliases:
+    - /docs/reference/config/helm/index.html
```
geeknoid: Only add aliases when moving pages. This is a new page, so it shouldn't have an alias.

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
+---
+title: Customise the Helm Installation
+description: Customise the Istio installation with the included Helm chart.
```
geeknoid: Custonise -> Customize

Probably better as:

"Describes the options available when installing Istio using the included Helm chart.

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
...
+weight: 30
+keywords: [kubernetes,helm]
+aliases:
+    - /docs/reference/config/helm/index.html
+---
+
+The Helm chart ships with reasonable defaults.  There may be circumstances in which defaults require overrides.
```
geeknoid: You can install Istio using Helm (provide link to install page), and although the included Helm chart uses reasonable defaults, there are circumstances that require customization. You can use Helm's `--set key=value` option to override one or more values. The set of supported keys is shown in the table below.  

The Istio helm chart includes reasonable default.

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
...
+To override Helm values, use `--set key=value` argument during the `helm install` command.  Multiple `--set` operations
+may be used in the same Helm operation.
+
+Helm charts expose configuration options which are currently in alpha.  The currently exposed options are explained in the
```
geeknoid: I've remove this paragraph.

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
...
+following table:
+
+<!-- AUTO-GENERATED-START -->
+| Parameter | Default | Description | Values |
```
geeknoid: Given the Helm command-line says "--set key=value", then should "Parameter" be called "Key" instead?

It doesn't seem as though the Values column should be using `` on the listed values. In fact, looking at the generated table, it doesn't appear as though the Values column is useful at all. There's a Description column which seems sufficient.

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
...
+| `kiali.dashboard.password` | `admin` |  | `` |
+<!-- AUTO-GENERATED-END -->
+
+The Helm chart also offers significant customization options per individual
+service. Customize these per-service options at your own risk. The per-service options are exposed via
+the [`values.yaml`]({{< github_file >}}/install/kubernetes/helm/istio/values.yaml) file.
+
+## Customization example: traffic management minimal set
+
+Istio is equipped with a rich and powerful set of features and some users may need only subset of those. For instance, users might be interested only
```
geeknoid: Istio has a rich feature set, but you may only want to use a subset of these. For instance, you might be only interested in installing the minimum necessary to support traffic management functionality.

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
...
+in installing the minimal set required to Istio's traffic management.
+Helm customization provides the option to install a subset by enabling those features of interest and disabling the ones that aren't required.
+
+In this example we will install Istio with only a minimal set of components necessary to conduct [traffic management](/docs/tasks/traffic-management/).
```
geeknoid: Generally, avoid using "we". Just say

This example shows how to install the minimal set of components necessary to use traffic management features

content/docs/reference/config/helm-customization/index.md
```diff
@@ -0,0 +1,283 @@
...
+NAME                                     READY     STATUS    RESTARTS   AGE
+istio-citadel-b48446f79-wd4tk            1/1       Running   0          1m
+istio-ingressgateway-7b77d995f7-t6ssx    1/1       Running   0          1m
+istio-pilot-58c65f74bc-2f5xn             2/2       Running   0          1m
+istio-sidecar-injector-86cc99578-4t58m   1/1       Running   0          1m
+{{< /text >}}
+
+With this minimal set you can proceed to installing the sample [Bookinfo](/docs/examples/bookinfo/) application or install your own application and [configure request routing](/docs/tasks/traffic-management/request-routing/) for instance.
+
+Of course that if no ingress is expected and sidecar is to be [injected manually](/docs/setup/kubernetes/sidecar-injection/#manual-sidecar-injection) then you can reduce this minimal set even further and only have Pilot and Citadel. However, Pilot depends on Citadel therefore you can't install it without the other.
```
geeknoid: If the minimal set can be further reduced, then it wasn't really the minimal set, right? Maybe you shold actually show the truly minimal set for traffic management?

content/docs/reference/config/helm-customization/index.md
ymesika: "The Istio helm chart includes reasonable default." is already mentioned in the first sentence so I suggest have only your first paragraph.

content/docs/reference/config/helm-customization/index.md
ymesika: Will update the python script as well

geeknoid:
Looks good, thanks. I'll let @sdake give it the final green light.

content/docs/reference/config/installation-options/index.md
```diff
@@ -0,0 +1,250 @@
...
+<!-- AUTO-GENERATED-END -->
+
+The Helm chart also offers significant customization options per individual
+service. Customize these per-service options at your own risk. The per-service options are exposed via
+the [`values.yaml`]({{< github_file >}}/install/kubernetes/helm/istio/values.yaml) file.
+
+## Example: traffic management minimal set
+
+Istio has a rich feature set, but you may only want to use a subset of these. For instance, you might be only interested in installing the minimum necessary to support traffic management functionality.
+
```
rshriram: This is burying the lead :). If this is meant to serve as an example of a la carte Istio, this should go into main docs, as its own thing: essentially, something like users wishing to play with Istio's networking features can start with a minimal Istio install consisting of Pilot and an Ingress Gateway, as shown below:..

content/docs/reference/config/installation-options/index.md
```diff
@@ -0,0 +1,250 @@
...
+This example shows how to install the minimal set of components necessary to use [traffic management](/docs/tasks/traffic-management/) features.
+
+Execute the following command to install the Pilot and Citadel:
+
+{{< text bash >}}
+$ helm install install/kubernetes/helm/istio --name istio --namespace istio-system \
+  --set ingress.enabled=false,gateways.istio-egressgateway.enabled=false,galley.enabled=false \
+  --set gateways.istio-ingressgateway.enabled=false,sidecarInjectorWebhook.enabled=false \
```
rshriram: You set ingress gateway to false. However, bookinfo requires gateway.

ymesika: Like @geeknoid commented in a previous review if ingress gateway is there then it isn't really a minimal set. Two options:
1. Remove the "installing the sample Bookinfo application or" part of the sentence
1. Add ingress gateway and change the title to Bookinfo minimal set

content/docs/reference/config/installation-options/index.md
ymesika: Should we move this example back to the helm setup page?

content/docs/reference/config/installation-options/index.md
```diff
@@ -0,0 +1,250 @@
...
+| `kiali.ingress.annotations` |  |  |
+| `kiali.ingress.tls` |  |  |
+| `kiali.dashboard` |  |  |
+| `kiali.dashboard.username` | `admin` |  |
+| `kiali.dashboard.password` | `admin` |  |
+<!-- AUTO-GENERATED-END -->
+
+The Helm chart also offers significant customization options per individual
+service. Customize these per-service options at your own risk. The per-service options are exposed via
+the [`values.yaml`]({{< github_file >}}/install/kubernetes/helm/istio/values.yaml) file.
```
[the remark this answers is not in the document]
ymesika: I suggest to remove this paragraph as there are no additional options when using the script to extract all into the table.

content/docs/reference/config/installation-options/index.md
sdake: yes please move this back to the main page.

content/docs/reference/config/installation-options/index.md
sdake: Would you consider instead of calling this the "minimal set" - "Running Istio without Mixer"?  For performance reasons, folks may not want to run with Mixer as it has a very large impact on QPS.

Another option is to have a minimal set and then add the "run without mixer"

Also, please use the syntax:
helm install install/kubernetes/helm/istio --name istio --namespace istio-system \
 --set ingress.enabled=false
--set gateways.istioegressgateway.enabled=false
--set galley.enabled=false \
--set gateways.istio-ingressgateway.enabled=false \
--set sidecarInjectorWebhook.enabled=false

This will render in a more readable format.

sdake:
@ymesika I will take care to update the autogenerator.  It needs a little more work anyway.

Cheers
-steve

sdake:
@gyliu513 can you fix the py autogenerator to remove the keys that are part of a list?

(Look at PR to see the errors produced by the autogenerator).

Cheers
-steve

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -86,282 +86,38 @@ to manage the lifecycle of Istio.
...
 
 {{< text bash >}}
 $ helm install install/kubernetes/helm/istio --name istio --namespace istio-system \
-  --set ingress.enabled=false,gateways.istio-egressgateway.enabled=false,galley.enabled=false \
-  --set mixer.enabled=false,prometheus.enabled=false,global.proxy.envoyStatsd.enabled=false
+  --set ingress.enabled=false \
+  --set gateways.istio-ingressgateway.enabled=false \
+  --set gateways.istio-egressgateway.enabled=false \
+  --set galley.enabled=false \
+  --set sidecarInjectorWebhook.enabled=false \
```
sdake: Rather than al la carte as the name, I'd use

"Installing Istio without Mixer" and show how that is done (meaning the webhook should be enabled).

Or another option is to just add that mode in.

I think in general we should provide 5-7 examples to create istio helm charts  that do meaningful things for people.

My main attraction to this model is to enable things like "How do I install Istio without Mixer?".  I'm sure folks will come up with other ideas as time passes.

Cheers
-steve

PR #1710, opened 2018-07-09
Put back reverted new troubleshooting material

frankbu:


geeknoid:
I think to fix the circleci failures, you need to make sure all the {{<text>}} and {{</text>}} sequences have a space between the {} and the <>. One/some of my regexes may be assuming this.

PR #1722, opened 2018-07-10
principal attributes

kyessenov:
/cc @douglas-reid @geeknoid @liminw 
Is this a better language to describe these attributes? Addresses two points:
- there is no request for TCP
- we don't have a well defined "application" concept

Signed-off-by: Kuat Yessenov <kuat@google.com>


geeknoid:
We really should be generating this stuff from the source manifest, just like we started to generate the description of all the Helm install options from the source yaml.

PR #1744, opened 2018-07-11
Fix the non-existent file in the test description of Citadel health checking

lei-tang:
Per the problem reported in https://github.com/istio/istio/issues/6922,  this PR fixes the non-existent file in the description of Citadel health checking in https://preliminary.istio.io/docs/tasks/security/health-check.html.
This PR depends on the security.healthCheckEnabled flag in the PR https://github.com/istio/istio/pull/6961, so it should be committed after the PR 6961 is committed.


sdake:
@lei-tang bad news - all the bookinfo files in the repository have been moved around.  Please let me know if you need help to fix these links to their proper location.

Cheers
-steve

PR #1746, opened 2018-07-11
add instruction to label the default ns for bookinfo example

rokii:
resolving this https://github.com/istio/istio/issues/6957

### DIFF
--- a/content/docs/examples/bookinfo/index.md
+++ b/content/docs/examples/bookinfo/index.md
@@ -93,8 +93,19 @@ To start the application, follow the instructions below corresponding to your Is
 
     *   If you are using a cluster with
         [automatic sidecar injection](/docs/setup/kubernetes/sidecar-injection/#automatic-sidecar-injection)
-        enabled, simply deploy the services using `kubectl`
+        enabled, label the `default` namespace with `istio-injection=enabled`
 
+        {{< text bash >}}
+        $ kubectl label namespace default istio-injection=enabled
+        $ kubectl get namespace -L istio-injection
+        NAME           STATUS    AGE       ISTIO-INJECTION
+        default        Active    1h        enabled
+        istio-system   Active    1h
+        kube-public    Active    1h
+        kube-system    Active    1h
+        {{< /text >}}
+
+        Then simply deploy the services using `kubectl`
         {{< text bash >}}
         $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo.yaml@
         {{< /text >}}


content/docs/examples/bookinfo/index.md
```diff
@@ -93,8 +93,19 @@ To start the application, follow the instructions below corresponding to your Is
 
     *   If you are using a cluster with
```
sdake: @rokii thanks for the contribution!  I'd prefer to just remove the automatic sidecar injection documentation almost entirely (except for the sidecar document itself) as the TOC has decided that sidecar injection will be a mandatory requirement.  Can you rework this document with that in mind?

Cheers
-steve

sdake: @frankbu see https://groups.google.com/forum/#!topic/istio-dev/ZdMbghc7sDE.  Perhaps I read too much into that, and your position is fair.  I just believe we shouldn't document manual injection in every single example page and should lean on automatic injection for most cases.

I do believe we should continue to document manual injection in the main automatic sidecar injection page, as it is very helpful.

Cheers
-steve

rokii:
@sdake I have no idea why this pr cause the lint check fails.
```
- ./public/docs/examples/bookinfo/index.html
  *  7:359: ERROR: Unexpected end tag : p (line 7)
htmlproofer 3.9.1 | Error:  HTML-Proofer found 1 failure!
LINTING FAILED
```
Could you help merge or fix this?

frankbu:
@geeknoid I think that either the html proofer is getting confused, or there's a a problem transforming command output with two commands plus output. Either way, it seems to display correctly. Any idea which it is?

geeknoid:
This is unfortunately a bug in the markdown converter that happens once in a while. The fix is simple. Just wrap the whole text block with a div. Please search the .md files for the few examples of this and use the same class name as those divs.

geeknoid:
BTW, when this happens, the generated HTML is incorrect. Whether it displays properly or not is up to the luck of the draw.

content/docs/examples/bookinfo/index.md
```diff
@@ -93,8 +93,19 @@ To start the application, follow the instructions below corresponding to your Is
...
+        {{< text bash >}}
+        $ kubectl label namespace default istio-injection=enabled
+        $ kubectl get namespace -L istio-injection
+        NAME           STATUS    AGE       ISTIO-INJECTION
+        default        Active    1h        enabled
+        istio-system   Active    1h
+        kube-public    Active    1h
+        kube-system    Active    1h
+        {{< /text >}}
+
```
frankbu: @rokii Please wrap your text block in a div like this:
```
<div class="workaround_for_hugo_bug">

{{< text bash >}}
...
{{< /text >}}

</div>
```

frankbu:
Merging and will fix hugo issue in followup PR

PR #1748, opened 2018-07-11
First round of edits.

navinger:
Replace "we" with "you" throughout. Use a numbered list in "Before you begin" and "Cleanup" because the order matters. Added info about the 3 versions in the Bookinfo sample. Added info about setting the default version before the command. Added a space between the number and 'qps' throughout.


geeknoid:
The circleci tests indicate:

content/docs/tasks/policy-enforcement/rate-limiting/index.md:25: MD009 Trailing spaces
content/docs/tasks/policy-enforcement/rate-limiting/index.md:263: MD029 Ordered list item prefix
content/docs/tasks/policy-enforcement/rate-limiting/index.md:269: MD029 Ordered list item prefix

which you'll need to fix. The broken links however aren't related to your changes and are being addressed separately.

PR #1751, opened 2018-07-11
Updating 3 diagrams

oaktowner:
Pilot Adapters, Traffic Management Overview, and Service Model Versions.


geeknoid:
In the what-is-istio diagram, at the bottom it says "config data to envoys" and "certs to Envoys", but the boxes above are labeled "proxy". I think we should label the boxes as "sidecar" and change the text below to use sidecar as well.

In the Pilot Architecture diagram, a line overwrites the third Envoy box at the bottom.

Remove the string "Service Version" at the bottom of that diagram, there is already a caption inserted by the web site.

rshriram:
I think we should have Envoys uniformly or sidecars uniformly. Also, if using envoys, the hexagon is good. But if not using Envoys, we should skip the envoy logo

oaktowner:
I agree -- we should probably refer to them as "proxies" rather than "Envoys". I think I prefer "proxy" to "sidecar" because the proxy isn't *always* a sidecar -- in ingress/egress situations, for example.

Aesthetically, I _do_ like the Envoy hexagon (and in complicated diagrams like Traffic Management Overview, that _greatly_ reduces the complication of the diagram.

So would you mind if I called them "proxies" but dropped the hexagons?

geeknoid:
In the first two traffic mgmt diagrams and fourth diagram, the spacing between the SvcA box and its proxy is different than the spacing between the proxy and Pod1/2/3 or SvcB

In those same diagram, we should use consistent names for the pods. They're called SvcA, SvcB, or Pod1/2/3. Maybe something like PodA-1, PodA-2, PodB-3 would work best for all diagrams.

PR #1757, opened 2018-07-11
Document to use flag security.selfSigned instead

axel7born:
of no longer existing template file.


frankbu:
In Istio 0.8 `install/kubernetes/istio-auth.yaml` is renamed `install/kubernetes/istio-demo-auth.yaml`. Assuming that is not going change in 1.0, we should leave both options in the doc.

loewenstein:
@frankbu Given that step 2 (Redeploy Citadel) uses Helm anyway, I am not sure if providing two options in the preparation section provides more worth or confusion.

frankbu:
Good point @loewenstein.

@geeknoid do you know who wrote this task and should probably review this change before we merge it?

frankbu:
fixes #1836

PR #1762, opened 2018-07-11
Edit Kubernetes Quick Start for clarity.

rcaballeromx:
This edit includes among other changes:

* Consistent use of markup, line length, and command formatting.
* Grammar, spelling, and other language fixes.
* Use of ordered lists for steps.
* New clear headings for installation options.

### DIFF
--- a/content/docs/setup/kubernetes/quick-start/index.md
+++ b/content/docs/setup/kubernetes/quick-start/index.md
@@ -36,17 +36,17 @@ The following instructions require:
 
   On Kubernetes **1.9**:
 
-  {{< text bash >}}
-  $ minikube start --memory=4096 --kubernetes-version=v1.9.4 \
-    --vm-driver=`your_vm_driver_choice`
-  {{< /text >}}
+    {{< text bash >}}
+    $ minikube start --memory=4096 --kubernetes-version=v1.9.4 \
+      --vm-driver=`your_vm_driver_choice`
+    {{< /text >}}
 
   On Kubernetes **1.10**:
 
-  {{< text bash >}}
-  $ minikube start --memory=4096 --kubernetes-version=v1.10.0 \
-    --vm-driver=`your_vm_driver_choice`
-  {{< /text >}}
+    {{< text bash >}}
+    $ minikube start --memory=4096 --kubernetes-version=v1.10.0 \
+      --vm-driver=`your_vm_driver_choice`
+    {{< /text >}}
 
 ### Google Kubernetes Engine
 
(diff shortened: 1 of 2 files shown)


geeknoid:

Built with commit b37c6a4916fb0975de55592279c64d679b7a273b

https://app.netlify.com/sites/preliminary-istio/deploys/5b46536edd28ef7796a6bc78

rcaballeromx:
This is my first pass at this document. However, the largest issue is the amount of information about requirements before the actual installation section. I want to create a separate PR where I create a separate document for the prerequisites and link to it. That change would provide a much better experience. @geeknoid Any thoughts?

content/docs/setup/kubernetes/quick-start/index.md
```diff
@@ -1,366 +1,463 @@
...
 
-To install Istio locally, install the latest version of
-[Minikube](https://kubernetes.io/docs/setup/minikube/) (version 0.28.0 or later).
+1. To install Istio locally, install the latest version of
+   [Minikube](https://kubernetes.io/docs/setup/minikube/), version **0.28.0 or
+   later**.
 
-Select a [VM driver](https://kubernetes.io/docs/tasks/tools/install-minikube/#install-a-hypervisor)
-and substitute the below `your_vm_driver_choice` with the installed VM driver.
+1. Select a [VM
```
sdake: this should be one line - I suspect the cause of liting problem

rcaballeromx: Thank you @sdake but that is not one of the issues the linter flags. There are several other links that are split like that due to line length but it does not flag any of them. 
My question there is, what should take precedence? Trying to keep the line length or keeping links on a single line?

geeknoid:
You are not allowed to break an URL on two lines, so that's the cause of the warnings.

Also note that if you look in the netlify preview, you'll see that many of the code blocks aren't rendered correctly and display an error. This is because you changed the indent in code blocks. The indent of the content of {{< text >}} blocks should align with the {{< text >}} sequence.

rcaballeromx:
@geeknoid I fixed the code blocks as you described but there are four that still won't render.
I also made sure all links were in a single line but the linter still complains. I will try removing the in-line markup within the link. I suspect that is what causes the flag. Do you know why the code blocks won't render even if they are on the same level as the tags?

geeknoid:
Code blocks must always be indented by multiples of 4 spaces, that's probably what the issue is.

rcaballeromx:
I finally got the linter to pass but there are two code blocks that are still not rendering correctly. Any suggestions on what the issue is?

PR #1799, opened 2018-07-13
Rewrite Rule configuration intro and some minor copy edits.

sgilsonapigee:
Signed-off-by: Stephen Gilson <gilsonsm@google.com>

### DIFF
--- a/content/docs/concepts/traffic-management/index.md
+++ b/content/docs/concepts/traffic-management/index.md
@@ -19,7 +19,7 @@ principles. It assumes that you've already read [What is Istio?](/docs/concepts/
 and are familiar with Istio's high-level architecture.
 
 Using Istio's traffic management model essentially decouples traffic flow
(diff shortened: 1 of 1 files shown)


sgilsonapigee:
@sdake "operator" doesn't convey that it is the person using/configuring/administrating Istio that is doing the work. It implies some other entity is responsible for doing the configuration. Also, from the Istio style guide, the preference is to address the reader as "you" in the doc, which is a standard convention.

-Stephen

geeknoid:
We've had guidelines for ever about using "you" instead of "we" throughout the site. Our tech writing friends have been purging uses of "we" they find as they go through the site.

"Operators" and "Users" fall in the same bucket IMO. We are addressing the text to the reader, they are the operators. So it seems reasonable to use "you" for this case as well.

Of course, I'm not a tech writer, and am not a native English speaker. So what do I know :-) But that's the motivation.

PR #1804, opened 2018-07-13
Add a platform setup section and link it in helm install page

wattli:
Otherwise people will not set up corresponding environment and the installation will fail.

For instance, you need to set up RBAC https://preliminary.istio.io/docs/setup/kubernetes/quick-start/#google-kubernetes-engine in GKE.

### DIFF
--- a/content/docs/setup/kubernetes/helm-install/index.md
+++ b/content/docs/setup/kubernetes/helm-install/index.md
@@ -15,6 +15,8 @@ plane and the sidecars for the Istio data plane.
 
 ## Prerequisites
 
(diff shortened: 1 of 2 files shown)


wattli:
I agree and I can work on it, but not sure everyone like this idea or we can just make a section called 'Platform setup' in that page, I feel the latter is safer. @geeknoid

sdake:
ya please have @geeknoid weigh in.

Cheers
-steve

wattli:
Cool, i will ping back when it's done.

wattli:
@geeknoid , PTAL

geeknoid:
Looks good.

@sdake?

PR #1808, opened 2018-07-13
Updated link to vm drivers link to list names of drivers

brian-avery:
This PR fixes https://github.com/istio/istio.github.io/issues/1806


brian-avery:
Thanks @geeknoid

PR #1834, opened 2018-07-16
Extract platform prerequisites 

andraxylia:
Extract platform prerequisites  and download of release in separate pages to be re-used by the quickstart and helm install.

Right now, helm instructions are confusing because the prereq send you to the full page of install with istio-auth.yaml


andraxylia:
I need some help with this PR: when running netlify, it does not find the new pages I added “platform-setup”. There seems to be some black magic in how links work. I used the hugo conventions but files are not being copied to the site.

I see faq directory also has a bunch of files that do not show up, there is a general problem @geeknoid @linsun

geeknoid:
All markdown files should be called index.md and be in a subdirectory whose name will be the URL. That's how all markdown is arranged in the content directory.

geeknoid:
The faq and glossary content are handled in special ways by the infrastructure. Although there are distinct .md files for each question and term, they are generated as agglomerated .html files instead.

From your repo, you can do "make gen" to generate the web site locally in the "public" folder. You can then easily inspect the output and get an idea of the translation process from markdown to HTML.

andraxylia:
@geeknoid is there any way to not add an entry in the left side menu? I noticed that if I add an index.md file for platform-setup, I will have a corresponding entry in the menu, and I do not want this.

andraxylia:
Even with index.md, it does not work if the sub-directory is at a deeper level, only if it is under kubernetes. Are these restrictions coming from hugo ? I could not find anything in their doc.

sdake:
@andraxylia I don't have ACLs to approve this PR as I'm not in the root owners file and don't have access to chinese translated commits.

Cheers
-steve

andraxylia:
please approve the Chinese changes, it's fixing the broken links as reported by lint
 @fleeto
 @gyliu513
 @rootsongjc

PR #1873, opened 2018-07-18
Fix the directory in which the Ansible playbooks can be found

geoand:
Relates to: https://github.com/istio/istio/pull/7207


geoand:
Thanks for pointing that out @sdake! Fixed

PR #1874, opened 2018-07-18
Update default metrics page

bianpengyuan:
Update according to https://github.com/istio/istio/pull/7034


bianpengyuan:
cc @douglas-reid

PR #1882, opened 2018-07-19
Add countdown timer to launch,.

geeknoid:
Add a countdown timer to 1.0. This counts down towards July 23rd.

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -38,6 +38,8 @@ title: Istio
 </script>
 
 <main class="landing">
+    <a id="timer" class="promotion" href="https://preliminary.istio.io"></a>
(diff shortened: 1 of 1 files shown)


geeknoid:
NB: See the netlify link for the preview

linsun:
Sorry just got back.... Looking at the preview... Is this correct that Istio 1.0 is in 4 days?   I had thought it is end of July.

geeknoid:
@linsun @smawson I don't know exactly at this point. I'm also just back from vacation and did this timer while on vacation. When I left, July 23rd was one of the dates being used.

So Sven, what's the plan?

Issue #1898, opened 2018-07-20
{CLUSTER_NAME} is undefined on control plane


sdake:
```
cd $WORK_DIR
kubectl create secret generic ${CLUSTER_NAME} --from-file ${CLUSTER_NAME} -n ${NAMESPACE}
kubectl label secret ${CLUSTER_NAME} istio/multiCluster=true -n ${NAMESPACE}
```
from:
https://preliminary.istio.io/docs/setup/kubernetes/multicluster-install/#instantiate-the-credentials-for-each-remote-cluster

sdake:
$WORK_DIR and ${NAMESPACE} are also undefined.

tiswanso:
The values are defined in this section:  https://preliminary.istio.io/docs/setup/kubernetes/multicluster-install/#generate-kubeconfigs-for-remote-clusters

Maybe we need to clarify that the values are from that section.  If the user uses the same shell and just switch between clusters using `kubectl config use-context <cluster context>` the env vars would be still set.  But if the user is using a different set of shells then these values wouldn't be defined in both shells after running the instructions in that section.

sdake:
@tiswanso agree - I am using the second workflow which is probably the cause of confusion.  The docs don't mention anything about `use-context`.  Whatever can be done to simplify the docs would be great 👍 .  Most people reading these docs will cut and paste, and not apply a whole lot of reasoning to their actions and expect things to work.  It took me a few hours (and I'm familiar with this stuff) to sort out the workflow using separate shells.

PR #1918, opened 2018-07-23
k8s Multicluster--clarify vars for remote-cluster secret

tiswanso:
Fixes: #1898

### DIFF
--- a/content/docs/setup/kubernetes/multicluster-install/index.md
+++ b/content/docs/setup/kubernetes/multicluster-install/index.md
@@ -211,6 +211,16 @@ to the remote cluster.
     EOF
     {{< /text >}}
 
+1. _(Optional)_  Create file with environment variables for creating the remote cluster secret:
+
+    {{< text bash >}}
+    $ cat <<EOF > remote_cluster_env_vars
+    export CLUSTER_NAME=${CLUSTER_NAME}
+    export KUBECFG_FILE=${KUBECFG_FILE}
+    export NAMESPACE=${NAMESPACE}
+    EOF
+    {{< /text >}}
+
 At this point, the remote clusters' `kubeconfig` files have been created in the current directory.
 The filename for a cluster is the same as the original `kubeconfig` cluster name.
 
@@ -219,6 +229,12 @@ The filename for a cluster is the same as the original `kubeconfig` cluster name
 Execute this work on the cluster running the Istio control
 plane using the `WORK_DIR`, `CLUSTER_NAME`, and `NAMESPACE` environment values set during the
 [Generate kubeconfig for remote clusters](#generate-kubeconfigs-for-remote-clusters) steps.
+
+* _(Optional)_  Source the environment variables file created for the remote cluster secret:
+
+    {{< text bash >}}
+    $ source remote_cluster_env_vars
+    {{< /text >}}
 
 Istio can be installed in a different namespace other than
 istio-system.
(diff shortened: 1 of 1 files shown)


geeknoid:
I'm not sure where the CI failures are coming from. Could you try rebasing your PR on the latest bits and see if that cures things?

Thanks.

content/docs/setup/kubernetes/multicluster-install/index.md
```diff
@@ -214,7 +214,8 @@ The filename for a cluster is the same as the original `kubeconfig` cluster name
 ## Instantiate the credentials for each remote cluster
 
 Execute this work on the cluster running the Istio control
-plane.
+plane using the `WORK_DIR`, `CLUSTER_NAME`, and `NAMESPACE` environment values set during the
+[Generate kubeconfig for remote clusters](#generate-kubeconfigs-for-remote-clusters) steps.
```
sdake: Might provide a URL to how that is done?  It is not obvious just from this text that a C&P from the website will result in a broken installation.

sdake: Right - might use an idea icon with a the link with text such as "One way to achieve this is using kubectl contexts.  Maybe that is the wrong approach for this documentation - I'm just not sure, but we expect people will cut and paste the docs, and if they cut and paste this doc, it won't work for them.

sdake: @tiswanso one thing you could do here is rename all the variables with a prefix of ISTIO_MC_, then run `export | grep ISTIO_MC > $HOME/mc_environment_to_source` - then use a block quote to copy the mc_environment_to_source to the remote clusters, then use `source $HOME/mc_environment_to_source`.  While not straight copy and paste, atleast it would not present as it currently does with "it didn't work, what went wrong" with no errors of any kind.  a source `$HOME/mc_environment_to_source` on the remote would surely yield an error if that file was not present, triggering the reader to read more closely.  (In copy and paste mode, readers don't read the prose much, they copy and paste :)

tiswanso: I'll see if I can spend some time on that.  I'm feeling like we're going pretty far with the make "copy paste work with a user in mindless mode".  I have scripted all of this for e2e tests and it might be more straight forward to add a form of that script to istio/bin so we can reference it in this doc.  Of course, it's also time consuming to script things for all cases and multicluster users would likely have their own flavors of scripts/utils to manage multiple clusters, e.g. sig-multicluster stuff, so an istio/bin script would pretty much be for eval setups only.

Additionally, I want to make an addendum to this doc with a GKE multicluster example.  I think that would be more valuable than making all the doc flow & copy/paste content super idiot proof.

sdake: @tiswanso the docs serve two masters.  The first is the copy and paste operators who want to get going quickly for an eval.  The second are operators that want to actually deploy and understand how this stuff works at a high level.  Scripting is not helpful for either of these cases, nor are the e2e tests usable for production.  We have been down the road of using gating code to implement installation (ref kolla-kubernetes) and it was super duper painful.

sdake: @tiswanso - if you can't get to this today - PLMK.  I think the current content is mergeable, and the cut and paste refinement can come in a followup and possibly backported.

tiswanso: Sorry... was focusing on gke example and re-validating latest 1.0s for e2e multicluster scenarios.  I'll push up something similar to your suggestion ASAP.

PR #1927, opened 2018-07-24
New topology diagrams

oaktowner:
Two new topology diagrams.


linsun:
@oaktowner should we still use mixer in the topo diagram since mixer is no longer a component in istio install now, it is broken into 2 pieces (telemtry and policies).  It could be helpful to show the 2 pieces to match what users are seeing and that the report call isn't holding up the request.

oaktowner:
@linsun I'm a little hesitant to do that -- the surrounding text on https://github.com/istio/istio.github.io/blob/master/content/docs/concepts/policies-and-telemetry/index.md is ALL about Mixer. 

I could do that, but it would need a serious rewrite (or at least a section explaining the concept of Mixer along with its relationship with istio-policy and istio-telemetry. 

@geeknoid WDYT?

linsun:
I see, Dan, seems our doc has not been updated since 0.8 after mixer is broken into 2 pieces.

PR #1928, opened 2018-07-24
New adapters diagram

oaktowner:
new Mixer adapters diagram


geeknoid:
Hey Dan,

Thanks for the onslaught of update diagrams. Unfortunately, we need a bit of markdown tweaking whenever the dimensions of a diagram change. In particular, the width and ratio values need to be adjusted within the markdown file. See https://preliminary.istio.io/about/contribute/writing-a-new-topic/#adding-images

To get the exact width & height of an SVG document, just load the file up into a text editor and the width & height are listed right there at the top of the file.

If the width & ratio values aren't specified correctly, the picture will be the wrong size and/or will have a bunch of whitespace around it.

Would you mind submitting updating markdown files for any of the other diagrams you've already changed?

linsun:
Dan, I don't see newrelic adapter in our repo, neither is datadog:  https://github.com/istio/istio/tree/master/mixer/adapter

Is the diagram representing some of the WIP mixer adapter or they aren't open sourced yet?

oaktowner:
@geeknoid will start updating ratios. Thanks for the tip.
@linsun I have just literally been recreating the existing diagrams as-is with a refresh of the style. There is a Datadog adapter (it's dogstatsd), so I'll leave that in. I will delete New Relic (not sure why it was there in the original!).

linsun:
@oaktowner thanks, sorry for the nits,  i was just reading based on what makes sense to me vs what is out there today... could not connect dogstatsd with data dog earlier.:)

PR #1931, opened 2018-07-24
New version of new authz diagram

oaktowner:
This is a new version of the new diagram that @liminw added last week. Like all the new diagrams, the Google Drawing is in the Istio Community Drive (Misc/Drawings for istio.io site)


geeknoid:
Please update markdown to have the proper width and ratio values for revised images.

linsun:
@oaktowner @liminw does authz require mixer (policies at least)?  I'm surprised not seeing it on the diagram.

liminw:
@linsun In the new diagram, Pilot distributes the policies to Envoy, and Envoy enforces the policies. Mixer is not required. Istio RBAC is RPC level access control policy. The syntax is simple enough that we think local enforcement makes sense. 

In the future, we may introduce resource level access control, which requires more complex logic and storage, those kind of access control will have to be done on Mixer or a backend connected from Mixer. We may even combine local (Envoy) and remote (Mixer) authorization depending on the use cases.

linsun:
@liminw thanks, didn't realize that is in place already, makes sense. :)

rcaballeromx:
@oaktowner @geeknoid I will update the MD for the right ratios for the images and remove the invisible objects causing the weird proportions.

sdake:
As https://github.com/istio/istio.github.io/pull/1943 merged, I've approved this.  Thanks @rcaballeromx.

/lgtm
/approve

PR #1933, opened 2018-07-24
New diagram for mesh expansion example

oaktowner:
Replaces the diagram at https://istio.io/docs/guides/integrating-vms/

(original is located in Community Drive -> Misc -> Drawings for Istio.io site)


geeknoid:
Please update markdown to have the proper width and ratio values for revised images.

linsun:
@oaktowner you have 2 boxes for reviews service... think one should be rating instead.  Also, do you think it will be better if we have all k8s services shown within k8s so it is clear which ones run in k8s?

PR #1943, opened 2018-07-24
Remove ghost elements from security diagrams.

rcaballeromx:
By removing ghost elements from the SVG files, I could fix the scaling of the
diagrams.


content/docs/concepts/security/index.md
```diff
@@ -30,7 +30,7 @@ and the `backend` service running as the `backend-team` service account. Istio
 supports services running on Kubernetes containers, virtual machines, and
 bare-metal machines.
 
-{{< image width="80%" ratio="56.25%"
+{{< image width="60%" ratio="60%"
```
geeknoid: I haven't looked at the images, but the fact you're using 60% for both width and ratio in two separate images seems suspect.

Width specifies the percentage of the page width that the image should occupy. Ratio represents the physical image height / physical image width * 100.

rcaballeromx: I took a look at the image and found having both at 60% made the image look best.

PR #1961, opened 2018-07-25
Added a missing 'one' to a sentence

holgero:


geeknoid:
This change needs to be applied to the original content in istio/api. @frankbu can you do the honors please?

frankbu:
@holgero This documentation is generated from the proto file here: https://github.com/istio/api/blob/ab4e69a0d8bd0658472a7b849c1ba913ad1a9849/networking/v1alpha3/virtual_service.proto#L656

I'll make your fix there. Thanks for pointing this out.

geeknoid:
Being fixed at the source.

Thanks.

PR #1969, opened 2018-07-25
CRDs must be installed separately

andraxylia:
Helm 2.9 install had issues with CRDs because validation failed. The fix for < 2.9 is to create CRDs separately using kubectl apply. Helm 2.10-beta fixes the issue but it is not stable. When 2.10 is released we will recommend users to use this version. Recommend users to wait a few seconds after applying CRDs.


content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -32,40 +32,52 @@ plane and the sidecars for the Istio data plane.
...
 
-## Option 2: Install with Helm and Tiller via `helm install`
+### Option 2: Install with Helm and Tiller via `helm install`
 
 This option allows Helm and
 [Tiller](https://github.com/kubernetes/helm/blob/master/docs/architecture.md#components)
 to manage the lifecycle of Istio.
 
-1. If a service account has not already been installed for Tiller, install one:
+* If a service account has not already been installed for Tiller, install one:
```
sdake: this is an ordered list, please keep the `1.`

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -32,40 +32,52 @@ plane and the sidecars for the Istio data plane.
...
 
     {{< text bash >}}
     $ kubectl create -f install/kubernetes/helm/helm-service-account.yaml
     {{< /text >}}
 
-1. Install Tiller on your cluster with the service account:
+* Install Tiller on your cluster with the service account:
```
sdake: ordered list, please keep the `1.`

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -32,40 +32,52 @@ plane and the sidecars for the Istio data plane.
...
 
     {{< text bash >}}
     $ helm init --service-account tiller
     {{< /text >}}
 
-1. Install Istio:
+* Install Istio:
```
sdake: ordered lsit, please keep the `1.`

content/docs/setup/kubernetes/quick-start/index.md
```diff
@@ -21,13 +21,18 @@ To install and configure Istio in a Kubernetes cluster, follow these instruction
...
-installations of Istio. With this installation, you can leverage all the
-options to configure and customize Istio to your needs.
+    {{< text bash >}}
+    $ kubectl apply -f install/kubernetes/helm/istio/templates/crds.yaml -n istio-system
+    {{< /text >}}
+
+1. To install Istio's core components you can choose any of the following four
+**mutually exclusive** options described below. However, for a production setup of Istio,
+we recommend installing with the
+[Helm Chart](/docs/setup/kubernetes/helm-install/), to leverage all the
```
sdake: please don't use the word leverage.  Instead I recommend the word plain word `use`.

sdake: all of the

content/docs/setup/kubernetes/quick-start/index.md
```diff
@@ -21,13 +21,18 @@ To install and configure Istio in a Kubernetes cluster, follow these instruction
...
+configuration options and to customize Istio to your needs.
```
sdake: options.  This will permit customization of Istio to operator specific needs.

content/docs/setup/kubernetes/helm-install/index.md
andraxylia: The problem was a could not do sub-numbered lists, and I already have Installation steps 1 and 2 above. If I transform this into a numbered list, it will be confusing with the steps above.

content/docs/setup/kubernetes/quick-start/index.md
andraxylia: ok.

andraxylia: Sorry, it sounds a bit like a phrasal : "to use all of the configuration option". 
vs

"use all the configuration options".

content/docs/setup/kubernetes/quick-start/index.md
andraxylia: ok.

content/docs/setup/kubernetes/helm-install/index.md
andraxylia: I changed it nevertheless, so we can get this in sooner, people need to start testing with the new candidate image and new instructions.

content/docs/setup/kubernetes/quick-start/index.md
sdake: your right.

linsun:
One issue I have with this approach is that I'm installing CRDs I don't need.  Secondly, I'm adding conditions in CRDs as it is a template file, so when mixer is enabled, install these, pilot is enabled installed the other CRDs.   see https://github.com/istio/istio/pull/7408. When the conditions are added, we won't be able to run `kubectl apply -f install/kubernetes/helm/istio/templates/crds.yaml` as it is a template file.

cc @costinm @rshriram @gyliu513 @ymesika

gyliu513:
@andraxylia @linsun I was a bit confused, with this fix, how can I install istio helm chart in one step via `helm install`? Does it mean I cannot install Isito helm chart in one step?

gyliu513:
I was also having some discussion at https://github.com/helm/helm/pull/3982#issuecomment-408444874 , as I think helm already support install CRDs before other resource can reference it at https://github.com/helm/helm/pull/3019 in helm 2.7, but do not know why helm community created another PR to fix it again for 2.10.

gyliu513:
+1 to not change user experience. Istio 1.0 is a milestone release, using two steps to install does not seems to be a good solution.

andraxylia:
@morvencao @gyliu513 Indeed, this is a two step installation instead of one, but it is the only solution that works until helm 2.10 is released. It is not a matter of preference, it is obvious we would not add an extra installation step for no reason. The matter was at large discussed on slack and in the environments meeting. 
"Helm 2.9 install had issues with CRDs because validation failed. The fix for < 2.9 is to create CRDs separately using kubectl apply. Helm 2.10-beta fixes the issue but it is not stable. When 2.10 is released we will recommend users to use this version. Recommend users to wait a few seconds after applying CRDs.
"

The corresponding change in the helm charts has already been made, I am merely updating the docs.

When helm 2.10 is released, we will remove this step.

gyliu513:
Thanks @andraxylia , I see, +1 to merge this first, but this breaking change is not good for Istio.

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -33,7 +33,19 @@ plane and the sidecars for the Istio data plane.
...
    appended to the end of the helm operation.
 
-## Option 1: Install with Helm via `helm template`
+## Installation steps
+
+1. Install Istio's [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)
+via `kubectl apply`, and wait a few seconds for the CRDs to be committed in the kube-apiserver:
+
+    {{< text bash >}}
+    $ kubectl apply -f install/kubernetes/helm/istio/templates/crds.yaml -n istio-system
```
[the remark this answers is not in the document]
linsun: +1 good catch, we can get this fixed in a seperate WI.

linsun:
This is a temp solution before we can get helm 2.10 which is expected next month cc @gyliu513 @morvencao

PR #1944, opened 2018-07-24
Update the Citadel certificate lifetime config document

lei-tang:
Update the Citadel certificate lifetime config document to be consistent with the implementation.

PR #1978, opened 2018-07-26
Revert "Update the Citadel cert lifetime config document (#1944)"

sdake:
This reverts commit 062d0e3baf2a6113c3a04a0a7649fd6af6777257.

Given that https://github.com/istio/istio/pull/7398 was closed, @myidpt asked for a revert of this PR I erroneously merged.

Cheers
-steve


myidpt:
Hi Steven,
I found that we already have https://preliminary.istio.io/docs/reference/commands/istio_ca/ to track all the flag default values already. So instead of reverting, I think it would be great if we can just add a link to that page :)

PR #1984, opened 2018-07-26
Wildcard https egress gateway

vadimeisenbergibm:


content/blog/2018/wildcard-https-egress-gateway/index.md
```diff
@@ -0,0 +1,186 @@
+---
+title: Configure Egress Gateway for HTTPS traffic to wildcarded domains
```
rshriram: This should not be a blog. This should be a task. Also, we really really need an advanced task for all variants of egress.

vadimeisenbergibm: Let me change it to a task. So which ideas do you have for the advanced task? Currently we have, in addition to this PR:

https://istio.io/docs/tasks/traffic-management/egress/
https://istio.io/docs/tasks/traffic-management/egress-tls-origination/
https://istio.io/docs/tasks/traffic-management/egress-gateway/
https://istio.io/blog/2018/egress-https/
https://istio.io/blog/2018/egress-tcp/
https://istio.io/blog/2018/egress-monitoring-access-control/

geeknoid:
Is this PR still needed, or should we close it?

geeknoid:
To me, this looks right at home where you have it today.

I do recommend shortening most of the section header names. They're quite verbose and end up not looking very good within the TOC on the right of the page.

frankbu:
I agree this belongs in examples, but I'm wondering if we should restructure the whole set of egress examples because right now it feels like there are just a bunch of random examples, instead of an organized set. Maybe we could just have 2 egress examples (1. Egress direct from sidecar, and 2. Egress via gateway). Each of these could have several sections organized the same (e.g., wildcard (simple and arbitrary), tls origination, mutual tls). That way it would be a more complete and organized coverage. ServiceEntry seems very complicated, there are so many variations, but this way we could try to cover lots of the common things uniformly.

Maybe the instructions for deploying a custom gateway could be split out and reused?

vadimeisenbergibm:
@frankbu The Egress control is pretty complex topic. The different dimensions are:

1. Different protocols: HTTP, TLS, TCP
2. Using/not using egress gateway
3. For HTTP: using/not using TLS origination
4. For egress gateway: with or without wildcards
5. For egress gateway and wildcards: wildcard domains for one specific host or arbitrary hosts
6. For egress gateway: with or without mTLS between a sidecar proxy and the egress gateway

The structuring is not easy, since multiple aspects are intertwined. For example, regarding: 
> Maybe the instructions for deploying a custom gateway could be split out and reused

 different ports are used for different protocols (HTTPS and MongoDB), so the instructions are also a bit different. It would be harder to follow the instructions like: "deploy a custom gateway with an SNI proxy, while changing the ports here, here and here."

vadimeisenbergibm:
@geeknoid I shortened the titles a bit. One issue is that the lint does not allow using the same title for multiple sections. Due to this, using "Cleanup" for the cleanup subsections is not possible, so long titles must be used for each of the cleanup subsections, e.g.: `Cleanup of HTTPS traffic configuration to arbitrary wildcarded domains`. Maybe, the lint can be changed to allow the same title for multiple subsections in the case of the "Cleanup" title only?

vadimeisenbergibm:
@frankbu Could you please review and merge this one? I will work on the MongoDB item next.

PR #1987, opened 2018-07-26
bug with new certs getting picked up

GregHanson:
certs were not getting picked up by ingress pod during the secure ingress task:
https://preliminary.istio.io/docs/tasks/traffic-management/ingress

### DIFF
--- a/content/docs/tasks/traffic-management/secure-ingress/index.md
+++ b/content/docs/tasks/traffic-management/secure-ingress/index.md
@@ -290,6 +290,12 @@ they have valid values, according to the output of the following commands:
     $ kubectl logs -n istio-system -l istio=ingressgateway
     {{< /text >}}
 
+1.  If the secret was created but the keys were not mounted, kill the ingress gateway pod and force it to reload certs:
+
+    {{< text bash >}}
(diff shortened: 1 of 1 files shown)


content/docs/tasks/traffic-management/secure-ingress/index.md
```diff
@@ -312,6 +318,12 @@ In addition to the steps in the previous section, perform the following:
     Subject: C=US, ST=Denial, L=Springfield, O=Dis, CN=httpbin.example.com
     {{< /text >}}
 
+1.  If the secret was created but the keys were not mounted, kill the ingress gateway pod and force it to reload certs:
```
frankbu: This is redundant, since the Mutual TLS troubleshooting instructions already say "in addition to the above ..."

vadimeisenbergibm: @frankbu note that it is another certificate, loaded from another volume/secret. There could be a case when the certificate above is loaded successfully, while this certificate is not not, and killing the pod would be required again.

frankbu: I see. I guess it would be clearer if the second one said "the CA certificate secret", instead of just "the secret", same as the first one.

Issue #2003, opened 2018-07-27
Add Chinese community portal on the website


rootsongjc:
For the following reasons, I recommend adding a Chinese community portal to the website:

- To acknowledge the community for their contributions to Chinese documents.
- To help them quickly get familiar with and understand the project.
- To help Chinese developers quickly integrate into the project.
- In order to overcome the jet lag problem, it is possible to communicate smoothly online.

**Responsibility**

- Update the Chinese documents
- Post blogs in bio-language
- Organise activities or meetups related to Istio
- Feedback to the upstream
- Promot Istio in China

**Action**

- Add a description about [ServiceMesher](http://www.servicemesher.com) community on the official website, the Chinese documents are translated by these guys
- Writing a document to introduce the Chinese how to take part in the community

@geeknoid If you have any questions? @fleeto @gyliu513 We'd like to add a document on `content_zh` to describe this.

myidpt:
https://istio.io/community/ might be a good place to add a link to the community http://www.servicemesher.com/.

If you want to add more content to introduce the Chinese how to take part in the community, you might want to put it somewhere under https://preliminary.istio.io/zh. This gives you more flexibility to present/advertise, and most Chinese developers will see the Chinese web pages.

rootsongjc:
Hi, @geeknoid Since istio 1.0 has been released for a few days, we need to consider when we can add the Chinese language switch button. 
There are a few things to do before adding the button:
- [x] check the links in Chinese content to ensure it refers to the right address, see #2014 
- [x] translate the keywords in the front-matter of documents, see #2127 
- [ ] evluate if the Chinese documents have been sycn with the English
- [ ] write a blog to announce Chinese documents release(in English and Chinese both) and acknowleage the ServcieMesher community's contribution
- [x] to strengthen internal collaboration, we created slack channel and a GitHub robot, see https://github.com/servicemesher/slackbot thanks to @fleeto 
- [ ] set up a milestone to trace the progress?

@fleeto @gyliu513 Do you have anything to add?

geeknoid:
I've got to say that you guys have been doing an amazing job. You're all
very dedicated, it's impressive. Thanks for doing this work, it'll make
Istio more accessible to more people.

PR #2013, opened 2018-07-27
Add Servicemesher portal on the Community page

rootsongjc:
Refer to #2003  @geeknoid @fleeto


content/about/community/index.md
```diff
@@ -66,13 +61,19 @@ There are quite a few ways to get plugged in:
...
       <td class="logo">{{< inline_image "community/group.svg" >}}</td>
       <td class="desc">
         If you'd like to contribute to the Istio project, consider participating in our
         <a href="https://github.com/istio/community/blob/master/WORKING-GROUPS.md">working groups</a>.
       </td>
     </tr>
+    <tr>
+      <td class="logo">{{< inline_image "community/servicemesher.svg" >}}</td>
+      <td class="desc">
+        Chinese content is maintained by the
```
geeknoid: Our Chinese-language documentation is maintained

layouts/partials/icons/servicemesher.svg
```diff
@@ -0,0 +1,14 @@
+<svg viewBox="0 0 310 310">
```
geeknoid: You only need this SVG if we put an icon in the page footer, and I don't think we should. So you can delete this file.

PR #2024, opened 2018-07-27
Initial checkin of telemetry FAQ.

geeknoid:
This is a markdown version of the telemetry FAQ.


geeknoid:
@sdake Thanks for the review, I addressed your feedback, except for the I->you thing. For the FAQ, these are supposed to be questions the user is asking him/herself.

PR #2028, opened 2018-07-28
Final release material

geeknoid:
- New home page layout and text

- Release 1.0 blog post

- New docs->concepts->what is istio page


sdake:
@geeknoid looks really good.  One small nitpick is that I have to scroll on my macbook 13" to get to the links and bottom of the page when in either full screen mode or not.  Yet there is free web space on the left and right that would avoid this scrolling.  Would be ideal if all the content could be present on one page on a typical 13" display. :)  Not sure if that is possible - if not, ignore this :)

geeknoid:
@sdake I'm continuing to play with this. It's still not right on small screens, but I think it should be pretty nice for a 13" display. WDYT?

geeknoid:
Now it should look good on mobile too.

sdake:
@geeknoid renders well at "67%" magnification on 13" MBP.  I run in can't-see-well mode of 100%+ by default :)

Definitely looks nice and I wouldn't get hung up about the magnification.

rshriram:
I don’t see a big difference between old and new except for less blue on the front page. It certainly looks simpler and easier on the eye. 

The learn more links are gone. Intentional ?

sdake:
@geeknoid perfect on fullscreen 13" MBP.

PR #2032, opened 2018-07-28
crd do not need namespace.

gyliu513:
/cc @linsun @andraxylia


gyliu513:
@sdake @geeknoid @linsun this should be land in 1.0 release.

PR #2059, opened 2018-07-30
GKE multicluster example: Fixes and uninstall section addition

tiswanso:
Reran through it and found a couple things.  Also, add uninstall info since GCP firewall rule will stay around if not explicitly deleted.

### DIFF
--- a/content/docs/examples/multicluster/gke/index.md
+++ b/content/docs/examples/multicluster/gke/index.md
@@ -5,7 +5,7 @@ weight: 65
 keywords: [kubernetes,multicluster]
 ---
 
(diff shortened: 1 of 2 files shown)


content/docs/examples/multicluster/gke/index.md
```diff
@@ -5,7 +5,7 @@ weight: 65
 keywords: [kubernetes,multicluster]
 ---
 
-This example demonstrates how to use Istio' multicluster feature to join 2
+This example demonstrates how to use Istio's multicluster feature to join 2
```
sdake: 2/two

PR #2061, opened 2018-07-30
Helm install not rendering properly

sdake:


sdake:
@geeknoid solved this problem with a `>` instead of a ordered list item

PR #2082, opened 2018-07-31
Minikube settings

### LINKED ISSUE #2017: preliminary / 1.0.0 docs do not document required admission controllers for minikube
Compare:

https://preliminary.istio.io/docs/setup/kubernetes/platform-setup/minikube/

to:

https://istio.io/docs/setup/kubernetes/quick-start/#minikube

And observe that critical details are missing-- namely required admission controllers are not explicitly enabled.

andraxylia:
We should advise the same minikube settings as tested by circleci.

Increased the memory and specified the number of CPUs.

Added the admission controllers plugins for k8s 1.9.0, they are enabled by default only starting with 1.10.0. Fixes #2017 .

### DIFF
--- a/content/docs/setup/kubernetes/platform-setup/minikube/index.md
+++ b/content/docs/setup/kubernetes/platform-setup/minikube/index.md
@@ -16,7 +16,9 @@ Follow these instructions to prepare Minikube for Istio.
 1. Select a
    [VM driver](https://kubernetes.io/docs/setup/minikube/#quickstart)
    and substitute `your_vm_driver_choice` below with the installed virtual
-   machine (VM) driver.
+   machine (VM) driver. To install Istio control plane components and add-ons,
+   as well as other applications,
+   we recommend starting minikube with 8192 MB of memory and 4 CPUs: 
 
     On Kubernetes **1.9**:
 
--- a/content/docs/setup/kubernetes/platform-setup/minikube/index.md
+++ b/content/docs/setup/kubernetes/platform-setup/minikube/index.md
@@ -21,11 +21,18 @@ Follow these instructions to prepare Minikube for Istio.
     On Kubernetes **1.9**:
 
     {{< text bash >}}
-    $ minikube start --memory=4096 --kubernetes-version=v1.9.4 --vm-driver=`your_vm_driver_choice`
+    $ minikube start --memory=8192 --cpus=4 --kubernetes-version=v1.9.4 \
+        --extra-config=controller-manager.cluster-signing-cert-file="/var/lib/localkube/certs/ca.crt" \
+        --extra-config=controller-manager.cluster-signing-key-file="/var/lib/localkube/certs/ca.key" \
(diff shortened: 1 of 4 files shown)


content/docs/tasks/policy-enforcement/rate-limiting/index.md
```diff
@@ -58,7 +58,7 @@ so the configuration to enable rate limiting on both adapters is the same.
     Run the following command to enable rate limits.
 
     {{< text bash >}}
-    $ kubectl apply -f @samples/bookinfo/policy/mixer-rule-productpage-ratelimit.yaml@
+    $ kubectl apply -f examples/bookinfo/policy/mixer-rule-productpage-ratelimit.yaml@
```
geeknoid: This change is not right. The original sequence is valid. Unfortunately, the will temporarily fail linting due to the fact the file only exists in the release-1.0 branch and has not been merged to master.

@mandarjog Do you need to merge this file to master, and will that be taken care of automatically?

content/docs/setup/kubernetes/platform-setup/minikube/index.md
```diff
@@ -21,11 +21,18 @@ Follow these instructions to prepare Minikube for Istio.
     On Kubernetes **1.9**:
 
     {{< text bash >}}
-    $ minikube start --memory=4096 --kubernetes-version=v1.9.4 --vm-driver=`your_vm_driver_choice`
+    $ minikube start --memory=8192 --cpus=4 --kubernetes-version=v1.9.4 \
+        --extra-config=controller-manager.cluster-signing-cert-file="/var/lib/localkube/certs/ca.crt" \
```
sdake: I am pretty sure the cluster-signing-cert-file is not needed - as minikube + Istio run without them in this exact same setup.  Could you try without them - as the command is a little clunky with them?  Otherwise looks good.

content/docs/tasks/policy-enforcement/rate-limiting/index.md
rshriram: This file is in master

geeknoid: I don't see it. I see it in release-1.0 but not master.

andraxylia: Ok, I can revert this, I added it because of the linter.

content/docs/setup/kubernetes/platform-setup/minikube/index.md
andraxylia: Then why do we have it in circleci?

The idea is we recommend exactely what we test. Until we have the equivalent removed from the circleci file, we should have this here.

duderino: Agree that circleci should test what we recommend.   If we think the settings are too high (e.g., 8GB) then another approach is to reduce the minikube memory used by circleci.

andraxylia: Checked with @ayj and we still need it if we use self-signed certificates or k8s CA. It's safer to leave it there.

sdake: wfm.

geeknoid:
Still a few spelling errors to deal with. "minikube" should be in backticks, the other errors probably need to be added to .spelling.

andraxylia:
I fixed the lint errors, can you please lgtm again?

PR #2086, opened 2018-07-31
Disable Mesh Expansion page.

geeknoid:
Per email discussion.

I'll cherry-pick this to the release-1.0 branch once this is in.


geeknoid:
@costinm Turns out removing the mesh expansion page completely leads to a bunch of broken links. So I've added the page back, but put a warning at the top of it about the feature being broken.

adeshmukh:
@geeknoid could the warning include a reference to issue tracking so we can be notified when the issue is addressed?

durkode:
Judging by the release notes of 1.0.1 (https://istio.io/about/notes/1.0.1/) should Mesh Expansion now work? If so can we remove the warning?

munnerz:
I'm trying to find information on why mesh expansion is broken, and tracking info on it becoming fixed again 😄 

As @adeshmukh says above, does anyone have links/references to the relevant email conversation mentioned, and/or any ongoing discussion on re-enabling mesh expansion support?

PR #2103, opened 2018-08-01
Full text RSS feeds

craigbox:
Currently, only the short description of each post is included in the RSS feed.  This change will update the feed to include the whole text of each blog post.

### DIFF
--- a/layouts/feed/single.rss.xml
+++ b/layouts/feed/single.rss.xml
@@ -20,7 +20,7 @@
       {{ if and (not $post.Draft) $post.IsPage }}
         <item>
           <title>{{ $post.Title }}</title>
-          <description>{{ $post.Description }}</description>
+          <description>{{ $post.Content | html }}</description>
           <pubDate>{{ $post.PublishDate.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
           <link>{{ $post.Permalink }}</link>
           <author>{{ $post.Params.attribution }}</author>


craigbox:
Confirmed in Feedly that this does as expected.
You can test by pasting https://deploy-preview-2103--preliminary-istio.netlify.com/feed.xml into your favourite RSS reader.

geeknoid:
I looked into this a while back and I had come to the conclusion that doing this was a non-standard extension to the RSS protocol and some readers would be unhappy about it.

craigbox:
While potentially non standard, it is how most feeds work: look at
kubernetes.io as an example.

craigbox:
@geeknoid Can we at least try the full feed in the content/content:encoded tag?

https://github.com/gohugoio/hugo/issues/4242

geeknoid:
Sorry for the delay. I'm going to spend a bit of time on this over the weekend and hopefully we'll get to a good place. Thanks for the two links for study, very useful.

geeknoid:
All right, let's go for it.

craigbox:
...can we backport it to the 1.0 branch?

(On that note, can we separate the blog from the docs branches?  It causes awkward bursts of blog posts that people think are released at commit time, but are actually released when the prelim becomes the release)

craigbox:
Thank you !

...can we backport it to the 1.0 branch?

(On that note, can we separate the blog from the docs branches?  It causes awkward bursts of blog posts that people think are released at commit time, but are actually released when the prelim becomes the release)

geeknoid:
I'll be doing another round of cherry picks from master into the release-1.0 branch in a couple days.

As for blog posts, it's just a matter of cherry picking them as they are published. It doesn't seem worth the overhead of a completely orthogonal way to manage them.

PR #2111, opened 2018-08-01
Update Option 1 and Option 2

chymy:
Option 1: Install with Helm via `helm template`  -->Option 1: Install with Istio via `helm template`
Option 2: Install with Helm and Tiller via `helm install` -->Option 2: Install with Istio and Tiller via `helm install`


geeknoid:
I don't think this is correct. This is about "installing Istio" not "Installing with Istio".

PR #2140, opened 2018-08-03
Add twitch livestream blog post

nibalizer:
This is for the all-day istio livestream on August 17th.

I am thinking it might be better to have a url like ``https://istio.io/livestream`` that redirects to the twitch page so it looks less IBMy, but I'm not sure how to do that technically.

### DIFF
--- a/content/blog/2018/istio-twitch-stream/index.md
+++ b/content/blog/2018/istio-twitch-stream/index.md
@@ -1,0 +1,26 @@
+---
+title: All Day Istio Twitch Stream
+subtitle:
+description: Istio hosting an all day twitch stream to celebrate the 1.0 release
+publishdate: 2018-08-03
+attribution: Spencer Krum, IBM
+weight: 83
+---
+
+To celebrate the 1.0 release and to promote the software to a wider audience, the Istio community is hosting an all day live stream on Twitch on August 17th.
+
+## What is Twitch?
+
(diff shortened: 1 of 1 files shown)


content/blog/2018/istio-twitch-stream/index.md
```diff
@@ -0,0 +1,27 @@
+---
+title: All Day Istio Twitch Stream
+subtitle:
+description: Istio hosting an all day twitch stream to celebrate the 1.0 release
+publishdate: 2018-08-02
+attribution: Spencer Krum, IBM
+weight: 84
```
geeknoid: Make this 83 so the blogs collate properly in the left nav.

linsun:
is this something we encourage local people to participate?  If not, the current content is LGTM.

nibalizer:
@linsun We're looking for istio developers, users, and presenters to come on the show if they are local, but we're asking viewers to watch on the livestream. Does that answer your question?

linsun:
yes thanks @nibalizer!

/lgtm

geeknoid:
You have a few lint errors to deal with.
Also, your publication date should be 8/3 instead of 8/2.

Note that pushing this to master will only publish it to preliminary.istio.io. After you've pushed it, I'll cherry pick it to the release-1.0 branch so it shows up on istio.io.

PR #2141, opened 2018-08-03
HTTPS->TLS in egress task

rshriram:


content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -80,7 +80,7 @@ from within your Istio cluster. In this task you access
       ports:
       - number: 443
         name: https
-        protocol: HTTPS
+        protocol: TLS
```
rcaballeromx: This change can confuse readers since the step text is left unchanged. The step on lines 68-69 is already convoluted. Having the protocol in the example be TLS for the HTTPS `ServiceEntry` without explanation other than "including HTTPS" can cause confusion. I suggest that you change the step text to explain the need of TLS better:
1.  The HTTPS `ServiceEntry` requires a `TLS`  protocol and the TLS `VirtualService`. To create a `ServiceEntry` and a `VirtualService` to allow access to an external HTTPS service using TLS, use the following command:

geeknoid:
@rshriram Do we still want this change?

PR #2176, opened 2018-08-07
Sidecar upgrade needs kubectl apply not replace

ostromart:
https://github.com/istio/istio/issues/7137


geeknoid:
Once this is in, please cherry pick this change to the master branch.

Thanks.

PR #2177, opened 2018-08-07
Sidecar upgrade needs kubectl apply not replace

ostromart:
Cherry pick of https://github.com/istio/istio.github.io/pull/2176


ostromart:
@ayj  any idea why circle is failing? Doesn't seem to be related to this change.

PR #2180, opened 2018-08-08
add Auth to .spelling for blog/2017/0.1-auth translate lint issue

DavadDi:


geeknoid:
I don't think we want this. The translated post needs to be fixed instead.

DavadDi:
@geeknoid The word `Istio Auth` will cause spelling-related errors in articles that have been translated into Chinese, and personally think it would be a good idea to add Auth to the .spelling file. The current period is temporarily working through `Isito Auth`, but it is very unattractive.

PR #2196, opened 2018-08-10
Highlight "Istio multicluster" in example document.

gyliu513:
/cc @sdake @geeknoid

### DIFF
--- a/content/docs/examples/multicluster/gke/index.md
+++ b/content/docs/examples/multicluster/gke/index.md
@@ -7,7 +7,7 @@ keywords: [kubernetes,multicluster]
 
 This example demonstrates how to use Istio's multicluster feature to join 2
 [Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/) clusters together,
-using the [Istio multicluster installation instructions](/docs/setup/kubernetes/multicluster-install/).
+using the [multicluster installation instructions](/docs/setup/kubernetes/multicluster-install/).
 
 ## Before you begin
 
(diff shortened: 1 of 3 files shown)


geeknoid:
Can we just say "multicluster installation instructions". We don't need to say Istio.

gyliu513:
> Can we just say "multicluster installation instructions". We don't need to say Istio.

Done @geeknoid

content/docs/examples/multicluster/gke/index.md
```diff
@@ -7,7 +7,7 @@ keywords: [kubernetes,multicluster]
 
 This example demonstrates how to use Istio's multicluster feature to join 2
 [Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/) clusters together,
-using the [Kubernetes multicluster installation instructions](/docs/setup/kubernetes/multicluster-install/).
+using the [multicluster installation instructions](/docs/setup/kubernetes/multicluster-install/).
```
rcaballeromx: Why are you removing the Kubernetes?
There are other multicluster solutions and readers should know what multicluster installation instructions they will be getting before they click the link.

gyliu513: Thanks @rcaballeromx , let me explain more for this.

The `Kubernetes multicluster installation instructions` makes me think this is a instruction for installing multiple Kubernetes clusters, but the document is actually focusing on how to configure service mesh cross multiple clusters.

As this is highlighting `Istio multicluster configuration` here, and the document is only for `Isito`, so I think it is ok to use `multicluster installation instructions` here.

PR #2199, opened 2018-08-10
fix typesetting error

Whitilied:
[relation](https://istio.io/zh/docs/tasks/traffic-management/ingress/)


Whitilied:
I signed it!

PR #2242, opened 2018-08-13
fixed: outdated reference to the field `httpReqTimeout`

fleeto:


content/docs/reference/config/istio.networking.v1alpha3/index.html
```diff
@@ -1645,7 +1645,7 @@ <h2 id="HTTPRetry">HTTPRetry</h2>
 <td>
 <p>REQUIRED. Number of retries for a given request. The interval
 between retries will be determined automatically (25ms+). Actual
-number of retries attempted depends on the httpReqTimeout.</p>
+number of retries attempted depends on the <em>timeout</em> field of the <a href="#HTTPRoute">route rule</a>.</p>
```
geeknoid: This file is auto-generated from the corresponding protos in istio/api. Please make the change in the proto file instead.

PR #2268, opened 2018-08-16
Egress Gateway mTLS origination advanced example

vadimeisenbergibm:
To demonstrate using additional certificates in `istio-egressgateway` to perform mTLS origination to external services.


content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,533 @@
...
+title: Mutual TLS origination by Egress Gateway
+description: Describes how to configure an Egress Gateway to perform mutual TLS origination to external services.
+weight: 45
+keywords: [traffic-management,egress]
+---
+
+The [Configure an Egress Gateway](/docs/examples/advanced-egress/egress-gateway) example describes how to configure
+Istio to direct the egress traffic through a dedicated service called _Egress Gateway_. This examples shows how to
+configure an Egress Gateway to perform mutual TLS origination to external services.
+You will deploy an [NGINX](https://www.nginx.com/) server in your Kubernetes cluster without injecting Istio sidecar
```
rshriram: @rcaballeromx whats the language format we need to use? `We` or `You` ?

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,533 @@
...
+proxy into it.
+This server will simulate a server outside the Istio service mesh. Then you will configure the Egress Gateway to
+perform mutual TLS with this server.
+Finally, you will direct the traffic from the application pods inside the mesh to the server outside the mesh through
+the Egress Gateway. The Egress Gateway will perform mutual TLS origination with the NGINX server.
+
+## Before you begin
+
+This examples assumes you deployed Istio with [mutual TLS Authentication](/docs/tasks/security/mutual-tls/)
+enabled. Follow the steps in the [Before you begin](/docs/examples/advanced-egress/egress-gateway/#before-you-begin)
```
rshriram: I dont think we need this sentence anymore. mututal tls is incrementally enabled. i.e. its always enabled in some way,

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,533 @@
...
+    spec:
+      hosts:
+      - my-nginx.mesh-external.svc.cluster.local
+      ports:
+      - number: 80
+        name: http-port
+        protocol: HTTP
+      - number: 443
+        name: http-port-for-tls-origination
+        protocol: HTTP
```
rshriram: See my comment in the HTTS blog.. we should have this as HTTPS. and ask applications to initiate connections to port 80, use a virtual service to forward from 80 to 443 and destination rule to initiate mtls to 443. Asking applications to use http://google:443 si very clunky. Lots of SDKs don't have such provisions. They typically allow http/https.

rshriram: Also, we don't need this service entry since the service already exists. In fact, you will have duplicate services because we don't allow service entries to overwrite what was objtained from the service registry

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
geeknoid: Always "you"

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
vadimeisenbergibm: @rshriram 

> Asking applications to use http://google:443 si very clunky

Agreed, I do not ask it here. This application uses port 80, port 443 in this service entry is probably not needed. 

Two issues:
1.  The service entry for the port 80 is required since the original service declares port 443 only. I guess the service entry is ignored for the port 443 where there is a duplication with the Kubernetes service.
2.  The idea of this example is to simulate mTLS to some external service. In that case, the ServiceEntry will be required for both ports. So I wanted to use the correct configuration for an external service, just to demonstrate it with a Kubernetes service.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
vadimeisenbergibm: When configuring a Gateway, the Gateway should know to expect mTLS or not. Otherwise, double instructions must be provided: "if you have mTLS enabled, create this definition, otherwise create that definition".

```
tls:
           mode: MUTUAL
           serverCertificate: /etc/certs/cert-chain.pem
           privateKey: /etc/certs/key.pem
           caCertificates: /etc/certs/root-cert.pem
```

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
vadimeisenbergibm: @rshriram I have fixed the port 443 to have HTTPS protocol and `https` name. I still propose to use the ServiceEntry, so this configuration will work for the cases when the external service is outside the cluster. Here I used an in-cluster out-of-mesh service just for simplicity of demonstration. A more complex demonstration would be to deploy NGINX in some other environment, in some other cluster, for example, but that would complicate the example.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
rcaballeromx: @rshriram We use the second person singular: you, your. However, we do not use the future tense. Thus, line 11 should read:
You deploy ...

rcaballeromx:
@vadimeisenbergibm Is this PR ready for me to review?

vadimeisenbergibm:
@rcaballeromx Yes, please review it. I will recheck the instructions once Istio 1.0.1 is released.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+          - name: nginx-server-certs
+            secret:
+              secretName: nginx-server-certs
+          - name: nginx-ca-certs
+            secret:
+              secretName: nginx-ca-certs
+    EOF
+    {{< /text >}}
+
+1.  Define a `ServiceEntry` and a `VirtualService` for `nginx.example.com` to instruct Istio to direct traffic destined
```
rcaballeromx: Please be specific. From the commands I cannot discern the context of `ServiceEntry` or `VirtualService`. Are they a service, a pod, a destination rule, a configuration of something else, etc? I would also encourage you to have a step for each command since you already have two separate commands with a ton of information for the readers.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+            host: nginx.example.com
+            port:
+              number: 443
+          weight: 100
+    EOF
+    {{< /text >}}
+
+## Deploy a container to test the NGINX deployment
+
+1.  Create Kubernetes [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to hold the client's and CA
```
rcaballeromx: See comment and suggestion for line 65.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+   certificates:
+
+    {{< text bash >}}
+    $ kubectl create secret tls nginx-client-certs --key nginx.example.com/4_client/private/nginx.example.com.key.pem --cert nginx.example.com/4_client/certs/nginx.example.com.cert.pem
+    $ kubectl create secret generic nginx-ca-certs --from-file=nginx.example.com/2_intermediate/certs/ca-chain.cert.pem
+    {{< /text >}}
+
+1.  Deploy the [sleep]({{< github_tree >}}/samples/sleep) sample with mounted client and CA certificates to test sending
+    requests to the NGINX server:
```
rcaballeromx: Thema-rhema structure. Begin with the known, the test, and move to the unknown, how to do it.
1.  To test sending requests to the NGINX server, deploy the [sleep sample]({{< github_tree >}}/samples/sleep) with the mounted client and CA certificates using the following command:

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+          - name: nginx-client-certs
+            secret:
+              secretName: nginx-client-certs
+          - name: nginx-ca-certs
+            secret:
+              secretName: nginx-ca-certs
+    EOF
+    {{< /text >}}
+
+1.  Define an environment variable to hold the name of the `sleep` pod:
```
rcaballeromx: To store the name of the `sleep` pod,  define an environment variable:

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+
+    {{< text bash >}}
+    $ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})
+    {{< /text >}}
+
+1.  Use the deployed [sleep]({{< github_tree >}}/samples/sleep) container to send requests to the NGINX server. Since
```
rcaballeromx: Keep consistency: don't use container and pod interchangeably. To fix, I suggest:
1.  Use the deployed pod with the [sleep sample app]({{< github_tree >}}/samples/sleep)to send requests to the NGINX server.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+    the host `nginx.example.com` does not exist and will not be resolved by DNS, use the `--resolve` option of `curl` to
+    resolve the hostname manually. You can provide any IP to the `--resolve` option, except for `127.0.0.1`, for example
+    `1.1.1.1`.
+    Istio will route the request correctly to your NGINX server. In the real life scenarios, a DNS entry will exist for
+    the destination hostname and providing `--resolve` option to `curl` will not be required.
```
rcaballeromx: Mind the 80 character line length. Future tense and other language issues. To fix, I suggest:
Since the `nginx.example.com` host  does not exist, the DNS cannot resolve the hostname. The following command uses the `--resolve` option of `curl` to resolve the hostname manually. You can provide any IP to the `--resolve` option, except for `127.0.0.1`. If you use, for example, `1.1.1.1`, Istio routes the request correctly to your NGINX server. Normally, a DNS entry exists for the destination hostname and you must not use the `--resolve` option of `curl`.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+    <center><h1>400 Bad Request</h1></center>
+    <center>No required SSL certificate was sent</center>
+    <hr><center>nginx/1.15.2</center>
+    </body>
+    </html>
+    {{< /text >}}
+
+## Redeploy the Egress Gateway with the client certificates
+
+1. Create Kubernetes [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to hold the client's and CA
```
rcaballeromx: See comment on line 65.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+   certificates.
+
+    {{< text bash >}}
+    $ kubectl create -n istio-system secret tls nginx-client-certs --key nginx.example.com/4_client/private/nginx.example.com.key.pem --cert nginx.example.com/4_client/certs/nginx.example.com.cert.pem
+    $ kubectl create -n istio-system secret generic nginx-ca-certs --from-file=nginx.example.com/2_intermediate/certs/ca-chain.cert.pem
+    {{< /text >}}
+
+1.  Generate the `istio-egressgateway` deployment with a volume to be mounted from the new secrets. Use the same options
```
rcaballeromx: Please be specific. In the command the `istio-egressgateway` text can be found multiple times but you are referring to the object after `--name`. A better description of the command would be to configure the deployment of the egress gateway, is it not? I would also refrain from referencing the "options you used for generating your `istio.yaml`. You provide all the options in the command and the sentence would block readers forcing them to try to go back and find those options only to realize you had them in the command all along. I suggest:
1.  Configure the deployment of the egress gateway with the following command. The command includes all the options the `istio-egressgateway` pod requires to mount a volume from the new secrets.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+    --set gateways.istio-egressgateway.secretVolumes[2].name=nginx-client-certs \
+    --set gateways.istio-egressgateway.secretVolumes[2].secretName=nginx-client-certs \
+    --set gateways.istio-egressgateway.secretVolumes[2].mountPath=/etc/nginx-client-certs \
+    --set gateways.istio-egressgateway.secretVolumes[3].name=nginx-ca-certs \
+    --set gateways.istio-egressgateway.secretVolumes[3].secretName=nginx-ca-certs \
+    --set gateways.istio-egressgateway.secretVolumes[3].mountPath=/etc/nginx-ca-certs > \
+    ./istio-egressgateway.yaml
+    {{< /text >}}
+
+1.  Redeploy `istio-egressgateway`:
```
rcaballeromx: Specificity issue. I suggest:
1. Redeploy the `istio-egressgateway` pod with the new configuration:

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+
+    {{< text bash >}}
+    $ kubectl apply -f ./istio-egressgateway.yaml
+    deployment "istio-egressgateway" configured
+    {{< /text >}}
+
+1.  Verify that the key and the certificate are successfully loaded in the `istio-egressgateway` pod:
```
rcaballeromx: Passive voice issue.
1.  Ensure that Kubernetes successfully loaded the key and the certificate in the `istio-egressgateway` pod:

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
+---
+title: Mutual TLS origination by Egress Gateway
+description: Describes how to configure an Egress Gateway to perform mutual TLS origination to external services.
+weight: 45
+keywords: [traffic-management,egress]
+---
+
+The [Configure an Egress Gateway](/docs/examples/advanced-egress/egress-gateway) example describes how to configure
+Istio to direct the egress traffic through a dedicated service called _Egress Gateway_. This examples shows how to
```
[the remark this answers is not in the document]
vadimeisenbergibm: The flow of the introduction is as follows:
1. The previous work, what the reader saw in the previous tasks/examples
2. This example

It follows the pattern of the scientific articles:
1. Previous Work
2. Current state of the research
3. This article
4. Future work

The goal is to introduce the reader to the context, to clarify how this example is different from the others.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+You will deploy an [NGINX](https://www.nginx.com/) server in your Kubernetes cluster without injecting Istio sidecar
+proxy into it. This server will simulate a host `nginx.example.com` outside the Istio service mesh.
+
+Then you will configure the Egress Gateway to perform mutual TLS with this server.
+Finally, you will direct the traffic from the application pods inside the mesh to the server outside the mesh through
+the Egress Gateway. The Egress Gateway will perform mutual TLS origination with the NGINX server.
+
+## Generate client and server certificates and keys
+
+Generate the certificates and keys in the same way as in the [Securing Gateways with HTTPS](/docs/tasks/traffic-management/secure-ingress/#generate-client-and-server-certificates-and-keys).
```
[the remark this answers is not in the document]
vadimeisenbergibm: I have doubts about redundant cross-references. I think that the examples and tasks should be cross referenced as much as possible. The reader should be told - here, you perform the same step as in the example X. This step is like in the example Y, but it is different and so on. The assumption is that the reader walked over multiple examples on istio.io and it could be useful to tell him/her how this examples is different from the others they performed.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+
+1.  Clone the <https://github.com/nicholasjackson/mtls-go-example> repository:
```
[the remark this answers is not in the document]
vadimeisenbergibm: Why not? It is an Open Source with Apache 2.0 license. We do not use the code, we just run the script there.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+
+    {{< text bash >}}
+    $ pushd mtls-go-example
+    {{< /text >}}
+
+1.  Generate the certificates for `nginx.example.com`.
+    Use any password with the following command:
+
+    {{< text bash >}}
+    $ ./generate.sh nginx.example.com <password>
```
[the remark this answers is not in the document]
vadimeisenbergibm: Note the convention https://istio.io/about/contribute/style-guide/#use-angle-brackets-for-placeholders

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+
+1.  Change directory back:
+
+    {{< text bash >}}
+    $ popd
+    {{< /text >}}
+
+## Deploy an NGINX server
+
+1.  Create a namespace `mesh-external` to represent services outside the Istio mesh. Note that the sidecar proxy will
```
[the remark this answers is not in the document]
vadimeisenbergibm: Are you sure about `outside` vs. `outside of`? See https://english.stackexchange.com/questions/9700/outside-or-outside-of

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+    not be automatically injected into the pods in this namespace since the automatic sidecar injection was not
+    [enabled](/docs/setup/kubernetes/sidecar-injection/#deploying-an-app) on it.
```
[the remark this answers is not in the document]
vadimeisenbergibm: Regarding the passive voice - I used it on purpose here. The goal is to spare the reader unnecessary details about who actually injects the sidecar proxy.

See passive voice https://istio.io/docs/setup/kubernetes/sidecar-injection/:

> The controller resource is unmodified. Sidecars can be updated selectively ...

In this case:
> Istio doesn't automatically inject the sidecar proxy into the pods 

Note that it is not Istio that injects the proxy, it is a mechanism in Kubernetes, which is not relevant to the example. The point of the sentence is that the sidecar proxies will not be injected, the deployed pods will not have sidecar proxies and will not be a part of the Istio mesh.

The same regarding:

>  Ensure that the key and the certificate are successfully loaded

vs. 

> Ensure that Kubernetes loaded the key and the certificates...

The passive voice means it does not matter who and how loaded the key/certificate, the reader should just verify that the key/certificate is there, in the file system of the pod.

@geeknoid @frankbu Could you please provide your opinion about the passive voice usage in istio.io ? I propose to use the passive voice to spare the readers unnecessary
technical details.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+
+    {{< text bash >}}
+    $ kubectl create namespace mesh-external
+    {{< /text >}}
+
+1. Create Kubernetes [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to hold the server's and CA
```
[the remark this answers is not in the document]
vadimeisenbergibm: That's an important point to agree upon. In my opinion, the goal of the examples is to provide users with working scenarios **without going too much into details and explanations**. No need to explain here what is Certificate Authority and how it will be used, or that there are two different secrets. I see it as follows - the user just clicks on the instructions one by one and sees the example working. These examples are like a set of recipes - you just follow them to complete a particular task, you do not have to understand all the details.

Also note that this one is an advanced example, at this point the reader should already be familiar with Istio, no need to go into details regarding each secret/virtual service/destination rule here.

@geeknoid @frankbu could you please provide your opinion here? This one is important, I think.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
vadimeisenbergibm: This is related to the point in 
https://github.com/istio/istio.github.io/pull/2268#discussion_r212181444.

> Please be specific. From the commands I cannot discern the context of ServiceEntry or VirtualService

1. Since it is an advanced example, the user should know what is `ServiceEntry` and `VirtualService`.

> I would also encourage you to have a step for each command since you already have two separate commands with a ton of information for the readers.

My goal is not to go into the details, but to let users complete the examples with as less clicks as possible. This is not _information for the readers_, it is a recipe to copy&paste, and to learn later if needed, to modify it for the readers requirements.

@geeknoid @frankbu - what do you think?

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
vadimeisenbergibm: "Store" may imply some persistent storage involved. Environment variables "hold" values, temporary, during the session. See this question on Stack Overflow - https://stackoverflow.com/questions/532155/linux-where-are-environment-variables-stored

Here https://en.wikipedia.org/wiki/Environment_variable stores/holds are used interchangeably. Here https://www.ibm.com/support/knowledgecenter/en/SSB27U_6.4.0/com.ibm.zvm.v640.dmsp2/dmsp248.htm, contains/holds is used, not stores.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
vadimeisenbergibm: "The sentence would block readers forcing them to try to go back and find those options only to realize you had them in the command all along"

Nope, I do not have the additional options the users used. See https://istio.io/docs/setup/kubernetes/helm-install/#customization-example-traffic-management-minimal-set:

> Installation Options has the full list of options allowing you to tailor the Istio installation to your needs.

The full list of options is here - https://istio.io/docs/reference/config/installation-options/. The point in the text is: Use the same options you used to install Istio, plus the options in this command.

vadimeisenbergibm: > A better description of the command would be to configure the deployment of the egress gateway, is it not?

This step generates a spec of `istio-egressgateway` `Deployment`, `Deployment` is a kind of configuration items in Kubernetes, the readers know it since it is an advanced example. Probably the word "spec" is missing here, or maybe the "deployment" should be `Deployment`. 

The next step _deploys_ the spec of the `Deployment`, or in `kubectl` terms _applies_ or _creates_ a `Deployment`.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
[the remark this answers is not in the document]
vadimeisenbergibm: Additional point - the reader may want to go over the examples one by one, so there is some order, some history of the examples.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+
+The [Configure an Egress Gateway](/docs/examples/advanced-egress/egress-gateway) example describes how to configure
+Istio to direct the egress traffic through a dedicated service called _Egress Gateway_. This examples shows how to
+configure an Egress Gateway to perform mutual TLS origination to external services.
+You will deploy an [NGINX](https://www.nginx.com/) server in your Kubernetes cluster without injecting Istio sidecar
+proxy into it. This server will simulate a host `nginx.example.com` outside the Istio service mesh.
+
+Then you will configure the Egress Gateway to perform mutual TLS with this server.
+Finally, you will direct the traffic from the application pods inside the mesh to the server outside the mesh through
+the Egress Gateway. The Egress Gateway will perform mutual TLS origination with the NGINX server.
```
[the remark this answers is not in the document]
vadimeisenbergibm: _outside of_ or _outside_? See https://english.stackexchange.com/questions/9700/outside-or-outside-of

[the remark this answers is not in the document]
vadimeisenbergibm: why not a new paragraph? The first paragraph described what this example shows in one sentence, high level. The second paragraph provides the details what the reader will actually configure.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+    $ git clone https://github.com/nicholasjackson/mtls-go-example
+    {{< /text >}}
+
+1.  Change directory to the cloned repository:
+
+    {{< text bash >}}
+    $ pushd mtls-go-example
+    {{< /text >}}
+
+1.  Generate the certificates for `nginx.example.com`.
```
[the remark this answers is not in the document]
vadimeisenbergibm: Note that the password here is a minor detail. The meaning is:
1. Run the following command to generate the certificates for `nginx.example.com`. BTW, regarding the password - just use any, it does not matter. You have to provide some password since this is how the script works, it doesn't matter for this example.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+
+    When prompted, select `y` for all the questions.
+
+1.  Move the certificates into `nginx.example.com` directory:
+
+    {{< text bash >}}
+    $ mkdir ~+1/nginx.example.com && mv 1_root 2_intermediate 3_application 4_client ~+1/nginx.example.com
+    {{< /text >}}
+
+1.  Change directory back:
```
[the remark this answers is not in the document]
vadimeisenbergibm: There is a directory where the user starts the example, let's call it `/somepath/somedirectory`.
The user clones `mtls-go-example` into this directory and performs `cd mtls-go-example`. Now the user is in `/somepath/somedirectory/mtls-go-example`. The user has to change the directory since the script expects to run inside the cloned directory. The user performs the steps, the directory `/somepath/somedirectory/nginx.example.com` is created. Finally, the user goes back to `/somepath/somedirectory`. 

Maybe a better wording would be: _go back to your previous location_. See a related question https://stackoverflow.com/questions/10655896/get-back-to-the-previous-location-after-cd-command

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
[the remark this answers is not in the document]
vadimeisenbergibm: How about the following?

Create a namespace to represent services outside the Istio mesh, namely `mesh-external`.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
vadimeisenbergibm: I have doubts about this as well. The structure goes as follows - this example is a recipe how to do things. The important thing is what the parts of the recipe do: 1. Create 2. Deploy 3. Send Request. The secondary part is why each part of the recipe is done: to hold certificate, to perform testing, to simulate external service.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
vadimeisenbergibm: I am not aware about the 80 characters line length convention for istio.io. I used 120.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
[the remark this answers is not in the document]
frankbu: @vadimeisenbergibm I don't have a strong opinion on the passive voice question. I think that passive voice can have a tendency to not flow as smoothly, but I understand your wanting to not say things that may imply things that are inaccurate or irrelevant to the task at hand. That said, these tasks are not legal documents, so saying things like "Istio will do ..." is generally understood to mean something will happen because of Istio, not necessarily that Istio will literally do it.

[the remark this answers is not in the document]
vadimeisenbergibm: @frankbu @geeknoid @rcaballeromx Additional point - I think Istio is one of the most complex software systems existing. You have Kubernetes (which is complex in itself) and you have Istio on top of it, and you have interplay of the layers (Kubernetes/Istio), and you may have Istio on top of frameworks other than Kubernetes. The amount of information, multiple configuration items, interplay of Kubernetes and Istio can be overwhelming for the users. In my opinion, we should spare the users any irrelevant information, and also prevent any distraction and confusion.

vadimeisenbergibm:
@rcaballeromx Do you agree with any of my responses to your comments? How should we proceed?

frankbu:
I suggest we go ahead and merge this PR now.
@vadimeisenbergibm: can you sync it with latest master (the document folder is still the old name).
@rcaballeromx: if you have additional concerns, we can address in a followup PR.

content/docs/examples/advanced-egress/egress-gateway-mtls-origination/index.md
```diff
@@ -0,0 +1,563 @@
...
+    {{< /text >}}
+
+1.  Generate the certificates for `nginx.example.com`.
+    Use any password with the following command:
+
+    {{< text bash >}}
+    $ ./generate.sh nginx.example.com <password>
+    {{< /text >}}
+
+    When prompted, select `y` for all the questions.
```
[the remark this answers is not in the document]
vadimeisenbergibm: Done

vadimeisenbergibm:
@frankbu synced with the latest master, any comments are welcome any time.

frankbu:
Force merging because errors circleci errors are unrelated.
@rcaballeromx: if you have additional concerns, we can address in a followup PR.

PR #2273, opened 2018-08-16
Document DNS resolution requirements

elevran:
Service names must be resolved to IP address in any cluster, regardless of endpoint location.


content/docs/setup/kubernetes/multicluster-install/index.md
```diff
@@ -315,6 +315,11 @@ $ ORIGINAL_SVC_MANIFEST=mysvc-v1.yaml
 $ istioctl kube-inject --injectConfigMapName istio-sidecar-injector --meshConfigMapName istio -f ${ORIGINAL_SVC_MANIFEST} | kubectl apply -f -
 {{< /text >}}
 
+## Accessing Services from different clusters
```
geeknoid: Lowercase "services" please.

content/docs/setup/kubernetes/multicluster-install/index.md
```diff
@@ -315,6 +315,11 @@ $ ORIGINAL_SVC_MANIFEST=mysvc-v1.yaml
...
+
+To access a service, the client first needs to resolve the service DNS name to an IP address. Once the name is resolved, Istio will handle routing to service endpoints in any cluster.
+Kubernetes DNS resolution is scoped to the cluster, so you should ensure services objects are defined in every cluster where a client may be running, regardless of the location of their endpoints. The simplest approach is to duplicate (e.g., using `kubectl`) the service object to every cluster. This will ensure that the service name can be resolved in that cluster. Note that since service objects are namespace scoped, cluster should also define the required namespaces in all clusters.
```
gyliu513: s/cluster should also define the required namespaces in all clusters./cluster admin should also define the required namespaces in all clusters./g

rcaballeromx:
Taking a look now, sorry but this week was crazy.

content/docs/setup/kubernetes/multicluster-install/index.md
```diff
@@ -315,6 +315,11 @@ $ ORIGINAL_SVC_MANIFEST=mysvc-v1.yaml
 $ istioctl kube-inject --injectConfigMapName istio-sidecar-injector --meshConfigMapName istio -f ${ORIGINAL_SVC_MANIFEST} | kubectl apply -f -
 {{< /text >}}
 
+## Accessing Services from different clusters
+
+To access a service, the client first needs to resolve the service DNS name to an IP address. Once the name is resolved, Istio will handle routing to service endpoints in any cluster.
```
rcaballeromx: Line length and passive voice issues: These apply everywhere in the document but I will only mention them once. The acceptable line length is 80 characters. Avoid passive voice. Specifically here, you should avoid future tense too.  It creates a grammatical issue by breaking verb tense congruence. Remember to add articles before nouns.
I suggest:
Once the client resolves the name, Istio handles the routing to the service endpoints in any cluster.

content/docs/setup/kubernetes/multicluster-install/index.md
rcaballeromx: This paragraph is very complex and hard to follow. To simplify and make it clearer, I suggest you avoid unnecessary adverbs and adjectives and keep the sentences short and concise. I suggest the following edits:
Kubernetes resolves DNS on a cluster basis. Because DNS resolution is tied to the cluster, you must define the service object in every cluster where a client runs, regardless of the location of the client's endpoints. 
To ensure this is the case, duplicate the service object to every cluster using `kubectl`. Duplication ensures Kubernetes can resolve the service name in the cluster. Since the service objects are tied to namespaces, the definitions in all clusters must include the required namespaces.
Keep in mind that "scoped" should not be used since it is considered ambiguous jargon. My suggestion avoids the term fully and tries to explain the intended meaning with more common words.

elevran:
@rcaballeromx PTAL. Submitted update addressing your input

elevran:
I will resubmit a PR without the merged history changes, focusing on this change only

PR #2277, opened 2018-08-17
 translation:/content/docs/examples/multicluster/index.md

zy475459736:
translation:/content/docs/examples/multicluster/index.md


rootsongjc:
Why there are so many commits with such little files? Can you squash them while committing?

PR #2294, opened 2018-08-19
TLS ingress gateway for HTTPS

vadimeisenbergibm:
Requested by a user in https://github.com/istio/istio/issues/8029.

### DIFF
--- a/content/docs/examples/ingress-for-https/index.md
+++ b/content/docs/examples/ingress-for-https/index.md
@@ -83,7 +83,7 @@ Generate the certificates and keys in the same way as in the [Securing Gateways 
     {{< /text >}}
 
 1.  Create a Kubernetes [ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/)
-to hold the configuration of the NGINX:
+to hold the configuration of the NGINX server:
 
     {{< text bash >}}
     $ kubectl create configmap nginx-configmap  --from-file=nginx.conf=./nginx.conf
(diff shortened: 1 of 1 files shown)


frankbu:
I don't think this belongs as a top-level article in the `Examples` section. How about we rename the `Advanced egress traffic control` subsection to `Advanced gateway traffic control` and put it there?

Also, I think it would be good to get @rcaballeromx to do his English magic on this article too.

vadimeisenbergibm:
> How about we rename the Advanced egress traffic control subsection to Advanced gateway traffic control and put it there?

@rshriram What do you think?

rshriram:
This sort of merged prematurely. 
 1. Please rename the title to SNI passthrough via Ingress gateway or Gateway without TLS termination. The title as it is, is very misleading as most people tend to think of a gateway with HTTPS as a terminating entity. 
 2. This article, when fixed could go into the advanced gateway traffic control section as @frankbu says. But the basic one should have the "basic" https TLS termination, http exposure.

frankbu:
How about:

```
title: Ingress gateway without TLS termination
description: Describes how to configure SNI passthrough for an ingress gateway.
```

frankbu:
My preference is "Advanced gateway traffic control" ... If you create a PR (use /assign or /hold to make sure it does not get merged prematurely), then we can discuss it further there. WDYT?

frankbu:
Good point, but looking at it, I wonder if maybe that it might be better to merge that one into the [Control egress task](https://preliminary.istio.io/docs/tasks/traffic-management/egress/). It's not that advanced anyway, since it doesn't require a gateway, so just adding the TLS origination part as a new section 2 (before the IP bypass section), might be better.

vadimeisenbergibm:
@frankbu Note that https://preliminary.istio.io/docs/examples/advanced-egress/egress-tls-origination is pretty long, it includes motivation for TLS origination. I do not think it would be useful to merge it into Control egress task. We can leave it just there, considering it as a pre-requisite example to the egress gateway example.

frankbu:
Another possibility: maybe it could be moved to Tasks. Then we have 2 ingress tasks and 2 egress tasks. Everything else goes to the advanced examples.

PR #2314, opened 2018-08-20
policy describe error

davidstack:
set “sidecar.istio.io/inject:false” can disable inject


davidstack:
I signed it!

davidstack:
@ayj i  update the desc，tks

PR #2319, opened 2018-08-20
fix localhost links

vadimeisenbergibm:
@geeknoid Sorry, I left those links not fixed. Note that the links check passed for some reason.


geeknoid:
We have an exclusion for localhost: links which is why these weren't detected. I'll see if I can take out the exception to avoid this happening again.

PR #2318, opened 2018-08-20
Update ingress multiple host

jsenon:
Add warning. #8073

PR #2322, opened 2018-08-20
Update Reference: Traffic routing

jsenon:
Add note when you need to configure multiple host gateway with SSL. Linked to #2318


frankbu:
This documentation is generated from comments in the proto file in istio/api. The change needs to be made there.

We may be fixing https://github.com/istio/istio/issues/8073, which will remove this restriction, so we may not need this documented at all.

PR #2333, opened 2018-08-21
add link from authN policy to ingress jwt policy

quanjielin:


content/docs/tasks/security/authn-policy/index.md
```diff
@@ -570,6 +570,9 @@ $ for i in `seq 1 10`; do curl --header "Authorization: Bearer $TOKEN" $INGRESS_
 401
 {{< /text >}}
 
+JWT policy could also be setup on `istio-ingressgateway`, see the
```
liminw: You can add a JWT policy on `istio-ingressgateway`. This is often used in the scenario that you want to define JWT policies that apply at the Gateway level, instead of at individual services. See the ... for more details.

content/docs/tasks/security/authn-policy/index.md
```diff
@@ -570,6 +570,11 @@ $ for i in `seq 1 10`; do curl --header "Authorization: Bearer $TOKEN" $INGRESS_
 401
 {{< /text >}}
 
+You can add a JWT policy on `istio-ingressgateway`. This is often used in the
```
geeknoid: Simplify:

This is often used to define JWT policies that apply at the gateway, instead of for individual services.

quanjielin: thanks, updated.

PR #2342, opened 2018-08-22
Rewrite ingress without tls termination example

vadimeisenbergibm:


frankbu:
@vadimeisenbergibm That's too bad. So, let's just merge this one, and then I'll create a PR myself with a few wording tweaks.

PR #2344, opened 2018-08-22
Replace a few instances of route rule with VirtualService

cmluciano:
Signed-off-by: Christopher M. Luciano <cmluciano@us.ibm.com>

### DIFF
--- a/content/docs/concepts/traffic-management/index.md
+++ b/content/docs/concepts/traffic-management/index.md
@@ -451,7 +451,7 @@ spec:
     timeout: 10s
 {{< /text >}}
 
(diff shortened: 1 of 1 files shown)


frankbu:
Changing only these two places seems kind of arbitrary, since the entire doc uses "route rules" pretty consistently when referring to the configuration in a VirtualService, which is by definition essentially a list of route rules.

I think that if we want to use more precise wording, we need to do it consistently in this doc.

cmluciano:
Yes it's arbitrary because this was the section of the docs I was looking at today. I recognize that we use this elsewhere, but shouldn't we be removing these references moving forward? These are a few that I found so far, but I can skim the rest of the doc to replace more if that helps.

PR #2347, opened 2018-08-22
Consuming External MongoDB Services

vadimeisenbergibm:


vadimeisenbergibm:
@frankbu Could you please review this PR? This is the last major PR regarding egress traffic configuration. It describes TCP and TLS traffic control using MongoDB as an example, with and without egress gateway, and also with wildcarded domains. 

More posts will follow regarding applying Mixer policies on the egress traffic, and maybe some summarizing post about egress traffic control. This one, fortunately, should be the last post regarding the egress **traffic configuration** aspect of the egress traffic control.

frankbu:
@vadimeisenbergibm I'm worried about all the setup part of this example. Seems that the mongodb setup is a large part of the example, and really has nothing to do with what the example is trying to teach. I wonder if maybe we could setup a public mongodb server and just use it in the example. If, not, maybe this would be better suited as a blog post, instead of an example.

content/docs/examples/advanced-gateways/egress-mongo/index.md
```diff
@@ -0,0 +1,1111 @@
...
+      name: mongo
+    spec:
+      hosts:
+      - $MONGODB_HOST
+      addresses:
+      - $MONGODB_IP/32
+      ports:
+      - number: $MONGODB_PORT
+        name: tcp
+        protocol: TCP
```
frankbu: I see that the various examples in this article use `protocol: TCP` or `protocol: TLS`. Given that `MONGO` is one of the [possible values](https://preliminary.istio.io/docs/reference/config/istio.networking.v1alpha3/#Port), shouldn't `protocol: MONGO` be used?

content/docs/examples/advanced-gateways/egress-mongo/index.md
```diff
@@ -0,0 +1,1111 @@
...
+      resolution: DNS
+      location: MESH_EXTERNAL
+    EOF
+    {{< /text >}}
+
+1.  Refresh the web page of the application and verify that the ratings are displayed correctly.
+
+1.  Proceed to the following section.
+
+#### Mutual TLS between the sidecar proxies and the egress gateway
```
frankbu: This seems like the less common, more advanced, case, so I would show it last.

content/docs/examples/advanced-gateways/egress-mongo/index.md
```diff
@@ -0,0 +1,1111 @@
...
+            host: $MONGODB_HOST
+            port:
+              number: $MONGODB_PORT
+          weight: 100
+    EOF
+    {{< /text >}}
+
+1.  Proceed to [Verify that the TCP egress traffic is directed through the egress gateway](#verify-that-the-tcp-egress-traffic-is-directed-through-the-egress-gateway).
+
+#### Direct TCP traffic through an egress gateway without mutual TLS
```
frankbu: Maybe we should switch this one to use the new simpler external network endpoint ServiceEntry approach.

content/docs/examples/advanced-gateways/egress-mongo/index.md
```diff
@@ -0,0 +1,1111 @@
...
+#### Cleanup directing TLS Egress traffic through an egress gateway
+
+{{< text bash >}}
+$ kubectl delete serviceentry mongo
+$ kubectl delete gateway istio-egressgateway
+$ kubectl delete virtualservice direct-mongo-through-egress-gateway
+$ kubectl delete destinationrule egressgateway-for-mongo
+{{< /text >}}
+
+### Enable MongoDB TLS egress traffic to arbitrary wildcarded domains
```
frankbu: This part seems like it really could be a separate example of blog.

content/docs/examples/advanced-gateways/egress-mongo/index.md
```diff
@@ -0,0 +1,1111 @@
...
+1.  Refresh the web page of the application. The application should display the ratings without error.
+
+#### Cleanup of the egress configuration for TLS
+
+{{< text bash >}}
+$ kubectl delete serviceentry mongo
+$ kubectl delete virtualservice mongo
+{{< /text >}}
+
+### Direct TLS Egress traffic through an egress gateway
```
frankbu: This would also be good to show how to do it using the new external network endpoint ServiceEntry approach

content/docs/examples/advanced-gateways/egress-mongo/index.md
frankbu: Also, isn't this significant overlap with the wildcard example?

vadimeisenbergibm:
@frankbu My thoughts about this documentation item:

It is pretty large, it comprises setting up a database for the bookinfo example, configuring the bookinfo example to use the database, and shows all the options for the egress traffic control. There is certain overlap with other tasks/example/blog posts, but I would like to put one documentation item to cover all the egress traffic options, on a single demo example of bookinfo + MongoDB. The only new part in this item is directing TCP traffic thru the egress gateway, all other parts are variations of other egress tasks/examples/blog posts. The difference is in the ports, for example, for HTTPS the same 443 port is used in the existing examples. For the mongodb, you can rewrite the mongodb port to 443 and then rewrite it back to the mongodb port. In other case you have to redeploy egress gateway with the MongoDB port. So I cannot just reference other parts in this item and instruct the reader to follow the instructions in the references, there are a non-trivial games with the ports, which are also hard to debug for a user once something goes wrong. This is why I am in favor of a single documentation item with all the options for the egress traffic.

I do not think it is feasible to create a public mongodb instance, managing load and security of it. Also, the instructions show the readers how to setup their own instance of the MongoDB database, which could valuable in some settings. Consider a reader who already has their database, firewall, environment set to work with a specific database instance. The reader can run bookinfo in their environment and connect to their real database, to POC egress traffic control of the database.

content/docs/examples/advanced-gateways/egress-mongo/index.md
vadimeisenbergibm: Done

content/docs/examples/advanced-gateways/egress-mongo/index.md
vadimeisenbergibm: @frankbu I am still not convinced that the endpoints approach is better. It requires specifying both the endpoints and the virtual services, while it spares the definition of the DestinationRule. In the current approach, you do not specify the endpoints, but specify a DestinationRule instead. I propose that we handle it separately for this documentation item, and for other egress gateway examples.

content/docs/examples/advanced-gateways/egress-mongo/index.md
vadimeisenbergibm: @frankbu Please see my summary comment regarding the both issues
https://github.com/istio/istio.io/pull/2347#issuecomment-438935643.

content/docs/examples/advanced-gateways/egress-mongo/index.md
vadimeisenbergibm: @frankbu Replied here https://github.com/istio/istio.io/pull/2347#discussion_r233744203.

vadimeisenbergibm:
@frankbu Thank you for your feedback, I have addressed your comments. I reverted my previous commits to move this item from blog post to examples, so now it is a blog post again. I can revert the reverting commits back, if needed, to move this item back to examples.

content/docs/examples/advanced-gateways/egress-mongo/index.md
frankbu: I think it would be good to mention `protocol: MONGO` in the article and explain why you are using TCP instead, since readers would probably expect a ServiceEntry for mongodb would use it.

content/blog/2018/egress-mongo/index.md
```diff
@@ -0,0 +1,1120 @@
...
+In the following sections you will configure egress access to the external MongoDB service, using different options for
+egress control in Istio.
+
+## Egress control for TCP
+
+Since [MongoDB Wire Protocol](https://docs.mongodb.com/manual/reference/mongodb-wire-protocol/) runs on top of TCP, you
+can control the egress traffic to your MongoDB as traffic to any other [external TCP service](/blog/2018/egress-tcp/). To
+control TCP traffic, a block of IPs in the [CIDR](https://tools.ietf.org/html/rfc2317) notation that includes the IP
+address of your MongoDB host must be specified. The caveat here is that the IP of the MongoDB host is not stable and
+could not be known in advance.
```
frankbu: As worded, this sentence is basically saying that this approach is not possible since MongoDB IPs are not stable or known. Is that what you meant to say? If not, it should say something like this instead:

```
The caveat here is that sometimes the IP of the MongoDB host is not stable or known in advance.
```

content/blog/2018/egress-mongo/index.md
```diff
@@ -0,0 +1,1120 @@
...
+      resolution: DNS
+      location: MESH_EXTERNAL
+    EOF
+    {{< /text >}}
+
+1.  Refresh the web page of the application and verify that the ratings are displayed correctly.
+
+1.  Proceed to the following section.
+
+#### Direct TCP traffic through an egress gateway
```
frankbu: This heading is almost exactly the same as the previous one. Suggest changing to:

```
#### Configure TCP traffic from sidecars to the egress gateway
```

content/blog/2018/egress-mongo/index.md
```diff
@@ -0,0 +1,1120 @@
...
+single configuration, you use *wildcarded* hosts.
+
+In this section you configure egress traffic for a wildcarded domain, namely `*.com` and direct that traffic through
+an egress gateway. You deploy a custom egress gateway with [an additional SNI proxy](/docs/examples/advanced-gateways/wildcard-https-egress-gateway/#https-traffic-to-arbitrary-wildcarded-domains)
+which you must deploy due to current limitations of Envoy, the proxy of the standard Istio egress gateway.
+
+#### Prepare a new egress gateway with an SNI proxy
+
+In this subsection you deploy an egress gateway with an SNI proxy, in addition to the standard Istio Envoy proxy. You
+can use any SNI proxy that is capable to route traffic according to arbitrary, not-preconfigured SNI values; we used
```
frankbu: capable of routing traffic

content/blog/2018/egress-mongo/index.md
```diff
@@ -0,0 +1,1120 @@
...
+        route:
+        - destination:
+            host: sni-proxy.local
+            port:
+              number: $MONGODB_PORT
+          weight: 100
+    EOF
+    {{< /text >}}
+
+1.  Refresh the web page of the application again and verify that the ratings are still displayed correctly.
```
frankbu: I'm missing something here. How is the application using this config for `*.com`? Isn't the ratings service using `MONGO_DB_URL=mongodb://bookinfo:$BOOKINFO_PASSWORD@$MONGODB_HOST:$MONGODB_PORT/test?authSource=test&ssl=true` for it's requests?

vadimeisenbergibm: @frankbu Good point - the application does not really use multiple mongodb hosts. The usecase is applicable when the cluster has to access multiple MongoDB instances, with urls for example, like: mongodb1.composedb.com, mongodb2.composedb.com. Instead of specifying a configuration per each MongoDB host, the user can specify a single configuration for *.composedb.com. Since the application uses a single MongoDB host, it would be possible to specify exactly that mongodb host, as it was done in the previous section.

This section shows that you can specify `*.composedb.com` to allow the traffic to `mongodb1.composedb.com`. The reader can deploy a second mongodb host, e.g. `mongodb2.composedb.com`, let bookinfo use it and see that everything works with the second host without any reconfiguration, since the wildcard was used. I can propose it in the blog post as an exercise for the reader.

What do you think?

frankbu: Explaining this and leaving it as an exercise for the user sounds good. I think you also need to make clear that `*.com` works for your compose, become the mongodb host you are using ends in `.com` (e.g., `mongodb1.composedb.com`). If someone is trying this out with some other mongodb instance, that ends in `.org` for example, they would need to use something like `*.org` for the config instead. I think you just need to explain what's going on a little better

vadimeisenbergibm: Sounds good, please see https://github.com/istio/istio.io/pull/2347/commits/723e7be8a7a86c40d6362ce7a5c860ef27f39674

content/blog/2018/egress-mongo/index.md
```diff
@@ -0,0 +1,1137 @@
...
+$ kubectl delete serviceentry mongo
+$ kubectl delete gateway istio-egressgateway
+$ kubectl delete virtualservice direct-mongo-through-egress-gateway
+$ kubectl delete destinationrule egressgateway-for-mongo
+{{< /text >}}
+
+### Enable MongoDB TLS egress traffic to arbitrary wildcarded domains
+
+Sometimes you want to configure egress traffic to multiple hostnames from the same domain, for example traffic to all
+MongoDB services from `*.<your company domain>.com`. You do not want to create multiple configuration items, one per each
```
frankbu: one for each

content/blog/2018/egress-mongo/index.md
```diff
@@ -0,0 +1,1137 @@
...
+single configuration, you use *wildcarded* hosts.
+
+In this section you configure egress traffic for a wildcarded domain. I used a MongoDB instance at `composedb.com`
+domain, so configuring egress traffic for `*.com` worked for me (I could have used `*.composedb.com` as well).
+You can pick a wildcarded domain according to your MongoDB host.
+
+As part of the egress traffic configuration you direct that traffic through
+an egress gateway. You deploy a custom egress gateway with
+[an additional SNI proxy](/docs/examples/advanced-gateways/wildcard-https-egress-gateway/#https-traffic-to-arbitrary-wildcarded-domains)
+which you must deploy due to current limitations of Envoy, the proxy of the standard Istio egress gateway.
```
frankbu: This paragraph is poorly worded. How about:

```
To configure egress gateway traffic for a wildcarded domain, you will first need to deploy a custom egress
gateway with [an additional SNI proxy](/docs/examples/advanced-gateways/wildcard-https-egress-gateway/#https-traffic-to-arbitrary-wildcarded-domains).
This is needed due to current limitations of Envoy, the proxy used by the standard Istio egress gateway.
```

content/blog/2018/egress-mongo/index.md
```diff
@@ -0,0 +1,1137 @@
...
+    {{< text bash >}}
+    $ kubectl logs $(kubectl get pod -l istio=egressgateway-with-sni-proxy -n istio-system -o jsonpath='{.items[0].metadata.name}') -n istio-system -c sni-proxy
+    127.0.0.1 [23/Aug/2018:03:28:18 +0000] TCP [<your MongoDB host>]200 1863 482 0.089
+    127.0.0.1 [23/Aug/2018:03:28:18 +0000] TCP [<your MongoDB host>]200 2590 1248 0.095
+    {{< /text >}}
+
+#### Understanding what happened
+
+In this section you configured egress traffic to your MongoDB host using a wildcarded domain. While for a single MongoDB
+host there is no gain in using wildcarded domains (an exact hostname can be specified), it could be beneficial for the
```
frankbu: s/for the/for/

content/blog/2018/egress-mongo/index.md
vadimeisenbergibm: Done in 00434dd2813b34632e379c6c17a128f2956493a1.

content/blog/2018/egress-mongo/index.md
vadimeisenbergibm: Done in 0e8e7d081d4ff5937d63ee7b698b16c2137edde4.

content/blog/2018/egress-mongo/index.md
```diff
@@ -0,0 +1,1137 @@
...
+cases when the applications in the cluster access multiple MongoDB hosts that match some wildcarded domain. For example,
+if the applications access `mongodb1.composedb.com`, `mongodb2.composedb.com` and `mongodb3.composedb.com`, the egress
+traffic can be configured by a single configuration for the wildcarded domain `*.composedb.com`. I will leave it
+as an exercise for the readers to configure their apps to use another instance of MongoDB with the hostname that
+matches the wildcarded domain they used in this section. This way the readers can verify that no additional Istio
+configuration is required. The configuration created in this section will work for multiple hostnames from the same
```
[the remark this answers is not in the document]
vadimeisenbergibm: Done in b9c040aa027411d2e636858fe9aea458d1180ba1.

content/docs/examples/advanced-gateways/egress-mongo/index.md
vadimeisenbergibm: Addressed in 9bac39be5c5fa5b85e322dc7bfe89e8431d16150.

content/blog/2018/egress-mongo/index.md
vadimeisenbergibm: Fixed in 2fb5405ee730e8dc7a5c004a4f8b05476bb6c49c.

content/blog/2018/egress-mongo/index.md
vadimeisenbergibm: Done in 1b66580f290d6c41a54c26f79c3cc33a619adaf8.

content/blog/2018/egress-mongo/index.md
vadimeisenbergibm: Done in ddc0da9cac224db35f685a67c569945dfc418870.

PR #2362, opened 2018-08-23
fix code block

yangchuansheng:


yangchuansheng:
@fleeto Yeah, better move into "/docs/examples/advanced-gateways/" after modifying

PR #2370, opened 2018-08-24
Fix typos

rootsongjc:
Fix typos @fleeto


rootsongjc:
@fleeto I think we can merge it.

PR #2373, opened 2018-08-24
fix code block

yangchuansheng:


gyliu513:
@yangchuansheng  I would like to see some descriptin why this is needed, such as you can add some commit message in description.

PR #2378, opened 2018-08-24
Change `mTLS` in YAML config examples to `mtls`

bluk:
- The `mtls` key is case sensitive.

If you use:
```
apiVersion: "authentication.istio.io/v1alpha1"
kind: "MeshPolicy"
metadata:
  name: "default"
spec:
  peers:
  - mTLS: {}
```

You will get an error:

```
- mTLS: {}
 unknown field "mTLS" in v1alpha1.PeerAuthenticationMethod
```

Changing it to:

```
apiVersion: "authentication.istio.io/v1alpha1"
kind: "MeshPolicy"
metadata:
  name: "default"
spec:
  peers:
  - mtls: {}
```

works.

### DIFF
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -492,7 +492,7 @@ recommendations to avoid disruption when updating your authentication policies:
 
 {{< text yaml >}}
 peers:
-- mTLS:
+- mtls:
     mode: PERMISSIVE
 {{< /text >}}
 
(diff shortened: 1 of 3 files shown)


sdake:
Thanks @bluk.  I'm not sure where `MeshPolicy` comes from - it is not in the api repo.  Instead `Policy` is in the repository.  Policy is definitely documented - but I'm not a proto expert.

The peers.mtls field is definitely lowercase:
https://github.com/istio/api/blob/release-1.0/authentication/v1alpha1/policy.proto#L272

/lgtm
/approve

PR #2391, opened 2018-08-25
Enlarge command container

ivanbeldad:
Make command container larger so it can be scrolled in phones easily.
Fix background color when output is wider than screen (commands).

It was almost impossible to scroll in mobile devices due to the small size.

**Without output**
_Before_
istio_before
_After_
istio_after

**With output**
_Before_
istio_multi_before
_After_
istio_multi_after


geeknoid:
Thanks for the fixes. I redid this PR so that these changes only apply to the mobile case, and did a few other mobile improvements as well. See #2397.

Issue #2136, opened 2018-08-03
oraganize helm configuration table


linsun:
From the helm doc file (https://preliminary.istio.io/docs/reference/config/installation-options/ ), the configuration is a huge table. We should categorize them a bit, based on category and common configurations maybe listed on the top for each category and make them bold...

Is it possible to automate this in the py file ( https://github.com/istio/istio.github.io/blob/master/scripts/tablegen.py)?

sdake:
@linsun as I was working on a refactor for tablegen.py that I had to table for a short while, I was thinking it would be helpful to render the docs with each major section as a header (e.g.)

## Global Variables

A table

## Security variables

A table

`...`

Unfortunately I am wrapped up in other things at present, and can't immediately work on this.  I'll add to my backlog.

Cheers
-steve

PR #2399, opened 2018-08-26
Fixes: #2136

sdake:
This is a first take at a table-creating tablegen that categorizes
the output of tablegen.py.  Still needed is the actual text that
goes in the description field.

### DIFF
--- a/scripts/tablegen.py
+++ b/scripts/tablegen.py
@@ -147,7 +147,7 @@ with open('index.md', 'r') as f:
 
         # Print encoded string dictionary
         for k, v in od.items():
-            print ("## `%s` configuration options\n" % k)
+            print ("## `%s` options\n" % k)
             print '| Key | Default Value | Description |'
             print '| --- | --- | --- |'
             for value in v:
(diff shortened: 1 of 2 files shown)


scripts/tablegen.py
```diff
@@ -120,17 +131,29 @@ def decode_helm_yaml(s):
...
         code = yaml.load(d_v)
         yaml.explicit_start = True
         yaml.dump(code, sys.stdout, transform=decode_helm_yaml)
 
+        # Order the encoded string dictionary
+        od = collections.OrderedDict(sorted(prdict.items(), key=lambda t: t[0]))
+
+        # Print encoded string dictionary
+        for k, v in od.items():
+            print ("## `%s` configuration options\n" % k)
```
geeknoid: I'd drop the 'configuration' string in there, it's redundant and causes too much wrapping in the right-hand TOC.

sdake: done.

linsun:
thanks Steve!

/lgtm

PR #2427, opened 2018-08-27
Adds PlanGrid to the Istio in Action page

PatrickAuld:


PatrickAuld:
The two sites that failed linting appear to just be down currently. `http://www.graphviz.org/` and `http://www.libressl.org` (which is ironically not https...). I not sure how to ignore those errors and don't have permissions to kick off another Circleci build to see if they are available again.

@sdake let me know if there are changes I can make on my side to get around the failure.

PR #2439, opened 2018-08-28
fix css style

songrgg:


geeknoid:
Thanks for the bug fix, but this is a generated file. Instead of this change, the source material needs to be corrected. In this case, it would be the file mixer/template/reportnothing/template.proto in the istio repo.

PR #2444, opened 2018-08-28
Add a note about mTLS in the Circuit Breaking task

jwendell:
If mTLS is enabled we need an additional instruction in the
DestinationRule object, otherwise we break traffic to httpbin
service.
    
While on that, also change the Mirroring task note to be the same.

### DIFF
--- a/content/docs/tasks/traffic-management/circuit-breaking/index.md
+++ b/content/docs/tasks/traffic-management/circuit-breaking/index.md
@@ -40,6 +40,8 @@ configuration by intentionally "tripping" the circuit breaker.
 1.  Create a [destination rule](/docs/reference/config/istio.networking.v1alpha3/#DestinationRule) to apply circuit breaking settings
 when calling the `httpbin` service:
 
+    > If you installed/configured Istio with mutual TLS Authentication enabled, you must add a TLS traffic policy `mode: ISTIO_MUTUAL` to the `DestinationRule` before applying it. Otherwise requests will generate 503 errors as described [here](/help/ops/traffic-management/deploy-guidelines/#503-errors-after-setting-destination-rule).
+
     {{< text bash >}}
     $ cat <<EOF | kubectl apply -f -
     apiVersion: networking.istio.io/v1alpha3
--- a/content/docs/tasks/traffic-management/mirroring/index.md
+++ b/content/docs/tasks/traffic-management/mirroring/index.md
@@ -125,6 +125,8 @@ In this step, you will change that behavior so that all traffic goes to `v1`.
 
 1.  Create a default route rule to route all traffic to `v1` of the service:
 
+    > If you installed/configured Istio with mutual TLS Authentication enabled, you must add a TLS traffic policy `mode: ISTIO_MUTUAL` to the `DestinationRule` before applying it. Otherwise requests will generate 503 errors as described [here](/help/ops/traffic-management/deploy-guidelines/#503-errors-after-setting-destination-rule).
+
     {{< text bash >}}
     $ cat <<EOF | kubectl apply -f -
     apiVersion: networking.istio.io/v1alpha3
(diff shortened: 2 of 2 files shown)


content/docs/tasks/traffic-management/circuit-breaking/index.md
```diff
@@ -40,6 +40,36 @@ configuration by intentionally "tripping" the circuit breaker.
...
+          baseEjectionTime: 3m
+          maxEjectionPercent: 100
+        tls:
+          mode: ISTIO_MUTUAL
+
+    EOF
+    {{< /text >}}
+
+    otherwise, run this command:
+
```
frankbu: The mirroring task had this same issue, but instead of duplicating the whole `DestinationRule` we just added a note.

https://preliminary.istio.io/docs/tasks/traffic-management/mirroring/#creating-a-default-routing-policy

I think we should do the same thing here, but the note should be a little more clear. I would suggest the following:

> Note: If you installed/configured Istio with mutual TLS Authentication enabled, you must add a TLS traffic policy `mode: ISTIO_MUTUAL` to the `DestinationRule` before applying it. Otherwise requests will generate 503 errors as described [here](https://istio.io/help/ops/traffic-management/deploy-guidelines/#503-errors-after-setting-destination-rule).

andraxylia:
Could you please fix also the typo "DesintationRule" here: https://istio.io/help/ops/traffic-management/deploy-guidelines/#503-errors-after-setting-destination-rule

content/docs/tasks/traffic-management/circuit-breaking/index.md
jwendell: I thought about adding a note too, but for the novice user, it might be confusing to add that piece of config in the main DestinationRule object. It breaks the copy&paste model of following a tutorial. A better alternative (IMO) which has less clutter would be having tabs, for example, like [this](https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer).

Do we have support for tabs?

andraxylia: Having those tabs would be nice @geeknoid

geeknoid: We don't support for these kinds of tabs at this point, but there is definitely a need. I should be able to put something like that together reasonably quickly though. I might even be able to steal some logic from the k8s doc repo in fact :-)

jwendell: So, can we go ahead with those two slightly different DestinationRules objects and use tabs when they are available? Or do you still prefer having the note?

frankbu: I think we should fix both tasks (circuit-breaker and mirroring) the same way and I think we should have the note that I suggested in the doc whether or not we have tabs. Even if we have tabs, I think that the note is important to help people figure out which tab to use, they might not even be sure if they have mTLS enabled or not. The note I suggested also has a link to the doc that helps users understand what happened if they get the 503s.

So, my proposal is that we should add the note for now (both tasks) and wait until we have tabs to add the second yaml. Adding it without tabs is too much noise IMO and I think most beginners start without mTLS, so it's not too bad for now. As soon as we have tabs, I agree it would be good to put both yamls in, but like I said, leave the note there as well.

jwendell:
@frankbu done.

content/docs/tasks/traffic-management/mirroring/index.md
```diff
@@ -157,7 +157,7 @@ In this step, you will change that behavior so that all traffic goes to `v1`.
     EOF
     {{< /text >}}
 
-    > NOTE: If you installed/configured Istio with mutual TLS Authentication enabled, you must add the [TLSSettings.TLSmode]( /docs/reference/config/istio.networking.v1alpha3/#TLSSettings-TLSmode), `mode: ISTIO_MUTUAL` as noted in the [TLSSettings](/docs/reference/config/istio.networking.v1alpha3/#TLSSettings) reference.
+    > Note: If you installed/configured Istio with mutual TLS Authentication enabled, you must add a TLS traffic policy mode: ISTIO_MUTUAL to the DestinationRule before applying it. Otherwise requests will generate 503 errors as described [here](/help/ops/traffic-management/deploy-guidelines/#503-errors-after-setting-destination-rule).
```
sdake: Please remove the word `Note:` . The `>` will block quote this section.

jwendell:
@sdake done.

PR #2482, opened 2018-08-31
Control Ingress Traffic clean up command error

249043822:
delete virtualservice or gateway should use istioctl, using kuberctl will encounter an error:
`[root@localhost istio-1.1.0.snapshot.0]#  kubectl delete gateway httpbin-gateway
Error from server (NotFound): gateways.config.istio.io "httpbin-gateway" not found
`
`[root@localhost istio-1.1.0.snapshot.0]# istioctl delete Gateway httpbin-gateway
Deleted config: Gateway httpbin-gateway
`


frankbu:
`istioctl delete` (along with other CRUD operations) is going to be deprecated and removed soon. `kubectl delete` is the command that should be used and should work. If not, there is probably something wrong in your environment.

PR #2547, opened 2018-09-06
Added default mTLS behavior with no MeshPolicy.

### LINKED ISSUE #2542: Default mutual TLS setting is unclear in documentation
When no MeshPolicy is defined, mTLS is disabled in the cluster by default.

While the default behavior is implied by explaining how to enable mTLS for the cluster with a MeshPolicy (see [Policy Storage Scope](https://istio.io/docs/concepts/security/#policy-storage-scope) and [Globally enabling Istio mTLS](https://istio.io/docs/concepts/security/#policy-storage-scope)), the default behavior is not explicitly documented. This is important information that would further clarify the use of the MeshPolicy resource.

m-eaton:
Fixes #2542

The default mutual TLS behavior when no MeshPolicy is defined is implied, but
not explicitly documented. This default behavior documentation has been added
under Concepts -> Security -> Authentication in the Policy storage scope
section.

### DIFF
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -367,7 +367,9 @@ namespace. Policies in mesh-scope can affect all services in the mesh. To
 prevent conflict and misuse, only one policy can be defined in mesh-scope
 storage. That policy must be named `default` and have an empty
 `targets:` section. You can find more information on our
-[target selectors section](/docs/concepts/security/#target-selectors).
+[target selectors section](/docs/concepts/security/#target-selectors). By
+default if a mesh-scoped policy is not installed, mutual TLS is considered
+disabled.
 
 Kubernetes currently implements the Istio configuration on Custom Resource
 Definitions (CRDs). These CRDs correspond to namespace-scope and


content/docs/concepts/security/index.md
```diff
@@ -367,7 +367,9 @@ namespace. Policies in mesh-scope can affect all services in the mesh. To
 prevent conflict and misuse, only one policy can be defined in mesh-scope
 storage. That policy must be named `default` and have an empty
 `targets:` section. You can find more information on our
-[target selectors section](/docs/concepts/security/#target-selectors).
+[target selectors section](/docs/concepts/security/#target-selectors). By
+default if a mesh-scoped policy is not installed, mutual TLS is considered
+disabled.
```
linsun: I think this message is vague.  It would be helpful to specify how the mesh-scoped policy is not installed?  I assume it is one of the helm flag or other ways.

sdake: @linsun a mesh scoped policy is covered in this document: https://istio.io/docs/tasks/security/authn-policy/#globally-enabling-istio-mutual-tls.  Its simply a `MeshPolicy` CR.  Another way to word this sentence would be

```
If a global `MeshPolicy` resource is not installed, mutual TLS is disabled by default
```

Of course, this isn't quite accurate, as with `--set global.enableControlPlaneSecurity=true` the control plane ends up with a bunch of MUTUAL_TLS configuration.  It is also possible to specify per namespace and per application TLS policies, so I agree the message is vague and unconvincing given the other documentation :)

m-eaton: @linsun @sdake do you have a suggestion here for a preferred update to the docs?

rcaballeromx: Please write in active voice, based on @sdake's text, I suggest:
You can only install a global data-plane mutual TLS policy with the \`--set global.mtls.enabled=true\` Helm installation flag. If you do not specify this flag, Istio does not enable the mutual TLS for the data plane by default.

geeknoid:
Closing due to inactivity. Please reopen if you intend to push this change in.

PR #2561, opened 2018-09-09
zh: reference template quota and front matters of each entries.

fleeto:


rootsongjc:
@fleeto This branch has conflicts that must be resolved

PR #2597, opened 2018-09-12
Improve the release note link shortcode.

geeknoid:
- When displaying a "Changes" button, we now display a diff from the previous patch to the current
patch, instead of a diff from the base release to the current patch.

- Properly localize the release note buttons so these can be translated.


linsun:
preview looks good 
/lgtm

PR #2613, opened 2018-09-14
add Huawei to the partner list

lichuqiang:
We a team of folks from Huawei would like to (and have been) involved in istio contribution.
So wanna add Huawei to the partner list here.

/assign @geeknoid


rootsongjc:
@lichuqiang You also need to submit a PR to the Chinese content.

PR #2616, opened 2018-09-14
WIP: Grab charts and create helm repository

aslafy-z:
Following up https://github.com/istio/istio/issues/6877 https://github.com/istio/istio/pull/6878

Helm repo folder: https://github.com/aslafy-z/istio.github.io/tree/feature/helm-repo/static/charts
Target urls: 
- github pages: https://istio.github.io/static/charts
- netlify: https://istio.io/charts

Note that the first url will be available after this pr has been merged and the second after someone triggered the deploy manually.


scripts/gen_helm_repo.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# This script generate and copies helm charts within the helm tree of this repo.
+
+ISTIO_BASE=$(cd "$(dirname "$0")" ; pwd -P)/..
+export GOPATH=$(mktemp -d)
+WORK_DIR=${GOPATH}/src/istio.io
+CHART_OUTPUT_DIR=$ISTIO_BASE/charts
```
geeknoid: I presume this stuff should be at the root of the web site? So https://istio.io/chars? If so, then the output needs to be put into the "public" directory, which is the root of the tree that gets published.

"public" is generated when you run "make gen".

aslafy-z: I added a command to the CI pipeline to copy ./charts to ./public/charts.
Does this fits you ?

geeknoid:
Oh wait a minute, this won't work.

We only use CircleCi for validation. What builds the site we actually publish is Netlify, triggered by "make netlify".

How about we switch this to work like grab_reference_docs.sh, say grab_charts.sh. We manually run this script and it puts the charts in static/charts, and this gets checked into GitHub. Stuff in static gets copied as-is to the generated web site.

So once in a while, we run grab_reference_docs.sh and grab_charts.sh to refresh the site content, and we're all set.

geeknoid:
Cool. This seems to be working.

Do you still want to do more, or can this go in?

aslafy-z:
Should I assume that grab_charts.sh will be ran with helm installed ?

geeknoid:
Yes, I think assuming Helm is installed is reasonable. Although is there a docker image for Helm? If there is, running from Docker is better. Otherwise, running the straight binary is OK too.

aslafy-z:
I can't find any official docker image.
If we don't need to stick to official, we can use this automaticaly-build image from docker hub: https://hub.docker.com/r/devth/helm/

geeknoid:
I think using that image would be fine.

linsun:
@aslafy-z could you pls provide a convenient preview link?  I can't find where the statics/chart link.

aslafy-z:
I updated OP @linsun

geeknoid:
This looks great. Thanks.

geeknoid:
Done, this is now on istio.io.

Note that when I reran the script against the release-1.0, it produced charts labeled as 1.0.1 and not 1.0.2. Is that expected?

aslafy-z:
It expected as
https://github.com/istio/istio/blob/release-1.0/install/kubernetes/helm/istio/Chart.yaml
Defines version as 1.0.1. Helm used this version field to label releases.

Can't find this on either istio.io/charts and istio.github.io/static/charts
actually, maybe it's due to some static cache on both ends ?

sdake:
@geeknoid ya - its a known issue with 1.0.2.  1.0.3 fixes the problem for 1.0.3.

Cheers
-steve

errordeveloper:
I tried to make a quick PR to update this, but the shell script didn't just work on Docker for Mac... it'd be great to automate chart repo updates!

aslafy-z:
This has been dropped and replaced by https://github.com/istio/istio.io/pull/2842

PR #2630, opened 2018-09-16
Fix plugin cert doc and improve helm instructions

myidpt:


content/docs/tasks/security/plugin-ca-cert/index.md
```diff
@@ -45,9 +50,15 @@ The following steps enable plugging in the certificates and key into Citadel:
         --from-file=samples/certs/cert-chain.pem
     {{< /text >}}
 
-1.  Redeploy Citadel, which reads the certificates and key from the secret-mount files by using [Helm](/docs/setup/kubernetes/helm-install/)
+1.  Redeploy Citadel, which reads the certificates and key from the secret-mount files by using Helm
     with `global.mtls.enabled` set to `true` and `security.selfSigned` to `false`.
 
+    {{< text bash >}}
+    $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system -x charts/security/templates/deployment.yaml \
+    --set global.mtls.enabled=true,security.selfSigned=false > $HOME/citadel-plugin-cert.yaml
```
sdake: I would recommend two separate --set operations here, rather than jamming it into one set operation.  I am not sure one set operation works - it may, but its a little harder to read.

Cheers
-steve

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -62,6 +62,12 @@ via `kubectl apply`, and wait a few seconds for the CRDs to be committed in the
     $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system > $HOME/istio.yaml
     {{< /text >}}
 
+    > If you want to enable global mutual TLS, set `global.mtls.enabled` to `true`:
```
geeknoid: I don't think you need the leading >

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -93,6 +99,12 @@ to manage the lifecycle of Istio.
     $ helm install install/kubernetes/helm/istio --name istio --namespace istio-system
     {{< /text >}}
 
+    > If you want to enable global mutual TLS, set `global.mtls.enabled` to `true`:
```
geeknoid: Same here, no need for the >

There's another preexisting instance of the same pattern in this file which should be removed.

content/docs/tasks/security/plugin-ca-cert/index.md
```diff
@@ -45,9 +50,15 @@ The following steps enable plugging in the certificates and key into Citadel:
         --from-file=samples/certs/cert-chain.pem
     {{< /text >}}
 
-1.  Redeploy Citadel, which reads the certificates and key from the secret-mount files by using [Helm](/docs/setup/kubernetes/helm-install/)
+1.  Redeploy Citadel, which reads the certificates and key from the secret-mount files by using Helm
```
geeknoid: This implies Citadel uses Helm to read the files. I think what you want instead is:

Redeploy Citadel using Helm. Citadel reads certificates...

content/docs/setup/kubernetes/helm-install/index.md
myidpt: Done.

content/docs/setup/kubernetes/helm-install/index.md
myidpt: Done.

content/docs/tasks/security/plugin-ca-cert/index.md
myidpt: Done. Thanks for pointing this out.

content/docs/tasks/security/plugin-ca-cert/index.md
myidpt: Done.

content/docs/setup/kubernetes/helm-install/index.md
linsun: It may be helpful to explain (or at least point a link) why someone wants to set global.mtls.enabled=true.

I believe this isn't what we recommend to users unless they absolutely know what they are doing and have got their app running without mTLS.

content/docs/tasks/security/plugin-ca-cert/index.md
linsun: is security.selfSigned default to false today?  I don't see any mentioning of it in the install doc being revised above.

myidpt: We do have "By default, Citadel generates self-signed root certificate and key" at the beginning of the doc.

content/docs/setup/kubernetes/helm-install/index.md
myidpt: Sure. added the links to our concept doc.

PR #2633, opened 2018-09-16
Site improvements

geeknoid:
- Use a new approach to managing icons. This has two primary benefits:

  - It makes it possible to color the icons such that they look good in the
    dark theme. Previously, the icons were rendered in black on dark grey when
    using the dark theme.

  - The average payload size for our web pages is reduced and we better use the
    browser cache.

- The new icon approach makes it possible to remove our dependency on the fontawesome
  package, which further slims down our payload requirement

- Refresh our iconography for a slightly lighter look.

- Remove the extra thick left-hand border of text blocks to lighten the
look.

- Added a "NN minutes to read" indication on top of each page. This is
only displayed if the count is > 1 minute.

- Added a calendar icon next to the blog post date.

- Exposed a bunch of strings that were buried in CSS/JS to translation.

- Add the 'keywords:' front-matter fields to the Hugo archetypes.


geeknoid:
@sdake If you could just review the aesthetic to make sure everything looks right.

Thanks.

sdake:
@geeknoid under blog posts, this icon looks a little off:

screen shot 2018-09-17 at 8 17 58 am

- dark theme works well
- icon changes look good (minus above mentioned one)
- timed minutes to read of a few pages - fairly accurate
- not sure I spotted this change in an A/B comparison: `Remove the extra thick left-hand border of text blocks to lighten the ...`

Everything looks good to me @geeknoid - minus that one icon in dark mode.

IMO merge - and address icon separately.

sdake:
The icons are a bit bright in dark mode:

screen shot 2018-09-17 at 8 34 26 am

sdake:
The purple color usage could be a shade brighter.  I have poor vision and its a little challenging to read in dark mode:

screen shot 2018-09-17 at 8 35 45 am

geeknoid:
Thanks for the great feedback. I'll address these in the next round of improvements.

PR #2647, opened 2018-09-19
Fix the error code when running the end-user authentication guide [release 1.0 branch]

lei-tang:
Fix the error code when running the end-user authentication guide [release 1.0 branch]


diemtvu:
/lgtm
 
Can you please make these corrections for master .

lei-tang:
@diemtvu Checked the master and the error code on the master branch is correct: https://preliminary.istio.io/docs/tasks/security/authn-policy/#end-user-authentication

PR #2668, opened 2018-09-21
Add a page about the open source project.

geeknoid:


content/about/community/project/index.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Open Source
+description: How we create Istio.
+weight: 40
+icon: construction
+---
+
+The Istio software is a result of the [Istio open source project](https://github.com/istio). You can learn about how we run our
+project or get involved by visiting the [community repository on GitHub](https://github.com/istio/community). There you can learn about
```
rcaballeromx: I would suggest you make the different areas inside the community repo as a list
To learn about how we run our project or get involved, visit the [community repository on GitHub](https://github.com/istio/community) There, you can find information on:
* Our project's [engineering values](https://github.com/istio/community/blob/master/VALUES.md)

smawson:
Do you have a preview of this on the site?

Also, we already have a link from https://istio.io/about/community/ (which is I think the join page) to the github project, maybe we should just expand that section a bit? We also have a link to working groups from there too. So not sure how the new page plays with the existing ones.

geeknoid:
@smawson Good point, this does have redundancy with the other page. Maybe I should separate this in two then, one page for the user community and one page for the developer community. I'll noodle on this.

Regarding previews, for doc PRs there is always an automatic preview available by click on the netlify link within the "checks" section.

PR #2690, opened 2018-09-25
Add support for the last_update front-matter field.

geeknoid:


i18n/zh.toml
```diff
@@ -85,6 +85,9 @@ other = "Privacy Policy"
 [page_publish_date_format]
 other = "January 2, 2006"
 
+[page_last_update_format]
```
sdake: should a bug be filed to translate this page?

PR #2694, opened 2018-09-26
some little mistakes

pandaamanda:


geeknoid:
These are auto-generated files. The original source material is in istio/istio/mixer/adapters/*. Would you mind applying those fixes there?

Thanks

PR #2704, opened 2018-09-28
Add support for partner adapters.

geeknoid:
Partner adapters (adapters not controlled by the Istio org) are now
called out as such on their page. Adapter author have a chance to put
all sorts of info about their adapter to make it easy for customers
to find and use their adapter.


theganyo:
The example looks great, @geeknoid

PR #2708, opened 2018-09-30
Remove extra-config for cert-manager

rstarmer:
Minikube does the right thing (as of 0.28.1 at least) with creating the embedded CA.  The extra-config parameters appear to have been necessary previously and were resolved to use the "right" credentials built by Minikube directly. In fact, passing those parameters appears to break current minikube deployments, making it impossible to create new service accounts and resources that rely on them. (like a tiller service account for a helm deployment of Istio...)

I found this bug that referenced this issue: https://github.com/kubernetes/minikube/issues/1647 which is now closed.

### DIFF
--- a/content/docs/setup/kubernetes/platform-setup/minikube/index.md
+++ b/content/docs/setup/kubernetes/platform-setup/minikube/index.md
@@ -9,7 +9,7 @@ keywords: [platform-setup,kubernetes,minikube]
 Follow these instructions to prepare Minikube for Istio.
 
 1. To run Istio locally, install the latest version of
-   [Minikube](https://kubernetes.io/docs/setup/minikube/), version **0.28.0 or
+   [Minikube](https://kubernetes.io/docs/setup/minikube/), version **0.28.1 or
    later**.
 
 1. Select a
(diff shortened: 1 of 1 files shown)


sdake:
@rstarmer you will need to fill out the CLA to have this work merged.

Cheers
-steve

rstarmer:
Well, my experience (minikube 0.29.0), was that pilot would end up in a
crashloop complaining about a lack of access to the paths defined. I didn't
dig much further, removed the paths, and everything works...

Robert

rstarmer:
Have done, apparently I submitted the patch on my personal github account,
so I signed a personal one, and one for my company too.
Robert

andraxylia:
My reasoning was we should advise the same config as tested by circleci. Change was in this PR:

https://github.com/istio/istio.io/pull/2082

At the time, I said  "Checked with ayj and we still need it if we use self-signed certificates or k8s CA. It's safer to leave it there."

Maybe we need them in some case and not in other cases?

rstarmer:
I wonder if there is additional configuration needed then, because when I
launch minikube (0.29.0) with the extra commands creating a 1.10.0 K8S
system, the environment doesn't completely form properly. For example, you
can create a service account for helm (which is really a DB operation), but
tiller doesn't launch. And in fact, it appears that other control plane
services aren't properly running.

Robert

adarshaj:
For those who are coming from the istio.io 's page, which doesn't yet reflect this change. The latest minikube's path for certs had to be updated from `/var/lib/localkube` to `/var/lib/minikube`.. (which isn't necessary as demonstrated by this PR), however just leaving here the command that worked for me:

```bash
minikube start --memory=8192 --cpus=4 --kubernetes-version=v1.12.1 \
    --extra-config=controller-manager.cluster-signing-cert-file="/var/lib/minikube/certs/ca.crt" \
    --extra-config=controller-manager.cluster-signing-key-file="/var/lib/minikube/certs/ca.key" \
    --vm-driver=kvm2
```

CodeJjang:
@adarshaj I don't see `/var/lib/minikube...` either in version v.0.30.0.  
Where is it located now?

PR #2712, opened 2018-10-01
sync latest IBM Cloud change from master to 1.0

linsun:


content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -114,7 +114,13 @@ This guide installs the current release version of Istio.
     $ helm del istio --purge
     {{< /text >}}
 
-1. Delete the Istio custom resource definitions:
+    If your Helm version is less than 2.9.0, then you need to manually cleanup extra job resource before redeploy new version of Istio chart:
```
frankbu: Not sure I understand this. What exactly is this saying?

... then you need to manually remove another job resource. Otherwise, if you try to redeploy Istio in the same namespace, it will fail.

linsun: I copied this from https://istio.io/docs/setup/kubernetes/helm-install/#uninstall :)   I think it means helm prior to 2.9 won't clean up the jobs for you automatically thus you'd have to clean it up manually.

PR #2715, opened 2018-10-01
Add Kiali Task to istio.io

jmazzitelli:
This adds a Task to introduce Kiali.
This also adds a deprecation notice to the Service Graph Task.

See Issue https://github.com/istio/istio/issues/9066


content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,314 @@
...
+## About the Kiali API
+
+In addition to visualizing your mesh and Istio configuration, [Kiali](https://www.kiali.io)
+also provides API endpoints for generating JSON representations of what the Kiali
+user-interface visualizes.
+
+For example, if you make an request to `https://$GATEWAY_URL/api/namespaces/bookinfo/graph?graphType=app`,
+you will get a JSON representation of your graph:
+
+```
```
geeknoid: Add `json` annotation

geeknoid: Actually, this should use the {{< text json >}} annotation

content/docs/tasks/telemetry/servicegraph/index.md
```diff
@@ -1,10 +1,13 @@
 ---
-title: Generating a Service Graph
+title: Generating a Service Graph*
 description: This task shows you how to generate a graph of services within an Istio mesh.
 weight: 50
 keywords: [telemetry,visualization]
 ---
 
+> \* Note: The Service Graph has been deprecated.
```
sdake: I'd recommend a warning icon here:

https://preliminary.istio.io/about/contribute/writing-a-new-topic/#adding-icons-emojis

`{{ warning }} The Service graph has been deprecated and replaced ...`

content/docs/tasks/telemetry/servicegraph/index.md
```diff
@@ -1,10 +1,13 @@
...
+> It has been replaced with [Kiali](../kiali).
```
sdake: this note in general should go in the release notes:

https://preliminary.istio.io/about/notes/

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,314 @@
...
+as well as your Istio configuration objects such as virtual services
+and destination rules. You will also use the Kiali API to generate graph data
+in the form of consumable JSON.
+
+The [Bookinfo](/docs/examples/bookinfo/) sample application is used as
+the example application throughout this task.
+
+## Before you begin
+
+* [Install Istio](/docs/setup/) in your cluster and deploy an
```
sdake: The docs/setup don't show how to enable Kiali and Kiali atm is disabled by default.  Can you change the docs or the state.  IIRC from the last TOC meeting was to keep kiali disabled until there was some basic smoke testing in the gate to make sure it comes up properly.

jmazzitelli: Yes, good point. It will be disabled by default (except for the demo yamls, which I believe we agreed to have everything plus the kitchen sink enabled in those :))

I will add some blurb here explaining how to at least get it to be installed via the helm chart (involves creating a secret and passing `--set kiali.enabled=true` to helm as described in https://github.com/morvencao/istio/blob/master/install/kubernetes/helm/istio/README.md#installing-the-chart)

content/docs/tasks/telemetry/servicegraph/index.md
```diff
@@ -1,12 +1,11 @@
...
-title: Generating a Service Graph*
+title: Generating a Service Graph
 description: This task shows you how to generate a graph of services within an Istio mesh.
 weight: 50
 keywords: [telemetry,visualization]
 ---
 
-> \* Note: The Service Graph has been deprecated.
-> It has been replaced with [Kiali](../kiali).
+> {{< warning_icon >}} The Service Graph has been deprecated and replaced with [Kiali](../kiali).
```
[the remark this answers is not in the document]
jmazzitelli: How is this (from commit https://github.com/istio/istio.io/pull/2715/commits/bdac17c2fea695d0649f2df7e58c7e93ff510102):

`The Service Graph is being deprecated in Istio 1.1 in favor of Kiali. It will be removed in the 1.2 timeframe.`

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: I tried addressing some of this in the commit: https://github.com/istio/istio.io/pull/2715/commits/271fa9980e2e679563157d538565656549f4c7e9

jmazzitelli:
> Instead of `$GATEWAY_URL` and `port-forward` should we be targeting the kiali vs/gateway that was recently added? I feel like that is a more natural way of accessing the content. Thoughts?

That would be better; I just haven't tested that vs/gw that was just added so... if I don't see it work, it doesn't exist :) I'll test it and update this if it works.

jmazzitelli:
> Let me know if you want me to perform a detailed review.

@rcaballeromx Yes, I would like a review - but I'll finish the content first, and when I'm ready for a detailed review, and I will ping you back. Thanks.

sdake:
@jmazzitelli using gateways/vs would be cool although the rest of the documentation would be inconsistent.  `kubectl port-forward` is used all over the place for telemetry at present.  We need to fix this for 1.1 across the documentation as operators are often confused about how to obtain an IP address for the addons.  `kubectl port-forward` is sadly limited to binding to `127.0.0.1` and hence is typically unusable in an eval environment (which is what we are after with kiali as well as telemetry).

Cheers
-steve

content/about/notes/1.1/index.md
```diff
@@ -2,9 +2,12 @@
 title: Istio 1.1
 weight: 90
```
sdake: is this the correct weight?  I believe you want 89.

jmazzitelli:
@sdake so should I leave this PR as-is wrt using port-forward  (i.e. do NOT mention the vs/gw)?

If so, then I think the content of this PR is done. We'd need folks to just start reviewing formatting/syntax/grammar - but as for the actual content, I think this is done barring objections from others.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,345 @@
+---
+title: Visualizing Your Mesh
+description: This task shows you how to visualize your services within an Istio mesh.
+weight: 50
```
sdake: I am not sure how the website caching responds to using the same weight for a previously weighted item.  @geeknoid is using the same weight as Service Graph (which is deprecated) ok here?

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,345 @@
...
+keywords: [telemetry,visualization]
+---
+
+This task shows you how to visualize different aspects of your Istio mesh.
+
+As part of this task, you will install the [Kiali](https://www.kiali.io) add-on and use
+the web-based graphical user interface for viewing service graphs of the mesh
+as well as your Istio configuration objects such as virtual services
```
sdake: i think you can remove such as ... until period.  Seems extraneous and makes the sentence longer and harder to read.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,345 @@
...
+and destination rules. You will also use the Kiali API to generate graph data
+in the form of consumable JSON.
+
+The [Bookinfo](/docs/examples/bookinfo/) sample application is used as
+the example application throughout this task.
+
+## Before you begin
+
+{{< idea_icon >}} If you already have Istio installed and need Kiali installed, you can follow the [Kiali install instructions here](https://www.kiali.io/gettingstarted/).
```
sdake: this looks ok, but the recommended way to achieve what you want is to use `helm template --set kiali.enabled --set (other kiali flags needed)` and kubectl apply it.  There is a similar workflow for helm upgrade which does the same thing.  What is proposed here would result in an inconsistent deployment. I'd recommend keeping the link around, but simply as a reference, not as the cononical way to install Kiali on Istio.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,345 @@
...
+
+Before you install [Istio](/docs/setup/) in your cluster, you first need to create a secret in your Istio namespace with credentials that users will use to authenticate to Kiali. See the [Helm README](https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/README.md#installing-the-chart) for details, but here is a command you can run to create a secret:
+
+{{< text ascii >}}
```
sdake: `{{< text yaml >}}` or `{{<text bash>}}` should render better.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,345 @@
...
+USERNAME=$(echo -n 'admin' | base64) ; \
+  PASSPHRASE=$(echo -n 'mysecret' | base64) ; \
+  NAMESPACE=istio-system ; \
+  kubectl create namespace $NAMESPACE ; \
+  cat <<EOF | kubectl apply -f -
```
sdake: 26-30 start the line with $ and remove the trailing `; \`

IOW make them separate commands.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,345 @@
...
+  labels:
+    app: kiali
+type: Opaque
+data:
+  username: $USERNAME
+  passphrase: $PASSPHRASE
+EOF
+{{</ text >}}
+
+Once a Kiali secret is created, you can now install Istio and Kiali via Helm by following [these instructions](/docs/setup/kubernetes/helm-install/). Note that you will want to use the `--set kiali.enabled=true` option when running the `helm` command. For example:
```
sdake: nice improvement over the last revs 👍

jmazzitelli:
For the record, I just tested this and it works for me - if we want to, this is how we can tell people to access Kiali via the gateway (as opposed to port-forward):

---

3. Open the Kiali UI.

In Kubernetes environments, execute the following command to determine the Kiali URL:

{{< text bash >}}
$ echo "$(kubectl -n istio-system get svc istio-ingressgateway -o jsonpath='{.spec.clusterIP}'):15029"
172.30.214.70:15029
{{< /text >}}

Point your web browser to that IP:port to access Kiali (e.g. `http://172.30.214.70:15029`).

sdake:
> For the record, I just tested this and it works for me - if we want to, this is how we can tell people to access Kiali via the gateway (as opposed to port-forward):
> 
> 1. Open the Kiali UI.
> 
> In Kubernetes environments, execute the following command to determine the Kiali URL:
> 
> {{< text bash >}}
> $ echo "$(kubectl -n istio-system get svc istio-ingressgateway -o jsonpath='{.spec.clusterIP}'):15029"
> 172.30.214.70:15029
> {{< /text >}}
> 
> Point your web browser to that IP:port to access Kiali (e.g. `http://172.30.214.70:15029`).

You are welcome 👍.  Such a change LGTM.

Cheers
-steve

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: @sdake I did use text bash before, and I can switch back to using it again, but it doesn't render as nice as you'd think. It thinks the "here document" is part of the output:

image

So I can't click the "copy" button and have the full set of commands copied to the clipboard nicely. The way it is now (with text ascii), I can click the "copy" button, go over to my console window, "paste" and the command executes and everything is created on one shot.

Does anyone know how to get this markdown "test bash" to understand the "here document" ?

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: This is done. I also changed it - no longer uses text bash (since it wasn't rendering the "here document" properly), I changed it to use "```bash" and that looks better and the reader can still click the "copy" button and get all the commands on the clipboard so it can be pasted to a console.

image

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: FYI: I switched it to use "```bash" and it renders it better while retaining the ability copy-n-paste into a console window.

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: @sdake OK, I moved that down to the bottom of the "before you begin" section just as a reference in case someone wants an alternative to helm.

I wasn't sure if we should assume the reader had Istio already installed or not - I tried not to assume that, but rather try to keep it relevant whether they have Istio installed or not (i.e. the instructions have them run helm template and then kubectl apply - which, as you say, can install Istio, too, or just update it with a kiali install).

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: Perhaps I should make this 49 instead - just one less of the deprecated ServiceGraph, since its meant to take its place later. This would still leave the forties (40-48) available. We can bump this back up to 50 when ServiceGraph Task is removed in the ~1.2 timeframe.

content/about/notes/1.1/index.md
jmazzitelli: this is done

content/docs/tasks/telemetry/servicegraph/index.md
```diff
@@ -5,6 +5,8 @@ weight: 50
 keywords: [telemetry,visualization]
 ---
 
+> {{< warning_icon >}} The Service Graph is being deprecated in Istio 1.1 in favor of [Kiali](../kiali). It will be removed in the 1.2 timeframe.
```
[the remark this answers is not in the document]
jmazzitelli: This is done.

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: This is done.

jmazzitelli:
> Such a change LGTM.

:-) OK, the latest change has this - no longer using kubectl port-forward but instead just gets the ingress gateway URL.

jmazzitelli:
> Instead of `$GATEWAY_URL` and `port-forward` should we be targeting the kiali vs/gateway that was recently added? I feel like that is a more natural way of accessing the content. Thoughts?

FYI: this is done.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,363 @@
...
+description: This task shows you how to visualize your services within an Istio mesh.
+weight: 49
+keywords: [telemetry,visualization]
+---
+
+This task shows you how to visualize different aspects of your Istio mesh.
+
+As part of this task, you will install the [Kiali](https://www.kiali.io) add-on
+and use the web-based graphical user interface for viewing service graphs of
+the mesh as well as your Istio configuration objects. You will also use the
```
JStickler: for viewing service graphs of the mesh and your Istio configuration objects.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,363 @@
...
+
+The [Bookinfo](/docs/examples/bookinfo/) sample application is used as the
+example application throughout this task.
+
+## Before you begin
+
+You first need to create a secret in your Istio namespace with credentials that
+users will use to authenticate to Kiali. See the
+[Helm README](https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/README.md#installing-the-chart)
+for details, but here is a command you can run to create a secret:
```
JStickler: ...here is an example command you can modify and run....   (Just so people don't blindly copy and paste without adding their own configuration data)

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,363 @@
...
+  labels:
+    app: kiali
+type: Opaque
+data:
+  username: $USERNAME
+  passphrase: $PASSPHRASE
+EOF
+```
+
+Once a Kiali secret is created, you can now install Kiali via Helm by following
```
JStickler: Once you have created the Kiali secret, you can install Kiali via Helm...

(Passive voice)

jmazzitelli:
> Let me know if you want me to perform a detailed review.

OK @rcaballeromx   - let me have it :)

The content should be complete. I don't anticipate adding anything new. So feel free to give a detailed review and suggest/change anything that doesn't match proper syntax, grammar, or style.

Thanks.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,363 @@
...
+to use the `--set kiali.enabled=true` option when running the `helm` command.
+For example:
+
+{{< text bash >}}
+$ helm template --set kiali.enabled=true install/kubernetes/helm/istio --name istio --namespace istio-system > $HOME/istio.yaml
+$ kubectl apply -f $HOME/istio.yaml
+{{</ text >}}
+
+Once Istio and Kiali are installed, you should then deploy an application such
+as the [Bookinfo](/docs/examples/bookinfo/) sample application.
```
JStickler: Could we add a link to instructions for installing bookinfo?

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,363 @@
...
+
+{{< idea_icon >}} The above installation instructions assume you have `helm`
```
JStickler: Move this note to the beginning of the installation instructions, so that people who don't have Helm know that there's an alternative path.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,363 @@
...
+    {{< /text >}}
+
+    Refresh the page a few times, or send the command a few times, to generate a small amount of traffic.
+    To continually send requests, you can use the `watch` command if available on your system:
+
+    {{< text bash >}}
+    $ watch -n 1 curl -o /dev/null -s -w %{http_code} $GATEWAY_URL/productpage
+    {{< /text >}}
+
+1.  Login to the Kiali UI.
```
JStickler: Break this out into two separate steps:
Determine the Kiali URL
Log into the console.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,363 @@
...
+    This shows a node for each workload in your service mesh. This graph type
+    does not require you to use the "app" and "version" labels so if you opt to
+    not use those labels on your components, this is the graph type you will
+    use.
+    {{< image width="70%" ratio="76%"
+    link="./kiali-workload.png"
+    caption="Example Workload Graph"
+    >}}
+
+1. Examine Details About The Istio Configuration.
```
JStickler: This step could probably be broken into several separate steps:
View Applications
View Application details
View Workloads
View Workload details
View Services
View Service Details

jmazzitelli: I was in a quandary - not sure how much detail to get into. This istio.io Task is really just supposed to show the user "this is how you can visualize your mesh" but not "these are all the things Kiali can do." So I wasn't sure how much detail I should get into and how many of the different screens I was going to introduce or tell the user to go to. I think going to all of those in separate steps would be overwhelming to a user who just wants an introduction to Kiali and just wants to know "what does my mesh look like?"

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: @sdake what do you think of this suggestion? I did have this at the top before - but I moved it to the bottom as a reaction to one of your earlier comments. I can go either way. :)

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: That /docs/examples/bookinfo does result in a link to the install instructions - it ends up giving the reader a link to this: https://istio.io/docs/examples/bookinfo/

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: this is done.

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: this is done.

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: This is done.

content/docs/tasks/telemetry/kiali/index.md
jmazzitelli: This is done

content/docs/tasks/telemetry/kiali/index.md
JStickler: Fair enough.  I wasn't sure how much detail you wanted to go into either.

jmazzitelli:
> 17 spelling errors found in 253 files
To learn how to address spelling errors, please see https://github.com/istio/istio.io#linting
content/docs/tasks/telemetry/kiali/index.md:60: MD031 Fenced code blocks should be surrounded by blank lines

The problem with this failure is if I add a blank line to satisfy the lint check, the text is not rendering in a way I want. I want it to look like this:

image

But to pass the lint check, I have to put a blank line above it which renders it like this (notice the code is now outside the grey "notice" box):

image

But, to get the lint job to pass, I'll add the blank line.

jmazzitelli:
OK, well, I don't get the lint thing - its telling me about spelling errors in code blocks and json that it should not be doing. I'll assume we can ignore those errors and won't block us from merging.

jmazzitelli:
- I updated the content to better conform to the [Istio style guide](https://istio.io/about/contribute/style-guide/).
- I fixed all lint errors - this should pass the lint check now.
- Finally, I squashed all commits.

Hopefully, this is in good shape now.

vnugent:
When trying to install Kiali on a minikube cluster using helm I used the following to determine the URL for Kiali console:

```
HOST_IP=$(minikube ip)
KIALI_PORT= $(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="http-kiali")].nodePort})
```

jmazzitelli:
> When trying to install Kiali on a minikube cluster using helm I used the following to determine the URL for Kiali console:

@sdake @rcaballeromx @douglas-reid What's the rule of thumb for how we should document something like this? In the Task doc, we are telling the user to go through the ingressgateway to access Kiali, however, depending on their environment, it appears we need to tell them different things to get the gateway IP and port.

frankbu:
I'm not sure, what's the best way to instruct people on finding the gateway URL in general, but in this case it's just a different port on the same host as the Bookinfo GATEWAY_URL, which the user has already set. So you could just do something like this:
```
KIALI_URL="$(echo $GATEWAY_URL | sed -e s/:.*/:15029/)"
```

frankbu:
Actually my previous comment isn't quite right, since you still need different instructions for the port as well, since it may or may not be a mapped nodeport.

Maybe the best thing to do something like I suggested to get the host and then say that if are running in an environment that doesn't have an external loadbalancer (e.g., minikube) then you should use the kiali nodeport otherwise the port

jmazzitelli:
@frankbu @rcaballeromx @vnugent I added some more details in the step "Determine the Kiali URL".

frankbu:
@jmazzitelli I wish it could be simpler yet, but here's my suggestion:

```
1.  Determine the Kiali URL.

Kiali is exposed through the same Istio gateway as the Bookinfo application,
only on a different port. Therefore you can use the `GATEWAY_URL` that you set
in the previous step to determine the Kiali URL.

If you are running in an environment that has external loadbalancers:

    {{< text bash >}}
    $ KIALI_URL="http://$(echo $GATEWAY_URL | sed -e s/:.*//):$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="http-kiali")].port}')"
    $ echo $KIALI_URL
    http://172.30.141.9:15029
    {{< /text >}}

If you are running in an environemnt that does not support external loadbalancers (e.g., minikube):

    {{< text bash >}}
    $ KIALI_URL="http://$(echo $GATEWAY_URL | sed -e s/:.*//):$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="http-kiali")].nodePort}')"
    $ echo $KIALI_URL
    http://192.168.99.100:31758
    {{< /text >}}
```

jmazzitelli:
thanks, @frankbu - I made that change.

I see need @rcaballeromx to review and remove his "request change" block on this PR before we can approve/merge.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,403 @@
...
+weight: 49
+keywords: [telemetry,visualization]
+---
+
+This task shows you how to visualize different aspects of your Istio mesh.
+
+As part of this task, you will install the [Kiali](https://www.kiali.io) add-on
+and use the web-based graphical user interface for viewing service graphs of
+the mesh and your Istio configuration objects. You will also use the Kiali
+Public API to generate graph data in the form of consumable JSON.
```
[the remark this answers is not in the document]
jmazzitelli: I will replace "graph" with "diagram" - however, keep in mind, the Kiali UI uses the term "graph" in places such as "Graph Type" and the "Graph View". This is in relation to the topology graph pages of the GUI. Where the word "graph" relates directly to a component in the Kiali UI, I will leave the word "graph" in place to avoid confusion (e.g. there is a "Graph Type" drop down menu item in Kiali UI - I should not refer to this as "Diagram Type" since it would not match what the user sees in the UI).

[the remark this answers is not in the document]
jmazzitelli: Also, keep in mind that these diagrams are really, technically, called "graphs" in that we use different graph models/algorithms implemented by Cytoscape (a third party library which also calls these things "graphs" - http://js.cytoscape.org/#notation/graph-model )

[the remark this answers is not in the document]
jmazzitelli: Actually, now that I'm looking more closely at the terminology, I believe these should be called "graphs" - these have nodes/edges, are directed, and cyclic or acyclic depending on connections between nodes. These are all features of graphs (nomenclature of the technology in question). In other words, do not think of the word "graph" here as in a "line graph" or a "bar graph" where you have mathematical data represented visually - these are "graphs" in the sense of "directed acyclic graphs" or the like, which is what Kiali is dealing with.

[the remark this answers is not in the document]
rcaballeromx: I understand if you consider, and based on the UI examples you mention, that the term graph is better suited, even if it is being used incorrectly. If you agree, my advice is to leave the term alone for now.

rcaballeromx:
@jmazzitelli I notice you have made some changes to the verbs. I'll perform a detailed review based on your latest commits today.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+  }
+}
+{{< /text >}}
+
+The Kiali Public API is built on top of Prometheus queries and depends on the
+standard Istio metric configuration.  It also makes Kubernetes API calls to
+obtain additional details about your services.  For the best experience using
+Kiali, you should use the metadata labels `app` and `version` on your
+application components (the Bookinfo sample application follows this
+convention).
```
rcaballeromx: Grammar, punctuation, and word choice:
For the best experience using Kiali, use the metadata labels `app` and `version` on your
application components. As a template, the Bookinfo sample application follows this convention.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+## Cleanup
+
+If you are not planning to explore any follow-on tasks, remove the
+Bookinfo sample application and Kiali.
```
rcaballeromx: Idiom:
If you are not planning any follow-up tasks, remove the Bookinfo sample application and Kiali from your cluster.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+To remove the Bookinfo application, refer to the [Bookinfo cleanup](/docs/examples/bookinfo/#cleanup) instructions.
```
rcaballeromx: This should be a numbered step.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+To remove Kiali from a Kubernetes environment, remove all components with the `app=kiali` label:
```
rcaballeromx: Ditto

content/docs/tasks/telemetry/servicegraph/index.md
```diff
@@ -5,6 +5,8 @@ weight: 50
 keywords: [telemetry,visualization]
 ---
 
+> {{< warning_icon >}} The Service Graph is being deprecated in Istio 1.1 in favor of [Kiali](../kiali).
```
rcaballeromx: The note, IMO is not enough. If the Istio service graphs are to be deprecated, the content should be archived and unpublished. @sdake @geeknoid Is there a content deprecation process?

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+example application throughout this task.
+
+## Before you begin
+
+Create a secret in your Istio namespace with credentials that you use to
+authenticate to Kiali. See the
+[Helm README](https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/README.md#installing-the-chart)
+for details, but here is an example command you can modify and run to create a secret:
+
+```bash
```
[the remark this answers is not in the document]
jmazzitelli: I originally did have it this way, however, there is a bug in the text_bash markup implementation that caused this to not render properly. I mentioned this in this comment: https://github.com/istio/istio.io/pull/2715#discussion_r222068150

The problem is text bash markup does not render a "here document" properly. Using "``bash" does render it properly; doing it this way allows the reader to still easily click the "copy" button and get all the commands on the clipboard so it can be pasted to a console with a simple "paste" operation as opposed to copy-n-pasting each line individually (while not selecting the $ at the same time) - that would be what the reader would need to do if I put "$" prompt in front of the individual commands inside "```bash".

For comparison, here's the bug - notice when using text_bash, the "here document" is rendered as if it is the command output, so it looks like it is output and not part of the command and the "copy" button doesn't copy  the "here document" into the clipboard:

image


 Thus, when using "here documents", it looks nicer with "```bash":

image

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+$ helm template --set kiali.enabled=true install/kubernetes/helm/istio --name istio --namespace istio-system > $HOME/istio.yaml
+$ kubectl apply -f $HOME/istio.yaml
+{{< /text >}}
+
+> {{< idea_icon >}} This Task does not discuss Jaeger and Grafana; however, if
+you have them already installed in your cluster and you want to see how Kiali
+integrates with both of them, you need to pass additional arguments to the
+`helm` command. For example:
+
+{{< text bash >}}
```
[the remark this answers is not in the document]
jmazzitelli: I tried that, it doesn't render any differently. I did get the code block to become part of the note, but only by deleting the blank line between the note text and the text bash - however, when I do that, the linter gives me an error message:

```
content/docs/tasks/telemetry/kiali/index.md:57: MD031 Fenced code blocks should be surrounded by blank lines
```

So to avoid having the "make lint" job block this PR I put the blank line back.

Indenting the code block (I tried indenting the text bash markers as well as not indenting the text bash markers) does not help. It doesn't render any differently. However, when I commit my next round of changes, I will indent the code block with four spaces.

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+    NAME           CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
+    kiali          10.59.253.165   <none>        20001/TCP  30s
+    {{< /text >}}
+
+1.  Determine the Bookinfo URL.
+
+    Follow these [instructions](/docs/examples/bookinfo/#determining-the-ingress-ip-and-port)
+    to determine the Bookinfo `GATEWAY_URL`.
+
+1.  Send traffic to the mesh.
```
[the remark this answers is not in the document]
jmazzitelli: That's part of the instructions the previous step told the reader to execute. See http://localhost:1313/docs/examples/bookinfo/#determining-the-ingress-ip-and-port where it tells the user in step 4:

```
Set GATEWAY_URL:

$ export GATEWAY_URL=$INGRESS_HOST:$INGRESS_PORT
```

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+---
+
+This task shows you how to visualize different aspects of your Istio mesh.
+
+As part of this task, you install the [Kiali](https://www.kiali.io) add-on
+and use the web-based graphical user interface to view service graphs of
+the mesh and your Istio configuration objects. Lastly, you use the Kiali
+Public API to generate graph data in the form of consumable JSON.
+
+The [Bookinfo](/docs/examples/bookinfo/) sample application is used as the
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+example application throughout this task.
+
+## Before you begin
+
+Create a secret in your Istio namespace with credentials that you use to
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+authenticate to Kiali. See the
+[Helm README](https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/README.md#installing-the-chart)
+for details, but here is an example command you can modify and run to create a secret:
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+type: Opaque
+data:
+  username: $USERNAME
+  passphrase: $PASSPHRASE
+EOF
+```
+
+Once the Kiali secret is created, install Kiali via Helm by following
+[these instructions](/docs/setup/kubernetes/helm-install/). Note that you need
+to use the `--set kiali.enabled=true` option when running the `helm` command.
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+For example:
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+{{< text bash >}}
+$ helm template --set kiali.enabled=true install/kubernetes/helm/istio --name istio --namespace istio-system > $HOME/istio.yaml
+$ kubectl apply -f $HOME/istio.yaml
+{{< /text >}}
+
+> {{< idea_icon >}} This Task does not discuss Jaeger and Grafana; however, if
+you have them already installed in your cluster and you want to see how Kiali
+integrates with both of them, you need to pass additional arguments to the
+`helm` command. For example:
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+$ helm template \
+    --set kiali.enabled=true \
+    --set "kiali.dashboard.jaegerURL=http://$(kubectl get svc tracing -o jsonpath='{.spec.clusterIP}'):80" \
+    --set "kiali.dashboard.grafanaURL=http://$(kubectl get svc grafana -o jsonpath='{.spec.clusterIP}'):3000" \
+    install/kubernetes/helm/istio \
+    --name istio --namespace istio-system > $HOME/istio.yaml
+$ kubectl apply -f $HOME/istio.yaml
+{{< /text >}}
+
+Once Istio and Kiali are installed, deploy the [Bookinfo](/docs/examples/bookinfo/) sample application.
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+> {{< idea_icon >}} The above installation instructions assume you have `helm`
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+and want to use it to install Kiali. Alternatively, you can install Kiali by
+following the [Kiali install instructions](https://www.kiali.io/gettingstarted/).
+
+## Generating a service graph
+
+1.  Verify that the service is running in your cluster.
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+    In Kubernetes environments, execute the following command:
+
+    {{< text bash >}}
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+    $ kubectl -n istio-system get svc kiali
+    NAME           CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
+    kiali          10.59.253.165   <none>        20001/TCP  30s
+    {{< /text >}}
+
+1.  Determine the Bookinfo URL.
```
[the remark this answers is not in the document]
jmazzitelli: done - however, I used the term "GATEWAY_URL" since that is the term/variable name throughout. To avoid confusing the reader, and to be clear we are talking about the env var that needs to be set, I felt it best to use GATEWAY_URL always (rather then here say "gateway URL").

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+    Follow these [instructions](/docs/examples/bookinfo/#determining-the-ingress-ip-and-port)
+    to determine the Bookinfo `GATEWAY_URL`.
+
+1.  Send traffic to the mesh.
+
+    Generate a small amount of traffic by visiting `http://$GATEWAY_URL/productpage`
+    in your web browser or by issuing the following command multiple times:
```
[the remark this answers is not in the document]
jmazzitelli: done

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+          "id": "06d772ed417cd829b4b6f95a209ed1b5",
+          "source": "c219903556c3afdb05eda7e610aba628",
+          "target": "e96058d658030da2171c18d09086ee77",
+          "tcpSentRate": "94.452"
+        }
+      }
+    ]
+  }
+}
+{{< /text >}}
```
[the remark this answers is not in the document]
jmazzitelli: I prefer to leave this because it provides the reader with a sample of what the graph JSON actually looks like. Without it, they really don't have any idea of the API's usefulness because they won't know what pieces of information this API provides. It helps illustrate an answer to the question, "What data does this API give me?"

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+    {{< text bash >}}
+    $ KIALI_URL="http://$(echo $GATEWAY_URL | sed -e s/:.*//):$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="http-kiali")].nodePort}')"
+    $ echo $KIALI_URL
+    http://192.168.99.100:31758
+    {{< /text >}}
+
+1.  Log into the Kiali UI.
+
+    Visit `$KIALI_URL` in your web browser, and in the login screen enter the
```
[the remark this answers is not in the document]
jmazzitelli: done. split into two steps - see fef2e729d3d405a94dfb30229de88cb282edfc83

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+    Click on a namespace to see a graph of all services in that namespace.  For
+    example, if you have the Bookinfo sample application installed, click on
+    `bookinfo` in the Bookinfo namespace card.  The page looks similar to:
+
+    {{< image width="75%" ratio="89%"
+    link="./kiali-graph.png"
+    caption="Example Graph"
+    >}}
+
+1.  View summary metrics.
```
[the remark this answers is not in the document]
jmazzitelli: will reword: "To view a summary of metrics, select any node or edge in the graph to display its metric details in the summary details panel on the right." 51cd491e1c1766b63911b6fd7e13f2f022d9afb4

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+    {{< image width="75%" ratio="58%"
+    link="./kiali-overview.png"
+    caption="Example Overview"
+    >}}
+
+1.  View a namespace graph.
+
+    Click on a namespace to see a graph of all services in that namespace.  For
+    example, if you have the Bookinfo sample application installed, click on
+    `bookinfo` in the Bookinfo namespace card.  The page looks similar to:
```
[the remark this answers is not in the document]
jmazzitelli: done 51cd491e1c1766b63911b6fd7e13f2f022d9afb4

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+    {{< /text >}}
+
+1.  Log into the Kiali UI.
+
+    Visit `$KIALI_URL` in your web browser, and in the login screen enter the
+    credentials that you stored in the Kiali secret you created above. If you
+    used the example secret above, enter a username of `admin` with a password
+    of `mysecret`.
+
+1.  Get an overview of your mesh.
```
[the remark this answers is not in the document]
jmazzitelli: I just reworded it a little bit. This page shows up by default immediately after you log in - there is nothing for the reader to do other than what was done in the previous step (the login step) - I'm just bringing the reader's attention to the overview page and what it shows. 51cd491e1c1766b63911b6fd7e13f2f022d9afb4

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+    link="./kiali-graph.png"
+    caption="Example Graph"
+    >}}
+
+1.  View summary metrics.
+
+    Select any node or edge in the graph and notice the summary details panel
+    on the right displays additional data about that selected node or edge.
+
+1.  Experiment with different graph types.
```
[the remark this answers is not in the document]
jmazzitelli: restructured - 09d2a0a916d1620b09b67cc1ae90741e69248d42

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+    Visualize the graph using several different types of graphs such as
+    the **App**, **Versioned App**, and **Workload** graph types.
+
+    1. Select the **App** graph type from the **Graph Type** drop down menu.
```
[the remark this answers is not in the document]
jmazzitelli: used normal list structure, not numbered list. also added the new service graph type. 09d2a0a916d1620b09b67cc1ae90741e69248d42

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+    menu.  This shows a node for each workload in your service mesh. This graph
+    type does not require you to use the `app` and `version` labels so if
+    you opt to not use those labels on your components, this is the graph type
+    you will use.
+    {{< image width="70%" ratio="76%"
+    link="./kiali-workload.png"
+    caption="Example Workload Graph"
+    >}}
+
+1. Examine details about the Istio configuration.
```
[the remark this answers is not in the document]
jmazzitelli: done - 748b951e002e655c83833f0b0a8e90cba5cfcfec

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+   {{< image width="80%" ratio="56%"
+   link="./kiali-services.png"
+   caption="Example Details"
+   >}}
+
+## About the Kiali Public API
+
+Generate JSON representing the graphs, as well as other metric, health,
+and configuration information, by accessing the
+[Kiali Public API](https://www.kiali.io/api/paths).
```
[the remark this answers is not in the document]
jmazzitelli: done 3ca0cca79c9bf56cd1846d6aa4022662684d75d0

content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -0,0 +1,400 @@
...
+
+For example, make a request to
```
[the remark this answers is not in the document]
jmazzitelli: i removed the example output, I made it more clear the request can be made from the browser to the given example URL. 3ca0cca79c9bf56cd1846d6aa4022662684d75d0

content/docs/tasks/telemetry/kiali/index.md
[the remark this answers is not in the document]
jmazzitelli: done - example output removed 3ca0cca79c9bf56cd1846d6aa4022662684d75d0

content/docs/tasks/telemetry/servicegraph/index.md
jmazzitelli: @jeffmendoza @sdake @douglas-reid  - correct me if I'm wrong in what I say below. In short, Service Graph add-on and its docs are not to be removed yet (only marked as deprecated).

As I understand it, the Service Graph add-on is not being removed - just marked as deprecated for removal in a later Istio version. So the Service Graph will still exist and ship with Istio 1.1. Thus, it was my understanding the Service Graph docs (such as the Task and its README) will remain but will have a deprecation notice. Those docs will then be removed when a future Istio version (1.2 probably) is released.

In fact, see this [Istio PR #9094](https://github.com/istio/istio/pull/9094) for the deprecation notice added to the README (which was to be done in tandem with this change in the Service Graph Task that you see here).

content/docs/tasks/telemetry/kiali/index.md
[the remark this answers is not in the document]
jmazzitelli: I just tried adding a "$" to the beginning of the commands in this "```bash" block, but the copy-n-paste doesn't work. If a reader clicks the "copy" button and tries to paste it into a console to create the secret, a bunch of errors will result:

```
$ $ USERNAME=$(echo -n 'admin' | base64)
bash: $: command not found...
$ $ PASSPHRASE=$(echo -n 'mysecret' | base64)
bash: $: command not found...
$ $ NAMESPACE=istio-system
bash: $: command not found...
$ $ kubectl create namespace $NAMESPACE
bash: $: command not found...
$ $ cat <<EOF | kubectl apply -f -
...
> EOF
bash: $: command not found...
error: no objects passed to apply
```

So there are problems all around - can't use text_bash because of a bug in rendering "here documents", and can't put "$" in the "```bash" section without making the copy-paste feature fail for the user. To make it easier for the reader to run this task, the way I have it now seems the best. The reader need only press the "copy" button and "paste" the bash content into their bash command window and the secret will be created.

jmazzitelli:
@rcaballeromx This is ready for the next round of reviews. I finished all my edits based on previous comments/suggestions. Thanks!

content/docs/tasks/telemetry/servicegraph/index.md
jeffmendoza: Not sure about docs. Service Graph code certainly needs to stay in istio/istio, and installable via helm option, not removed or archived.

I can see the case for removing docs from the site. We don't want new users to discover and start using it. I'll defer to @geeknoid for a decision.

jmazzitelli:
This is the final PR to be merged in order to consider https://github.com/istio/istio/issues/9066 closed.

frankbu:
Let's merge this now. If @rcaballeromx has more suggestions, we can address in a followup PR.

rcaballeromx:
@jmazzitelli The changes look great. I'm making a note of the bug you described and will try to find a solution for it.

content/docs/tasks/telemetry/kiali/index.md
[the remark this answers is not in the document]
rcaballeromx: The copy button functionality is the only thing I would considered bugged. Having the configuration file as part of the command is considered a walk-around ideally you would have the command be something like:
```bash
kubectl apply -f this-conf.yaml
```
instead.

PR #2737, opened 2018-10-04
broke up long command line

navinger:
At the suggestion of mjog@google.com, I broke the long helm command into two separate commands. I also changed "<the flags you used to install Istio>" to "<the flags you used to install Helm>" because I didn't use Helm to install Istio, and I found that confusing.

Signed-off-by: Nancy Avinger <navinger@google.com>


navinger:
Okay @frankbu. I'll take your suggestion and close this PR and open an issue to clarify this section.

PR #2761, opened 2018-10-11
rbac: deprecate RbacConfig with ClusterRbacConfig

yangminzhu:
istio/istio#8825

The script is being reviewed in https://github.com/istio/istio/pull/9175, we can review the content of this PR first.

/cc @diemtvu

### DIFF
--- a/content/docs/setup/kubernetes/upgrading-istio/index.md
+++ b/content/docs/setup/kubernetes/upgrading-istio/index.md
@@ -260,10 +260,10 @@ spec:
 
 If you use [Istio authorization](/docs/concepts/security/#authorization) with custom resource of kind `RbacConfig`,
 you need to replace the custom resource of kind `RbacConfig` with kind `ClusterRbacConfig`.
-The rbac config has a [bug](https://github.com/istio/istio/issues/8825) that causes it to be namespace scoped in some cases. The cluster rbac config
-has exactly the same specification as the rbac config with correct cluster scope implementation.
+The `RbacConfig` has a [bug](https://github.com/istio/istio/issues/8825) that causes it to be namespace scoped in some cases. The `ClusterRbacConfig`
+has exactly the same specification as the `RbacConfig` with correct cluster scope implementation.
 
-For example, if you have a rbac config with the following config:
+For example, if you have a `RbacConfig` like the following one:
 
 {{< text yaml >}}
 apiVersion: "rbac.istio.io/v1alpha1"
@@ -277,8 +277,8 @@ spec:
 {{< /text >}}
 
 Kubernetes doesn't allow to change the kind of a custom resource once it's created, so you need to
-create a new cluster rbac config with the same specification as the existing rbac config. For the above
-rbac config, you will need to apply the following config:
+create a new `ClusterRbacConfig` with the same specification as the existing `RbacConfig`. For the above
+`RbacConfig`, you will need to apply the following `ClusterRbacConfig`:
 
 {{< text yaml >}}
 apiVersion: "rbac.istio.io/v1alpha1"
@@ -291,12 +291,12 @@ spec:
     namespaces: ["default"]
 {{< /text >}}
 
-Wait for a few seconds until the cluster rbac config take effect and then you can safely delete
-the rbac config.
+Wait for a few seconds until the `ClusterRbacConfig` take effect and then you can safely delete
+the `RbacConfig`.
 
-A script is provided to automate the migration for you. The script will get existing rbac config and
-apply a cluster rbac config with the same specification. It will then wait a few seconds and delete
-the rbac config if the new cluster rbac config is applied successfully.
+A script is provided to automate the migration for you. The script will get existing `RbacConfig` and
+apply a `ClusterRbacConfig` with the same specification. It will then wait a few seconds and delete
+the `RbacConfig` if the new `ClusterRbacConfig` is applied successfully.
 
 Depending on if you have the [Istio installation package](/docs/setup/kubernetes/download-release):
 
--- a/content/docs/setup/kubernetes/upgrading-istio/index.md
+++ b/content/docs/setup/kubernetes/upgrading-istio/index.md
@@ -261,7 +261,7 @@ spec:
 The `RbacConfig` is deprecated due to a [bug](https://github.com/istio/istio/issues/8825). You must
 migrate to `ClusterRbacConfig` if you are currently using `RbacConfig`. The bug reduces the scope of
 the object to be namespace-scoped in some cases. The `ClusterRbacConfig` follows the exact same
-specification as the `RbacConfig` but with correct cluster scope implementation.
+specification as the `RbacConfig` but with the correct cluster scope implementation.
 
 To automate the migration, we developed the `convert_RbacConfig_to_ClusterRbacConfig.sh` script.
 The script is included in the [Istio installation package](/docs/setup/kubernetes/download-release).
@@ -303,7 +303,7 @@ The script automates the following operations:
         namespaces: ["default"]
     {{< /text >}}
 
-1. The script applies the configuration and wait for a few seconds to let the configuration to take effect.
+1. The script applies the configuration and waits for a few seconds to let the configuration to take effect.
 
-1. The script deletes the previous RBAC configuration file after applying the cluster RBAC configuration
-   successfully.
+1. The script deletes the previous RBAC configuration custom resource after applying the cluster RBAC
+   configuration successfully.
--- a/content/docs/setup/kubernetes/upgrading-istio/index.md
+++ b/content/docs/setup/kubernetes/upgrading-istio/index.md
@@ -258,10 +258,33 @@ spec:
 
 ## Migrating the `RbacConfig` to `ClusterRbacConfig`
 
-If you use [Istio authorization](/docs/concepts/security/#authorization) with custom resource of kind `RbacConfig`,
-you need to replace the custom resource of kind `RbacConfig` with kind `ClusterRbacConfig`.
-The `RbacConfig` has a [bug](https://github.com/istio/istio/issues/8825) that causes it to be namespace scoped in some cases. The `ClusterRbacConfig`
-has exactly the same specification as the `RbacConfig` with correct cluster scope implementation.
+The `RbacConfig` has a [bug](https://github.com/istio/istio/issues/8825). The bug reduces the scope of
+the object to be namespace-scoped in some cases. The `ClusterRbacConfig` follows the exact same
+specification as the `RbacConfig` but with correct cluster scope implementation.
+
+1. Replace the custom resource of kind `RbacConfig` with kind `ClusterRbacConfig` if you use
+   [Istio authorization](/docs/concepts/security/#authorization) with custom resource of kind `RbacConfig`:
+
+    A script is provided to automate the migration for you. Depending on if you have the [Istio installation package](/docs/setup/kubernetes/download-release):
+
+    - If you have the Istio installation package
+
+        Move to the Istio package directory and run the script with the following command:
+
+        {{< text bash >}}
+        $ ./tools/convert_RbacConfig_to_ClusterRbacConfig.sh
+        {{< /text >}}
+
+    - If you don't have the Istio installation package
+
+        Download and run the script directly with the following command:
+
+        {{< text bash >}}
+        $ curl -L https://raw.githubusercontent.com/istio/istio/master/tools/convert_RbacConfig_to_ClusterRbacConfig.sh | sh -
+        {{< /text >}}
+
+The script will get existing `RbacConfig` and apply a `ClusterRbacConfig` with same specification as
+Kubernetes doesn't allow to change the kind of a custom resource once it's created.
 
 For example, if you have a `RbacConfig` like the following one:
 
@@ -276,9 +299,7 @@ spec:
     namespaces: ["default"]
 {{< /text >}}
 
-Kubernetes doesn't allow to change the kind of a custom resource once it's created, so you need to
-create a new `ClusterRbacConfig` with the same specification as the existing `RbacConfig`. For the above
-`RbacConfig`, you will need to apply the following `ClusterRbacConfig`:
+The script will create the following `ClusterRbacConfig`:
 
 {{< text yaml >}}
 apiVersion: "rbac.istio.io/v1alpha1"
@@ -291,27 +312,4 @@ spec:
     namespaces: ["default"]
 {{< /text >}}
 
-Wait for a few seconds until the `ClusterRbacConfig` take effect and then you can safely delete
-the `RbacConfig`.
-
-A script is provided to automate the migration for you. The script will get existing `RbacConfig` and
-apply a `ClusterRbacConfig` with the same specification. It will then wait a few seconds and delete
-the `RbacConfig` if the new `ClusterRbacConfig` is applied successfully.
-
-Depending on if you have the [Istio installation package](/docs/setup/kubernetes/download-release):
-
-- If you have the Istio installation package
-
-    Move to the Istio package directory and run the script with the following command:
-
-    {{< text bash >}}
-    $ ./tools/convert_RbacConfig_to_ClusterRbacConfig.sh
-    {{< /text >}}
-
-- If you don't have the Istio installation package
-
-    Download and run the script directly with the following command:
-
-    {{< text bash >}}
-    $ curl -L https://raw.githubusercontent.com/istio/istio/master/tools/convert_RbacConfig_to_ClusterRbacConfig.sh | sh -
-    {{< /text >}}
+The script will wait for a few seconds to let the `ClusterRbacConfig` to take effect and then delete the `RbacConfig`.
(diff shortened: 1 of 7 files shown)


content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,63 @@ spec:
     tls:
       mode: DISABLE
 {{< /text >}}
+
+## Migrating the `RbacConfig` to `ClusterRbacConfig`
+
+If you use [Istio authorization](/docs/concepts/security/#authorization) with custom resource of kind `RbacConfig`,
+you need to replace the custom resource of kind `RbacConfig` with kind `ClusterRbacConfig`.
+The rbac config has an implementation bug that causes it to be namespace scoped in some cases. The cluster rbac config
```
ayj: Is it worth describing the bug in more detail or at least providing a reference to one?

yangminzhu: I added a link to the bug.

yangminzhu:
@geeknoid Can we merge this if no changes needed? Thank you.

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,63 @@ spec:
     tls:
       mode: DISABLE
 {{< /text >}}
+
+## Migrating the `RbacConfig` to `ClusterRbacConfig`
```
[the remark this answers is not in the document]
yangminzhu: There is only 1 step to do the migration that is described in the end of this section. Other text are just explanation and examples.

yangminzhu:
@rcaballeromx Comments addressed, PTAL (The CircleCI fail is not related), Thanks!

content/docs/setup/kubernetes/upgrading-istio/index.md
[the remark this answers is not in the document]
rcaballeromx: Rewritten the first paragraph looks like:
1. Replace the custom resource of kind `RbacConfig` with kind `ClusterRbacConfig` if you use [Istio authorization](/docs/concepts/security/#authorization) with a custom resource of kind `RbacConfig`.

> The `RbacConfig` object has a [bug](https://github.com/istio/istio/issues/8825). The bug reduces the scope of the object to be namespace-scoped in some cases. The `ClusterRbacConfig` follows the exact same specification as the `RbacConfig` but with correct cluster scope implementation.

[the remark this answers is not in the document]
rcaballeromx: Consider that a reader of this page wants to complete the task at hand. Explanations and examples are all well and good but in this case they are hiding the actual actions the reader needs to take. My advice is for you to rewrite this section putting the actual steps of downloading the script, running it, and verifying it completed the migration successfully fist. Keep the explanations and examples but at the end of the section.

yangminzhu:
@rcaballeromx Reorganized and moved the task before the example/explanations, also deleted some redundant. PTAL.

rcaballeromx:
@yangminzhu It looks better than before but the text remains convoluted. I'm leaving comments pointing to the issues causing it.

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,61 @@ spec:
...
+
+The `RbacConfig` has a [bug](https://github.com/istio/istio/issues/8825). The bug reduces the scope of
+the object to be namespace-scoped in some cases. The `ClusterRbacConfig` follows the exact same
+specification as the `RbacConfig` but with correct cluster scope implementation.
```
rcaballeromx: This text explains the bug and the fact that `ClusterRbacConfig` does not have the buggy behavior. That implies why the migration is needed but fails to be explicit or to indicate: When should readers migrate? Under what conditions does the bug impact them?

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,61 @@ spec:
...
+
+1. Replace the custom resource of kind `RbacConfig` with kind `ClusterRbacConfig` if you use
+   [Istio authorization](/docs/concepts/security/#authorization) with custom resource of kind `RbacConfig`:
```
rcaballeromx: Everyone using RbacConfig must migrate? If that is the case, why not deprecate the current buggy version of RbacConfig in favor of the bug-free ClusterRbacConfig?

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,61 @@ spec:
...
+
+    A script is provided to automate the migration for you. Depending on if you have the [Istio installation package](/docs/setup/kubernetes/download-release):
```
rcaballeromx: Aside from the passive voice issue: "To automate the migration, we provide a script in the [Istio installation package] ..."
Please replace the non-specific "a script" with the specific filename: "the `convert_RbacConfig_to_ClusterRbacConfig.sh` script. Do not structure the content as an not-ordered list.
The text in lines 268-288 should be a series of steps.
Since you do not know whether the reader has downloaded the installation package, the flow should be:
1. Download the installation package or Download the script 
   <command to download the package or the script>
2. Run the script
   <command to run the script>
3. Verify the migration is complete
   <command to verify the migration was completed.

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,61 @@ spec:
...
+
+    - If you have the Istio installation package
+
+        Move to the Istio package directory and run the script with the following command:
```
rcaballeromx: The command does not move the reader to the Istio package directory. See my comment on line 268. There is no need for two options for users to get to the script.

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,61 @@ spec:
...
+
+    - If you don't have the Istio installation package
+
+        Download and run the script directly with the following command:
+
+        {{< text bash >}}
+        $ curl -L https://raw.githubusercontent.com/istio/istio/master/tools/convert_RbacConfig_to_ClusterRbacConfig.sh | sh -
+        {{< /text >}}
+
+The script will get existing `RbacConfig` and apply a `ClusterRbacConfig` with same specification as
```
rcaballeromx: Verb congruence: Do not use future tense in documentation.

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,61 @@ spec:
...
+Kubernetes doesn't allow to change the kind of a custom resource once it's created.
+
+For example, if you have a `RbacConfig` like the following one:
```
rcaballeromx: The verification content starting here until the end must be rewritten as part of the procedure:
1. To verify the migration was completed successfully, take a RBAC configuration file before migration:

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,61 @@ spec:
...
+kind: RbacConfig
+metadata:
+  name: default
+spec:
+  mode: 'ON_WITH_INCLUSION'
+  inclusion:
+    namespaces: ["default"]
+{{< /text >}}
+
+The script will create the following `ClusterRbacConfig`:
```
rcaballeromx: 1. Compare the previous file with the cluster RBAC configuration file the script created:

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,61 @@ spec:
...
+kind: ClusterRbacConfig
+metadata:
+  name: default
+spec:
+  mode: 'ON_WITH_INCLUSION'
+  inclusion:
+    namespaces: ["default"]
+{{< /text >}}
+
+The script will wait for a few seconds to let the `ClusterRbacConfig` to take effect and then delete the `RbacConfig`.
```
rcaballeromx: This line contains two steps:
1. Wait for a few seconds to let the configuration take effect.
1. Delete the previous configuration file:
   <The command to remove the file with the exact path and filename of the file that needs to be removed.

yangminzhu: The script will do all these steps (migrate, wait and delete), for user there is only one step that is to run the script.

content/docs/setup/kubernetes/upgrading-istio/index.md
yangminzhu: The RbacConfig is deprecated and everyone should migrate

content/docs/setup/kubernetes/upgrading-istio/index.md
yangminzhu: everyone should migrate if they are currently using RbacConfig

content/docs/setup/kubernetes/upgrading-istio/index.md
yangminzhu: What do you mean for verification content? This is not for verification, it just explains what would the script do.

content/docs/setup/kubernetes/upgrading-istio/index.md
yangminzhu: This makes it strange, here I want to point out what the script do, not to let the user to compare it manually.

rcaballeromx: What step or steps can the reader do to verify the script worked as intended?

content/docs/setup/kubernetes/upgrading-istio/index.md
rcaballeromx: I applaud the fact the script does all those things! A slight reword might be all you need @yangminzhu:
The script automates the following operations: 
1. Create the cluster RBAC configuration file.
2. Apply the configuration and wait for a few seconds to let the configuration take effect.
3. Delete the previous RBAC configuration file.

content/docs/setup/kubernetes/upgrading-istio/index.md
rcaballeromx: Readers expect to be able to verify what the script does after running it. It can be as simple as opening a file and verify the value was changed but there must be a verification step.

content/docs/setup/kubernetes/upgrading-istio/index.md
rcaballeromx: I do not see that statement anywhere in the doc. Please add a clear statement at the beginning of the section.

content/docs/setup/kubernetes/upgrading-istio/index.md
yangminzhu: The script will check if the new configuration is applied successfully and prompt the user accordingly.

yangminzhu:
@rcaballeromx I rephrased a lot based on your comments, PTAL, thanks

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,66 @@ spec:
     tls:
       mode: DISABLE
 {{< /text >}}
+
+## Migrating the `RbacConfig` to `ClusterRbacConfig`
+
+The `RbacConfig` is deprecated due to a [bug](https://github.com/istio/istio/issues/8825). You should
```
rcaballeromx: Wrong verb: You must

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,66 @@ spec:
...
+migrate to `ClusterRbacConfig` if you are currently using `RbacConfig`. The bug reduces the scope of
+the object to be namespace-scoped in some cases. The `ClusterRbacConfig` follows the exact same
+specification as the `RbacConfig` but with correct cluster scope implementation.
+
+To migrate the `RbacConfig` to `ClusterRbacConfig`:
+
+1. We provide the `convert_RbacConfig_to_ClusterRbacConfig.sh` script to automate the migration,
+   depending on if you have the [Istio installation package](/docs/setup/kubernetes/download-release):
```
rcaballeromx: I don't see how this dependence is important. IMO, mentioning that it is included in the installation package is enough to cover that situation. I would get rid of lines 271-279 and just provide the users with the command for direct download and execution. If you follow said advice, the text needs to change to:
To automate the migration, we developed the `convert_RbacConfig_to_ClusterRbacConfig.sh` script. The script is included in the [Istio installation package](/docs/setup/kubernetes/download-release). 
Download and run the script with the following command:
<command on lines 283 -285>

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,66 @@ spec:
...
+    kind: ClusterRbacConfig
+    metadata:
+      name: default
+    spec:
+      mode: 'ON_WITH_INCLUSION'
+      inclusion:
+        namespaces: ["default"]
+    {{< /text >}}
+
+1. Apply the configuration and wait for a few seconds to let the configuration to take effect.
```
rcaballeromx: Do not use the imperative here. This process is performed by the script, and thus the ordered list is descriptive rather than prescriptive. A simple present sentence does the trick, just use "The script" as the subject. For example:
1. The script applies the configuration and waits for a few seconds to let the configuration to take effect.

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,66 @@ spec:
...
+
+        Download and run the script directly with the following command:
+
+        {{< text bash >}}
+        $ curl -L https://raw.githubusercontent.com/istio/istio/master/tools/convert_RbacConfig_to_ClusterRbacConfig.sh | sh -
+        {{< /text >}}
+
+The script automates the following operations:
+
+1. Create a `ClusterRbacConfig` with same specification as the existing `RbacConfig`. This
```
rcaballeromx: See comment on line 318 for the reason behind this edit suggestion. Do not use the value in `kind` to refer to configuration files unless you are specifically referencing the code:
1. The script creates a cluster RBAC configuration file with same specification as the existing RBAC configuration file because

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,66 @@ spec:
...
+   is because Kubernetes doesn't allow to change the kind of a custom resource after it's created.
```
rcaballeromx: Avoid using "this" without a noun immediately after:
... existing RBAC configuration file because Kubernetes doesn't allow the value of `kind:` in a custom resource to change after it's created.

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,66 @@ spec:
...
+      name: default
+    spec:
+      mode: 'ON_WITH_INCLUSION'
+      inclusion:
+        namespaces: ["default"]
+    {{< /text >}}
+
+1. Apply the configuration and wait for a few seconds to let the configuration to take effect.
+
+1. Delete the previous `RbacConfig` after the `ClusterRbacConfig` is applied successfully.
```
rcaballeromx: Always leave and empty line at the end of MD files, it helps avoid some parsing issues. Also see comment on line 318 for the justification of the following suggestion:
1. The script deletes the previous RBAC configuration file after applying the new cluster RBAC configuration successfully.

content/docs/setup/kubernetes/upgrading-istio/index.md
yangminzhu: The Istio installation package is actually optional, user could install Istio without using the Istio installation package. That's why I provide two options here: if the user installed Istio via the installation package, they can find the script already there.
If the user installs it in some custom way, they can only download the script (not the Istio installation package) nad use the script directly .

rcaballeromx: I understand, but my point is: We don't need to tell the readers how to get the installation package here. That bit of information is given when they install. It is enough to mention they can find the script there if they already have the package, which my suggestions do.

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,55 @@ spec:
...
+    kind: ClusterRbacConfig
+    metadata:
+      name: default
+    spec:
+      mode: 'ON_WITH_INCLUSION'
+      inclusion:
+        namespaces: ["default"]
+    {{< /text >}}
+
+1. The script applies the configuration and wait for a few seconds to let the configuration to take effect.
```
rcaballeromx:
```suggestion
1. The script applies the configuration and waits for a few seconds to let the configuration to take effect.
```

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -255,3 +255,55 @@ spec:
     tls:
       mode: DISABLE
 {{< /text >}}
+
+## Migrating the `RbacConfig` to `ClusterRbacConfig`
+
+The `RbacConfig` is deprecated due to a [bug](https://github.com/istio/istio/issues/8825). You must
+migrate to `ClusterRbacConfig` if you are currently using `RbacConfig`. The bug reduces the scope of
+the object to be namespace-scoped in some cases. The `ClusterRbacConfig` follows the exact same
+specification as the `RbacConfig` but with correct cluster scope implementation.
```
rcaballeromx: Grammar:
```suggestion
specification as the `RbacConfig` but with the correct cluster scope implementation.
```

yangminzhu:
@rcaballeromx Also replaced `file` with `custom resource` in last sentence as the script only deletes it in k8s and won't be able to delete the actual config file.

yangminzhu:
Rigs, Thanks for helping making this better.
@geeknoid Could you take a look, Thanks!

PR #2775, opened 2018-10-16
Update Task/Enabling rate limits

jsenon:
Update Task/Enabling rate limits, remove validDuration in dimension that is not in redisquota, Move redisserverurl and connectionpoolsize


jsenon:
Does somebody can launch retest?

geeknoid:
Could you rebase and push again? That should fix the CircleCI failures.

Thanks.

PR #2779, opened 2018-10-17
Fix typo in gateway label example

abonas:


frankbu:
Thanks for the PR but this is generated documentation and has already been fixed at the source: https://github.com/istio/api/blob/0b70b6093baee68d6e1f4e30ecf28bb33f89710a/networking/v1alpha3/gateway.proto#L41

It will be fixed next time the reference docs are refreshed.

PR #2782, opened 2018-10-18
Add Network configuration objects content to the Traffic management concept

rcaballeromx:
The Network configuration objects concept provides a high-level overview of
the available network configuration capabilities of Istio. Cross-references,
diagrams, and examples provide a clear introduction to the objects with a clear
didactic approach and simple content structure. Links to and from existing
content help readers connect the different pieces of information into single
cognitive units. To aid that goal, I performed some consistency edits of
existing content.


content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+  hosts:
+    - ext-svc
+  gateways:
+    - ext-host-gwy
+{{< /text >}}
+
+The following diagram shows the configured routes for the external HTTPS
+traffic coming from the external host on port 443 to the `virtual-svc` virtual
+service through the `ext-host-gwy` gateway to reach the `my-svc` service
+within the mesh:
```
oaktowner: I think that this following diagram shows what the previous 2 YAMLs describe. Again: probably better to start with the prose description of what we want (e.g., "traffic from outside my mesh addressed to https:my-svc should get routed to the service (workload?) called my-svc").

And it would be much better to have something like "shopping-cart" than "my-svc"

rshriram:
This is actually pretty cool, especially given that you are not the subject matter expert in the area. I like the intent and the layout, love the diagrams that give a quick visual overview of the networking config. Glad you could glean so much info from the networking API docs. Thank you for doing this!

All said, there are a lot of inaccuracies here, which I can understand. I strongly suggest you let @frankbu incorporate the layout and the content into the existing traffic management doc. Its important to convey the concepts to users before the mechanisms. The traffic management doc does just that (describing our design philosophies for canary rollouts, etc.) and then takes the user through the networking config.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+Use ingress gateways to configure ingress traffic routes into the service mesh.
+You can add as many gateways as you need. For example, you can have one gateway
+for private traffic and another for public traffic. Thus, you balance the
+different loads independently from each other.
+
+### Egress gateways
+
+Use egress gateways to configure an exit route for traffic leaving the mesh.
+You can configure each egress gateway to use its own policies and telemetry.
+You can use egress gateways if only particular application services can or
```
oaktowner: Isn't there a 'must' here? That is, you _must_ use egress gateways if you want to permit traffic off the mesh?

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+## Virtual Services
+
+You can use virtual services for fine-grained traffic control. Istio can use
+virtual services to configure  ingress, canary, traffic rewrite, or traffic
+policing. A virtual service is not required for simple connectivity. A service
+can map one or more user-addressable destinations to the actual destination
+workloads inside the mesh without a virtual service. The Istio virtual services
+can configure traffic routes to:
+
+-  Other virtual services in the mesh.
```
frankbu: No they can't

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+-  Specific services or services' subsets in the mesh.
+-  Other network configuration objects in the mesh.
+
+{{< image width="80%"
+    link="./net-config-2.svg"
+    caption="Possible configurations between virtual services and workloads"
+    >}}
+
+As you can see, virtual services are very flexible and can accommodate a wide
+variety of topologies. With destination rules, virtual services configure the
```
frankbu: Destination rules are not required

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+services through a single Istio virtual service. To eliminate redundant rules,
+you can add multiple match conditions to a virtual service configuration. If
+your mesh uses Kubernetes, for example, you can configure a virtual service to
+handle all services in a specific namespace. With the Istio virtual services,
+you can configure routes for the canary and A/B testing for your application's
+services. You can configure each application service version as a virtual
+service subset. You can also configure the corresponding Istio destination rule
+to determine the set of pods or VMs belonging to these subsets.
+
+You can use the Istio virtual services in a variety of ways: to route traffic
```
frankbu: What do you mean by a variety of ways, given you only follow that up with two examples?

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: my-vtl-svc
+spec:
+  hosts:
+    - *.my-co.org
+      route:
+      - destination:
+        host: service
```
frankbu: This doesn't seem like a very realistic ... You're delegating everything all my-co.org subdomains to a single monolith service

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+implementations can co-exist within your mesh. Using a gateway, you can
+configure workload labels to reuse your existing network appliances such as:
+firewall functions, caching, authentication, network address translation, and
+IP address management.
+
+By separating the layer 4 and 6 specifications from the layer 7, the Istio
+gateways overcome the shortcomings of the layer 4 and 6 specifications. Istio
+only configures the functions that layer 7 load balancing proxies implement
+uniformly. To use standard Istio rules to control HTTP requests as well as TCP
+traffic entering a gateway, you must bind a virtual service to the gateway. The
```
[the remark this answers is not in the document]
frankbu: yes, all traffic entering via an ingress gateway needs to be routed somewhere using a virtual service.

content/docs/concepts/network/index.md
frankbu: no, sending egress traffic via a gateway is optional.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+  - ext-resource.com
+  ports:
+  - number: 443
+    name: https
+    protocol: HTTPS
+  location: MESH_EXTERNAL
+{{< /text >}}
+
+Because a service entry configuration simply adds the external resource to the
+internal service registry, you must use either a virtual service or destination
```
frankbu: Depending on the situation, you don't need anything more that a service entry to call an external service

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+external dependency through the `svc-entry` service entry on port 80 and using
+mutual TLS. The following diagram shows the configured traffic routes:
+
+{{< image width="50%"
+    link="./net-config-8.svg"
+    caption="Configurable traffic routes using service entries and destination rules"
+    >}}
+
+Service entries do not require virtual services to address services in the
+mesh. Use destination rules to configure routes to address services directly.
```
frankbu: Not sure what this is trying to say, but I think it's wrong.

rcaballeromx:
@rshriram This content is intended to provide conceptual information about the network objects. The Traffic routing doc focuses on What you can do with traffic and how to accomplish specific routing task, which is all well and good. This doc focuses on How the network configuration objects work and why they are needed. I did link the two documents to one another but merging them into one would only make the content too long and confusing.

rcaballeromx:
@frankbu I fully agree that the term "gateway" is overloaded. This document introduces consistency by avoiding referring to the configuration objects using the value in the `kind` key. This document is not talking about a specific configuration file with a `kind: Gateway` value, but rather describing the conceptual behavior of the gateway configuration object. Reading your comment does leave me thinking that additional disambiguation might be needed. I did not consider that the actual workload is also called "gateway". I am trying to establish a consistent pattern for these terms in this document. Using `VirtualService` instead of virtual service is very confusing because it, incorrectly, communicates that what the text refers to only applies to the example file provided.

rcaballeromx:
@oaktowner @frankbu @rshriram Thank you all for your comments and review! I have replied to the general comments and will be addressing the in-line comments soon.

rcaballeromx:
@oaktowner Regarding your raised issues about the diagrams:

1. I depicted the Envoy proxies because none of the configuration happens on the actual workloads. The configuration is always sent to the proxies. I could remove the proxies from the diagram but that could create a lot of confusion with folks thinking the configuration affects services.

1. They are neither E:R nor network diagrams. Lines 20-24 explain clearly what the diagrams are depicting.

rcaballeromx:
@frankbu I did follow an Istio-wide policy: https://istio.io/about/contribute/style-guide/
The case around virtual service vs `VirtualService` is the similar to Mixer vs mixer or "service mesh" vs "Service Mesh". 
The markup and spelling of things like `VirtualService` imply a connection to the actual value in the configuration file. If I was writing content that needed that direct connection, I would use it but the content here is not referring to that specific value.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+services together with destination rules allow you to configure the behavior of
+the traffic to fulfill the needs of your mesh. Together with a service entry,
+you can use virtual services to configure traffic routes to external
+dependencies.
+
+Visit our [virtual services reference documentation](/docs/reference/config/istio.networking.v1alpha3/#VirtualService) to review all the enabled keys and values.
+
+## Gateways
+
+The Istio gateways are layer 7 load balancers. With an Istio gateway, you can
```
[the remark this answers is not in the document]
rcaballeromx: Is it correct to say that every configuration object with `kind: Gateway` requires 1 or more workloads to work? How can I identify the workload in the configuration object? You mention that they only configure layer 4-6 but I read [here](https://istio.io/blog/2018/v1alpha3-routing/#gateway) that they implement L7. Can you help me by clarifying further?

content/docs/concepts/network/index.md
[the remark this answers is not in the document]
oaktowner: Thanks, @frankbu .  It would be _great_ to capture that.

content/docs/concepts/network/index.md
oaktowner: ack.

content/docs/concepts/network/index.md
oaktowner: Great to know. @rcaballeromx can you capture that?

content/docs/concepts/network/index.md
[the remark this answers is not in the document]
frankbu: Right, so this is the problem. A "gateway" is a workload, an L7 router implementation (typically an Envoy running in a pod). A `Gateway`, which is what you are talking about here, is the kind of the configuration resource that configures the L4-6 properties of the gateway that it is bound to. It's bound to a particular gateway (workload) with the `selector` field in the `Gateway`. A `VirtualService` is then used to configure the remaining L7 properties of the gateway. This is why I think its best to explicitly refer to the configuration resources by the value of there kind field. That seems pretty clear to me.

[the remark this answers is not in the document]
frankbu: Btw, the style guide doesn't really address the issue of the configuration resource kinds although it does have examples of referring to them as I suggest: https://istio.io/about/contribute/style-guide/#use-code-style-for-inline-code-and-commands

Notice the word `Deployment`.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
+---
+title: The Istio network configuration objects and your application
```
geeknoid: This is a bit long for our normal document titles, can this be made shorter?

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+    caption="Istio network configuration objects overview"
+    >}}
+
+This document describes the key configurations each configuration object
+enables. Additionally, this document provides high-level examples of how the
+Istio network API can help you handle internal, incoming, and outgoing traffic.
+Lastly, the document discusses the benefits of using the Istio network API for
+your cloud applications.
+
+# The Istio network configuration objects
```
geeknoid: For our docs, the highest-level header is H2. So this should be ##

content/docs/concepts/network/index.md
rcaballeromx: done

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+description: Describes Istio's network configuration objects and their functionality.
+weight: 30
+keywords: [network, virtual service, gateway, ingress, egress, service entry]
+aliases:
+---
+
+Cloud applications, unlike monolithic applications, are comprised of a
+collection of connected services. Throughout this document a service should be
+understood as an addressable destination workload inside the mesh. These
+services can connect to other services or to external resources. The Istio
```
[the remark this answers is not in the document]
rcaballeromx: Shifted focus to the readers application and rewrote the two sentences to provide a better intro to the document.

content/docs/concepts/network/index.md
rcaballeromx: Done

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+enables. Additionally, this document provides high-level examples of how the
+Istio network API can help you handle internal, incoming, and outgoing traffic.
+Lastly, the document discusses the benefits of using the Istio network API for
+your cloud applications.
+
+# The Istio network configuration objects
+
+The Istio network configuration objects provide you with four objects to
+configure the traffic routes between your services and your external resources:
+virtual services, destination rules, service entries, and gateways. You can add
```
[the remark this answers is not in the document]
rcaballeromx: Done

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+and configure these objects in a variety of ways to suit your needs. Each
+configuration object serves a specific purpose and provides specific
+configurations for your Istio service mesh. To create a solution for your
+specific mesh, you can combine the configuration objects in a way that best
+fulfills your needs.
+
+You can define the Istio network configuration objects as Kubernetes Custom
```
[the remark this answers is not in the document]
rcaballeromx: I used can because, theoretically, readers could be using other constructs if they do not use Kubernetes. That being said and since the entire doc focuses heavily on CRDs, I agree with you.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+Resource Definition [(CRD)](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
+objects and store their configuration in YAML files. The following sections
+describe the purposes and specific configurations of each configuration object
+and provide examples.
+
+## Virtual Services
```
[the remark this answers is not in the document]
rcaballeromx: I followed a third approach, a didactic one. I began with the simplest and most familiar, virtual service, and went to the more complex and unfamiliar, service entries. There is a dependency between gateways and virtual services though. Because of it, I went with the virtual services first.

content/docs/concepts/network/index.md
rcaballeromx: Understood.

content/docs/concepts/network/index.md
rcaballeromx: Understood. What functionality would you say a destination rule can add to a virtual service? What relationships are possible between the two?

content/docs/concepts/network/index.md
rcaballeromx: I'm trying to leave it open ended to communicate that even if what the reader is doing with the configuration objects does not match 1:1 what is described here, that does not mean that it should not be possible. I added a conditional prepositional phrase, "for example" to make clear that the two examples are just that.

content/docs/concepts/network/index.md
rcaballeromx: Realism was not my intent. This doc is about adding clarity to how the configuration objects behave. Starting with a simple example where everything goes to a single service is meant as an introduction. Further down the doc, more complex configurations are discussed.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+spec:
+  hosts:
+    - *.my-co.org
+      route:
+      - destination:
+        host: service
+        subset: v1
+{{< /text >}}
+
+Virtual services can specify the destination's DNS host name using the `hosts:`
```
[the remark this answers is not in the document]
rcaballeromx: I took another stab based on your suggestions. With the exception of your final suggestion. Overwhelming readers by mentioning all the other places where "service" is defined would defeat the main purpose of the doc, which is to teach how the configuration objects work. The configuration files in this doc are not meant to be tried out nor is anywhere implied they would "work" if the reader just copy pasted them somewhere. That type of content and examples is meant for procedures, not concepts.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+field under `spec:`. As you can see in the `my-vtl-svc` example, you can use a
+wildcard for the virtual service configurations to apply to multiple hosts. In
+the `destination:` field of each route, you specify the service and subset that
+the configuration targets. The example configures a route for traffic to reach
+the `v1` subset of the `my-svc` service from the `my-vtl-svc` virtual service.
```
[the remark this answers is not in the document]
rcaballeromx: Moving it up.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+    - uri:
+        prefix: /svc-2
+    route:
+    - destination:
+        host: svc-2
+{{< /text >}}
+
+The following diagram shows the configured traffic routes implementing a
+virtual service to configure traffic for services in a namespace for two
+distinct application services based on the URI prefixes:
```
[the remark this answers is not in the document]
rcaballeromx: Yes.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+
+{{< image width="50%"
+    link="./net-config-5.svg"
+    caption="Configurable traffic rules implementing a namespace for two  application services"
+    >}}
+
+Once you have identified and configured the virtual services suitable for your
+mesh, you can add gateways to route traffic in or out of your mesh. The virtual
+services together with destination rules allow you to configure the behavior of
+the traffic to fulfill the needs of your mesh. Together with a service entry,
```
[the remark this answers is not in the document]
rcaballeromx: No, they are introduced here as the piece a virtual service needs to "configure traffic routes to external dependencies". It's an ancient didactic technique. I'm using something the reader now knows, virtual services, to begin introducing the unknown, gateways, destination rules, and service entries.
I'm adding links to the the different sections on this doc covering the mentioned objects. Hopefully it does not end up feeling too link-cluttered.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+`my-namespace.svc.cluster.local` Kubernetes namespace:
+
+{{< text yaml >}}
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: my-namespace
+spec:
+  hosts:
+    - *.my-namespace.svc.cluster.local
```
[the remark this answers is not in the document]
rcaballeromx: @frankbu Please correct me if I'm wrong but, the way this is currently configured, means:
1. Only requests intended for the `my-namespace.svc.cluster.local` namespace will follow the route in this virtual service regardless of whether they are intended for svc1 or svc2.
2. Only requests intended for the `svc1.my-namespace.svc.cluster.local` follow the route to `host: svc-1`
3. Only requests intended for the `svc2.my-namespace.svc.cluster.local` follow the route to `host: svc-2`
Just to reiterate, this doc is trying to explain things, I'm aware that out in the wild developers probably create a virtual service for each pod if they need granular traffic control to them instead of doing it on a namespace level.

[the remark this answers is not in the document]
rcaballeromx: @oaktowner You can see the flow and how the hosts are intended to be resolved. IMU (In My Understanding), he wildcard in `hosts:` allow configured virtual service to apply to any pods (svc-X) in the `my-namespace.svc.cluster.local` namespace. Then the specific routes are configured for individual pods. The prefix is used because what distinguishes one pod's name from the others is what comes before.

content/docs/concepts/network/index.md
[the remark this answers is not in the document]
rcaballeromx: @frankbu Thank you for the clarification. I've modified the text using your outstanding explanation. To avoid the terminology issue, I'm using the term "gateway configuration object" to distinguish the actual "gateway workload" 
IMO using the value of `kind:` doesn't really solve the problem of us referring to two very different things with the same term, it just transforms it into a markup use consistency issue. 
The best practice in technical writing is clear here: https://docs.microsoft.com/en-us/style-guide/word-choice/use-technical-terms-carefully
Having some specific research on this might be helpful but until then, I'll use the workaround described above.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+the traffic to fulfill the needs of your mesh. Together with a service entry,
+you can use virtual services to configure traffic routes to external
+dependencies.
+
+Visit our [virtual services reference documentation](/docs/reference/config/istio.networking.v1alpha3/#VirtualService) to review all the enabled keys and values.
+
+## Gateways
+
+The Istio gateways are layer 7 load balancers. With an Istio gateway, you can
+configure routes for HTTP and TCP traffic regardless of where the gateway runs.
```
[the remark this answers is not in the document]
rcaballeromx: Your confusion is warranted, As @frankbu explained, gateway here is referring to the actual workload, which can run in a number of places. In a k8s mesh it would be in a pod. I'm changing the term "gateway" to "gateway workload" to avoid this confusion.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+Your mesh can have any number of gateways and multiple different gateway
+implementations can co-exist within your mesh. Using a gateway, you can
```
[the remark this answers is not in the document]
rcaballeromx: Done

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,437 @@
...
+configure workload labels to reuse your existing network appliances such as:
+firewall functions, caching, authentication, network address translation, and
+IP address management.
+
+By separating the layer 4 and 6 specifications from the layer 7, the Istio
+gateways overcome the shortcomings of the layer 4 and 6 specifications. Istio
+only configures the functions that layer 7 load balancing proxies implement
+uniformly. To use standard Istio rules to control HTTP requests as well as TCP
```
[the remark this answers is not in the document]
rcaballeromx: Done.

content/docs/concepts/network/index.md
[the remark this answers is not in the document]
rcaballeromx: Done

content/docs/concepts/network/index.md
rcaballeromx: That is exactly what it shows, I assume the text doesn't make that clear?
The diagrams are an effort to "show" rather than "tell" using prose. I've added a brief sentence though. 
The service can be called whatever but using "my-svc" and similar names makes it clear that these are placeholder names.

content/docs/concepts/network/index.md
rcaballeromx: Absolutely, done!

content/docs/concepts/network/index.md
rcaballeromx: It was trying to say what you stated in the previous comment only awkwardly. I've removed it.

rcaballeromx:
I've addressed all comments on the text. @frankbu @oaktowner Please let me know if any of the changes don't address your concerns with the text.
@rshriram The key difference between this content and the "Traffic management" concept is: 
Traffic management is about how to manage traffic for specific purposes.
Network configuration objects is about the configurable components available and how they work with one another.
They are certainly related and I've added cross-references between them to cement that relationship.
However, the documents address fundamentally different questions.

rcaballeromx:
@rshriram @frankbu @oaktowner Please review the text while I try to improve the diagrams based on @oaktowner 's comments.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,445 @@
+---
+title: Network configuration objects
```
geeknoid: We capitalize our titles, so "Network Configuration Objects"

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,445 @@
...
+description: Describes Istio's network configuration objects and their functionality.
+weight: 30
+keywords: [network, virtual service, gateway, ingress, egress, service entry]
+aliases:
+---
+
+## The Istio network configuration objects and your application
```
geeknoid: Please don't start with a title. The page as a whole already has a title rendered by the site infra.

content/docs/concepts/network/index.md
```diff
@@ -0,0 +1,445 @@
...
+    caption="Istio network configuration objects overview"
+    >}}
+
+This document describes the key configurations each configuration object
+enables. Additionally, this document provides high-level examples of how the
+Istio network API can help you handle internal, incoming, and outgoing traffic.
+Lastly, the document discusses the benefits of using the Istio network API for
+your cloud applications.
+
+## The Istio network configuration objects
```
geeknoid: Somewhat redundant, given the title of the page.

geeknoid:
I'm definitely concerned about this doc, relative to the Traffic Management doc. The TM doc already discusses the same kinds of objects.

Right now, in our Concepts, we have one page per major functional area in the product. The idea is to give our customers a single unambiguous place to go to find the overall description of the area. This new doc now introduces two conceptual overview of the networking functionality, which I think is confusing and somewhat redundant.

When I initiatlly read drafts of this doc, I was expecting it to be a full replacement for the existing TM concept doc, and I still think that's the right path forward. I propose to merge this doc's new content into the TM doc, thus maintaining the idea of having a single conceptual overview of each functional area.

Thanks.

rcaballeromx:
@frankbu @rshriram @geeknoid I have completed the integration of the new content with our existing Traffic management content. During that process, I had to restructure the entire section. You can see exactly how on the [commit message](https://github.com/istio/istio.io/pull/2782/commits/27c44739b5305566a3b71364a3a17b13b6e67bdc)
The new structure is clearer, easier to consume, and does not assume the content is read from beginning to end - all good things. 
PTAL

content/docs/concepts/traffic-management/routing-rules/index.md
```diff
@@ -0,0 +1,271 @@
...
+work as expected. **Pilot** would evaluate the weight-based rule first in that
+case and route all traffic to the instances in the `v1` subset, even requests
+including the matching `Foo` header.
+
+Once **Pilot** finds a rule that applies to the incoming request, the rule is
+executed and the rule-evaluation process terminates.
+
+{{< warning_icon >}} Consider the priorities of each rule in a configuration file when there is more than one.
+
+## Multiple match conditions {#multi-condition}
```
frankbu: Multiple match conditions are "conditional rules", so this section belongs after the following one (possibly even as a subsection of it).

rcaballeromx: Agreed, I'll move the content then.

content/docs/concepts/traffic-management/routing-rules/index.md
```diff
@@ -0,0 +1,271 @@
...
+
+- [Canary rollouts](../traffic-routing/#canary)
+
+- Staged rollouts with percentage-based [traffic splitting](../traffic-routing/#splitting)
+
+The behaviors described in the following sections apply to all routing rules
+regardless of the configuration object where you configure them. If you use the
+examples provided, you must adapt them to your application and apply the
+configuration in the appropriate configuration object. For simplicity, all
+examples use [virtual services](../virtual-services).
```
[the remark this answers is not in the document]
rcaballeromx: Thank you for the clarification. To enhance clarity and keep the content modular, I'll keep this content as a separate file but will emphasize what you just clarified. I will link both the virtual services section to this one and vice-versa. One question remains: Are routing rules only found on virtual services? It is my understanding that destination rules can also contain them, is that correct?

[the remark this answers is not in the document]
frankbu: No. Destination rules define the labels that identify service subsets and post-routing destination configuration. During routing, after a particular route rule (in a virtual service) is selected and consequently the exact destination is determined, then a destination rule configuration for that host (if one exists) is used to control policies, like circuit breakers, loadbalancing policy for the instances/pods implementing that service (or subset), and tls properties to use when calling that particular route destination.

content/docs/concepts/traffic-management/destination-rules/index.md
```diff
@@ -0,0 +1,180 @@
...
+
+{{< image width="50%"
+    link="./destination-rule.svg"
+    caption="Configurable route examples defined in the destination rule"
+    >}}
+
+Visit our [destination rules reference documentation](/docs/reference/config/istio.networking.v1alpha3/#DestinationRule)
+to review all the enabled keys and values.
+
+## Destination rules routing evaluation
```
frankbu: Destination rules have nothing to do with routing. Maybe a better title could be "Destination rules evaluation", but see my following comment about the appropriateness of describing the evaluation algorithm in the Destination rules section of the doc.

content/docs/concepts/traffic-management/destination-rules/index.md
```diff
@@ -0,0 +1,180 @@
...
+
```
frankbu: There is a pervasive incorrect characterization of the role of Pilot in many of these docs (i.e., not just in this section), but here specifically everything it says Pilot does is not correct. Pilot does NOT do any of these things. Pilot's job is to accept user configuration objects, validate them for correctness, and then distribute them to the appropriate Envoy proxies. Envoy does all the runtime evaluation based on the user's configuration it's been given.

content/docs/concepts/traffic-management/destination-rules/index.md
```diff
@@ -0,0 +1,180 @@
...
+Similar to route rules in virtual services, **Pilot** associates policies
+defined in a destination rule with a particular `host:` service. If the
```
frankbu: Pilot doesn't associate policies with hosts, users do.

content/docs/concepts/traffic-management/destination-rules/index.md
```diff
@@ -0,0 +1,180 @@
...
+policies are subset specific, the activation depends on the route rule
+evaluation results.
```
frankbu: Before saying this, you would need to say that policies can optionally be subset-specific.

content/docs/concepts/traffic-management/destination-rules/index.md
```diff
@@ -0,0 +1,180 @@
...
+
+The rule evaluation process is as follows:
```
frankbu: All 4 of the following steps are wrong. I also don't think that describing the rule evaluation algorithm in the Destination rule section of the docs makes sense.

content/docs/concepts/traffic-management/destination-rules/index.md
```diff
@@ -0,0 +1,180 @@
...
+
+1. **Pilot** evaluates if the route rules in the virtual service configuration
+   files correspond to the requested `host:`.
```
frankbu: Step 1. Envoy evaluates the routing rules of a configured virtual service corresponding to the incoming request host, if there is one.
Step 2. If no rule matching the request is found, the request falls through to lower-level round-robin routing.
Step 3. Otherwise, the matching rule's destination host (and optionally subset) is used to locate a destination rule policies that should be applied to the request, if there is one defined.
Step 4. Envoy then routes the request to the selected destination using the destination rule policies that it found, or default policies otherwise.

content/docs/concepts/traffic-management/pilot/index.md
```diff
@@ -0,0 +1,96 @@
...
+objects to configure the routing rules between your services and your external
+resources: virtual services, destination rules, service entries, and gateways.
+You can add and configure these objects in a variety of ways to suit your
+needs. Each configuration object serves a specific purpose and provides
+specific configurations for your Istio service mesh. To create a solution for
+your specific mesh, you can combine the configuration objects in a way that
+best fulfills your needs. You can configure routing rules using any of the four
+network configuration objects in Istio:
+
+* A [virtual service](../virtual-services) configures the routing rules
```
LisaFC: I'd be a bit stronger here, virtual services are kind of core to how the whole traffic management model works. Maybe something like

, which uses routing rules to describe the mapping between one or more user-addressable destinations to the actual destination workloads inside the mesh

content/docs/concepts/traffic-management/traffic-routing/index.md
```diff
@@ -0,0 +1,135 @@
+---
+title: Traffic Routing
```
LisaFC: I'm not sure about this title. Most users will think of what you do with Virtual Services as traffic routing (ie specifying exactly which destination services you want to get which traffic), which is not what this doc is about. Service subsets? Grouping services?

content/docs/concepts/traffic-management/traffic-routing/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+As described in [Pilot's architecture](../pilot), Pilot maintains an
+`Abstract Model` with the canonical representation of the services in a mesh.
+Istio's model of a service is independent of its representation in the
+underlying platform: Kubernetes, Mesos, Cloud Foundry, etc. The
+platform-specific adapters populate the `Abstract Model` representation with
+various fields from the metadata in the platform.
+
+To cover common configuration scenarios, Istio introduces the concept of a
+**service subset**, to subdivide service instances by versions, say `v1` and
```
LisaFC: **service subset**. Defining subsets of services lets you subdivide your service instances by versions....

content/docs/concepts/traffic-management/traffic-routing/index.md
```diff
@@ -0,0 +1,135 @@
...
+`v2`, or environment, say `staging` and `prod`. These variants are not
+necessarily different API versions: they can be iterative changes to the same
+service, deployed in different environments like prod, staging, dev, etc.
+Together with Istio's network configuration objects, you can use subsets to
```
LisaFC: Once you've defined your service subsets, you can use these subsets in Istio's network configuration objects, for example to direct a particular type of traffic to a particular service version in a [virtual service](), or to specify security policies for a particular service version in a [destination rule]().

content/docs/concepts/traffic-management/traffic-routing/index.md
```diff
@@ -0,0 +1,135 @@
...
+configure [routing rules](../routing-rules) that refer to service versions to
+provide additional control over the traffic between services.
+
+This concept covers the common configuration scenarios for services subsets,
+like:
```
LisaFC: This page describes some of the scenarios where you might want to use service subsets, including:

content/docs/concepts/traffic-management/traffic-routing/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+- [Canary rollout](#canary)
+- [Splitting traffic between versions](#splitting) for A/B testing
+- [Circuit breakers](../failures/#circuit)
+
+## Communication between services {#communication}
```
LisaFC: Again, odd title - having service subsets doesn't reeeeeally allow communication between services, but the Istio config/traffic model does allow for easy communication between services that have multiple versions/subsets.  Maybe "Subsets and Virtual Services"? This whole section is a bit of a repeat of some of the stuff we've covered in the other pages, particularly around virtual services. I'd be tempted to take it out, move some of it elsewhere, or cut it down dramatically and link to the virtual services and Pilot/Envoy pages for details.

content/docs/concepts/traffic-management/traffic-routing/index.md
```diff
@@ -0,0 +1,135 @@
...
+functions. The number of instances implementing the services can scale up and
+down based on traffic load completely orthogonal to the version traffic
+routing. This model makes managing a canary version in the presence of
+autoscaling a much simpler problem. To learn more about canary deployments
+and the interoperability of canary deployments and autoscaling when using
+Istio visit our blog post on [Canary Deployments](/blog/2017/0.1-canary/).
+
+## Splitting traffic between versions {#splitting}
+
+Each route rule identifies one or more weighted workloads to call when the rule
```
LisaFC: Each route rule in a [virtual service]...

content/docs/concepts/traffic-management/traffic-routing/index.md
```diff
@@ -0,0 +1,135 @@
...
+is activated. Each workload corresponds to a specific version of the
+destination service. You use subsets to express the versions. If there are
+multiple registered instances with the specified subset, **Pilot** routes them
```
LisaFC: Pilot -> Envoy

content/docs/concepts/traffic-management/traffic-routing/index.md
```diff
@@ -0,0 +1,135 @@
...
+        host: reviews
+        subset: v1
+      weight: 75
+    - destination:
+        host: reviews
+        subset: v2
+      weight: 25
+{{< /text >}}
+
+To learn more about configuring destination rules, visit our [Destination Rules concept](../destination-rules).
```
LisaFC: Shouldn't we be sending them to the virtual services page, not the destination rules page?

content/docs/concepts/traffic-management/virtual-services/index.md
```diff
@@ -0,0 +1,168 @@
+---
+title: Virtual Services
+description: Describes the architecture and behavior of virtual services and provides configuration examples.
+weight: 3
+keywords: [routing, request, traffic routing, requests routing, virtual service, subset, routing rules]
+aliases:
+---
+
```
LisaFC: As per comment above, I think we should combine this and the routing rules doc.

content/docs/concepts/traffic-management/virtual-services/index.md
```diff
@@ -0,0 +1,168 @@
...
+A [virtual service](/docs/reference/config/istio.networking.v1alpha3/#VirtualService)
+configures the rules controlling how **Pilot** routes requests for a service
+within an Istio service mesh. A virtual service contains an ordered list of
+[routing rules](../routing-rules/).
+
+You can use virtual services for fine-grained traffic control. Istio can use
+virtual services to configure ingress, canary, traffic rewrite, or traffic
+policing. A virtual service is not required for simple connectivity. A service
```
LisaFC: You can use virtual services for fine-grained traffic control, including configuring ingress traffic and routing to specific service versions. However, you don't require a virtual service for simple connectivity.

content/docs/concepts/traffic-management/virtual-services/index.md
```diff
@@ -0,0 +1,168 @@
...
+can map one or more user-addressable destinations to the actual destination
+workloads inside the mesh without a virtual service. The Istio virtual services
```
LisaFC: "A service can...."

Is this right?

LisaFC: Using virtual services you can configure traffic routes to:

content/docs/concepts/traffic-management/virtual-services/index.md
```diff
@@ -0,0 +1,168 @@
...
+-  Other network configuration objects in the mesh.
+
+{{< image width="60%"
+    link="./virtual-services-1.svg"
+    caption="Possible configurations between virtual services and workloads"
+    >}}
+
+As you can see, virtual services are very flexible and can accommodate a wide
+variety of topologies. Virtual services configure the [traffic routes](../traffic-routing)
+to your application services. The Istio virtual services can use any DNS names
```
LisaFC: Virtual services configure...

remove this sentence, redundant.

content/docs/concepts/traffic-management/virtual-services/index.md
LisaFC: Also I have a few more edits to do on this doc but I need to catch a train and I'm aware this is getting a bit epic! :)

rcaballeromx:
@frankbu Thank you for taking a look again. I'll work with @LisaFC to address what has become incorrect or inaccurate through my changes. However, let me explain why and how the new content introduced inaccuracies. The original content is accurate by providing little detail and remaining vague. My being more explicit and clear in the new content causes those flaws to surface.
I can correct the inaccuracies you find in the new content and I can change the structure to fix any flow issues but the original content requires a lot more than just minor tweaks.

geeknoid:
@rcaballeromx Do you still intend to move forward with this PR?

geeknoid:
I'm closing this PR. Please reopen if you decide to move forward with it.

rcaballeromx:
That is fine, the next iteration does not need this PR. Thanks for closing @geeknoid

PR #2784, opened 2018-10-18
WIP: Helm docs

jjasghar:
This will fix the script so you can run it and just get the values from the values.yaml.


scripts/tablegen/values.yaml
```diff
@@ -0,0 +1,278 @@
+# Description: Gateways Configuration, refer to the charts/gateways/values.yaml
```
sdake: shouldn't need to check this file in.  As you can tell now from looking at the script it can use some improvement :)

scripts/tablegen/tablegen.py
```diff
@@ -75,11 +78,12 @@ def decode_helm_yaml(s):
     totalNum = len(context)
 
     for lineNum in range(0, totalNum):
+
         if  context[lineNum].strip().startswith('- '):
             pass
         elif '#' in context[lineNum] and '#' == context[lineNum].lstrip()[0]:
-            if "Description: " in context[lineNum]:
-                desc = context[lineNum].strip()
+            desc = context[lineNum].strip()
```
sdake: Our original prototype for this work had this technique of generating desc from a tag.  We sort of decided it was unwieldy and might be prone to error, especially if developers didn't understand how to write the tag.  What would be preferred is parsing the comments character `#` that proceeds any K/V pair, joining it together, and removing the `#` and any extra whitespace.  Since we have to modify a whole lot of values.yaml (each of the subcharts has a `values.yaml` that needs parsing as well) this should yield a pretty quick docification of `values.yaml` throughout the codebase.

scripts/tablegen/requirements.txt
```diff
@@ -0,0 +1,2 @@
+wget==3.2
+ruamel.yaml==0.15.74
```
sdake: curious why the pin on this specific version of rumal?

scripts/tablegen/requirements.txt
```diff
@@ -0,0 +1,2 @@
+wget==3.2
```
sdake: if we can not use absolute pins that would be preferred 👍

scripts/tablegen/tablegen.py
```diff
@@ -18,6 +18,9 @@
 import linecache
 import string
 import sys
+import wget
```
sdake: should ideally be alphabetical.  I think once your improvements in this script are done, we can run it through pep8 and make sure it lints well.

scripts/tablegen/tablegen.py
```diff
@@ -132,7 +136,13 @@ def decode_helm_yaml(s):
         print d
         if "<!-- AUTO-GENERATED-START -->" in d:
             break
+    # if os.path.exists("values.yaml"):
```
sdake: ya, agree the usage of direct file access kind of sucks :)

scripts/tablegen/tablegen.py
```diff
@@ -132,7 +136,13 @@ def decode_helm_yaml(s):
...
+    #     os.remove("values.yaml")
+    # else:
+    #     print("values.yaml doesn't exist")
 
+    # values = 'https://raw.githubusercontent.com/istio/istio/master/install/kubernetes/helm/istio/values.yaml'
```
sdake: this is really cool!  I hadn't really considered this - I was instead thinking we would add a command line argument to pass the values in.  This is better.  We do need to be able to handle specific tags.  For the moment, its ok if this is manual, but perhaps the tag or branch could be passed in via cli?

scripts/tablegen/tablegen.py
```diff
@@ -132,7 +136,13 @@ def decode_helm_yaml(s):
...
+    # values_yaml = wget.download(values)
```
sdake: This will be a little challenging as the subcharts also need their values.yaml read.  The current code sorts the outputs by service name.  One option is to read the requirements.yaml (via wget) in the top level chart, and add any values.yaml found in the subcharts (via wget) to the yaml that is retrieved from the branch or tag.

scripts/tablegen/tablegen.py
jjasghar: Fixed :)

sdake:
@jjasghar any progress in your local branches on this PR?  We definitely need this work for 1.1 (early January).

Cheers
-steve

jjasghar:
No where I'd like it to be. I was going to go heads down over my holiday to give it another shot though.

geeknoid:
@jjasghar Do you still intend to move forward with this PR?

Thanks.

jjasghar:
I haven't been able to give it any focus lately. If you'd like to close this your are welcome to.

geeknoid:
All righty, closing for now. Please reopen when you're ready.

Thanks.

PR #2786, opened 2018-10-19
Clearly indicate which version of k8s we support.

geeknoid:


linsun:
Thanks!  Looking at the preview, I don't see the warning on the index and platform page.

Do you also want to add a warning to here?

https://deploy-preview-2786--preliminary-istio.netlify.com/docs/setup/kubernetes/

geeknoid:
@linsun I hope you meant you "do" see the note on those pages, right?

I've also got the note on the page you indicate. Sadly, it ends up showing up at the bottom of the page due to how directory pages are generated. I've made a note to extend the site framework to let me put text at the top of directory pages instead of just at the bottom. I'll do that separately.

linsun:
@geeknoid yes, i missed it earlier, i saw it at the bottom now.  I was expecting to be at the top, glad you are fixing it in a separate PR.

/lgtm

PR #2791, opened 2018-10-20
Update Task/Enabling rate limits for redisquota

jsenon:
Update Task/Enabling rate limits, remove validDuration in dimension that is not in redisquota, Move redisserverurl and connectionpoolsize


jsenon:
@geeknoid could you help me, I have already rebase as suggested

PR #2794, opened 2018-10-22
fix(content/docs/setup/kubernetes/minimal-install) filename typo

phspagiari:
Just a very small fix on minimal installation guide 😄


phspagiari:
I signed it!

PR #2818, opened 2018-10-30
Fix service format

supereagle:
Unify the service format.

xref: https://github.com/istio/istio.io/pull/2793#pullrequestreview-166753870


supereagle:
@frankbu  Find another place in the doc and have fixed it.

PR #2842, opened 2018-11-05
Remove static charts.

hklai:
Now that they are built along with daily releases and they can be found
in
https://gcsweb.istio.io/gcs/istio-prerelease/daily-build/master-latest-daily/charts/ for master
or https://gcsweb.istio.io/gcs/istio-prerelease/daily-build/release-1.1-latest-daily/charts/ for release-1.1.


geeknoid:
Don't we want these charts published on istio.io?

hklai:
These files were built manually and they go out of sync easily when helm
config files are updated.

Now that we build charts in daily/monthly releases and they are available
at stable locations like
https://gcsweb.istio.io/gcs/istio-prerelease/daily-build/master-latest-daily/charts/.
We should not need to maintain these one off copy in istio.io.

sdake:
@geeknoid - @hklai 's logic is sound.  I had asked @tiswanso to chase down a T&R process for the helm chart builds with @hklai while I was on PTO.  Looks like he did that 👍

We have never documented (that I am aware) how to use the charts from istio.io.  As long as they are built properly (helm package -u istio isito-remote istio-cni), I don't think a redirect is necessary.

It would however be ideal to be able to provide a redirect - although I'm not sure if helm can process one or not.

cheers
-steve

sdake:
@hklai I am curious, do you plan to build charts for 1.y.z or only start with 1.1.z releases?

hklai:
@sdake Tim mentioned that redirect likely won't work. See https://github.com/helm/helm/issues/3531

I think you did some surgery to the helm files/structure in order to build these charts. The same build step does not work in release-1.0 (I tried and explicitly disabled it for release-1.0). We will need to apply the same change to release-1.0 if we want these in 1.0.4 and onward.

For now, these are built in both master and release-1.1.

PR #2864, opened 2018-11-12
Add trulia.com to our exception list, since it is causing a lot of fa…

geeknoid:
…lse positive lint failures.


geeknoid:
Aha, got it!

PR #2870, opened 2018-11-13
add a restriction about application UID 1337

vadimeisenbergibm:
Solves https://github.com/istio/istio/issues/5408.


content/docs/setup/kubernetes/spec-requirements/index.md
```diff
@@ -29,3 +29,5 @@ cluster must satisfy the following requirements:
   `app` label is used to add contextual information in distributed
   tracing. The `app` and `version` labels are also used to add contextual information
   in the metric telemetry collected by Istio.
+
+* _**Application UIDs**_: Applications must not run as a user with user ID (UID) 1337.
```
rcaballeromx: Please avoid using both bold and italics. IMO bold would be enough. Do not use lists for emphasis. Every list requires at least two items. I suggest:
```suggestion
{{< warning_icon >}} **Application UIDs**: Do **not** run applications as a user with the user ID (UID) value of **1337**.
```

vadimeisenbergibm: @rcaballeromx Thank you for your review. I am OK with your comments, but I just followed the style of the original document. It uses bold and italics together, uses a list, it specifies the requirements in the third person and it does not use warning icons. I do not claim that the original style is better, my point is that we have to rewrite the whole document to make it uniform.

I propose the following - let's commit this PR according to the current style of the document. Then you can propose style changes to the whole document, or you may want to submit a PR that would change the style/wording of the whole document. Then we can discuss your proposal in the context of the whole document.

rcaballeromx: @vadimeisenbergibm I agree that the formatting can stay as it is on the rest of the doc. A new PR at a later date to address the is the right approach.
The changes in the language still need to happen though.
```suggestion
* _**Application UIDs**_: Do **not** run applications as a user with the user ID (UID) value of **1337**.
```

vadimeisenbergibm: @rcaballeromx Agree, done in https://github.com/istio/istio.io/pull/2870/commits/469ca0cbdbf48144c7330bf5ef544e563f548c55.

vadimeisenbergibm:
Closing this PR due to a failed merge.

PR #2874, opened 2018-11-14
Delete Service Graph

clyang82:
Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

istio/istio#9066 - ServiceGraph is deprecated in 1.1 in favor of Kiali. Remove the document accordingly.


clyang82:
from this PR - https://github.com/istio/istio.io/pull/2715, just add a note for service graph deprecated. We don't want new users to discover and start using it. so propose to delete this staff. @geeknoid it is up to you. Thanks.

PR #2876, opened 2018-11-14
update chart requirements

itaysk:
adding `helm dep up` which is required for the install to succeed

### DIFF
--- a/content/docs/setup/kubernetes/helm-install/index.md
+++ b/content/docs/setup/kubernetes/helm-install/index.md
@@ -98,6 +98,7 @@ to manage the lifecycle of Istio.
 1. Install Istio:
 
     {{< text bash >}}
+    $ helm repo add istio.io "https://storage.googleapis.com/istio-prerelease/daily-build/master-latest-daily/charts"
     $ helm dep update install/kubernetes/helm/istio
     $ helm install install/kubernetes/helm/istio --name istio --namespace istio-system
     {{< /text >}}
--- a/content/docs/setup/kubernetes/helm-install/index.md
+++ b/content/docs/setup/kubernetes/helm-install/index.md
@@ -38,6 +38,8 @@ plane and the sidecars for the Istio data plane.
 
 ## Installation steps
 
+The following commands has relative references in the Istio directory, and should be executed in the root of the Istio  directory.
+
 1. If using a Helm version prior to 2.10.0, install Istio's [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)
 via `kubectl apply`, and wait a few seconds for the CRDs to be committed in the kube-apiserver:
 
--- a/content/docs/setup/kubernetes/helm-install/index.md
+++ b/content/docs/setup/kubernetes/helm-install/index.md
@@ -38,7 +38,7 @@ plane and the sidecars for the Istio data plane.
 
 ## Installation steps
 
-The following commands has relative references in the Istio directory, and should be executed in the root of the Istio directory.
+The following commands have relative references in the Istio directory. You must execute the commands in Istio's root directory.
 
 1. If using a Helm version prior to 2.10.0, install Istio's [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)
 via `kubectl apply`, and wait a few seconds for the CRDs to be committed in the kube-apiserver:
--- a/content/docs/setup/kubernetes/helm-install/index.md
+++ b/content/docs/setup/kubernetes/helm-install/index.md
@@ -38,7 +38,7 @@ plane and the sidecars for the Istio data plane.
 
 ## Installation steps
 
-The following commands has relative references in the Istio directory, and should be executed in the root of the Istio  directory.
+The following commands has relative references in the Istio directory, and should be executed in the root of the Istio directory.
 
 1. If using a Helm version prior to 2.10.0, install Istio's [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)
 via `kubectl apply`, and wait a few seconds for the CRDs to be committed in the kube-apiserver:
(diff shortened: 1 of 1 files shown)


jmazzitelli:
This PR isn't entirely correct (at least it isn't complete). This PR only changes instructions for the option 2 (the tiller option). But additional things for option 1 also need to be done. And I believe additional things even for option 2 need to be done.

It is more than just `helm dep update`. You have to tell the user to "helm init" for option 1, we can tell the user `helm init --client-only` since Tiller is not used.

Finally, before the update, I believe `helm add repo` must be done - and I think this is true for both option 1 and 2.

So for option 1 it needs to be something like:

```
helm init --client-only
helm repo add istio.io "https://storage.googleapis.com/istio-prerelease/daily-build/master-latest-daily/charts"
helm dep update install/kubernetes/helm/istio
```

Option 2 I believe you need to add the `helm repo add`.

itaysk:
>before the update, I believe helm add repo must be done - and I think this is true for both option 1 and 2.

This is unnecessary since the chart is local: The guide starts with 'Download the Istio Release' and the `helm install` command points to the chart in tree and not in a repo.

>You have to tell the user to "helm init"

The guide starts with a prerequisite that clearly say: 'Install the Helm client' with a link to their docs. I don't know if the Istio docs are the right place to tech the user how to install Helm but that's not my call, and I think this is a separate discussion.

theute:
For the records, I was having an issue (with snapshot 2 release), (I was trying to reproduce a user issue and succeeded by just following the doc):

`Error: found in requirements.yaml, but missing in charts/ directory: sidecarInjectorWebhook, security, ingress, gateways, mixer, nodeagent, pilot, grafana, prometheus, servicegraph, tracing, galley, kiali, certmanager`

And the commands listed by mazz helped me pass that:
`helm init --client-only
helm repo add istio.io "https://storage.googleapis.com/istio-prerelease/daily-build/master-latest-daily/charts"
helm dep update install/kubernetes/helm/istio`

It really didn't come to mind that "installing helm client" would also require to initialize it, for me, it was installed because I could use "helm".

That said, I'm confused now, because when I delete ~/.helm and restart the whole thing without "helm init" and "helm repo add" it works...

Something fishy, but I hope we'll find the right documentation to help people getting started quickly.

jmazzitelli:
In Option 2, it explicitly specifies the `helm init` command (with the `--service-account tiller` option) -- see here:

https://preliminary.istio.io/docs/setup/kubernetes/helm-install/#option-2-install-with-helm-and-tiller-via-helm-install

So it seems to me proper for Option 1 to also tell the user to explicitly run `helm init` (albeit with the `--client-only` option, instead of the `--service-account` option that Option 2 shows).

linsun:
I think we need to add repo cmd too:   https://github.com/istio/istio/pull/9920

cc @clyang82 @hklai @sdake

sdake:
@theute `helm dep update` copies dependent charts into your charts directory.  While it is unintuitive that helm init --client-only is no longer needed after the first `helm dep update`, it does work.

I believe @hklai or @mandarjog is working out a solution using `helm package -u` which produces an index with tarballs of packaged helm-based manifests.  You can see this already with the cni repo which is hosted on `googleapis`  This is our mid-term solution to installation.  I'm not quite sure where this is in development as I've been on the road since Friday and pretty busy.

Cheers
-steve

hklai:
@sdake We already produce index and tarballs in https://storage.googleapis.com/istio-prerelease/daily-build/master-latest-daily/charts for each daily release.

In long term, we probably will put tarballs from different LTS/patch versions in one location with a generated index.

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -38,6 +38,8 @@ plane and the sidecars for the Istio data plane.
 
 ## Installation steps
 
+The following commands has relative references in the Istio directory, and should be executed in the root of the Istio  directory.
+
```
clyang82: remove one space between `Istio` and `directory`

itaysk: done

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -38,6 +38,8 @@ plane and the sidecars for the Istio data plane.
 
 ## Installation steps
 
+The following commands has relative references in the Istio directory, and should be executed in the root of the Istio directory.
```
rcaballeromx: Grammar and passive voice. I suggest:
```suggestion
The following commands have relative references in the Istio directory. You must execute the commands in Istio's root directory.
```

ayj:
@itaysk, @jmazzitelli, @sdake - what are the next steps to address this issue? As far as I can tell, the following command is required, regardless of `helm template` vs `helm install`.

```
helm dep update install/kubernetes/helm/istio
```

geeknoid:
I removed the approved label, since it's not clear to me that we've reach consensus on whether this PR is correct.

itaysk:
If you follow the instructions [here](https://preliminary.istio.io/docs/setup/kubernetes/helm-install/#option-2-install-with-helm-and-tiller-via-helm-install) step by step, you will fail. the missing step is to `helm dep up` (for 1.0.x).

For 1.1, istio-cni chart was added with a repo reference, which requires to add a repo, so I'll add that as well.

For the record, if we have an official, repo why not ship the umbrella chart there (along with the dependencies) and save the user from downloading all of istio just to install? (I can work on that if that's sounds ok).

itaysk:
also, I can add a .gitignore to the chart directory in istio/istio so that the local cache won't be tracked (as result from helm dep up)

sdake:
> @itaysk, @jmazzitelli, @sdake - what are the next steps to address this issue? As far as I can tell, the following command is required, regardless of `helm template` vs `helm install`.
> 
> ```
> helm dep update install/kubernetes/helm/istio
> ```

@ayj I would recommend using `helm package -u install/kubernetes/install/helm/istio" to produce a packaged tar.gz.  This tarball can then be distributed in the release.  `helm template` or `helm install` both understand how to read the tarball.  This eliminates the need for `helm dep` step (this is what the `-u` flag does.

Cheers
-steve

sdake:
@geeknoid the PR is correct, although I suspect we need a slight modification to the release process to fine-tune how the packaging work.  As such, I've approved this PR.  I think we will need another rev once we find out the final location of the release repo for the `helm repo add` step.  At that time if T&R has sorted out a method with `helm package -u` to produce tarballs in the released tree, we will need to revise.

@tiswanso was tracking this work - not sure where/if it was finished (it was part of the CNI effort).

@itaysk can you take a look at getting the CLA signed?

Cheers
-steve

PR #2879, opened 2018-11-15
new blog post on incremental istio usage for traffic management

crcsmnky:


content/blog/2018/incremental-traffic-management/index.md
```diff
@@ -0,0 +1,225 @@
...
+subtitle:
+attribution: Sandeep Parikh
+twitter: crcsmnky
+weight: 95
+keywords: [traffic-management, gateway]
+---
+
+Traffic management is one of the critical benefits provided by Istio. At the heart of Istio’s traffic management is the ability to decouple traffic flow and infrastructure scaling. This lets you control your traffic in ways that aren’t possible without a service mesh like Istio.
+
+For example, let’s say you want to execute a [canary deployment](https://martinfowler.com/bliki/CanaryRelease.html). With Istio, you can specify that **v1** of a service receives 90% of incoming traffic, while **v2** of that service only receives 10%. With standard Kubernetes deployments, the only way to achieve this is to manually control the number of available Pods for each version, for example 9 Pods running v1 and 1 Pod running v2. This type of manual control is hard to implement, and over time may have trouble scaling.
```
frankbu: There is a blog post for this that would be good to reference: https://istio.io/blog/2017/0.1-canary/

content/blog/2018/incremental-traffic-management/index.md
```diff
@@ -0,0 +1,225 @@
...
+  "stars": 4,
+  "color": "red"
+}
+{{< /text >}}
+
+Notice how we’re getting responses from all three versions of the reviews service (`null` is from reviews v1 which doesn’t have ratings) and not getting the even split across v1 and v2. This is expected behavior because the `curl` command is using Kubernetes service load balancing across all three versions of the reviews service. In order to access the reviews 50/50 split we need to access the service via the ingress Gateway:
+
+{{< text bash >}}
+$ for i in `seq 4`; do \ 
+  kubectl exec -it $SLEEP_POD curl http://istio-ingressgateway.istio-system/reviews/0 | \
```
frankbu: This seems a little hacky. If you were going to do something like this in production, you would probably want to configure DNS for a real host instead of configuring host `*` and then just calling the ingressgateway service directly. I wonder if you could use the new coredns addon that's used for multi-cluster config: https://preliminary.istio.io/docs/setup/kubernetes/multicluster-install/gateways/#configure-dns

Of course that starts to move away from the minimal Istio install theme of this article.

content/blog/2018/incremental-traffic-management/index.md
```diff
@@ -0,0 +1,225 @@
...
+    labels:
+      version: v3
+EOF
+{{< /text >}}
+
+Finally, **deploy a Pod that you can use for testing** with `curl`:
+
+{{< text bash >}}
+$ kubectl apply -f samples/sleep/sleep.yaml
+{{< /text >}}
```
frankbu: You should point out that the automatic sidecar injection is not enabled, so this is just starting the sleep pod with 1 container, instead of the usual 2 (app and sidecar proxy) that people are used to when using it.

content/blog/2018/incremental-traffic-management/index.md
crcsmnky: Addressed by adding a link to the recommended post to the "canary" paragraph.

content/blog/2018/incremental-traffic-management/index.md
crcsmnky: Agreed, this moves in the direction of increased complexity so keeping it simpler for this post.

content/blog/2018/incremental-traffic-management/index.md
crcsmnky: Addressed by including a comment that the sleep Pod is deployed **without** the sidecar container

crcsmnky:
Updated post to reflect requested feedback from @frankbu. And while biased, my preference is that is appears as a blog post since it's focused on an early stage example of how to start adopting Istio.

frankbu:
Please fix the circleci errors. Otherwise looks good.

frankbu:
@crcsmnky The "correct" way to handle spelling issues for resource types like VirtualService and DestinationRule is to put back-ticks around the resource name, e.g., `DestinationRule` instead of adding them to the .spelling file. I see we have some cases where people have done the latter, but it would be best if we try to not continue with the incorrect solution. Could you please remove DestinationRule from .spelling and use the back-ticks instead. Thanks.

crcsmnky:
Hi @frankbu I'd like to see this post go out sooner rather than waiting on the next release. Should I file a PR against 1.0.3 branch or can you backport this post into that branch yourself. Let me know what you'd prefer.

frankbu:
@crcsmnky Send another PR against the release-1.0 branch with a comment that says it's the same as this PR (#2879). When that one merges, it will immediately appear on the istio.io site.

PR #2881, opened 2018-11-15
Fix sni_host option

ThomasWo:
This tripped me up as this page doesn't agree with the examples[1] and the pilot validator seems to want to take `sni_host`.

[1]: https://istio.io/docs/tasks/traffic-management/egress/


geeknoid:
This file is auto-generated from the original source code in the istio/api repo. Please make these fixes in that repo, and they'll get copied here.

Thanks.

PR #2893, opened 2018-11-17
clarify match example

itaysk:


geeknoid:
This file is automatically generated from the original sources in the istio/api repo. Please make this change in that repo, and it'll automatically be migrated into this doc.

Thanks.

PR #2916, opened 2018-11-26
An example for configuring and verifying split horizon EDS

ymesika:
This is a new example to be added to the multicluster examples that demonstrates how to set up a two-clusters multicluster using the Split Horizon EDS that will be introduced in v1.1.


content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
+---
+title: Multicluster with Split Horizon EDS
+description: Example of leveraging Istio's Split Horizon EDS to create a Multi-Cluster.
```
linsun: Multi-Cluster -- should we use Multicluster to be consistent?

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
+---
+title: Multicluster with Split Horizon EDS
```
linsun: probably good to spell out EDS once.

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+description: Example of leveraging Istio's Split Horizon EDS to create a Multi-Cluster.
+weight: 85
+keywords: [kubernetes,multicluster]
+---
+
+This example demonstrates how to use a variation of the [Istio multicluster](/docs/setup/kubernetes/multicluster-install/vpn/) together with _Split Horizon EDS_ (introduced in Istio 1.1) to join two Kubernetes clusters together to allow access to remote instances of a service via the remote Istio ingress gateway. By following the steps in this example you will setup a topology of two clusters as described in the following diagram:
+
```
linsun: I think it is confusion to say a variation of the multicluster vpn ... because you are not using VPN for your solution. :).  should we call it single mesh multicluster?

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+  {{< image width="80%" ratio="36.01%"
+  link="./diagram.svg"
+  caption="Istio mesh spanning multiple Kubernetes clusters with the Split Horizon EDS configured"
```
linsun: Istio mesh -> Single Istio mesh?

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+  >}}
+
+The `local` cluster hosts the Istio Pilot and other Istio components while the `remote` cluster only runs the Istio Citadel, Sidecar Injector and Ingress gateway.
```
linsun: other Istio components -> other Istio control plane components

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+However, unlike the requirements of [Istio multicluster](/docs/setup/kubernetes/multicluster-install/vpn/), no VPN connectivity is required nor direct network access between workload from different clusters.
+
```
linsun: It would be useful to use this reasoning highlight why you use split EDS over the vpn.

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+## Before you begin
+
+In addition to the prerequisites for installing Istio the following setup is required for this example:
+
+* Two Kubernetes clusters with Load Balancers configured (also known as `local` and `remote`)
```
linsun: can user use ingress gateway as the load balancer?

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+{{< text bash >}}
+$ export CTX_LOCAL=<KUBECONFIG_LOCAL_CONTEXT>
+$ export CTX_REMOTE=<KUBECONFIG_REMOTE_CONTEXT>
+{{< /text >}}
+
+## Example multicluster setup
+
+This example procedure installs Istio with mutual TLS enabled for the control plane and application pods. We create the `cacerts` secret on both local and remote by using the same Istio certificate from samples as the shared root CA.
+
+The steps below also set up a remote cluster with a selector-less service and an endpoint for `istio-pilot.istio-system` that has the address of the local Istio ingress gateway. To allow that we are securely exposing the local pilot to remote cluster(s) through the ingress gateway without terminating the mTLS.
```
linsun: To allow that.... -- this reads odd.   I think you meant something like This enables us to securely expose the local pilot to remote cluster(s)....

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+
+    By default the `global.meshNetworks` value for Istio is empty but we will need to modify it to declare a new network for endpoints on the remote cluster. Modify `install/kubernetes/helm/istio/values.yaml` and add a `network2` declaration:
+
+    {{< text yaml >}}
+    meshNetworks:
+      network2:
+        endpoints:
+        - fromRegistry: remote_kubecfg
+        gateways:
+        - address: 192.23.120.32
```
linsun: How would user know this remote gateway addr before istio gateway is installed onto the remote cluster (which is step after this)?

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+    {{< /text >}}
+
+    Although deployed locally this Gateway instance will also affect the remote cluster by configuring it to passthrough incoming traffic to the relevant remote service (SNI-based) but keeping the mTLS all the way from the source to destination sidecars.
+
+1. Deploy the files:
+
+    {{< text bash >}}
+    $ kubectl create --context=$CTX_LOCAL -f helloworld-v1.yaml -n sample
+    $ kubectl create --context=$CTX_LOCAL -f helloworld-gateway.yaml -n sample
+    {{< /text >}}
```
linsun: Any virtual service and dr need to be deployed?  I don't see how you bind the helloworld gateway to the helloworld service.  Or is that not needed with SNI based routing?

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,377 @@
...
+$ kubectl delete --context=$CTX_REMOTE ns istio-system
+$ kubectl delete --context=$CTX_REMOTE -f helloworld-v2.yaml -n sample
+{{< /text >}}
+
+Cleanup the local cluster:
+
+{{< text bash >}}
+$ kubectl delete --context=$CTX_LOCAL -f istio-auth.yaml
+$ kubectl delete --context=$CTX_LOCAL ns istio-system
+$ kubectl delete --context=$CTX_LOCAL -f helloworld-v1.yaml -n sample
```
linsun: do we need to add instruction to clean up CRDs?

ymesika: It's kind of redundant if you delete the `istio-system` namespace but I don't mind adding it.

content/docs/examples/multicluster/split-horizon-eds/index.md
ymesika: Nope. Using the new AUTO_PASSTHROUGH tls mode in the `Gateway` object. It is also being used for the gateways-based multicluster.

content/docs/examples/multicluster/split-horizon-eds/index.md
linsun: Is it?   the CRDs are not namespace scoped, so I was not sure if removing istio-system ns would remove it.

content/docs/examples/multicluster/split-horizon-eds/index.md
linsun: yes, that would be good, thanks @ymesika

content/docs/examples/multicluster/split-horizon-eds/index.md
linsun: Ok I have one more q, how does AUTO_PASSTHROUGH tls mode in the gateway object knows to pass through to the helloworld service when you have multiple sample services deployed in the sample namespace?

content/docs/examples/multicluster/split-horizon-eds/index.md
ymesika: Yes, you are right.

content/docs/examples/multicluster/split-horizon-eds/index.md
ymesika: SNI-based

ymesika:
@rshriram I'm not sure I understand what are the changes you are asking for or is it just a generic rejection of yours to adding this example?

rshriram:
my point being this doc will needlessly complicate the already complicated multicluster setup. If you want, you could replace the existing multicluster setup with VPN completely, and use this doc (along the way remove any mention of words like split horizon blah)..And please dont use words like single mesh as it generates unnecessary confusion and implies a mesh is same as a control plane.

frankbu:
I agree with @rshriram that we need to get the multicluster docs aligned around some clear terminology and not use words like split-horizon in user docs. I think the gist of this example is to show how you can setup a multicluster configuration using a single controlplane in an environment without VPN connectivity between the clusters. So I think it needs to be organized in the docs accordingly but the problem is that the current organization of the whole bunch of multi-xxx docs is a bit of a mess, and with no clear overview of what it's all about.

I'm planning to take a pass at a high level overview and organization in the next few days, so I would propose putting this PR on hold until that's done, and hopefully it should then be more clear where this material should go and how the wording should be changed.

frankbu:
@ymesika #2991 gives an overview of all the multicluster concepts. Please link to it and update your title and wording to align with it.

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -0,0 +1,379 @@
+---
+title: Multicluster with Split Horizon EDS
```
frankbu: How about: `title: Location-aware Multicluster Service Routing` ?

content/docs/examples/multicluster/_index.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Single L3 Network Multcluster Mesh
-description: A variety of fully working VPN-based multicluster examples for Istio that you can experiment with.
+title: Single/Multi L3 Network Multicluster Mesh
```
frankbu: How about `title: Multicluster Service Mesh` ?

content/docs/examples/multicluster/_index.md
```diff
@@ -1,6 +1,6 @@
...
+description: A variety of fully working multicluster examples for Istio that you can experiment with.
 weight: 100
 keywords: [multicluster]
 ---
```
frankbu: Maybe add a link to concept doc, like I did in the setup section in #2991.

PR #2936, opened 2018-12-01
zh:content/docs/setup/kubernetes/minimal-install/index.md

MJ-CJM:


MJ-CJM:
@fleeto thank you very much for your review，I have taken your comments and revised it.

PR #2941, opened 2018-12-03
update kiali screenshots for the kiali task documentation

jmazzitelli:


jmazzitelli:
/hold

looks like the chinese docs refers to the images, and I renamed one of them. I'm putting a hold on this PR so I can rename it back. Plus, I think I need to take different snapshots.

PR #2946, opened 2018-12-04
Fix typo

pilhuhn:
spec.target should be spec.targets


pilhuhn:
@googlebot CLA has been signed - I am part of Red Hat CLA

Issue #2947, opened 2018-12-05
Document ingressgateway service externalIps in helm chart


cwocwo:
Add document for `gateways.istio-ingressgateway.externalIPs` in `reference/config/installation-options/index.md`.

PR #2948, opened 2018-12-05
Gateway service support externalIps in helm charts #2947

cwocwo:
Add document for `gateways.istio-ingressgateway.externalIPs` in `reference/config/installation-options/index.md`. 
Related to istio issue: https://github.com/istio/istio/issues/9557


linsun:
is this a config nob we expose?  I don't see it.

I had thought this page is auto generated too cc @sdake FYI

linsun:
as the PR is for master... i believe the doc is targeted for 1.1

PR #2960, opened 2018-12-07
An example of configuring access to an external legacy HTTPS proxy

vadimeisenbergibm:
Addresses https://github.com/istio/istio/issues/7302

### DIFF
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -1,0 +1,233 @@
+---
+title: Connect to an external HTTPS proxy
+description: Describes how to configure Istio to let applications use an external HTTPS proxy.
+weight: 60
+keywords: [traffic-management,egress]
+---
+The [Configure an Egress Gateway](docs/examples/advanced-gateways/egress-gateway/) showed how you can direct traffic to
+external services from your mesh via an Istio edge component called _Egress Gateway_. However, there are cases when you
+must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your company may already
+have such a proxy in place and all the applications within the organization may be required to direct their traffic
+through such proxy. This task shows how to enable access to an external HTTPS proxy.
+
+## Before you begin
+
+*   Setup Istio by following the instructions in the [Installation guide](/docs/setup/).
+
+*   Start the [sleep]({{< github_tree >}}/samples/sleep) sample
+    which will be used as a test source for external calls via the proxy.
+
+    If you have enabled
+    [automatic sidecar injection](/docs/setup/kubernetes/sidecar-injection/#automatic-sidecar-injection), do
+
+    {{< text bash >}}
+    $ kubectl apply -f @samples/sleep/sleep.yaml@
+    {{< /text >}}
+
+    otherwise, you have to manually inject the sidecar before deploying the `sleep` application:
+
+    {{< text bash >}}
+    $ kubectl apply -f <(istioctl kube-inject -f @samples/sleep/sleep.yaml@)
+    {{< /text >}}
+
+    Note that any pod that you can `exec` and `curl` from would do.
+
+*   Create a shell variable to hold the name of the source pod for sending requests to external services.
+    If you used the [sleep]({{<github_tree>}}/samples/sleep) sample, run:
+
+    {{< text bash >}}
+    $ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})
+    {{< /text >}}
+
+## Deploy an HTTPS proxy
+
+For this example, to simulate a legacy proxy, you deploy an HTTPS proxy inside your cluster. To simulate the fact that
+in the real life your proxy may be external to your cluster, you will address the pod of the proxy by its IP address and
+not by a Kubernetes service.
+You can use any HTTPS proxy that supports HTTP Connect. We used [Squid](http://www.squid-cache.org).
+
+1.  Create a namespace for the HTTPS proxy. Note that since you do not label it for Istio automatic sidecar injection,
+    Istio will not control traffic in this namespace.
+
+    {{< text bash >}}
+    $ kubectl create namespace external
+    {{< /text >}}
+
+1.  Create a configuration file for the Squid proxy.
+
+    {{< text bash >}}
+    $ cat <<EOF > ./proxy.conf
+    http_port 3128
+
+    acl SSL_ports port 443
+    acl CONNECT method CONNECT
+
+    http_access deny CONNECT !SSL_ports
+    http_access allow localhost manager
+    http_access deny manager
+    http_access allow all
+
+    coredump_dir /var/spool/squid
+    EOF
+    {{< /text >}}
+
+1.  Create a Kubernetes [ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/)
+    to hold the configuration of the proxy:
+
+    {{< text bash >}}
+    $ kubectl create configmap proxy-configmap -n external --from-file=squid.conf=./proxy.conf
+    {{< /text >}}
+
+1.  Deploy a container with Squid:
+
+    {{< text bash >}}
+    $ kubectl apply -f - <<EOF
+    apiVersion: extensions/v1beta1
+    kind: Deployment
+    metadata:
+      name: squid
+      namespace: external
+    spec:
+      replicas: 1
+      template:
+        metadata:
+          labels:
+            app: squid
+        spec:
+          volumes:
+          - name: proxy-config
+            configMap:
+              name: proxy-configmap
+          containers:
+          - name: squid
+            image: sameersbn/squid:3.5.27
+            imagePullPolicy: IfNotPresent
+            volumeMounts:
+            - name: proxy-config
+              mountPath: /etc/squid
+              readOnly: true
+    EOF
+    {{< /text >}}
+
+1.  Deploy the [sleep]({{< github_tree >}}/samples/sleep) sample in the `external` namespace to test traffic to the
+    proxy without Istio traffic control.
+
+    {{< text bash >}}
+    $ kubectl apply -n external -f @samples/sleep/sleep.yaml@
+    {{< /text >}}
+
+1.  Define an environment variable to hold the IP address of the proxy pod:
+
+    {{< text bash >}}
+    $ export PROXY_IP=$(kubectl get pod -n external -l app=squid -o jsonpath={.items..podIP})
+    {{< /text >}}
+
+1.  Define an environment variable to hold the port of your proxy. The deployment of Squid in this example uses port
+    3128.
+
+    {{< text bash >}}
+    $ export PROXY_PORT=3128
+    {{< /text >}}
+
+1.  Send a request from the `sleep` pod in the `external` namespace to an external service via the proxy:
+
+    {{< text bash >}}
+    $ kubectl exec -it $(kubectl get pod -n external -l app=sleep -o jsonpath={.items..metadata.name}) -n external -- sh -c "HTTPS_PROXY=$PROXY_IP:$PROXY_PORT curl https://en.wikipedia.org/wiki/Main_Page" | grep -o "<title>.*</title>"
+    <title>Wikipedia, the free encyclopedia</title>
+    {{< /text >}}
+
+1.  Check the access log of the proxy and see your request:
+
+    {{< text bash >}}
+    $ kubectl exec -it $(kubectl get pod -n external -l app=squid -o jsonpath={.items..metadata.name}) -n external -- tail -f /var/log/squid/access.log
+    1544160065.248    228 172.30.109.89 TCP_TUNNEL/200 87633 CONNECT en.wikipedia.org:443 - HIER_DIRECT/91.198.174.192 -
+    {{< /text >}}
+
+## Configure traffic to external HTTPS proxy
+
+1.  Define a TCP (!) Service Entry for the HTTPS proxy. Note that despite the fact that HTTP
+    [CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6) method is used to communicate with HTTPS proxies, the traffic between the application and the proxies is TCP (a TCP tunnel), and not HTTP.
+
+    {{< text bash >}}
+    $ kubectl apply -f - <<EOF
+    apiVersion: networking.istio.io/v1alpha3
+    kind: ServiceEntry
+    metadata:
+      name: proxy
+    spec:
+      hosts:
+      - my-company-proxy.com # ignored
+      addresses:
+      - $PROXY_IP/32
+      ports:
+      - number: $PROXY_PORT
+        name: tcp
+        protocol: TCP
+      location: MESH_EXTERNAL
+    EOF
+    {{< /text >}}
+
+1.  Send a request from the `sleep` pod in the `default` namespace. The `sleep` pod has Istio sidecar injected and its
+    traffic is controlled by Istio.
+
+    {{< text bash >}}
+    $ kubectl exec -it $SOURCE_POD -c sleep -- sh -c "HTTPS_PROXY=$PROXY_IP:$PROXY_PORT curl https://en.wikipedia.org/wiki/Main_Page" | grep -o "<title>.*</title>"
+    <title>Wikipedia, the free encyclopedia</title>
+    {{< /text >}}
+
+1.  Check the Istio sidecar proxy's logs and see your request appear:
+
+    {{< text bash >}}
+    $ kubectl logs $SOURCE_POD -c istio-proxy
+    [2018-12-07T10:38:02.841Z] "- - -" 0 - 702 87599 92 - "-" "-" "-" "-" "172.30.109.95:3128" outbound|3128||my-company-proxy.com 172.30.230.52:44478 172.30.109.95:3128 172.30.230.52:44476 -
+    {{< /text >}}
+
+1.  Check the access log of the proxy and see your request:
+
+    {{< text bash >}}
+    $ kubectl exec -it $(kubectl get pod -n external -l app=squid -o jsonpath={.items..metadata.name}) -n external -- tail -f /var/log/squid/access.log
+    1544160065.248    228 172.30.109.89 TCP_TUNNEL/200 87633 CONNECT en.wikipedia.org:443 - HIER_DIRECT/91.198.174.192 -
+    {{< /text >}}
+
+## Understanding what happened
+
+In this example you configured access to an external legacy HTTPS proxy from the Istio service mesh. To simulate a
+remote HTTPS proxy, you deployed an HTTPS proxy in a separate namespace, without Istio automatic sidecar injection. In addition, for simulation of a remote proxy, you addressed the HTTPS proxy by its IP address and not by a Kubernetes
+service.
+
+To enable Istio-controlled traffic to the external HTTPS proxy you created a TCP ServiceEntry with the IP address and
+ the port of the proxy.
+
+## Cleanup
+
+1.  Shutdown the [sleep]({{<github_tree>}}/samples/sleep) service:
+
+    {{< text bash >}}
+    $ kubectl delete -f @samples/sleep/sleep.yaml@
+    {{< /text >}}
+
+1.  Shutdown the [sleep]({{<github_tree>}}/samples/sleep) service in the `external` namespace:
+
+    {{< text bash >}}
+    $ kubectl delete -f @samples/sleep/sleep.yaml@ -n external
+    {{< /text >}}
+
+1.  Shutdown the Squid proxy, remove the config map and the configuration file:
+
+    {{< text bash >}}
+    $ kubectl delete -n external deployment squid
+    $ kubectl delete -n external configmap proxy-configmap
+    $ rm ./proxy.conf
+    {{< /text >}}
+
+1.  Delete the `external` namespace:
+
+    {{< text bash >}}
+    $ kubectl delete namespace external
+    {{< /text >}}
+
+1.  Delete the Service Entry:
+
+    {{< text bash >}}
+    $ kubectl delete serviceentry proxy
+    {{< /text >}}    
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -4,7 +4,7 @@ description: Describes how to configure Istio to let applications use an externa
 weight: 60
 keywords: [traffic-management,egress]
 ---
-The [Configure an Egress Gateway](docs/examples/advanced-gateways/egress-gateway/) showed how you can direct traffic to
+The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gateway/) showed how you can direct traffic to
 external services from your mesh via an Istio edge component called _Egress Gateway_. However, there are cases when you
 must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your company may already
 have such a proxy in place and all the applications within the organization may be required to direct their traffic
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -8,7 +8,11 @@ The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gatewa
 external services from your mesh via an Istio edge component called _Egress Gateway_. However, there are cases when you
 must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your company may already
 have such a proxy in place and all the applications within the organization may be required to direct their traffic
-through it. This example shows how to enable access to an external HTTPS proxy.
+through it.
+
+This example shows how to enable access to an external HTTPS proxy. Since access to HTTPS proxies is performed by the
+HTTP [CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6) method, configuring traffic to an external HTTPS
+proxy is different from configuring traffic to external HTTP and HTTPS services.
 
 ## Before you begin
 
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -8,7 +8,7 @@ The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gatewa
 external services from your mesh via an Istio edge component called _Egress Gateway_. However, there are cases when you
 must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your company may already
 have such a proxy in place and all the applications within the organization may be required to direct their traffic
-through such proxy. This example shows how to enable access to an external HTTPS proxy.
+through it. This example shows how to enable access to an external HTTPS proxy.
 
 ## Before you begin
 
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -143,6 +143,10 @@ You can use any HTTPS proxy that supports HTTP Connect. We used [Squid](http://w
     1544160065.248    228 172.30.109.89 TCP_TUNNEL/200 87633 CONNECT en.wikipedia.org:443 - HIER_DIRECT/91.198.174.192 -
     {{< /text >}}
 
+At this point the proxy has been deployed and tested by using `curl` to access `wikipedia.org` through the proxy, all
+without Istio. It's just plain Kubernetes setting so far, which simulates an external HTTPS proxy.
+In the following section you are going to configure traffic from Istio-enabled pods to the HTTPS proxy.
+
 ## Configure traffic to external HTTPS proxy
 
 1.  Define a TCP (!) Service Entry for the HTTPS proxy. Note that despite the fact that HTTP
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -195,7 +195,7 @@ In this example you configured access to an external legacy HTTPS proxy from the
 remote HTTPS proxy, you deployed an HTTPS proxy in a separate namespace, without Istio automatic sidecar injection. In addition, for simulation of a remote proxy, you addressed the HTTPS proxy by its IP address and not by a Kubernetes
 service.
 
-To enable Istio-controlled traffic to the external HTTPS proxy you created a TCP ServiceEntry with the IP address and
+To enable Istio-controlled traffic to the external HTTPS proxy you created a TCP service entry with the IP address and
  the port of the proxy.
 
 ## Cleanup
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -150,7 +150,8 @@ In the following section you are going to configure traffic from Istio-enabled p
 ## Configure traffic to external HTTPS proxy
 
 1.  Define a TCP (!) Service Entry for the HTTPS proxy. Note that despite the fact that HTTP
-    [CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6) method is used to communicate with HTTPS proxies, the traffic between the application and the proxies is TCP (a TCP tunnel), and not HTTP.
+    [CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6) method is used to communicate with HTTPS proxies,
+    the traffic between the application and the proxies is TCP (a TCP tunnel), and not HTTP.
 
     {{< text bash >}}
     $ kubectl apply -f - <<EOF
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -201,7 +201,9 @@ remote HTTPS proxy, you deployed an HTTPS proxy in a separate namespace, without
 service.
 
 To enable Istio-controlled traffic to the external HTTPS proxy you created a TCP service entry with the IP address and
- the port of the proxy.
+ the port of the proxy. Note that you must not create service entries for the external services you access though the
+ external proxy, like `wikipedia.org`. This is because from Istio's point of view the requests are sent to the
+ external proxy only; Istio is not aware of the fact that the external proxy forwards the requests further.
 
 ## Cleanup
 
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -8,7 +8,7 @@ The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gatewa
 external services from your mesh via an Istio edge component called _Egress Gateway_. However, there are cases when you
 must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your company may already
 have such a proxy in place and all the applications within the organization may be required to direct their traffic
-through such proxy. This task shows how to enable access to an external HTTPS proxy.
+through such proxy. This example shows how to enable access to an external HTTPS proxy.
 
 ## Before you begin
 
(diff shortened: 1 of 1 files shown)


content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Understanding what happened
+
+In this example you configured access to an external legacy HTTPS proxy from the Istio service mesh. To simulate a
+remote HTTPS proxy, you deployed an HTTPS proxy in a separate namespace, without Istio automatic sidecar injection. In addition, for simulation of a remote proxy, you addressed the HTTPS proxy by its IP address and not by a Kubernetes
+service.
+
+To enable Istio-controlled traffic to the external HTTPS proxy you created a TCP service entry with the IP address and
+ the port of the proxy.
+
```
frankbu: I think you should elaborate here. For example, explain what is fundamentally different than when you don't use a proxy like why didn't you need to create service entries for the external host (en.wikipedia.org or wildcard), like you would normally do.

vadimeisenbergibm:
@frankbu Thank you for your review, I have applied your remarks. More comments are welcome.

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    If you used the [sleep]({{<github_tree>}}/samples/sleep) sample, run:
+
+    {{< text bash >}}
+    $ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})
+    {{< /text >}}
+
+## Deploy an HTTPS proxy
+
+For this example, to simulate a legacy proxy, you deploy an HTTPS proxy inside your cluster. To simulate the fact that
+in the real life your proxy may be external to your cluster, you will address the pod of the proxy by its IP address and
```
frankbu: I suggest changing this:
```
To simulate the fact that in the real life your proxy may be external to your cluster,
```
to this:
```
Also, to simulate a more realistic proxy that is running outside of your cluster,
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    {{< /text >}}
+
+1.  Send a request from the `sleep` pod in the `external` namespace to an external service via the proxy:
+
+    {{< text bash >}}
+    $ kubectl exec -it $(kubectl get pod -n external -l app=sleep -o jsonpath={.items..metadata.name}) -n external -- sh -c "HTTPS_PROXY=$PROXY_IP:$PROXY_PORT curl https://en.wikipedia.org/wiki/Main_Page" | grep -o "<title>.*</title>"
+    <title>Wikipedia, the free encyclopedia</title>
+    {{< /text >}}
+
+1.  Check the access log of the proxy and see your request:
```
frankbu: s/and see/for/

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    {{< /text >}}
+
+At this point the proxy has been deployed and tested by using `curl` to access `wikipedia.org` through the proxy, all
+without Istio. It's just plain Kubernetes setting so far, which simulates an external HTTPS proxy.
+In the following section you are going to configure traffic from Istio-enabled pods to the HTTPS proxy.
+
+## Configure traffic to external HTTPS proxy
+
+1.  Define a TCP (!) Service Entry for the HTTPS proxy. Note that despite the fact that HTTP
+    [CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6) method is used to communicate with HTTPS proxies,
```
frankbu: that the HTTP

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Fixed in https://github.com/istio/istio.io/pull/2960/commits/9aa4998ec562ca281ee7c85f45e1a851794055f7

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Fixed in https://github.com/istio/istio.io/pull/2960/commits/c0c2d9ff62d6e3e8a31ef7cb691013821f1cc297

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Fixed in https://github.com/istio/istio.io/pull/2960/commits/2ea97e6e2b9c4682a29b7e108ae88b8b60c80822

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    the traffic between the application and the proxies is TCP (a TCP tunnel), and not HTTP.
```
frankbu: and the proxy is TCP

vadimeisenbergibm: Fixed in https://github.com/istio/istio.io/pull/2960/commits/869e7f46763e91b3ee6672afe396c3b3ffc39c04

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+title: Connect to an external HTTPS proxy
+description: Describes how to configure Istio to let applications use an external HTTPS proxy.
+weight: 60
+keywords: [traffic-management,egress]
+---
+The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gateway/) showed how you can direct traffic to
+external services from your mesh via an Istio edge component called _Egress Gateway_. However, there are cases when you
+must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your company may already
+have such a proxy in place and all the applications within the organization may be required to direct their traffic
+through it. This example shows how to enable access to an external HTTPS proxy.
```
frankbu: I would start a new paragraph after `through it.` and then explain why this scenario is different. I'm not sure this is exactly right, but something like this anyway:

```
though it.

This example shows how to enable access to an external HTTPS proxy.
Configuring access via such a proxy is complicated by the fact that Istio does not support
the HTTP protocol expected of proxy clients and therefore needs to be configured to simply
pass through the requests using a TCP tunnel.
```

frankbu: Yes, I think you are getting to heart of what is bothering me about this example. It seems like a lot of work to setup the proxy, test it, configure istio, test again, just for something that is not very complicated. It almost seems like a lot of noise that could be replaced by a paragraph or two in the docs somewhere. That said, if we do want the example, I think we at least have to give a very brief intro that explains why it's special. I agree that it should be short and high level but not quite as high level as you suggest (just needs to be configured "in some other way") :-)

vadimeisenbergibm: @frankbu Note this comment https://github.com/istio/istio/issues/7302#issuecomment-420725490:

> would love to make this a task or an entry in the guide

and it got three likes. So my guess is that people are interested in such a task/example. See the issue itself https://github.com/istio/istio/issues/7302, it seems it caused some pain to some people.

Regarding Istio examples - my view on them is as follows. An example does not necessarily has to present some complex or special solution, it is just a scenario that some people can implement. In this case, the scenario is special, the solution is simple (with the catch of TCP entry instead of HTTP). The people who have to work with an HTTPS proxy, can follow this example and see that it works (or does not work) in their environment. The example demonstrates to both Istio developers and Istio users that this particular setting works in Istio. Ideally, we should automate all our examples  and test them for every PR.

>  I think we at least have to give a very brief intro that explains why it's special.

Let me add a sentence about HTTP CONNECT.

vadimeisenbergibm: Addressed in https://github.com/istio/istio.io/pull/2960/commits/57fdcd843fa927612770fe774ba35579b71cfe4b

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+    otherwise, you have to manually inject the sidecar before deploying the `sleep` application:
+
+    {{< text bash >}}
+    $ kubectl apply -f <(istioctl kube-inject -f @samples/sleep/sleep.yaml@)
+    {{< /text >}}
+
+    Note that any pod that you can `exec` and `curl` from would do.
+
+*   Create a shell variable to hold the name of the source pod for sending requests to external services.
```
rcaballeromx: Please refer to the variable as an environment variable. It also helps if you use the name of the variable in the text for readers to link the text and the code block. For example:
```suggestion
*   Create the `SOURCE_POD` environment variable to store the `kubectl` command to send requests to external services from the source pod.
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    If you used the [sleep]({{<github_tree>}}/samples/sleep) sample, run:
```
rcaballeromx: Not needed, you already established that you are using the sleep sample. It helps to focus on only one flow. In this case, the flow concerning the sleep sample. Avoid referencing other possible flows to avoid confusing your readers.

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+    {{< text bash >}}
+    $ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})
+    {{< /text >}}
+
+## Deploy an HTTPS proxy
+
+For this example, to simulate a legacy proxy, you deploy an HTTPS proxy inside your cluster. Also, to simulate a more
```
rcaballeromx: Readers already are on this example, the conditional clause is thus redundant. I suggest:
```suggestion
To simulate a legacy proxy, deploy an HTTPS proxy inside your cluster. To simulate a more
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+realistic proxy that is running outside of your cluster, you will address the pod of the proxy by its IP address and
```
rcaballeromx: Verb congruence and simplicity issues. Avoid switching verb tenses, this paragraph has sentences in present, past, and future which is confusing to readers. To fix this issue and simplify the content, I suggest:
```suggestion
realistic proxy running outside of your cluster, address the proxy's pod through its IP address instead of```

File: content/docs/examples/advanced-gateways/http-proxy/index.md [hunk 11]
```diff
@@ -0,0 +1,240 @@
...
+not by a Kubernetes service.
```
rcaballeromx:
```suggestion
through the Kubernetes service.
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+You can use any HTTPS proxy that supports HTTP Connect. We used [Squid](http://www.squid-cache.org).
```
rcaballeromx: Emphasize and focus on the current flow. By switching the order of the sentences, you achieve that emphasis. Remember to avoid past tense.
```suggestion
This example uses [Squid](http://www.squid-cache.org) but you can use any HTTPS proxy that supports HTTP Connect.
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+1.  Create a namespace for the HTTPS proxy. Note that since you do not label it for Istio automatic sidecar injection,
```
rcaballeromx: How can readers know that no automatic sidecar injection is in use? Avoid assuming that readers will be able to identify the same things as you do from the commands. Try to be specific and to write the steps devoid of assumptions, for example:
```suggestion
1.  Create the `external` namespace for the HTTPS proxy. The command does not configure automatic sidecar injection for the `external` namespace.
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    Istio will not control traffic in this namespace.
```
rcaballeromx:
```suggestion
    Thus, Istio does not control the traffic in this namespace.
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+    {{< text bash >}}
+    $ kubectl create namespace external
+    {{< /text >}}
+
+1.  Create a configuration file for the Squid proxy.
+
+    {{< text bash >}}
```
rcaballeromx: I would advise you to avoid using here files in the doc. They can overwhelm the readers. If possible, store the configuration file in the repo and link to it when you introduce the command. The document becomes a lot easier to consume if the commands do not contain the entire configuration. If you want to reference and explain specific configuration values, which I would welcome, you can still do so if the configuration file is stored in the repo.

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    to hold the configuration of the proxy:
+
+    {{< text bash >}}
+    $ kubectl create configmap proxy-configmap -n external --from-file=squid.conf=./proxy.conf
+    {{< /text >}}
+
+1.  Deploy a container with Squid:
+
+    {{< text bash >}}
+    $ kubectl apply -f - <<EOF
```
rcaballeromx: This is a great example of a configuration file that does not need to be a here document. I would suggest you have this config stored as squid.yaml and then the commands would become much simpler and easier to consume.

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    {{< /text >}}
+
+1.  Deploy the [sleep]({{< github_tree >}}/samples/sleep) sample in the `external` namespace to test traffic to the
+    proxy without Istio traffic control.
+
+    {{< text bash >}}
+    $ kubectl apply -n external -f @samples/sleep/sleep.yaml@
+    {{< /text >}}
+
+1.  Define an environment variable to hold the IP address of the proxy pod:
```
rcaballeromx: Be specific, name the variable you are creating to provide an easy way to link text and code example for readers. Think of your audience, will they be able to tell that you are obtaining the IP address through the command? I would advise you change the text to make that part of the command more explicit, since it is more important than the creation of the environment variable:
```suggestion
1.  Obtain the IP address of the proxy pod and define the `PROXY_IP` environment variable to store it:
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+    {{< text bash >}}
+    $ export PROXY_IP=$(kubectl get pod -n external -l app=squid -o jsonpath={.items..podIP})
+    {{< /text >}}
+
+1.  Define an environment variable to hold the port of your proxy. The deployment of Squid in this example uses port
```
rcaballeromx: Simplify the content to make it easier to read:
```suggestion
1.  The Squid deployment in this example uses **port 3128**. Store the value of the proxy's port in the `PROXY_PORT` environment variable:
```

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+    {{< text bash >}}
+    $ export PROXY_PORT=3128
+    {{< /text >}}
+
+1.  Send a request from the `sleep` pod in the `external` namespace to an external service via the proxy:
+
+    {{< text bash >}}
+    $ kubectl exec -it $(kubectl get pod -n external -l app=sleep -o jsonpath={.items..metadata.name}) -n external -- sh -c "HTTPS_PROXY=$PROXY_IP:$PROXY_PORT curl https://en.wikipedia.org/wiki/Main_Page" | grep -o "<title>.*</title>"
+    <title>Wikipedia, the free encyclopedia</title>
```
rcaballeromx: Please do not combine the command and the output in the same code block. There is a discrepancy between the description of what the command does in line 132 and what the actual command is doing. You are not only sending a request, you are sending that request to a specific service and grepping for a specific pattern. Please describe accurately what operations the reader is taking with a command.

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    {{< /text >}}
+
+1.  Check the access log of the proxy for your request:
+
+    {{< text bash >}}
+    $ kubectl exec -it $(kubectl get pod -n external -l app=squid -o jsonpath={.items..metadata.name}) -n external -- tail -f /var/log/squid/access.log
+    1544160065.248    228 172.30.109.89 TCP_TUNNEL/200 87633 CONNECT en.wikipedia.org:443 - HIER_DIRECT/91.198.174.192 -
```
rcaballeromx: Please do not include the output in the same code block as the command and without any explanation or context for the reader. In this specific instance, I don't think you need to include the output but, if you want to explain the output, that would be welcome. Just make sure you provide an explanation to the readers.

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    {{< /text >}}
+
+At this point the proxy has been deployed and tested by using `curl` to access `wikipedia.org` through the proxy, all
```
rcaballeromx: I would advise you to change this to a list for clarity and readability.
```suggestion
So far, you completed the following tasks without Istio:

- You deployed the HTTPS proxy.
- You used `curl` to access the `wikipedia.org` external service through the proxy.

The next step is to configure the traffic from the Istio-enabled pods to use the HTTPS proxy.
```

content/docs/examples/advanced-gateways/http-proxy/index.md
rcaballeromx: Passive voice issue: What is using the HTTP CONNECT method to communicate with the HTTPS proxies?

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+without Istio. It's just plain Kubernetes setting so far, which simulates an external HTTPS proxy.
+In the following section you are going to configure traffic from Istio-enabled pods to the HTTPS proxy.
+
+## Configure traffic to external HTTPS proxy
+
+1.  Define a TCP (!) Service Entry for the HTTPS proxy. Note that despite the fact that the HTTP
+    [CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6) method is used to communicate with HTTPS proxies,
+    the traffic between the application and the proxy is TCP (a TCP tunnel), and not HTTP.
```
rcaballeromx: This clause only adds to the confusion create by the passive voice. Why is this fact important? How does it relate to the service entry?

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    1544160065.248    228 172.30.109.89 TCP_TUNNEL/200 87633 CONNECT en.wikipedia.org:443 - HIER_DIRECT/91.198.174.192 -
+    {{< /text >}}
+
+At this point the proxy has been deployed and tested by using `curl` to access `wikipedia.org` through the proxy, all
+without Istio. It's just plain Kubernetes setting so far, which simulates an external HTTPS proxy.
+In the following section you are going to configure traffic from Istio-enabled pods to the HTTPS proxy.
+
+## Configure traffic to external HTTPS proxy
+
+1.  Define a TCP (!) Service Entry for the HTTPS proxy. Note that despite the fact that the HTTP
```
rcaballeromx: Avoid using capitalization for emphasis, it should be "service entry". I would end the step with the first sentence of this line to introduce the command. What does (!) mean? Is that something you want to make readers particularly aware of? I'm all for it but consider rewriting the text to make that emphasis more apparent. You can also use {{< warning_icon >}} to get the symbol I think you wanted to use.

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    [CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6) method is used to communicate with HTTPS proxies,
+    the traffic between the application and the proxy is TCP (a TCP tunnel), and not HTTP.
+
+    {{< text bash >}}
+    $ kubectl apply -f - <<EOF
```
rcaballeromx: See my previous comments regarding the use of here documents.

rcaballeromx:
@frankbu Why was this content merged so quickly? I was barely able to start reviewing it and it has a ton of issues. Can you give me some guidance as to what time window I have to review content? It is incredibly frustrating to lose so much work.

vadimeisenbergibm:
@rcaballeromx I will submit an additional PR to address your comments, so your work will not be lost.

frankbu:
@rcaballeromx sorry if I merged too quickly. I suppose we need a better way to coordinate. If you are willing to review every doc PR, that is great, but we need to know what you are doing. How about you mark any PR that you are looking at ASAP with /hold, so that we know not to merge prematurely?

As @rcaballeromx said, he will make the changes in a followup PR, so there's no problem continuing to review after the merge. This material won't actually go live at isito.io until 1.1 is released anyway.

One thing that I did notice is that your suggested changes in the before you begin section apply to a whole bunch of other examples as well. Those same setup instructions have been copy/pasted into several of the examples, so they will need to be updated in all of them.

vadimeisenbergibm:
@frankbu Should I create a separate task "deploy sleep container" and reference it from all other tasks/examples? To prevent duplication.

vadimeisenbergibm:
@frankbu ^ or just reference this section from some task in all other tasks/examples?

frankbu:
@vadimeisenbergibm I think the instructions are too short to warrant a separate task. I would reference them from one task or better yet, since it is so short and simple, just fix the grammar as @rcaballeromx suggests, and then copy/paste the result into all of the tasks/examples that need it. After that's done it shouldn't be a problem anymore.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: @rcaballeromx Let me disagree with you here. Grepping the output is a minor, unimportant detail that is not worth mentioning. It is done just to give the user a clue that the Wikipedia site was indeed accessed successfully. The user can just output the whole page to the screen - it will be less neat and the user will have to look at the long output to see if the request was successful.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: @rcaballeromx I would like to raise the issue of the passive voice again. cc @frankbu @geeknoid 

According to https://writing.stackexchange.com/questions/16812/technical-documentation-voice-preference-active-voice-passive-voice-a, which cites an APA manual:

> The passive voice is acceptable in expository writing and when you want to focus on the object or recipient of the action rather than on the actor. For example, "The speakers were attached to either side of the chair" emphasizes the placement of speakers, not who placed them—the more appropriate focus in the Method section. "The President was shot" emphasizes the importance of the person shot. (p. 77)

Kubernetes documentation uses passive voice extensively:

Setup https://kubernetes.io/docs/setup/

> These solutions allow you to create Kubernetes clusters with only a few commands and are actively developed

> They can also be hosted on a range of Cloud IaaS providers

Concepts: https://kubernetes.io/docs/concepts/

> The master can also be replicated for availability and redundancy.

Install kubectl
https://kubernetes.io/docs/tasks/tools/install-kubectl/

> which is created automatically
>  The completion script itself is generated by kubectl
>  which is not installed by default
> so it is automatically loaded

My claim is that passive voice should be used to omit obvious or unimportant details.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: It explains why a TCP service entry is used and not an HTTP one.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: My reservation about the present tense here is as follows: the reader creates a new namespace and then, after the creation, Istio **will not** control the traffic in the newly created namespace. 

Saying "Create a namespace and Istio controls traffic in it" sounds strange to me, though I may be wrong here.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: @rcaballeromx cc @geeknoid 
So where should we put the files? I used `here` files extensively until now. @geeknoid should we put the files in some folder in this repository and then address them by `kubectl` commands?

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your company may already
+have such a proxy in place and all the applications within the organization may be required to direct their traffic
+through it. This example shows how to enable access to an external HTTPS proxy.
+
+## Before you begin
+
+*   Setup Istio by following the instructions in the [Installation guide](/docs/setup/).
+
+*   Start the [sleep]({{< github_tree >}}/samples/sleep) sample
+    which will be used as a test source for external calls via the proxy.
```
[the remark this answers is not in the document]
vadimeisenbergibm: Addressed in https://github.com/istio/istio.io/pull/2973/commits/47d6a2118cadb288ffc7e20d9aa8fbb75d57cf19

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+    If you have enabled
```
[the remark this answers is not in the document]
vadimeisenbergibm: The meaning is different: if you previously enabled the automatic sidecar injection, now you can do ... otherwise you have to manually inject the sidecar...

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -0,0 +1,240 @@
...
+    $ kubectl apply -f @samples/sleep/sleep.yaml@
+    {{< /text >}}
+
+    otherwise, you have to manually inject the sidecar before deploying the `sleep` application:
+
+    {{< text bash >}}
+    $ kubectl apply -f <(istioctl kube-inject -f @samples/sleep/sleep.yaml@)
+    {{< /text >}}
+
+    Note that any pod that you can `exec` and `curl` from would do.
```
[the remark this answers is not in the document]
vadimeisenbergibm: How about the following sentence: `Note that you can use as a test source any pod that has curl installed.` ?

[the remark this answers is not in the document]
vadimeisenbergibm: Addressed in https://github.com/istio/istio.io/pull/2973/commits/400829c272e31ffe6ec4ef412711ac9d40913e14

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Addressed in a379b4e13a4e749a039b7aab08f199d902b76a82

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Addressed in a138dba0c850651b88f5af7f2d558486b183c9ca.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: The intention is "For the sake of this example only, to simulate a legacy proxy"... it stresses the fact that the proxy is deployed only for example, it does not have to be deployed in the real life.

vadimeisenbergibm: Addressed in 4862dc5a387052bca4d5c3cf10c51de80e7f808c

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Addressed in c6b560784a8c04bb596bbb58fa6cafedfffaf72c, the meaning is "by the domain name of a Kubernetes service".

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: > We used Squid

The issue is a bit complicated. If I say "we use Squid", in the present tense, it would sound like we and not the reader use Squid in this example. The meaning here is "We decided to use Squid to describe the example, so you, the reader, will use the Squid in this example".

rcaballeromx:
@vadimeisenbergibm Thank you for opening the new pull request.
@frankbu I won't be able to review every single PR, there are just too many. I will follow your advise and will comment with /hold before I start my review to better sync our efforts.
Regarding the Before you begin section, I agree with @frankbu that having them as a separate task, although easier to maintain, is not adequate due to the size of the content and the possible flexibility required for different tasks. I've created issue #2975 to track that effort.

content/docs/examples/advanced-gateways/http-proxy/index.md
rcaballeromx: @vadimeisenbergibm This topic is indeed a challenging one but let me try to explain why, passive voice is not appropriate for this type of content.
1. As you point out, passive voice does have a role in writing. There are reasons why it exists and the APA manual cites them very well. However, the American Press Association manual pertains to journalistic content, not technical content. The oldest and best established style guide for developer facing content is the [Microsoft Style guide](https://docs.microsoft.com/en-us/style-guide/welcome/) which recommends active voice. If you are interested, I can provide you a full literature list of technical writing books which emphasize the importance of active voice in technical content. OpenStack, the Linux kernel, Google and many other projects also have technical content guidelines recommending the use of active voice.
2. It is very difficult to accurately know what is obvious or unimportant to your readers. The best way we have found to leave as few readers behind as possible, is to attempt to be as explicit, clear, specific, and accurate as possible. In this particular case, emphasizing the target and omitting the actor makes your content harder to grasp and less specific.
3. I will raise the issue of passive voice in Kubernetes with that technical writing team but please keep in mind that, similar to Istio, not every PR from every contributor gets a full review by an editor or a tech. writer. Thus, instances of passive voice might slip through. Not because they are considered correct, but because there was no chance to correct them. The [Kubernetes content best practices](https://kubernetes.io/docs/contribute/style/style-guide/#content-best-practices) make the clear recommendation to use active voice, with the sole exception its use leading to strange constructs. That is not the case in this instance.
The use of active voice in technical content is and has been a technical writing best practice for many years and across many languages. If you want more information or more resources regarding this topic, please let me know.

frankbu: I understand the general guidance but I also think that some of @vadimeisenbergibm's examples fall into a legitimate exception category (active voice adds unimportant details). In [this document](https://writing.wisc.edu/Handbook/CCS_activevoice.html) they call out exactly this case as an exception.
```
To describe a condition in which the actor is unknown or unimportant
```
Passive voice:
```
Every year, thousands of people are diagnosed as having cancer. 
```
Active voice would be adding unimportant detail to the message, e.g.:
```
Every year, doctors diagnose thousands of people as having cancer.
```
I think this is the same issue Vadim is having, at least in some of the cases.

vadimeisenbergibm: @rcaballeromx Thank you for the extensive answer and for the references. My point is as follows: While I agree we should strive to use active voice as much as possible, there could be cases where passive voice could be beneficial. So using passive voice should not be rejected outright, we should weigh the alternatives in each case.

My concrete proposal is: in every PR the authors will document all the cases of the passive voice and explain why they used them and which alternatives they considered. This way we will have all the passive voice cases under control, and it will be clear that passive voice was not added by omission, but intentionally by the author. Then we can discuss if each case merits using passive voice.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Applied in 4954637b98186dbb4b447c448651d4fad7b90e3e

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Note that the name of the namespace is unimportant, the readers can be confused and will think that the name `external` has some special meaning. This is why I wrote "Create a namespace for the HTTPS proxy [the name does not matter]".

vadimeisenbergibm: The lack of automatic sidecar injection clarified in 1d64eb7c284bdcdbd650b76e88e1aaf6efdefbcb

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Fixed in 726c6a05c7535005b32e1d93d8803d15a09c52cd

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: @rcaballeromx Additional point here: some readers will use a real proxy and not the simulated one. For them, the important part is the first one: "Define an environment variable to store the port of your proxy". The sentence about the port 3128 is secondary.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Applied in e26ef515948f351e821a632b6b9886def8e32bfe

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: @rcaballeromx Agreed, for the new PR I listed the passive voice cases in the PR's first comment.

frankbu: @rcaballeromx I agree with you, although your wording in this example is not quite correct and misses the point. Let me give it a try.

Note that even though applications use the HTTP CONNECT method to communicate with HTTPS proxies, Istio will not handle these requests as HTTP traffic. It will instead forward them as opaque TCP requests (i.e., a TCP tunnel).

rcaballeromx: @frankbu That's why I love collaborating on content, if I may improve upon your version:
Although applications use the HTTP CONNECT method to communicate with HTTPS proxies, Istio doesn't handle these requests as HTTP traffic. Instead, Istio forwards them as opaque TCP requests, for example, as a TCP tunnel.
Unlike before, my suggestions here are mostly copy-editing to fix things like verb congruence, unclear pronoun references, and that sort of thing.

frankbu: just change "for example", to "that is" and I like it.

@vadimeisenbergibm WDYT?

frankbu: Actually I think you need to also change "doesn't" to "won't" and "forwards" to "will forward", because this will only be the case after you apply the configuration that comes right after this sentence.

Although applications use the HTTP CONNECT method to communicate with HTTPS proxies, Istio won't handle these requests as HTTP traffic. Instead, Istio will forward them as opaque TCP requests, that is, as a TCP tunnel.

frankbu: Another try:

Although applications use the HTTP CONNECT method to establish connections with HTTPS proxies, you must configure the proxy for TCP traffic, instead of HTTP, because once a connection is established the proxy simply acts as a TCP tunnel.

vadimeisenbergibm: Fixed in https://github.com/istio/istio.io/pull/2973/commits/96f9545aa7c16404aa79479b710daaaae8759002

PR #2970, opened 2018-12-11
Document browser connection reuse problem

frankbu:


content/help/ops/traffic-management/deploy-guidelines/index.md
```diff
@@ -232,3 +232,31 @@ To make sure services will have zero down-time when configuring routes with subs
     1. Wait a few seconds for the `VirtualService` configuration to propagate to the Envoy sidecars.
 
     1. Update the `DestinationRule` to remove the unused subsets.
+
+## Browser problem when multiple gateways configured with same TLS cert
```
geeknoid: cert -> certificate

content/help/ops/traffic-management/deploy-guidelines/index.md
```diff
@@ -232,3 +232,31 @@ To make sure services will have zero down-time when configuring routes with subs
...
+
+Configuring more than one gateway using the same TLS certificate will cause browsers
+that leverage [HTTP/2 connection reuse](https://httpwg.org/specs/rfc7540.html#reuse)
+(i.e., most browsers) to produce 404 errors when accessing a second host after a
+connection to another host has already been established.
+
+For example, let's say you have 2 hosts that share the same TLS cert like this:
```
geeknoid: cert -> certificate

content/help/ops/traffic-management/deploy-guidelines/index.md
```diff
@@ -232,3 +232,31 @@ To make sure services will have zero down-time when configuring routes with subs
...
+
+* Wildcard certificate `*.test.com` installed in `istio-ingressgateway`
+* `Gateway` configuration `gw1` with host `service1.test.com`, selector `istio: ingressgateway`, and TLS config using gateway's mounted (wildcard) cert
+* `Gateway` configuration `gw2` with host `service2.test.com`, selector `istio: ingressgateway`, and TLS config using gateway's mounted (wildcard) cert
+* `VirtualService` configuration `vs1` with host `service1.test.com` and gateway `gw1`
+* `VirtualService` configuration `vs2` with host `service2.test.com` and gateway `gw2`
+
+Since both gateways are served by the same workload (i.e., selector `istio: ingressgateway`) requests to both services
+(`service1.test.com` and `service2.test.com`) will resolve to the same IP. If `service1.test.com` is accessed first, it
+will return the wildcard cert (`*.test.com`) indicating that connections to `service2.test.com` can use the same cert.
```
geeknoid: cert -> certificate

.spelling
```diff
@@ -223,6 +223,7 @@ camelCase
 canaried
 canarying
 colocated
+config
```
geeknoid: I don't think we want this in the dictionary.

haf:
This isn't a solution; all Istio has to do is to add logic to merge the config in exactly the same way that you now want every user, ever, to do.

frankbu:
Detecting and then merging arbitrary gateways for this scenario is hardly a trivial piece of logic, but as soon as the problem is fixed (one way or another) we can remove this doc. Otherwise, it is the current solution (workaround).

haf:
That's exactly why it should go into istio core or envoy core rather than forcing every user to deal with it. Until Envoy works, this should be worked around inside Istio IMO.

It should be clear that you're working around a major security issue and that Istio is broken, and it's not the fault of the user.

PR #2972, opened 2018-12-12
Fix external link check error

SataQiu:
Fix external link check error


geeknoid:
I'm going ahead with a different fix for the broken link.

Thanks.

PR #2973, opened 2018-12-12
Improve the HTTPS proxy example

vadimeisenbergibm:
Based on comments of @rcaballeromx in https://github.com/istio/istio.io/pull/2960

Passive voice cases:
* "Istio sidecars will not be injected" since the sidecars are injected automatically, it does not matter which component injects them

### DIFF
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -202,16 +202,19 @@ Next, you must configure the traffic from the Istio-enabled pods to use the HTTP
     {{< /text >}}
 
 ## Understanding what happened
+In this example, you took the following steps to simulate an external legacy HTTPS proxy and configure access to it from
+ an Istio service mesh:
 
-In this example you configured access to an external legacy HTTPS proxy from the Istio service mesh. To simulate a
-remote HTTPS proxy, you deployed an HTTPS proxy in a separate namespace, without Istio automatic sidecar injection.
-In addition, for simulation of a remote proxy, you addressed the HTTPS proxy by its IP address and not by the domain
-name of a Kubernetes service.
+1. Deployed an HTTPS proxy in a separate namespace without Istio automatic sidecar injection to simulate a remote HTTPS
+proxy.
+1. Addressed the HTTPS proxy by its IP address and not by the domain name of a Kubernetes service to maintain the
+simulation of a remote proxy.
+1. Created a TCP service entry with the IP address and the port of the proxy to enable Istio-controlled traffic to the
+proxy.
 
-To enable Istio-controlled traffic to the external HTTPS proxy you created a TCP service entry with the IP address and
- the port of the proxy. Note that you must not create service entries for the external services you access through the
- external proxy, like `wikipedia.org`. This is because from Istio's point of view the requests are sent to the
- external proxy only; Istio is not aware of the fact that the external proxy forwards the requests further.
+Note that you must not create service entries for the external services you access through the external proxy, like
+`wikipedia.org`. This is because from Istio's point of view the requests are sent to the external proxy only; Istio is
+not aware of the fact that the external proxy forwards the requests further.
 
 ## Cleanup
 
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -148,9 +148,12 @@ This example uses [Squid](http://www.squid-cache.org) but you can use any HTTPS 
     1544160065.248    228 172.30.109.89 TCP_TUNNEL/200 87633 CONNECT en.wikipedia.org:443 - HIER_DIRECT/91.198.174.192 -
     {{< /text >}}
 
-At this point the proxy has been deployed and tested by using `curl` to access `wikipedia.org` through the proxy, all
-without Istio. It's just plain Kubernetes setting so far, which simulates an external HTTPS proxy.
-In the following section you are going to configure traffic from Istio-enabled pods to the HTTPS proxy.
+So far, you completed the following tasks without Istio:
+
+* You deployed the HTTPS proxy.
+* You used `curl` to access the `wikipedia.org` external service through the proxy.
+
+The next step is to configure the traffic from the Istio-enabled pods to use the HTTPS proxy.
 
 ## Configure traffic to external HTTPS proxy
 
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -37,8 +37,7 @@ proxy is different from configuring traffic to external HTTP and HTTPS services.
     Note that you can use as a test source any pod that has `curl` installed.
 
 *   Create the `SOURCE_POD` environment variable to store the name of the source pod for sending requests to external
-    services.
-    If you used the [sleep]({{<github_tree>}}/samples/sleep) sample, run:
+    services:
 
     {{< text bash >}}
     $ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -48,7 +48,7 @@ proxy is different from configuring traffic to external HTTP and HTTPS services.
 To simulate a legacy proxy and only for this example, you deploy an HTTPS proxy inside your cluster.
 Also, to simulate a more realistic proxy that is running outside of your cluster, you will address the proxy's pod
 by its IP address and not by the domain name of a Kubernetes service.
-You can use any HTTPS proxy that supports HTTP Connect. We used [Squid](http://www.squid-cache.org).
+This example uses [Squid](http://www.squid-cache.org) but you can use any HTTPS proxy that supports HTTP Connect.
 
 1.  Create a namespace for the HTTPS proxy. Note that since you do not label it for Istio automatic sidecar injection,
     Istio will not control traffic in this namespace.
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -4,7 +4,7 @@ description: Describes how to configure Istio to let applications use an externa
 weight: 60
 keywords: [traffic-management,egress]
 ---
-The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gateway/) example showed how you can direct
+The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gateway/) example shows how you to direct
 traffic to external services from your mesh via an Istio edge component called _Egress Gateway_. However, there are
 cases when you must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your
 company may already have such a proxy in place and all the applications within the organization may be required to
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -156,9 +156,9 @@ The next step is to configure the traffic from the Istio-enabled pods to use the
 
 ## Configure traffic to external HTTPS proxy
 
-1.  Define a TCP (not HTTP!) Service Entry for the HTTPS proxy. Note that despite the fact that the HTTP
-    [CONNECT](https://tools.ietf.org/html/rfc7231#section-4.3.6) method is used to communicate with HTTPS proxies,
-    the traffic between the application and the proxy is TCP (a TCP tunnel), and not HTTP.
+1.  Define a TCP (not HTTP!) Service Entry for the HTTPS proxy. Although applications use the HTTP CONNECT method to
+    establish connections with HTTPS proxies, you must configure the proxy for TCP traffic, instead of HTTP. Once the
+    connection is established, the proxy simply acts as a TCP tunnel.
 
     {{< text bash >}}
     $ kubectl apply -f - <<EOF
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -36,7 +36,8 @@ proxy is different from configuring traffic to external HTTP and HTTPS services.
 
     You can use any pod with `curl` installed as a test source.
 
-*   To send requests to external services, create the `SOURCE_POD` environment variable to store the name of the source pod:
+*   To send requests to external services, create the `SOURCE_POD` environment variable to store the name of the source
+    pod:
 
     {{< text bash >}}
     $ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -45,9 +45,9 @@ proxy is different from configuring traffic to external HTTP and HTTPS services.
 
 ## Deploy an HTTPS proxy
 
-For this example, to simulate a legacy proxy, you deploy an HTTPS proxy inside your cluster. Also, to simulate a more
-realistic proxy that is running outside of your cluster, you will address the pod of the proxy by its IP address and
-not by a Kubernetes service.
+For the sake of this example only, to simulate a legacy proxy, you deploy an HTTPS proxy inside your cluster.
+Also, to simulate a more realistic proxy that is running outside of your cluster, you will address the pod of the proxy
+by its IP address and not by a Kubernetes service.
 You can use any HTTPS proxy that supports HTTP Connect. We used [Squid](http://www.squid-cache.org).
 
 1.  Create a namespace for the HTTPS proxy. Note that since you do not label it for Istio automatic sidecar injection,
--- a/content/docs/examples/advanced-gateways/http-proxy/index.md
+++ b/content/docs/examples/advanced-gateways/http-proxy/index.md
@@ -179,7 +179,7 @@ Next, you must configure the traffic from the Istio-enabled pods to use the HTTP
     EOF
     {{< /text >}}
 
-1.  Send a request from the `sleep` pod in the `default` namespace. The `sleep` pod has Istio sidecar injected so
+1.  Send a request from the `sleep` pod in the `default` namespace. Because the `sleep` pod has a sidecar, 
     Istio controls its traffic.
 
     {{< text bash >}}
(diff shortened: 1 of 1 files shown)


frankbu:
Just noticed error on line 7, it should say: 

The Configure an Egress Gateway example showed how ...

vadimeisenbergibm:
@frankbu Addressed in https://github.com/istio/istio.io/pull/2973/commits/e5ef00f06db87025080601a47dd9fd74f4cbcc4a

rcaballeromx:
I've created issue #2975 to track the effort around fixing all "Before you begin" sections.

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -34,20 +34,20 @@ proxy is different from configuring traffic to external HTTP and HTTPS services.
...
     $ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})
     {{< /text >}}
 
 ## Deploy an HTTPS proxy
 
-For this example, to simulate a legacy proxy, you deploy an HTTPS proxy inside your cluster. Also, to simulate a more
-realistic proxy that is running outside of your cluster, you will address the pod of the proxy by its IP address and
-not by a Kubernetes service.
+For the sake of this example only, to simulate a legacy proxy, you deploy an HTTPS proxy inside your cluster.
+Also, to simulate a more realistic proxy that is running outside of your cluster, you will address the pod of the proxy
```
rcaballeromx: Avoid using words with little to no semantic value to simplify your text. Use the imperative present tense instead of the future tense. Lastly, I suggest using the possessive to move the noun "pod" closer to the pronoun "its" to make the reference more evident:
 ```suggestion
To simulate a more realistic proxy running outside of your cluster, address the proxy's pod
```

**Comment by vadimeisenbergibm** [comment]
@rcaballeromx Thank you for your review, let me perform the changes.

File: content/docs/examples/advanced-gateways/http-proxy/index.md [hunk 2]
```diff
@@ -4,11 +4,11 @@ description: Describes how to configure Istio to let applications use an externa
 weight: 60
 keywords: [traffic-management,egress]
 ---
-The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gateway/) showed how you can direct traffic to
-external services from your mesh via an Istio edge component called _Egress Gateway_. However, there are cases when you
-must use an external, legacy (non-Istio) HTTPS proxy to access external services. For example, your company may already
-have such a proxy in place and all the applications within the organization may be required to direct their traffic
-through it.
+The [Configure an Egress Gateway](/docs/examples/advanced-gateways/egress-gateway/) example showed how you can direct
```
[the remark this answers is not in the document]
vadimeisenbergibm: Fixed in https://github.com/istio/istio.io/pull/2973/commits/2223452c4a53338feade62f918ac45d9eaa3b878

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -19,7 +19,7 @@ proxy is different from configuring traffic to external HTTP and HTTPS services.
 *   Setup Istio by following the instructions in the [Installation guide](/docs/setup/).
 
 *   Start the [sleep]({{< github_tree >}}/samples/sleep) sample
-    which will be used as a test source for external calls via the proxy.
+    which you will use as a test source for external calls via the proxy.
```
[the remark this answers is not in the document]
vadimeisenbergibm: Note that there is a guidance in istio.io to use "you" for the steps the reader takes. So it is not the example uses, it is the reader who **will** use the sleep sample after he/she starts it.

So which one is better?

1. Start the sleep sample which you use in this example
2. Start the sleep sample which you **will** use in this example

cc @frankbu

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: `Also` is added to connect with the previous sentence:

> To simulate a legacy proxy... Moreover, [to be closer to the real life situation], you will address the proxy's pod by its IP address

Please advise.

vadimeisenbergibm: > Use the imperative present tense instead of the future tense

The imperative could confuse the readers - they can ask themselves: "What should I do to address the proxy's pod by its IP address?". 

The meaning here is "in the following instructions, when you will address the pod, you will address it by its IP address, as we will instruct you".

Please advise.

vadimeisenbergibm: Alternatively, the meaning is: in the following steps you will address the pod by its IP, do not be surprised - you will do it to simulate a real life situation better.

rcaballeromx:
> Passive voice cases:
> 
> * "access to HTTPS proxies is performed by the HTTP CONNECT method" The access is performed by clients, which should be obvious in networking - clients send requests to servers. In this case, clients send requests to servers via an HTTPS proxy. The clients use HTTPS CONNECT method to communicate with the HTTPS proxy (this is the protocol).
> * "HTTP CONNECT method is used to communicate with HTTPS proxies" - the same as above.
> * "Istio sidecars will not be injected" since the sidecars are injected automatically, it does not matter which component injects them

See [this comment](https://github.com/istio/istio.io/pull/2960#discussion_r241522911) for the fix to the first two cases.
The third case is a bit more tricky. Although the sidecar injection is performed automatically, there is value in letting the reader know what component performs said action. Readers might still find useful to learn what component is performing that task in case they wish to stop the automatic injection, learn more about it, or configure special behaviors, if that is possible. Just because the action happens automatically, does not mean that omitting the actor is necessarily a good idea.

Now to the specific case: Passive voice isn't the biggest issue, grammatically speaking, that would be the verb congruence issue caused by the future tense. 
Can I suggest using instead: "Istio doesn't inject the sidecars when ..."

content/docs/examples/advanced-gateways/http-proxy/index.md
rcaballeromx: Let me address your concerns in order:
1. The use of `Also`, `Moreover`, and similar sentence connecting words:
As we learn to write, we are taught that the sentences in a paragraph must be linked using these words. What most of us forget, is that we are also taught to do so sparingly and that not all lead words are the same. Readers connect sentences in a paragraph with an implicit `AND`. My advice to you would be to only use leading words when the connection differs from `AND`. For example, to connect two sentences in a paragraph with `BUT` I would use "However, ..."
2. The use of imperative:
I agree fully, readers will ask  "How do I address the proxy's pod by its IP address?". If you know this, my advise would be to link to the instructions on how to do that. The imperative will help you identify those questions and then ask: Will my audience know the answer or not? If not, link to the answer, define it briefly or link to a definition. 
3. The future meaning conflicts with the use of present tense: 
I understand that from your perspective, you are describing the actions readers will take and, thus, feel the need to use the future tense in your text. However, I would advise you to switch from a descriptive approach, writing a descriptions of the future actions to take, to a prescriptive approach, writing the actions readers need to take at that moment. What action do readers need to take at the beginning of your document: Keep reading! Therefore, I suggest:
Keep reading to learn how to address the pod by its IP and how this approach allows you to simulate a real life situation better.

Hopefully, I've been clear enough and answered your questions. Let me know if further discussion is needed.

vadimeisenbergibm:
@rcaballeromx We can say "Istio doesn't inject the sidecars...", but it will be a bit imprecise. From https://istio.io/docs/setup/kubernetes/sidecar-injection/#automatic-sidecar-injection (also passive voice)

> Sidecars can be automatically added to applicable Kubernetes pods using a mutating webhook admission controller. 

To explain what it means: the Istio users, as part of Istio installation, create the "mutating webhook admission controller" for Istio sidecars. Then, when any new pod is created, the "mutating webhook admission controller" adds the Istio sidecars to the created pod.

content/docs/examples/advanced-gateways/http-proxy/index.md
vadimeisenbergibm: Note that addressing the pod by IP is done as part of the following steps. The sentence means: when you will perform the steps below, note that you will address the pod by IP (as part of the steps, not for the sake of addressing the pod by IP).

content/docs/examples/advanced-gateways/http-proxy/index.md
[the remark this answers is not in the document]
rcaballeromx: Best is:

    To have test source for external calls via the proxy, start the sleep sample.

Before the reader starts the sample, they should have read a sentence similar to:

    This example uses the sleep sample as a test source for external calls via the proxy.

That way you let readers know how the sleep sample is used throughout and you don't need to mention it on each step.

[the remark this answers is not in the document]
vadimeisenbergibm: Applied in https://github.com/istio/istio.io/pull/2973/commits/a359411b2acd1ecfd00b59f5a403fce037d25e8f

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -201,11 +205,12 @@ In the following section you are going to configure traffic from Istio-enabled p
 ## Understanding what happened
 
 In this example you configured access to an external legacy HTTPS proxy from the Istio service mesh. To simulate a
-remote HTTPS proxy, you deployed an HTTPS proxy in a separate namespace, without Istio automatic sidecar injection. In addition, for simulation of a remote proxy, you addressed the HTTPS proxy by its IP address and not by a Kubernetes
-service.
+remote HTTPS proxy, you deployed an HTTPS proxy in a separate namespace, without Istio automatic sidecar injection.
+In addition, for simulation of a remote proxy, you addressed the HTTPS proxy by its IP address and not by the domain
+name of a Kubernetes service.
```
rcaballeromx: I love the fact that you are summarizing what the readers did on the example. I would advise you to do so as a numbered list instead:

```suggestion
In this example, you took the following steps to configure access to an external legacy HTTPS proxy from an Istio service mesh:

1. Deployed an HTTPS proxy in a separate namespace without Istio automatic sidecar injection to simulate a remote HTTPS proxy.
2. Addressed the HTTPS proxy by its IP address and not by the domain name of a Kubernetes service to maintain the simulation of a remote proxy.
```
You would need to continue with the content of the next paragraph. The numbered steps provide additional clarity to readers. If you can, try to be more high-level with this overarching flow. You already covered the nit and grit of a lot of it above.

rcaballeromx:
> @rcaballeromx We can say "Istio doesn't inject the sidecars...", but it will be a bit imprecise. From https://istio.io/docs/setup/kubernetes/sidecar-injection/#automatic-sidecar-injection (also passive voice)
> 
> > Sidecars can be automatically added to applicable Kubernetes pods using a mutating webhook admission controller.
> 
> To explain what it means: the Istio users, as part of Istio installation, create the "mutating webhook admission controller" for Istio sidecars. Then, when any new pod is created, the "mutating webhook admission controller" adds the Istio sidecars to the created pod.

I love precision! Can we say:
The [mutating webhook admission controller](https://istio.io/docs/setup/kubernetes/sidecar-injection/#automatic-sidecar-injection) doesn't inject the sidecars ...

I added the link to make sure that we provide readers that do not know what the mutating webhook admission controller is a way to learn more.

vadimeisenbergibm:
@rcaballeromx @frankbu let's handle the remaining case of the passive voice, which happens a lot on istio.io, namely the automatic sidecar injection.

Background: automatic sidecar injection in Istio is implemented by a mutating webhook admission controller. The mechanism itself is not important for the user, in my opinion. The users can learn this mechanism once, if needed, and forget about it later. The important things for the users are:
1. They can inject Istio sidecars either manually or they can enable automatic sidecar injection per namespace
2. To enable automatic sidecar injection, they have to label the namespace

Apart from that, it does not matter how the automatic sidecar injection is implemented. It could be by some magic, from the user's point of view. Mentioning  mutating webhook admission controller every time automatic sidecar injection is mentioned is needlessly overwhelming users with technical details irrelevant for the current task. Istio is complicated enough, we should strive to simplify the documentation as much as possible, in particular by omitting unnecessary technical details.

The following sentence could appear multiple times on istio.io. Let's consider three alternatives to express it:

1. Do not label the namespace so Istio automatic sidecar injection will not be performed on the pods in the namespace.
2. Do not label the namespace so Istio will not inject sidecars in the pods in the namespace.
3. Do not label the namespace so the mutating webhook admission controller will not inject Istio sidecars in the pods in the namespace.

The sentence 3 provides an irrelevant technical detail - mutating webhook admission controller. The sentence 2 is imprecise (it is not an Istio component that injects the sidecars). The sentence 1 uses passive voice.

What do you think?

content/docs/examples/advanced-gateways/http-proxy/index.md
```diff
@@ -34,24 +34,25 @@ proxy is different from configuring traffic to external HTTP and HTTPS services.
...
+To simulate a legacy proxy and only for this example, you deploy an HTTPS proxy inside your cluster.
+Also, to simulate a more realistic proxy that is running outside of your cluster, you will address the proxy's pod
+by its IP address and not by the domain name of a Kubernetes service.
+This example uses [Squid](http://www.squid-cache.org) but you can use any HTTPS proxy that supports HTTP Connect.
 
-1.  Create a namespace for the HTTPS proxy. Note that since you do not label it for Istio automatic sidecar injection,
-    Istio will not control traffic in this namespace.
+1.  Create a namespace for the HTTPS proxy. Since the automatic sidecar injection is disabled in Istio by default,
+    Istio sidecars will not be injected and Istio will not control traffic in this namespace. You need this behavior to
+    simulate the proxy outside of the cluster.
```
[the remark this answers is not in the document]
vadimeisenbergibm: Addressed in https://github.com/istio/istio.io/pull/2973/commits/29f2e1c8c1629fdc3d3e3b983c6e3b6c72273b04

rcaballeromx:
> @rcaballeromx @frankbu let's handle the remaining case of the passive voice, which happens a lot on istio.io, namely the automatic sidecar injection.
> 
> Background: automatic sidecar injection in Istio is implemented by a mutating webhook admission controller. The mechanism itself is not important for the user, in my opinion. The users can learn this mechanism once, if needed, and forget about it later. The important things for the users are:
> 
> 1. They can inject Istio sidecars either manually or they can enable automatic sidecar injection per namespace
> 2. To enable automatic sidecar injection, they have to label the namespace
> 
> Apart from that, it does not matter how the automatic sidecar injection is implemented. It could be by some magic, from the user's point of view. Mentioning mutating webhook admission controller every time automatic sidecar injection is mentioned is needlessly overwhelming users with technical details irrelevant for the current task. Istio is complicated enough, we should strive to simplify the documentation as much as possible, in particular by omitting unnecessary technical details.

I agree and let me clarify: my suggestion was not to mention the mutating webhook admission controller every single time, just on the one instance in this one document. 

> 
> The following sentence could appear multiple times on istio.io. Let's consider three alternatives to express it:
> 
> 1. Do not label the namespace so Istio automatic sidecar injection will not be performed on the pods in the namespace.
> 2. Do not label the namespace so Istio will not inject sidecars in the pods in the namespace.
> 3. Do not label the namespace so the mutating webhook admission controller will not inject Istio sidecars in the pods in the namespace.

IMO, none of the versions you proposed should be used because of a verb congruence issue caused by the future tense. I propose two alternatives:

a. To stop the automatic sidecar injection in the pods of a namespace, do not label the namespace with ....

b. To stop the mutating webhook admission controller from injecting the Istio sidecars in the pods of a namespace, do not label the namespace with ...

Version a has the least technical detail and version b the most. Both of my versions and all three of yours are missing one important piece of information for the reader: What is the label in question?
To me, that is more important to include than the bit about the admission controller.

> 
> The sentence 3 provides an irrelevant technical detail - mutating webhook admission controller. The sentence 2 is imprecise (it is not an Istio component that injects the sidecars). The sentence 1 uses passive voice.
> 
> What do you think?

The level of technical detail needed in the content is different for each document based on the target audience, the context of the document, and purpose of the document. Because of that, I would advise we focus on the specific instance. 

With that in mind, I'm leaning towards a. as long as we include the label. I also wanted to suggest we change the sentence to positive instead of negative. Although that might involve changing the text around it. Within the context though, I think the negative version works best since it is intended as a warning.

vadimeisenbergibm:
@rcaballeromx In this concrete example, the meaning is as follows:
Step1: You create a new namespace.
Step2: Since you did not label the namespace in Step1 for automatic sidecar injection, no automatic sidecar injection will be applied from now on.

How this should be expressed in present tense and in one step? Note the time relationships here: 
1. You create a namespace. You do not stop automatic sidecar injection in the namespace, since automatic sidecar injection was never enabled on that namespace
2. Since you did not label the namespace you created, from now on automatic sidecar injection is disabled.

The label here is less important, since the reader should not use it. The sentence basically should say: just create a namespace and don't do anything else,  this is enough to make sure that automatic sidecar injection is disabled from now on.

rcaballeromx:
> @rcaballeromx In this concrete example, the meaning is as follows:
> Step1: You create a new namespace.
You create a new namespace without labeling for automatic sidecar injection in Step 1. That detail is important to mention on Step 1 for readers to know that when they create the namespace, not after.
> Step2: Since you did not label the namespace in Step1 for automatic sidecar injection, no automatic sidecar injection will be applied from now on.
Here you don't need to mention the label again, but I remain unsure as a reader regarding what stops performing the injection.
That might be a technical detail you might not want to add. 
> 
> How this should be expressed in present tense and in one step? Note the time relationships here:
> 
Let me give it a try:
Create the namespace without labeling it for sidecar injection. Without the label, sidecar injection is disabled from now on.
The text conveys more clearly the action the reader needs to take and the consequence of that action. 
> 1. You create a namespace. You do not stop automatic sidecar injection in the namespace, since automatic sidecar injection was never enabled on that namespace
> 2. Since you did not label the namespace you created, from now on automatic sidecar injection is disabled.
> 
> The label here is less important, since the reader should not use it. The sentence basically should say: just create a namespace and don't do anything else, this is enough to make sure that automatic sidecar injection is disabled from now on.
I think my suggestion above conveys that meaning clearly. Do you agree?

frankbu:
@rcaballeromx 

`Create the namespace without labeling it for sidecar injection. Without the label, sidecar injection is disabled from now on.`

"from now on" makes no sense, since it was never on. 

Maybe something like this: `Create the namespace without labeling it for sidecar injection. Without the label, sidecar injection is disabled in the new namespace.`

Btw, I think this discussion is going on too long. Is it really that important? We need to make sure all of these changes get in before 1.1 is released.

vadimeisenbergibm:
@frankbu Done in https://github.com/istio/istio.io/pull/2973/commits/c9636e3298aa04efbfb8b99f5433f5a4b0545f1f

vadimeisenbergibm:
Thank you @rcaballeromx @frankbu for your patience. I am glad we had this discussion and agreed on the wording.

PR #2980, opened 2018-12-13
Add a review process to our contribution guide.

rcaballeromx:
To avoid PRs from being merged before the content has been reviewed, contributors
must add a comment with the appropriate command beforehand. I've added this
process. Additionally, I cleaned up the process on how to contribute new
documents.

### DIFF
--- a/content/about/contribute/github/index.md
+++ b/content/about/contribute/github/index.md
@@ -83,7 +83,7 @@ content in detail, follow these steps:
 
     {{< text markdown >}}
     Use present tense to avoid verb congruence issues and
-    make the text easier to understand:
+    to make the text easier to understand:
 
     ```suggestion
 
@@ -98,7 +98,7 @@ content in detail, follow these steps:
     {{< warning_icon >}} If you do not publish your review, the PR owner and
     the community cannot see your comments.
 
-1.  Once you publish your review, leave a comment with the words:
+1.  Once you publish your review, leave a comment with the text:
     `/hold cancel`. That command unblocks the PR from being merged.
 
 ## Previewing your work
--- a/content/about/contribute/github/index.md
+++ b/content/about/contribute/github/index.md
@@ -82,7 +82,8 @@ content in detail, follow these steps:
     example:
 
     {{< text markdown >}}
-    Use present tense to avoid verb congruence issues:
+    Use present tense to avoid verb congruence issues and
+    make the text easier to understand:
 
     ```suggestion
 
@@ -92,12 +93,12 @@ content in detail, follow these steps:
     {{< /text >}}
 
 1.  Publish your review to share your comments and suggestions with us and the
-    PR owner.
+    PR owner. Request changes as the review warrants.
 
     {{< warning_icon >}} If you do not publish your review, the PR owner and
     the community cannot see your comments.
 
-1.  Once your comments have been addressed, leave a comment with the words:
+1.  Once you publish your review, leave a comment with the words:
     `/hold cancel`. That command unblocks the PR from being merged.
 
 ## Previewing your work
--- a/content/about/contribute/github/index.md
+++ b/content/about/contribute/github/index.md
@@ -52,11 +52,11 @@ To add content you must create a fork of the repository and a PR from
 your fork to the docs main repository. The following steps describe the
 process:
 
-1.  Click the button below to visit the GitHub repository.
+<a class="btn btn-istio"
+href="https://github.com/istio/istio.io/">Browse this site's source
+code</a>
 
-    <a class="btn btn-istio"
-    href="https://github.com/istio/istio.io/">Browse this site's source
-    code</a>
+1.  Click the button above to visit the GitHub repository.
 
 1.  Click the **Fork** button in the upper-right corner of the screen to
     create a copy of our repository in your GitHub account.
--- a/content/about/contribute/github/index.md
+++ b/content/about/contribute/github/index.md
@@ -28,29 +28,77 @@ The documentation is published under the [Apache
 
 ## How to contribute
 
-There are two ways you can contribute to the Istio documentation:
+There are three ways you can contribute to the Istio documentation:
 
 * If you want to edit an existing page, you can open up the page in your
   browser and select the **Edit This Page on GitHub** option from the gear menu
   at the top right of each page. This takes you to GitHub to edit and
   submit the changes.
 
-* If you want to work on the site in general, you must create a fork of the
-  repository. Click the button below to visit the GitHub repository. Then, you
-  must click the **Fork** button in the upper-right corner of the screen to
-  create a copy of our repository in your GitHub account. Create a clone of
-  your fork and make any changes you want. When you are ready to send those
-  changes to us, push the changes to your fork, go to the index page for your
-  fork, and click **New Pull Request** to let us know about it.
+* If you want to work on the site in general, follow the steps in our
+  [How to add content section](#add).
 
-<a class="btn btn-istio"
-href="https://github.com/istio/istio.io/">Browse this site's source
-code</a>
+* If you want to review an existing pull request (PR), follow the steps in our
+  [How to review content section](#review)
 
 Once your changes are merged, they show up immediately on
 `preliminary.istio.io`. However, the changes only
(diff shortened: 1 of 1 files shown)


content/about/contribute/github/index.md
```diff
@@ -28,30 +28,78 @@ The documentation is published under the [Apache
...
+
+    ```suggestion
+
+    Pilot maintains an abstract model of the mesh.
+
+    ```
+    {{< /text >}}
+
+1.  Publish your review to share your comments and suggestions with us and the
+    PR owner.
```
frankbu: I think that the `/hold cancel` should happen now. If you requested changes in your review, that will prevent the PR from merging.

rcaballeromx: Will it? I've seen examples where it hasn't.

rcaballeromx:
@geeknoid The spell checker is complaining about the values in the HTML tags. Is there something I need to do for it to ignore it? I know I can add it to the exceptions file but that means we need to add exceptions for every id text used in html.

content/about/contribute/github/index.md
```diff
@@ -28,30 +28,78 @@ The documentation is published under the [Apache
...
 
+### How to add content {#add}
+
+To add content you must create a fork of the repository and a PR from
+your fork to the docs main repository. The following steps describe the
+process:
+
+1.  Click the button below to visit the GitHub repository.
+
+    <a class="btn btn-istio"
```
geeknoid: This is causing a CI failure. I think the markdown process is not happy with a <a> block starting indented like that. Maybe you would just put the <a> at the end of the previous paragraph, but precede it with a <br> or two. That will likely give you the rendering you want.

rcaballeromx: @geeknoid Thank you!

content/about/contribute/github/index.md
```diff
@@ -28,30 +28,78 @@ The documentation is published under the [Apache
...
+    fork.
+1.  Go to the index page for your fork, and click **New Pull Request** to let
+    us know about it.
+
+### How to review content {#review}
+
+If your review is small, simply comment on the PR directly. If you review the
+content in detail, follow these steps:
+
+1.  Leave a comment on the PR with the text `/hold`. This command prevents the
```
sdake: this step is overkill imo :)

content/about/contribute/github/index.md
sdake: yes it will prevent changes from merging if you click the "request changes button".  We generally don't use this button though, as the red turns away contributors.  Instead, we leave comments and reviewers don't merge the PR until the comments are resolved.

This solves a bit of a social problem (the negativity of the "request changes in Red" vs just guiding the committer towards a successful merge.

I suspect /hold and /hold cancel do the same thing.  As a reviewer, when I see a /hold, I take a good hard look at every review comment and change and am more thorough in the review.  Since /holds are so infrequent, it is warranted to be extra cautious.

I'd recommend removing hold from the workflow for this reason.

Cheers
-steve

rcaballeromx: @sdake Sorry but I am not sure I followed. In your opinion: should the /hold be removed as soon as the **request changes button** is selected by the reviewer when they publish the review or after the all the comments and changes have been addressed?

sdake: I think pressing the "request changes button" is counter to sending a positive message to contributors.  It shows up as a big red X.  Other reviewers will not approve a review until all relevant comments have been addressed.  I would advocate a workflow that doesn't expect reviewers to use hold/request changes features of github - but instead put a little extra work on reviewers to ensure all relevant issues are resolved.  Reviewers should be doing this work anyway, especially folks in the OWNERS file.

content/about/contribute/github/index.md
sdake: That is a fair complaint.  Typically how we handle this in the code base, is if a patch merges prematurely, the onus is on the committer to fix the problems found post merge.  If the patch problems are egregious, the onus is on the reviewer to revert the commit and trigger the commiter to resubmit their work.

Perhaps we should have different set of workflows for docs vs code, as code reviews rarely merge ahead of a very solid review (and there are 30+ checks).  Docs are different - less checks, and more human work needed during the review process.

Cheers
-steve

content/about/contribute/github/index.md
rcaballeromx: @sdake I agree with your sentiment regarding sending a positive message to contributors. However, we need a process that scales well for both contributors and reviewers. I do like the use `/hold` over the request changes because it does not show a big red `X` beside the reviewers and because it ensures that no one inadvertently merges a contribution that is not approved by the reviewer. Historically, that "little extra work" on part of the reviewers has not happened and as the number of contributors grow, it becomes a lot harder for owners to keep track of both reviewers and contributors.

content/about/contribute/github/index.md
rcaballeromx: I agree. I don't think the right approach to revert someone's contribution to our content. I much rather iterate on it because, unlike in code, there are several ways of writing the same idea. My proposal here is to add provide a transparent way to ensure feedback is applied to reduce the quality-debt of the content. So far, only @vadimeisenbergibm has followed the process you describe @sdake. In my experience, it is hard enough to get the initial contribution to the docs. After most contributions, authors never revisit the content unless there was a code change that requires them to and, even then, that doesn't always happen. With this process, we can ensure, at the very least, the technical accuracy and basic clarity of content.

content/about/contribute/github/index.md
rcaballeromx: @geeknoid Unfortunately, the fix doesn't work since the list still breaks if I do not indent the button. This is definitely a bug for the linter, where should I report it? 
I fixed it by placing the button before the list. It's not ideal but it works as a workaround.

rcaballeromx:
I've rebased and fixed the linting issue caused by the the button. I've also answered the questions that were brought up regarding the process. Only content where a formal review is done will be subject to this process, not all content updates fall under this category.

content/about/contribute/github/index.md
frankbu: I think that whether or not "request changes" looks like it's sending a positive message, it is the standard github process, so we should be following it. I think that `/hold` is a good way to temporarily block automatic merging until a review is complete. If that review is asking for changes, then marking the review as "request changes" is the proper way to control the merge at that point. Of course, if either the committer or reviewer is not responding quickly enough, and the PR is basically done and important enough, the owners/administrators can always go ahead and override and merge the PR at their discretion. Additional changes/fixes, if necessary, can then be done in a followup PR. This is one of the nice things about the way we initially publish to preliminary.istio.io, so it's not really live until the next release goes out.

rcaballeromx: @frankbu You then agree that the `hold cancel` comments should happen once the review is published with **request changes** but before the changes are addressed?
I agree that the Owners can always override and merge at their discretion, that remains unchanged. Additional changes in subsequent PRs are also a possibility but keep in mind that we cannot force contributors to make them. 
@frankbu @sdake @geeknoid If you all agree that the `hold cancel` should happen right after the review is published, I'll change the steps to reflect it. What do you think?

geeknoid:

Built with commit 06cdfbdefbecd2d6754ed335c2d1a73135545b00

https://app.netlify.com/sites/preliminary-istio/deploys/5c3f7975550ed00008f6962e

rcaballeromx:
The text now makes it clear that the comment with `/hold cancel` should be done right after the review is published.
@geeknoid @sdake @frankbu PTAL

content/about/contribute/github/index.md
```diff
@@ -28,30 +28,79 @@ The documentation is published under the [Apache
...
+    ```
+    {{< /text >}}
+
+1.  Publish your review to share your comments and suggestions with us and the
+    PR owner. Request changes as the review warrants.
+
+    {{< warning_icon >}} If you do not publish your review, the PR owner and
+    the community cannot see your comments.
+
+1.  Once you publish your review, leave a comment with the words:
```
frankbu: nit: change "with the words:" to "with the text" for consistency with step 1.

otherwise, looks good.

rcaballeromx: Done.

PR #2803, opened 2018-10-26
Add new blog entry to create a custom gateway manually

jsenon:
I propose to add a new blog entry for "How to create you own custom gateway?"

PR #2996, opened 2018-12-19
Blog custom gateway

jsenon:
I propose to add a new blog entry for "How to create you own custom gateway?"
Replace old PR as mention by @geeknoid

### DIFF
--- a/content/blog/2018/custom-gateway/index.md
+++ b/content/blog/2018/custom-gateway/index.md
@@ -22,7 +22,7 @@ Have cert-manager installed
 
 Cert-manager needs to be installed, if it's has been install with helm you can check installation has bellow:
 
-{{< text sh >}}
+{{< text bash >}}
 $ helm ls
 NAME         	REVISION	UPDATED                 	STATUS  	CHART                    	APP VERSION 	NAMESPACE
 istio        	1       	Thu Oct 11 13:34:24 2018	DEPLOYED	istio-1.0.2              	1.0.2       	istio-system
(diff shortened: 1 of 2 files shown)


geeknoid:
@jsenon Julien, can you follow up on Frank and Vadim's comments from the other PR?

@frankbu Why don't you feel this is appropriate as a task?

frankbu:
@geeknoid, there are too many things that we are demonstrating in the category of ingress/egress, so we decided to just stick to a couple of basic tasks and put the rest of the things in the edge traffic examples folder: https://preliminary.istio.io/docs/examples/advanced-gateways/

jsenon:
@geeknoid I've taken into account Frank and Vadim comment except Helm remark because I think the most important part of this blog entry is to explain how, all components are linked together by an example, and not with an automated installation

geeknoid:
So Frank:

If not a task, do you think this doc should be a blog post or be moved to docs/examples/advanced-gateway?

And Julien thinks he's addressed your and Vadim's feedback from #2803. So what's missing?

frankbu:
I think it can be moved to a blog as is. Example would be fine with me too, but I think the instructions need a little more detail to be consistent with our other examples. Our examples spell out everything you need to do to run it. This one is missing steps or at least notes. A couple of examples are: In step 9, what VirtualService? To run the curl command in step 10 you need to have DNS set up.

@vadimeisenbergibm WDYT?

frankbu:
My vote is for blog entry, which is what Julien initially proposed.

vadimeisenbergibm:
I am OK with blog entry.

PR #3012, opened 2018-12-30
Fix the typo on the secure ingress page

SeriyBg:
Fixed in the documentation command
- Typo, the `jsonpath` contain extra dot char: `.items[0]..metadata.name` instead of `.items[0].metadata.name`
- The jsonpath without a weapping in quote chars won't work on all the systems and could lead to some errors of type: `no matches found: jsonpath={.items[0].metadata.name}`


SeriyBg:
@sdake I'm running it on MacOS with zsh as a shell and it fails with the error that I've posted in PR

PR #3013, opened 2019-01-01
Update index.md

buptliuwei:
To better distinguish between the two ways to call external services from an Istio mesh, we should remove the rules about `ServiceEntry`.

### DIFF
--- a/content/docs/tasks/traffic-management/egress/index.md
+++ b/content/docs/tasks/traffic-management/egress/index.md
@@ -217,7 +217,7 @@ $ helm template install/kubernetes/helm/istio <the flags you used to install Ist
 Note that you should use the same Helm command you used [to install Istio](/docs/setup/kubernetes/helm-install),
 in particular, the same value of the `--namespace` flag. In addition to the flags you used to install Istio, add `--set global.proxy.includeIPRanges="10.0.0.1/24" -x templates/sidecar-injector-configmap.yaml`.
 
-Redeploy the `sleep` application as described in the [Before you begin](/docs/tasks/traffic-management/egress/#before-you-begin) section.
+**You should remove the previously deployed ServiceEntry and VirtualService**.Redeploy the `sleep` application as described in the [Before you begin](/docs/tasks/traffic-management/egress/#before-you-begin) section.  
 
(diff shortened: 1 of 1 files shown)


buptliuwei:
I signed it!

content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -217,7 +217,7 @@ $ helm template install/kubernetes/helm/istio <the flags you used to install Ist
 Note that you should use the same Helm command you used [to install Istio](/docs/setup/kubernetes/helm-install),
 in particular, the same value of the `--namespace` flag. In addition to the flags you used to install Istio, add `--set global.proxy.includeIPRanges="10.0.0.1/24" -x templates/sidecar-injector-configmap.yaml`.
 
-Redeploy the `sleep` application as described in the [Before you begin](/docs/tasks/traffic-management/egress/#before-you-begin) section.
+**You should remove the previously deployed ServiceEntry and VirtualService**.Redeploy the `sleep` application as described in the [Before you begin](/docs/tasks/traffic-management/egress/#before-you-begin) section.
```
frankbu: I would suggest putting this as a warning after the "Redeploy the ..." sentence, something like this:

Redeploy the `sleep` application ...

{{< warning_icon >}}  Make sure to remove the previously deployed `ServiceEntry` and `VirtualService`.

buptliuwei: thanks, it looks more clear :)

buptliuwei: Maybe we can also express it in the form of a quote, just like other tips.

Redeploy the sleep application ...
> Make sure to remove the previously deployed `ServiceEntry` and `VirtualService`.

frankbu: Yes, that would be fine too. It's up to you, whichever way you prefer, but please add a blank line before the warning, so it displays a little nicer, and remove the space character at the end of the line that is causing the circleci failure. Thanks.

buptliuwei:
@geeknoid @frankbu thanks

PR #3026, opened 2019-01-03
Fix GKE cluster create command (platform setup)

m-okeefe:
Fixes cluster create command in: https://preliminary.istio.io/docs/setup/kubernetes/platform-setup/gke/ 

Bug: without a trailing slash after `num-nodes`, any other arguments after (zone, project) are ignored.  (And thus defaults to a different zone, eg. `us-central1-b`, not the zone specified in the flag.) 

Solution: 
- Add trailing slash after `num-nodes` 
- Also added `cluster-version=latest` for best-practices reasons.


ahmetb:
any ideas when this goes live? isn't there auto-publish?

https://istio.io/docs/setup/kubernetes/platform-setup/gke/ still shows the old content even after 1+ month.

PR #3053, opened 2019-01-08
Update the minimal install Helm flags

ymesika:
Disable the useMCP as Galley is not installed in a minimal install and Pilot it throwing errors.


geeknoid:
Dupe of #3056

PR #3056, opened 2019-01-08
Add useMCP=false for minimal install

clyang82:
useMCP is true by default in `release-1.1`. we should disable it for minimal install (only pilot).

Signed-off-by: clyang82 <clyang@cn.ibm.com>

### DIFF
--- a/content/docs/setup/kubernetes/minimal-install/index.md
+++ b/content/docs/setup/kubernetes/minimal-install/index.md
@@ -41,6 +41,7 @@ via `kubectl apply`, and wait a few seconds for the CRDs to be committed in the 
       --set mixer.telemetry.enabled=false \
       --set prometheus.enabled=false \
       --set global.proxy.envoyStatsd.enabled=false \
+      --set global.useMCP=false \
       --set pilot.sidecar=false > $HOME/istio-minimal.yaml
     {{< /text >}}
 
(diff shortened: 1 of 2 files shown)


geeknoid:
Yeah, given that Galley is on by default, the minimal install of Istio should include Galley.

There's a flag somewhere to control whether to use Galley or not in the 1.1 timeframe (since Galley is only in Beta mode). I think that should be sufficient control.

clyang82:
Yes. Galley will be the default configuration mechanism for Pilot in Istio 1.1+. for 1.1 timeframe, use Galley and Pilot for minimal install is not working.

geeknoid:
@ozevren 

The minimal install of Istio for 1.1 should include Galley, and Pilot should be getting its config through MCP even in the minimal install.

linsun:
so what is the conclusion on galley for minimal profile?  I don't think it includes galley by default.

if we can't get galley running independently by itself (istio/istio#8736), I'd prefer to leave it out for minimal profile for 1.1.

@geeknoid @clyang82 @ozevren @ayj @rshriram

geeknoid:
For 1.1, we want Galley to be enabled by default in all configurations, including in the minimal install.

Because Galley is a Beta component, we should have an option to turn Galley off for exceptional cases. But the mainline deployments of Istio should always include Galley.

So I'd like to close this PR as a result. We might have some bugs to fix, that's fine. But our instructions should never tell users to not install Galley.

PR #2959, opened 2018-12-07
Revised mesh expansion instructions

LisaFC:
Updated from #2089

PR #3057, opened 2019-01-08
Updated mesh expansion guide

LisaFC:
Supercedes #2959 - same change, minus many many old changes in the same branch.

### DIFF
--- a/content/docs/setup/kubernetes/mesh-expansion/index.md
+++ b/content/docs/setup/kubernetes/mesh-expansion/index.md
@@ -166,7 +166,7 @@ After setup, the machine can access services running in the Kubernetes cluster
 or on other mesh expansion machines.
 
 The following example shows accessing a service running in the Kubernetes cluster from a mesh expansion VM using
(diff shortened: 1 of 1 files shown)


costinm:
Hi Lisa, would it be possible to cherry-pick this into 1.0 branch as well ?

LisaFC:
Done so in #3104

incfly:
After walking through this updated instructions with latest istio daily build for 1.1, I run into issues.

So this PR removes the `istioctl register <service-name> endpints protoco:port` using Kubenetetes selector-less to generate DNS mappings. However the default installment does not generate any DNS mappings based on ServiceEntry (multi cluster does, when we install a CoreDNS plugin pod).

This means we don't have a service VIP for services on VM, and I hit `404`.

Running `istioctl register will fix that`.

Do I miss something? What's our intended solution regarding VIP for `ServiceEntry` mesh expansion services? @rshriram @costinm 

Without that, customers can't get it work.

PR #3060, opened 2019-01-08
Organize the setup/kubernetes sidebar

jlbutler:
The docs Setup->Kubernetes sidebar is a bit muddy, with install items, maintenance items and platform quickstart guides all scrambled.

This commit resets page weights to organize the sidebar a bit, with common articles grouped together.

I had considered putting a "Quickstart Guides" subcategory in place, but I think it's better to leave them as-is. Open to either way though.

### DIFF
--- a/content/docs/setup/kubernetes/upgrading-istio/index.md
+++ b/content/docs/setup/kubernetes/upgrading-istio/index.md
@@ -1,7 +1,7 @@
 ---
 title: Upgrading Istio
 description: Demonstrates how to upgrade the Istio control plane and data plane independently.
-weight: 70
+weight: 37
 keywords: [kubernetes,upgrading]
 ---
 
--- a/content/docs/setup/kubernetes/advanced-install/index.md
+++ b/content/docs/setup/kubernetes/advanced-install/index.md
@@ -1,7 +1,7 @@
 ---
 title: Advanced Install Options
 description: Instructions for customizing the Istio installation.
-weight: 20
+weight: 35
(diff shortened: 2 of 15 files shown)


jlbutler:
/assign @sdake

ping @geeknoid (related to other work recently reviewed)

jlbutler:
Good question, I kept it there just because it was lowest. Maybe it should move after install items?

geeknoid:
Yes, I think so.

jlbutler:
Moved 'Platform Setup' above the Install items. Seems like a good place for it, let me know what you think. Thanks!

content/docs/setup/kubernetes/upgrading-istio/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Upgrading Istio
 description: Demonstrates how to upgrade the Istio control plane and data plane independently.
-weight: 70
+weight: 90
```
sdake: upgrading should come right after the installation.  At the moment it appears like n information island.

Cheers
-steve

jlbutler: Yeah, I intended for them to go after the series of "quick starts" as they are "installs" technically... but, you're right. I dropped it between the draft 'Advanced Install' and 'Installing the Sidecar'. Thanks!

jlbutler:
Move upgrade notes up and rebased, PTAL thanks!

PR #3061, opened 2019-01-08
Add blog entry blog post for custom ingress gateway with cert manager

jsenon:
Replace #2996


jsenon:
@geeknoid Can you close old PR #2996 ?

content/blog/2019/custom-ingress-gateway/index.md
```diff
@@ -0,0 +1,226 @@
...
+title: Deploy a custom ingress gateway using cert-manager
+description: Describes how to deploy a custom ingress gateway using cert-manager manually.
+subtitle: Custom ingress gateway
+publishdate: 2019-01-08
+weight: 89
+keywords: [ingress,traffic-management]
+attribution: Julien Senon
+---
+
+This post provides instructions to manually create a custom ingress [gateway](/docs/reference/config/istio.networking.v1alpha3/#Gateway) with automatic provisioning of certificates based on cert-manager.
```
linsun: would this be more appropriate as a task in istio.io?  This seems something we want to keep it as a living doc.

linsun: Would be great if you can highlight why users would need to create a custom ingress gateway vs create an ingress gateway resource using the default ingress gateway provided by istio?   for example https://istio.io/docs/tasks/traffic-management/secure-ingress/

geeknoid: This started out as a blog, then we moved it to a task, and now it's back to a blog. @frankbu and @vadimeisenbergibm have been involved in the discussion.

I don't care either way, I'll be happy either way.

frankbu: As I said in the previous PR, this should not be a task. If we want it in examples, then it needs more work, but as is it's best as a blog.

content/blog/2019/custom-ingress-gateway/index.md
```diff
@@ -0,0 +1,226 @@
...
+* We will use `demo.mydemo.com` for our example,
+  it must be resolved with your DNS
+
+## Configuring the custom ingress gateway
+
+1. Check if [cert-manager](https://github.com/helm/charts/tree/master/stable/cert-manager) was installed using Helm with the following command:
+
+    {{< text bash >}}
+    $ helm ls
+    {{< /text>}}
```
frankbu: I think this is causing the circleci errors. Missing space.
```
{{< /text >}}
```

geeknoid: Yep, that would do it. I really should improve the regex in scripts/lint_site.sh to deal with this case correctly

content/blog/2019/custom-ingress-gateway/index.md
linsun: As much as I like this in a task, I do agree that it needs more work for task. :).

geeknoid:
Yep, the space was the problem.

@jsenon Please add the missing space on line 23, rebase, and we should be good to go.

Thanks.

PR #3064, opened 2019-01-09
simplify the platform setup notes in helm install

jlbutler:
Change "Installation with Helm" to point to the "Platform Setup" page rather than specifying each platform in a list. This is similar to the "Installation with Ansible" page. This simplifies adding a new platform, leaving just the one list in "Quick Start with Kubernetes" to update.


jlbutler:
/assign @sdake
ping @geeknoid

PTAL, thanks!

PR #3065, opened 2019-01-09
fix envoy reference in docs/reference/config/istio.networking.v1alpha…

emaildanwilson:
The links look like `_` was replaced w/ `<em>` somehow. 
This should make them work again but maybe there is something else this is generated from which needs to be fixed?


geeknoid:
This file is generated based on the original .proto file in the istio/api repo. I fixed the original, and regenerated the file in #3069. So I'm going to close this PR.

Thanks.

PR #3093, opened 2019-01-16
Add breadcrumbs at the top of each content page.

geeknoid:
This is an experiment to seek feedback.

Right now, main content pages have a link at the top of the page to go up one level in the doc hierarchy. This PR adds a breadcrumb display at the top as well. So some questions:

- Do you prefer to "go up one level" link or the breadcrumb line?

- Should the breadcrumb line include the top-most part of the web site (namely "Istio") or should it start at the second level of the hierarchy?

- Should the current document title be included in the breadcrumb line, or should the breadcrumb line only show the path leading to the current document?

What do you think?


frankbu:
I kind of like it the way you have it in this PR:

1. breadcrumb line
2. include top-most part
3. include current document

geeknoid:
I'm going to push this in. Let me know if you'd like to see something different...

PR #3099, opened 2019-01-16
Add the formal process to become a member of an Istio repo.

rcaballeromx:
This process is detailed as agreed upon by the Istio Technical Oversight Committee.


geeknoid:
We don't want this information on istio.io for two reasons:

- The source of truth for our governance material on github.com/istio/community. We have a pointer to this in the community page on istio.io.

- istio.io is for customers, not contributors (the exception being around doc contributions). We want to keep all contributor material in GitHub.

rcaballeromx:
@geeknoid The content now points to the info in the community repo. PTAL

PR #3102, opened 2019-01-17
fix from non-Istio services to Istio service

idouba:
request from non-Istio services to Istio service should NOT get 200 response when mtls is enabled.


geeknoid:
Please also submit this change to the master branch if it still applies.

Thanks.

frankbu:
This looks like it's already fixed in `master` and `release-1.1`.

idouba:
Yes, checked is fixed in master. @frankbu @geeknoid

PR #3104, opened 2019-01-17
Add updated meshex instructions to 1.0 branch

LisaFC:


LisaFC:
This is cherry-picked from #3057

geeknoid:
1.1 is right around the corner. Do we still need this PR?

LisaFC:
Yeah, possibly, though it does mean existing users will have to wait for 1.1 (or go to preliminary) for working meshex instructions. (also is mesh expansion going to change at all in 1.1?)

PR #3109, opened 2019-01-20
Updates to the Split Horizon EDS example

ymesika:
Updates to the doc after verifying the example with latest v1.1 build

### DIFF
--- a/content/docs/examples/multicluster/split-horizon-eds/index.md
+++ b/content/docs/examples/multicluster/split-horizon-eds/index.md
@@ -81,6 +81,7 @@ This will be used to access the `local` pilot securely using the ingress gateway
     --set security.selfSigned=false \
     --set mixer.telemetry.enabled=false \
     --set mixer.policy.enabled=false \
+    --set global.useMCP=false \
     --set global.controlPlaneSecurityEnabled=true \
     --set gateways.istio-egressgateway.enabled=false \
     --set global.meshExpansion.enabled=true \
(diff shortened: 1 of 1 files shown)


linsun:
@ymesika any particular reason why MCP needs to be disabled for this scenario?  I asked because that is not obvious and it is enabled by default.

linsun:
cc @rvennam

ymesika:
@linsun I'm not sure the instruction I have in this example fits the scenario where Pilot fetches the configuration from MCP and MCP is watching the remote k8s api. This still needs to be tested and examined and with the `useMCP=false` flag we choose the classical way of Pilot watching k8s.

I think it has no affect on `istio-remote` though so I will remove it.

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -92,7 +93,7 @@ This will be used to access the `local` pilot securely using the ingress gateway
     {{< text bash >}}
     $ kubectl create --context=$CTX_LOCAL ns istio-system
     $ kubectl create --context=$CTX_LOCAL secret generic cacerts -n istio-system --from-file=samples/certs/ca-cert.pem --from-file=samples/certs/ca-key.pem --from-file=samples/certs/root-cert.pem --from-file=samples/certs/cert-chain.pem
-    $ kubectl apply --context=$CTX_LOCAL -f install/kubernetes/helm/istio/templates/crds.yaml
+    $ helm install install/kubernetes/helm/istio-init --name istio-init
```
frankbu: Which context is `helm install` working in? This command implies some required setup that was not mentioned in the earlier instructions - e.g., tiller should be running in the local cluster

frankbu:
@linsun, `/lgtm` on doc PRs cause auto-merge so please don't `/lgtm` unless it's has been reviewed and you really want to trigger the merge.

content/docs/examples/multicluster/split-horizon-eds/index.md
ymesika: Right. How do we install CRDs without Helm? Latest releases don't have a `crds.yaml` in them anymore.

frankbu: Good question. I think we do need a non-helm way to install the CRDs.

@andraxylia Do you know what the non-helm plan is?

linsun: You could do it via a loop to install CRDs one by one.  See https://preliminary.istio.io/docs/setup/kubernetes/helm-install/#option-1-install-with-helm-via-helm-template

However using istio-init is also valid which is why it was created in the first place.

linsun:
@frankbu I did review it, I missed the good context question you raised.  @ymesika can help address your concern in a separate WI.

PR #3112, opened 2019-01-21
clarify secure ingress gateway troubleshooting regarding killing the pod to reload certificates

vadimeisenbergibm:
put the instructions to kill the pod after checking that the key/certificate are loaded. Killing the pod is required if the key/certificates were not loaded, so killing-the-pod step should be put after checking-keys-are-loaded step. There is not point to proceed with other steps if the keys were not loaded.


content/docs/tasks/traffic-management/secure-ingress/index.md
```diff
@@ -470,6 +470,13 @@ they have valid values, according to the output of the following commands:
 
     `tls.crt` and `tls.key` should exist in the directory contents.
 
+1.  If you created the secret `istio-ingressgateway-certs`, but the key and the certificate are not loaded, kill the
```
geeknoid: created the secret `istio-ingressgateway-certs` -> created the `istio-ingressgateway-certs` secret
kill -> delete

Issue #3124, opened 2019-01-23
Instruction for use of IP address ranges in IKS is incorrect


mbanikazemi:
The instructions for using `global.proxy.includeIPRanges` for allowing access to outside the mesh contain an incorrect CIDR. Services are in the 172.21.0.0/16 range on IKS.

PR #3126, opened 2019-01-23
Correcting the IP ranges used in IKS

mbanikazemi:


mbanikazemi:
@frankbu 172.20.0.0 was never used. It must be a typo. The 10.10.10.0/24 range was used very briefly very early on and was changed to 172.21.0.0/16 to allow for larger number of services. So perhaps we should remove the 10.10.10.0/24 as well. What do you think?

linsun:
This change looks right to me, when comparing to my IKS clusters... all services being used are 172.21.x

Examples:

```
$ kubectl get services
NAME           TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
kubernetes     ClusterIP   172.21.0.1     <none>        443/TCP    48d
redis-master   ClusterIP   172.21.94.10   <none>        6379/TCP   47d
redis-slave    ClusterIP   172.21.142.8   <none>        6379/TCP   47d```

linsun:
@kitch can you comment?

kitch:
LGTM. The service IP range for IKS has always been `172.21.0.0/16`.

linsun:
@kitch thanks!  should we remove the other two, e.g. `172.30.0.0/16\,10.10.10.0/24`?

frankbu:
In that case, this has never been documented correctly, which I suppose implies nobody has used it.
Nevertheless, we should probably also fix it in at least the release-1.0 branch.

linsun:
+1 on fix it in 1.0 :)

linsun:
I'd suggest leave as it is.

mbanikazemi:
Addressed issue #3124 
A new pull request for release-1.0

PR #3149, opened 2019-01-28
Deprecate Deployment Manager install, point to Istio on GKE

ostromart:
afaik deployment manager is not being maintained anymore, this PR is to point people to the Istio on GKE docs by default.


geeknoid:
If its deprecated, let's just remove the whole page. Then go to the other page and add an alias so that any bookmarks to the old page will land on the new page. See https://preliminary.istio.io/about/contribute/creating-and-editing-pages/#renaming-moving-or-deleting-pages for info on aliases.

ostromart:
Done. I did not update the link since the top level title is still correct.

geeknoid:
I mean to just delete the .md file completely. It's just not useful to have this thing that's not supported listed on the web site.

ostromart:
The .md page title is "Quick start with GKE" which we still support, just not the Deployment Manager flavor of GKE install. I don't think we have DM referenced anywhere specifically, it's just a section of "Quick start with GKE".

geeknoid:
Sorry, I was reacting to the .md file path, which includes -dm, which ends up in the URL.

So how about renaming the directory to remove the -dm? Then you can add an alias in the .md to deal with old bookmarks.

There are also DM-related images to delete in the directory.

Finally, the deleted text references DM-related content in GitHub. Do we need to delete those too?

ostromart:
Done for renaming the dir, but the images cannot be removed since they are still referenced from the Chinese page. When that page is updated we can remove the images. How's your Chinese?

geeknoid:
@rootsongjc @fleeto When you guys update the Chinese version of this page, please nuke the superfluous image files.

Thanks.

PR #3151, opened 2019-01-29
Fix Invalid Example Manifests in Authentication Reference

jacobboykin:
A few **Policy** examples include invalid **Name** fields (k8s api server rejects them as they contain uppercase letters and underscores) and invalid spec fields of **target** (**targets** is the correct field, and istio will reject examples including this)


geeknoid:
This is a generated file. Please go apply this fix to the original content, which is in a .proto file in the istio/api repo.

Thanks.

PR #3157, opened 2019-01-29
fail with 503 instead of 000 for injected-app

idouba:
As a bad destination rule is set to disable client side mTLS and receiving side is mTLS enabled. At this point, Running the curl command between sidecar injected Istio services  all requests will fail with a 503 error code as the client side is still using plain-text.


rcaballeromx:
@idouba Will the user only see the returned 503 code or will the system provide additional information?

content/docs/tasks/security/mutual-tls/index.md
```diff
@@ -125,8 +125,7 @@ You can also confirm that requests from `sleep` to `httpbin` are now failed:
```
frankbu: Can you also please change the above sentence to say `are now failing:`

rcaballeromx:
@idouba Any update?

idouba:
@rcaballeromx Sorry for getting back to you late，I will show you the details of 503 when I meet it again.  
: )

PR #3159, opened 2019-01-29
should specify the resource type:destinationrule.

idouba:
should specify the resource type:destinationrule in the delete statement.


rcaballeromx:
The updated command looks good.
/approve

PR #3162, opened 2019-01-29
fail with 503 instead of 000 for injected-app

idouba:
As a bad destination rule is set to disable client side mTLS and receiving side is mTLS enabled. At this point, Running the curl command between sidecar injected Istio services all requests will fail with a 503 error code as the client side is still using plain-text.


rcaballeromx:
Will the fail command only return 503 without any further text?

linsun:
Seems a reasonable update - @wattli - can you review?

geeknoid:
Does this fix also need to be applied to the master branch?

PR #3163, opened 2019-01-29
Update monitoring and access control for egress traffic blog post

vadimeisenbergibm:
From https://archive.istio.io/v0.8/blog/2018/egress-monitoring-access-control/


content/blog/2018/egress-monitoring-access-control/index.md
```diff
@@ -0,0 +1,614 @@
+---
+title: Monitoring and Access Policies for HTTP Egress Traffic
+description: Describes how to configure Istio for monitoring and access policies of HTTP egress traffic.
+publishdate: 2018-06-22
```
geeknoid: Please add a new front-matter field:

last_update: 2019-11-29

vadimeisenbergibm: Done by https://github.com/istio/istio.io/pull/3163/commits/4adf48b4cf57c87e17bb94ab5101ee74119ce0e9. Nice feature!

geeknoid:
@rcaballeromx @frankbu Any comments?

rcaballeromx:
I'm adding this to my queue now but I don't think I'll be able to finish the review today.
@vadimeisenbergibm Is it ok if push it back to Monday?
/hold

vadimeisenbergibm:
@rcaballeromx Sure, please take your time. The quality is the most important concern, looking forward to your comments.

content/blog/2018/egress-monitoring-access-control/index.md
```diff
@@ -0,0 +1,613 @@
...
+source pod's name and the container's name is `sleep`.
+
+## Configure monitoring and access policies
+
+Note that since you want to accomplish your tasks in a _secure way_, you must direct egress traffic through
+_egress gateway_, as described in the [Egress Gateway with TLS Origination](/docs/examples/advanced-gateways/egress-gateway-tls-origination/)
+task. The _secure way_ here means that you want to prevent malicious applications from bypassing Istio monitoring and
+policy enforcement.
+
+According to the scenario of this blog post, the organization performed the instructions in the
```
rcaballeromx: To help readers identify with your message, consider including the reader in the narrative. This technique can be effective to keep the text clear and simple.
```suggestion
According to our scenario, the organization performed the instructions in the
```

content/blog/2018/egress-monitoring-access-control/index.md
```diff
@@ -0,0 +1,613 @@
...
+[Before you begin](#before-you-begin) section. It enabled HTTP traffic to _edition.cnn.com_ and configured that traffic
```
rcaballeromx:
```suggestion
[Before you begin](#before-you-begin) section, enabled HTTP traffic to _edition.cnn.com_, and configured that traffic
```

content/blog/2018/egress-monitoring-access-control/index.md
```diff
@@ -0,0 +1,613 @@
...
+to pass through the egress gateway. The egress gateway performs TLS origination to _edition.cnn.com_, so the traffic
+leaves the mesh encrypted. Now the organization is ready to configure Istio for monitoring and access policies for
```
rcaballeromx: "Now" is probably not the best word choice. It can confuse readers thinking that you are referring to time but I understand you want to establish the situation before you move on. Consider instead:
```suggestion
leaves the mesh encrypted. At this point, the organization is ready to configure Istio to monitor and apply access policies for
```

rcaballeromx:
@vadimeisenbergibm 
I've taken a cursory look and began reviewing but I have a question: Why are you submitting this as a blog post instead of a Task or an Example?
I began leaving comments on how to soften the language and make it more conversational to fit the style of a blog but after the prerequisites the content turned into a tutorial.
IMO, this would be a better task than a blog post since I see the value of having this information updated for future releases.
@geeknoid @frankbu Your thoughts?

vadimeisenbergibm:
@rcaballeromx @geeknoid @frankbu I am never sure what is a blog post and what is an example.

The arguments in favor of a blog post:

1. It does not introduce some new concept, in this case telemetry and policies have their own tasks. It shows how to apply telemetry and policies on Egress traffic.
2. It has pretty large content, summary-like
3. It has some story: _Consider an organization that runs applications that process content from cnn.com._

But I am OK with making it an example too. Task is for small, focused items in my opinion.

rcaballeromx:
I agree, an Example is probably best for most of the content. However, you can have some of the additional context information in a blog post announcing the new Example.

frankbu:
My take on this:

* tasks - a short-and-sweet set of instructions to accomplish something with not much explanation of how things work, just explain what's being done
* examples - a more complete set of instructions for how to do something more involved or complete. Initially, we started creating examples that were just pointers to a set of tasks that could be run to accomplish some larger goal.
* blogs - explains something clearly, with or without the exact instructions for how to follow along.

Over time, things have gotten a little fuzzy, because the articles are a mix. Alos sometimes we decided that what was initially a blog, is a really important piece of documentation, so we moved it. This [early post](https://istio.io/blog/2017/0.1-canary/) of mine is an example of one that we left as a blog but ended up pointing to it from various other places in the docs. It does conform to the original idea of a blog. It explains concepts and has an example, but not one with explicit steps in the article.

To summarize, I think we have a bunch articles that may or may not be in the right place. We really should revisit the criteria, esp for examples. For this article, I think it is definitely not a task, but example or blog would probably be fine, but blog is better IMO since it has a lot of explanatory text.

rcaballeromx:
Thank you @frankbu I agree, we have content that does not reside where readers would expect it.
The key issue for me, and the reason why I would make this content an Example, is the fact that we are updating this content. The main difference between an Example and a Blog post is that posts are not updated over time as the project changes. The Examples in the docs do.
We can have the explanatory text as part of the blog post announcing the example. That text is unlikely to change over time and the post can point to the example which will be updated.

rcaballeromx:
To move things forward, I suggest @vadimeisenbergibm @frankbu, to publish this as a blog post, with after we finish the review process, and file an issue to split the explanatory and the procedural sections later.
Please leave a +1 if you agree.

frankbu:
+1, although the blog post I referenced (above) was updated when the networking model changed. I still think that one belongs as a blog post nevertheless. I do agree that a blog post should not be one that needs to change frequently (mine was updated only once due to the significant change in the Istio APIs). If this article falls into the same category, i.e., it won't likely need to change again, then I think it still could be a blog. That said, I think that if the article is really a follow along to do something set of instructions, it should be an example (or like you suggest, split up).

vadimeisenbergibm:
@rcaballeromx I update the blogs I write, so if something will be changed or not, should not be a criteria. Blogs have "update date", @geeknoid probably add this field exactly due to this fact.

This PR is exactly an update of a blog post from https://archive.istio.io/v0.8/blog/2018/egress-monitoring-access-control/.

In my opinion, a blog posts can have some scenario and a set of steps to address the scenario (the procedural part), since these are technical blog posts.

Let's proceed with this PR as a blog post, and discuss whether we should split it later.

content/blog/2018/egress-monitoring-access-control/index.md
vadimeisenbergibm: @rcaballeromx According to the istio.io style guide (https://istio.io/about/contribute/style-guide/#avoid-using-we), "we" should be avoided, so I assumed that "our" should be avoided too.

content/blog/2018/egress-monitoring-access-control/index.md
vadimeisenbergibm: Now -> at this point, configure for monitoring -> configure to monitor

rcaballeromx:
@frankbu @geeknoid @vadimeisenbergibm 
IMO It would be better for our users to publish clean updated task/s instead of updating the old post.
Beyond my opinion, we recently updated our guide on content types. The key difference in our content type guidance is whether or not the content should be updated over time, not how frequently we do so. The content here very clearly needs to be updated when the APIs change. According to the guidance that means, at least the sections with steps, should be either tasks or examples. 
A compromise would be to publish new clean task/s now and hold publishing the clean version of the conceptual content for later.

vadimeisenbergibm:
@geeknoid @rcaballeromx @frankbu 

Another take of what I propose for a blog post criteria:

* It has a large content, summary-like, provides multiple options.
* It has some story: _Consider an organization that runs applications that process content from cnn.com._
* It can be opinionated.

geeknoid:
Deploy preview for *preliminary-istio* processing.

Building with commit 0bb1d881367a6d5b466aa1bb2ccd27dac785c073

https://app.netlify.com/sites/preliminary-istio/deploys/5c769dc540610c00080c895d

geeknoid:

Built with commit b1371282ebb3edae94f33c9c5f10922012c09971

https://app.netlify.com/sites/preliminary-istio/deploys/5c76a94bea86f800087200b6

vadimeisenbergibm:
@rcaballeromx I have opened an issue to rewrite this blog post as an example https://github.com/istio/istio.io/issues/3411. I have applied your comments. Please approve this PR so I will be able to proceed with a PR that depends on it - https://github.com/istio/istio.io/pull/3179. I will rewrite this blog post as an example later.

rcaballeromx:
@vadimeisenbergibm Thank you for doing this. I'll do a very light review just to catch typos and the like then.

PR #3175, opened 2019-01-30
Replace deployment manager with Istio on GKE

ostromart:
Cherry pick of https://github.com/istio/istio.io/pull/3149.


geeknoid:
Should this be in the release 1.1 branch instead?

Issue #2566, opened 2018-09-09
tablegen.py should parse comments and add them to the generated documentation


sdake:
The https://github.com/istio/istio.github.io/blob/master/scripts/tablegen.py script does not possess any logic to process comments.

The choice of YAML parser (ruamel) is a good one as it permits the ordering and consumption of comments (vs pyyaml which throws away comments and does not preserve ordering).

The code needs enhancements to read comments from the parsed yaml and store them in the appropriate dictionary https://github.com/istio/istio.github.io/blob/master/scripts/tablegen.py#L116-L117

Other enhancements could be made to this implementation as well - its not my finest wine :)

Cheers
-steve

sdake:
Looking for community help across any issues in Istio, and especially this one :)

johnma14:
@sdake I will take a look at this. Thanks for referencing this.

sdake:
@johnma14 nice work on this area!  This really adds alot of value.  We plan to release the docs Friday for 1.1.  I'd like to see this in 1.1, but if it's not ready by Friday don't be concerned - we can easily backport this work.  I'd make this a P0, but it just doesn't block the release.

johnma14:
@sdake I am pretty confident I can get this done by tomorrow. Will do my best towards that goal. Thanks.

johnma14:
Hopefully this latest PR should clear all the related linter issues.

johnma14:
The PR for this was merged, so closing this issue. Thanks.

PR #3188, opened 2019-01-31
Update tablegen.py to update the installation option table for 1.1

johnma14:
Updated tablegen.py to process the configuration options from the values.yaml
files under /istio/install/kubernetes/heml/subcharts directory and the
remaining configuration options like global, istiocoredns, istio_cni from
values.yaml under /istio/install/kubernetes/helm/istio directory.

### DIFF
--- a/scripts/tablegen.py
+++ b/scripts/tablegen.py
@@ -18,14 +18,23 @@ import collections
 import linecache
 import string
 import sys
+import os
 
 from ruamel.yaml import YAML
+from __builtin__ import file
 
+#
 # Reads a documented Helm values.yaml file and produces a
 # MD formatted table.  pip install ruamel to obtain the proper
 # YAML decoder.  ruamel maintains ordering and comments.  The
 # comments are needed in order to decode the commented helm
 # values.yaml file
+#
+YAML_CONFIG_DIR = "istio/install/kubernetes/helm/subcharts"
+ISTIO_CONFIG_DIR = "istio/install/kubernetes/helm/istio"
+VALUES_YAML = "values.yaml"
+ISTIO_IO_DIR = os.path.abspath(__file__ + "/../../")
+CONFIG_INDEX_DIR = "content/docs/reference/config/installation-options/index.md"
 
 def endOfTheList(context, lineNum, lastLineNum, totalNum):
     flag = 0
@@ -39,10 +48,13 @@ def endOfTheList(context, lineNum, lastLineNum, totalNum):
 
     for nextLineNum in range(lineNum + 1, totalNum):
         nextLine = context[nextLineNum]
+
         if  len(nextLine.lstrip()) != 0 and '#' != nextLine.lstrip()[0] and ':' in nextLine:
             if whitespaces >= (len(nextLine) - len(nextLine.lstrip())) / 2:
                 if flag == 0:
                     valueList.append(currentLine.split(':', 1)[1].strip())
+                return True, valueList
+            else:
                 return True, valueList
         elif len(nextLine.lstrip()) != 0 and '#' !=  nextLine.lstrip()[0] and ':' not in nextLine and len(nextLine.strip()) != 0:
             value = nextLine.replace(' ', '')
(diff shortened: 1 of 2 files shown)


scripts/tablegen.py
```diff
@@ -37,13 +46,17 @@ def endOfTheList(context, lineNum, lastLineNum, totalNum):
     if lineNum != lastLineNum:
         return False, valueList
 
+
```
sdake: extra whitespace

scripts/tablegen.py
```diff
@@ -37,13 +46,17 @@ def endOfTheList(context, lineNum, lastLineNum, totalNum):
...
     for nextLineNum in range(lineNum + 1, totalNum):
         nextLine = context[nextLineNum]
+
         if  len(nextLine.lstrip()) != 0 and '#' != nextLine.lstrip()[0] and ':' in nextLine:
             if whitespaces >= (len(nextLine) - len(nextLine.lstrip())) / 2:
                 if flag == 0:
                     valueList.append(currentLine.split(':', 1)[1].strip())
                 return True, valueList
+            else:
```
sdake: if this renders correctly, then the whitespace comparison is not needed and you should be able to delete line 54/58/59

scripts/tablegen.py
```diff
@@ -58,73 +71,119 @@ def endOfTheList(context, lineNum, lastLineNum, totalNum):
...
+
+            storekey = key
+            sk = storekey.split('.', 2)
+            if len(sk) > 1:
+                storekey = '.'.join(sk[:1]).lstrip('.')
+            else:
+                storekey = '.'.join(sk[:0]).lstrip('.')
+
+            #
+            # If we are processing the configurations options within the values.yaml under istio,
```
sdake: how does this condition occur?

scripts/tablegen.py
sdake: Note iirc, the reason for this whitespace comparison was to produce a description field - but my memory is fuzzy. @gyliu513 may know more.

linsun:
@johnma14 great work!!!!   This is so well needed.

The rendered table looks good to me.  I agree with Steve's comment on parsing the comment for each field in the table, that would be really nice.

johnma14:
@sdake @linsun thank you for your kind comments and review. Appreciate it. I will make those changes you recommended in this PR and I can take a look at filling the description fields also. Is there a WI associated to that already or should we create one?

scripts/tablegen.py
johnma14: @sdake without that if-else statement it doesn't render correctly. It only shows the content from the values.yaml under istio and not the values from the sub charts dir.

scripts/tablegen.py
johnma14: I have included comments in the latest PR. This condition is used to check if the configuration option is in values.yaml under istio directory and if this option is a new option or processed before (for example, if this was found in the sub charts directory). The assumption here is that the configuration options under the sub charts directory has the latest values and we want to keep them and only process the configuration options that are only in the istio/values.yaml dir like 'global' option for example. 

Hope this helps. Thanks for the review

scripts/tablegen.py
johnma14: fixed in the latest PR

scripts/tablegen.py
sdake: oh right that makes sense.

scripts/tablegen.py
sdake: The code currently (or did) work by replacing all contents between  `<!-- AUTO-GENERATED-START --` and `<!-- AUTO-GENERATED-END -->`.  We can't select out keys that are in the existing MD as the comments or default may have changed.  Can you take a look at ensuring this case works?  The code takes fractions of a second to run, so we aren't all that concerned about effeciency.

sdake: > @sdake @linsun thank you for your kind comments and review. Appreciate it. I will make those changes you recommended in this PR and I can take a look at filling the description fields also. Is there a WI associated to that already or should we create one?

Can handle the description field as a separate PR. The existing issue for this is: https://github.com/istio/istio.io/issues/2566

johnma14: @sdake This code works exactly like you mentioned, i.e, replace everything in the markdown file between .  That new if-else block processes the configuration options in the values.yaml under the istio/install/kubernetes/helm/istio directory. There are configuration options in there like:
- kiali : we want to make sure that the parameters for this option that will be populated in the index.md are from istio/install/kubernetes/helm/subcharts/kiali/values.yaml and not istio/install/kubernetes/helm/istio/values.yaml

- global: This does not have a corresponding directory under istio/install/kubernetes/helm/subcharts/ so we will get the parameters for this option from stio/install/kubernetes/helm/istio/values.yaml.

That was the thought process.

PR #3200, opened 2019-02-06
Use port 80 for TLS origination by egress gateway

vadimeisenbergibm:
Resolves https://github.com/istio/istio/issues/9546.

The problem was that HTTP and HTTPS were mixed on port 443. The solution is to use port 80 of the gateway for HTTP with or without mTLS. (In case of HTTP/mTLS, specify the protocol as HTTPS).
The 443 port must be used for TLS only, with or without mTLS.


content/docs/examples/advanced-gateways/egress-gateway-tls-origination/index.md
```diff
@@ -86,8 +86,14 @@ be done by the egress gateway, as opposed to by the sidecar in the previous exam
...
     sidecar requests that will be directed to the egress gateway.
 
-    Choose the instructions corresponding to whether or not you have
-    [mutual TLS authentication](/docs/tasks/security/mutual-tls/) enabled in Istio.
+    Choose the instructions corresponding to whether or not you want to enable
+    [mutual TLS Authentication](/docs/tasks/security/mutual-tls/) between the source pod and the egress gateway.
+
+    {{< idea >}}
+    You may want to enable mutual TLS so the traffic between the source pod and the egress gateway will be encrypted.
+    In addition, mutual TLS will allow the egress gateway monitor the identity of the source pods and enable Mixer
```
frankbu: to monitor

frankbu:
@vadimeisenbergibm If I understand correctly, changing the doc like this will avoid the problem, but is there also some code changes needed to protect the gateway from crashing when configured the old way? That seems pretty nasty.

content/docs/examples/advanced-gateways/egress-gateway-tls-origination/index.md
vadimeisenbergibm: Fixed in https://github.com/istio/istio.io/pull/3200/commits/23f3d78a790334368dbf134daa7ab59ce0a08278.

castlemilk:
@frankbu @vadimeisenbergibm, pilot doesn't crash, nor does the gateway. But as the issue describes in [stio/istio#9546](https://github.com/istio/istio/issues/9546), the egressgateway it enter a state that no longer allows egress traffic due to the config rejection. Effectively "crashing" or preventing all egress from that gateway. Not sure why this would be described as dramatic, it's a fairly severe defect if I can break egress simply by applying a gateway configuration which defines servers that specify `protocol: HTTPS` and `protocol: TLS` on the same port [443] (but different domains).

Please let me know if you aren't observing this behaviour.

vadimeisenbergibm:
@castlemilk I addressed your comment https://github.com/istio/istio.io/pull/3200#issuecomment-461199940 in https://github.com/istio/istio/issues/9546#issuecomment-461285270.

geeknoid:
If you rebase, the CI failure will go away.

Issue #3228, opened 2019-02-11
Update Chinese content to use new <warning>, <tip>, and <idea> annotations


geeknoid:
Instead of blockquotes, Istio content now uses annotations to denote special kind of content:

{{< tip >}}
{{< idea >}}
{{< warning >}}

The Chinese content needs to have corresponding changes made.

You can find all places where block quotes are used by enabling the new lint check in scripts/lint_site.sh and then running "make lint".

rootsongjc:
There are 51 files need to change. To change so many files will cost a lot of time.

geeknoid:
All done. Thanks.

PR #3234, opened 2019-02-12
zh: Update Chinese contents

rootsongjc:
Update <tip>, <idea> and <warning> tags in Chinese contents https://github.com/istio/istio.io/issues/3228
@geeknoid Some files have problems, can you figure out why the lint failed in those files?


geeknoid:
The lint errors in the faq section are due to use of `>` in some of the individual faq pages. For example, content_zh/help/faq/mixer/writing-custom-adapters.md has that in it.

Note that at this point, you can't turn on the lint rule fully since there are still a few legitimate uses of blockquotes in the English content which I need to find a way to hide from the lint check. I'm probably just going to introduce {{< quote >}} as an annotation for this purpose.

geeknoid:
I added support for {{< quote >}} and fixed the few remaining English warnings for the new lint rule. If you sync, you should be able to fix all the Chinese warnings and enable the lint rule for everyone.

rootsongjc:
@geeknoid #3228 fixed

rootsongjc:
@geeknoid Could you merge this PR?

rootsongjc:
@geeknoid Rebased, it's ok to merge.

PR #3240, opened 2019-02-12
Add release notes for 1.0.6

ostromart:


geeknoid:
Please also update data/args.yml's full_version field to 1.0.6. This will update the dowload link on the home page.

And please also submit a PR for the master branch with the same notes.

geeknoid:
Let's not include links to pull requests in user-facing material. Let's focus on the feature-level customer-visible effect, rather than on the implementation details.

geeknoid:
Also, let's give some context for user "Fix concurrency problems" isn't informative for someone trying to assess what's different.

Keep in mind that there's a link on the page to let readers access the full source code diff if they are so inclined. But for normal users, we need to explain things in terms that makes sense to someone that uses Istio, not develops for Istio.

ostromart:
> Let's not include links to pull requests in user-facing material. Let's focus on the feature-level customer-visible effect, rather than on the implementation details.

How about linking to the issues instead?
The reason I did this is because I found the full diff quite noisy. For someone who's not an Istio developer, it's hard to tell which are the "user impacting" PR's vs. things most users don't care about. I think it would be useful to the target audience (I guess mostly devops) to have some additional level of detail between something like "Fix security vulnerabilities" and having to look at all PRs in the release for the details. 
At least I would argue for including links to some of the items:
- in mem registry (because wth is an in memory registry and why are we removing it?)
AND
- mutex handling 
- map concurrency
- leak in pilot 
- kubeenv mixer adapter
because without a lot more text it's hard to tell how serious is the problem, and if the problem/fix is relevant to a particular use case.

I added a little more detail to concurrency problems but imo it is still not very informative without more context.

geeknoid:
Our doc set is generally focused on non-programmer users. We don't delve into the implementation details and focus on features and customer-visible experience. We don't expect our customers to be GitHub experts/users.

Take a step back and think of what matters to someone that's just trying to use this stuff. This person doesn't program in Go, and doesn't care about PR. This person just wants to know:

- Which bad behavior that they have seen has been fixed.
- Which new features they can use have been added?
- Which latent security issues have been discovered and fixed.

The key is to strip away all the engineering geek-speak and focus on the surface of the product experienced by users.

ostromart:
We might be in violent agreement here. I'm totally with you about focusing on the problem as experienced by the (devops) user. I'm suggesting linking to the issues rather than PRs and imo this is a level of detail a devops person would be comfortable with reading to answer the questions below.

> This person just wants to know:
> 
> * Which bad behavior that they have seen has been fixed.
> * Which new features they can use have been added?
> * Which latent security issues have been discovered and fixed.
> 
Without lots of explanatory text, it's hard for a devops reader to judge if they need the fix for (say) updating the Go libraries, or "mutex handling". 
The auto diff feature is not very useful, because it dumps the user right into all the implementation details (with extra noise thrown in).

geeknoid:
Oh, and release notes need to come with a quicky blog post announcing the release. See the many examples in the /content/blog directory.

ostromart:
Reworked the text to be more "what does it mean to me as a devops user" oriented, including issue links where appropriate. 
Added blog entry.

content/about/notes/1.0.6/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Update Go `requests` and `urllib3` libraries in Bookinfo sample code per [`CVE-2018-18074`](https://nvd.nist.gov/vuln/detail/CVE-2018-18074) and [`CVE-2018-20060`](https://nvd.nist.gov/vuln/detail/CVE-2018-20060).
+- Do not expose username and password in `Grafana` and `Kiali` ([Issue 7446](https://github.com/istio/istio/issues/7476), [Issue 7447](https://github.com/istio/istio/issues/7447)).
+- Remove in-memory service registry in Pilot. This allowed adding endpoints to proxy configurations from within the cluster through a Pilot debug API.
+
+## Robustness improvements
+
+- Fix Pilot failing to push configuration under load ([Issue 10360](https://github.com/istio/istio/issues/10360)).
+- Fix concurrent read/write to map leading to Pilot restart ([Issue 10868](https://github.com/istio/istio/issues/10868)).
+- Fix goroutine leak in pilot ([Issue 10822](https://github.com/istio/istio/issues/10822)).
```
geeknoid: Fixed a memory leak in Pilot.

content/about/notes/1.0.6/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Security vulnerability fixes
+
+- Update Go `requests` and `urllib3` libraries in Bookinfo sample code per [`CVE-2018-18074`](https://nvd.nist.gov/vuln/detail/CVE-2018-18074) and [`CVE-2018-20060`](https://nvd.nist.gov/vuln/detail/CVE-2018-20060).
+- Do not expose username and password in `Grafana` and `Kiali` ([Issue 7446](https://github.com/istio/istio/issues/7476), [Issue 7447](https://github.com/istio/istio/issues/7447)).
+- Remove in-memory service registry in Pilot. This allowed adding endpoints to proxy configurations from within the cluster through a Pilot debug API.
+
+## Robustness improvements
+
+- Fix Pilot failing to push configuration under load ([Issue 10360](https://github.com/istio/istio/issues/10360)).
```
geeknoid: Fix -> Fixed

content/about/notes/1.0.6/index.md
```diff
@@ -0,0 +1,23 @@
...
+- Fix concurrent read/write to map leading to Pilot restart ([Issue 10868](https://github.com/istio/istio/issues/10868)).
```
geeknoid: - Fixed race condition that would lead Pilot to crash and restart

content/about/notes/1.0.6/index.md
```diff
@@ -0,0 +1,23 @@
...
+- Fix goroutine leak in pilot ([Issue 10822](https://github.com/istio/istio/issues/10822)).
+- Fix to `kubeenv` mixer adapter memory leak ([Issue 10393](https://github.com/istio/istio/issues/10393)).
```
geeknoid: Fix a memory leak in Mixer.

content/about/notes/1.0.6/index.md
ostromart: Done

content/about/notes/1.0.6/index.md
ostromart: Done

content/about/notes/1.0.6/index.md
ostromart: Done

PR #3255, opened 2019-02-13
remove the portion for egress gateway exposed as nodeport

linsun:


linsun:
we should not recommend users to expose egress gateway on node port imho.

linsun:
@rcaballeromx @sdake @costinm PTAL and approve.

PR #3260, opened 2019-02-13
Kiali jaegerURL should use jaeger-query service.

dlinsley:
Prior to change, while using the tracing service on port 80 for the jaegerURL, kiali would never get metrics to compose the service graphs.

Switching jaegerURL to jaeger-query service on port 16686 fixes the issue.  After traffic is generated the service graphs are built and visible in kiali.

### DIFF
--- a/content/docs/tasks/telemetry/kiali/index.md
+++ b/content/docs/tasks/telemetry/kiali/index.md
@@ -72,8 +72,8 @@ integrates with them, you must pass additional arguments to the
 {{< text bash >}}
 $ helm template \
     --set kiali.enabled=true \
-    --set "kiali.dashboard.jaegerURL=http://$(kubectl get svc jaeger-query --namespace istio-system -o jsonpath='{.spec.clusterIP}'):16686" \
-    --set "kiali.dashboard.grafanaURL=http://$(kubectl get svc grafana --namespace istio-system -o jsonpath='{.spec.clusterIP}'):3000" \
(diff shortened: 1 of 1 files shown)


content/docs/tasks/telemetry/kiali/index.md
```diff
@@ -72,7 +72,7 @@ integrates with them, you must pass additional arguments to the
 {{< text bash >}}
```
sdake: I know you didn't introduce this text, but it seems a little dubious.  The expectation here is that Kiali has already been installed (otherwise the jsonpath for clusterIP would not function properly).  Could you add a warning indicating that Istio and Kiali must already be installed?  Looking at the full file view, I see its sort of there already, but there isn't a strong warning - just an idea icon.

How does one install these components without first installing Istio, then reapplying an updated manifest?

dlinsley: Good point @sdake !  I actually didn't use the clusterIP query because I was trying to apply this config at install time.  Using service names as host address was adequate.  I can clean that up to cover a compatible install time and post install update.

rcaballeromx: @sdake There is a note at the begining of the document stating that: The following instructions assume you have installed Helm and use it to install Kiali.
I don't believe a further warning on that regard is needed.
The second question sound like a great suggestion for a related task, but I don't think it is within the scope of what this document is trying to do.

PR #3274, opened 2019-02-16
Update mesh expansion doc with birectional example.

incfly:
https://github.com/istio/istio.io/pull/3274

- It should be `values.meshExpansion.enabled=true` 
- Examples for both through VM -> Mesh and Mesh -> VM.
- Explain the need of DNS and restore `istioctl register`


incfly:
/assign @costinm 
for DNS part correctness.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -213,11 +221,41 @@ spec:
       ports:
         http: 8080
       labels:
-        app: vm1
+        app: vmhttp
         version: 1
```
justinburke: Did you actually run this command and verify that it works? When I tried it, I got this error output: "json: cannot unmarshal number into Go value of type string". Removing "version: 1" fixed the problem.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,38 +174,45 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+    {{< /text >}}
+
+1. Configure Service Discovery for VM by `ServiceEntry`. You add VM services to the mesh by configuring a
 [`ServiceEntry`](/docs/reference/config/istio.networking.v1alpha3/#ServiceEntry). A service entry lets you manually add
 additional services to Istio's model of the mesh so that other services can find and direct traffic to them. Each
 `ServiceEntry` configuration contains the IP addresses, ports, and labels (where appropriate) of all VMs exposing a
 particular service, as in the following example.
 
 {{< text bash yaml >}}
-$ kubectl -n test apply -f - <<EOF
```
justinburke: Why the removal of the namespace? And what isn't this object being created in `${SERVICE_NAMESPACE}`?

content/docs/setup/kubernetes/mesh-expansion/index.md
justinburke: I should add, while removing "version: 1" fixed the error I was seeing, I don't actually know if that was the correct thing to do. No idea why "version: 1" is there to begin with...

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -213,11 +221,41 @@ spec:
...
         version: 1
 EOF
 {{< /text >}}
 
+1. Workloads in Kubernetes cluster need a DNS mapping to resolve the service running on the VM.
+You can integrate with your own DNS system. For illustration purpose, we use `istioctl register`
+which creates a Kubernetes `selector-less` service.
+
+    {{< text bash >}}
+    $ istioctl  register vmhttp 0.0.0.0 8080
```
justinburke: The IP address here, why is it '0.0.0.0'?

content/docs/setup/kubernetes/mesh-expansion/index.md
frankbu: A version label is optional. It would be used if the endpoint represents a specific version of the service that we want to use in subsets for routing.

The proper fix is to put quotes around the number value:
```
version: "1"
```

incfly: I tried this version 1 stuff and not working due to this string/numeric stuff as you mentioned.

@frankbu  I agreed with @justinburke that we should make a minimal working tutorial setup. And we don't have any follow-up steps that depend on this version/routing, stuff.

I think we can remove it.

frankbu: @incfly if you want the simplest possible tutorial, then you can remove the app label too.

Either way, the text just above says: `Each ServiceEntry configuration contains the IP addresses, ports, and labels (where appropriate) ...`, so I think the `labels (where appropriate)` part needs some clarification about what that means. It means needed for route rules to use, which usually includes some kind of version label.

incfly: Got it, thanks for explaination. I changed "1" to "v1" and verified this cmd now works.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: I personally dont think namespace are relevant here  using default is fine. But changed to `SERVICE_NAMESPACE` for consistency.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Good catch. Changed. The actual value of this does not matter as explained in the doc, because this is just to generate a DNS entry for the vmhttp service. But 0.0.0.0 won't work.

justinburke: I'm not finding an explanation in the doc about why that parameter doesn't matter. Could you point me to it?

Also, are there cases where that `istio register` parameter does matter? If not, can we just remove that parameter from the command?

incfly:
PTAL.

I walked through all the steps myself and verify the accuracy of the doc.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +110,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
-    to `/etc/hosts` or to
-    the DNS server. In our example we'll use `/etc/hosts` as it is the easiest way to get things working. The following is
-    an example of updating an `/etc/hosts` file with the Istio gateway address:
+to `/etc/hosts` or to
+the DNS server. In our example we'll use `/etc/hosts` as it is the easiest way to get things working. The following is
+an example of updating an `/etc/hosts` file with the Istio gateway address:
 
     {{< text bash >}}
-    $ echo "35.232.112.158 istio-citadel istio-pilot istio-pilot.istio-system" >> /etc/hosts
+    $ sudo echo "35.232.112.158 istio-citadel istio-pilot istio-pilot.istio-system" >> /etc/hosts
```
justinburke: This command will not work. Use `tee`.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +110,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
     {{< /text >}}
 
 1.  Install `root-cert.pem`, `key.pem` and `cert-chain.pem` under `/etc/certs/`.
 
     {{< text bash >}}
-    $ sudo mkdir /etc/certs
+    $ sudo mkdir -p /etc/certs
```
justinburke: This change is not needed since `/etc/` will already exist.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +185,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+    10.55.246.247
     {{< /text >}}
 
 1.  Then on the mesh expansion machine, add the service name and address to its `etc/hosts` file. You can then connect to
     the cluster service from the VM, as in the example below:
 
     {{< text bash >}}
-$ sudo echo "10.55.246.247 productpage.bookinfo.svc.cluster.local" >> /etc/hosts
-$ curl productpage.bookinfo.svc.cluster.local:9080
+$ sudo echo "10.55.246.247 productpage.default.svc.cluster.local" >> /etc/hosts
```
justinburke: Same as above, this command will not work. Use `tee`.

content/docs/setup/kubernetes/mesh-expansion/index.md
justinburke: I think it is helpful to use the namespace consistently throughout this doc since that will be representative of how an actual user is going to use this doc; we don't expect users to stuff everything in the default namespace, so why not show them how to use a custom namespace?

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +185,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+    <li><a href=".ssh/">.ssh/</a>
+    </body>
+    {{< /text >}}
+
+## Cleanup
+
+After finishes the setup, you can clean up the mesh expansion.
+
+{{< text bash >}}
+$ istioctl deregister -n ${SERVICE_NAMESPACE} vmhttp 10.128.0.17
```
justinburke: Do you mean to reference `${GCE_IP}` here?

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -64,10 +65,11 @@ cluster for mesh expansion, run the following commands on a machine with cluster
     more about customizing Helm charts in the [Helm documentation](https://docs.helm.sh/using_helm/#using-helm).
     {{< /tip >}}
 
-1.  Find the IP address of the Istio ingress gateway, as this is how the mesh expansion machines will access [Citadel](/docs/concepts/security/) and [Pilot](/docs/concepts/traffic-management/#pilot-and-envoy).
+1.  Find the IP address of the Istio ingress gateway, as this is how the mesh expansion machines will access [Citadel](/docs/concepts/security/) and [Pilot](/docs/concepts/traffic-management/#pilot-and-envoy). Also throughout this doc, we assume the GCE VM is part of `$SERVICE_NAMESPACE`.
```
justinburke: This change isn't clear to me. At this point in the doc, the VM is not part of any namespace. Maybe you mean something like: ... we assume that the GCE VM will be added to the namespace `$SERVICE_NAMESPACE` ?

But that's also kind of weird since this is the first mention of SERVICE_NAMESPACE; how is a reader to know what is meant by SERVICE_NAMESPACE at this point in the doc?

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +110,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
           -o jsonpath='{.data.cert-chain\.pem}' |base64 --decode > cert-chain.pem
```
justinburke: We should add some introductory text about SERVICE_NAMESPACE. It's used here for the first time with no mention about what it means.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: I used this and it works. Not sure what you mean not working.

Well it does give the error about permission not enough, and works if we change to `sudo su`. But that's not related to 'tee|echo".

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: I use `-p` this won't produce error.

We don't know what image and linux ditro customer is using, and this doesn't hurt.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: I'm okay to use a non default namespace.

But that adds up more complexity for this tutorial setup. Customer needs to create the namespace, create istio.default service account and verify Citadel catches up generate corresponding key/cert. This makes the tutorial more fragile and complicated.

I think the purpose of the doc is more of illustration rather the real usage in prod. editing `/etc/hosts/` manual download deb package, etc. So robustness and simplicity is more important than being real.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: I add some sentence to explain. Feel free to suggest to expand it.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: This is mentioned in step 1 now.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Good catch, done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Pilot uses ServiceEntry for service discovery to support mesh expansion. istioctl register creates a selectorless DNS, that's all we need for it. For HTTP service, we don't care DNS because we use host header to route. TCP service that might matter. I'm not aware of any doc mentioning that.

Yeah these are all arcane knowledge not well we explained throughout.

And we can improve istioctl register. I don't think is up-to-date with current mesh expansion design.

Let's focus on getting e2e working in this PR for now and address these improvements as follow-up.

content/docs/setup/kubernetes/mesh-expansion/index.md
justinburke: The stdout redirection, ">>", takes place in the non-root shell, the `sudo` command never even sees that redirection command. The above command just executes `echo` as root, and then the non-root shell attempts to append to `/etc/hosts`, which is going to fail unless you're running sudo as root. Changing the command to `sudo su` does not fix this problem for the same reasons.

One way to solve this to do: `echo "1.2.3.4 blah" | sudo tee -a /etc/hosts`.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -64,10 +65,12 @@ cluster for mesh expansion, run the following commands on a machine with cluster
     more about customizing Helm charts in the [Helm documentation](https://docs.helm.sh/using_helm/#using-helm).
     {{< /tip >}}
 
-1.  Find the IP address of the Istio ingress gateway, as this is how the mesh expansion machines will access [Citadel](/docs/concepts/security/) and [Pilot](/docs/concepts/traffic-management/#pilot-and-envoy).
+1.  Find the IP address of the Istio ingress gateway, as this is how the mesh expansion machines will access [Citadel](/docs/concepts/security/) and [Pilot](/docs/concepts/traffic-management/#pilot-and-envoy). VM workloads belong to certain namespace, which means their SPIFFE certificates and `ServiceEntry` should match to that namespace.
```
justinburke: If you're open to feedback on the text, I'd suggest splitting the second and third sentences here, plus the `export SERVICE_NAMESPACE` code bit, into a dedicated step since the namespace discussion has nothing to do with the gateway address. I think this will make it easier to a reader to understand.

Then update the namespace text to something like: In this step, we define the namespace to which the VM will join. Throughout this example, we'll reference the `SERVICE_NAMESPACE` variable; the value of this variable needs to match the namespace setting in configuration files that we later write.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +186,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+    <hr>   
+    <ul>                                                                
+    <li><a href=".bashrc">.bashrc</a>
+    <li><a href=".ssh/">.ssh/</a>
+    </body>
+    {{< /text >}}
+
+## Cleanup
+
+After finishes the setup, you can clean up the mesh expansion.
```
justinburke: s/finishes/completing/

incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Thanks for suggestion! Applied.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Thanks for explanation! Changed and verified on a VM machine.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -42,16 +42,17 @@ cluster for mesh expansion, run the following commands on a machine with cluster
 
     {{< text bash >}}
     $ cd install/kubernetes/helm/istio
-    $ helm upgrade --set global.meshExpansion=true istio-system .
+    $ helm upgrade --set global.meshExpansion.enabled=true istio-system .
```
myidpt: Please fix the `--set global.meshExpansion=true` at line 37 as well.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +115,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
           -o jsonpath='{.data.cert-chain\.pem}' |base64 --decode > cert-chain.pem
     {{< /text >}}
 
-### Setting up the machines
+### Setting up VM machines
```
myidpt: s/VM machines/the VMs

incfly:
I think you're commenting on a outdated commit? The most recent commit
should have both updated.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Oh my bad... It's not there.

incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: I feel definite articles are not needed here for plural forms. Not quite confident since not a native speaker, @justinburke thoughts?

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +115,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
-1.  Install the Debian package with the Envoy sidecar:
+<!-- TODO: find a better way to reference release in the URL instead of manually sync... -->
+
+1.  Copy the `cluster.env` and `*.pem` files that you created in the previous section to the VM. For example
 
     {{< text bash >}}
-    $ curl -L https://storage.googleapis.com/istio-release/releases/1.0.0/deb/istio-sidecar.deb > istio-sidecar.deb
-    $ dpkg -i istio-sidecar.deb
+    $ export GCE_NAME="your-gce-instance"
+    $ gcloud compute scp {key.pem,cert-chain.pem,cluster.env,root-cert.pem} ${GCE_NAME}:~
```
myidpt: This may not work since users may have a wrong project and zone set up.

content/docs/setup/kubernetes/mesh-expansion/index.md
myidpt: I meant "VM machines" is literally "Virtual Machine machines". You can remove the second "machine".

justinburke: Maybe we can resolve this by saying "Setting up the VM" since this guide explicitly shows the user how to setup a single VM, not a collection.

content/docs/setup/kubernetes/mesh-expansion/index.md
justinburke: If there's concern about this, maybe we could add something like this: See [this link](https://cloud.google.com/sdk/gcloud/reference/compute/scp) for additional parameters to point `gcloud compute scp` to different GCP projects or zones.

Could add that above, before "For example".

incfly: I just added `--project` and `--zone` argument since we already referenced them as env var.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -34,24 +34,25 @@ The first step when adding non-Kubernetes services to an Istio mesh is to config
...
     {{< /text >}}
 
     *   If you installed Istio without Helm and Tiller, use `helm template` to update your configuration with the option and
         reapply with `kubectl`:
 
     {{< text bash >}}
+    $ for i in install/kubernetes/helm/istio-init/files/crd*yaml; do kubectl apply -f $i; done
     $ cd install/kubernetes/helm/istio
-    $ helm template --set global.meshExpansion=true --namespace istio-system . > istio.yaml
+    $ helm template --set global.meshExpansion.enabled=true --namespace istio-system . > istio.yaml
```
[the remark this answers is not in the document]
incfly: done.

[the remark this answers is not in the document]
justinburke: This suggested change is not necessary. The first resource in the `istio.yaml` file is the `istio-system` namespace.
Please remove.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +190,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+        name: http
+        protocol: HTTP
+    resolution: STATIC
+    endpoints:
+        - address: ${GCE_IP}
+        ports:
+            http: 8080
+        labels:
+            app: vmhttp
+            label: "v1"
```
frankbu: version: "v1"

content/docs/setup/kubernetes/mesh-expansion/index.md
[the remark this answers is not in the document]
justinburke: Throughout this doc, we don't need to guess which commands are necessary or which commands work. We can run these commands and verify. IMO, verified commands should be the bar we need to meet for adding to documentation.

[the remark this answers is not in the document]
incfly: It maybe env or release discrepancy, but I **verified** this on my workstation before taking this suggestion,

```
helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
  --set global.meshExpansion.enabled=true > ./istio.yaml  && ack -i "kind.*namespace" istio.yaml 
```

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -34,25 +34,26 @@ The first step when adding non-Kubernetes services to an Istio mesh is to config
...
-    $ helm upgrade --set global.meshExpansion=true istio-system .
+    $ helm upgrade --set global.meshExpansion.enabled=true istio-system .
     $ cd -
     {{< /text >}}
 
     *   If you installed Istio without Helm and Tiller, use `helm template` to update your configuration with the option and
         reapply with `kubectl`:
 
     {{< text bash >}}
+    $ for i in install/kubernetes/helm/istio-init/files/crd*yaml; do kubectl apply -f $i; done
```
frankbu: There's a `helm template` way to do this now

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +115,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
 
-### Setting up the machines
+### Setting up the VM
 
 Next, run the following commands on each machine that you want to add to the mesh:
 
-1.  Install the Debian package with the Envoy sidecar:
+<!-- TODO: find a better way to reference release in the URL instead of manually sync... -->
+
+1.  Copy the `cluster.env` and `*.pem` files that you created in the previous section to the VM. For example
```
frankbu: for example:

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +190,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
-You add VM services to the mesh by configuring a
-[`ServiceEntry`](/docs/reference/config/networking/v1alpha3/service-entry/). A service entry lets you manually add
+1. We first setup a HTTP server on the VM instance, serving HTTP traffic on port 8080.
+
+    {{< text bash >}}
+    $ gcloud compute ssh ${GCE_NAME}
+    $ python -m SimpleHTTPServer 8080
+    {{< /text >}}
+
+1. Determine the GCE instance IP address.
```
frankbu: For GCE specific things, you should say something like: `For example, the command to ... is:`. Otherwise it should say at the top these instructions assume GCE.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +190,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+        - address: ${GCE_IP}
+        ports:
+            http: 8080
+        labels:
+            app: vmhttp
+            label: "v1"
+    EOF
+    {{< /text >}}
+
+1. Workloads in Kubernetes cluster need a DNS mapping to resolve the service running on the VM.
```
frankbu: in a Kubernetes cluster

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +190,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+You can integrate with your own DNS system. For illustration purpose, we use `istioctl register`
+which creates a Kubernetes `selector-less` service.
+
+    {{< text bash >}}
+    $ istioctl  register -n ${SERVICE_NAMESPACE} vmhttp ${GCE_IP} 8080
+    {{< /text >}}
+
+1. We deploy a sleep pod in Kubernetes cluster, and wait for the pod to be ready.
```
frankbu: sleep pd in the Kubernetes

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +190,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+    {{< text bash >}}
+    $ kubectl apply -f samples/sleep/sleep.yaml
+    $ kubectl get po
+    NAME                             READY     STATUS    RESTARTS   AGE
+    productpage-v1-8fcdcb496-xgkwg   2/2       Running   0          1d
+    sleep-88ddbcfdd-rm42k            2/2       Running   0          1s
+    ...
+    {{< /text >}}
+
+1. Send a request from sleep container to VM HTTP service.
```
frankbu: from the sleep

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Do you mean the `helm template` generated YAML will include CRD definition as well? In the 
```
helm template --set global.meshExpansion.enabled=true --namespace istio-system . > istio.yaml
ack -i "CustomResourceDefinition istio.yaml
```

I didn't see istio API are there.

Could you provide some cmd or links?

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +190,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+    <hr>   
+    <ul>                                                                
+    <li><a href=".bashrc">.bashrc</a>
+    <li><a href=".ssh/">.ssh/</a>
+    </body>
+    {{< /text >}}
+
+## Cleanup
+
+After completing the setup, you can clean up the mesh expansion.
```
frankbu: Suggest reword to something like this:

Run the following commands to remove the expansion VM from the mesh.

content/docs/setup/kubernetes/mesh-expansion/index.md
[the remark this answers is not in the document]
justinburke: Just to be clear, I was responding to the suggestion that the `istio-system` namespace be explicitly created by the user. That's not necessary since the `istio.yaml` will do that automatically.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Good point. Done.

incfly:
Hi @frankbu can you grant stamp via `/lgtm and /approve` otherwise semes like `content/OWENRS` approval can't be recognized.

content/docs/setup/kubernetes/mesh-expansion/index.md
frankbu: There's an `istio-init` template that installs the CRDs.

https://preliminary.istio.io/docs/setup/kubernetes/helm-install/#option-1-install-with-helm-via-helm-template

frankbu:
@incfly I don't see fixes for the comments I added. You also need to fix all the circleci errors before it can be merged.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -46,14 +46,14 @@ cluster for mesh expansion, run the following commands on a machine with cluster
     $ cd -
     {{< /text >}}
 
-    *   If you installed Istio without Helm and Tiller, use `helm template` to update your configuration with the option and
-        reapply with `kubectl`:
+    *   If you installed Istio without Helm and Tiller, use `helm template` to update your configuration with the option and reapply with `kubectl`:
```
justinburke: I've previously tried adding the `global.meshExpansion.enabled=true` flag to a pre-existing Istio deployment and never was able to get that to work. What did work for me was to do a full, clean install of Istio with that flag.

Has anyone here verified that adding this flag to a pre-existing Istio deployment actually works?

rcaballeromx:
This content needs a detailed review.
/hold

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: I lack the helm/tiller/upgrade experience to tell if these doc are inaccurate or we miss some prerequisite/info ourselves...

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -64,15 +69,21 @@ cluster for mesh expansion, run the following commands on a machine with cluster
     more about customizing Helm charts in the [Helm documentation](https://docs.helm.sh/using_helm/#using-helm).
     {{< /tip >}}
 
-1.  Find the IP address of the Istio ingress gateway, as this is how the mesh expansion machines will access [Citadel](/docs/concepts/security/) and [Pilot](/docs/concepts/traffic-management/#pilot-and-envoy).
+1. In this step, we define the namespace to which the VM will join. Throughout this example, we'll reference the `SERVICE_NAMESPACE` variable; the value of this variable needs to match the namespace setting in configuration files that we later write. For example, `ServiceEntry` below should be stored in that namespace.
 
     {{< text bash >}}
-    $ GWIP=$(kubectl get -n istio-system service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+    $ export SERVICE_NAMESPACE="default"
+    {{< /text>}}
```
frankbu:
```suggestion
    {{< /text >}}
```

frankbu: ^^^ this missing space is causing many of the circleci errors

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -24,6 +24,10 @@ server, or configuring the IPs in any other DNS server accessible from the VM.
 * If you haven't already enabled mesh expansion [at install time with Helm](/docs/setup/kubernetes/helm-install/), you have
 installed the [Helm client](https://docs.helm.sh/using_helm/). You'll need it to enable mesh expansion for the cluster.
 
+{{< tip >}}
+The following instructions assume that the expansion VM is running on GCE and use Google platform-specific command for some of the steps.
```
frankbu:
```suggestion
The following instructions assume that the expansion VM is running on GCE and use Google platform-specific commands for some of the steps.
```

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +119,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
-1.  Install the Debian package with the Envoy sidecar:
+<!-- TODO: find a better way to reference release in the URL instead of manually sync... -->
+
+1.  Copy the `cluster.env` and `*.pem` files that you created in the previous section to the VM. For example:
 
     {{< text bash >}}
-    $ curl -L https://storage.googleapis.com/istio-release/releases/1.0.0/deb/istio-sidecar.deb > istio-sidecar.deb
-    $ dpkg -i istio-sidecar.deb
+    $ export GCE_NAME="your-gce-instance"
+    $ gcloud compute scp --project=${MY_PROJECT} --zone={MY_ZONE} {key.pem,cert-chain.pem,cluster.env,root-cert.pem} ${GCE_NAME}:~
```
wattli: Add '$' in front of {MY_ZONE}

wattli: It is worth to call out 'MY_ZONE' is the zone of the VM, not the k8s cluster.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +119,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
     {{< /text >}}
 
-1.  Copy the `cluster.env` and `*.pem` files that you created in the previous section to the VM.
+1.  Install the Debian package with the Envoy sidecar.
+
+    {{< text bash >}}
+    $ gcloud compute ssh "${GCE_NAME}"
```
wattli: This actually would fail if the vm and k8s cluster are in different zone.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +119,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
+    $ curl -L https://storage.googleapis.com/istio-release/releases/1.1.0-rc.0/deb/istio-sidecar.deb > istio-sidecar.deb
+    $ sudo dpkg -i istio-sidecar.deb
```
wattli: Got some warning messages, not sure if it is expected or not:

lita@lita-test-vm:~$ sudo dpkg -i istio-sidecar.deb
Selecting previously unselected package istio-sidecar.
(Reading database ... 63810 files and directories currently installed.)
Preparing to unpack istio-sidecar.deb ...
Unpacking istio-sidecar (1.1.0-rc.0) ...
Setting up istio-sidecar (1.1.0-rc.0) ...
Adding group `istio-proxy' (GID 113) ...
Done.
Warning: The home dir /var/lib/istio you specified already exists.
Adding system user `istio-proxy' (UID 109) ...
Adding new user `istio-proxy' (UID 109) with group `istio-proxy' ...
The home directory `/var/lib/istio' already exists.  Not copying from `/etc/skel'.
adduser: Warning: The home directory `/var/lib/istio' does not belong to the user you are currently creating.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -108,32 +119,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
+    {{< /text >}}
 
 1.  Add the IP address of the Istio gateway (which we found in the [previous section](#preparing-the-kubernetes-cluster-for-expansion))
```
wattli: Can we move the command to get gateway ip to this section so people don't have to jump around? Considering gateway ip is not used in previous section.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -82,22 +93,22 @@ cluster for mesh expansion, run the following commands on a machine with cluster
...
     ISTIO_CP_AUTH=MUTUAL_TLS
     ISTIO_SERVICE_CIDR=10.55.240.0/20
     {{< /text >}}
 
-1.  (Optional)  If the VM only calls services in the mesh, you can skip this step. Otherwise, add the ports the VM exposes
+1. (Optional)  If the VM only calls services in the mesh, you can skip this step. Otherwise, add the ports the VM exposes
     to the `cluster.env` file with the following command. You can change the ports later if necessary.
 
     {{< text bash >}}
     $ echo "ISTIO_INBOUND_PORTS=3306,8080" >> cluster.env
```
wattli: It is worthwhile to mention that 3306/8080 are just examples and people should change this accordingly in their scenario.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +194,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
     {{< /text >}}
 
 1.  Then on the mesh expansion machine, add the service name and address to its `etc/hosts` file. You can then connect to
     the cluster service from the VM, as in the example below:
 
     {{< text bash >}}
-$ sudo echo "10.55.246.247 productpage.bookinfo.svc.cluster.local" >> /etc/hosts
-$ curl productpage.bookinfo.svc.cluster.local:9080
+$ echo "10.55.246.247 productpage.default.svc.cluster.local" | sudo tee -a /etc/hosts
+$ curl productpage.default.svc.cluster.local:9080
```
wattli: add a '-v' to the end, otherwise you won't see the envoy field.

content/docs/setup/kubernetes/mesh-expansion/index.md
```diff
@@ -173,49 +194,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+
+    {{< text bash >}}
+    $ gcloud compute ssh ${GCE_NAME}
+    $ python -m SimpleHTTPServer 8080
+    {{< /text >}}
+
+1. Determine the VM instance IP address. For example, commands to find GCE instance IP are as below.
+
+    {{< text bash >}}
+    $ export GCE_IP=$(gcloud --format="value(networkInterfaces[0].networkIP)" compute instances describe  ${GCE_NAME})
```
wattli: Mention this command should run on cluster admin machine. We frequently switch from VM and cluster admin machines, it would make the doc more readable to call it out clearly.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: We have this above
> Otherwise, add the ports the VM exposes  to the `cluster.env` file with the following command.

Do you think we should be more specific and clear?

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: They can be the same, right? Also would be better for simplicity. I actually always use the same zone for my setup. Not sure if cross-zone setup would work or not.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Good point to call out where to run each cmd since we need to jump around a lot. but I am a bit of hesitate to say "cluster admin machine" since this is a new terminology that might need more explaination.

@rcaballeromx thoughts?

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done. Good catch.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Not sure. The /var/lib/istio is not supposed to exist on common linux distro. Might because you have installed istio.deb on this particular GCE instane before?

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Done. I add --zone cmd now.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: Oops. Done.

content/docs/setup/kubernetes/mesh-expansion/index.md
incfly: You mean in `Set up the VM` section and step 1?

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -166,6 +185,8 @@ Next, run the following commands on each machine that you want to add to the mes
     $ sudo systemctl start istio
     {{< /text >}}
 
+## VM workloads send request to Kubernetes services
```
rcaballeromx:
```suggestion
## Send requests from VM workloads to Kubernetes services
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+$ echo "10.55.246.247 productpage.default.svc.cluster.local" | sudo tee -a /etc/hosts
+$ curl -v productpage.default.svc.cluster.local:9080
+< HTTP/1.1 200 OK
+< content-type: text/html; charset=utf-8
+< content-length: 1836
+< server: envoy
 ... html content ...
     {{< /text >}}
 
+The `server: envoy` header indicates the traffic indeed is intercepted by sidecar.
```
rcaballeromx:
```suggestion
The `server: envoy` header indicates that the sidecar intercepted the traffic.
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+
 ## Running services on a mesh expansion machine
 
-You add VM services to the mesh by configuring a
-[`ServiceEntry`](/docs/reference/config/networking/v1alpha3/service-entry/). A service entry lets you manually add
+1. We first setup a HTTP server on the VM instance, serving HTTP traffic on port 8080.
```
rcaballeromx:
```suggestion
1. Setup an HTTP server on the VM instance to serve HTTP traffic on port 8080:
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+
+    {{< text bash >}}
+    $ gcloud compute ssh ${GCE_NAME}
+    $ python -m SimpleHTTPServer 8080
+    {{< /text >}}
+
+1. Determine the VM instance IP address. For example, commands to find GCE instance IP are as below.
```
rcaballeromx:
```suggestion
1. Determine the VM instance's IP address. For example, find the IP address
    of the GCE instance with the following commands:
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+
+    {{< text bash >}}
+    $ export GCE_IP=$(gcloud --format="value(networkInterfaces[0].networkIP)" compute instances describe  ${GCE_NAME})
+    $ echo ${GCE_IP}
+    {{< /text >}}
+
+1. Configure Service Discovery for VM by `ServiceEntry`. You add VM services to the mesh by configuring a
```
rcaballeromx: Please update the entire text of the step to avoid clunky content for our users.
```suggestion
1. Configure a service entry to enable service discovery for the VM. You can add VM services to the mesh using a 
    [service entry](/docs/reference/config/networking/v1alpha3/service-entry/). Service entries let you manually add
    additional services to Pilot's abstract model of the mesh. Once VM services are part of the mesh's abstract model,  
    other services can find and direct traffic to them. Each service entry configuration contains the IP addresses, ports, 
    and appropriate labelsof all VMs exposing a particular service, for example:
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+        - address: ${GCE_IP}
+        ports:
+            http: 8080
+        labels:
+            app: vmhttp
+            version: "v1"
+    EOF
+    {{< /text >}}
+
+1. Workloads in a Kubernetes cluster need a DNS mapping to resolve the service running on the VM.
```
rcaballeromx:
```suggestion
1. The workloads in a Kubernetes cluster need a DNS mapping to resolve the the domain names of VM services. To 
    integrate the mapping with you own DNS system, use `istioctl register` and creates a Kubernetes `selector-less` 
    service, for example:
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+You can integrate with your own DNS system. For illustration purpose, we use `istioctl register`
+which creates a Kubernetes `selector-less` service.
+
+    {{< text bash >}}
+    $ istioctl  register -n ${SERVICE_NAMESPACE} vmhttp ${GCE_IP} 8080
+    {{< /text >}}
+
+1. We deploy a sleep pod in the Kubernetes cluster, and wait for the pod to be ready.
```
rcaballeromx:
```suggestion
1. Deploy a pod running the `sleep` service in the Kubernetes cluster, and wait until it is ready:
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+    {{< text bash >}}
+    $ kubectl apply -f samples/sleep/sleep.yaml
+    $ kubectl get po
+    NAME                             READY     STATUS    RESTARTS   AGE
+    productpage-v1-8fcdcb496-xgkwg   2/2       Running   0          1d
+    sleep-88ddbcfdd-rm42k            2/2       Running   0          1s
+    ...
+    {{< /text >}}
+
+1. Send a request from the sleep container to the VM HTTP service.
```
rcaballeromx:
```suggestion
1. Send a request from the `sleep` service on the pod to the VM's HTTP service:
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+    <title>Directory listing for /</title>
+    <body>
+    <h2>Directory listing for /</h2>
+    <hr>   
+    <ul>                                                                
+    <li><a href=".bashrc">.bashrc</a>
+    <li><a href=".ssh/">.ssh/</a>
+    </body>
+    {{< /text >}}
+
```
rcaballeromx: Add text letting the readers know what they achieved:
```suggestion

**Congratulations!**

You successfully configured a service running in a pod within the mesh to send 
traffic to an external service running on a VM outside of the mesh and 
tested that the configuration worked. 

```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -175,49 +196,115 @@ The following example shows accessing a service running in the Kubernetes cluste
...
+## Cleanup
+
+Run the following commands to remove the expansion VM from the mesh.
```
rcaballeromx:
```suggestion
Run the following commands to remove the VM services from the mesh's abstract model.
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -110,32 +121,40 @@ cluster for mesh expansion, run the following commands on a machine with cluster
           -o jsonpath='{.data.cert-chain\.pem}' |base64 --decode > cert-chain.pem
     {{< /text >}}
 
-### Setting up the machines
+### Setting up the VM
 
 Next, run the following commands on each machine that you want to add to the mesh:
 
-1.  Install the Debian package with the Envoy sidecar:
+<!-- TODO: find a better way to reference release in the URL instead of manually sync... -->
```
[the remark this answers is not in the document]
incfly: Filed https://github.com/istio/istio.io/issues/3496

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Adopted. but do we want to break the word `ServiceEntry`, which is a specific terminology in Istio?

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Although not matter in markdown, why newline in these cases? I searched a bit on [style guide](https://preliminary.istio.io/about/contribute/style-guide) but didn't find any recommended line length.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Thanks for pointing out, this is helpful :)

But the wording is not quite  accurate right? The whole point of the `mesh expansion` is to make VM part of the mesh beyond Kubernetes clusters. So the VM are also part of the mesh, not external of it. Or your intention is to replace that with "Kubernetes" cluster?

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Given the page is titled as "mesh expansion" I think "remove the VM from the mesh" is clear enough? Any reason to mention "abstract model"? I'm not sure if we have settled down a model for VM.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -36,24 +40,25 @@ The first step when adding non-Kubernetes services to an Istio mesh is to config
...
     {{< text bash >}}
     $ cd install/kubernetes/helm/istio
-    $ helm upgrade --set global.meshExpansion=true istio-system .
+    $ helm upgrade --set global.meshExpansion.enabled=true istio-system .
     $ cd -
     {{< /text >}}
 
-    *   If you installed Istio without Helm and Tiller, use `helm template` to update your configuration with the option and
-        reapply with `kubectl`:
+    *   If you installed Istio without Helm and Tiller, use `helm template` to update your configuration with the option and reapply with `kubectl`:
```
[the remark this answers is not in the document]
incfly: Yes. I think that's accurate given the mesh expansion option is only available as a helm value flag.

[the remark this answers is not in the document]
incfly: I updated the prerequisite sections to reflect `helm` client is a must anyway.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -26,6 +26,10 @@ server, or configuring the IPs in any other DNS server accessible from the VM.
 * If you haven't already enabled mesh expansion [at install time with Helm](/docs/setup/kubernetes/install/helm/), you have
```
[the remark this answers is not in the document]
incfly: Done.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
rcaballeromx: Yes, we do. See:  https://preliminary.istio.io/about/contribute/style-guide/#use-consistent-capitalization for details.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
rcaballeromx: This is a recommendation from GitHub. Up to 80 characters are properly shown in a single line for review.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
rcaballeromx: I agree, you are adding the VM to the mesh but not the cluster. Changing "mesh" to cluster on the second line of my suggestion should clarify.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
rcaballeromx: So, the abstract model is not the model of the VM but the abstract model Pilot maintains for the mesh. Remove VM from mesh is not false but it's too general. The added specificity helps readers understand the behavior.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Gotcha. Done.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -23,8 +23,11 @@ is not required to have access to the cluster IP addresses assigned by Kubernete
 include exposing the Kubernetes DNS server through an internal load balancer, using a Core DNS
 server, or configuring the IPs in any other DNS server accessible from the VM.
 
-* If you haven't already enabled mesh expansion [at install time with Helm](/docs/setup/kubernetes/install/helm/), you have
-installed the [Helm client](https://docs.helm.sh/using_helm/). You'll need it to enable mesh expansion for the cluster.
+* Mesh expansion has to be enabled by Helm. You need to install the [Helm client](https://docs.helm.sh/using_helm/).
+
+{{< tip >}}
```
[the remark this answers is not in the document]
incfly: Removed

incfly:
Now all the suggests are applied and circle ci is passing, PTAL, thanks! @rcaballeromx

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -212,25 +210,30 @@ $ curl -v productpage.default.svc.cluster.local:9080
...
     $ export GCE_IP=$(gcloud --format="value(networkInterfaces[0].networkIP)" compute instances describe  ${GCE_NAME})
     $ echo ${GCE_IP}
     {{< /text >}}
 
-1. Configure Service Discovery for VM by `ServiceEntry`. You add VM services to the mesh by configuring a
+1. Configure a service entry to enable service discovery for the VM. You can add VM services to the mesh using a
+    [service entry](/docs/reference/config/networking/v1alpha3/service-entry/). Service entries let you manually add
+    additional services to Pilot's abstract model of the mesh. Once VM services are part of the mesh's abstract model,
+    other services can find and direct traffic to them. Each service entry configuration contains the IP addresses, ports,
+    and appropriate labels of all VMs exposing a particular service, for example:
```
rcaballeromx: Thanks for accepting my suggestion but please remove lines 237-238.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -260,15 +263,17 @@ particular service, as in the following example.
     EOF
     {{< /text >}}
 
-1. Workloads in a Kubernetes cluster need a DNS mapping to resolve the service running on the VM.
+1. The workloads in a Kubernetes cluster need a DNS mapping to resolve the the domain names of VM services. To
+    integrate the mapping with you own DNS system, use `istioctl register` and creates a Kubernetes `selector-less`
+    service, for example:
```
rcaballeromx: Thanks for accepting my suggestion but please remove lines 269-270.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -260,15 +263,17 @@ particular service, as in the following example.
     EOF
     {{< /text >}}
 
-1. Workloads in a Kubernetes cluster need a DNS mapping to resolve the service running on the VM.
+1. The workloads in a Kubernetes cluster need a DNS mapping to resolve the the domain names of VM services. To
```
rcaballeromx:
```suggestion
1. The workloads in a Kubernetes cluster need a DNS mapping to resolve the domain names of VM services. To
```

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -288,12 +293,18 @@ which creates a Kubernetes `selector-less` service.
...
     </body>
     {{< /text >}}
 
+**Congratulations!**
+
+You successfully configured a service running in a pod within the mesh to send
+traffic to an external service running on a VM outside of the mesh and
+tested that the configuration worked.
+
 **Congratulations!**
```
rcaballeromx: Thanks for accepting my suggestion but please remove lines 309-311.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Done.

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
incfly: Done.

incfly:
Finally we got it in!!!

Thanks everyone for the feedback on this PR!

wattli:
Great job!

PR #3275, opened 2019-02-16
Fix too many redirects path in Kiali task

liamawhite:
When following along with the docs this path gives me an ERR_TOO_MANY_REDIRECTS but changing it to `/kiali/console` fixes it.

Signed-off-by: Liam White <liam@tetrate.io>


sdake:
cc @jmazzitelli 

FYI - not sure if the too many redirects in intentional but that result seems odd.

Cheers
-steve

jmazzitelli:
@sdake This fix to the docs is fine - it will work for Kiali v0.12 and newer. However, just FYI: this redirect bug has been fixed in Kiali v0.13 so when we move Istio to pull in that version (or newer) of Kiali, this doc fix wouldn't be necessary (though, as I say, it will still work if you go directly to /kiali/console)

If interested - here is the Kiali bug report on this:

reported on Istio: https://github.com/istio/istio/issues/11491
reported on Kiali JIRA: https://issues.jboss.org/browse/KIALI-2188
the PR that fixed this: https://github.com/kiali/kiali/pull/774

PR #3291, opened 2019-02-19
Fixes markdown conversion issues

eoftedal:


geeknoid:
This is a generated file. Plesse fix this in the original source in the istio/api repo.

Thanks.

PR #3296, opened 2019-02-19
Rejigger the setup section to make room for upgrade instructions.

geeknoid:
Here's a prototype updated layout for setup & upgrade. There's a whole Upgrading Istio section which only contains a single doc at the moment, but this is where we can pour more docs as they become available.

WDYT?

I still need to fix up a bunch of aliases before committing this, but I thought I'd get feedback being doing that work.


geeknoid:
Closing this PR, Rigs will take this work over.

PR #3303, opened 2019-02-20
verify-install command doc

irisdingbj:
verify-install command


geeknoid:
This is a generated file. Please fix the problem in the original source in the istio/istio repo.

Thanks.

Issue #3321, opened 2019-02-21
IBM Cloud quick start guide use istio helm repo


linsun:
In addition to #3285, I think we should switch to use the istio helm repo, vs the ibm chart repo because it was not kept updated.

@Rachael-Graham

linsun:
@ericvn

ericvn:
I believe this was tackled and completed in the 1.1 docs. Currently, we point to the base common Istio docs from the IBM Cloud Public which uses the Istio repo or download.

PR #3327, opened 2019-02-21
Update the IBM Cloud Quick start

ericvn:
- Use a demo profile for lower resources
- Update to mention tested Kubernetes releases
- use helm-service-account-yaml
- updated helm commands for CRDS, etc.


ericvn:
@frankbu I just noticed there is another Quick Start PR out there, but would you review this one as well. I can rebase on the other if needed. The other PR includes a separate Managed Istio section which I Slacked with Lin about this am which I don't have.

ericvn:
The helm chart updates are to cover: https://github.com/istio/istio.io/issues/3321

content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -61,69 +39,95 @@ This guide installs the current release version of Istio.
     $ helm init --service-account tiller
     {{< /text >}}
 
-1. Add the IBM Cloud Helm repository to your Helm instance:
+### Deploy the Istio Helm charts
+
+1. Install the `istio-init` chart to bootstrap all the Istio's CRDs:
```
frankbu: all the Istio CRDs

content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -61,69 +39,95 @@ This guide installs the current release version of Istio.
...
     {{< text bash >}}
-    $ helm repo add ibm-charts https://registry.bluemix.net/helm/ibm-charts
+    $ helm install install/kubernetes/helm/istio-init --name istio-init --namespace istio-system
     {{< /text >}}
 
-### Deploy the Istio Helm chart
-
-1. If using a Helm version prior to 2.10.0, install Istio’s Custom Resource Definitions via `kubectl apply`, and wait a few seconds for the CRDs to be committed
-to the Kubernetes API server:
+    To verify all Istio's CRDs were committed in the Kubernetes api-server, check the number of CRDs with the specified suffix with the following command. Verify that the number of total CRDs created was `56` for Istio:
```
frankbu: To verify that all ...

content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -61,69 +39,95 @@ This guide installs the current release version of Istio.
...
 
     {{< text bash >}}
-    $ kubectl apply -f https://raw.githubusercontent.com/IBM/charts/master/stable/ibm-istio/templates/crds.yaml
+    $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | wc -l
```
frankbu: You can display the expected output, 56, here:
```
    {{< text bash >}}
    $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | wc -l
    56
    {{< /text >}}
```

content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -61,69 +39,95 @@ This guide installs the current release version of Istio.
...
+    prometheus-5488844b5c-vwd9p               1/1     Running     0          57m
+    servicegraph-86b55fc8b8-k87n9             1/1     Running     0          57m
     {{< /text >}}
 
+     _Note:_ The IBM Cloud Kubernetes Service free cluster contains a single worker providing fewer resources, so the `demo` profile was specified. If using a paid cluster of sufficient size, one can remove the --values parameter to use the `values.yaml` values.
+
 ### Upgrade
 
-1. To upgrade your Istio Helm chart to the latest version:
+1. To keep all the Istio [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions) (CRDs) upgraded, you must upgrade the `istio-init` chart. If the chart doesn't exist you must install the new release.
```
frankbu: This is confusing.
Also, why would the istio-init chart not exist?
Maybe reword something like this:

```
1. Upgrade the `istio-init` chart to keep all the Istio [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions) (CRDs) up to date:
    {{< warning >}}
    If If the chart doesn't exist yet, you'll need to install the release instead. <--- is that what you are trying to say?
    {{< /warning >}}
```

content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -61,69 +39,95 @@ This guide installs the current release version of Istio.
...
+    istio-policy-7b6bfcf94d-v6sr7             2/2     Running     2          57m
+    istio-sidecar-injector-6657dd87b9-ccg87   1/1     Running     0          57m
+    istio-telemetry-77d557f66-nsn87           2/2     Running     2          57m
+    istio-tracing-6994cd89bb-gcssk            1/1     Running     0          57m
+    kiali-69d6978b45-5zjzl                    1/1     Running     0          57m
+    prometheus-5488844b5c-vwd9p               1/1     Running     0          57m
+    servicegraph-86b55fc8b8-k87n9             1/1     Running     0          57m
     {{< /text >}}
 
+     _Note:_ The IBM Cloud Kubernetes Service free cluster contains a single worker providing fewer resources, so the `demo` profile was specified. If using a paid cluster of sufficient size, one can remove the --values parameter to use the `values.yaml` values.
```
frankbu: I think this kind of thing is supposed to be a tip: `{{< tip >}} ... {{< /tip >}}`.
Would also probably move it up to right after the command.

content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -61,69 +39,95 @@ This guide installs the current release version of Istio.
...
     {{< text bash >}}
-    $ kubectl apply -f https://raw.githubusercontent.com/IBM/charts/master/stable/ibm-istio/templates/crds.yaml
+    $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | wc -l
     {{< /text >}}
 
 1. Install the Helm chart to your cluster:
 
     {{< text bash >}}
-    $ helm install ibm-charts/ibm-istio --name=istio --namespace istio-system
+    $ helm install install/kubernetes/helm/istio --name istio --namespace istio-system --values install/kubernetes/helm/istio/values-istio-demo.yaml
```
frankbu: Suggest move up and clarify tip:
```
    {{< tip >}}
    The Istio `demo` profile (`install/kubernetes/helm/istio/values-istio-demo.yaml`) is specified in the above command to support the IBM Cloud Kubernetes Service free cluster, whcih only contains a single worker providing fewer resources than needed.
    If using a paid cluster of sufficient size, you can remove the `--values` parameter which will use the default Istio configuration values instead.
    {{< /tip >}}
```

ericvn:
@frankbu I'm fine with @Rachael-Graham going in first and I can rebase. One question for you and @linsun would be if I should update this to no longer use Helm/Tiller but to simply do a `kubectl apply -f ../istio-demo.yaml and then use the istioctl validate to make sure things are running? I'm fine making that change in light of a perceived move from using Tiller.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: The `Istio-init` chart doesn't exist pre 1.1 so when _upgrading_ from 1.0 the user won't find a chart to _update_. In that case, the user will _install_ the chart from the current release.

I'll work on the wording.

ericvn: One might argue that we could remove the *Upgrade* section since a Quick Start should be new and not an upgrade of an existing install.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Done and updated to 58.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Done.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Done.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Done.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Done.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Done.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Done.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Done.

linsun:
@ericvn I'd keep tiller in as an option as the main istio helm install also has that option.  There are users who may prefer tiller however I do think most users will use helm template and kubectl apply.

content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -69,69 +46,99 @@ Make sure to use the `kubectl` CLI version that matches the Kubernetes version o
...
+    $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | wc -l
+    58
     {{< /text >}}
 
 1. Install the Helm chart to your cluster:
 
+    {{< tip >}}
+    The Istio `demo` profile (`install/kubernetes/helm/istio/values-istio-demo.yaml`) is specified in the above command to support the IBM Cloud Kubernetes Service free cluster, which only contains a single worker providing fewer resources than needed.
+    If using a paid cluster of sufficient size, you can remove the `--values` parameter which will use the default Istio configuration values instead.
+    {{< /tip >}}
```
frankbu: This tip should be after the bash section. Or it needs to be changed to say "in the following command".

ericvn: Done.

content/docs/setup/kubernetes/quick-start-ibm/index.md
```diff
@@ -69,69 +46,99 @@ Make sure to use the `kubectl` CLI version that matches the Kubernetes version o
...
     {{< text bash >}}
-    $ helm repo add ibm-charts https://registry.bluemix.net/helm/ibm-charts
+    $ helm install install/kubernetes/helm/istio-init --name istio-init --namespace istio-system
     {{< /text >}}
 
-### Deploy the Istio Helm chart
-
-1. If using a Helm version prior to 2.10.0, install Istio’s Custom Resource Definitions via `kubectl apply`, and wait a few seconds for the CRDs to be committed
-to the Kubernetes API server:
+    To verify that all Istio CRDs were committed in the Kubernetes api-server, check the number of CRDs with the specified suffix with the following command. Verify that the number of total CRDs created was `58` for Istio:
```
frankbu: I don't understand what CRDs with the specified suffix means. What suffix is it talking about? Why not just one simple sentence like this:

Verify that all `58` Istio CRDs were committed to the Kubernetes api-server using the following command:

ericvn: I admit to pulling that section from the base install documentation. I believe the original author is describing the `grep` piece of the command. The command gets all the CRDS, greps those with the specific suffix, and then gets the count of lines.

That said, the suggested text is more concise.

Done.

frankbu: It greps, but not for any suffix. Maybe it could say "with one of the substrings as specified in the following grep command". Then it would be correct, but way too verbose.

I notice that this verify step is also in the install with helm instructions (2 places), although the wording there is a little different (correct), and also in the minimal install instructions.

I suggest you change it to the short version that I suggested in all 4 places and then when we merge this PR it will be the same everywhere. It should probably be moved to a fragment file so that it doesn't need to be copied everywhere, but I think that's a cleanup exercise that can be done later.

frankbu:
@linsun tiller is not just one option in these instructions, it's the only option. If we want to include both tiller and non tiller options, the doc needs to be changed a lot, but I think this option1/option2 style is becoming a mess in all the install instructions, not just this one.

I suggest leaving as is for now, and we need to figure out a better way to document tiller and non-tiller ways, and then change all the docs consistently.

content/docs/setup/kubernetes/quick-start-ibm/index.md
ericvn: Updated so all 4 places match suggested text.

PR #3333, opened 2019-02-21
IBM Platform update

ericvn:
- Added a section for using the new Istio add-on for IKS clusters
- update IKS command to use --export. Mpt sure if we can totally move
  new ks command syntax yet.


ericvn:
@frankbu @linsun Update to the IBM Platform setup doc. I did take Managed Istio from an earlier PR.

PR #3337, opened 2019-02-21
More instructions for setting up GKE for Istio

utka:


content/docs/setup/kubernetes/quick-start-gke/index.md
```diff
@@ -30,6 +30,12 @@ Then add the `Kubernetes Engine Admin` role:
 
 {{< image width="70%" link="./dm_gcp_iam_role.png" caption="GKE-IAM Role" >}}
 
+Create `clusterrolebinding` for Istio to be able to manage the cluster.
```
rcaballeromx: This instruction comes out of nowhere for the reader. Why do readers need to create a `clusterrolebinding`? What is that? Where would they run the command?
Is this something that is already covered and better on the GKE docs?

geeknoid:
@utka This seems like useful information. Can you give a bit more context so we can merge this in?

Thanks.

geeknoid:
Please reopen if you decide to move this forward.

Issue #3191, opened 2019-02-01
Reference page documenting annotations


esnible:
Istio supports many "annotations" which affect deployment.  For example, _pkg/kube/inject/inject.go_ includes sidecar.istio.io/inject, sidecar.istio.io/status, sidecar.istio.io/proxyImage, sidecar.istio.io/interceptionMode, status.sidecar.istio.io/port, readiness.status.sidecar.istio.io/initialDelaySeconds, readiness.status.sidecar.istio.io/periodSecond, readiness.status.sidecar.istio.io/failureThreshold, readiness.status.sidecar.istio.io/applicationPorts", validateReadinessApplicationPorts, traffic.sidecar.istio.io/includeOutboundIPRanges, traffic.sidecar.istio.io/excludeOutboundIPRanges", ValidateExcludeIPRanges, traffic.sidecar.istio.io/includeInboundPorts, and traffic.sidecar.istio.io/excludeInboundPorts.  They may be more.

There is documentation for sidecar.istio.io/inject ( https://preliminary.istio.io/docs/setup/kubernetes/sidecar-injection.html ) but the other annotations don't show up searching https://preliminary.istio.io .

geeknoid:
Where does these annotations show up in the Istio source code?

esnible:
Perhaps related to https://github.com/istio/istio/issues/9710

esnible:
https://github.com/istio/istio.io/pull/3349#issuecomment-466539382 says we need to fix in the code.  Will need to be post 1.1.

geeknoid:
Coming shortly...

forkjoseph:
is this still coming shortly?

kbruner:
Also, will this cover the precedence when a resource annotation performs the same functionality as a field in an Istio resource?  Example: `networking.istio.io/exportTo` vs the `exportTo` key in a `ServiceEntry` or `VirtualService`

PR #3349, opened 2019-02-21
Reference for all Istio annotations

esnible:
A single page listing all object annotations provided by Istio.

To resolve issue https://github.com/istio/istio.io/issues/3191

Note that this PR has some TODOs because I couldn't figure out what the annotation was for.

### DIFF
--- a/content/docs/reference/config/annotations/index.html
+++ b/content/docs/reference/config/annotations/index.html
@@ -1,0 +1,290 @@
+---
+title: Annotations
+description: Istio Annotations that can be applied to other Mesh resources.
+location: https://istio.io/docs/reference/config/annotations.html
+layout: protoc-gen-docs
+generator: protoc-gen-docs
+number_of_entries: 1
+---
+<p>Istio Annotations that can be applied to other Mesh resources.</p>
+
+<h2 id="Sidecar">Sidecar</h2>
+<section>
+<p>Sidecar annotations alter the behavior of sidecars.</p>
+
+<p>See <a href="/docs/setup/kubernetes/sidecar-injection/#policy">Automatic Sidecar Injection</a>
+
+<table class="message-fields">
+<thead>
+<tr>
+<th>Annotation</th>
+<th>Type</th>
+<th>Description</th>
+</tr>
+</thead>
+<tbody>
+<tr id="sidecar.istio.io/inject">
+<td><code>sidecar.istio.io/inject</code></td>
+<td><code>boolean</code></td>
+<td>
+<p>The sidecar injector will inject the sidecar into pods by default. Add the sidecar.istio.io/inject annotation with value false to the pod template spec to override the default and disable injection.</p>
+
+</td>
+</tr>
+<tr id="sidecar.istio.io/status">
+<td><code>sidecar.istio.io/status</code></td>
+<td><code>string</td>
+<td>
+<p>TODO</p>
+
+</td>
+</tr>
+<tr id="sidecar.istio.io/proxyImage">
+<td><code>sidecar.istio.io/proxyImage</code></td>
+<td><code>string</td>
+<td>
+<p>TODO Docker image</p>
+
+</td>
+</tr>
+<tr id="sidecar.istio.io/interceptionMode">
+<td><code>sidecar.istio.io/interceptionMode</code></td>
+<td><code>string</td>
+<td>
+<p>See <a href="/docs/setup/kubernetes/istio-cni-install/#traffic-redirection-details">Traffic Redirection</a>.</p>
+
+</td>
+</tr>
+<tr id="status.sidecar.istio.io/port">
+<td><code>status.sidecar.istio.io/port</code></td>
+<td><code>string</td>
+<td>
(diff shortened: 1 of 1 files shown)


content/docs/reference/config/annotations/index.html
```diff
@@ -0,0 +1,290 @@
+---
+title: Annotations
+description: Istio Annotations that can be applied to other Mesh resources.
+location: https://istio.io/docs/reference/config/annotations.html
```
geeknoid: Please remove location, layout, generate, and number_of_entries, these are only used for the auto-generated docs. You should also have a weight: value to control the sorting order within the contanining directory.

Any reason you chose to do this doc in HTML instead of markdown?

mandarjog:
Is this extracted automagically? If not @geeknoid this is ripe for automation.

content/docs/reference/config/annotations/index.html
esnible: I would like to do this doc in markdown.  I based this page on its siblings in _content/docs/reference_.  Is there a better example that I could follow?

geeknoid: All the other docs in istio.io are in markdown. The only HTML material is what the proto conversion tool generates.

Check out https://preliminary.istio.io/about/contribute/creating-and-editing-pages/

geeknoid:
@mandarjog Where is this information in the source base? Any ideas how to automate extraction?

geeknoid:
OK, I found the definitions in istio/istio/pilot/pkg/kube/inject.go. Is that where all annotations are defined? We should simply:

- Extend the registeredAnnotation type to carry a description string and whatever other metadata is useful (like perhaps a type).

- Fill in the new fields when initiatlizing the array of annotations.

- Add a hidden command-line option to pilot to dump out the set of annotations on stdout in some site-friendly form.

- Update istio.io/scripts/grab_reference_docs.sh to trigger pilot with the hidden option to extract the data and dump it into the right place in the site.

esnible:
@geeknoid I had to hunt through the code for these and I am not sure if I have the correct list.  I expect that I will need to contact OWNERS for each area to see if the annotations their source code processes are documentable. The TODOs are for that.

It is possible https://github.com/istio/istio/issues/9710 is to create a single place in the code where all things like annotations are defined.

content/docs/reference/config/annotations/index.html
esnible: @geeknoid Let me be concrete.   https://preliminary.istio.io/docs/reference/config/authorization/istio.rbac.v1alpha1/ includes the text "A ServiceRole specification includes a list of rules".  I searched for that text on my system and found that text in just two places, _istio.io/content/docs/reference/config/authorization/istio.rbac.v1alpha1/index.html_ and _api/rbac/v1alpha1/rbac.proto_.  For annotation we don't have a .md and we don't have a .proto.

Please advise.

geeknoid:
So Ed, the procedures I outlined below is what I think we need to do. Basically, we need to go in the original source code where these annotations are defined and add a descriptive string to them. Then we add a command-line option to have the corresponding binary dump out all the annotations it knows about. We can then grab this output and inject into the site.

This is how we collect command-line arguments to publish on the site today.

So if you want to go to the process of adding the needed descriptions and command-line option to the binaries, I can handle the work for getting that stuff into the web site. I would recommend adding something to istio/pkg/collateral for this use (that's what collect command-lines and outputs them for use by the web site).

content/docs/reference/config/annotations/index.html
```diff
@@ -0,0 +1,290 @@
...
+</tr>
+<tr id="traffic.sidecar.istio.io/excludeInboundPorts">
+<td><code>traffic.sidecar.istio.io/excludeInboundPorts</code></td>
+<td><code>string</td>
+<td>
+<p>TODO comma separated list or *</p>
+
+</td>
+</tr>
+<tr id="sidecar.istio.io/v1alpha1/statsInclusionPrefixes">
```
mandarjog: Several corrections here
1. change annotation name
`sidecar.istio.io/statsInclusionPrefixes`
2. Instead of `Example` add the default values for inclusion which are `listener_manager,cluster_manager,cluster.xds-grpc,server,http_mixer_filter,tcp_mixer_filter`
3. State that this replaces the default list above.  These following stats:`listener_manager,cluster_manager` , are required for proper functioning of the readiness subsystem therefore you should include these two with your list.

@geeknoid what do you think is it better to just fix this in code, or doc is fine?

esnible: Is there a PR for `sidecar.istio.io/statsInclusionPrefixes` in the code itself or should I also do that work separately now?

When I implemented statsInclusionPrefixes I did not have this guidance.  The current implementation works as it says.  I am unclear about alpha/beta versioning of annotations.  Should we use `alpha.sidecar.istio.io/statsInclusionPrefixes` to indicate it is an experimental feature?

geeknoid:
@mandarjog We should fix all this in the code and make this file so it is auto-generated, as I outlined above.

geeknoid:
I'm going to close this. What we need is automation as mentioned above.

Issue #3269, opened 2019-02-15
Istio 1.1 Minikube install instructions need to be updated


brian-avery:
The quickstart guide says that Istio 1.1 is tested on Kubernetes 1.10, 1.11, and 1.12 and then the minikube section talks about how to set up Kubernetes 1.9 or 1.10 (https://preliminary.istio.io/docs/setup/kubernetes/platform-setup/minikube/). The 1.9 option should be removed. Should the 1.10 instructions continue to refer to 1.10 or one of the newer tested versions? 

On Kubernetes 1.9:

$ minikube start --memory=8192 --cpus=4 --kubernetes-version=v1.9.4 \
    --extra-config=apiserver.admission-control="NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota" \
    --vm-driver=`your_vm_driver_choice`

On Kubernetes 1.10:

$ minikube start --memory=8192 --cpus=4 --kubernetes-version=v1.10.0 \
    --vm-driver=`your_vm_driver_choice`

rcaballeromx:
@brian-avery I can update the guide.
@geeknoid Do we want to maintain the 1.9 command or should I only include the Kubernetes 1.10 command?

geeknoid:
@rcaballeromx Yes, all the 1.9 stuff should be excised. But it's likely that 1.10 needs to go too, and we need to add 1.13 stuff.

@linsun @costinm @smawson @louiscryan @rshriram  Guys, what is our k8s support window for 1.1? For 1.0 we had:

supported_kubernetes_versions: ["1.10", "1.11", "1.12"]

Should we drop 1.10 and add 1.13?

louiscryan:
I think we should retire 1.9 and if probably 1.10 seeing as its almost a
year old.

brian-avery:
Happy to make the changes, or @rcaballeromx can if he likes. Just need to know what versions to put.

geeknoid:
Already done. I've got it pegged at 1.11, 1.12, and 1.13 for Istio 1.1

brian-avery:
Awesome. Thanks. I see the changes on "Quick Start with Kubernetes" already, but not the minikube setup guide.

brian-avery:
Fixed by #3351

PR #3351, opened 2019-02-21
Update to Kubernetes 1.13

brian-avery:
Updating this to suggest using Kubernetes 1.13.


brian-avery:
@geeknoid Can you review? Looks like we missed a page. Fixes #3269

PR #3352, opened 2019-02-21
Remove In-Depth Telemetry Example

douglas-reid:
The "example" is just a curated set of links against existing tasks and does little in the way of adding any information on top of those tasks. The content definitely isn't an "example" (it used to be a "guide") and doesn't fit the rest of the content in that heading.


geeknoid:
We should add an alias on a page somewhere so that people with a bookmark to this page will land someplace real. Perhaps the root of the example section? Or the root of the telemetry tasks?

See https://preliminary.istio.io/about/contribute/creating-and-editing-pages/#renaming-moving-or-deleting-pages for info on aliases.

douglas-reid:
Added alias. PTAL.

PR #3358, opened 2019-02-22
new networking reference docs

rshriram:
One page per proto
Signed-off-by: Shriram Rajagopalan <shriramr@vmware.com>


linsun:
I think this looks better than what we had before with 1 huge page.
@frankbu any objections?

geeknoid:
This will need a great many aliases in order to ensure old links keep working.

geeknoid:
This looks pretty good.

First step is to fix all the broken links reported by the linter. Once those are all fixed, next step is to add aliases to redirect old bookmarks to the new pages.

content/docs/reference/config/istio.mesh.v1alpha1/index.html
```diff
@@ -75,7 +75,7 @@ <h2 id="ConfigSource">ConfigSource</h2>
 </tr>
 <tr id="ConfigSource-tls_settings">
 <td><code>tlsSettings</code></td>
-<td><code><a href="/docs/reference/config/istio.networking.v1alpha3.html#TLSSettings">istio.networking.v1alpha3.TLSSettings</a></code></td>
+<td><code><a href="/docs/reference/config/networking/v1alpha3/virtual-service.html#TLSSettings">istio.networking.v1alpha3.TLSSettings</a></code></td>
```
frankbu: This is a generated link that the generator is messing up. TLSSettings is in destination-rule.html, but the generator seems to have arbitrarily picked virtual-service.html instead.

@geeknoid Do you know if there is a way to annotate or otherwise fix the generator here?

geeknoid: This is exactly what I was warning @rshriram about links between these protos. He keeps just dismissing it with "we don't have links across protos".

Support for "one HTML file per proto" was added to the proto extraction tool by someone at Google that needed that feature for another project. I was never convinced the change would deal with cross-proto links correctly, and here's evidence.

I'll take a look at the tool to see if this is an easy fix.

geeknoid: @rshriram 

I've updated the proto doc extraction tool in istio/tools and have submitted https://github.com/istio/api/pull/812 to start using there. If the original PR in istio/api is retried using this latest version of the tool, it should fix these types of link problem.

scripts/grab_reference_docs.sh
```diff
@@ -12,7 +12,7 @@
 # The repos to mine for docs, just add new entries here to pull in more repos.
 REPOS=(
     https://github.com/istio/istio.git@release-1.1
-    https://github.com/istio/api.git@release-1.1
+    https://github.com/rshriram/api.git@sep_html
```
geeknoid: Now that your PR is in istio/api, please revert this change.

PR #3367, opened 2019-02-24
Separate out the uninstall directions

sdake:
Also fix the 58->56.  Looking at the istio-init chart, only 56
"CustomResourceDefinition" objects are present.  I am not sure where
this number 57 or 58 originates.

### DIFF
--- a/content/docs/setup/kubernetes/helm-install/index.md
+++ b/content/docs/setup/kubernetes/helm-install/index.md
@@ -1,4 +1,3 @@
-
 ---
 title: Installation with Helm
 description: Install Istio with the included Helm chart.
@@ -11,8 +10,8 @@ icon: helm
 ---
 Quick start instructions for the setup and configuration of Istio using Helm.
 This is the recommended install method for installing Istio to your
-production environment as it offers rich customization to the Istio control
-plane and the sidecars for the Istio data plane.
+production environment as it offers rich customization for both the Istio control
+plane and the sidecars used witin the Istio data plane.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


geeknoid:

Built with commit 8864549c19a1297cdbcc787731d83c48ce8ff796

https://app.netlify.com/sites/preliminary-istio/deploys/5c7295cf714ea500083fc542

ericvn:
@sdake There were originally 56 CRDs and then we saw in one PR from last week's testing day, it was changed to 58. I hadn't seen the 58 up until that time using snapshot 6, but when using RC0 I did see 58, so Frank had me change them all to 58. I did more testing with snapshot 6, RC0 and the current code and 56 is what I am seeing now.  I do see there another PR for the 56/58 issue https://github.com/istio/istio.io/pull/3380 and in that one I do note two other files that have 58 and also need updating.

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -159,11 +138,11 @@ to manage the lifecycle of Istio.
 
     {{< /text >}}
 
-1. Verify that all `58` Istio CRDs were committed to the Kubernetes api-server using the following command:
+1. Verify that all `56` Istio CRDs were committed to the Kubernetes api-server using the following command:
```
rcaballeromx:
```suggestion
1. Verify that all `56` Istio CRDs are committed to the Kubernetes api-server using the following command:
```

content/docs/setup/kubernetes/helm-install/index.md
```diff
@@ -176,14 +155,47 @@ to manage the lifecycle of Istio.
 
 1. Uninstall steps:
 
+
+## Uninstalling Istio
+
+### Option 1: Uninstall with Helm via `helm template`
```
rcaballeromx: Two headings one after the other are the first indicator that there is a problem with the content's structure. I suggest we split the document into two one document without tiller and another with.
However, I also suggest we do it after 1.1. 
@sdake Do you agree?

sdake:
@andraxylia is solving this one.

PR #3368, opened 2019-02-24
Reorder production install text to avoid later issues with Option 4

chris-hamper:
When running through the steps on this page, if a user runs "Installation steps" #1, and then continues to Option 4 when using Helm 2.10.0 or later, an error is encountered due to the Istio CRDs already being installed (see https://github.com/istio/istio/issues/10588), preventing the install from succeeding. Reordering the recommended production install method link will help prevent users from running into this issue, as they will not have already run step #1 before clicking on the link to https://istio.io/docs/setup/kubernetes/helm-install/

### DIFF
--- a/content/docs/setup/kubernetes/quick-start/index.md
+++ b/content/docs/setup/kubernetes/quick-start/index.md
@@ -28,6 +28,12 @@ To install and configure Istio in a Kubernetes cluster, follow these instruction
 
 1. Check the [Requirements for Pods and Services](/docs/setup/kubernetes/spec-requirements/).
 
+## Production install
+
+For a production setup of Istio, we recommend installing with the
+[Helm Chart](/docs/setup/kubernetes/helm-install/), to use all the
+configuration options. This permits customization of Istio to operator specific requirements.
+
 ## Installation steps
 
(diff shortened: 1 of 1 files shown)


frankbu:
This change makes sense, but I think the whole quickstart section of the docs is going to be changed quite a bit, which may fix this problem in a different way.

^^^ @rcaballeromx

content/docs/setup/kubernetes/quick-start/index.md
```diff
@@ -28,6 +28,12 @@ To install and configure Istio in a Kubernetes cluster, follow these instruction
 
 1. Check the [Requirements for Pods and Services](/docs/setup/kubernetes/spec-requirements/).
 
+## Production install
```
rcaballeromx: Please do not add this here. we are already providing additional guidance in a new landing page:
https://docs.google.com/document/d/1b6j7F8ogtC1DAJIK9fSaPPCSAvKaUIgh8t8g2wnln_c/edit?ts=5c758b15
The content of this section conflicts with that effort.

content/docs/setup/kubernetes/quick-start/index.md
```diff
@@ -28,6 +28,12 @@ To install and configure Istio in a Kubernetes cluster, follow these instruction
...
+
+For a production setup of Istio, we recommend installing with the
```
rcaballeromx: If you want to warn users installing on production environments, please use a warning callout: https://preliminary.istio.io/about/contribute/creating-and-editing-pages/#callouts
The warning would be best placed just before the Prerequisites heading.

chris-hamper:
Sounds like this will be resolved by pending changes, so I've closed this PR

PR #3380, opened 2019-02-25
There are only 56 CRDs, not 58

brian-avery:
1.1 is only creating 56 CRDs, not 58. It looks like the missing ones are servicecontrolreports.config.istio.io and servicecontrols.config.istio.io, both of which have been removed. 

adapters.config.istio.io
apikeys.config.istio.io
attributemanifests.config.istio.io
authorizations.config.istio.io
bypasses.config.istio.io
certificates.certmanager.k8s.io
checknothings.config.istio.io
circonuses.config.istio.io
cloudwatches.config.istio.io
clusterissuers.certmanager.k8s.io
clusterrbacconfigs.rbac.istio.io
deniers.config.istio.io
destinationrules.networking.istio.io
dogstatsds.config.istio.io
edges.config.istio.io
envoyfilters.networking.istio.io
fluentds.config.istio.io
gateways.networking.istio.io
handlers.config.istio.io
httpapispecbindings.config.istio.io
httpapispecs.config.istio.io
instances.config.istio.io
issuers.certmanager.k8s.io
kubernetesenvs.config.istio.io
kuberneteses.config.istio.io
listcheckers.config.istio.io
listentries.config.istio.io
logentries.config.istio.io
memquotas.config.istio.io
meshpolicies.authentication.istio.io
metrics.config.istio.io
noops.config.istio.io
opas.config.istio.io
policies.authentication.istio.io
prometheuses.config.istio.io
quotas.config.istio.io
quotaspecbindings.config.istio.io
quotaspecs.config.istio.io
rbacconfigs.rbac.istio.io
rbacs.config.istio.io
redisquotas.config.istio.io
reportnothings.config.istio.io
rules.config.istio.io
serviceentries.networking.istio.io
servicerolebindings.rbac.istio.io
serviceroles.rbac.istio.io
sidecars.networking.istio.io
signalfxs.config.istio.io
solarwindses.config.istio.io
stackdrivers.config.istio.io
statsds.config.istio.io
stdios.config.istio.io
templates.config.istio.io
tracespans.config.istio.io
virtualservices.networking.istio.io
zipkins.config.istio.io

PR #3389, opened 2019-02-26
Update CRD count from 58 to 56

ericvn:


ericvn:
This should cover the changes I requested in 
https://github.com/istio/istio.io/pull/3380.

PR #3408, opened 2019-02-27
[Cluster-Aware] Enable the access log on primary cluster

ymesika:
Enabling the access log because [there is a section](http://localhost:1313/docs/examples/multicluster/split-horizon-eds/#split-horizon-eds-in-action) that verifies accessing local/remote endpoints by looking at the access log.

__Alternatively__, the section can be removed.


geeknoid:
Please rebase, then we can merge.

Thanks.

PR #3414, opened 2019-02-27
Moving the scalability and sizing section to the top

oaktowner:
People keep asking me for this section even though it exists. Since most people are coming this page to find this information, I'm moving it to the top.


geeknoid:
That's not quite right. As it is, the page starts with a title and immediately a section header. I think you should move this section below the first lines of intro text instead of right to the top of the page.

geeknoid:
I'm going to close this for now. I filed #3532 to clean up the whole scalability thing ASAP.

PR #3429, opened 2019-02-28
[WIP] update egress doc for the new default

linsun:


content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -7,11 +7,10 @@ aliases:
...
 ---
 
-By default, Istio-enabled services are unable to access URLs outside of the cluster because the pod uses
-iptables to transparently redirect all outbound traffic to the sidecar proxy,
-which only handles intra-cluster destinations.
+By default, Istio-enabled services are able to access URLs outside of the cluster.  If you choose to tighten up the outbound access control to services running outside of the cluster, you can configure `global.outboundTrafficPolicy.mode` to `REGISTRY_ONLY`.  In this `REGISTRY_ONLY` mode, Istio configures the pod to use iptables to transparently redirect all outbound traffic to the sidecar proxy, which only handles intra-cluster destinations.
+
 This task describes how to configure Istio to expose external services to Istio-enabled clients
-in three different ways:
+in three different ways, when `REGISTRY_ONLY` mode is enabled:
```
vadimeisenbergibm: @linsun So it is two ways. The third ways was to disable `REGISTRY_ONLY`.

vadimeisenbergibm: @linsun I mean there are three ways, one of them is the default, when `REGISTRY_ONLY` mode is disabled.

frankbu: This sentence should not be changed. The task shows 3 different ways period. The fact that the third way is enabled by default is not relevant here.

content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -7,11 +7,10 @@ aliases:
 keywords: [traffic-management,egress]
 ---
 
-By default, Istio-enabled services are unable to access URLs outside of the cluster because the pod uses
-iptables to transparently redirect all outbound traffic to the sidecar proxy,
-which only handles intra-cluster destinations.
+By default, Istio-enabled services are able to access URLs outside of the cluster.  If you choose to tighten up the outbound access control to services running outside of the cluster, you can configure `global.outboundTrafficPolicy.mode` to `REGISTRY_ONLY`.  In this `REGISTRY_ONLY` mode, Istio configures the pod to use iptables to transparently redirect all outbound traffic to the sidecar proxy, which only handles intra-cluster destinations.
```
frankbu:
```suggestion
Depending on your Istio configuration, Istio-enabled services may not be able to access URLs outside of the cluster.
```

content/docs/tasks/traffic-management/egress/index.md
vadimeisenbergibm: OK, I see your suggestion, "
Depending on your Istio configuration, Istio-enabled services may not be able to access URLs outside of the cluster", together it is clear for me.

content/docs/tasks/traffic-management/egress/index.md
linsun: I want to reflect the default config here as it is a change we made in 1.1 across all profiles.

linsun: any suggestions?  or you don't think we should highlight that?

frankbu: I see your point. It's tricky because the point of this task is to talk about all three ways of allowing egress and pointing out pros and cons. The ALLOW_ANY default is just to help people get started, but it's not really recommended, so we need to be careful not to lose the message just because we changed the default profile.

I was thinking that we should update the task as little as possible, just make it "correct", given the change in default. Looking at it more, though, I think that it would be better to change the flow of the entire task (e.g., switch the order of the options, etc.).

How about we close this PR, and I'll take a pass at reworking it?

linsun: That works, thank you @frankbu, pls let me know if you need anything from me.

PR #3461, opened 2019-03-02
race condition fix for mTLS steps

GregHanson:
Race condition exists when the secret is created, and then the Gateway rule in the following step, and when pilot tries to send the LDS update for port 443, it is rejected by the gateway if the secret has not been mounted in ingress gateway yet


geeknoid:
Please rebase so we can merge this change.

Thanks.

geeknoid:
Ping. Please rebase and repush.

Thanks.

geeknoid:
Replaced by #3722

PR #3479, opened 2019-03-03
add a warning about already-existing istio-system

vadimeisenbergibm:
To prevent errors in case `istio-system` already exists.
`istio-system` could exist if additional installation steps are required,
for example to create a secret in istio-system, which will be used by the installed Istio components.
Currently such scenario occurs in setting the "Gateway Connectivity" multicluster, see https://preliminary.istio.io/docs/setup/kubernetes/multicluster/gateways/#deploy-the-istio-control-plane-in-each-cluster.

The pattern of using _kubectl create dry run, piping to kubectl apply_, is described in
https://github.com/kubernetes/kubernetes/issues/65066#issuecomment-397187601


content/docs/setup/kubernetes/install/helm/index.md
```diff
@@ -84,7 +84,7 @@ Choose this option if your cluster doesn't have [Tiller](https://github.com/kube
 
     {{< text bash >}}
 
-    $ kubectl create namespace istio-system
+    $ kubectl create namespace istio-system --dry-run -o yaml | kubectl apply -f -
```
frankbu: Hmm, this is kind of a confusing command. Wouldn't it be better to just change the instruction text to say `create the istio-system namespace if necessary`? If the create command is blindly run, it's just returns an error saying it exists already, which is harmless.

vadimeisenbergibm: Right, it is harmless, though I do not want to confuse the reader. Just one command, no errors, no    "if necessary" reservations. The user does not have to understand the commands, just click and paste them, step by step, as smoothly as possible. In my opinion.

sdake: this operation doesn't look correct.  --dry-run is an input to `helm` and this pattern is typical of those wishing `helm template` had existed prior to its exitstance.  kubectl for example does not take the --dry-run input...

frankbu: I don't really see it that way. I think many people do look at the command and will wonder what it's trying to do, like I did. I started wondering if it even does the right thing all the time. What if the namespace had some labels set? This command will remove them. It's not necessarily a NOOP when the NS already exists.

rcaballeromx: I agree with @frankbu We want to keep things simple for readers but without introducing new sources of confusion. @vadimeisenbergibm Can you help me understand why do you think the conditional "if needed" would not work in this instance?

vadimeisenbergibm: @rcaballeromx It forces the user to think why this namespace is necessary. In the proposed  version, the user just clicks on the steps without thinking. I agree with @frankbu that some users can be confused by the command itself which is less trivial in the proposed version, and the command can remove the labels on the namespace if they existed. So let's agree on the wording that is simplest for the user.

The ideal situation would be if `kubectl` had _create-if-not-exists_ command, so we would run it like _kubect create-if-not-exists namespace istio-system_. 

Such a command does not exist. So we can tell the user, run `kubectl create namespace istio-system`, and if you get an error that such a namespace exists, ignore it.

Alternatively, we can tell the user, run `kubectl create namespace istio-system` if `istio-system` does not exist. We will force the user to figure out if he/she created istio-system before, or to run `kibectl get namespace istio-system` and check if it exists.

Yet another option is to use bash if: 
`if ! kubectl get namespace istio-system 2> /dev/null; then kubectl create namespace istio-system; fi`

rcaballeromx: @vadimeisenbergibm The bash script is not ideal since it complicates the command. Based on the context you provide, we have two options to document the command:

1. Before the command, let readers know they only need to create the namespace if it is not already there.
2. Use a {{< warning >}} after the command with the exact text of the error message and letting readers know they can safely ignore it.

I prefer 2 because you could create the warning once in boilerplates and use it for every occurrence.

vadimeisenbergibm: @rcaballeromx, @frankbu So what do you think now?

linsun:
/hold
looks like there is concern to add this in.

content/docs/setup/kubernetes/install/helm/index.md
vadimeisenbergibm: @rcaballeromx, @frankbu So now, when the Istio 1.1 is out, could you please revisit 
https://github.com/istio/istio.io/pull/3479#discussion_r262530805 (based on 
https://github.com/istio/istio.io/pull/3479#discussion_r262527360)?

PR #3487, opened 2019-03-04
Spelling

loverto:
fix spelling errors in kubernetes install minimal


loverto:
[spelling errors](https://circleci.com/gh/istio/istio.io/9425?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

image

english spelling is normal
image

geeknoid:
The English version is a link, the Chinese version isn't. The spell checker doesn't check the spelling of links.

Please just make the Chinese version use links too.

loverto:
However, my writing is consistent with the English version. In the end, there is no way to change the spelling file. Can you help me?
@geeknoid

loverto:
https://github.com/istio/istio.io/pull/3475/files

fleeto:
If 

> The English version is a link, the Chinese version isn't. The spell checker doesn't check the spelling of links.
> 
> Please just make the Chinese version use links too.

I think @loverto is talking about the anchor of the format `{#*}`. My opinion is to skip the checking for the anchors in the curly braces, and I'd made a fix in #3494

geeknoid:
I think the right solution is to simply rename option-1 and option-2 to something that makes sense in Chinese. Then it wouldn't be a spelling error.

loverto:
[node-markdown-spellcheck](https://github.com/lukeapage/node-markdown-spellcheck#cli-options) itself does not support Chinese spell checking. The current option is invalid.

fleeto:
I'd spent hours on this, I found that mdspell doesn't support  the option `--zh-cn`. 
When we are trying to write the anchor code in a more beautiful and readable way like `## chinese_title {#english_anchor}` in `/content_zh/*.md`, I think the result is unexpected. 

So  may be the old way (just keep `## Chinese_title`) is a better choice, 1.1 is the most important. 

https://github.com/istio/istio.io/pull/3504/commits/dafc7a071b6fbcad52e1e3dadae8715e2f29275a#diff-b669befb45c78b87d6fd3312d173bd7a

scripts/lint_site.sh
```diff
@@ -91,7 +91,7 @@ check_content() {
 }
 
 check_content content --en-us
-check_content content_zh --zh-cn
+check_content content_zh --en-us
```
geeknoid: Was this intentional?

geeknoid:
@fleeto So you're saying that we're getting no Chinese spell checking at the moment? If not, then is it worth running mdspell on the Chinese content at all?

fleeto:
@geeknoid I think its working in a strange way .

~~~
$ cp content/docs/setup/kubernetes/install/minimal/index.md content_zh/docs/setup/kubernetes/install/minimal/inde
x2.md
$ make lint
...
htmlproofer 3.9.2
>> 308 files are free from spelling errors
    content_zh/docs/setup/kubernetes/install/minimal/index2.md
       26 | th Helm via `helm template` {#option-1}
       50 | d Tiller via `helm install` {#option-2}

>> 2 spelling errors found in 320 files
~~~ 
We will see, the same file in different folders, will get different result.

And then 
~~~
$ docker run -it --rm gcr.io/istio-testing/website-builder:2019-03-03 mdspell --zh-cn
error: unknown option `--zh-cn'
$ docker run -it --rm gcr.io/istio-testing/website-builder:2019-03-03 mdspell --en-us
...
~~~

It seems that mdspell don't support that option.

But it can still check the **English typo** in content_zh, with `--no-suggestions --zh-cn` or `--en-us`, so it's still useful now.

geeknoid:
There's definitely something funny happening. In general, if you do "make lint", you'll see that the Chinese content is indeed being spell-checked. So mdspell is generally working. I'm able to run mdspell on the command-line directly or through docker, using the --zh-cn option and it works. But I need to use a full command-line, just passing --zh-cn as you showed does fail. I think you need to pass --zh-cn with some other options for it to work correctly.

geeknoid:
So it's the --report option. If you remove that option from the mdspell command-line, then mdspell complains that --zh-cn is an unknown option. But with --report in place, then things work correctly.

I'm guessing what mdspell is trying to say is that "--zh-cn requires the --report option" but it's printing the wrong error message.

loverto:
@geeknoid 
this is mdspell support [language](https://github.com/lukeapage/node-markdown-spellcheck/blob/master/es5/cli.js#L52)

geeknoid:
Ah, thanks for digging that up. It's too bad then we can't check Chinese :-(

PR #3503, opened 2019-03-05
Restore ingress doc order

JimmyCYJ:
Move the "Securing Ingress Gateway" tasks below "Control Ingress Traffic" task, this is to keep the order of tasks unchanged.


JimmyCYJ:
cc @myidpt

PR #3396, opened 2019-02-26
Fix helm values for demo and uninstall steps

andraxylia:
Doc changes related to istio/istio#11964

PR #3513, opened 2019-03-05
delete the CRD clean up section

linsun:
as it is covered earlier in each of the 2 helm install options.


andraxylia:
Can we merge this after  https://github.com/istio/istio.io/pull/3396 ?

PR #3515, opened 2019-03-05
update instructions for quick start

linsun:
1) I'd prefer to call it quick start so it is consistent with it when folks looking for it.  Also, since it is mainly for quick start and demo purpose :)

2) I think calling mtls enabled is confusing because within istio-values-demo yaml file we actually set it to global.mtls.enabled=false (by reusing what is in the values.yaml).


linsun:
I think there are more changes on the page to position it as a quick start install guide.  I'd like to wait for folks' opinion before proceeding any of the changes.

frankbu:
I've reworked this entire page in #3520 

I'm still thinking about how best to link to it as the "official quick start instructions", but that will be another PR.

linsun:
Thanks @frankbu - closing this because my comment are addressed in your PR

Issue #3521, opened 2019-03-05
"Installing the sidecar" verification step incorrect


esnible:
https://preliminary.istio.io/docs/setup/kubernetes/additional-setup/sidecar-injection/

The step says "Verify that the sidecar has been injected into the deployment."

```
$ kubectl get deployment sleep -o wide
NAME      DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE       CONTAINERS          IMAGES                             SELECTOR
sleep     1         1         1            1           2h        sleep,istio-proxy   tutum/curl,unknown/proxy:unknown   app=sleep
```

I tested with _kubectl_ version 1.13 against a server running 1.13.  Although the injection was correct, the IMAGES and CONTAINERS fields show only one container because the injection happens to the _pod_, not the _deployment_.

This is what I actually see:

```
$ kubectl get deployment sleep -o wide
NAME    READY   UP-TO-DATE   AVAILABLE   AGE   CONTAINERS   IMAGES           SELECTOR
sleep   1/1     1            1           8d    sleep        pstauffer/curl   app=sleep
```

A better way to check might be `kubectl get pod <pod> -o jsonpath='{.metadata.annotations.sidecar\.istio\.io\/status}'`

linsun:
how about this?

```
$ kubectl get pod  -l app=sleep
NAME                     READY   STATUS    RESTARTS   AGE
sleep-64c6f57bc8-f5n4x   2/2     Running   0          24s
```

linsun:
BTW, this returned nothing because I didn't use the annotation to inject sidecar.

```
$ kubectl get pod  -l app=sleep -o jsonpath='{.metadata.annotations.sidecar\.istio\.io\/status}'
(⎈ |linistio9:default)
```

PR #3527, opened 2019-03-06
update validate sidecar cmd

linsun:
fixes: https://github.com/istio/istio.io/issues/3521


esnible:
If we want to see the container names, as in previous versions, this works provided we have a recent version of Kubernetes.
```
kubectl get pods -o=custom-columns=NAME:.metadata.name,CONTAINERS:.spec.containers[*].name
```

We should probably merge what you have because it will work with all versions of Kubernetes.

linsun:
this doesn't work with my k8s:

```
$ kubectl get pod  -l app=sleep  -o=custom-columns=NAME:.metadata.name,CONTAINERS:.spec.containers[*].name
zsh: no matches found: -o=custom-columns=NAME:.metadata.name,CONTAINERS:.spec.containers[*].name
```

PR #3533, opened 2019-03-06
no need to make stateful set red

linsun:


content/docs/setup/kubernetes/install/kubernetes/index.md
```diff
@@ -69,7 +69,7 @@ Choose this variant for:
 * Applications that use
   [liveness and readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/),
 * Headless services, or
-* `StatefulSets`
+* StatefulSets
```
geeknoid: That should actually just be 'stateful sets'

Two words, no caps.

linsun: It is also a k8s concept - see https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

linsun: should I add this to the ignore list?

geeknoid:
No, this should definitely not be in the ignore list.

It's either a programming symbol (`StatefulSet`) or it is a concept
(stateful set). Our docs currently confuse these two notions throughout and
@frankbu was going to do a pass to eliminate the most common cases.

In this context, this should be a 'stateful set' since it is talking about
the concept and not the specific programming synbol.

linsun:
I just feel it is odd it is red. Looking at k8s doc (https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) it is using concept with StatefulSet(s).

geeknoid:
The k8s docs are equally confused about the way the prose should be written :-)

When talking about concepts, we should be using "a virtual service routes traffic to a destination rule, which identifies a Kubernetes service". When talking about symbols, we should be using "create a `VirtualService` resource with a reference to an existing `DestinationRule` resource."

content/docs/setup/kubernetes/install/kubernetes/index.md
rcaballeromx: Since it is a K8s object/concept, then it should be in back-ticks. That is what that markup indicate: https://istio.io/about/contribute/style-guide/#use-code-style-for-object-field-names

linsun: reading this -- https://istio.io/about/contribute/style-guide/#address-the-reader-as-you

I don't think this is common to highlight in red for `Deployment` in this example in k8s docs.

rcaballeromx:
@linsun I agree with @geeknoid The K8s docs use the markup inconsistently. This is an issue we have been tackling.

PR #3537, opened 2019-03-06
Fix the Istio CNI doc per the rel-1.1 helm chart decoupling changes

tiswanso:
There is now a 2 step procedure to install Istio CNI and Istio separately, rather than the prior single/coupled helm install.

### DIFF
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -47,15 +47,13 @@ replaces the functionality provided by the `istio-init` container.
     1.  Refer to the [Hosted Kubernetes Usage](#hosted-kubernetes-usage) section for any non-default
         settings required.
 
-1.  Update Helm's local package cache with the instructions found in the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/). 
+1.  Update Helm's local package cache with the instructions found in the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/).
     To install using the `helm template`, fetch the `istio-cni` template in addition
     to the other Istio templates you fetched when following [option 1 of the Istio Helm installation procedure](/docs/setup/kubernetes/install/helm/#option-1-install-with-helm-via-helm-template).
 
-
-
-        {{< text bash >}}
-        $ helm fetch istio.io/istio-cni --untar --untardir $HOME/istio-fetch
-        {{< /text >}}
+    {{< text bash >}}
+    $ helm fetch istio.io/istio-cni --untar --untardir $HOME/istio-fetch
+    {{< /text >}}
 
 1.  Create a namespace for the Istio control-plane components:
 
@@ -65,13 +63,13 @@ replaces the functionality provided by the `istio-init` container.
 
 1.  Install Istio CNI.
 
-    - Option 1.  Render and apply the Istio CNI Helm template.
+    *  Option 1.  Render and apply the Istio CNI Helm template.
 
         {{< text bash >}}
         $ helm template $HOME/istio-fetch/istio-cni --name=istio-cni --namespace=istio-system | kubectl apply -f -
         {{< /text >}}
 
-    - Option 2.  Install via Helm and Tiller via `helm install`.
+    *  Option 2.  Install via Helm and Tiller via `helm install`.
 
         {{< text bash >}}
         $ helm install istio.io/istio-cni --name=istio-cni --namespace=istio-system
@@ -155,10 +153,9 @@ of many common Kubernetes environments.
     `--set cniBinDir=/home/kubernetes/bin`. For example, the following `helm install`
     command sets the `cniBinDir` value for GKE setups:
 
-
-        {{< text bash >}}
-        $ helm install istio.io/istio-cni --name=istio-cni --namespace=istio-system --set cniBinDir=/home/kubernetes/bin
-        {{< /text >}}
+    {{< text bash >}}
+    $ helm install istio.io/istio-cni --name=istio-cni --namespace=istio-system --set cniBinDir=/home/kubernetes/bin
+    {{< /text >}}
 
 ## Sidecar injection compatibility
 
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -47,10 +47,11 @@ replaces the functionality provided by the `istio-init` container.
     1.  Refer to the [Hosted Kubernetes Usage](#hosted-kubernetes-usage) section for any non-default
         settings required.
 
-1.  Update Helm's local package cache per the instructions in the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/).
+1.  Update Helm's local package cache with the instructions found in the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/). 
+    To install using the `helm template`, fetch the `istio-cni` template in addition
+    to the other Istio templates you fetched when following [option 1 of the Istio Helm installation procedure](/docs/setup/kubernetes/install/helm/#option-1-install-with-helm-via-helm-template).
 
-    1.  If installing via `helm template` fetch the `istio-cni` template in addition to the other Istio
-        templates fetched per the [Istio Helm installation procedure](/docs/setup/kubernetes/install/helm/#option-1-install-with-helm-via-helm-template)
+
 
         {{< text bash >}}
         $ helm fetch istio.io/istio-cni --untar --untardir $HOME/istio-fetch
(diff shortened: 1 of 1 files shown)


tiswanso:
@john-a-joyce  Please review

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -45,30 +45,73 @@ replaces the functionality provided by the `istio-init` container.
     1.  Refer to the [Hosted Kubernetes Usage](#hosted-kubernetes-usage) section for any non-default
         settings required.
 
-1.  Add the `--set istio_cni.enabled=true` setting to the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/) to include the installation of the Istio CNI plugin. For example:
+1.  Update Helm's local package cache per the instructions in the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/).
+
+    1.  If installing via `helm template` fetch the `istio-cni` template in addition to the other Istio
+        templates fetched per the [Istio Helm installation procedure](/docs/setup/kubernetes/install/helm/#option-1-install-with-helm-via-helm-template)
+
+        {{< text bash >}}
```
john-a-joyce: I think you need some kind of note here to indicate the CNI charts are pushed to the same locations that the istio charts are pushed to.   As we discussed directly I don't think you want to get in the version quagmire of which specific location to grab from.   For most people it may not be obvious that the CNI charts are pushed to the same locations.

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -45,30 +45,73 @@ replaces the functionality provided by the `istio-init` container.
...
+    {{< /text >}}
+
+1.  Create and apply Istio CNI manifest.  Override the default configuration of the `istio-cni` Helm
+    chart's `logLevel` and `excludeNamespaces` parameters:
+
+    {{< text bash >}}
+    $ helm template $HOME/istio-fetch/istio-cni --name=istio-cni --namespace=istio-system \
     --set istio-cni.logLevel=info \
-    --set istio-cni.excludeNamespaces={"istio-system,foo_ns,bar_ns"} > $HOME/istio.yaml
+    --set istio-cni.excludeNamespaces={"istio-system,foo_ns,bar_ns"} > $HOME/istio-cni.yaml
```
john-a-joyce: Do you want to keep the "istio-cni" part of "istio-cni.excludeNamespaces".  I am not sure if it breaks including it,  but it shouldn't be required.

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -105,8 +148,8 @@ of many common Kubernetes environments.
     enable [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) in your cluster.
     *  Note for existing clusters this redeploys the nodes.
 
-1.  Install Istio via Helm including these options
-    `--set istio_cni.enabled=true --set istio-cni.cniBinDir=/home/kubernetes/bin`
+1.  Install Istio CNI via Helm including these options
```
john-a-joyce: You lost the "istio_cni.enabled=true" part of the instructions.  Maybe reword to say "Install Istio CNI via Helm as described above, but also with this option."  Note also use of "these options" in original text is not proper.

content/docs/setup/kubernetes/install/cni/index.md
tiswanso: IMHO, since this is user facing doc they shouldn't care about those release & src repo details.  Doing the `helm repo add istio.io ...` instructions always makes these CNI instructions work.  So, I'm wary to add more complicated info here.

content/docs/setup/kubernetes/install/cni/index.md
tiswanso: you're right ... will fix

content/docs/setup/kubernetes/install/cni/index.md
tiswanso: I removed the `istio_cni.enabled` since the below refers to the istio-cni param.  Maybe I will add an example it like this:

```
$ helm install istio.io/istio-cni --name=istio-cni --namespace=istio-system --set cniBinDir=/home/kubernetes/bin
```

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -45,30 +45,73 @@ replaces the functionality provided by the `istio-init` container.
     1.  Refer to the [Hosted Kubernetes Usage](#hosted-kubernetes-usage) section for any non-default
         settings required.
 
-1.  Add the `--set istio_cni.enabled=true` setting to the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/) to include the installation of the Istio CNI plugin. For example:
+1.  Update Helm's local package cache per the instructions in the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/).
```
rcaballeromx:
```suggestion
1.  Update Helm's local package cache with the instructions found in the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/). 
    To install using the `helm template`, fetch the `istio-cni` template in addition
    to the other Istio templates you fetched when following [option 1 of the Istio Helm installation procedure](/docs/setup/kubernetes/install/helm/#option-1-install-with-helm-via-helm-template).

```

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -45,30 +45,73 @@ replaces the functionality provided by the `istio-init` container.
...
+
+    1.  If installing via `helm template` fetch the `istio-cni` template in addition to the other Istio
```
rcaballeromx: To help simplify things for readers, I suggest you add this content to the main step. I've left a suggestion that does this for you to be able to quickly implement it. I'm also removing the lines that then become redundant.
```suggestion
```

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -45,30 +45,73 @@ replaces the functionality provided by the `istio-init` container.
...
+        templates fetched per the [Istio Helm installation procedure](/docs/setup/kubernetes/install/helm/#option-1-install-with-helm-via-helm-template)
```
rcaballeromx:
```suggestion
```

content/docs/setup/kubernetes/install/cni/index.md
rcaballeromx: @john-a-joyce We can add the details of the implementation in a separate document at a later date. Adding the details here could confuse and over-complicate the procedure.

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -45,30 +45,73 @@ replaces the functionality provided by the `istio-init` container.
...
 
     {{< text bash >}}
-    $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
-      --set istio_cni.enabled=true > $HOME/istio.yaml
+    $ kubectl create namespace istio-system
     {{< /text >}}
 
+1.  Install Istio CNI.
+
+    1.  Option 1.  Render and apply the Istio CNI Helm template.
```
rcaballeromx: Do not use numbered lists for alternative options. The numbers tell readers that they should run one command after the other and that is not the case here. 

```suggestion
    - Option 1.  Render and apply the Istio CNI Helm template.
```

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -45,30 +45,73 @@ replaces the functionality provided by the `istio-init` container.
...
+
+        {{< text bash >}}
+        $ helm template $HOME/istio-fetch/istio-cni --name=istio-cni --namespace=istio-system | kubectl apply -f -
+        {{< /text >}}
+
+    1.  Option 2.  Install via Helm and Tiller via `helm install`.
```
rcaballeromx:
```suggestion
    - Option 2.  Install via Helm and Tiller via `helm install`.
```

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -105,8 +148,14 @@ of many common Kubernetes environments.
     enable [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) in your cluster.
     *  Note for existing clusters this redeploys the nodes.
 
-1.  Install Istio via Helm including these options
-    `--set istio_cni.enabled=true --set istio-cni.cniBinDir=/home/kubernetes/bin`
+1.  Install Istio CNI via Helm including this option
+    `--set cniBinDir=/home/kubernetes/bin`
```
rcaballeromx:
```suggestion
    `--set cniBinDir=/home/kubernetes/bin`. For example, the following `helm install`
    command sets the `cniBinDir` value for GKE setups:
```

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -105,8 +148,14 @@ of many common Kubernetes environments.
...
+
+    1.  For example, the following `helm install` command sets the `cniBinDir` value for GKE setups:
```
rcaballeromx:
```suggestion
```

content/docs/setup/kubernetes/install/cni/index.md
```diff
@@ -105,8 +149,14 @@ of many common Kubernetes environments.
     enable [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) in your cluster.
     *  Note for existing clusters this redeploys the nodes.
 
-1.  Install Istio via Helm including these options
-    `--set istio_cni.enabled=true --set istio-cni.cniBinDir=/home/kubernetes/bin`
+1.  Install Istio CNI via Helm including this option
+    `--set cniBinDir=/home/kubernetes/bin`. For example, the following `helm install`
+    command sets the `cniBinDir` value for GKE setups:
+
+
```
tiswanso: need to fix up the spacing and newlines here... it rendered funny

tiswanso:
@rcaballeromx -- I took all your suggestions basically as is and rebased and fix lint issues.  Should be good to go, IMO.

PR #3543, opened 2019-03-07
Remove a the specified headers

idouba:
Remove  the specified headers


geeknoid:
This is a generated file. Please fix the original source in the istio/api repo.

Thanks.

PR #3551, opened 2019-03-07
redirect should be part of HTTPRoute

idouba:


fleeto:
These html files is generated from the source code.

https://github.com/istio/api/blob/5f7cfed8429f257f0a0fce4d0b493df8fd914bd6/networking/v1alpha3/virtual_service.proto#L817

geeknoid:
This is an auto-generated file. Please fix the original source in the istio/api repo.

Thanks.

PR #3584, opened 2019-03-08
forgot the other helm path

linsun:
for #3582


linsun:
chatted with @frankbu offline - we may need to revert this.  our helm install instruction is confusing because https://github.com/istio/istio.io/issues/3587

PR #3598, opened 2019-03-09
Using the term canary rollouts consistency

johnma14:
Replace canary deployments, releases with consistent term canary rollouts.


rshriram:
curious: why replace deployments with rollouts? I had initially chosen the term deployments carefully to emphasize that this is not an inplace replacement of compute units (pods) as the term rollouts implies in k8s. But rather, an additional set of compute units, and then route traffic.

PR #3611, opened 2019-03-10
Update uninstall step for helm.

morvencao:
The current uninstall step is not working since all the pods and clusterrole and clusterrolebing... will not be removed.

### DIFF
--- a/content/docs/setup/kubernetes/install/helm/index.md
+++ b/content/docs/setup/kubernetes/install/helm/index.md
@@ -248,9 +248,38 @@ $ helm install istio.io/istio --name istio --namespace istio-system \
 
 1. If you installed Istio with the `helm template`, uninstall with this command:
 
-    {{< text bash >}}
-    $ kubectl delete namespace istio-system
-    {{< /text >}}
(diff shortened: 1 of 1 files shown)


geeknoid:
Shoot, I apparently broke the tab logic somewhere along the way. I'll push a fix later today. Sorry about that.

geeknoid:
OK, I pushed a fix to the tabs. Please rebase and repush so that the preview will look good.

Thanks.

content/docs/setup/kubernetes/install/helm/index.md
```diff
@@ -246,11 +246,40 @@ $ helm install istio.io/istio --name istio --namespace istio-system \
...
 
-    {{< text bash >}}
-    $ kubectl delete namespace istio-system
-    {{< /text >}}
+{{< tabset cookie-name="helm_profile" >}}
+
+{{% tab name="default" cookie-value="default" %}}
+
+{{< text bash >}}
+$ helm template $HOME/istio-fetch/istio --name istio --namespace istio-system | kubectl delete -f -
```
linsun: I agree, it is better to run this cmd then delete the ns.

PR #3617, opened 2019-03-11
Highlight apply default destination rules for bookinfo.

gyliu513:
Some people often forget to apply default destination rules for bookinfo
wheh running this example, it is better to highlight it here.


gyliu513:
Already fixed in master at https://preliminary.istio.io/docs/tasks/traffic-management/request-routing/#apply-a-virtual-service

PR #3626, opened 2019-03-11
Shuffle the setup directory a bit.

geeknoid:
I'm expecting some broken links I'll need to go fix, but you can see the general layout in the meantime.


frankbu:
This looks really good to me. Additionally, I would suggest moving `Additional Setup Guides` under `Install`, since they are all install related.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -23,7 +23,7 @@ networking functionality but without requiring Istio users to enable elevated
 Kubernetes RBAC permissions.
 
 The Istio CNI plugin performs the Istio mesh pod traffic redirection in the Kubernetes pod lifecycle's network
-setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/setup/kubernetes/requirements//)
+setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/setup/kubernetes/prepare/requirements//)
```
frankbu:
```suggestion
setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/setup/kubernetes/prepare/requirements/)
```

rcaballeromx:
> This looks really good to me.

Agreed

> Additionally, I would suggest moving `Additional Setup Guides` under `Install`, since they are all install related.

IMO they are setup related but not necessarily install related. Readers might need to expand the mesh or manually inject sidecars long after they completed the installation, for example.
I would leave them as they are in this PR.

frankbu:
> IMO they are setup related but not necessarily install related. Readers might need to expand the mesh or manually inject sidecars long after they completed the installation, for example.
I would leave them as they are in this PR.

If we leave it, at least change the title to something short and sweet, like the rest of the sections, maybe:

* Prepare
* Download
* Install
* Upgrade
* More Guides

geeknoid:
OK, all bad links fixed.

Should I check in?

Issue #3618, opened 2019-03-11
A few invalid values in Installation Options table


rvennam:
https://preliminary.istio.io/docs/reference/config/installation-options/

![image](https://user-images.githubusercontent.com/5502967/54097082-efd57a80-4384-11e9-8461-57a4551a6e39.png)

This table is generated using scripts/tablegen.py  Looks like it's struggling with values with comments on the same line:
https://github.com/istio/istio/blob/release-1.1/install/kubernetes/helm/istio/values.yaml#L272
![image](https://user-images.githubusercontent.com/5502967/54097190-52c71180-4385-11e9-8a85-c1a7238aa623.png)

rvennam:
cc @johnma14

rvennam:
Thanks for the fix @johnma14! Closing.

PR #3627, opened 2019-03-11
Escape pipe symbol if it appears in the configuration options

johnma14:
Fix for invalid values in Installation Options table (#3618)

### DIFF
--- a/content/docs/reference/config/installation-options/index.md
+++ b/content/docs/reference/config/installation-options/index.md
@@ -174,7 +174,7 @@ To customize Istio install using Helm, use the `--set <key>=<value>` option in H
 | `global.enableTracing` | `true` |  |
 | `global.tracer.lightstep.address` | `""                # example: lightstep-satellite:443` |  |
 | `global.tracer.lightstep.accessToken` | `""            # example: abcdefg1234567` |  |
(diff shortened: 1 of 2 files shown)


rvennam:
In the future, it might be worth looking into moving everything after the # into the description field

```
| `global.tracer.lightstep.secure` | `true` | example: true\|false |
```

frankbu:
Not to mention `true|false` isn't really an example. Just `false` would make more sense.

johnma14:
@rvennam there is already an issue open to do this: https://github.com/istio/istio.io/issues/2566. And I really need to push that PR out this week itself.

johnma14:
@frankbu I agree. Do you think it makes more sense to make the change in the value.yaml under istio/kubernetes/helm directory?

frankbu:
Yes, it should be changed back in the source. Now that you mention it, your `\|` fix will get wiped out next time the doc is regenerated anyway.

johnma14:
@frankbu I actually made the change in the script (tablegen.py) to escape the pipe symbol if it appears in the value of a configuration option.  It is the updated script that generated that output you see in index.md. That way, in case this happens in other yaml files, this fix should help them as well.

PR #3628, opened 2019-03-11
Fix bugs in multicluster gateway Istio installation

lei-tang:
This PR fixes the issue: https://github.com/istio/istio/issues/12405

### DIFF
--- a/content/docs/setup/kubernetes/install/multicluster/gateways/index.md
+++ b/content/docs/setup/kubernetes/install/multicluster/gateways/index.md
@@ -62,8 +62,9 @@ on **each** Kubernetes cluster.
     {{< /warning >}}
 
     {{< text bash >}}
+    $ cat install/kubernetes/helm/istio-init/files/crd-* > $HOME/istio.yaml
     $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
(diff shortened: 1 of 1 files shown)


content/docs/setup/kubernetes/install/multicluster/gateways/index.md
```diff
@@ -75,6 +76,7 @@ on **each** Kubernetes cluster.
     * Create a Kubernetes secret for your generated CA certificates using a command similar to the following. See [Certificate Authority (CA) certificates](/docs/tasks/security/plugin-ca-cert/#plugging-in-the-existing-certificate-and-key) for more details.
 
         {{< text bash >}}
+        $ kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user="$(gcloud config get-value core/account)"
```
frankbu: @lei-tang this can't be added here because it is using a google-specific command (`gcloud config ...`). If it is only needed for GCP, then there should be a warning icon telling GCP users to run this extra command. These instructions need to be platform-independent.

lei-tang: https://github.com/istio/istio.io/pull/3959.

PR #3635, opened 2019-03-11
Remove outdated intelligent routing example

douglas-reid:
The "example" is just a curated set of links against existing tasks and does little in the way of adding any information on top of those tasks. The content definitely isn't an "example" (it used to be a "guide") and doesn't fit the rest of the content in that heading.


sdake:
@douglas-reid can you please rebase.

Thanks
-steve

geeknoid:
Being replaced by #3721

Issue #3170, opened 2019-01-30
Installing the sidecar doc refer to an yaml file which is not exist


rootsongjc:
This file doesn't exist `install/kubernetes/istio-sidecar-injector-configmap-release.yaml`but be refered on https://istio.io/docs/setup/kubernetes/sidecar-injection/ doc.

geeknoid:
This is a P0 blocking the 1.1 release. When you believe this issue can be addressed? If you don't have time, please assign to someone else that can address this issue.

Thanks.

rootsongjc:
@geeknoid Could you assign it to someone else, I am a little busy nowadays.

sdake:
assign @howardjohn

PR #3654, opened 2019-03-12
Update sidecar-injector setup

howardjohn:
Fixes https://github.com/istio/istio.io/issues/3170

The file moved and is now double templated (helm first, then the code here) so I think it is too complicated to include the real file, but adding the actual file doesn't add much value anyways - it is just explaining how it works


sdake:
@howardjohn I agree its strange to document how this works, however, some community members have written their own injectors...

Cheers
-steve

PR #3662, opened 2019-03-13
Rework IBM Cloud Public instructions to call common instructions

ericvn:


ericvn:
@linsun  FYI, I did slack with Long and got back: 
```Yes, seems that the steps in `public` piece is same to the common istio installation steps. I’m OK with that.

For `private` piece, we need to update the step once we have `ibm-istio` chart published to `catalog`.
```

frankbu:
If the generic instructions are all your pointing at, then I would not even have this section. Instead, just add a sentence at the top that says for IBM Cloud Public, just follow the Istio instructions and point to the Kubernetes getting started index.

That said, I think the IBM platform setup section there needs to be updated to remove the free cluster stuff, and add the necessary warnings and such that you just did in your previous PR.

ericvn:
@frankbu I further simplified the `IBM Cloud Public` section to a single line. I did leave the section since we have links for the 3 ways to use with IBM Cloud in the first paragraph.

A rebase should have fixed up the IKS free changes I did in the other PR.

One other note is that in the simplification, I removed a tip:

```
 {{< tip >}}
	    The Istio `demo` profile (`install/kubernetes/helm/istio/values-istio-demo.yaml`) is specified in the following command to support smaller machine-typed IBM Cloud Kubernetes Service clusters. If using a cluster of sufficient size, you can remove the `--values` parameter which will use the default Istio configuration values instead. See Istio's built-in [configuration profiles](/docs/setup/kubernetes/additional-setup/config-profiles/) for a description of the available profiles.
	    {{< /tip >}}
```

I think this is OK as the wording applies to all platforms and it is noted on the Profiles page that the **demo** profile is for modest resources. I added the _tip_ when we needed to use the **demo** profile to make things work on _free_ clusters and they are no longer supported.

content/docs/setup/kubernetes/install/platform/ibm/index.md
```diff
@@ -22,131 +22,7 @@ To install the managed Istio add-on in IBM Cloud Public, see the [IBM Cloud Kube
...
-1. If desired, run the following command to delete all CRDs:
-
-    {{< warning >}}
-    Deleting CRDs deletes any configuration changes that you have made to Istio.
-    {{< /warning >}}
-
-    {{< text bash >}}
-    $ for i in install/kubernetes/helm/istio-init/files/crd*yaml; do kubectl delete -f $i; done
-    {{< /text >}}
+To use Helm with IBM Cloud Public, follow the instructions for [install/uninstall](/docs/setup/kubernetes/install/helm/) and [upgrading](/docs/setup/kubernetes/upgrade).
```
frankbu:
```suggestion
Follow [these instructions](/docs/setup/kubernetes/install/helm/) to install and run the current release version of Istio in [IBM Cloud Public](https://www.ibm.com/cloud/) using Helm and the IBM Cloud Kubernetes Service (IKS).

To upgrade Istio in an existing IKS cluster, follow the [upgrade instructions](/docs/setup/kubernetes/upgrade) instead.
```

PR #3677, opened 2019-03-14
Site improvements

geeknoid:
- Added call-to-action buttons in the 1.1 announcements. We can use the same
buttons in future release announcements.
    
- Fixed broken large warning icon on the 404 error page.
    
- Fix order of blog posts in the side bar.
    
- Add support to not expand @@ notations in text blocks and use
it for a text blocks that's currently triggering a build failure
    
- Fixed broken rendering of some text blocks when syntax coloring
is turned off via the menu.


oaktowner:
OMG you're really doing it.

oaktowner:
OMG you're really doing it.

geeknoid:
Closed by accident...

PR #3681, opened 2019-03-14
zh: correct a Chinese translation error

javafuns:
If I understand correctly, the translation for this state is not correct and should be re-translated:

> admissionregistration.k8s.io/v1beta1#MutatingWebhookConfiguration configures when the webhook is invoked by Kubernetes.


SataQiu:
@javafuns Thanks for your contribution! But it looks like you need to sign the CLA first.

javafuns:
> I signed it!

PR #3682, opened 2019-03-14
Site improvements

geeknoid:
- Add a test suite for various compositions of site features.

- Substantially improve composition of various site features to eliminate bad
interactions. It's not perfect yet, but it's much better. The one thing still
broken is "a text block inside a warning in a list in a tab". There's still
something funny going on there, but I have some ideas for later.

- Cleaned up boilerplate logic. This removes broken logic. Boilerplates
now need to stsrt with empty front-matter. This addresses the problem of
having a single warning block in a boilerplate not working correctly.

- Remove extraneous space at top and bottom of multi-paragraph callouts.
There is still some oddball spacing going on in tabs. I'll get to that soon.


geeknoid:
Still not ready...

PR #3689, opened 2019-03-14
Use helm charts from downloaded release

frankbu:
This changes the helm install instructions to use the charts in the release image, instead of fetching them from the helm repo. This is better for the following reasons:

* Aligns these instructions with the rest of the setup instructions, which tell users to download the release first, and then proceed to the helm install instructions. If we use the helm repo, then that download wasn't even necessary.
* Since most of the setup instructions are targeted at new users, they will need to download the release anyway to access `istioctl` and the samples, so we might as well use the charts that are already downloaded.
* Makes the helm instructions work with any downloaded release/build. Just run them in the appropriate root directory.

From discussion with @sdake I think we should add additional instructions targetted at production users that don't need access to the samples and would benefit from installing with the helm repo with no release download necessary. But, I think this is part of a bigger doc reorg to provide separate instructions for SREs/operators vs. new users, but I think it can wait until after 1.1 is out.

### DIFF
--- a/content/blog/2018/incremental-traffic-management/index.md
+++ b/content/blog/2018/incremental-traffic-management/index.md
@@ -56,7 +56,7 @@ $ kubectl create clusterrolebinding cluster-admin-binding \
    --user=$(gcloud config get-value core/account)
 {{< /text >}}
 
-Next, [install Helm](https://docs.helm.sh/using_helm/) and [generate a minimal Istio install](/docs/setup/kubernetes/install/helm/) -- only traffic management components:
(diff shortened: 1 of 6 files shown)


sdake:
```From discussion with @sdake I think we should add additional instructions targetted at production users that don't need access to the samples and would benefit from installing with the helm repo with no release download necessary. But, I think this is part of a bigger doc reorg to provide separate instructions for SREs/operators vs. new users, but I think it can wait until after 1.1 is out.```

works for me, as long as we are willing to backport the basics of these new improvements rather soon after 1.1 release.

/lgtm
/approve

Issue #3671, opened 2019-03-13
Document how to re-enable istio-policy in docs (tasks)


douglas-reid:
With disabled-by-default, we should add clear instructions on enable policy in the appropriate locations (and possibly in a FAQ).

Issue #3670, opened 2019-03-13
Introduce in-process adapter model depracation notices


geeknoid:
We need to advertise that the in-process adapter model is being deprecated in 1.1 and will be completely removed in 1.2. So:

- Put advisory on the in-process adapter page wiki page.

- Put a bullet in the 1.1 release notes

- Add an entry in the feature status page.

- Post a note on discuss.istio.io

douglas-reid:
I've posted a note, updated the wiki, and sent a PR for the other two action items.

douglas-reid:
Since that PR has been merged I will close this issue.

PR #3694, opened 2019-03-14
Add doc on policy check enablement and clean-up P&T release notes

douglas-reid:
Also, includes deprecation notice for in-proc adapters in release notes (#3670).

Fixes: #3671

### DIFF
--- a/content/docs/tasks/policy-enforcement/enabling-policy/index.md
+++ b/content/docs/tasks/policy-enforcement/enabling-policy/index.md
@@ -12,7 +12,7 @@ This task shows you how to enable Istio policy enforcement.
 In the default Istio installation profile, policy enforcement is disabled. To install Istio
 with policy enforcement on, use the `--set global.disablePolicyChecks=false` Helm install option.
 
-Alternatively, you may [install Istio using the demo profile](https://preliminary.istio.io/docs/setup/kubernetes/install/kubernetes/),
+Alternatively, you may [install Istio using the demo profile](/docs/setup/kubernetes/install/kubernetes/),
 which enables policy checks by default.
 
 ## For an existing Istio mesh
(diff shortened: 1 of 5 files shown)


content/docs/tasks/policy-enforcement/enabling-policy/index.md
```diff
@@ -0,0 +1,48 @@
...
+---
+
+This task shows you how to enable Istio policy enforcement.
+
+## At install time
+
+In the default Istio installation profile, policy enforcement is disabled. To install Istio
+with policy enforcement on, use the `--set global.disablePolicyChecks=false` Helm install option.
+
+Alternatively, you may [install Istio using the demo profile](https://preliminary.istio.io/docs/setup/kubernetes/install/kubernetes/),
```
geeknoid: Remove "https://preliminary.istio.io", shoud just be /docs/...

kyessenov:
Does this really work?
Don't you need to restart pilot (at the very least)?

douglas-reid:
@kyessenov I tested this in a cluster and was able to see `Check()` calls without restarting Pilot.

kyessenov:
Hmm, that didn't work for me. But I guess it might just take time.

douglas-reid:
@kyessenov ok. can you check your test cluster after a minute or two? i was verifying by checking the metrics on `istio-policy` pods directly.  but I don't recall the time between testing the commands to update the config and actually sending traffic.

maybe we need to add a "wait 2 minutes" step?

kyessenov:
We should get to the bottom of this, and figure out if this is even supposed to work. Last time I asked, I was told configmaps don't auto-update in pilot.

kyessenov:
What you might be seeing is that some other update event triggers configmap reload. But it doesn't guarantee that configmaps themselves trigger a reload.

PR #3696, opened 2019-03-14
Site improvements

geeknoid:
- Add a test suite for various compositions of site features.

- Substantially improve composition of various site features to eliminate bad
interactions. It's not perfect yet, but it's much better. The one thing still
broken is "a text block inside a warning in a list in a tab". There's still
something funny going on there, but I have some ideas for later.


geeknoid:
Er, sorry, I didn't ask for a review for this yet since it wasn't actually ready. I found some pages that don't render correctly with these changes :-(

I'll undo this.

PR #3697, opened 2019-03-15
tweaks to release notes

rshriram:


content/about/notes/1.1/index.md
```diff
@@ -56,6 +56,11 @@ to create global namespaces. Locality-aware routing is enabled by default in the
 [Securing Kubernetes Ingress with Cert-Manager](/docs/examples/advanced-gateways/ingress-certmgr/) example for more details on how
 to use Kubernetes Ingress resources with [gateways](/docs/concepts/traffic-management/#gateways).
 
+- **Behavioral changes**.
+    - Outbound traffic policy is set to `ALLOW_ANY` by default. As a result, traffic to unknown ports will be forwarded as-is. Traffic to known ports (e.g., port 80) will be matched with one of the services in the system and forwarded accordingly.
+    - Destination rule resolution order has been set. For a given sidecar, when routing to a particular service, destination rules for the target service in the same namespace as the sidecar will first be considered, followed by destination rules in the service’s namespace, and finally followed by destination rules in `configRootNamespace` if defined or any other namespace.
```
linsun: destination rules for the target service in the same namespace as the sidecar will first be considered
--> destination rules for the target service in the same namespace as the source sidecar will first be considered?

followed by destination rules in the service’s namespace
-> followed by destination rules in the target service’s namespace?

content/about/notes/1.1/index.md
```diff
@@ -56,6 +56,11 @@ to create global namespaces. Locality-aware routing is enabled by default in the
...
+    - We recommend storing gateway resources in the same namespace as the gateway workload. E.g., `istio-system` in case of `istio-ingressgateway`. When referring to gateway resources in virtual services, use the `namespace/name` format instead of the `name.namespace.svc.cluster.local` format.
```
linsun: I'm confused by this, because if I look at our ingress doc, we don't recommend users to deploy the gateway resource in the istio-system ns.  https://preliminary.istio.io/docs/tasks/traffic-management/ingress/#configuring-ingress-using-an-istio-gateway

content/about/notes/1.1/index.md
```diff
@@ -32,7 +32,7 @@ TBD: need full content & links
 
 - **New `Sidecar` Resource**. Added support to limit the set of services visible to sidecar proxies in a given namespace using the `Sidecar` resource.
 This limit reduces the amount of configuration computed and transmitted to the proxy. On large clusters, we recommend adding
-a sidecar object per namespace. TBD LINK: how to add a sidecar per namespace?
+a sidecar object per namespace.
```
geeknoid: object -> resource

content/about/notes/1.1/index.md
```diff
@@ -87,15 +92,6 @@ See [Istio Vault CA Integration](/docs/tasks/security/vault-ca) for more informa
 - TBD: How about adding [11667](https://github.com/istio/istio/issues/11667) as well? As this is also a significant feature for security which can enable end user to set
 different CA and Certs for different namespaces.
 
-## Multicluster
```
geeknoid: Why remove this section?

@sdake @costinm

rshriram: coz I covered it already in my first rework of release notes, under traffic management

content/about/notes/1.1/index.md
rshriram: we should I guess.. so that in future we can make the full switch

content/about/notes/1.1/index.md
frankbu: > and finally followed by destination rules in `configRootNamespace` if defined or any other namespace.

The `or any other namespace` part is confusing. You mean `or any other namespace if configRootNamespace is not set`.

Didn't we make `configRootNamespace` default to `istio-system` anyway, so it would need to be explicitly unset for the any namespace lookup to kick in?

rshriram: nope.. it was not set to istio-system

frankbu: https://github.com/istio/istio/pull/12234

What this reverted?

PR #3706, opened 2019-03-15
Site improvements.

geeknoid:
- Added a test suite to test out the different compositions of
site features.

- Substantially improve the composability of site features
(callouts, tabs, text blocks, boilerplates, lists). You can
now more confidentally mix & match these in any combination
and have a pretty good chance it'll render correctly.


geeknoid:
REVIEWERS: This rewrites some of the core logic to support the various site features.
Please browse the site in general to look for any anomalous formatting. Of particular
interest should be tabs, and lists in tabs, or tabs in lists. Text blocks have been pretty
challenging to get right in all cases too, so keep a lookout for bad text blocks,
especially when they're embedded in tabs/lists/warnings.

Thanks.

PR #3713, opened 2019-03-15
[PoC] Proposal: Add test coverage section to task docs

douglas-reid:
In one of the Code Mauve discussions, it was mentioned that it would be nice to know which sections of the istio.io site were validated by tests -- and known to be working.

I've thrown together this PoC to further discussion as to what this might look like. My original line of thinking was that we could have some sort of test badge, based on individual test links in CI. Moving forward, we could require that all tasks have links to specific tests (and their statuses).

I've used the testgrid.k8s.io embed links to present coverage for the `Policies` tasks of rate limits and denials in this PR.

It looks like:

image

Hopefully, someone with better JS skills and/or understanding of Prow links could just extract the latest status from both the release branch and master.

Would be interested in thoughts on general approach, enhancements, etc.

/hold

### DIFF
--- a/content/docs/tasks/policy-enforcement/denial-and-list/index.md
+++ b/content/docs/tasks/policy-enforcement/denial-and-list/index.md
@@ -289,3 +289,7 @@ subnet.
 * If you are not planning to explore any follow-on tasks, refer to the
   [Bookinfo cleanup](/docs/examples/bookinfo/#cleanup) instructions
   to shutdown the application.
+
+## Test coverage
+
+<iframe src="https://testgrid.k8s.io/istio-release#e2e-mixer-no_auth&include-filter-by-regex=Denials&embed-without-column-labels=" height="125" width="100%"></iframe>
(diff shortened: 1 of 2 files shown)


geeknoid:
I like the concept, but I think it needs to be a bit more subtle.

As a customer, _of course_ you expect the examples being properly tested and working. So showing a bunch of info about this seems redundant to them.

How about if instead of the build stats & such, we just have a DOWNLOAD EXAMPLE button or something like it. If there is such a button, we know that the example is coming from GitHub and thus has a test suite wrapped around it.

rcaballeromx:
I do like the idea of having a test coverage section on every task. Before starting a task, readers need to know what Istio and Kubernetes releases they should have for the instructions to work. In their current state, the tasks don't provide that insight making it a gamble whether or not the instructions will work for a specific release.
The proposed table probably provides that insight if I go and look at the tests. Would that be the case? @douglas-reid Could we expose the more high-level info such as release number instead?

douglas-reid:
@rcaballeromx I think we could expose that information. It would take a bit of work to get the right information, but it should be doable to cover Istio releases and kube versions that have been tested.

douglas-reid:
I'm going to close this issue, as I believe the proposals coming out of the Tiger teams supercede this proposal and will provide better feature coverage observability.

PR #3727, opened 2019-03-17
Improve text blocks.

geeknoid:
- Add tests for the various text block features.

- Combine text_file and text_dynamic into text_import

- Add support for snippet extraction for text downloaded from a URL

- Add support for formatting output blocks in a different style for
content downloaded from a URL

- Report some errors when unable to fetch content from a URL.

- Fixed a few small bugs along the way.


geeknoid:
@sdake I know you can't review the JavaScript and Hugo changes, but if you could browse the site and make sure text blocks look as they should, that'd be appreciated.

geeknoid:
BTW, the intent for this work is to get things fully ready for Code Mauve work, where we want to test all the examples in the site. Now, we'll be able to pull content & snippets from either local files or URLs, all with the same features as we use with inlined content.

PR #3728, opened 2019-03-18
[WIP] update warning for release-1.1

linsun:
missing a link to @johnma14 's helm API diff page.

PR #3732, opened 2019-03-18
Improve upgrade notice.

geeknoid:


content/docs/setup/upgrade-notice/index.md
```diff
@@ -0,0 +1,36 @@
+---
+title: 1.1 Upgrade Notice
+description: Breaking or surprising changes in Istio 1.1.
```
rshriram: stating this as "breaking" changes is kind of bad, given that 1.0 is supposed to be a stable release.. I would state something like "behavioral changes" or something else?

linsun:
btw, i'm also working on https://github.com/istio/istio.io/pull/3728 that highlight warnings.

content/docs/setup/upgrade-notice/index.md
duderino: I personally don't care because I think we are just being transparent, but I will say that this is the third complaint I've heard about using the phrase 'breaking change' so it will probably be an issue for many people

.spelling
```diff
@@ -53,7 +53,6 @@ admissionregistration.k8s.io
 AES-NI
 AKS
 Alibaba
-ALLOW_ANY
```
duderino: for my own education, we don't have to put technical words here if we enclose them in backticks?

content/docs/setup/upgrade-notice/index.md
geeknoid: Word games, right? Behavioral changes is a euphemism for "breaking" IMO. How about:

"Important changes operators must understand before upgrading".

.spelling
geeknoid: Correct. Backticks render as `code blocks`. Symbols are treated as a code blocks through the docs.

geeknoid:
Updated the description.

duderino:
double final super mega /lgtm

Issue #3753, opened 2019-03-19
update CR procedure for backup to include certmanager.


sdake:
1.0 only had certmanager, 1.1 separates certmanager crds from Istio's CRDs at @stono's request.

The current backup procedure only handles istio custom resources.

Issue #3752, opened 2019-03-19
update doc missing usage of ISTIO_RESTORE_DATA.yaml


linsun:
We asked folks to create this file (ISTIO_RESTORE_DATA.yaml) however we never mention how it may be used in case of issues.

The instruction also isn't clear on which scenarios user should be creating this ISTIO_RESTORE_DATA.yaml file.

Is this only applicable when user has used `helm install` previously?

linsun:
@sdake @frankbu cc @howardjohn

howardjohn:
> Restore if necessary:
> 
> $ kubectl apply -f $HOME/ISTIO_RESTORE_DATA.

maybe we should expand on "if necessary" - its necessary if after you install all your CRDs managed to get delete

And it should only be needed by users that use Tiller, but its just a backup of CRDs so anyone could do it

linsun:
there is also missing a space in this cmd. -o yaml
```
kubectl get crds | grep istio.io | cut -f1-1 -d "." | \
    xargs -n1 -i sh -c "kubectl get --all-namespaces -oyaml {}; echo ---" > $HOME/ISTIO_RESTORE_DATA.yaml
```

howardjohn:
`-oyaml` should work

linsun:
ok cool, it did :)

sdake:
@linsun i had an appointment today and mentioned to @frankbu on slack mobile the PR looked good - but mobile github is a suboptimal review experience.

This backup procedure should only apply to helm.

I agree the restore procedure needs some attention.  Perhaps we should have operators create a temporary virtual service prior to upgrading, validate it is present, if it isn't, they know their custom resource data was deleted, ask them to `helm rollback` and follow the restore if the temporary vs isn't present?

This leaves their deployment in an un-upgraded state at which point they should probably hit slack or discuss so we can root cause...

That is the best I've got ;)

Cheers
-steve

frankbu:
I think we may be over analyzing this. Basically the warning message in the upgrade instructions says that we don't think there will be any problems with the upgrade, but back up the CRDs just in case things go bad. It provides both the backup command and restore command, if needed. I think that's good enough for now. If people do, in fact, run into problems, we'll hear from them and we can update the warning when we know more.

linsun:
thanks all for the feedback?

Should we move the warning down to the helm upgrade option?  This can reduce confusion and steps for our users.

PR #3755, opened 2019-03-19
back up certmanager CRDs as well

sdake:
We want to backup certmanager CRDs in the case of Tiller usage.

Also clarify thee filename slightly to indicate it is CRD data and
related to Istio 1.0.

Fixes: #3753


sdake:
Output: https://deploy-preview-3755--preliminary-istio.netlify.com/docs/setup/kubernetes/upgrade/steps/#control-plane-upgrade

sdake:
@howardjohn we are debating how to handle restore here:

https://github.com/istio/istio.io/issues/3752

Issue #3748, opened 2019-03-18
CRD Backup restore fails


howardjohn:
Steps:
Create a CRD (can be whatever):
```
apiVersion: networking.istio.io/v1alpha3
kind: DestinationRule
metadata:
  name: outlier
spec:
  host: "google.com"
  trafficPolicy:
    outlierDetection:
      consecutiveErrors: 7
      interval: 5m
      baseEjectionTime: 15m
```

Run backup script

It will create
```
apiVersion: v1
items:
- apiVersion: networking.istio.io/v1alpha3
  kind: DestinationRule
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"networking.istio.io/v1alpha3","kind":"DestinationRule","metadata":{"annotations":{},"name":"outlier","namespace":"locality"},"spec":{"host":"google.com","trafficPolicy":{"outlierDetection":{"baseEjectionTime":"15m","consecutiveErrors":7,"interval":"5m"}}}}
    creationTimestamp: "2019-03-18T16:17:06Z"
    generation: 1
    name: outlier
    namespace: locality
    resourceVersion: "14208615"
    selfLink: /apis/networking.istio.io/v1alpha3/namespaces/locality/destinationrules/outlier
    uid: 45c7cf1e-4999-11e9-8a7a-42010a80013d
  spec:
    host: google.com
    trafficPolicy:
      outlierDetection:
        baseEjectionTime: 15m
        consecutiveErrors: 7
        interval: 5m
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

Apply this.

Now apply original config and get: `The destinationrules "outlier" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update`

howardjohn:
@sdake

sdake:
@howardjohn I attempted to duplicate.  Here was my workflow:

```
sdake@beast-01:~/istio-1.1.0/install$ kubectl apply -f kubernetes/helm/istio-init/files/
customresourcedefinition.apiextensions.k8s.io/virtualservices.networking.istio.io created
customresourcedefinition.apiextensions.k8s.io/destinationrules.networking.istio.io created
customresourcedefinition.apiextensions.k8s.io/serviceentries.networking.istio.io created
customresourcedefinition.apiextensions.k8s.io/gateways.networking.istio.io created
[...]

PR #3756, opened 2019-03-19
Resolve restore problem

sdake:
Fixes: #3748

Restore was defective in some circumstances.  This backup/restore
procedure isn't very elegant.  It would be possible to use jq with
del() to filter the unneeded data from the backup.  To take minimal
risk, we instead recommend backing up the full custom resource data
set, and then only filter in a restore scenario.;

### DIFF
--- a/content/docs/setup/kubernetes/upgrade/steps/index.md
+++ b/content/docs/setup/kubernetes/upgrade/steps/index.md
@@ -42,7 +42,8 @@ $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | cut -f1-1 -d "." | \
 Restore if necessary:
 
 {{< text bash >}}
-$ kubectl apply -f $HOME/ISTIO_1_0_RESTORE_CRD_DATA.yaml
+$ grep -v selfLink /home/sdake/ISTIO_1_0_RESTORE_CRD_DATA.yaml | grep -v resourceVersion | \
+    grep -v creationTimestamp | grep -v uid | grep -v generation | kubectl apply -f -
 {{< /text >}}
 
 {{< /warning >}}
--- a/content/docs/setup/kubernetes/upgrade/steps/index.md
+++ b/content/docs/setup/kubernetes/upgrade/steps/index.md
@@ -42,7 +42,7 @@ $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | cut -f1-1 -d "." | \
 Restore if necessary:
 
 {{< text bash >}}
-$ grep -v "selfLink|resourceVersion|creationTimestamp|uid|generation" \
+$ grep -v "selfLink\|resourceVersion\|creationTimestamp\|uid\|generation" \
     $HOME/ISTIO_1_0_RESTORE_CRD_DATA.yaml | kubectl apply -f -
 {{< /text >}}
 
--- a/content/docs/setup/kubernetes/upgrade/steps/index.md
+++ b/content/docs/setup/kubernetes/upgrade/steps/index.md
@@ -42,8 +42,8 @@ $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | cut -f1-1 -d "." | \
 Restore if necessary:
 
 {{< text bash >}}
-$ grep -v selfLink /home/sdake/ISTIO_1_0_RESTORE_CRD_DATA.yaml | grep -v resourceVersion | \
-    grep -v creationTimestamp | grep -v uid | grep -v generation | kubectl apply -f -
+$ grep -v "selfLink|resourceVersion|creationTimestamp|uid|generation" \
+    $HOME/ISTIO_1_0_RESTORE_CRD_DATA.yaml | kubectl apply -f -
 {{< /text >}}
 
 {{< /warning >}}


sdake:
https://deploy-preview-3756--preliminary-istio.netlify.com/docs/setup/kubernetes/upgrade/steps/

sdake:
note this restore shell isn't particularly elegant either :)  If anyone has improvements, please submit them.

content/docs/setup/kubernetes/upgrade/steps/index.md
```diff
@@ -42,7 +42,8 @@ $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | cut -f1-1 -d "." | \
 Restore if necessary:
 
 {{< text bash >}}
-$ kubectl apply -f $HOME/ISTIO_1_0_RESTORE_CRD_DATA.yaml
+$ grep -v selfLink /home/sdake/ISTIO_1_0_RESTORE_CRD_DATA.yaml | grep -v resourceVersion | \
```
frankbu:
```suggestion
$ grep -v selfLink $HOME/ISTIO_1_0_RESTORE_CRD_DATA.yaml | grep -v resourceVersion | \
```

frankbu: Actually, can't this just be one grep command?
```
$ grep -v "selfLink|resourceVersion|creationTimestamp|uid|generation" $HOME/ISTIO_1_0_RESTORE_CRD_DATA.yaml | kubectl apply -f -
```
```

Comment (frankbu) [reply content/docs/setup/kubernetes/upgrade/steps/index.md]: Also, are you sure that using `kubectl get --export=true` in the backup command doesn't solve this problem?

Comment (sdake) [reply content/docs/setup/kubernetes/upgrade/steps/index.md]: @frankbu I tried that grep command as one, but it iddn't seem to work.  I will try again.

I also tried --export=true, didn't seem to do the job.  @howardjohn had same results.

Comment (sdake) [reply content/docs/setup/kubernetes/upgrade/steps/index.md]: @frankbu with your grep command the CR on a restore looks like
```
apiVersion: v1
items:
- apiVersion: networking.istio.io/v1alpha3
  kind: DestinationRule
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"networking.istio.io/v1alpha3","kind":"DestinationRule","metadata":{"annotations":{},"name":"outlier","namespace":"default"},"spec":{"host":"google.com","trafficPolicy":{"outlierDetection":{"baseEjectionTime":"15m","consecutiveErrors":7,"interval":"5m"}}}}
    creationTimestamp: "2019-03-19T04:04:14Z"
    generation: 1
    name: outlier
    namespace: default
    resourceVersion: "6623"
    selfLink: /apis/networking.istio.io/v1alpha3/namespaces/default/destinationrules/outlier
    uid: 0f501338-49fc-11e9-83fb-0042688109a5
```

Comment (sdake) [reply content/docs/setup/kubernetes/upgrade/steps/index.md]: this however looks good ```grep -v "selfLink\|resourceVersion\|creationTimestamp\|uid\|generation" $HOME/ISTIO_1_0_RESTORE_CRD_DATA.yaml > zz
```

content/docs/setup/kubernetes/upgrade/steps/index.md
```diff
@@ -42,7 +42,8 @@ $ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | cut -f1-1 -d "." | \
 Restore if necessary:
 
 {{< text bash >}}
-$ kubectl apply -f $HOME/ISTIO_1_0_RESTORE_CRD_DATA.yaml
+$ grep -v "selfLink|resourceVersion|creationTimestamp|uid|generation" \
```
frankbu:
```suggestion
$ grep -v "selfLink\|resourceVersion\|creationTimestamp\|uid\|generation" \
```

howardjohn:
Filtering with grep could removed other fields that the user added if they have the same phrases. The ones we filtering aren't exactly super rare either

sdake:
@howardjohn I tend to agree on the uniqueness problem.  My morning is sucked up by torture in a dentist chair - hence won't be able to address a rework of the backup using `jq` and `del()` until post release.

We have two options:
1. someone else takes over this PR and reimplements the backup/restore and test using `jq`
2. merge as is, and backport a change

frankbu:
Maybe a quick solution to make it safer is to only `grep -v resourceVersion`, and leave the others alone. Have you guys checked if that's sufficient? I think that's the problem field. The others will just get overridden when you apply.

sdake:
Unfortunately `uid` must also be filtered, probably among others:

```
The DestinationRule "outlier" is invalid: metadata.uid: Invalid value: "9b1e53c1-49f2-11e9-83fb-0042688109a5": field is immutable
```

`uid` is the most troubling field in the restore as it could be used by other components.  I can take a look post 1pm if I'm not in a dental drug induced haze at using jq and del for backup, but I am leaving in 20 mins for the dentist.  This could then quickly be backported.

Cheers
-steve

geeknoid:
@sdake @frankbu Last chance before publication. Wanna fix the conflicts and push this in?

12 minutes before I pull the plug...

frankbu:
@geeknoid I pushed #3778 as an interim solution. This one can be finished later.

sdake:
I htink this works well enough, but for the moment closing - we can be a little reactive if someone needs to restore their data rather than giving them a cut and paste command which may crater their system.

Cheers
-steve

PR #3771, opened 2019-03-19
Remove overuse of backticks and geek-speak.

geeknoid:


content/docs/concepts/performance-and-scalability/index.md
```diff
@@ -66,7 +66,7 @@ Data plane performance depends on many factors, for example:
 - Number of client connections
 - Target request rate
 - Request size and Response size
-- Enabling `mTLS`
+- Enabling mutual TLS
```
mandarjog: While you are changing this, can you remove mutual TLS ? It has minimal impact as compared to others on the list

PR #3778, opened 2019-03-19
remove restore command that won't work

frankbu:
For now we won't document how to restore. Hopefully nobody will need to restore, but @sdake and @howardjohn will backport a proper restore command, once they have one.

Related to #3756


sdake:
thanks @frankbu !

PR #3797, opened 2019-03-20
Site improvements

geeknoid:
- Darken the text color of the selected tab in a tab set.

- Remove extra bottom border line in collapsed sidebar panels.

- Cleaned up accessibility annotations for the sidebar.


frankbu:
Doesn't this PR need to be sent to `release-1.1` branch, not `master`?

I think the darker text in the tabs looks better, but wonder if the selected tab can be made to be highlighted, instead of the opposite. Right now, when I look at the tabs (esp when there are only 2), the not-selected one stands out, which intuitively seems like it is the selected one. Could the selected one be colored differently, or maybe darker than all the other ones?

PR #3800, opened 2019-03-20
Update index.md (#3793)

linsun:
cherry pick from master


geeknoid:
I don't understand why the lint check went all bananas here. I'll just push this change in for now.

Issue #3794, opened 2019-03-20
Recent change to Bookinfo example what's next section not acceptable


frankbu:
Part of the change in #3721 is very bad: the change in the "What's next" section. After starting the bookinfo application, what's next needs to point to some tasks that use the bookinfo application. It used to point to the "Inteligent Routing" example which used to walk through a set of tasks that were all using the bookinfo application, but now we are just pointing at other examples that have nothing to do with bookinfo. We need to change this what's next to point to the related tasks.

geeknoid:
#3721 was a redo of #3635, do adding @douglas-reid .

frankbu:
I'll send a PR to the `release-1.1` branch to fix this.

douglas-reid:
Aren't the "What's Next?" sections auto-generated?

douglas-reid:
Oh, nevermind, that's the "See Also" sections in the Tasks.

frankbu:
Fixed by #3806

PR #3806, opened 2019-03-20
Bookinfo example fixes

frankbu:
Fixes: #3794


geeknoid:
The CI failure is benign. Let me know when you're happy with the content and I can push it in.

PR #3808, opened 2019-03-20
fixed: add retryOn field to retries

fleeto:
https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/#HTTPRetry

### DIFF
--- a/content/docs/concepts/traffic-management/index.md
+++ b/content/docs/concepts/traffic-management/index.md
@@ -513,6 +513,7 @@ spec:
     retries:
       attempts: 3
       perTryTimeout: 2s
+      retryOn: gateway-error,connect-failure,refused-stream
 {{< /text >}}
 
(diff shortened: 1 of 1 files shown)


content/docs/concepts/traffic-management/index.md
```diff
@@ -513,6 +513,7 @@ spec:
     retries:
       attempts: 3
       perTryTimeout: 2s
+      retryOn: gateway-error,connect-failure,refused-stream
```
frankbu: What's the default value for retryOn. If it is reasonable, it might be better to leave this field out of the simple example in this concepts section. If it is needed, we should probably at least mention it in the text above the example too.

fleeto: No 5xx isn't in the default  values list of `retryOn`.

PR #3823, opened 2019-03-21
Fix the link of credential distribution

wattli:


geeknoid:
Please push this change to the release-1.1 branch instead.

PR #3829, opened 2019-03-22
Updates for Z shell.

ericvn:
- Support Z Shell
 - Change k get po to not show `Completed` pods to match output
 - Add instruction to wait for `sleep` pod.

### DIFF
--- a/content/docs/examples/multicluster/split-horizon-eds/index.md
+++ b/content/docs/examples/multicluster/split-horizon-eds/index.md
@@ -60,9 +60,9 @@ This will be used to access pilot on `cluster1` securely using the ingress gatew
       --set global.controlPlaneSecurityEnabled=true \
       --set global.proxy.accessLogFile="/dev/stdout" \
       --set global.meshExpansion.enabled=true \
-      --set global.meshNetworks.network2.endpoints[0].fromRegistry=n2-k8s-config \
-      --set global.meshNetworks.network2.gateways[0].address=0.0.0.0 \
-      --set global.meshNetworks.network2.gateways[0].port=443 \
+      --set 'global.meshNetworks.network2.endpoints[0].fromRegistry=n2-k8s-config' \
+      --set 'global.meshNetworks.network2.gateways[0].address=0.0.0.0' \
+      --set 'global.meshNetworks.network2.gateways[0].port=443' \
       install/kubernetes/helm/istio > istio-auth.yaml
     {{< /text >}}
 
@@ -84,7 +84,7 @@ This will be used to access pilot on `cluster1` securely using the ingress gatew
     Wait for the Istio pods on `cluster1` to become ready:
 
     {{< text bash >}}
-    $ kubectl get pods --context=$CTX_CLUSTER1 -n istio-system
+    $ kubectl get pods --context=$CTX_CLUSTER1 -n istio-system --field-selector=status.phase!=Succeeded
     NAME                                      READY     STATUS    RESTARTS   AGE
     istio-citadel-5b9d878756-bwnxx            1/1       Running   0          2m
     istio-galley-6f7594c9f4-7s9db             1/1       Running   0          2m
@@ -173,7 +173,7 @@ This will be used to access pilot on `cluster1` securely using the ingress gatew
     Wait for the Istio pods on `cluster2`, except for `istio-ingressgateway`, to become ready:
 
     {{< text bash >}}
-    $ kubectl get pods --context=$CTX_CLUSTER2 -n istio-system -l istio!=ingressgateway
+    $ kubectl get pods --context=$CTX_CLUSTER2 -n istio-system -l istio!=ingressgateway --field-selector=status.phase!=Succeeded
     NAME                                      READY     STATUS    RESTARTS   AGE
     istio-citadel-958c4b596-kpmj4             1/1       Running   0          40s
     istio-sidecar-injector-77599f75f6-tnj7s   1/1       Running   0          39s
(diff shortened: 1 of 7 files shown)


content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -344,10 +344,17 @@ We will call the `helloworld.sample` service from another in-mesh `sleep` servic
...
+    {{< text bash >}}
+    $ kubectl get po --context=$CTX_CLUSTER1 -n sample -l app=sleep
+    sleep-754684654f-n6bzf           2/2     Running   0          5s
+    {{< /text >}}
+
 1. Call the `helloworld.sample` service several times:
 
     {{< text bash >}}
-    $ kubectl exec --context=$CTX_CLUSTER1 -it -n sample -c sleep $(kubectl get pod --context=$CTX_CLUSTER1 -n sample -l app=sleep -o jsonpath={.items[0].metadata.name}) -- curl helloworld.sample:5000/hello
+    $ kubectl exec --context=$CTX_CLUSTER1 -it -n sample -c sleep $(kubectl get pod --context=$CTX_CLUSTER1 -n sample -l app=sleep -o jsonpath='{.items[0].metadata.name}') -- curl helloworld.sample:5000/hello
```
frankbu: This seems inconsistent with the previous ones. There you put `'` around the whole argument value while here you're just doing the `=value` part of the `-o` argument.

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -84,7 +84,7 @@ This will be used to access pilot on `cluster1` securely using the ingress gatew
     Wait for the Istio pods on `cluster1` to become ready:
 
     {{< text bash >}}
-    $ kubectl get pods --context=$CTX_CLUSTER1 -n istio-system
+    $ kubectl get pods --context=$CTX_CLUSTER1 -n istio-system --field-selector=status.phase!=Succeeded
```
frankbu: I'm wondering if this is really needed here. Isn't it better to let the user see all the pods, whether they've done something during startup and completed, or are up and running?

content/docs/examples/multicluster/split-horizon-eds/index.md
```diff
@@ -60,9 +60,9 @@ This will be used to access pilot on `cluster1` securely using the ingress gatew
       --set global.controlPlaneSecurityEnabled=true \
       --set global.proxy.accessLogFile="/dev/stdout" \
       --set global.meshExpansion.enabled=true \
-      --set global.meshNetworks.network2.endpoints[0].fromRegistry=n2-k8s-config \
-      --set global.meshNetworks.network2.gateways[0].address=0.0.0.0 \
-      --set global.meshNetworks.network2.gateways[0].port=443 \
+      --set 'global.meshNetworks.network2.endpoints[0].fromRegistry=n2-k8s-config' \
```
frankbu: This is needed for z shell? Is it the `[]` chars? Is this not also a problem in many other docs?

content/docs/examples/multicluster/split-horizon-eds/index.md
ericvn: I was trying to be consistent with the other files which added ' or "s in their -o jsonpath=... (ex: content/docs/examples/advanced-gateways/egress-gateway-tls-origination/index.md)

content/docs/examples/multicluster/split-horizon-eds/index.md
ericvn: I found that some other docs didn't have this issue. The used \' or \"s so added that here.

A global search `[]` and [0]`, does highlight that other docs may have this issue. I'll add another commit fixing those up.

content/docs/examples/multicluster/split-horizon-eds/index.md
ericvn: I was trying to get the command to generate the output that was included  in the text bash which had no `Completed` pods. I'm fine going back to the old commands, updating the sample output to include the `Completed` pods, and possibly adding a remark the all pods should be `Running` or `Completed`. Thoughts @frankbu on which makes sense.

frankbu: Yeah, I don't know. I notice that every time you update istio, you get another set of these post-install and cleanup `Completed` pods, so the output from `get po` keeps getting more and more messy. I think we need to look into a way to delete these pods, rather than deal with them in the docs.

content/docs/examples/multicluster/split-horizon-eds/index.md
frankbu: I think that we should fix all of the z shell problem for all of the docs in a consistent way and in one PR.

ericvn: Commit #2 in this PR should handle the z shell problem constantly across all docs. Standardized on single quoting on one side of the `=` unless single quotes were already used, then left the double-quote.

content/docs/examples/multicluster/split-horizon-eds/index.md
ericvn: I agree with the messiness of the `k get po` output. Trying to generate an acceptable command string can be difficult and error prone. This is where having something in `istioctl` to verify readiness would be a good thing.

We could do something about removing the pods, but then I wonder if we would want people to be able to gather logs for diagnosis of a problem and then they can't.

content/docs/examples/multicluster/split-horizon-eds/index.md
ericvn: Commit # 2 covers files with similar missing quotes around []'s.

content/docs/examples/multicluster/split-horizon-eds/index.md
ericvn: Discussion with @frankbu: We will leave the commands to show the `Completed` pods and add the `Completed` pods to the sample output for consistency.

ericvn: Commit # 3 has this work.

ericvn:
All current code review comments have been addressed.

PR #3856, opened 2019-03-26
Add namespace flag to get CRDs

comps3:

PR #3863, opened 2019-03-27
Add namespace flag to get CRDs

comps3:
Re-opened PR from #3856 but merging to release-1.1.

### DIFF
--- a/content/boilerplates/verify-crds.md
+++ b/content/boilerplates/verify-crds.md
@@ -5,6 +5,6 @@ If cert-manager is enabled, then the CRD count will be `58` instead.
 {{< /warning >}}
 
 {{< text bash >}}
-$ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | wc -l
+$ kubectl get crds --namespace istio-system | grep 'istio.io\|certmanager.k8s.io' | wc -l
 53
 {{< /text >}}


content/boilerplates/verify-crds.md
```diff
@@ -5,6 +5,6 @@ If cert-manager is enabled, then the CRD count will be `58` instead.
 {{< /warning >}}
 
 {{< text bash >}}
-$ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | wc -l
+$ kubectl get crds --namespace istio-system | grep 'istio.io\|certmanager.k8s.io' | wc -l
```
frankbu: Why is this needed? I thought that CRDs are global, so they appear in all namespaces.

comps3: In the installation instructions, CRDs are installed in the istio-system namespace. 
 `helm template install/kubernetes/helm/istio-init --name istio-init --namespace istio-system | kubectl apply -f -`

I verified when running that command with the namespace flag produces the expected result.

frankbu: Yes, but I think that running the command without the namespace flag produces the exact same output, so the flag is not needed. CRDs are very special resources, so the namespace doesn't matter (they are effectively in `every namespace`).

A separate question is whether we need to set the namespace for the helm install command for the istio-init chart. I think that doesn't make any difference either. @sdake: do you know?

comps3: After doing a fresh installation, it turns out namespace is not needed. Thank you for pointing that out! Although, I am also curious about whether namespace is needed for istio-init.

PR #3871, opened 2019-03-28
zh： help/ops/setup/app-health-check

loverto:
move and update


fleeto:
`[trans-bot:N/A]`

Thank you @loverto, there are more than 1 files in your PR, will not be reported to the task issues.

PR #3884, opened 2019-03-28
Fix sidecar.istio.io/inject usage in EFK example

MrLuje:
**sidecar.istio.io/inject** is not properly used in [example-fluentd-elasticsearch-kibana-stack](https://istio.io/docs/tasks/telemetry/logs/fluentd/#example-fluentd-elasticsearch-kibana-stack)

The annotation is used on deployment's metadata whereas is should be on the pod template spec's metadata (cf [docs](https://istio.io/docs/setup/kubernetes/additional-setup/sidecar-injection/#policy))


MrLuje:
I signed the CLA

Issue #3703, opened 2019-03-15
Need to produce helm repo related doc


linsun:
it was suggested in SRE doc by @sdake and @frankbu -- We'll need some doc on how to use helm repo.

from https://github.com/istio/istio.io/issues/3131

geeknoid:
This can wait till after 1.1 launches, right?

sdake:
post 1.1 is fine @geeknoid.  See https://github.com/istio/istio.io/issues/3131#issuecomment-473301078.

tiswanso:
@howardjohn added some comments to #3852 that really belong here:

> recent issues that came out of not documenting the Helm repo:
>  istio/istio#12889
>  istio/istio#12879
> 
> Another one istio/istio#12923

sdake:
@tiswanso had added a comment here:https://github.com/istio/istio.io/issues/3852#issue-425468176

That the location of the URL includes the version number.  This isn't correct.  All Helm charts for all versions should go in one directory and an index built for the files within that dir.

This permits an easy upgrade without having to reset the istio repo location.

@utka ^^

tiswanso:
@sdake -- Could you clarify whether you're saying we need enhancements to the release process to push versioned charts to a common helm repo?  The current release process is creating versioned helm chart repos--which I made the download doc page refer to in #3894.

sdake:
@tiswanso Thanks for the doc PR. That doesn't solve this particular problem.

This is incorrect in our current infrastructure:

 helm repo add istio.io https://storage.googleapis.com/istio-release/releases/{{< istio_full_version >}}/charts/
Instead, it should be

 helm repo add istio.io https://storage.googleapis.com/istio-release/releases/charts
All versions should be stored in that path, and a proper index file built to permit installing any supported version of Istio (this is especially necessary for upgrade from an helm repo

Thanks for the doc PR. That doesn't solve this particular problem.

Cheers
-steve

howardjohn:
This is a build/release issue though, not a docs issue. 

` helm repo add istio.io https://storage.googleapis.com/istio-release/releases/{{< istio_full_version >}}/charts/` is correct today, we release the Helm repo versioned.

Right?

If so I agree, seems like all other Helm repos work this way, we should open another issue under istio/istio for that maybe.

tiswanso:
[...]

PR #3894, opened 2019-03-29
Add release helm chart repo info to Istio download instructions

tiswanso:
Partial fix for #3703

### DIFF
--- a/content/docs/setup/kubernetes/download/index.md
+++ b/content/docs/setup/kubernetes/download/index.md
@@ -43,3 +43,11 @@ services from all other namespaces.
     {{< text bash >}}
     $ export PATH=$PWD/bin:$PATH
     {{< /text >}}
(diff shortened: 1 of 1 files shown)


sdake:
@tiswanso Thanks for the doc PR.  That doesn't solve this particular problem.

This is incorrect in our current infrastructure:

```
 helm repo add istio.io https://storage.googleapis.com/istio-release/releases/{{< istio_full_version >}}/charts/
```

Instead, it should be

```
 helm repo add istio.io https://storage.googleapis.com/istio-release/releases/charts
```

All versions should be stored in that path, and a proper index file built to permit installing any supported version of Istio (this is especially necessary for upgrade from an helm repo

Thanks for the doc PR.  That doesn't solve this particular problem.

Cheers
-steve

PR #3899, opened 2019-03-30
Egress by Kubernetes services

vadimeisenbergibm:
To demonstrate that the "legacy" Kubernetes services work for egress traffic. To answer the questions like https://stackoverflow.com/questions/55347688/istio-to-outside-cluster-communication-issue/55393208#55393208.


content/docs/examples/advanced-gateways/egress-kubernetes-services/index.md
```diff
@@ -0,0 +1,244 @@
...
+    $ kubectl get pod $SOURCE_POD_WITHOUT_ISTIO -n without-istio
+    NAME                     READY   STATUS    RESTARTS   AGE
+    sleep-66c8d79ff5-8tqrl   1/1     Running   0          32s
+    {{< /text >}}
+
+## Kubernetes ExternalName service to access an external service
+
+1.  Create a Kubernetes
+    [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) service for Wikipedia:
+
```
frankbu: s/Wikipedia/httpbin.org/

frankbu:
This is a fine example, but my concern that it's really just a lot more reading material, just to say one thing: you need to disable mutual TLS to call kubernetes external services. I'm worried that we have so much egress material, users will be overwhelmed. This whole example could be replaced by 1 or 2 sentences in the doc somewhere IMO. I short blog post that walks through this example might be more appropriate than adding another egress example.

I think we should discuss this issue in the doc WG. It's part of a bigger discussion about what exactly we want to have in tasks vs. examples vs. blogs.

vadimeisenbergibm:
@frankbu Notice the Stackoverflow question https://stackoverflow.com/questions/55347688/istio-to-outside-cluster-communication-issue/55393208#55393208.

The goal of this item is to show the following:
1. It is possible to continue using legacy Kubernetes Services for accessing external services
2. You have to disable mTLS for that.
3. You do not need to define Service Entries, as for other external services
4. If you need to use Kubernetes Services for some reason, this how to do it, two ways
5. This is how you can quickly check how Kubernetes Services can be used for egress with Istio

I am OK to move this item to some other section, a blog post, or, maybe, the [Operations Guide](https://istio.io/help/ops/). What do you think?

BTW, when is the next meeting of the Docs WG?

frankbu:
@vadimeisenbergibm I'm not sure what to do with it. Your 5 points, above, highlight what my concern is, not just with this one example, but with how we are doing examples in general.

The 5 simple list items you provided, above, is all the information users need. Just adding that to the operations guide vs. making people read through and try a long example that shows it indirectly, seems like a much easier way to provide the information to users. I really feel that by demonstrating everything with long examples, we are hiding important information from people that don't need to see it working.

frankbu:
Maybe we should use this example as a good case study for a more general discussion about examples in the documentation WG next week?

@geeknoid, what do you think about this?

vadimeisenbergibm:
@frankbu I am OK with whatever decision. The devil is in the details, though, you can never know which field or which value could interfere with Istio and what can go wrong, without seeing a working documentation item. Let's say it is a sanity check, if you want to make sure that Kubernetes Services can work for egress with Istio, this how you can perform that check.

geeknoid:
@frankbu Should we merge this in?

frankbu:
@geeknoid, we can't merge it as is. It's still in the old examples subsection that we move all to tasks. That said, I still think task or example is not the best way to explain this. I talked to Vadim about this, and he's going to take a pass at adding something to the operations guide to cover this information, and also, more generally, how/if it works for protocols other than http.

vadimeisenbergibm:
Closing in favor of https://github.com/istio/istio.io/pull/4710. This one got out-of-sync with the master branch.

PR #3924, opened 2019-04-02
Fix broken helm link in boilerplate security warn

sdake:


sdake:
I can't read Chinese so I am not sure if the links are correct for the ZH translation.

Cheers
-steve

PR #3925, opened 2019-04-02
Fix html rendering issue for mesh expansion

pitlv2109:
Put html tags from html response to `{{< text html >}}`
Currently html tags (part of html response) within `{{< text bash >}}` are rendered, which is not what we want.

### DIFF
--- a/content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
@@ -285,19 +285,24 @@ The `server: envoy` header indicates that the sidecar intercepted the traffic.
 
     {{< text bash >}}
     $ kubectl exec -it sleep-88ddbcfdd-rm42k -c sleep -- curl vmhttp.${SERVICE_NAMESPACE}.svc.cluster.local:8080
+    {{< /text >}}
+
+    You should see something similar to the output below.
+
+    ```html
     <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN"><html>
     <title>Directory listing for /</title>
     <body>
     <h2>Directory listing for /</h2>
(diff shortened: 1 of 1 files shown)


pitlv2109:
@geeknoid When adding the `text html` to the markdown file for this PR, the corresponding html was generated with an extra `</p>` tag, which made it fail the circleci test. Do you know how I could fix this issue?

geeknoid:
Unfortunately, Hugo has some problems with  embedded HTML tags which sometimes confuse the parser. Things to try:

- Instead of  two blocks, try {{< text bash html >}}. This is more inline with how we want to render commands and outputs overall and may solve the problem.

- If that doesn't work, try {{< text bash plain >}}

- If that doesn't  work, try using {{< text bash >}} for the first block as you have, and use triple-backticks as in standard YAML (```html)

content/docs/setup/kubernetes/additional-setup/mesh-expansion/index.md
```diff
@@ -285,14 +285,18 @@ The `server: envoy` header indicates that the sidecar intercepted the traffic.
 
     {{< text bash >}}
     $ kubectl exec -it sleep-88ddbcfdd-rm42k -c sleep -- curl vmhttp.${SERVICE_NAMESPACE}.svc.cluster.local:8080
+    {{< /text >}}
+    You should see something familiar to the output below.
```
geeknoid: familiar -> similar

pitlv2109: Oops. Thanks :)

pitlv2109:
@geeknoid Option 1 and 2 have a problem, when using the "Copy to clipboard" feature, it will copy the whole thing (bash command + html response). I chose option 3 instead. Thanks!

PR #3939, opened 2019-04-04
[WIP]remove http 1.0 workaround portion

linsun:
since it is already supported in 1.1 as experimental feature via

	HTTP10 = os.Getenv("PILOT_HTTP10") == "1"

### DIFF
--- a/content/help/ops/traffic-management/troubleshooting/index.md
+++ b/content/help/ops/traffic-management/troubleshooting/index.md
@@ -230,29 +230,3 @@ Check your `ulimit -a`. Many systems have a 1024 open file descriptor limit by d
 {{< /text >}}
 
 Make sure to raise your ulimit. Example: `ulimit -n 16384`
-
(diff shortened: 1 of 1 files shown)


linsun:
Need some validation first.  Think the support in place is only for sidecar, not the gateway.

frankbu:
Instead of removing this doc, we should probably update it instead, i.e., mention the PILOT_HTTP10 env variable, and also where exactly it is supported.

geeknoid:
To document environment variables, you can go to istio/istio and look for instances of RegisterBoolVar and then fill in the description parameter.

There's an issue open in istio/istio to add descriptions for all env vars and annotations.

geeknoid:
@linsun What do you want to do with this PR?

linsun:
@geeknoid i need to test a bit the instruction, i'll rework the PR.

geeknoid:
@linsun Ping.

PR #3940, opened 2019-04-04
fix regex to get multiple lines in between snippet tags

tvieira:
The text_import feature, using URL, is ignoring multiline snippets, and it is taking only the first line. This commit fixes the regex to get all lines in between `# $snippet XXXX` and `# $endsnippet`.

This commit was tested with this example:
https://raw.githubusercontent.com/tvieira/istio/snippets/docs/snippets/fault-injection.txt

At least, all 3 snippets worked as expected.

### DIFF
--- a/src/ts/codeBlocks.ts
+++ b/src/ts/codeBlocks.ts
@@ -315,7 +315,7 @@ function handleCodeBlocks() {
                 })
                 .then(data => {
                     if (code.dataset.snippet) {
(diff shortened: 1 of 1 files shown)


geeknoid:
YOu need to do "make build" and check in the resulting built files, otherwise this change will have no effect.

Try out the examples in the test/tb directory to make sure it all renders correctly.

geeknoid:
Thanks, I've incorporated your change in a different PR which addresses your issues.

https://github.com/istio/istio.io/pull/3942

So I'm going to close this one.

PR #3942, opened 2019-04-04
Site improvements

geeknoid:
- Fix multi-line snippets when loading a text block from a URL.

- Fix outputis for text blocks loaded from a URL

- Fix @@ expansion for text blocks loaded from a URL.

- Added more text block test cases


geeknoid:
@tvieira This replaces https://github.com/istio/istio.io/pull/3940

tvieira:
thanks @geeknoid ! Looks great.

PR #3945, opened 2019-04-05
Fixed a simple typo

SeriyBg:
can useful -> can be useful


frankbu:
This is a generated file. Please make the change in the source repo.
https://github.com/istio/istio.io/blob/6af3f7294005af776ad840fe745135017591f566/content/docs/reference/config/policy-and-telemetry/templates/reportnothing/index.html#L2

SeriyBg:
@frankbu oops, by bad!

PR #3949, opened 2019-04-05
make enabling of liveness HTTP probe rewrite a separate subsection

vadimeisenbergibm:
so it can be referenced from other documents

also, edit the sidecar injector config map instead of reinstalling Istio

### DIFF
--- a/content/help/ops/setup/app-health-check/index.md
+++ b/content/help/ops/setup/app-health-check/index.md
@@ -145,21 +145,26 @@ This approach rewrites the application `PodSpec` liveness probe, such that the p
 [Pilot agent](/docs/reference/commands/pilot-agent/). Pilot agent then redirects the
 request to application, and strips the response body only returning the response code.
 
-To use this approach, you need to install Istio with Helm option `sidecarInjectorWebhook.rewriteAppHTTPProbe=true`.
-Note this is a global flag. **Turning it on means all Istio app deployment will be affected.**
-Please be aware of the risk.
+To use this approach, you need to configure Istio to rewrite the liveness HTTP probes.
+
+##### Configure Istio to rewrite liveness HTTP probes
+
+Update the configuration map of Istio sidecar injection:
 
 {{< text bash >}}
-$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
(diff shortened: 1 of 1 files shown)


content/help/ops/setup/app-health-check/index.md
```diff
@@ -145,21 +145,26 @@ This approach rewrites the application `PodSpec` liveness probe, such that the p
...
-$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
-    --set global.mtls.enabled=true --set sidecarInjectorWebhook.rewriteAppHTTPProbe=true \
-    -f @install/kubernetes/helm/istio/values.yaml@ > $HOME/istio.yaml
-$ kubectl apply -f $HOME/istio.yaml
+$ kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e "s/ rewriteAppHTTPProbe: false/ rewriteAppHTTPProbe: true/" | kubectl apply -f -
 {{< /text >}}
 
-Re-deploy the liveness health check app.
+{{< warning >}}
+This is a global update. All Istio app deployments will be affected. Be aware of the risk.
```
incfly: Thanks for adding more options. I think it's okay to mention both, but using helm flag is more recommended and more manageable solution compared with editing configmap directly.

Can we put helm flag one in the first and config map editing as alternative?

vadimeisenbergibm: Thanks @incfly. I have put the helm option before the ConfigMap editing in https://github.com/istio/istio.io/pull/3949/commits/2d859f191c8dcb6829981dd7a4a86f5512f8b468

incfly: Thanks. Just one thing, here we mentioned "this is a global update." and below we have alternative to update config map. It can mislead users changing config map won't have that global effect. Maybe we can put the "global update" to the end of both methods?

Also "global update" seems not quite accurate, because we may not update anything. Just "global option/global flag/global effect" or  just remove it.

vadimeisenbergibm: @incfly I have addressed it in https://github.com/istio/istio.io/pull/3949/commits/09be4723bccfda2955fc08c4ff5019c24ee8ed83, thanks.

PR #3955, opened 2019-04-05
update version 1.0.7 for downloading and release notes.

yangminzhu:
This is for branch `release-1.0`

### DIFF
--- a/content/about/notes/1.0.7/index.md
+++ b/content/about/notes/1.0.7/index.md
@@ -1,0 +1,10 @@
+---
+title: Istio 1.0.7
+weight: 85
+icon: /img/notes.svg
+---
+
+This release includes an important security update. All customers using prior versions of Istio are advised to upgrade immediately.
+Please refer to the [announcement blog post](https://istio.io/blog/2019/announcing-1.1.2/) for a description of the security issue.
+
+{{< relnote_links >}}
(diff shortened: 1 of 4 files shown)


yangminzhu:
@geeknoid the test failed with error ```/bin/sh: scripts/gen_site.sh: not found Exited with code 127``` which seems unrelated to this PR, do you know what caused it?

geeknoid:
The release 1.0 branch has been archived at this point and is only available from archive.istio.io. We shouldn't be updating any content in that branch.

yangminzhu:
@geeknoid If some followed the download page in https://archive.istio.io/v1.0/docs/setup/kubernetes/download-release/, they will  get the latest version `1.1.2`, not the `1.0.7`. I thought the 1.0 site should always provide information about 1.0 version for user still using it. That's why we publish the fix for 1.0.7, right?

geeknoid:
Yes, in general we want the releases to be self consistent even over time. The 1.0 doc set had a couple issues in that regard, including the specific link you mention. When 1.1 is archived in a few months, it won't have this problem.

geeknoid:
I'm not sure why CI is failing like that. That's awfully strange. But we shouldn't be modifying the archived site anyway, so let's just close this PR.

PR #3961, opened 2019-04-08
Add SkyWalking in istio mixer adaptor page

wu-sheng:
SkyWalking has supported Istio mixer bypass adaptor from 6.0.0-alpha, with GA released one month ago, we plan to add the project in the [adaptor page](https://preliminary.istio.io/docs/reference/config/policy-and-telemetry/adapters/)

Because of using the mixer bypass, rather than write our own adaptor in hard code, we provide a very simple page, https://github.com/apache/incubator-skywalking-data-collect-protocol/blob/master/service-mesh-probe/istio/skywalking.config.pb.html


wu-sheng:
@geeknoid From the preview page, https://deploy-preview-3961--preliminary-istio.netlify.com/docs/reference/config/policy-and-telemetry/adapters/, look like it hasn't working.

Could you tell me why?

geeknoid:
After the change to grab_reference_docs.sh, then you also need to run that script and then checkin anything that's changed. That should cause your addition to actually show up in the rendered HTML.

wu-sheng:
@geeknoid 
Page generated and commit. https://github.com/istio/istio.io/blob/aca82f0c005c6844ae11f51716516b91defe3e89/scripts/content/docs/reference/config/policy-and-telemetry/adapters/apache-skywalking/index.html

But the deploy page looks like same, still haven't included the new one. https://deploy-preview-3961--preliminary-istio.netlify.com/docs/reference/config/policy-and-telemetry/adapters/

Do I miss something else?

geeknoid:
I'm not sure what's going on. Let me try to do this myself and see how it turns out.

wu-sheng:
> I'm not sure what's going on. Let me try to do this myself and see how it turns out.

@geeknoid Thanks.

geeknoid:
I've got this covered by https://github.com/istio/istio.io/pull/3971, so I'm closing this PR.

Thanks.

PR #3971, opened 2019-04-09
Update reference docs.

geeknoid:
Adds the Skywalking adapter docs.


wu-sheng:
Why I didn't generate the right one? Do you have a clue?
I hope we could make it right in next time.

PR #3973, opened 2019-04-09
Improved egress task

### LINKED ISSUE #3841: Issue with docs/tasks/traffic-management/egress/index.md
I noticed that the option global.outboundTrafficPolicy.mode=ALLOW_ANY is enabled by default, it's really help for beginners, thanks. But I have some confusions with the following expression: 

> For example, if you don’t register an HTTP service or define a service entry for port 8000 within the mesh, the sidecar proxy will forward requests to any external service on that port. **However, if you later create an HTTP service inside the mesh on port 8000 or define a service entry for any host on port 8000, Istio will block all external access to port 8000**; Istio will fall back to a blocking-by-default behavior for that port.

As I can see, a serviceentry is defined for allowing accessing an outside service, but why after defining a service entry for any host on port 8000, istio will **block** all all external access to port 8000? 
Anyone who can explain this please help. Thanks in advance.
[https://istio.io/docs/tasks/traffic-management/egress/#envoy-passthrough-to-external-services](url)

### LINKED ISSUE #3883: Clarify interaction of includeIPRanges and outboundTrafficPolicy
The [docs here](https://istio.io/docs/tasks/traffic-management/egress/) cover using the new default `global.outboundTrafficPolicy.mode=ALLOW_ANY` setting to allow all outbound traffic by default.

It also covers using `global.proxy.includeIPRanges` to restrict proxy interception to a known cluster CIDR.

There are two issues here:
1. `ALLOW_ANY` is misleading. By default it doesn't actually allow traffic on port 80, for example -- it gets intercepted and returns a 404, which means systems like `kube2iam` won't work. There are workarounds with a `ServiceEntry` but the description makes it sound like those aren't needed... and they don't work at all for IP addresses (like 169.254.169.254).
2. The interaction between these options are unclear. If `ALLOW_ANY` is set, does that take precedence over `includeIPRanges`? Do they somehow intersect? The Istio chart side of this should probably simplify the configuration here too so it's clear if only one is used at a time.

frankbu:
Fixes #3883 
Fixes #3841

### DIFF
--- a/content/docs/tasks/traffic-management/egress/index.md
+++ b/content/docs/tasks/traffic-management/egress/index.md
@@ -8,10 +8,14 @@ keywords: [traffic-management,egress]
 ---
 
 Because all outbound traffic from an Istio-enabled pod is redirected to its sidecar proxy by default,
-Istio must be configured in a way that allows access to URLs outside of the cluster before external
-services can be called. This task will show you how to do it in three different ways:
+accessability of URLs outside of the cluster depends on the configuration of the proxy.
+By default, Istio configures the Envoy proxy to passthourgh requests for unknown services.
+Although this provides a convenvient way to get started with Istio, configuring
+stricter control is usually preferrable.
 
-1. Configure the Envoy proxy to pass requests through to external services on ports that are not configured inside the mesh.
+This task shows you how to access external services in three different ways:
+
+1. Configure the Envoy proxy to pass requests through to external services that are not configured inside the mesh.
 1. Configure [service entries](/docs/reference/config/networking/v1alpha3/service-entry/) to provide controlled access to external services.
 1. Bypass the Envoy proxy for a specific range of IPs.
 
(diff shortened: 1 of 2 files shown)


content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -425,8 +462,10 @@ Note that configuration examples in this task **do not enable secure egress traf
 A malicious application can bypass the Istio sidecar proxy and access any external service without Istio control.
 {{< /warning >}}
 
-To implement egress traffic control in a secure way, you must [direct egress traffic through an egress gateway](/docs/examples/advanced-gateways/egress-gateway) and address the security concerns expressed in
-[Configure an Egress Gateway example, Additional Security Considerations](/docs/examples/advanced-gateways/egress-gateway#additional-security-considerations).
+To implement egress traffic control in a completely secure way, you must
+[direct egress traffic through an egress gateway](/docs/examples/advanced-gateways/egress-gateway)
+and address the security concerns described in
```
rcaballeromx:
```suggestion
and review the security concerns described in our
```

content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -425,8 +462,10 @@ Note that configuration examples in this task **do not enable secure egress traf
...
+[Additional security considerations](/docs/examples/advanced-gateways/egress-gateway#additional-security-considerations).
```
rcaballeromx:
```suggestion
[additional security considerations](/docs/examples/advanced-gateways/egress-gateway#additional-security-considerations).
```

content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -52,30 +66,38 @@ you cannot use this approach for services using those ports.
 
     {{< /tip >}}
 
-1.  Make a couple of requests to external HTTPS services from `SOURCE_POD`:
+1.  Make a couple of requests to external HTTPS services from `SOURCE_POD` and confirm that they produce
```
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx Note that once `SOURCE_POD` is defined, the sleep app is irrelevant. SOURCE_POD can contain any pod used as a source for sending curl requests.

content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -52,30 +66,38 @@ you cannot use this approach for services using those ports.
...
+    successful `200` responses:
 
     {{< text bash >}}
-    $ kubectl exec -it $SOURCE_POD -c sleep -- curl -s https://www.google.com | grep -o "<title>.*</title>"; kubectl exec -it $SOURCE_POD -c sleep -- curl -s https://edition.cnn.com | grep -o "<title>.*</title>"
-    <title>Google</title>
```
vadimeisenbergibm: I thought it is a nice visual clue. It also shows that each request arrived to its intended destination, and not to some other service that mistakenly returned 200.

frankbu: `grep -o` doesn't work so well no my machine. It doesn't seem to add a <CR> character so the output keeps moving to the right. The 200 codes print nice and clean and I think it's also very clear that the external request is going out

vadimeisenbergibm:
@frankbu `--set global.outboundTrafficPolicy.mode=ALLOW_ANY --set pilot.env.PILOT_ENABLE_FALLTHROUGH_ROUTE=1` should not be added to the boiler plate, since they are not relevant for all the egress examples, except for https://istio.io/docs/tasks/traffic-management/egress/.

vadimeisenbergibm:
@frankbu Added an issue https://github.com/istio/istio.io/issues/3989.

Issue #3851, opened 2019-03-26
Update docs regarding Mixer CRD usage


geeknoid:
Need to update a bunch of docs per https://github.com/istio/istio/pull/12689

PR #3975, opened 2019-04-09
Update mixer configs

kyessenov:
Fixes https://github.com/istio/istio.io/issues/3851
We provide two versions of samples temporarily, while we complete deprecation of old style CRDs.
Links to samples instead of hard-coding them inside the content for easier automation.

/assign @geeknoid


geeknoid:
So in general:

- Let's only merge this into master since this is about post-1.1 work.

- The text does not need to say "when using 1.1.2 or earlier" and should not include that text. The next update to the doc set will be documenting Istio 1.2 and it doesn't need to cover previous versions, that' why we have archive.istio.io.

So just focus on documenting the desired modern behavior.

kyessenov:
@geeknoid The problem with that is that we need to support both types of configs for at least a version with documentation for both. We already have questions about what the difference between the two. The fact that current builds of istio have a problem with not processing the new style makes the problem worse. So that means if 1.2 is the first version that uses the new style of config, then the users have an issue with moving their config over from 1.1.

geeknoid:
OK, then replace all uses of "version 1.1.2" with "Istio 1.1.2"

kyessenov:
@geeknoid I manually compared old and new generated websites for any typos I could have made and did not find any. I think this is ready for review and merge. Thanks!

kyessenov:
Since @geeknoid is OoO, @linsun -- it would be great to merge this when the release goes out.

geeknoid:
So this should be merged after 1.1.3 is out, correct?

kyessenov:
Sometime around 1.1.3. The samples are already showing up, however.

content/docs/tasks/policy-enforcement/rate-limiting/index.md
```diff
@@ -244,18 +99,8 @@ so the configuration to enable rate limiting on both adapters is the same.
...
-      name: requestcount
-      namespace: istio-system
-    spec:
-      dimensions:
-        source: request.headers["x-forwarded-for"] | "unknown"
-        destination: destination.labels["app"] | destination.service.host | "unknown"
-        destinationVersion: destination.labels["version"] | "unknown"
+    {{< text bash >}}
+    $ kubectl -n istio-system get instance requestcountquota -o yaml
     {{< /text >}}
```
jwendell: While on that, could you fix the following Paragraph, which mentions `destination.service.host`. The yaml file uses `destination.service.name`.

kyessenov: Yes, will do.

PR #3980, opened 2019-04-10
add ygrene to list of companies using Istio in Action

zparnold:
This adds Ygrene Energy Fund to the list of companies present and also removes the CSS animation that makes logos spin. The first one I'd like to have for sure, the animation I can remove, but I feel that it detracts from the companies when the `:hover` state is fired in the DOM


zparnold:
I signed it!

geeknoid:
I've received a lot of comments about folks loving the whimsy of the fidget-spinner animations we use there, so I'd like to keep that as-is. Would you just add the company logo without the animation change please?

zparnold:
Yep! @geeknoid will do. Thanks 😄

zparnold:
@geeknoid Just pushed

PR #3981, opened 2019-04-11
fix Debugging Envoy and Pilot

yangminzhu:
Istio 1.1.2 changed to `global.outboundTrafficPolicy.mode: Always` by default which means Istio lets calls to unknown services pass through.

See:
https://istio.io/docs/tasks/traffic-management/egress/#envoy-passthrough-to-external-services
https://istio.io/docs/reference/config/installation-options-changes/#new-global-key-value-pairs

### DIFF
--- a/content/help/ops/traffic-management/proxy-cmd/index.md
+++ b/content/help/ops/traffic-management/proxy-cmd/index.md
@@ -28,16 +28,16 @@ receiving configuration or is out of sync then `proxy-status` will tell you this
 
 {{< text bash >}}
 $ istioctl proxy-status
-PROXY                                                  CDS        LDS        EDS               RDS          PILOT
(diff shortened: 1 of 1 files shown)


yangminzhu:
@sdake @frankbu @geeknoid Could you take a look? This updates the doc to match the 1.1.2 default behavior.

howardjohn:
The original docs were wrong, black hole cluster wont 404, it won't send any HTTP response at all. In 1.1 we changed to ALLOW_ANY mode as the default which makes a PassThrough cluster as the default, which behaves like this doc suggests - allow all traffic through. But before 1.1.3 this had a bug/"feature" that if you sent a request to a service/external site on the same port as a service you had (and keep in mind, Istio ships with something on port 80), it would not reach the PassThrough cluster, resulting in a 404.

Now with 1.1.3 and the flag external traffic works how you would expect

PR #3983, opened 2019-04-11
Add the diagram creation guidelines.

rcaballeromx:
These guidelines include an SVG file with all the shapes and styles needed. The
steps to get someone started are meant to entice people to create their own
diagrams.


content/about/contribute/diagrams/index.md
```diff
@@ -0,0 +1,45 @@
+---
```
geeknoid: I think we need to state an important requirement. SVG is typically an output format and we need to have the original source of diagrams. So we can add something about:

1) Put the original images in our team drive: https://drive.google.com/corp/drive/u/0/folders/1jczscJueUBR3IOvH30q9HAZX0jC7GSyW

2) Put a comment in the markdown with the URL to the original source.

geeknoid: Along the same lines, it would be considerably more useful, at least for Googlers using Google Drawing, if the sample content were supplied as a Google drawing rather than an SVG.

frankbu: +1 on both of these

rcaballeromx: @geeknoid @frankbu I agree but SVG is an editable format, not a read-only format. I already have a Google drawing version of the guide ready. I can add it to the community drive and add a link.
However, it would be better for people planning on reusing the assets to do so with a proper vector graphics editing tool, such as InkScape, using SVG.

rcaballeromx: @geeknoid What I want to avoid is having two sources of truth for diagrams, the Gdrive and the repo. IMO, only the repo should be the source of truth for content. The guidelines accomplish that by requiring the diagrams be added to the repo in SVG format with editable text but I am open to other formats.

geeknoid: Most of our diagrams have been created using Google Drawing, which doesn't support SVG as an input format. I think we need to recognize the reality of the tools most of our developers use and are familiar with.

I agree that having source content in one format and published content be in a different format isn't ideal. But it has generally been working for us, so it's a manageable strategy.

rcaballeromx: @frankbu @geeknoid I've added the guide to the community's Misc folder and a link to it in the doc. Google Draw allows you to export your drawing as an SVG which means contributors can include their Google Draw diagrams in their pull requests as SVG very easily.
PTAL
@geeknoid Any idea how to solve the spellchecker failure due to the SVG quotation marks?

geeknoid:
The CI failure is because of an overly aggressive grep (a few actually) done in scripts/lint_sites. I should be checking just .html or .md files, but it's checking everything in the content folder. Let me fix that and then you can rebase and the CI failure should go away.

I'm surprised this hasn't happen before.

geeknoid:
OK, if you rebase the CI error should go away.

content/about/contribute/diagrams/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+Our goal is to drive consistency across all diagrams in our website to ensure
+diagrams are clear, technically accurate, and accessible.
+
+Keeping the text editable allows the community to improve and change the
+diagrams as needed.
+
+Follow these steps to begin using the guideline:
+
+1. [Download the guidelines as an SVG file.](./diagram-guidelines.svg)
```
geeknoid: I think normally the . would go outside the link.

content/about/contribute/diagrams/index.md
```diff
@@ -0,0 +1,47 @@
...
+helps make diagrams easier to understand and the content more accessible.
+
+Thank you for contributing to the Istio documentation!
+
+{{< image width="75%"
+    link="./diagram-guidelines.svg"
+    alt="The Istio diagram creation guidelines in SVG format."
+    title="The Istio Diagram Creation Guidelines"
+    >}}
+
```
geeknoid: I'm surprised you're not getting warnings for having these blank lines...

content/about/contribute/diagrams/index.md
```diff
@@ -0,0 +1,47 @@
...
+Keeping the text editable allows the community to improve and change the
+diagrams as needed.
+
+Follow these steps to begin using the guideline:
+
+1. [Download the guidelines as an SVG file.](./diagram-guidelines.svg)
+1. Select the shapes you want to use.
+1. Copy them into a new SVG file.
+1. Make as many duplicates as you need for your diagram.
+1. Connect the shapes with the appropriate style of line.
```
geeknoid: I think the first four bullets are bit too hand-holdy for folks. Can we just summarize in one sentence:

Use the diagram for guidance, cut and paste from it as needed, and then follow these steps:

rcaballeromx:
@geeknoid  I've addressed all your comments and implemented your suggestions. Thanks for fixing the linter's grep; I rebased and everything seems to be working fine.

PR #3984, opened 2019-04-12
Optimize user experiences by replacing a manual step with a command

lei-tang:
Optimize user experiences of Isito Vault CA integration by replacing a manual step with a command.


lei-tang:
https://github.com/istio/istio/issues/13439

PR #3988, opened 2019-04-14
Update the Egress MongoDB blog post

vadimeisenbergibm:
1. replace checking Envoy's statistics with printing Envoy's logs to verify that the traffic flows through the egress gateway
1. add enabling Envoy's logging
1. use fake hostname in TCP routing + static endpoints, to enable routing to MongoDB services without an FQDN hostname, since now IPs are not allowed as hostnames.
1. add a flag to enable egress gateway since now it is disabled by default

### DIFF
--- a/content/blog/2018/egress-mongo/index.md
+++ b/content/blog/2018/egress-mongo/index.md
@@ -217,7 +217,7 @@ instructions in this section. Alternatively, if you do want to direct your traff
     Also note that when the protocol `TCP` is specified, the configuration is not specific for MongoDB, but is the same
     for any other database with the protocol on top of TCP.
 
-    Note that the host of your MongoDB is not used in TCP routing, so you can use any host, for example `my-mongo-fake-host.com`. Notice the `STATIC` resolution and the endpoint with the IP of your MongoDB service. Once you define such an endpoint, you can access MongoDB services which do not have an FQDN hostname.
+    Note that the host of your MongoDB is not used in TCP routing, so you can use any host, for example `my-mongo-fake-host.com`. Notice the `STATIC` resolution and the endpoint with the IP of your MongoDB service. Once you define such an endpoint, you can access MongoDB services that do not have an FQDN hostname.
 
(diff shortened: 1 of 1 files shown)


content/blog/2018/egress-mongo/index.md
```diff
@@ -214,6 +217,8 @@ instructions in this section. Alternatively, if you do want to direct your traff
     Also note that when the protocol `TCP` is specified, the configuration is not specific for MongoDB, but is the same
     for any other database with the protocol on top of TCP.
 
+    Note that the host of your MongoDB is not used in TCP routing, so you can use any host, for example `my-mongo-fake-host.com`. Notice the `STATIC` resolution and the endpoint with the IP of your MongoDB service. Once you define such an endpoint, you can access MongoDB services that do not have an FQDN hostname.
```
frankbu: I think the name `my-mongo-fake-host.com` seems like a hack. Although the host isn't used for runtime routing, it is used for config matching (gateway and destination rules hosts need to match), so I think we should come up with a better recommended naming convention to use for TCP service entries like this. We should document a convention somewhere, and use it in this blog IMO.

Maybe something like `my-mongo.tcpsvc` ?

frankbu:
Why do you have two sections: no egressgateway and egressgateway? Now that the only difference is to deploy the egressgateway, why not just have the no-gateway section first, and then in the following section the service entry is already there so you just deploy the gateway and start configuring the redirection?

content/blog/2018/egress-mongo/index.md
vadimeisenbergibm: @frankbu `my-mongo.tcpsvc` sounds good, I will use it.

vadimeisenbergibm:
@frankbu 

> Why do you have two sections: no egressgateway and egressgateway

You are right, this is a leftover from the previous setting. I will do as you proposed.

vadimeisenbergibm:
@frankbu I have applied your comments, thank you for the review.

PR #3990, opened 2019-04-14
1.1.3 release notes

duderino:


geeknoid:
Also need to update the last line of static/_redirects to point to the right chart version. Updating this is needed in the 1.1 branch, but has been automated in master for > 1.1.x releases.

geeknoid:
Found another file that tracks versions, which needs to be updated: data/release.yml.

geeknoid:
Can we remove links to commits and links to PRs? This is for customers, not developers. Links to issues is fine, but commits and PRs should not be part of our customer-facing material.

duderino:
I think it’s useful information and should be preserved because many of our
customers are quite savvy.   If customers complain about it I’d happily
remove it.  Otherwise I think we should let them decide.

geeknoid:
Our general policy for istio.io is to treat Istio as a product, and not as an OSS project. We've always kept a separation between what we show our customers, and what we use in our development process. 

Folks that care can go and find this information. In fact, the release note page contains a diff button to get the full diff if needed. So the mainline content doesn't need to repeat this stuff and has not done so in the past.

PR #3994, opened 2019-04-15
Remove warning from egress boilerplate

### LINKED ISSUE #3989: remove --set global.outboundTrafficPolicy.mode=ALLOW_ANY --set pilot.env.PILOT_ENABLE_FALLTHROUGH_ROUTE=1 from the egress boilerplate

frankbu:
Fixes #3989 

Put copy in egress task for now, because the warning is temporary. In the future we can switch back to using the boilerplate version (without warning) in egress task too.

### DIFF
--- a/content/docs/tasks/traffic-management/egress/index.md
+++ b/content/docs/tasks/traffic-management/egress/index.md
@@ -19,7 +19,45 @@ This task shows you how to access external services in three different ways:
 1. Configure [service entries](/docs/reference/config/networking/v1alpha3/service-entry/) to provide controlled access to external services.
 1. Completely bypass the Envoy proxy for a specific range of IPs.
 
-{{< boilerplate before-you-begin-egress >}}
+## Before you begin
+
+*   Setup Istio by following the instructions in the [Installation guide](/docs/setup/).
+
+    {{< warning >}}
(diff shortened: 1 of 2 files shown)


content/docs/tasks/traffic-management/egress/index.md
```diff
@@ -19,7 +19,45 @@ This task shows you how to access external services in three different ways:
 1. Configure [service entries](/docs/reference/config/networking/v1alpha3/service-entry/) to provide controlled access to external services.
 1. Completely bypass the Envoy proxy for a specific range of IPs.
 
-{{< boilerplate before-you-begin-egress >}}
+## Before you begin
+
+*   Setup Istio by following the instructions in the [Installation guide](/docs/setup/).
+
+    {{< warning >}}
```
vadimeisenbergibm: Don't you want to put this warning before the "Before you begin" section, and then use the boilerplate?

frankbu: I really want it at this step of the instructions, so there's no nice way to use the boilerplate (as is) and add the warning IMO.

vadimeisenbergibm: OK, sounds good.

PR #4025, opened 2019-04-24
Release notes for 1.1.4

geeknoid:


content/boilerplates/notes/1.1.4.md
```diff
@@ -0,0 +1,42 @@
...
+- Fixed bug where [`ConsistentHashLB.minimumRingSize`](/docs/reference/config/networking/v1alpha3/destination-rule/#LoadBalancerSettings-ConsistentHashLB)
+was defaulting to 0 instead of the documented 1024 ([Issue 13261](https://github.com/istio/istio/issues/13261)).
+
+## Small enhancements
+
+- Updated to the latest version of the [Kiali](https://www.kiali.io) add-on.
+
+- Updated to the latest version of [Grafana](https://grafana.com).
+
+- Added validation to ensure Citadel is only deployed with a single replica ([Issue 13383](https://github.com/istio/istio/issues/13383)).
```
linsun: How does this work for upgrade?  I'm asking because we always wanted folks to increase number of replicas before upgrade for each control plane components.

content/boilerplates/notes/1.1.4.md
```diff
@@ -0,0 +1,42 @@
+## Behavior change
+
+- Changed the default behavior for Pilot to enable fall through routes which allows workloads to access external host using the same port as any
+internal HTTP service. This behavior can be controlled by the `PILOT_ENABLE_FALLTHROUGH_ROUTE` environment variable.
```
linsun: @howardjohn any reason why any one would want to set this `PILOT_ENABLE_FALLTHROUGH_ROUTE`  to false?

howardjohn: https://xkcd.com/1172/

Alternatively, if something goes wrong and it turns out there is a bug, it causes perf regressions, etc.

There shouldn't be though, we won't have the flag in 1.2, its just as a safety switch because its a new change mid-release

content/boilerplates/notes/1.1.4.md
```diff
@@ -0,0 +1,42 @@
+## Behavior change
+
+- Changed the default behavior for Pilot to enable fall through routes which allows workloads to access external host using the same port as any
```
howardjohn: Fallthrough route isn't really something a user needs to know about (or a envoy/istio term). Maybe just make this say that all external traffic will now be allowed, even if it is on the same port as an internal service.?

content/boilerplates/notes/1.1.4.md
geeknoid: @lei-tang I believe you made this change?

content/boilerplates/notes/1.1.4.md
geeknoid: Rephrased. PTAL.

linsun: yes, i like that better.  That is essentially why i asked the q :)

content/boilerplates/notes/1.1.4.md
lei-tang: Multi-replica Citadel is still under development; currently for secret-mount-based Citadel when multiple Citadel are deployed, they have race condition. Per the quote from @jcetkov in istio/istio#13383: "it might be appropriate to modify the helm charts not to allow setting the security replica count via variable, if such case is really not supported and results in hard to trace issues."

linsun: ok, thanks, i validated we only allow 1 replica now.  should we update this?  https://istio.io/docs/setup/kubernetes/upgrade/steps/  -- `To minimize downtime, please ensure your Istio control plane components and your applications are highly available with multiple replicas` to exclude citadel?

lei-tang: Yes, Citadel should be excluded from this. I will create a PR to update it.

lei-tang: https://github.com/istio/istio.io/pull/4026 has been created to update this.

Issue #4024, opened 2019-04-24
Issue with docs/examples/advanced-gateways/egress-gateway/index.md


zuxahiz:
When doing with example "Configure an Egress Gateway" with mtls enabled, I found that after applying DestinationRule: egressgateway-for-cnn, the tls-check between sleep pod is failed:
```
$ istioctl authn tls-check sleep-57f9d6fd6b-7w2sw istio-egressgateway.istio-system.svc.cluster.local
HOST:PORT                                                 STATUS       SERVER     CLIENT     AUTHN POLICY     DESTINATION RULE
istio-egressgateway.istio-system.svc.cluster.local:80     CONFLICT     mTLS       HTTP       default/         egressgateway-for-cnn/default
```

It seems that after applying DestinationRule: egressgateway-for-cnn, the mtls on the egressgateway side is set to disabled infact.
Has someone face the same problem or you can give some advises on debugging?
https://istio.io/docs/examples/advanced-gateways/egress-gateway/#egress-gateway-for-http-traffic

vadimeisenbergibm:
The conflicting cluster is not used, another cluster is used to communicate with the egress gateway service and that cluster has mTLS enabled. In any case, I fixed the documentation so `istiocl authn tis-check` will not complain.

```
istioctl authn tls-check $SOURCE_POD istio-egressgateway.istio-system.svc.cluster.local
HOST:PORT                                                 STATUS     SERVER     CLIENT     AUTHN POLICY     DESTINATION RULE
istio-egressgateway.istio-system.svc.cluster.local:80     OK         mTLS       mTLS       default/         egressgateway-for-cnn/default
```

zuxahiz:
@vadimeisenbergibm Thanks for your quick response! This problem has been solved.

But I still have something unclear: Sometimes I found it doesn't work and got a "**_HTTP/1.1 503 Service Unavailable_**" error. That means if you combine those commands into a script and execute it many times, it will fail sometimes. 
Due to my limited knowledge of mtls, I am not sure if is because of the cleanup procedure don't  sweep all the configuration of this example, or mtls connection don't closed before re-connection? 
Also I noticed that the second response from "http://edition.cnn.com/politics" is "**_HTTP/2 200_**" now, rather than "**_HTTP/1.1 200 OK_**", maybe the website has upgrade their http version to http2. The docs may need to change. And do you think it will influence the above matter?

vadimeisenbergibm:
[...]

PR #4039, opened 2019-04-25
apply ISTIO_MUTUAL to egress gateway in the general cases

vadimeisenbergibm:
Apply ISTIO_MUTUAL in DestinationRules not only in the used subsets, but also in the general cases.
While only the subsets are used to communicate with the egress gateway
service, not applying ISTIO_MUTUAL in the general cases causes istiocl authn tls-check to produce conflict errors

fixes https://github.com/istio/istio.io/issues/4024

### DIFF
--- a/content/blog/2018/egress-mongo/index.md
+++ b/content/blog/2018/egress-mongo/index.md
@@ -386,6 +386,14 @@ enable Mixer policy enforcement based on that identity. By enabling mutual TLS y
       name: egressgateway-for-mongo
     spec:
       host: istio-egressgateway.istio-system.svc.cluster.local
+      trafficPolicy:
+        loadBalancer:
+          simple: ROUND_ROBIN
+        portLevelSettings:
+        - port:
+            number: 443
+          tls:
+            mode: ISTIO_MUTUAL
       subsets:
       - name: mongo
         trafficPolicy:
@@ -597,6 +605,14 @@ to be 443. The egress gateway accepts the MongoDB traffic on the port 443, match
       name: egressgateway-for-mongo
     spec:
       host: istio-egressgateway.istio-system.svc.cluster.local
+      trafficPolicy:
+        loadBalancer:
+          simple: ROUND_ROBIN
+        portLevelSettings:
+        - port:
+            number: 443
+          tls:
+            mode: ISTIO_MUTUAL
       subsets:
       - name: mongo
         trafficPolicy:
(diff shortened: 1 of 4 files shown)


content/blog/2018/egress-mongo/index.md
```diff
@@ -386,6 +386,14 @@ enable Mixer policy enforcement based on that identity. By enabling mutual TLS y
       name: egressgateway-for-mongo
     spec:
       host: istio-egressgateway.istio-system.svc.cluster.local
+      trafficPolicy:
```
frankbu: Since this isn't used anyway, wouldn't it be good enough to make the traffic policy same as default, i.e., just globally set the mode and nothing else?
```
trafficPolicy:
  tls:
    mode: ISTIO_MUTUAL
```
This is much simpler and all you really need to avoid the auth conflict, right?

vadimeisenbergibm: I think the issue is with the subset, which I need to define to set SNI. Once I define a subset without the tls mode, it will cancel the global mode, I am afraid.

vadimeisenbergibm: The issue is the same as with https://github.com/istio/istio/blob/master/samples/bookinfo/networking/destination-rule-all-mtls.yaml vs. https://github.com/istio/istio/blob/master/samples/bookinfo/networking/destination-rule-all.yaml.

When you define a subset, you have to specify the TLS mode.

frankbu: Yes, I understand. I meant you should leave the trafficPolicy in the subsets as they are, but when you add the top level trafficPolicy, you can simplify it. Instead of adding:
```
      trafficPolicy:
        loadBalancer:
          simple: ROUND_ROBIN
        portLevelSettings:
        - port:
            number: 443
          tls:
            mode: ISTIO_MUTUAL
```
you can simplify it to just:
```
      trafficPolicy:
        tls:
          mode: ISTIO_MUTUAL
```

vadimeisenbergibm: But why should I interfere with other ports? Suppose some user would want to use port 80 without mTLS and without TLS origination for some external services, while using port 443 with mTLS. I prefer to be specific here.

frankbu: My point is that if you had not created a DestinationRule at all, the [default destination rule in istio-system](https://github.com/istio/istio/blob/f03880b23dfd1aeef0b734e5766d87bdec3e1c3f/install/kubernetes/helm/istio/charts/security/templates/enable-mesh-mtls.yaml#L23) uses mTLS on all ports by default. So why change that for ports you are not using in the example? If users want to turn off mTLS for some other port, they would add another port level setting for it, rather then relying on the overridden default (which isn't even explicit - it's only overridden because this DestinationRule doesn't declare a top level trafficPolicy).

PR #4073, opened 2019-04-29
set environment variables for cluster contexts automatically

vadimeisenbergibm:
fix the wording, remove passive voice

### DIFF
--- a/content/boilerplates/kubectl-multicluster-contexts.md
+++ b/content/boilerplates/kubectl-multicluster-contexts.md
@@ -1,7 +1,7 @@
 ---
 ---
 *   You can use the `kubectl` command to access both the `cluster1` and `cluster2` clusters with the `--context` flag,
-for example `kubectl get pods --context cluster1`.
+    for example `kubectl get pods --context cluster1`.
     Use the following command to list your contexts:
 
     {{< text bash >}}
--- a/content/boilerplates/kubectl-multicluster-contexts.md
+++ b/content/boilerplates/kubectl-multicluster-contexts.md
@@ -20,3 +20,8 @@ for example `kubectl get pods --context cluster1`.
     $ echo CTX_CLUSTER1 = ${CTX_CLUSTER1}, CTX_CLUSTER2 = ${CTX_CLUSTER2}
     CTX_CLUSTER1 = cluster1, CTX_CLUSTER2 = cluster2
     {{< /text >}}
+
+{{< tip >}}
+If you have more than two clusters in the context list and you want to configure your mesh using clusters other than the
+first two, you will need to manually set the environment variables to the appropriate context names
+{{< /tip >}}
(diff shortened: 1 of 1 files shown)


content/boilerplates/kubectl-multicluster-contexts.md
```diff
@@ -10,9 +12,11 @@
               cluster2   cluster2   user@foo.com   default
     {{< /text >}}
 
-* Export the following environment variables with the context names of your configuration:
+*   Store the context names of your clusters in environment variables:
 
     {{< text bash >}}
-    $ export CTX_CLUSTER1=<KUBECONFIG_CONTEXT_NAME_FOR_CLUSTER_1>
-    $ export CTX_CLUSTER2=<KUBECONFIG_CONTEXT_NAME_FOR_CLUSTER_2>
+    $ export CTX_CLUSTER1=$(kubectl config view -o jsonpath='{.contexts[0].name}')
```
frankbu: I like this, but we should probably make it clear that you can only do this if the two clusters you want to use are the first two contexts that you have. I think it might be a good idea to say (e.g., tip) something like `If you have more than two clusters in the context list and you want to configure your mesh using clusters other than the first two, you will need to manually set the environment variables to the appropriate context names.`

vadimeisenbergibm: @frankbu Thanks, I have added this tip in https://github.com/istio/istio.io/pull/4073/commits/ea3462be3c7840fcb5e14e7e77e2dfab004f2814.

content/boilerplates/kubectl-multicluster-contexts.md
```diff
@@ -10,9 +12,16 @@
...
+    $ export CTX_CLUSTER2=$(kubectl config view -o jsonpath='{.contexts[1].name}')
+    $ echo CTX_CLUSTER1 = ${CTX_CLUSTER1}, CTX_CLUSTER2 = ${CTX_CLUSTER2}
+    CTX_CLUSTER1 = cluster1, CTX_CLUSTER2 = cluster2
     {{< /text >}}
+
+{{< tip >}}
```
frankbu: Maybe this should be indented 4 spaces too, since it's associated with the above step.

vadimeisenbergibm: @frankbu Done in https://github.com/istio/istio.io/pull/4073/commits/6f4f0c9e0cdd9df98f89523ecb36475206d0e7eb.

content/boilerplates/kubectl-multicluster-contexts.md
```diff
@@ -1,7 +1,9 @@
 ---
 ---
-* The `kubectl` command is used to access both the `cluster1` and `cluster2` clusters with the `--context` flag.
-  Use the following command to list your contexts:
+You can use the `kubectl` command to access both the `cluster1` and `cluster2` clusters with the `--context` flag,
```
frankbu: This needs to start with a `*` or it breaks the flow of the item list in the `before you begin` sections of the owning document. I think something like this should work:
```
*   You can use the `kubectl` command to access both the `cluster1` and `cluster2` clusters with the `--context` flag,
    for example `kubectl get pods --context cluster1`.
    Use the following command to list your contexts:
```

vadimeisenbergibm: @frankbu Fixed in https://github.com/istio/istio.io/pull/4073/commits/5639f6765a32d51a915593ba85e9e0ab17462ccc.

PR #4094, opened 2019-05-04
zh: sync release notes

fleeto:


fleeto:
`[trans-bot:N/A]`

Thank you @fleeto, I can only process the PR with 1 file included, will not be reported to the task issues.

PR #4105, opened 2019-05-07
[WIP] Experiment with snippets

frankbu:
Trying out some ways to move config and command fragments out of the docs and into separate files that can be used for testing.

Snippets can be used to move the fragments to an external test file, but there are several issues (see inline comments). Seems that a test framework will need to be able to understand the doc text block formatting rules (e.g., first line of a bash fragment starts with `$`)


content/docs/tasks/traffic-management/egress/egress-sample.txt
```diff
@@ -0,0 +1,23 @@
+# $snippet SNIP1
```
frankbu: A test program can't just run this example, because only the first line is a command. The remaining lines are expected output.

content/docs/tasks/traffic-management/egress/egress-sample.txt
```diff
@@ -0,0 +1,23 @@
...
+$ kubectl exec -it $SOURCE_POD -c sleep -- curl -I https://www.google.com | grep  "HTTP/"; kubectl exec -it $SOURCE_POD -c sleep -- curl -I https://edition.cnn.com | grep "HTTP/"
+command terminated with exit code 35
+command terminated with exit code 35
+# $endsnippet
+
+# $snippet SNIP2
```
frankbu: This snippet is a proper command, but the first line has an extra `$` character at the beginning, so this one also can't just be run as a test command.

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -82,7 +82,7 @@ To confirm there is a conflict, check whether the `STATUS` field in the output o
 is set to `CONFLICT` for your service. For example:
 
 {{< text bash >}}
-$ istioctl authn tls-check httpbin.default.svc.cluster.local
+$ istioctl authn tls-check istio-ingressgateway-6b454db869-5zn2j.istio-system httpbin.default.svc.cluster.local
```
frankbu: This is an actual bug in the current docs. It's an example of what we're trying to avoid with testing. The `istioctl authn tls-check` command syntax has changed to require an extra source pod argument, but this doc wasn't updated.

But, since the command requires the running instance name in the syntax, moving the command to a snippet wouldn't be a command that a test could run directly. The test would need to preprocess the snippet to change the pod name to the running one.

content/docs/tasks/traffic-management/egress/egress-sample.txt
geeknoid: So we need a mechanism to capture test output and put into the site? Kind of makes my head hurt, not sure how to wire that up.

content/docs/tasks/traffic-management/egress/egress-sample.txt
geeknoid: Oh, I hadn't thought about the $ stuff. Not sure how to handle that :-(

geeknoid:
Don't need this anymore.

PR #4113, opened 2019-05-11
1.1.6 release notes

duderino:


content/boilerplates/notes/1.1.6.md
```diff
@@ -0,0 +1,11 @@
+## Bug fixes
+
+- Fix Galley Helm charts so that the `validatingwebhookconfiguration` object can now deployed to a namespace other than `istio-system` ([Issue 13625](https://github.com/istio/istio/issues/13625)).
+- Additional Helm chart fixes for anti-affinity support: fix `gatewaypodAntiAffinityRequiredDuringScheduling` and `podAntiAffinityLabelSelector` match expressions and fix the default value for `podAntiAffinityLabelSelector` ([Issue 13892](https://github.com/istio/istio/issues/13892)).
+- Make Pilot handle a condition where Envoy continues to request routes for a deleted gateway while listeners are still draining ([Issue 13739](https://github.com/istio/istio/issues/13739)).
+
+## Small enhancements
+
+- Add access logs for the `passthrough` listener.
```
linsun: Access logs aren’t enabled by default right?  Would be good to explain why a user would need to care about this.   I assume it is for helping with debugging.

PR #4117, opened 2019-05-13
Remove `kubectl create` tip in Authentication Policy task

wxdao:
The command should be `kubectl create`.

### DIFF
--- a/content/docs/tasks/security/authn-policy/index.md
+++ b/content/docs/tasks/security/authn-policy/index.md
@@ -729,10 +729,6 @@ spec:
 EOF
 {{< /text >}}
 
-{{< tip >}}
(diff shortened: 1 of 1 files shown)


content/docs/tasks/security/authn-policy/index.md
```diff
@@ -730,7 +730,7 @@ EOF
 {{< /text >}}
 
 {{< tip >}}
-Use `istio create` if the `jwt-example` policy hasn't been submitted.
+Use `kubectl create` if the `jwt-example` policy hasn't been submitted.
```
frankbu: A better fix would be to remove this tip completely (delete lines 732-735), because `kubectl apply` works in all cases (it creates if necessary).

wxdao: I agree too.

PR #4126, opened 2019-05-15
Experiment with showing the search bar all the time on large-enough screens.

geeknoid:
Here's what it looks like if we make the search box visible at all times on wide screens. Sadly, this ends up showing the "Google Custom Search" text all the time, which I find very distracting and inappropriate for  an OSS project....

WDYT?


frankbu:
Yeah, I guess this doesn't look too good. It also seems to be missing (covering?) the other menu items, but maybe that's just your quick demo?

geeknoid:
Oops, I didn't move over the header links enough to make room for the gear.

PR #4151, opened 2019-05-17
zh: Update index.md

maonaoda:
This DestinationRule's  namespace should be istio-system.


fleeto:
`[trans-bot:N/A]`

Thank you @maonaoda, the [related task issue](https://github.com/servicemesher/istio-official-translation/issues/927) has no assignee. will not be reported to the task issues

maonaoda:
@fleeto   zh seems too old.

PR #4152, opened 2019-05-17
Update index.md

maonaoda:
This DestinationRule's  namespace should be default.

### DIFF
--- a/content/docs/tasks/security/authn-policy/index.md
+++ b/content/docs/tasks/security/authn-policy/index.md
@@ -236,7 +236,6 @@ apiVersion: networking.istio.io/v1alpha3
 kind: DestinationRule
 metadata:
  name: "api-server"
- namespace: istio-system
 spec:
  host: "kubernetes.default.svc.cluster.local"
  trafficPolicy:


content/docs/tasks/security/authn-policy/index.md
```diff
@@ -236,7 +236,6 @@ apiVersion: networking.istio.io/v1alpha3
 kind: DestinationRule
 metadata:
  name: "api-server"
- namespace: istio-system
```
frankbu: This change looks wrong. Putting the DestinationRule in the `istio-system` namespace is best. Then it applies to services in all namespaces.

As documented  [here](https://istio.io/docs/tasks/security/authn-policy/#globally-enabling-istio-mutual-tls):
> Starting with Istio 1.1, only destination rules in the client namespace, server namespace and global namespace (default is istio-system) will be considered for a service, in that order.

maonaoda: Yes,this is a different place from the zh version. The zh version is missing the description you  linked, and DestinationRule is placed in the default.

PR #4166, opened 2019-05-18
Move content from Examples to Tasks.

rcaballeromx:
Moves the content found in examples/advanced-gateways/ to
tasks/traffic-management/edge-traffic and
the content found in examples/multicluster/ to tasks/multicluster/
Fixes all broken links caused by the move and adds aliases to the moved pages.
The changes are applied to both, English and Chinese, websites.


frankbu:
The moved examples look great, but now I think the Traffic Management tasks could be better organized. The two subfolders `Securing Ingress Gateway` and `Edge Traffic Management` could be renamed to `Ingress Traffic Control` and `Egress Traffic Control` and then we could put all the ingress and egress tasks into the corresponding folders.

We can do that in this or a followup PR. WDYT?

rcaballeromx:
@frankbu I agree that's a great way to group the related tasks.

PR #4187, opened 2019-05-18
zh：http-proxy

loverto:
update


fleeto:
`[trans-bot:N/A]`

Thank you @loverto, the [related task issue](https://github.com/servicemesher/istio-official-translation/issues/937) has no assignee. will not be reported to the task issues

PR #4190, opened 2019-05-19
zh：collecting-logs

loverto:
udpate


fleeto:
`[trans-bot:N/A]`

Thank you @loverto, I can only process the PR with 1 file included, will not be reported to the task issues.

PR #4229, opened 2019-05-27
Add ha-demo example to show how istio behaves for ha

c0d1ngm0nk3y:
New simple demo which shows that istio is highly available. This demo can later also be used to show that istio can be updated without downtime for the client.

Possible enhancements:
* We could route to several services (round robin)  instead of only one.

### DIFF
--- a/content/docs/examples/ha-demo/index.md
+++ b/content/docs/examples/ha-demo/index.md
@@ -1,0 +1,99 @@
+---
+title: HA-Demo
+description: Deploys a sample application in a high available fashion to show that Istio features are high available as well.
+weight: 10
+aliases:
+    - /docs/samples/ha-demo.html
+    - /docs/guides/ha-demo/index.html
+---
+
+This example deploys a small sample application composed of two components used
+to demonstrate Istio features in a high available fashion. The service shows the latest news article.
+
+The HA-demo application consists of two components:
+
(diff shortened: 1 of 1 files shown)


c0d1ngm0nk3y:
Do you think this kind of demo makes sense? We created it for our internal tests to check if istio can be used in a ha environment.

c0d1ngm0nk3y:
see https://github.com/istio/istio/pull/14387

jsPlachetka:
I give my consent.

frankbu:
This does not look like an appropriate Istio example since it doesn't configure things in the "Istio way". Maybe making it a blog entry would be a better fit, but even then maybe it should be an Envoy, instead of Istio, blog.

An alternative would be to do something similar with proper Istio configuration, and ideally using existing Istio services like bookinfo, helloworld, httpbin, or similar.

c0d1ngm0nk3y:
@frankbu The alternative with a blog post sounds like a good idea. Just out of curiosity, how would the "istio way" look like? There is a `VirtualService` that allows the sidecar to re-route the traffic from news->article. Sounds like istio to me. Please let me know what should be done differently. The `article` in front is needed imho to free the client of the burden to restart in case of a new istio just because of the sidecar.

I don't think it is that easy to achieve ha with the bookinfo app, is it? Especially to be able to update istio in a ha manner. That is what we are checking with that.

c0d1ngm0nk3y:
see #14387 - I will close this PR since the effort for rounding this up is questionable because there is no need for testing ha features.

PR #4238, opened 2019-05-30
Update secure naming in security concept page

wattli:


wattli:
@wenchenglu please take a look, thanks

content/docs/concepts/security/index.md
```diff
@@ -287,6 +287,10 @@ When a client calls the `datastore` service, it extracts the `test-team` identit
 and checks whether `test-team` is allowed to run `datastore` with the secure naming information.
 The client detects that `test-team` is **not** allowed to run the `datastore` service and the authentication fails.
 
+Secure naming is able to protect against general network hijackings for HTTPS traffic. However, it would fail to work
```
wenchenglu: Please cite network hijacking examples: 
... against network hijackings (i.e. DNS spoofing, BGP/route hijacking, ARP spoofing, etc.) for HTTPS traffic. It can also protect TCP traffic from network hijackings except DNS spoofing.

content/docs/concepts/security/index.md
```diff
@@ -287,6 +287,10 @@ When a client calls the `datastore` service, it extracts the `test-team` identit
 and checks whether `test-team` is allowed to run `datastore` with the secure naming information.
 The client detects that `test-team` is **not** allowed to run the `datastore` service and the authentication fails.
 
+Secure naming is able to protect against general network hijackings for HTTPS traffic. However, it would fail to work 
+for TCP traffic if the attacker hijacks the DNS and modifies the IP address of the destination. This is because TCP 
+traffic does not contain the hostname information and we can only rely on the IP address for routing.    
+
```
wenchenglu: And this DNS hijack can happen even before the client-side Envoy receives the traffic.

content/docs/concepts/security/index.md
```diff
@@ -287,6 +287,10 @@ When a client calls the `datastore` service, it extracts the `test-team` identit
 and checks whether `test-team` is allowed to run `datastore` with the secure naming information.
```
wenchenglu: I cannot comment on previous line in place, here is the line I would suggest to change "Suppose the malicious user successfully hacked the discovery service or DNS to map the `datastore` service name to the forged server." ->
Suppose the malicious user successfully hijacked (through DNS spoofing, BGP/route hijacking, ARP spoofing, etc.) the traffic sent to the `datastore` and redirect it to the forged server.

myidpt:
Not related to this PR. But maybe we can investigate if it's possible to use DNS Sec to provide secure DNS information, as a security enhancement for TCP traffic.

content/docs/concepts/security/index.md
wattli: Done.

content/docs/concepts/security/index.md
wattli: Done

content/docs/concepts/security/index.md
wattli: Done

PR #4240, opened 2019-05-30
Add a link to a guide for configuring k8s auth method on Vault [release 1.1 branch]

lei-tang:
Add a link to a guide for configuring k8s auth method on Vault, which includes detailed examples of configuring Vault to authenticate and authorize Kubernetes service accounts.

### DIFF
--- a/content/docs/tasks/security/vault-ca/index.md
+++ b/content/docs/tasks/security/vault-ca/index.md
@@ -105,6 +105,8 @@ certificate signing requests to Vault.
     on the testing Vault CA.
     To learn more about configuring a Vault CA for Kubernetes authentication and authorization,
     visit the [Vault Kubernetes `auth` method reference documentation](https://www.vaultproject.io/docs/auth/kubernetes.html).
+    The guide in [the link](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) includes
+    detailed examples of configuring Vault to authenticate and authorize Kubernetes service accounts.
 
     {{< text bash >}}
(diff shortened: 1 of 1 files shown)


lei-tang:
The counterpart on master branch: https://github.com/istio/istio.io/pull/4241.

myidpt:
Thanks, Lei. Can we have two sections in this task: section one is what you have, which uses your testing Vault server; section two is instructing users how to set up their own Vault server. I think it's still missing in this task. We don't need to go through all the details, but we should give a story, and give pointers to references of configuration (like the link you gave) to tell users how to set it up by themselves. With this, the users have good reason to believe they can build a secure and solid identity management system with Vault.

Does this make sense?

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -105,6 +105,8 @@ certificate signing requests to Vault.
     on the testing Vault CA.
     To learn more about configuring a Vault CA for Kubernetes authentication and authorization,
     visit the [Vault Kubernetes `auth` method reference documentation](https://www.vaultproject.io/docs/auth/kubernetes.html).
+    The guide in [the link](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) includes
```
geeknoid: The [Integration Kuberentes with Vault - auth](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) post includes....

lei-tang: Done.

PR #4255, opened 2019-06-05
Injection caveats docs

### LINKED ISSUE #4253: Docs should include additional caveats around sidecar injection
Docs for sidecar injection should mention the following two caveats:
* Sidecar injection seems to be impossible in `kube-system` and `public` namespaces, as per the declaration of `ignoredNamespaces` [here](https://github.com/istio/istio/blob/13cb3422ae47f95dfbc829a327724a56929cab3b/pilot/pkg/kube/inject/initializer.go#L28) and the call to `injectRequired` [here](https://github.com/istio/istio/blob/13cb3422ae47f95dfbc829a327724a56929cab3b/pilot/pkg/kube/inject/webhook.go#L558).
* Sidecar injection doesn't work for pods on the host network, as per code [here](https://github.com/istio/istio/blob/13cb3422ae47f95dfbc829a327724a56929cab3b/pilot/pkg/kube/inject/inject.go#L403)

Both these facts seem to be undocumented. Happy to provide a PR tomorrow.

Niksko:
Closes #4253

### DIFF
--- a/content/docs/ops/setup/injection/index.md
+++ b/content/docs/ops/setup/injection/index.md
@@ -11,14 +11,19 @@ time. Injection can be scoped to particular sets of namespaces using
 the webhooks `namespaceSelector` mechanism. Injection can also be
 enabled and disabled per-pod with an annotation.
 
-Whether or not a sidecar is injected is depends on three pieces of configuration:
+Whether or not a sidecar is injected is depends on three pieces of configuration and two limitations:
 
+Configuration:
 * webhooks `namespaceSelector`
 * default `policy`
 * per-pod override annotation
 
+Limitations:
+* sidecars cannot be injected in the `kube-system` or `kube-public` namespaces
+* sidecars cannot be injected into pods that use the host network
+
 The following truth table shows the final injection status based on
-these three variables.
+the three configuration items. The limitations above cannot be overridden.
 
 | `namespaceSelector` match | default `policy` | Pod override annotation `sidecar.istio.io/inject` | Sidecar injected? |
 |---------------------------|------------------|---------------------------------------------------|-----------|
@@ -39,6 +44,16 @@ these three variables.
 
 This includes an injected sidecar when it wasn't expected and a lack
 of injected sidecar when it was.
+
+1. Ensure your pod is not in the `kube-system` or `kube-public` namespace.
+   Automatic sidecar injection cannot be used in these namespaces.
+
+1. Ensure your pod does not have `hostNetwork: true` in its PodSpec.
+   Automatic sidecar injection cannot be used with pods that are on the host network.
+   
+   The sidecar model assumes that the iptables changes required for Envy to intercept
+   traffic are within the pod. For pods on the host network this assumption is violated,
+   and this can lead to routing failures at the host level.
 
 1. Check the webhook's `namespaceSelector` to determine whether the
    webhook is scoped to opt-in or opt-out for the target namespace.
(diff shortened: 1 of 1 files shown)


content/docs/ops/setup/injection/index.md
```diff
@@ -40,6 +45,16 @@ these three variables.
 This includes an injected sidecar when it wasn't expected and a lack
 of injected sidecar when it was.
 
+1. Ensure your pod is not in the `kube-system` or `kube-public` namespace.
+   Automatic sidecar injection cannot be used in these namespaces.
+
+1. Ensure your pod does not have `hostNetwork: true` in its pod spec.
+   Automatic sidecar injection cannot be used with pods that are on the host network.
+
+   The sidecar model assumes that the iptables changes required for Envy to intercept
```
frankbu: Envoy

content/docs/ops/setup/injection/index.md
```diff
@@ -11,14 +11,19 @@ time. Injection can be scoped to particular sets of namespaces using
 the webhooks `namespaceSelector` mechanism. Injection can also be
 enabled and disabled per-pod with an annotation.
 
-Whether or not a sidecar is injected is depends on three things:
+Whether or not a sidecar is injected depends on three pieces of configuration and two limitations:
```
linsun: I would not position is as 2 limitations... it is something by design and don't want to confuse users.

content/docs/ops/setup/injection/index.md
```diff
@@ -11,14 +11,19 @@ time. Injection can be scoped to particular sets of namespaces using
...
 
+Configuration:
 * webhooks `namespaceSelector`
 * default `policy`
 * per-pod override annotation
 
+Limitations:
+* sidecars cannot be injected in the `kube-system` or `kube-public` namespaces
```
linsun: what about istio-system namespace?

content/docs/ops/setup/injection/index.md
```diff
@@ -40,6 +45,16 @@ these three variables.
 This includes an injected sidecar when it wasn't expected and a lack
 of injected sidecar when it was.
 
+1. Ensure your pod is not in the `kube-system` or `kube-public` namespace.
+   Automatic sidecar injection cannot be used in these namespaces.
```
frankbu: Can you be more clear about the limitation, e.g., what would happen if you turn on an annotation? Does Istio ignore it, or will very bad things happen :-) Saying something like `Automatic sidecar injection configuration is ignored in these namespaces.` would be less mysterious IMO.

content/docs/ops/setup/injection/index.md
sdake: instead of two limitations, you might try `two security rules`.

tangiel:
@googlebot rescan

content/docs/ops/setup/injection/index.md
Niksko: My reading of the code didn't show any limitations on `istio-system`, but happy to add it

content/docs/ops/setup/injection/index.md
```diff
@@ -11,14 +11,19 @@ time. Injection can be scoped to particular sets of namespaces using
 the webhooks `namespaceSelector` mechanism. Injection can also be
 enabled and disabled per-pod with an annotation.
 
-Whether or not a sidecar is injected is depends on three things:
+Whether or not a sidecar is injected depends on three pieces of configuration and two security rules:
 
+Configuration:
```
frankbu: These lists are displaying correctly in the generated preview: https://deploy-preview-4255--preliminary-istio.netlify.com/

I think Hugo only accepts `-` (not `*`) for list items, so it needs to be:
```
Configuration:
- webhooks `namespaceSelector`
- default `policy`
- per-pod override annotation
```

Niksko: Turns out it needs whitespace around it

PR #4262, opened 2019-06-05
Change RBAC v1 methods field usage doc

pitlv2109:
We will need to update the API for the Reference page as well.

### DIFF
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -628,9 +628,8 @@ Each rule has the following standard fields:
 - **`services`**: A list of service names. You can set the value to `*` to
   include all services in the specified namespace.
 
-- **`methods`**: A list of HTTP method names, for permissions on gRPC requests,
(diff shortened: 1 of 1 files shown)


content/docs/concepts/security/index.md
```diff
@@ -628,9 +628,8 @@ Each rule has the following standard fields:
 - **`services`**: A list of service names. You can set the value to `*` to
   include all services in the specified namespace.
 
-- **`methods`**: A list of HTTP method names, for permissions on gRPC requests,
-  the HTTP verb is always `POST`. You can set the value to `*` to include all
-  HTTP methods.
+- **`methods`**: A list of HTTP methods. You can set the value to `*` to include all
+  HTTP methods. This field should not be used for TCP and gRPC services.
```
liminw: NIT: "This field should not be set for TCP and gRPC services. "

used => set.

pitlv2109: Done. Thanks!

PR #4273, opened 2019-06-06
Root transition announcement.

myidpt:
https://github.com/istio/istio/issues/14516

PR #4282, opened 2019-06-07
Patch #4273 to 1.0 branch

wattli:


geeknoid:
The 1.0 bramch is archived and we don't update anything there.

PR #4285, opened 2019-06-07
Patch #4273 to master branch

wattli:
@geeknoid , the difference between this one and #4273 is that I use 1.x.8 (x can be 0 or 1) instead of 1.1.8 since it affects both 1.0 and 1.1 

Everything else is the same.


geeknoid:
The 1.x.8 thing is tripping up the spell checker, and wouldn't be correct for 1.2.8 in the future. So please replace all uses with "1.0.8, 1.1.8, or later"

wattli:
@geeknoid  Done.

PR #4291, opened 2019-06-09
Document the Helm whitespace problem

sdake:
Document the helm whitespace command line problem.  We should consider
producing a list of known Helm problems as they relate to Istio - post
1.2 (with a quick backport to all of the branches including deprecated
Istio versions).  It will take some time to generate this list, but we
could replace this 1 warning with all of the various Helm issues we
see in production usage along with the upstream tracking bug.


sdake:
upstream issue filed: https://github.com/helm/helm/issues/5863

Issue #4309, opened 2019-06-10
CRD count is off in boilerplate for 1.2


sdake:
This file needs a revision:
https://github.com/istio/istio.io/blob/master/content/boilerplates/verify-crds.md

A bunch of `CRDs` were removed from 1.2 IIRC relating to mixer.

sdake:
@morvencao can you take assignment on this one?

morvencao:
@sdake Sure, let me check that.

PR #4311, opened 2019-06-10
update expected CRDs number.

morvencao:
Resolve: https://github.com/istio/istio.io/issues/4309


richardwxn:
Shouldn't it be 23 for the number without cert-manager?

morvencao:
@richardwxn Thanks for reminder, I double checked the code, found that `sidecar` CRD exists both in [crd-10.yaml](https://github.com/istio/istio/blob/1.2.0-rc.0/install/kubernetes/helm/istio-init/files/crd-10.yaml#L152-L171) and [crd-11.yaml](https://github.com/istio/istio/blob/1.2.0-rc.0/install/kubernetes/helm/istio-init/files/crd-11.yaml) 

Will raise another PR to fix this.

PR #3382, opened 2019-02-26
Remove unnecessary section to turn on permissive

wattli:

PR #4316, opened 2019-06-10
Remove references to "PERMISSIVE" mode

vngzs:
In c4b714ec99954882f8d0d56a5fca2ab922934913 (#3382), `PERMISSIVE` mode
references were removed from the documentation as it is no longer
required.

"PERMISSIVE" mode is not described elsewhere in the docs.

Remove references to this mode in the mainline documentation.

### DIFF
--- a/content/docs/tasks/security/mtls-migration/index.md
+++ b/content/docs/tasks/security/mtls-migration/index.md
@@ -13,16 +13,13 @@ may be desirable to gradually migrate them to Istio. During the migration, some 
 sidecars while some do not. For a service with a sidecar, if you enable
 mutual TLS on the service, the connections from legacy clients (i.e., clients without
 Envoy) will lose communication since they do not have Envoy sidecars and client certificates.
-To solve this issue, Istio authentication policy provides a "PERMISSIVE" mode to solve
-this problem. Once "PERMISSIVE" mode is enabled, a service can take both HTTP
-and mutual TLS traffic.
 
 You can configure Istio services to send mutual
 TLS traffic to that service while connections from legacy services will not
 lose communication. Moreover, you can use the
 [Grafana dashboard](/docs/tasks/telemetry/using-istio-dashboard/) to check which services are
-still sending plain text traffic to the service in "PERMISSIVE" mode and choose to lock
-down once the migration is done.
+still sending plain text traffic to the service and choose to lock down once
+the migration is done.
 
 ## Before you begin
 


content/docs/tasks/security/mtls-migration/index.md
```diff
@@ -13,16 +13,13 @@ may be desirable to gradually migrate them to Istio. During the migration, some
...
-this problem. Once "PERMISSIVE" mode is enabled, a service can take both HTTP
-and mutual TLS traffic.
 
 You can configure Istio services to send mutual
 TLS traffic to that service while connections from legacy services will not
 lose communication. Moreover, you can use the
 [Grafana dashboard](/docs/tasks/telemetry/using-istio-dashboard/) to check which services are
-still sending plain text traffic to the service in "PERMISSIVE" mode and choose to lock
-down once the migration is done.
+still sending plain text traffic to the service and choose to lock down once
```
sdake: lock down how?  permissive is gone?

vngzs: That's the odd bit to me, too. The referenced commit to the documentation, c4b714ec99954882f8d0d56a5fca2ab922934913, removes the last reference to `PERMISSIVE` mode in the documentation. If we look at [https://github.com/istio/istio/](istio/istio).

We can [search the commit history](https://github.com/istio/istio/search?q=permissive&type=Commits) we see that [permissive mode has been enabled in the default deployment](https://github.com/istio/istio/commit/d0841f8d9ad16d1b0b7e576762417cf795fc1751). Disabling mTLS in the Helm chart implies enabling `PERMISSIVE` mode, as it creates the following resource:

```
apiVersion: authentication.istio.io/v1alpha1
kind: MeshPolicy
metadata:
  name: "default"
spec:
  peers:
    - mtls:
        mode: PERMISSIVE
```

Thus, the policy

```
spec:
   targets:
   - name: httpbin
   peers:
   - mtls:
       mode: STRICT
```

should be enough to "lock down" a cluster to `STRICT` mode.

I believe that the Reference should, indeed, describe `PERMISSIVE` mode, but I do not believe that this document is the place for that. This document roughly describes how to convert a cluster with default settings into one with strict mTLS.

If `PERMISSIVE` mode were to be described here, it could be explained at the end, as a potential rollback option.

vngzs:
@sdake jk

content/docs/tasks/security/mtls-migration/index.md
rcaballeromx: I concur. This document should focus on mTLS migration. You can link to the reference doc describing the modes:
/docs/reference/config/istio.authentication.v1alpha1/#MutualTls-Mode

rcaballeromx:
@vngzs We do have the modes documented in the reference docs: /docs/reference/config/istio.authentication.v1alpha1/#MutualTls-Mode

PR #4318, opened 2019-06-10
Update tracing task doc

bianpengyuan:
Change namespace from istio-control back to istio-system. Also update tracing screenshot since policy spans won't be included by default.


content/docs/tasks/telemetry/distributed-tracing/overview/index.md
```diff
@@ -81,8 +81,8 @@ When you make downstream calls in your applications, make sure to include these
 
 ## Trace sampling
 
-Istio captures a trace for all requests by default. For example, when
-using the Bookinfo sample application above, every time you access
+When installing with demo profile, Istio captures a trace for all requests by default.
```
sdake: `Istio captures a trace for all requests by default when installing with the demo profile.`

bianpengyuan: Done.

bianpengyuan:
@douglas-reid @sdake I updated the overview a bit. Please take another look. Thanks!

geeknoid:
You have a few CI failures to take care of.

bianpengyuan:
@geeknoid It's weird that lint thinks [Lightstep](../lightstep/) is not correctly spelled (with the text part highlighted). Changing the relative path to full path makes lint happy.

geeknoid:
We don't support relative paths in the docs, so that was definitely part of the problem.

PR #4339, opened 2019-06-11
Docker Desktop updates

### LINKED ISSUE #4334: Update Docker Desktop page memory requirements
The current Docker Desktop page simply status you may want to increase the memory requires from the default of 2GB saying pilot needs 4GB and showing a picture showing 5GB. 

Per #4293, minikube wants 16GB, so we may want to increase the Docker Desktop requirements as well.

ericvn:
Fixes #4334


ericvn:
There is some discussion in Slack as to the amount of memory to specify. In 1.1, we moved to 8.0 GB, and I haven't seen failures with testing on 1.2. Some of the other platforms call out 16 GB (minikube and IBM Cloud for example).

sdake:
8GB should be sufficient for docker desktop as docker runs in the Host OS.  Docker uses a little north of 2GB of ram, so 16GB on IBM Cloud and minikube makes more sense.

PR #4344, opened 2019-06-11
fix(telemetry docs): replace p&t concept doc with observability doc

douglas-reid:
This PR replaces the existing concept guide for *Policies and Telemetry* with a doc on *Observability*.

The previous doc focused on Mixer and how it works. The new guide attempts to highlight the observability features of Istio at a high-level, without focusing on the implementation details. Policy details are removed to avoid confusion (they are also covered in the *Security* concept guide).

The old `Concepts > Policies and Telemetry` content is moved to the reference section and given a new name: `Reference > Configuration > Policies and Telemetry > Mixer Configuration Model`. The *Reliability and latency * section is removed (as this information is outdated and now covered in the perf-and-scalability sections of the docs).

The *Default Metrics* documentation is also updated to reflect Prometheus metrics names, to make it easier to consume.

NOTE: The *Control Plane Metrics* subsection of the new doc does not currently include links to the documentation on those metrics, as that has not been generated. I plan on working on that in a separate PR.


content/docs/concepts/observability/index.md
```diff
@@ -0,0 +1,159 @@
...
+  source_workload_namespace="default"
+} 214
+```
+
+### Control Plane Metrics
+
+Each Istio component (Pilot, Galley, Mixer) also provides a collection of self-monitoring metrics. These metrics allow monitoring of the behavior
+of Istio itself (as distinct from that of the services within the mesh).
+
+For more information on which metrics are maintained, please the reference docs for each of the components:
```
bianpengyuan: Please see?

content/docs/concepts/observability/index.md
```diff
@@ -0,0 +1,159 @@
...
+
+- Pilot
```
bianpengyuan: Should these be link? Also should citadel be added here?

content/docs/concepts/observability/index.md
geeknoid: Actually, we're not that polite in our docs. That should just be "see the reference docs"

content/docs/concepts/observability/index.md
douglas-reid: yes, but the target of those links doesn't exist yet.

douglas-reid:
closing in favor of #4448

Issue #4348, opened 2019-06-12
503 error after mTLS policy is applied on RBAC groups user guide


lei-tang:
503 error after mTLS policy is applied on RBAC groups user guide: https://preliminary.istio.io/docs/tasks/security/rbac-groups/.

Fix for this issue: a DestinationRule should be applied on sleep after httpbin requires mTLS.

lei-tang:
Fixes for this issue: https://github.com/istio/istio.io/pull/4349, https://github.com/istio/istio.io/pull/4350

pitlv2109:
Close?

PR #4350, opened 2019-06-12
Fix RBAC groups user guide [release-1.1 branch]

lei-tang:
This PR fixes the issue: https://github.com/istio/istio.io/issues/4348

### DIFF
--- a/content/docs/tasks/security/rbac-groups/index.md
+++ b/content/docs/tasks/security/rbac-groups/index.md
@@ -99,7 +99,7 @@ are supported.
     EOF
     {{< /text >}}
 
-1.  Apply a `DestinationRule` policy on `sleep` to use mutual TLS when
+1.  Apply a policy through a destination rule for the `sleep` service to use mutual TLS when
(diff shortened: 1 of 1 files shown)


geeknoid:
@wattli @lei-tang In general, we should not be pushing stuff on an on-going basis to the release branch unless it is an emergency. Just push to master and the fix will be published the next time we release (just like fixing bugs in the software).

Especially now, since 1.2 is a week away.

content/docs/tasks/security/rbac-groups/index.md
```diff
@@ -99,6 +99,23 @@ are supported.
     EOF
     {{< /text >}}
 
+1.  Apply a `DestinationRule` policy on `sleep` to use mutual TLS when
```
rcaballeromx:
```suggestion
1.  Apply a policy through a destination rule for the `sleep` service to use mutual TLS when
```

lei-tang: Done.

PR #4354, opened 2019-06-12
Add clarification on behaviour in absence of policy

### LINKED ISSUE #2542: Default mutual TLS setting is unclear in documentation

Niksko:
This closes #2542.

Seems like previous attempts to add clarification here were hampered by the specific wording around how the mesh could get into such a state, and introduced complications with Helm charts cf. https://github.com/istio/istio.io/pull/2547

I attempted to phrase this in the most neutral possible manner, and hopefully I actually got the behaviour right? Is this clear, or should we omit the example policy and just state that in the absence of a policy both transport and origin auth are disabled?

### DIFF
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -444,18 +444,8 @@ only one authentication policy per mesh and one authentication policy per
 namespace. Istio also requires mesh-wide and namespace-wide policies to have
 the specific name `default`.
 
-If a service has no matching policies, the service behaves as if the following
-`empty-policy` policy applies:
-
-{{< text yaml >}}
-apiVersion: "authentication.istio.io/v1alpha1"
-kind: "Policy"
-metadata:
-  name: "empty-policy"
-spec:
-{{< /text >}}
-
-This policy disables both transport authentication and origin authentication.
+If a service has no matching policies, both transport authentication and
+origin authentication are disabled.
 
 #### Transport authentication
 
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -444,6 +444,19 @@ only one authentication policy per mesh and one authentication policy per
 namespace. Istio also requires mesh-wide and namespace-wide policies to have
 the specific name `default`.
 
+If a service has no matching policies, the service behaves as if the following
+'empty' policy applies:
+
+{{< text yaml >}}
+apiVersion: "authentication.istio.io/v1alpha1"
+kind: "Policy"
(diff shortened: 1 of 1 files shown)


content/docs/concepts/security/index.md
```diff
@@ -444,6 +444,19 @@ only one authentication policy per mesh and one authentication policy per
 namespace. Istio also requires mesh-wide and namespace-wide policies to have
 the specific name `default`.
 
+If a service has no matching policies, the service behaves as if the following
+'empty' policy applies:
```
rcaballeromx:
```suggestion
`empty-policy` applies:
```

content/docs/concepts/security/index.md
```diff
@@ -444,6 +444,19 @@ only one authentication policy per mesh and one authentication policy per
...
+
+{{< text yaml >}}
+apiVersion: "authentication.istio.io/v1alpha1"
+kind: "Policy"
+metadata:
+  name: "empty-policy"
+spec:
+{{< /text >}}
+
+This disables both transport authentication and origin authentication.
```
rcaballeromx:
```suggestion
This policy disables both transport authentication and origin authentication.
```

Niksko:
@rcaballeromx ready for another review please

content/docs/concepts/security/index.md
```diff
@@ -444,6 +444,19 @@ only one authentication policy per mesh and one authentication policy per
 namespace. Istio also requires mesh-wide and namespace-wide policies to have
 the specific name `default`.
 
+If a service has no matching policies, the service behaves as if the following
+`empty-policy` policy applies:
```
frankbu: I don't think this `empty-policy` example is helpful, since it's not even valid. Istio would reject it, because its name must be `default`. Isn't it sufficient to just say that both transport authentication and origin authentication are disabled by default?

Niksko: Yep, agreed. Hopefully the sections on transport and origin auth are clear enough that it is obvious that omitting them both is equivalent to omitting the policy

PR #4356, opened 2019-06-12
Add clarity to transport auth section, inc new mode param

Niksko:
Arguably the wording here before was incorrect, because the mtls
parameter does have an argument, the mode parameter. This documents
STRICT and PERMISSIVE modes, as well as discussing the equivalence
between STRICT mode and omission of the mode key. It also adds clarity
as to what happens when the section is omitted.

### DIFF
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -474,7 +474,7 @@ peers:
 
 The mutual TLS setting has an optional `mode` parameter that defines the
 strictness of the peer transport authentication. These modes are documented
-in the [Authentication Policy reference document](/docs/reference/config/istio.authentication.v1alpha1/#MutualTls).
+in the [Authentication Policy reference document](/docs/reference/config/istio.authentication.v1alpha1/#MutualTls-Mode).
 
 {{< tip >}}
 The default mutual TLS mode is `STRICT`. Therefore, `mode: STRICT` is equivalent
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -490,9 +490,6 @@ When you do not specify a mutual TLS mode, peers cannot use transport
 authentication, and Istio rejects mutual TLS connections bound for the sidecar.
 At the application layer, services may still handle their own mutual TLS sessions.
 
-In the future, the mutual TLS setting may carry additional arguments to
-provide different mutual TLS implementations.
-
 #### Origin authentication
 
 The `origins:` section defines authentication methods and associated parameters
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -477,8 +477,7 @@ strictness of the peer transport authentication. These modes are documented
 in the [Authentication Policy reference document](/docs/reference/config/istio.authentication.v1alpha1/#MutualTls-Mode).
 
 {{< tip >}}
-The default mutual TLS mode is `STRICT`. Therefore, `mode: STRICT` is equivalent
-to all of the following:
+The default mutual TLS mode is `STRICT`. Therefore, `mode: STRICT` is equivalent to all of the following:
 
 - `- mtls: {}`
 - `- mtls:`
--- a/content/docs/concepts/security/index.md
+++ b/content/docs/concepts/security/index.md
@@ -462,8 +462,7 @@ The `peers:` section defines the authentication methods and associated
 parameters supported for transport authentication in a policy. The section can
 list more than one method and only one method must be satisfied for the
 authentication to pass. However, as of the Istio 0.7 release, the only
-transport authentication method currently supported is mutual TLS. If you do not
-need transport authentication, skip this section entirely.
+transport authentication method currently supported is mutual TLS.
 
 The following example shows the `peers:` section enabling transport
 authentication using mutual TLS.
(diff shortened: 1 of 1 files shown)


content/docs/concepts/security/index.md
```diff
@@ -461,10 +460,24 @@ peers:
...
+is identical to `-mtls: {}`, `- mtls:` or `- mtls: null`
+
+- `mode: PERMISSIVE` allows peers to use mututal TLS while still allowing
+peers to not use transport authentication. This mode is useful for transitioning
+existing workloads to peer authentication gradually, as it allows interoperation
+between mTLS and plaintext traffic.
+
+If this section is omitted completely, peers may not use transport
+authentication, and mutual TLS connections bound for the sidecar pod
+will be rejected. However clients are free to handle their own mutual
```
Niksko: Perhaps this wording is confusing. Happy to remove the part about clients handling mTLS, but I thought it might be prudent to hint that this is an option, and to make sure users know that the 'HTTP' language provided by istioctl isn't really HTTP, it's more just 'plaintext' / 'auth how you like'

content/docs/concepts/security/index.md
```diff
@@ -461,10 +460,24 @@ peers:
...
 {{< /text >}}
 
-Currently, the mutual TLS setting doesn't require any parameters. Hence,
-`-mtls: {}`, `- mtls:` or `- mtls: null` declarations are treated the same. In
-the future, the mutual TLS setting may carry arguments to provide different
-mutual TLS implementations.
+The mutual TLS setting has an optional `mode` parameter that defines the
+strictness of the peer transport authentication:
+
+- `mode: STRICT` enforces that all peers must use mutual TLS. This declaration
```
Niksko: Am I over-complicating things or duplicating information by putting this here? Should I just link further up to the section on strict vs permissive, link to the Policy reference docs or link to the mTLS migration guide?

Niksko:
I think the key parts are to correct the statement that there aren't any arguments to the mtls field, clarify the behaviour when you omit the `peer` section, and to explain the equivalence between `mode: STRICT` and the various ways you can leave `mtls` empty

content/docs/concepts/security/index.md
```diff
@@ -461,10 +460,24 @@ peers:
...
+is identical to `-mtls: {}`, `- mtls:` or `- mtls: null`
```
rcaballeromx: It would be better if you link to the reference doc that has all the declarations instead but this is not a deal breaker.

content/docs/concepts/security/index.md
```diff
@@ -461,10 +460,24 @@ peers:
...
+
+- `mode: PERMISSIVE` allows peers to use mutual TLS while still allowing
+peers to not use transport authentication. This mode is useful for transitioning
```
rcaballeromx: Please indent multi-line lists with four (4) spaces. I'm only suggesting here but please fix everywhere.
```suggestion
    peers to not use transport authentication. This mode is useful for transitioning
```

content/docs/concepts/security/index.md
rcaballeromx: The paragraph reads a bit awkwardly, consider:
When you do not specify a mutual TLS mode, peers can't use transport
authentication, and Istio rejects mutual TLS connections bound for the sidecar. 
At the application layer, services can still handle their own mutual TLS sessions.

content/docs/concepts/security/index.md
```diff
@@ -461,10 +460,24 @@ peers:
...
+existing workloads to peer authentication gradually, as it allows interoperation
+between mutual TLS and plaintext traffic.
+
+If this section is omitted completely, peers may not use transport
+authentication, and mutual TLS connections bound for the sidecar pod
+will be rejected. However clients are free to handle their own mutual
+TLS sessions at the application layer.
+
+In the future, the mutual TLS setting may carry additional arguments to
```
rcaballeromx: To avoid confusing readers, please do not describe possible future behavior. Opt instead to update the docs when the new features are implemented.
```suggestion
```

content/docs/concepts/security/index.md
```diff
@@ -461,10 +460,24 @@ peers:
...
+provide different mutual TLS implementations.
```
rcaballeromx:
```suggestion
```

content/docs/concepts/security/index.md
Niksko: I've linked to the reference docs, but I think it's key to stress that `STRICT` is default mode and that if you don't specify a mode it's equivalent to specifying `STRICT` mode

content/docs/concepts/security/index.md
Niksko: Noted, thanks. FYI this was already there, I merely moved it, it seemed strange to me as well.

Niksko: It was also wrong, given that we now do have an additional option: mode

content/docs/concepts/security/index.md
```diff
@@ -461,10 +460,23 @@ peers:
...
-`-mtls: {}`, `- mtls:` or `- mtls: null` declarations are treated the same. In
-the future, the mutual TLS setting may carry arguments to provide different
-mutual TLS implementations.
+The mutual TLS setting has an optional `mode` parameter that defines the
+strictness of the peer transport authentication. These modes are documented
+in the [Authentication Policy reference document](/docs/reference/config/istio.authentication.v1alpha1/#MutualTls-Mode).
+
+{{< tip >}}
+The default mutual TLS mode is `STRICT`. Therefore, `mode: STRICT` is equivalent
+to all of the following:
```
Niksko: Hugo seems to insist on putting a unnecessary `</p>` tag right after the colon here, which causes the site to fail linting. Any ideas as to how to fix this?

Niksko:
Anyone able to help out with the linting errors I'm getting here? Still happening after a rebase. That `</p>` shouldn't be there, but I can't seem to get rid of it.

geeknoid:
Sorry about the problem you're having with Hugo. This is a bug I've reported previously to the Hugo community.

This is probably because of the list in the tip. If you rewrite that content as a normal paragraph instead of a tip, then it should be OK.

Issue #4361, opened 2019-06-12
Remove GDB from the official documentation


sdake:
https://istio.io/help/ops/component-debugging/#with-gdb

gdb is no longer shipped in the containers, and if it were, the sidecar injected containers do not have the proper capabilities.

For more detail see:
https://github.com/istio/istio/issues/8383

PR #4377, opened 2019-06-13
Remove GDB documentation

sdake:
GDB is no longer shipped in the containers.  Whether it should be or not
is not relevant for 1.2 (its too late to add GDB to 1.2).  Remove
GDB from the documentaiton.  We don't really expect operators would
use GDB in production.  If GDB were reintroduced to the containers,
it may make sense to document GDB in the developer wiki.

Fixes: #4361


sdake:
Looks like the URL: https://www.googleapis.com/books/v1/volumes?q=isbn:0486424618 fails often with a transient failure:

```
{
 "error": {
  "errors": [
   {
    "domain": "global",
    "reason": "backendFailed",
    "message": "Service temporarily unavailable.",
    "locationType": "other",
    "location": "backend_flow"
   }
  ],
  "code": 503,
  "message": "Service temporarily unavailable."
 }
}
```

PR #4379, opened 2019-06-13
Fix doc 'istioctl apply' -> 'kubectl apply'

colas-pomies:
Example of code in documentation ask to execute the command `istioctl apply` but it has to be `kubectl apply`


colas-pomies:
I signed it!

PR #4381, opened 2019-06-13
The dotted lines turned solid on import

sdake:
cloudconvert seems to not handle dotted lines.  This has been
fixed.  SVG was minified, although it seems much larger than
the original.


geeknoid:
Have you tried "make opt"? This runs the SVG optimizer on the site content. Maybe that'll shrink it further

PR #4321, opened 2019-06-10
Add new Traffic management concept.

rcaballeromx:
Wrote a new "Traffic management" concept to include the latest developments.
Organized the content into smaller pages to make it easier to consume. The new
structure improves visibility and each page includes keywords to improve the
"See Also" pieces of suggested content and SEO. The new concept includes
diagrams and examples. Each section starts with the larger picture and then
moves to explain the specifics.

Signed-off-by: rcaballeromx <grca@google.com>

PR #4396, opened 2019-06-13
Updated traffic management to point to ingress docs

brian-avery:
Link directly to the ingress gateways documentation on finding the host instead of going through bookinfo to get there.


rcaballeromx:
@brian-avery PTAL at #4321 And comment there to avoid conflicts.

PR #4397, opened 2019-06-13
fix the egress gateway example for multi cluster gateways

linsun:
need to change static because one address was host the other was ip.

It is possible (non-default scenario when not following our example) that the cluster2_gw_addr is a hostname so added a note for that too.

### DIFF
--- a/content/docs/tasks/multicluster/gateways/index.md
+++ b/content/docs/tasks/multicluster/gateways/index.md
@@ -144,9 +144,10 @@ use the following service entry for `httpbin.bar` instead of the one in the prev
 
 {{< tip >}}
 The egress gateway used in this configuration cannot also be used for other, non inter-cluster, egress traffic.
+{{< /tip >}}
+
 
 If `$CLUSTER2_GW_ADDR` is an IP address, use the `$CLUSTER2_GW_ADDR - IP address` option.  If `$CLUSTER2_GW_ADDR` is a hostname, use the `$CLUSTER2_GW_ADDR - hostname` option.
-{{< /tip >}}
 
 {{< tabset cookie-name="profile" >}}
 
--- a/content/docs/tasks/multicluster/gateways/index.md
+++ b/content/docs/tasks/multicluster/gateways/index.md
@@ -145,19 +145,21 @@ use the following service entry for `httpbin.bar` instead of the one in the prev
 {{< tip >}}
 The egress gateway used in this configuration cannot also be used for other, non inter-cluster, egress traffic.
 
-If $CLUSTER2_GW_ADDR is an IP address, use the `$CLUSTER2_GW_ADDR - IP address` option.  If $CLUSTER2_GW_ADDR is a hostname, use the `$CLUSTER2_GW_ADDR - hostname` option.
+If `$CLUSTER2_GW_ADDR` is an IP address, use the `$CLUSTER2_GW_ADDR - IP address` option.  If `$CLUSTER2_GW_ADDR` is a hostname, use the `$CLUSTER2_GW_ADDR - hostname` option.
 {{< /tip >}}
 
 {{< tabset cookie-name="profile" >}}
 
 {{< tab name="$CLUSTER2_GW_ADDR - IP address" cookie-value="option1" >}}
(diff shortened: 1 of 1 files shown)


content/docs/tasks/multicluster/gateways/index.md
```diff
@@ -139,42 +139,79 @@ running in a second cluster.
...
-    ports:
-      http1: 15443
-EOF
-{{< /text >}}
+1. Export the `cluster1` egress gateway address:
+    {{< text bash >}}
+    export CLUSTER1_EGW_ADDR=$(kubectl get --context=$CTX_CLUSTER1 svc --selector=app=istio-egressgateway \
+        -n istio-system -o yaml -o jsonpath='{.items[0].spec.clusterIP}')
+    {{< /text >}}
+
```
sdake: might add a tip callout here indicating option 1 is for systems that use IP addresses for their gateway ip, option 2 for systems that use DNS for their gateway ip.

content/docs/tasks/multicluster/gateways/index.md
```diff
@@ -139,42 +139,79 @@ running in a second cluster.
...
+        network: external
+        ports:
+          http1: 15443 # Do not change this port value
+      - address: ${CLUSTER1_EGW_ADDR}
+        ports:
+          http1: 15443
+    EOF
+    {{< /text >}}
+    
+    If the `${CLUSTER2_GW_ADDR}` is a hostname, you can use `resolution: DNS` for the endpoint resolution:
```
sdake: might change this to option 2

(using a numbered list) and add the tip box at the top of the section as peeps may cut and paste (incorrectly) the first box when they need to paste the second (on AWS afaik, but possibly other cloud platforms).

linsun:
@sdake I tried to addr your comment, PTAL.  The lint failure is unrelated to my change IIUC.

content/docs/tasks/multicluster/gateways/index.md
```diff
@@ -139,14 +139,61 @@ running in a second cluster.
...
+    protocol: http
+  resolution: STATIC
+  addresses:
+  - 127.255.0.2
+  endpoints:
+  - address: ${CLUSTER2_GW_ADDR}
+    network: external
+    ports:
+      http1: 15443 # Do not change this port value
+  - address: ${CLUSTER1_EGW_ADDR}
```
frankbu:
```suggestion
  - address: ${CLUSTER1_GW_ADDR}
```

content/docs/tasks/multicluster/gateways/index.md
```diff
@@ -139,14 +139,61 @@ running in a second cluster.
...
 
-If you want to route traffic from `cluster1` via a dedicated
-egress gateway, instead of directly from the sidecars,
+If you want to route traffic from `cluster1` via a dedicated egress gateway, instead of directly from the sidecars,
 use the following service entry for `httpbin.bar` instead of the one in the previous section.
 
 {{< tip >}}
 The egress gateway used in this configuration cannot also be used for other, non inter-cluster, egress traffic.
+
+If `$CLUSTER2_GW_ADDR` is an IP address, use the `$CLUSTER2_GW_ADDR - IP address` option.  If `$CLUSTER2_GW_ADDR` is a hostname, use the `$CLUSTER2_GW_ADDR - hostname` option.
```
frankbu: This should come after the tip, not inside it.

frankbu: Is there more to consider than that?
^^^ @rshriram

rshriram: Yeah. this is not a tip. this is essentially two slightly different service entries.

content/docs/tasks/multicluster/gateways/index.md
linsun: I think CLUSTER1_EGW_ADDR is correct because it is the egress gateway ip.

frankbu: Sorry, my mistake.

linsun:
all comments are addressed, merging this.  lint errors aren't related.

Issue #4327, opened 2019-06-11
istio-cni documentation issues


sdake:
As a new consumer of CNI, I took notes during my first interaction with the system.  Looks pretty good!  The docs could be refined as follows:

1. Where is the step 3 tarball?  I cloned the repo to obtain the `deployments` directory which contains the helm templates - but this is the wrong approach.
1. The install has a cut and paste, but is missing the crd install step, which may confuse people (it confused me on first attempting this work…)
1. Remove the word ‘Example:”
1. Don’t we always wish to exclude kube-system and kube-public?
1. Reword from: `Not all hosted Kubernetes clusters are created with the kubelet configured to use the CNI plugin so compatibility with this istio-cni solution is not ubiquitous`. Reword to:
`The istio-cni solution is not ubiquitous.  Some platforms, especially hosted Kubernetes environments, do not enable the CNI plugin in the kubelet configuration.`

1. The first row in the hosted table list is confusing.  As I run on bare metal (which is obviously not a hosted platform) - I was misled into thinking that first row applied to my use case.   I’d rework the table into 4  columns.  rename “Required non-default settings” to “Configuration Overrides” and add a new field “Platform Overrides”.  For GKE I’d put “Enable Network Policy.  This allows you to delete the first row entirely.  If no overrides are required, make sure to put the word “none” for the override.  An empty table field could mean many things to many people :)

1. Remove single bullet: https://preliminary.istio.io/docs/setup/kubernetes/additional-setup/cni/#gke-setup.  Change the single bullet into a preceding warning using {{< warning >}} - single bullets don’t make any sense.

1. What does This refer to?  The preceding section?  A little confusing..
```
This should have no effect on the operations performed by the base CNI plugin configuring the pod’s networking setup, although not all CNI’s have been validated.
```

We want people to be able to burn through these docs, not spend 30 minutes (as I did) on first exposure to them.  If they want details, cool, they can read them more slowly.

john-a-joyce:
Regarding the tarball.  That should get a different issue because we need to get the tar ball built and decide where it is pushed, before we can fix the documentation side of things

tiswanso:
The Istio release process toolchain seemed to change and no longer create the istio-cni-*.tgz in the release archive.  Maria created this fix but it looks like it never made it:  https://github.com/istio/istio/pull/13615

tiswanso:
@sdake -- thanks for tracking this stuff.  Some Q's/comments for clarification:

[...]

PR #4398, opened 2019-06-13
Update CNI docs

rlenglet:
Fixes: https://github.com/istio/istio.io/issues/4327

### DIFF
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -68,12 +68,13 @@ The following table shows all the options that the `istio-cni` Helm chart suppor
 ### Excluding specific Kubernetes namespaces
 
 This example uses Helm to perform the following tasks:
-- Install the Istio CNI plugin.
-- Configure its log level.
-- Ignore the pods in the following namespaces:
-    - `istio-system`
-    - `foo_ns`
-    - `bar_ns`
+
+* Install the Istio CNI plugin.
+* Configure its log level.
+* Ignore the pods in the following namespaces:
+    * `istio-system`
+    * `foo_ns`
+    * `bar_ns`
 
 Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) for complete instructions.
 
@@ -186,15 +187,17 @@ Due to this, the `kubelet` process writes the plugin's log entries into its log.
 
 The Istio CNI plugin may cause networking connectivity problems for any application `initContainers`. When using Istio CNI, `kubelet`
 starts an injected pod with the following steps:
+
 1. The Istio CNI plugin sets up traffic redirection to the Istio sidecar proxy within the pod.
 1. All init containers execute and complete successfully.
 1. The Istio sidecar proxy starts in the pod along with the pod's other containers.
 
 Init containers execute before the sidecar proxy starts, which can result in traffic loss during their execution.
 Avoid this traffic loss with one or both of the following settings:
-- Set the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to disable redirecting traffic to any
+
+* Set the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to disable redirecting traffic to any
   CIDRs the init containers communicate with.
-- Set the `traffic.sidecar.istio.io/excludeOutboundPorts` annotation to disable redirecting traffic to the
+* Set the `traffic.sidecar.istio.io/excludeOutboundPorts` annotation to disable redirecting traffic to the
   specific outbound ports the init containers use.
 
 ### Compatibility with other CNI plugins
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -31,7 +31,7 @@ replaces the functionality provided by the `istio-init` container.
 
 1.  Install Kubernetes with the container runtime supporting CNI and `kubelet` configured
     with the main [CNI](https://github.com/containernetworking/cni) plugin enabled via `--network-plugin=cni`.
-    *  Kubernetes installations for AWS EKS, Azure AKS, and IBM Cloud IKS clusters have this capability.
+    *  AWS EKS, Azure AKS, and IBM Cloud IKS clusters have this capability.
     *  Google Cloud GKE clusters require that the
        [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) feature
        is enabled to have Kubernetes configured with `network-plugin=cni`.
@@ -40,78 +40,18 @@ replaces the functionality provided by the `istio-init` container.
     *  The Kubernetes documentation highly recommends this for all Kubernetes installations
        where `ServiceAccounts` are utilized.
 
-1.  [Download the Istio release](/docs/setup/kubernetes/#downloading-the-release).
-
 ## Installation
 
 1.  Determine the Kubernetes environment's CNI plugin `--cni-bin-dir` and `--cni-conf-dir` settings.
+    Refer to [Hosted Kubernetes settings](#hosted-kubernetes-settings) for any non-default settings required.
 
-    1.  Refer to the [Hosted Kubernetes Usage](#hosted-kubernetes-usage) section for any non-default
-        settings required.
-
-1.  Create a namespace for the Istio control-plane components:
-
-    {{< text bash >}}
-    $ kubectl create namespace istio-system
-    {{< /text >}}
-
-1.  Expand the release's compressed Istio CNI chart.
-
-    {{< text bash >}}
-    $ tar -xzf install/kubernetes/helm/charts/istio-cni-*.tgz -C install/kubernetes/helm/
-    {{< /text >}}
-
-1.  Install Istio CNI.
-
-    *  Option 1.  Render and apply the Istio CNI Helm template.
-
-        {{< text bash >}}
-        $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system | kubectl apply -f -
-        {{< /text >}}
-
-    *  Option 2.  Install via Helm and Tiller via `helm install`.
-
-        {{< text bash >}}
-        $ helm install install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system
-        {{< /text >}}
+1.  Install Istio CNI and Istio using Helm.
+    Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) instructions and the **default with CNI enabled** profile.
+    Pass `--set cniBinDir=...` and/or `--set cniConfDir=...` options when installing `istio-cni` if non-default, as determined in the previous step.
 
     {{< tip >}}
-    Refer to the full set of `istio-cni` Helm parameters in the chart's [`values.yaml`](https://github.com/istio/cni/blob/master/deployments/kubernetes/install/helm/istio-cni/values.yaml).
+    Refer to [Helm chart parameters](#helm-chart-parameters) below for the full set of `istio-cni` Helm parameters.
     {{< /tip >}}
-
-1.  Add the `--set istio_cni.enabled=true` setting to the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/) to include the installation of the Istio CNI plugin. For example:
-
-    {{< text bash >}}
-    $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
-      --set istio_cni.enabled=true | kubectl apply -f -
-    {{< /text >}}
-
-### Example: excluding specific Kubernetes namespaces
-
-The following example configures the Istio CNI plugin to ignore pods in the namespaces `istio-system`,
-`foo_ns`, and `bar_ns`.
-
-1.  Create an `istio-system` namespace for the Istio control-plane components.
-
-    {{< text bash >}}
-    $ kubectl create namespace istio-system
-    {{< /text >}}
-
-1.  Create and apply Istio CNI manifest.  Override the default configuration of the `istio-cni` Helm
-    chart's `logLevel` and `excludeNamespaces` parameters:
-
-    {{< text bash >}}
-    $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system \
-    --set logLevel=info \
-    --set excludeNamespaces={"istio-system,kube-system,foo_ns,bar_ns"} | kubectl apply -f -
-    {{< /text >}}
-
-1.  Create and apply Istio manifest with the Istio CNI plugin enabled:
-
-    {{< text bash >}}
-    $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
-    --set istio_cni.enabled=true | kubectl apply -f -
-    {{< /text >}}
 
 ### Helm chart parameters
 
@@ -125,31 +65,46 @@ The following example configures the Istio CNI plugin to ignore pods in the name
 | `cniConfDir` | | `/etc/cni/net.d` | Must be the same as the environment's `--cni-conf-dir` setting (`kubelet` parameter). |
 | `cniConfFileName` | | None | Leave unset to auto-find the first file in the `cni-conf-dir` (as `kubelet` does).  Primarily used for testing `install-cni` plugin configuration.  If set, `install-cni` will inject the plugin configuration into this file in the `cni-conf-dir`. |
 
-### Hosted Kubernetes Usage
+### Excluding specific Kubernetes namespaces
 
-Not all hosted Kubernetes clusters are created with the `kubelet` configured to use the CNI plugin so
-compatibility with this `istio-cni` solution is not ubiquitous.  The `istio-cni` plugin is expected
-to work with any hosted Kubernetes leveraging CNI plugins.  The below table indicates the known CNI status
-of many common Kubernetes environments.
+This example configures the Istio CNI plugin via `helm template` to modify the log level and
+ignore pods in the namespaces `istio-system`, `foo_ns`, and `bar_ns`.
+Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) for complete instructions.
 
-| Hosted Cluster Type | Uses CNI | Required Non-default settings |
-|---------------------|----------|-------------------------------|
-| GKE 1.9+ default | N | |
-| GKE 1.9+ w/ [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) | Y | `cniBinDir=/home/kubernetes/bin` |
-| IKS (IBM cloud) | Y | |
-| EKS (AWS) | Y | |
-| AKS (Azure) | Y | |
-| Red Hat OpenShift 3.10+ | Y | |
+Render and apply Istio CNI components and override the default configuration of the `istio-cni` Helm
+    chart's `logLevel` and `excludeNamespaces` parameters:
 
-#### GKE setup
+{{< text bash >}}
+$ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system \
+    --set logLevel=info \
+    --set excludeNamespaces={"istio-system,kube-system,foo_ns,bar_ns"} | kubectl apply -f -
+{{< /text >}}
+
+### Hosted Kubernetes settings
+
+The Istio CNI solution is not ubiquitous. Some platforms, especially hosted Kubernetes environments, do not enable the CNI plugin in the kubelet configuration.
+The `istio-cni` plugin is expected to work with any hosted Kubernetes leveraging CNI plugins.
+The below table indicates the required settings for many common Kubernetes environments.
+
+| Hosted Cluster Type | Required Istio CNI Setting Overrides | Required Platform Setting Overrides |
+|---------------------|--------------------------------------|-------------------------------------|
+| GKE 1.9+ (see [GKE setup](#gke-setup) below for details)| `cniBinDir=/home/kubernetes/bin` | enable [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) |
+| IKS (IBM cloud) | _(none)_ | _(none)_ |
+| EKS (AWS) | _(none)_ | _(none)_ |
+| AKS (Azure) | _(none)_ | _(none)_ |
+| Red Hat OpenShift 3.10+ | _(none)_ | _(none)_ |
+
+### GKE setup
 
 1.  Refer to the procedure to [prepare a GKE cluster for Istio](/docs/setup/kubernetes/platform-setup/gke/) and
     enable [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) in your cluster.
-    *  Note for existing clusters this redeploys the nodes.
 
-1.  Install Istio CNI via Helm including this option
-    `--set cniBinDir=/home/kubernetes/bin`. For example, the following `helm install`
-    command sets the `cniBinDir` value for GKE setups:
+    {{< warning >}}
+    For existing clusters, this redeploys all the nodes.
+    {{< /warning >}}
+
+1.  Install Istio CNI via Helm including the `--set cniBinDir=/home/kubernetes/bin` option.
+    For example, the following `helm install` command sets the `cniBinDir` value for a GKE cluster:
 
     {{< text bash >}}
     $ helm install install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system --set cniBinDir=/home/kubernetes/bin
@@ -215,7 +170,7 @@ corresponding application pod annotation key.
 
 ### Logging
 
-The Istio CNI plugin is run by the container runtime process space, and, therefore, the log entries are added within
+The Istio CNI plugin is run by the container runtime process space. Therefore, the log entries are logged by the
 the `kubelet` process.
 
 ## Compatibility with other CNI plugins
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -185,7 +185,7 @@ the `kubelet` process.
 
 ### Compatibility with application init containers
 
-Installing the Istio CNI plugin may cause networking problems for any application `initContainers`, since the plugin
+The Istio CNI plugin may cause networking problems for any application `initContainers`, since the plugin
 sets up traffic redirection for the pod prior to any init container starting, and Istio's sidecar proxy is not
 started up until all init containers have successfully completed, potentially resulting in traffic loss during the
 execution of the init containers.
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -186,5 +186,5 @@ CNI plugin only performs actions to setup the application pod's traffic redirect
 sidecar (using `iptables` in the pod's network namespace).
 
 {{< warning >}}
-This _should_ have no effect on the operations performed by the base CNI plugin configuring the pod's networking setup, although not all CNI's have been validated.
+The Istio CNI plugin should have no effect on the operations performed by the base CNI plugin configuring the pod's networking setup, although not all CNI plugins have been validated.
 {{< /warning >}}
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -174,7 +174,7 @@ application pod annotation key.
 | `traffic.sidecar.istio.io/excludeOutboundIPRanges` | `<IPCidr1>,<IPCidr2>,...` | | Comma separated list of IP ranges in CIDR form to be excluded from redirection.  Only applies when `includeOutboundIPRanges` is `"*"`. |
 | `traffic.sidecar.istio.io/includeInboundPorts` | `<port1>,<port2>,...` | Pod's list of `containerPorts` | Comma separated list of inbound ports for which traffic is to be redirected to the Istio proxy sidecar.  The value of `"*"` redirects all ports. |
 | `traffic.sidecar.istio.io/excludeInboundPorts` | `<port1>,<port2>,...` | | Comma separated list of inbound ports to be excluded from redirection to the Istio sidecar proxy.  Only valid when `includeInboundPorts` is `"*"`. |
-| `traffic.sidecar.istio.io/excludeOutboundPorts` | `<port1>,<port2>,...` | | Comma separated list of outbound ports to be excluded from redirection to Envoy. | 
+| `traffic.sidecar.istio.io/excludeOutboundPorts` | `<port1>,<port2>,...` | | Comma separated list of outbound ports to be excluded from redirection to Envoy. |
 | `traffic.sidecar.istio.io/kubevirtInterfaces` | `<ethX>,<ethY>,...` | | Comma separated list of virtual interfaces whose inbound traffic (from VM) will be treated as outbound. |
 
 ### Logging
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -187,14 +187,14 @@ Due to this, the `kubelet` process writes the plugin's log entries into its log.
 The Istio CNI plugin may cause networking connectivity problems for any application `initContainers`. When using Istio CNI, `kubelet`
 starts an injected pod with the following steps:
 1. The Istio CNI plugin sets up traffic redirection to the Istio sidecar proxy within the pod.
-1. All init containers execute and successfully complete.
-1. The Istio sidecar proxy is started in the pod along with the pod's other containers.
+1. All init containers execute and complete successfully.
+1. The Istio sidecar proxy starts in the pod along with the pod's other containers.
 
-Init containers execute before the sidecar proxy is started, which may result in traffic loss during their execution.
-This traffic loss can be avoided by:
-- setting the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to disable redirecting traffic to any
-  CIDRs the init containers communicate with,
-- and/or setting the `traffic.sidecar.istio.io/excludeOutboundPorts` annotation to disable redirecting traffic to the
+Init containers execute before the sidecar proxy starts, which can result in traffic loss during their execution.
+Avoid this traffic loss with one or both of the following settings:
+- Set the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to disable redirecting traffic to any
+  CIDRs the init containers communicate with.
+- Set the `traffic.sidecar.istio.io/excludeOutboundPorts` annotation to disable redirecting traffic to the
   specific outbound ports the init containers use.
 
 ### Compatibility with other CNI plugins
@@ -210,5 +210,6 @@ CNI plugin only performs actions to setup the application pod's traffic redirect
 sidecar (using `iptables` in the pod's network namespace).
 
 {{< warning >}}
-The Istio CNI plugin should have no effect on the operations performed by the base CNI plugin configuring the pod's networking setup, although not all CNI plugins have been validated.
+The Istio CNI plugin should not interfere with the operations of the base CNI plugin that configures the pod's
+networking setup, although not all CNI plugins have been validated.
 {{< /warning >}}
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -46,7 +46,7 @@ replaces the functionality provided by the `istio-init` container.
     Refer to [Hosted Kubernetes settings](#hosted-kubernetes-settings) for any non-default settings required.
 
 1.  Install Istio CNI and Istio using Helm.
-    Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) instructions and the **default with CNI enabled** profile.
+    Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) instructions and the **Istio CNI enabled** profile.
     Pass `--set cniBinDir=...` and/or `--set cniConfDir=...` options when installing `istio-cni` if non-default, as determined in the previous step.
 
     {{< tip >}}
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -49,11 +49,9 @@ replaces the functionality provided by the `istio-init` container.
     Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) instructions and the **Istio CNI enabled** profile.
     Pass `--set cniBinDir=...` and/or `--set cniConfDir=...` options when installing `istio-cni` if non-default, as determined in the previous step.
 
-    {{< tip >}}
-    Refer to [Helm chart parameters](#helm-chart-parameters) below for the full set of `istio-cni` Helm parameters.
-    {{< /tip >}}
-
 ### Helm chart parameters
+
+The following table indicates all the options supported by the `istio-cni` Helm chart:
 
 | Option | Values | Default | Description |
 |--------|--------|---------|-------------|
@@ -72,7 +70,7 @@ ignore pods in the namespaces `istio-system`, `foo_ns`, and `bar_ns`.
 Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) for complete instructions.
 
 Render and apply Istio CNI components and override the default configuration of the `istio-cni` Helm
-    chart's `logLevel` and `excludeNamespaces` parameters:
+chart's `logLevel` and `excludeNamespaces` parameters:
 
 {{< text bash >}}
 $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system \
@@ -84,7 +82,7 @@ $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=i
 
 The Istio CNI solution is not ubiquitous. Some platforms, especially hosted Kubernetes environments, do not enable the CNI plugin in the kubelet configuration.
 The `istio-cni` plugin is expected to work with any hosted Kubernetes leveraging CNI plugins.
-The below table indicates the required settings for many common Kubernetes environments.
+The following table indicates the required settings for many common Kubernetes environments.
 
 | Hosted Cluster Type | Required Istio CNI Setting Overrides | Required Platform Setting Overrides |
 |---------------------|--------------------------------------|-------------------------------------|
@@ -153,7 +151,7 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
 1.  The pod has more than 1 container.
 1.  The pod has no annotation with key `sidecar.istio.io/inject` OR the value of the annotation is `true`.
 
-### Traffic redirection details
+### Traffic redirection parameters
 
 To redirect traffic in the application pod's network namespace to/from the Istio proxy sidecar, the Istio
 CNI plugin configures the namespace's iptables.  The following table describes the parameters to the
@@ -170,6 +168,16 @@ corresponding application pod annotation key.
 | `traffic.sidecar.istio.io/excludeOutboundPorts` | `<port1>,<port2>,...` | | Comma separated list of outbound ports to be excluded from redirection to Envoy. | 
 | `traffic.sidecar.istio.io/kubevirtInterfaces` | `<ethX>,<ethY>,...` | | Comma separated list of virtual interfaces whose inbound traffic (from VM) will be treated as outbound. |
 
+Cluster-wide defaults for these parameters can be set using the corresponding
+[`global.proxy.*` parameters](/docs/reference/config/installation-options/#global-options) when installing the `istio`
+Helm chart.  For instance, to set a default `traffic.sidecar.istio.io/excludeOutboundIPRanges` option for all pods in
+the cluster when installing Istio:
+
+{{< text bash >}}
+$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
+    --set global.proxy.excludeOutboundPorts=53 | kubectl apply -f -
+{{< /text >}}
+
 ### Logging
 
 The Istio CNI plugin is run by the container runtime process space. Therefore, the log entries are logged by the
@@ -178,13 +186,13 @@ the `kubelet` process.
 ### Compatibility with application init containers
 
 Installing the Istio CNI plugin may cause networking problems for any application `initContainers`, since the plugin
-sets up traffic interception for the pod prior to any init container starting, and Istio's sidecar proxy is not
+sets up traffic redirection for the pod prior to any init container starting, and Istio's sidecar proxy is not
 started up until all init containers have successfully completed, potentially resulting in traffic loss during the
 execution of the init containers.
 
-This traffic loss can be avoided by setting the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to avoid
-intercepting traffic to any CIDRs the init containers communicate with, and/or the
-`traffic.sidecar.istio.io/excludeOutboundPorts` annotation to avoid intercepting traffic to the specific outbound ports
+This traffic loss can be avoided by setting the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to disable
+redirecting traffic to any CIDRs the init containers communicate with, and/or the
+`traffic.sidecar.istio.io/excludeOutboundPorts` annotation to disable redirecting traffic to the specific outbound ports
 the init containers use.
 
 ### Compatibility with other CNI plugins
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -177,16 +177,6 @@ application pod annotation key.
 | `traffic.sidecar.istio.io/excludeOutboundPorts` | `<port1>,<port2>,...` | | Comma separated list of outbound ports to be excluded from redirection to Envoy. | 
 | `traffic.sidecar.istio.io/kubevirtInterfaces` | `<ethX>,<ethY>,...` | | Comma separated list of virtual interfaces whose inbound traffic (from VM) will be treated as outbound. |
 
-To set cluster-wide defaults for these parameters, use the corresponding
-[`global.proxy.*` parameters](/docs/reference/config/installation-options/#global-options) when installing Istio via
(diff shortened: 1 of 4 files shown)


content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -27,63 +27,135 @@ setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/set
...
+   $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system | kubectl apply -f -
+   {{< /text >}}
+
+1. Install all the Istio
+    [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)
+    (CRDs) using `kubectl apply`, and wait a few seconds for the CRDs to be committed in the Kubernetes API-server:
 
     {{< text bash >}}
-    $ tar -xzf install/kubernetes/helm/charts/istio-cni-*.tgz -C install/kubernetes/helm/
+    $ helm template install/kubernetes/helm/istio-init --name istio-init --namespace istio-system | kubectl apply -f -
```
john-a-joyce: All istio install related text should be removed.  Istio and Istio-cni are now two very distinct steps and we don't want to refer to Isito install here in case it becomes stale and doesn't match how the main Istio install steps evolve.   The only thing that should be preserved and highlighted AS MUCH AS POSSIBLE is that the istio install need to be done with  `--set istio_cni.enabled=true`

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -27,63 +27,135 @@ setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/set
...
+    $ helm init --service-account tiller
+    {{< /text >}}
+
+1. Install the `istio-cni` chart:
+
+    {{< text bash >}}
+    $ helm install install/kubernetes/helm/istio-cni --name istio-cni --namespace istio-system
+    {{< /text >}}
+
+1. Install the `istio-init` chart to bootstrap all the Istio's CRDs:
```
john-a-joyce: Ditto above remove all Istio installation steps

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -27,63 +27,135 @@ setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/set
...
+    with the `--set istio_cni.enabled=true` setting.
+    The **default** profile is recommended for production deployments.
+    Refer to the the [Istio Installation with Helm procedure](/docs/setup/kubernetes/install/helm/#option-2-install-with-helm-and-tiller-via-helm-install)
+    to use other profiles.
+
+    {{< text bash >}}
+    $ helm install install/kubernetes/helm/istio --name istio --namespace istio-system --set istio_cni.enabled=true
     {{< /text >}}
 
 ### Example: excluding specific Kubernetes namespaces
```
john-a-joyce: I know this is WIP but Steve wanted "Exampled" removed

john-a-joyce:
@rlenglet - thanks very much for getting going on this.  I was actually planning to start on this today so its great that you started this.   You can see my inline comments but the main theme of all of them is that we should decouple this from the Istio install as much as possible.  There was a decision 6 months back to really decouple the CNI install form the Istio install as much as possible,  but the docs never caught up with that.   While refactoring we should make that split more obvious and it will greatly simplify keeping this up to date as Istio install process evolves.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -27,63 +27,135 @@ setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/set
...
-    *  Kubernetes installations for AWS EKS, Azure AKS, and IBM Cloud IKS clusters have this capability.
-    *  Google Cloud GKE clusters require that the
-       [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) feature
-       is enabled to have Kubernetes configured with `network-plugin=cni`.
+## Prerequisites
 
-1.  Install Kubernetes with the [ServiceAccount admission controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#serviceaccount) enabled.
-    *  The Kubernetes documentation highly recommends this for all Kubernetes installations
-       where `ServiceAccounts` are utilized.
+1. [Download the Istio release](/docs/setup/kubernetes/#downloading-the-release).
```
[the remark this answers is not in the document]
rlenglet: The CNI chart is now shipped in the Istio release.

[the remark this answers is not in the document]
rlenglet: Cf. https://github.com/istio/istio/pull/14752

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -27,63 +27,135 @@ setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/set
...
         settings required.
 
-1.  Create a namespace for the Istio control-plane components:
+1. Check the [Requirements for Pods and Services](/docs/setup/kubernetes/additional-setup/requirements/).
+
+1. [Install a Helm client](https://github.com/helm/helm/blob/master/docs/install.md) with a version higher than 2.10.
+
+## Helm chart release repositories
+
+The commands in this guide use the Helm charts that are included in the Istio release image.
```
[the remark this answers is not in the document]
rlenglet: As I commented above, the CNI chart now ships in the Istio archive, cf. https://github.com/istio/istio/pull/14752.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -27,63 +27,135 @@ setup phase, thereby removing the [`NET_ADMIN` capability requirement](/docs/set
...
+{{< text bash >}}
+$ helm repo add istio.io https://storage.googleapis.com/istio-release/releases/{{< istio_full_version >}}/charts/
+{{< /text >}}
+
+## Installation steps
+
+Change directory to the root of the release and then
+choose one of the following two **mutually exclusive** options:
+
+1. To deploy Istio without using Tiller, follow the instructions for [option 1](/docs/setup/kubernetes/additional-setup/cni/#option-1-install-with-helm-via-helm-template).
```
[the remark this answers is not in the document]
rlenglet: Ditto.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: There is still a lot in common that was missing, e.g. the proper installation of Tiller. We need to give at least a copy-and-pastable example of a command to install `istio` with `--set istio_cni.enabled=true`, and stop there. And that's exactly what I've done here. I've removed all other profiles, for example. I don't see what else I can remove.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Will do.

tiswanso:
@rlenglet @john-a-joyce @sdake 
It looks like we're in danger of having a lot of the same discussion that was had when this doc was updated for release-1.1 (#3537).   

The gist of that was that this doc is not a recipe to install Istio and it should not be a repeat of the content in `/docs/setup/kubernetes/install/helm/`, `/docs/setup/kubernetes/download/`, `/docs/setup/kubernetes/`, etc.  IMO, the goal is to contextualize the istio-cni installation with references to the main Istio installation documents.  I strongly disagree with trying to make an exact full-on cut/paste example of installation of Istio in this doc since it will diverge from the actual getting-started instructions and likely cause confusion with istio-cni users that its dependent on specific ways to install Istio other than the single cni-enabled helm flag.

IMO, the doc as it exists without this PR is not far off and likely needs some clarifying text about how to use it in conjunction with an Istio install.

content/docs/setup/kubernetes/additional-setup/cni/index.md
john-a-joyce: I don't agree the more istio install content we put here the quicker it becomes out of date.  This was originally written at a time when the Istio and Istio-cni installs where tightly coupled.   Now the only two intersections points are that the istio-cni helm charts are available as part of the istio CNI and the istio install needs to set the istio_cni.enabled flag.  We should refactor this to only touch on those two intersection points.  If you want to provide links to create the coupling that is ok because at least if the link becomes stale the intent would still be clear.

content/docs/setup/kubernetes/additional-setup/cni/index.md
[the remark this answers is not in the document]
john-a-joyce: I know the chart now ships with the Istio archive,  but that doesn't change the point of my comment.   It is not the only way to get the chart.

content/docs/setup/kubernetes/additional-setup/cni/index.md
tiswanso: @john-a-joyce -- for release-1.1 the doc was updated with the separation of the istio-cni v. istio.  A lot of the stuff @rlenglet is adding is putting in repeated content from the Istio getting started guide and I disagree with that.  The existing istio-cni doc installation section-- https://istio.io/docs/setup/kubernetes/additional-setup/cni/#installation--it just refers to the details for installing the istio-cni and then an example command for passing the helm setting for `istio_cni.enabled=true`.   

@sdake's confusion triggering the issue creation was 
1) the istio helm chart / test-and-release changes broke the procedure by removing the istio-cni tarfile from the istio release charts 
2) he thought he could copy paste everything in just this doc and get Istio installed with the CNI.  (maybe auto-pilot from other doc usage)  

For 1) that's been addressed by the PR for the release/ stuff.  For 2) we can improve the commentary about the procedure but the  procedure still is the same as release-1.1.

rcaballeromx:
@john-a-joyce @sdake @rlenglet Please let me know when you feel like I can/should take a look and review.

rlenglet:
I'll try a different approach with moving the CNI-specific installation bits into the Helm docs, in a different tab in the tabsets, and then just refer to that page from the CNI page.

rlenglet:
Please review. I believe I have addressed the comments.

rlenglet:
The CircleCI failure is unrelated to this PR:
```
- ./public/docs/tasks/policy-enforcement/rate-limiting/index.html
  *  External link http://redis.io/ failed: 301 SSL peer certificate or SSH remote key was not OK
- ./public/zh/docs/tasks/security/authz-permissive/index.html
  *  External link https://raw.githubusercontent.com/istio/istio/master/samples/bookinfo/platform/kube/bookinfo-add-serviceaccount.yaml failed: 404 No error
- ./public/zh/docs/tasks/security/authz-tcp/index.html
  *  External link https://raw.githubusercontent.com/istio/istio/master/samples/bookinfo/platform/kube/rbac/ratings-v2-add-serviceaccount.yaml failed: 404 No error
htmlproofer 3.9.2 | Error:  HTML-Proofer found 3 failures!
```

yangminzhu:
Can you cherrypick https://github.com/istio/istio.io/pull/4429 to this PR? It fixes the broken link in the zh RBAC tasks.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -40,79 +40,19 @@ replaces the functionality provided by the `istio-init` container.
...
-        $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system | kubectl apply -f -
-        {{< /text >}}
-
-    *  Option 2.  Install via Helm and Tiller via `helm install`.
-
-        {{< text bash >}}
-        $ helm install install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system
-        {{< /text >}}
+1.  Install Istio CNI and Istio using Helm.
+    Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) instructions and the **default with CNI enabled** profile.
```
john-a-joyce: This link didn't seem to work for me maybe this is expected with things in transit.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -40,79 +40,19 @@ replaces the functionality provided by the `istio-init` container.
...
-        {{< text bash >}}
-        $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system | kubectl apply -f -
-        {{< /text >}}
-
-    *  Option 2.  Install via Helm and Tiller via `helm install`.
-
-        {{< text bash >}}
-        $ helm install install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system
-        {{< /text >}}
+1.  Install Istio CNI and Istio using Helm.
```
john-a-joyce: Per my previous comments I am not sure why feel compelled to include the Istio install here.  The rest of this page is not dependent on any Istio installation steps so it isn't required to install the CNI.  If you few strongly that the Istio install should be left here at least change the wording to say something "it is also recommended to install Istio at this time".  Practically it would be better to not install Istio at the same time as any containers that require a sidecar are dependent on the CNI being up before they can commence.

john-a-joyce:
This looks much better if you still think it is important to include a reference to the Istio install in this guide at least indicate it is not a required step and the CNI can be install has without any dependency on the Istio install.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: @john-a-joyce  What do you mean it doesn't work? The `#cni` hash at the end highlights the tabs that install the Istio CNI. Doesn't that work for you?

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: It _is_ intertwined with the Istio installation steps. Especially, the following steps must be completed beforehand: download the release, install Helm (incl. creating the account for Tiller), create the `istio-system` namespace.

rlenglet:
> This looks much better if you still think it is important to include a reference to the Istio install in this guide at least indicate it is not a required step and the CNI can be install has without any dependency on the Istio install.

I think that would be neat, but I don't think that's realistic. For instance, when installing the CNI with Tiller, Tiller must first be installed with the service account defined in Istio, in order not to mess up with any subsequent Istio install. If we don't give compatible instructions, users will have surprises when they install Istio afterwards.

content/docs/setup/kubernetes/additional-setup/cni/index.md
john-a-joyce: A couple points.   You don't have to download the release its just the most convenient way to get the helm chart.   You also don't need to install the CNI in the istio-system namespace it works just fine in its own namespace.  We also document an option that doesn't require Tiller.   So it is absolutely not required to install Istio to get the CNI installed.   If you want to keep it the Istio install in the documentation at least note it is recommended and not required.  Alternatively and slightly better would be to note why it is recommended to install Istio as you have noted in your response above (i.e. Chart, ns & accounts)    Also the implied order here is wrong as you only want to follow enough of the Istio install to get the helm chart, create the necessary accounts and create the namespace.   Then you want to complete the CNI install before completing the Istio install otherwise some of the Istio pods will not install until the CNI is up.

tiswanso: @john-a-joyce  -- IMO, for the purposes of this doc the Istio and istio-cni installs are intertwined.  I went through the same discussion in release 1.1 about just describing purely the istio-cni install and the opinions were that it should be contextualized along with the getting started guide's procedure for Istio install.  There also was a strong push for documentation simplicity over detailed scenarios so a lot of my original explanations about the independence of the CNI install were removed during review.  

@rlenglet -- We should convey that the istio-cni install works with any of the Istio install profiles and explain the step in "default with CNI enabled" tab for installing the istio-cni is identical for all profiles.  Possibly we want to pull out the istio-cni apply into its own step in the procedure like it was prior but add it to the helm install doc like you've started with the new tab.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -215,7 +170,7 @@ corresponding application pod annotation key.
 
 ### Logging
 
-The Istio CNI plugin is run by the container runtime process space, and, therefore, the log entries are added within
+The Istio CNI plugin is run by the container runtime process space. Therefore, the log entries are logged by the
 the `kubelet` process.
 
 ## Compatibility with other CNI plugins
```
tiswanso: In this section, @sdake had a comment that the `warning` wording was ambiguous.  Could you change "this should have no effect" to "The Istio CNI plugin should have no effect" in this PR.

tiswanso:
An additional thing we didn't add to the documentation was that the use of the istio-cni could cause networking problems for any application initContainers since the istio-cni will add traffic redirect rules prior to the initContainers starting and the envoy proxy is not brought up until the pod's normal containers phase (after the initContainers complete).   The workarounds are not great--e.g. use `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation for any CIDRs the initContainer talks to; use `traffic.sidecar.istio.io/excludeOutboundPorts` annotation to avoid redirect for any outbound ports the initContainer would use.

content/docs/setup/kubernetes/additional-setup/cni/index.md
john-a-joyce: @tiswanso - I recall the history - but at the same time the Istio and istio-cni where more interwound at that time now they are almost entirely separated.   Although i feel strongly about this I seem to be in the minority on this point.   If the Istio install is left in these instructions then there MUST be some type of note to indicate that the CNI install should be completed before the bulk of the Istio install should be attempted.

john-a-joyce: I still disagree that the istio install should be mention here at all.  As currently written it goes in circles.  
Istio Install docs - "go install CNI"
CNI install instructions - "go install Istio"

That is plain stupid in my opinion,  I don't really care what history go us to this point.  It is still stupid and confusing.  I am tired of arguing about it so this will be my last comment on the topic.  If people insist that Istio install should be included here in the CNI install steps I implore you to at least make sure the reader understands that the CNI should be installed first.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
tiswanso: @rlenglet -- I spoke with @john-a-joyce.  His concerns would be satisfied with a warning or blurb in the CNI installation doc to the effect of:
"The Istio installation may fail if the Istio CNI plugin is not installed prior."

tiswanso:
IMO, since we're adding istio-cni install info inline in the helm install instructions, we can remove the note in the helm install prereqs:  https://deploy-preview-4398--preliminary-istio.netlify.com/docs/setup/kubernetes/install/helm/#prerequisites

rlenglet:
> Need to add row for traffic.sidecar.istio.io/excludeOutboundPorts to the table in the "Traffic redirection details" section.

Done. Also added a row for `traffic.sidecar.istio.io/kubevirtInterfaces`.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Sorry, catching up on this thread.

> You don't have to download the release its just the most convenient way to get the helm chart.

More than "the most convenient way", my view point is that this doc should specify **the recommended way** to install the Istio CNI plugin, which guarantees that it it will work regardless of how users decide to install Istio. If users decide to install Istio using Tiller, which is one of the recommended ways to install Istio, the doc should make it easy to work for them.

> We also document an option that doesn't require Tiller.

True, but we should document precisely how the Istio CNI needs to be installed in the Tiller case, since Tiller is a recommended way to install Istio, and installing Istio CNI in this scenario is far from trivial.

> Then you want to complete the CNI install before completing the Istio install otherwise some of the Istio pods will not install until the CNI is up.

How could that happen if the user followed the new install doc step-by-step?

> explain the step in "default with CNI enabled" tab for installing the istio-cni is identical for all profiles.

OK, I'll do that.

> Possibly we want to pull out the istio-cni apply into its own step in the procedure like it was prior but add it to the helm install doc like you've started with the new tab.

I decided against doing that because the installation of the `istio-cni` chart is strongly coupled with passing `--set istio_cni.enabled=true`. Putting those commands together highlights the strong dependency.

rlenglet:
> An additional thing we didn't add to the documentation was that the use of the istio-cni could cause networking problems for any application initContainers since the istio-cni will add traffic redirect rules prior to the initContainers starting and the envoy proxy is not brought up until the pod's normal containers phase (after the initContainers complete). The workarounds are not great--e.g. use traffic.sidecar.istio.io/excludeOutboundIPRanges annotation for any CIDRs the initContainer talks to; use traffic.sidecar.istio.io/excludeOutboundPorts annotation to avoid redirect for any outbound ports the initContainer would use.

I added a `Compatibility with application init containers` subsection, before the `Compatibility with other CNI plugins` section.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -157,7 +157,7 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
 
 To redirect traffic in the application pod's network namespace to/from the Istio proxy sidecar, the Istio
 CNI plugin configures the namespace's iptables.  The following table describes the parameters to the
-redirect functionality.  The default values for the parameters are able to be overridden by setting the
+redirect functionality.  The default values for the parameters can be overridden by setting the
```
john-a-joyce: It might be worth mentioning that these parameters can also be set via helm values.

rlenglet:
> IMO, since we're adding istio-cni install info inline in the helm install instructions, we can remove the note in the helm install prereqs

Good catch, I missed that one. Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -202,23 +157,37 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
 ### Logging
 
-The Istio CNI plugin is run by the container runtime process space, and, therefore, the log entries are added within
+The Istio CNI plugin is run by the container runtime process space. Therefore, the log entries are logged by the
 the `kubelet` process.
 
-## Compatibility with other CNI plugins
+### Compatibility with application init containers
+
+Installing the Istio CNI plugin may cause networking problems for any application `initContainers`, since the plugin
```
tiswanso: nit... remove the word "Installing"

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

rlenglet:
I believe I have addressed all comments. Thanks for the thorough reviews!

tiswanso:
@rcaballeromx @sdake -- I think the PR is ready for your review.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rcaballeromx: @rlenglet I understand what you are trying to do is supply information that works for as many readers as possible. I understand that you want CNI to work for readers that are using Tiller and those that aren't, for example. Istio offers configuration profiles for installation. My advice is that you pick a configuration profile during install and write the document based on that configuration profile. Then you can just introduce this doc by saying in the Before you begin section: "Install Istio using [the configuration profile you select]."

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -40,81 +40,19 @@ replaces the functionality provided by the `istio-init` container.
     *  The Kubernetes documentation highly recommends this for all Kubernetes installations
        where `ServiceAccounts` are utilized.
 
-1.  [Download the Istio release](/docs/setup/kubernetes/#downloading-the-release).
```
rcaballeromx: The Istio installation needs to be part of the prerequisites. Readers need to be able to enable CNI independently from installing Istio.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -40,81 +40,19 @@ replaces the functionality provided by the `istio-init` container.
...
-    $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
-    --set istio_cni.enabled=true | kubectl apply -f -
-    {{< /text >}}
+1.  Install Istio CNI and Istio using Helm.
+    Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) instructions and the **Istio CNI enabled** profile.
+    Pass `--set cniBinDir=...` and/or `--set cniConfDir=...` options when installing `istio-cni` if non-default, as determined in the previous step.
 
 ### Helm chart parameters
 
+The following table indicates all the options supported by the `istio-cni` Helm chart:
```
rcaballeromx:
```suggestion
The following table shows all the options that the the `istio-cni` Helm chart supports:
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -125,31 +63,46 @@ The following example configures the Istio CNI plugin to ignore pods in the name
 | `cniConfDir` | | `/etc/cni/net.d` | Must be the same as the environment's `--cni-conf-dir` setting (`kubelet` parameter). |
 | `cniConfFileName` | | None | Leave unset to auto-find the first file in the `cni-conf-dir` (as `kubelet` does).  Primarily used for testing `install-cni` plugin configuration.  If set, `install-cni` will inject the plugin configuration into this file in the `cni-conf-dir`. |
 
-### Hosted Kubernetes Usage
+### Excluding specific Kubernetes namespaces
+
+This example configures the Istio CNI plugin via `helm template` to modify the log level and
```
rcaballeromx: Readers can be confused if they don't know that `helm template` is just a Helm command. User lists instead of paragraphs to make the content more scannable. This applies to all lists in paragraphs but I am only leaving a suggestion here:

```suggestion
This example uses Helm to perform the following tasks:

- Configure the Istio CNI plugin 
- Modify the log level
- Ignore the pods in the following namespaces:
    - `istio-system`
    - `foo_ns`
    - `bar_ns`
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -125,31 +63,46 @@ The following example configures the Istio CNI plugin to ignore pods in the name
...
+ignore pods in the namespaces `istio-system`, `foo_ns`, and `bar_ns`.
```
rcaballeromx:
```suggestion
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -125,31 +63,46 @@ The following example configures the Istio CNI plugin to ignore pods in the name
...
+Refer to the [Customizable Install with Helm](/docs/setup/kubernetes/install/helm/#cni) for complete instructions.
+
+Render and apply Istio CNI components and override the default configuration of the `istio-cni` Helm
```
rcaballeromx:
```suggestion
Use the following command to render and apply Istio CNI components and override the default configuration of the `istio-cni` Helm
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -125,31 +63,46 @@ The following example configures the Istio CNI plugin to ignore pods in the name
...
+    --set excludeNamespaces={"istio-system,kube-system,foo_ns,bar_ns"} | kubectl apply -f -
+{{< /text >}}
+
+### Hosted Kubernetes settings
 
-Not all hosted Kubernetes clusters are created with the `kubelet` configured to use the CNI plugin so
-compatibility with this `istio-cni` solution is not ubiquitous.  The `istio-cni` plugin is expected
-to work with any hosted Kubernetes leveraging CNI plugins.  The below table indicates the known CNI status
-of many common Kubernetes environments.
+The Istio CNI solution is not ubiquitous. Some platforms, especially hosted Kubernetes environments, do not enable the CNI plugin in the kubelet configuration.
```
rcaballeromx: Please respect the 80 character length.
```suggestion
The Istio CNI solution is not ubiquitous. Some platforms, especially hosted Kubernetes environments,
do not enable the CNI plugin in the kubelet configuration.
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -125,31 +63,46 @@ The following example configures the Istio CNI plugin to ignore pods in the name
...
 
 1.  Refer to the procedure to [prepare a GKE cluster for Istio](/docs/setup/kubernetes/platform-setup/gke/) and
     enable [network-policy](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy) in your cluster.
-    *  Note for existing clusters this redeploys the nodes.
 
-1.  Install Istio CNI via Helm including this option
-    `--set cniBinDir=/home/kubernetes/bin`. For example, the following `helm install`
-    command sets the `cniBinDir` value for GKE setups:
+    {{< warning >}}
+    For existing clusters, this redeploys all the nodes.
```
rcaballeromx:
```suggestion
    For existing clusters, this redeploys all nodes.
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
john-a-joyce: FWIW you can't install Istio first it will fail or at least spin its wheels until the CNI comes up.  By install I mean the actual kube apply of the Istio YAML some of the Istio install steps must be done first (e.g to download the release.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +151,51 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
 1.  The pod has more than 1 container.
 1.  The pod has no annotation with key `sidecar.istio.io/inject` OR the value of the annotation is `true`.
 
-### Traffic redirection details
+### Traffic redirection parameters
 
 To redirect traffic in the application pod's network namespace to/from the Istio proxy sidecar, the Istio
 CNI plugin configures the namespace's iptables.  The following table describes the parameters to the
-redirect functionality.  The default values for the parameters are able to be overridden by setting the
+redirect functionality.  The default values for the parameters can be overridden by setting the
```
rcaballeromx:
```suggestion
redirect functionality.  To override the default values for the parameters, set the
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +151,51 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
-| `traffic.sidecar.istio.io/includeOutboundIPRanges` | `<IPCidr1>,<IPCidr2>,...` | "*" | Optional comma separated list of IP ranges in CIDR form to redirect to the sidecar proxy.  The default value of "*" redirects all traffic. |
-| `traffic.sidecar.istio.io/excludeOutboundIPRanges` | `<IPCidr1>,<IPCidr2>,...` | | Optional comma separated list of IP ranges in CIDR form to be excluded from redirection.  Only applies when `includeOutboundIPRanges` is "*". |
-| `traffic.sidecar.istio.io/includeInboundPorts` | `<port1>,<port2>,...` | Pod's list of `containerPorts` | Comma separated list of inbound ports for which traffic is to be redirected to the Istio proxy sidecar.  The value of "*" redirects all ports. |
-| `traffic.sidecar.istio.io/excludeInboundPorts` | `<port1>,<port2>,...` | | Comma separated list of inbound ports to be excluded from redirection to the Istio sidecar proxy.  Only valid when `includeInboundPorts` is "*". |
+| `traffic.sidecar.istio.io/includeOutboundIPRanges` | `<IPCidr1>,<IPCidr2>,...` | `"*"` | Optional comma separated list of IP ranges in CIDR form to redirect to the sidecar proxy.  The default value of `"*"` redirects all traffic. |
```
rcaballeromx: Please use bold to emphasize the fields that are optional. For example:
```suggestion
| `traffic.sidecar.istio.io/includeOutboundIPRanges` | `<IPCidr1>,<IPCidr2>,...` | `"*"` | **Optional** Comma separated list of IP ranges in CIDR form to redirect to the sidecar proxy.  The default value of `"*"` redirects all traffic. |
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +151,51 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+| `traffic.sidecar.istio.io/excludeOutboundIPRanges` | `<IPCidr1>,<IPCidr2>,...` | | Optional comma separated list of IP ranges in CIDR form to be excluded from redirection.  Only applies when `includeOutboundIPRanges` is `"*"`. |
+| `traffic.sidecar.istio.io/includeInboundPorts` | `<port1>,<port2>,...` | Pod's list of `containerPorts` | Comma separated list of inbound ports for which traffic is to be redirected to the Istio proxy sidecar.  The value of `"*"` redirects all ports. |
+| `traffic.sidecar.istio.io/excludeInboundPorts` | `<port1>,<port2>,...` | | Comma separated list of inbound ports to be excluded from redirection to the Istio sidecar proxy.  Only valid when `includeInboundPorts` is `"*"`. |
+| `traffic.sidecar.istio.io/excludeOutboundPorts` | `<port1>,<port2>,...` | | Comma separated list of outbound ports to be excluded from redirection to Envoy. | 
+| `traffic.sidecar.istio.io/kubevirtInterfaces` | `<ethX>,<ethY>,...` | | Comma separated list of virtual interfaces whose inbound traffic (from VM) will be treated as outbound. |
+
+Cluster-wide defaults for these parameters can be set using the corresponding
```
rcaballeromx:
```suggestion
To set cluster-wide defaults for these parameters, use the corresponding
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +151,51 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+[`global.proxy.*` parameters](/docs/reference/config/installation-options/#global-options) when installing the `istio`
```
rcaballeromx:
```suggestion
[`global.proxy.*` parameters](/docs/reference/config/installation-options/#global-options) when installing Istio via
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +151,51 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+Helm chart.  For instance, to set a default `traffic.sidecar.istio.io/excludeOutboundIPRanges` option for all pods in
```
rcaballeromx:
```suggestion
the `istio` Helm chart.  For example, you can set a default `traffic.sidecar.istio.io/excludeOutboundIPRanges` option for all pods in
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +151,51 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+the cluster when installing Istio:
```
rcaballeromx:
```suggestion
the cluster using the following command to install Istio:
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +151,51 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+
+{{< text bash >}}
+$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
+    --set global.proxy.excludeOutboundPorts=53 | kubectl apply -f -
+{{< /text >}}
 
 ### Logging
 
-The Istio CNI plugin is run by the container runtime process space, and, therefore, the log entries are added within
+The Istio CNI plugin is run by the container runtime process space. Therefore, the log entries are logged by the
```
rcaballeromx: No passive voice please. Readers, particularly non-native English readers have a very difficult time untangling passive voice. Opt instead for the simple present:
```suggestion
The container runtime process space runs the Istio CNI plugin. Due to this, the `kubelet` process creates the entries in the log.
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +151,51 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
 the `kubelet` process.
 
-## Compatibility with other CNI plugins
+### Compatibility with application init containers
+
+The Istio CNI plugin may cause networking problems for any application `initContainers`, since the plugin
```
rcaballeromx: This entire paragraph is very difficult to follow. What can cause problems? Are the problems caused by my container, by Istio, by CNI? Where do I need to go to fix it? Should I change the sidecar? Those are just some of the questions a reader might have by the end of the paragraph. 
Please rewrite. IMO explaining the process as it works normally first and then pointing out which step can be problematic because the redirect gets put in place before the Istio initialization container runs, would go a long way. Ideally, explain it in a numbered list.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: No, it can't be done before installing Istio CNI.
The installation of Istio and the Istio CNI is done in the 2nd step of the "Installation" section.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: That would result in a lot of churn in this file. It seems like it's been edited with ~100 columns. I'll stick to that and fix the lines longer than 100 columns.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: They are all optional, with default values. I'll just remove the occurrences of "Optional" in this table.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: > What can cause problems?

I think this is pretty clear: "The Istio CNI plugin may cause networking problems"

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: Split into a numbered list.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,54 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+| `traffic.sidecar.istio.io/excludeInboundPorts` | `<port1>,<port2>,...` | | Comma separated list of inbound ports to be excluded from redirection to the Istio sidecar proxy.  Only valid when `includeInboundPorts` is `"*"`. |
+| `traffic.sidecar.istio.io/excludeOutboundPorts` | `<port1>,<port2>,...` | | Comma separated list of outbound ports to be excluded from redirection to Envoy. | 
+| `traffic.sidecar.istio.io/kubevirtInterfaces` | `<ethX>,<ethY>,...` | | Comma separated list of virtual interfaces whose inbound traffic (from VM) will be treated as outbound. |
+
+To set cluster-wide defaults for these parameters, use the corresponding
+[`global.proxy.*` parameters](/docs/reference/config/installation-options/#global-options) when installing Istio via
+the `istio` Helm chart.  For example, you can set a default `traffic.sidecar.istio.io/excludeOutboundIPRanges` option
+for all pods in the cluster using the following command to install Istio:
```
tiswanso: The PR for making the global params get into annotations on the pod wasn't allowed to get backported to 1.2.  https://github.com/istio/istio/pull/14858

rlenglet: Yeah, you're right. I had doubts about this part.
I removed this paragraph.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
 ### Logging
 
-The Istio CNI plugin is run by the container runtime process space, and, therefore, the log entries are added within
-the `kubelet` process.
+The Istio CNI plugin runs in the container runtime process space.
+Due to this, the `kubelet` process writes the plugin's log entries into its log.
+
+### Compatibility with application init containers
+
+The Istio CNI plugin may cause networking connectivity problems for any application `initContainers`. When using Istio CNI, `kubelet`
```
rcaballeromx:
```suggestion
The Istio CNI plugin may cause networking connectivity problems for any `initContainers` application.  When using Istio CNI, `kubelet`
```

rlenglet: I disagree with this suggestion. We explicitly use `application` here is to distinguish from non-application init containers such as the `istio-init` init container injected by Istio in the non-CNI case. `initContainers application` has a different meaning.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -231,5 +210,5 @@ CNI plugin only performs actions to setup the application pod's traffic redirect
 sidecar (using `iptables` in the pod's network namespace).
 
 {{< warning >}}
-This _should_ have no effect on the operations performed by the base CNI plugin configuring the pod's networking setup, although not all CNI's have been validated.
+The Istio CNI plugin should have no effect on the operations performed by the base CNI plugin configuring the pod's networking setup, although not all CNI plugins have been validated.
```
rcaballeromx: I tried to untangle the passive voice here as well. It's difficult because you have two plugins involved. What is the core message you want to tell readers? Is there anyway of phrasing this a bit more clearly?
```suggestion
The base CNI plugin that configures the pod's networking setup performs operations that the Istio CNI plugin should not affect, although not all CNI plugins have been validated.
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
-the `kubelet` process.
+The Istio CNI plugin runs in the container runtime process space.
+Due to this, the `kubelet` process writes the plugin's log entries into its log.
+
+### Compatibility with application init containers
+
+The Istio CNI plugin may cause networking connectivity problems for any application `initContainers`. When using Istio CNI, `kubelet`
+starts an injected pod with the following steps:
+1. The Istio CNI plugin sets up traffic redirection to the Istio sidecar proxy within the pod.
+1. All init containers execute and successfully complete.
```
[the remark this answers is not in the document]
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+1. The Istio sidecar proxy is started in the pod along with the pod's other containers.
```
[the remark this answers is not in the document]
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+
+Init containers execute before the sidecar proxy is started, which may result in traffic loss during their execution.
+This traffic loss can be avoided by:
```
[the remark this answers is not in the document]
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+- setting the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to disable redirecting traffic to any
```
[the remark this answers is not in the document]
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+  CIDRs the init containers communicate with,
+- and/or setting the `traffic.sidecar.istio.io/excludeOutboundPorts` annotation to disable redirecting traffic to the
```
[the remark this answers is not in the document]
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
rlenglet: I'm suggesting this:
> The Istio CNI plugin should not interfere with the operations of the base CNI plugin that configures the pod's networking setup, ...

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+
+### Compatibility with application init containers
+
+The Istio CNI plugin may cause networking connectivity problems for any application `initContainers`. When using Istio CNI, `kubelet`
+starts an injected pod with the following steps:
+1. The Istio CNI plugin sets up traffic redirection to the Istio sidecar proxy within the pod.
+1. All init containers execute and successfully complete.
+1. The Istio sidecar proxy is started in the pod along with the pod's other containers.
+
+Init containers execute before the sidecar proxy is started, which may result in traffic loss during their execution.
```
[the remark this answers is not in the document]
rlenglet: Done.

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -123,33 +63,55 @@ The following example configures the Istio CNI plugin to ignore pods in the name
 | `excludeNamespaces` | `[]string` | `[ istio-system ]` | List of namespaces to exclude from Istio pod check. |
 | `cniBinDir` | | `/opt/cni/bin` | Must be the same as the environment's `--cni-bin-dir` setting (`kubelet` parameter). |
 | `cniConfDir` | | `/etc/cni/net.d` | Must be the same as the environment's `--cni-conf-dir` setting (`kubelet` parameter). |
-| `cniConfFileName` | | None | Leave unset to auto-find the first file in the `cni-conf-dir` (as `kubelet` does).  Primarily used for testing `install-cni` plugin configuration.  If set, `install-cni` will inject the plugin configuration into this file in the `cni-conf-dir`. |
+| `cniConfFileName` | | | Leave unset to auto-find the first file in the `cni-conf-dir` (as `kubelet` does).  Primarily used for testing `install-cni` plugin configuration.  If set, `install-cni` will inject the plugin configuration into this file in the `cni-conf-dir`. |
+
+### Excluding specific Kubernetes namespaces
+
+This example uses Helm to perform the following tasks:
+- Install the Istio CNI plugin.
```
rcaballeromx: You need a blank line here for the list to render properly.
```suggestion

- Install the Istio CNI plugin.
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
-The Istio CNI plugin is run by the container runtime process space, and, therefore, the log entries are added within
-the `kubelet` process.
+The Istio CNI plugin runs in the container runtime process space.
+Due to this, the `kubelet` process writes the plugin's log entries into its log.
+
+### Compatibility with application init containers
+
+The Istio CNI plugin may cause networking connectivity problems for any application `initContainers`. When using Istio CNI, `kubelet`
+starts an injected pod with the following steps:
+1. The Istio CNI plugin sets up traffic redirection to the Istio sidecar proxy within the pod.
```
rcaballeromx: Add a blank line here for the list to render properly.
```suggestion

1. The Istio CNI plugin sets up traffic redirection to the Istio sidecar proxy within the pod.
```

content/docs/setup/kubernetes/additional-setup/cni/index.md
```diff
@@ -198,27 +160,44 @@ accompanying Istio proxy sidecar by checking that the pod meets all of the follo
...
+1. All init containers execute and complete successfully.
+1. The Istio sidecar proxy starts in the pod along with the pod's other containers.
+
+Init containers execute before the sidecar proxy starts, which can result in traffic loss during their execution.
+Avoid this traffic loss with one or both of the following settings:
+- Set the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to disable redirecting traffic to any
```
rcaballeromx: Add a blank line here for the list to render properly.
```suggestion

- Set the `traffic.sidecar.istio.io/excludeOutboundIPRanges` annotation to disable redirecting traffic to any
```

PR #4403, opened 2019-06-13
Address errors on the 'Secure Ingress Mount' page

philrud:
Fixes https://github.com/istio/istio/issues/14695

### DIFF
--- a/content/docs/tasks/traffic-management/ingress/secure-ingress-mount/index.md
+++ b/content/docs/tasks/traffic-management/ingress/secure-ingress-mount/index.md
@@ -347,7 +347,7 @@ only this time for host `bookinfo.com` instead of `httpbin.example.com`.
     used for generating your `istio.yaml`:
 
     {{< text bash >}}
-    $ helm template $HOME/istio-fetch/istio --name istio --namespace istio-system -x charts/gateways/templates/deployment.yaml --set gateways.istio-egressgateway.enabled=false \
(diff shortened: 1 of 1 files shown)


philrud:
cc @duderino

content/docs/tasks/traffic-management/ingress/secure-ingress-mount/index.md
```diff
@@ -347,7 +347,7 @@ only this time for host `bookinfo.com` instead of `httpbin.example.com`.
     used for generating your `istio.yaml`:
 
     {{< text bash >}}
-    $ helm template $HOME/istio-fetch/istio --name istio --namespace istio-system -x charts/gateways/templates/deployment.yaml --set gateways.istio-egressgateway.enabled=false \
+    $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system -x charts/gateways/templates/deployment.yaml --set gateways.istio-egressgateway.enabled=false \
```
rcaballeromx: Using the env. var. is better than hard coding the path. Could you keep that approach?

philrud: The rest of the doc expects that the user resides in the Istio distribution directory (see how bookinfo/httpbin samples are installed). This change just makes this snippet consistent with the rest of the doc. It was likely copied from another page as is, please see the referenced issue for more context.

The path is relative to the current directory, not sure what exactly is hardcoded here that should be moved to a var.

frankbu: That's correct. @rcaballeroxmx, please remove your change request so the PR can merge

PR #4405, opened 2019-06-13
Remove [x]PM unless necessary

robswhitmore:
These instructions are now for both LightStep [x]PM and LightStep Tracing.


robswhitmore:
I have a CLA (this is my second commit).

robswhitmore:
Actually, it looks like the content reverted back to what was there two revisions ago?

robswhitmore:
Can someone help? It looks like the LightStep content has reverted back to its original state - none of my changes are there now.

frankbu:
It looks like this was sent to the wrong branch, `release-1.1`. It needs to go to the `release-1.2` branch to appear in the current docs. If you send a cherry-pick PR to the `master` branch and add the label `action/merge-to-release-branch`, then it will go to both the current docs at istio.io and also stay around in future releases.

PR #4432, opened 2019-06-14
Add instructions of configuring a basic Vault server for illustration purpose

lei-tang:
Add instructions of configuring a basic Vault server for illustration purpose.

### DIFF
--- a/content/en/docs/tasks/security/vault-ca/index.md
+++ b/content/en/docs/tasks/security/vault-ca/index.md
@@ -5,12 +5,19 @@ weight: 10
 keywords: [security,certificate]
 ---
 
+{{< warning >}}
+The following information describes an experimental feature, which is intended
+for evaluation purposes only.
+{{< /warning >}}
+
 This task shows you how to integrate a [Vault Certificate Authority (CA)](https://www.vaultproject.io/) with Istio to issue certificates
 for workloads in the mesh. Istio Vault integration is an experimental feature. This task includes a demo of Istio mutual TLS using certificates issued by a Vault CA.
 
 ## Before you begin
 
 * Create a new Kubernetes cluster to run the example in this tutorial.
+* Istio Vault CA integration uses Kubernetes service account for authentication
+so it only works in Kubernetes environments. This task uses Kubernetes version 1.11.
 
 ## Certificate request flow
 
@@ -103,14 +110,13 @@ certificate signing requests to Vault.
 1.  Since the Vault CA requires the authentication and authorization of Kubernetes service accounts,
     you must edit the `vault-citadel-sa` service account to use the example Kubernetes service account
     that has already been configured on the testing Vault CA for authentication and authorization.
-    Since the testing Vault CA used in this guide has been configured to accept this example Kubernetes service
-    account for the certificate signing requests, you do not need to configure the authentication and authorization.
     When integrating your Vault server with Istio for issuing certificates, you are responsible to
     configure your Vault server's authentication and authorization for Kubernetes service accounts.
+
     To learn more about configuring a Vault CA for Kubernetes authentication and authorization,
     visit the [Vault Kubernetes `auth` method reference documentation](https://www.vaultproject.io/docs/auth/kubernetes.html).
-    A later [section](#configuring-a-basic-vault-server) includes an example to
-    configure a basic Vault server to authenticate and authorize Kubernetes service accounts.
+    See the [configuring a basic Vault server section](#configuring-a-basic-vault-server) for an example on
+    how to authenticate and authorize Kubernetes service accounts.
 
     {{< text bash >}}
     $ export SA_SECRET_NAME=$(kubectl get serviceaccount vault-citadel-sa -o=jsonpath='{.secrets[0].name}')
@@ -171,8 +177,13 @@ signing requests.
 
 The following instructions configure an example basic Vault server to authenticate and authorize a CSR
 based on the Vault Kubernetes auth method.
-**The instructions here are for illustrative purposes only. Please consult with security experts
-to set up the security configuration and certificate issuance policies of your Vault servers.** The instructions are based on the
+
+{{< warning >}}
+The instructions here are for illustrative purposes only. Please consult with security experts
+to set up the security configuration and certificate issuance policies of your Vault servers.
+{{< /warning >}}
+
+The instructions are based on the
 posts [1](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) and
 [2](https://github.com/coreos/vault-operator/blob/master/doc/user/kubernetes-auth-backend.md).
 
--- a/content/en/docs/tasks/security/vault-ca/index.md
+++ b/content/en/docs/tasks/security/vault-ca/index.md
@@ -101,12 +101,13 @@ certificate signing requests to Vault.
     {{< /text >}}
 
 1.  Since the Vault CA requires the authentication and authorization of Kubernetes service accounts,
-    you must edit the `vault-citadel-sa` service account to use the example JWT configured
-    on the testing Vault CA.
+    you must edit the `vault-citadel-sa` service account to use the example Kubernetes service account
+    that has been configured on the testing Vault CA.
     To learn more about configuring a Vault CA for Kubernetes authentication and authorization,
     visit the [Vault Kubernetes `auth` method reference documentation](https://www.vaultproject.io/docs/auth/kubernetes.html).
-    The [Integration Kubernetes with Vault - auth](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) post includes
-    detailed examples of configuring Vault to authenticate and authorize Kubernetes service accounts.
+    The example Kubernetes service account used here has been configured for authentication and authorization
+    on the testing Vault server. The [Appendix](#appendix) includes an example to
+    configure a basic Vault server to authenticate and authorize a Kubernetes service account.
 
     {{< text bash >}}
     $ export SA_SECRET_NAME=$(kubectl get serviceaccount vault-citadel-sa -o=jsonpath='{.secrets[0].name}')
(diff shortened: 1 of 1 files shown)


content/docs/tasks/security/vault-ca/index.md
```diff
@@ -101,12 +101,13 @@ certificate signing requests to Vault.
     {{< /text >}}
 
 1.  Since the Vault CA requires the authentication and authorization of Kubernetes service accounts,
-    you must edit the `vault-citadel-sa` service account to use the example JWT configured
-    on the testing Vault CA.
+    you must edit the `vault-citadel-sa` service account to use the example Kubernetes service account
+    that has been configured on the testing Vault CA.
```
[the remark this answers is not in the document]
lei-tang: This paragraph has been revised to address potential confusions.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -101,12 +101,13 @@ certificate signing requests to Vault.
...
     To learn more about configuring a Vault CA for Kubernetes authentication and authorization,
     visit the [Vault Kubernetes `auth` method reference documentation](https://www.vaultproject.io/docs/auth/kubernetes.html).
-    The [Integration Kubernetes with Vault - auth](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) post includes
-    detailed examples of configuring Vault to authenticate and authorize Kubernetes service accounts.
+    The example Kubernetes service account used here has been configured for authentication and authorization
```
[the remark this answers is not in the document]
lei-tang: This paragraph has been revised to address the places that are unclear.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -101,12 +101,13 @@ certificate signing requests to Vault.
...
+    on the testing Vault server. The [Appendix](#appendix) includes an example to
```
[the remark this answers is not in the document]
lei-tang: Changed to use a descriptive heading.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +155,123 @@ between workloads using the certificates the Vault CA issued.
 After completing this tutorial, you may delete the testing cluster created
 at the beginning of this tutorial.
 
+## Appendix
```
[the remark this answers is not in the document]
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +155,123 @@ between workloads using the certificates the Vault CA issued.
...
+
+Vault servers are hosted and managed by their owners. When signing a CSR, Istio Citadel Agent
```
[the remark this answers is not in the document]
lei-tang: Done. I chose to start here because it clarifies that Istio does not manage or configure Vault server. The owner of a Vault server is responsible of managing, configuring, and securing the server.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+        "path": {
+            "istio_ca/sign/istio-pki-role": {
+                "capabilities": ["update", "read"]
+            }
+        }
+    }
+    EOF
+    $ vault write sys/policy/istio-cert policy=@./policy.hcl
+    $ vault write auth/kubernetes/role/istio-cert \
+        bound_service_account_names=default \
```
jonmoter: Why are we using the `default` service account here?

Looking at the helm template for the `nodeagent` daemonset, it appears to be using a service account named `istio-nodeagent-service-account`.

https://github.com/istio/istio/blob/ed099bc2fefd3dbef9319634ab507738608663d4/install/kubernetes/helm/istio/charts/nodeagent/templates/daemonset.yaml#L27

So shouldn't we use that as the service account name rather than `default`?  Then you shouldn't need to do the patching of secrets up above.

If you change that, we should also change `bound_service_account_namespaces` to `istio-system`

lei-tang: The k8s service account used for Vault authentication is from a workload pod (e.g., htttpbin, sleep), so istio-nodeagent-service-account is not used.

jonmoter: Does that mean that to use Vault as a CA, I need to enable every list every single service account that will be used for workloads running in my cluster? That doesn't seem scalable at all.

My understanding is the workload makes a request to the node-agent, and then the node-agent is the one authenticating with vault to sign the CSR. So why isn't the node-agent using its service account to do that?

lei-tang: The current design is that Vault authenticates the k8s service account from a workload. IMO, the approach that node-agent uses its service account to authenticate with Vault may have a few issues:
1. Different workloads on the node share the same node agent service account, while they may run with their individual service accounts and need certificates corresponding to their individual service accounts.
2. There may be multiple nodes. If different node agents have different service accounts, these service accounts still need to be configured on Vault. If different node agents share the same service account, then the security of the system hinges on this single node agent service account, which may not be very secure.

Different designs come with different tradeoffs, some approaches are easier to configure but less secure while other approaches take more efforts to configure but is more secure.

jonmoter: So the workload passes its service account token to node-agent, and node-agent passes that along to Vault? Isn't that a separate security concern, where you've now allowed node-agent to authenticate to the Kubernetes API with different permissions?

I recognize I'm challenging a design decision in a documentation review, which probably isn't the right approach.  I'll open up a separate issue to discuss this.

Is this requirement (that the Vault role needs to include a list of all whitelisted service accounts or namespaces) documented somewhere? IMHO, that needs to be clearly spelled out.

lei-tang: node-agent does not authenticate to the Kubernetes API and does not authenticate to Vault neither; instead, node-agent passes the k8s service account from a workload to Vault and Vault does the authentication. 
It is up to the Vault owners to decide their authentication and certificate issuance policies. There is no requirement that Vault must include a list of whitelisted service accounts or namespaces; Vault owners can enforce their own authentication and authorization policies based on the k8s service account and the CSR. The instructions here only configures a basic Vault server for illustration purpose based on Vault k8s auth method.

jonmoter: > There is no requirement that Vault must include a list of whitelisted service accounts or namespaces

To configure the k8s auth method in Vault, you need to create a role. According to the [Vault docs for creating a role](https://www.vaultproject.io/api/auth/kubernetes/index.html#create-role), you need to specify one or both of:

> `bound_service_account_names (array: <required>)` - List of service account names able to access this role. If set to "*" all names are allowed, both this and bound_service_account_namespaces can not be "*".
> `bound_service_account_namespaces (array: <required>)` - List of namespaces allowed to access this role. If set to "*" all namespaces are allowed, both this and bound_service_account_names can not be set to "*". 

Since they both can't be `*`, it is necessary to have a list of either service account names or namespaces. That's what I mean by requiring a whitelist.

If nodeagent was passing in its own ServiceAccount credentials, I could just use the nodeagent's service account in `bound_service_account_names`. But since it's the workload's tokens being passed through, that won't work.

> The instructions here only configures a basic Vault server for illustration purpose based on Vault k8s auth method.

But as far as I can tell, this is the _only_ documentation there is on the Istio/Vault integration. Does NodeAgent support other methods for authenticating with Vault, other than the k8s auth method? If so, is that documented anywhere?

lei-tang: Node Agent does not mandate the authentication method on a Vault endpoint and the authentication methods used by Vault endpoints are transparent to Node Agent. Node Agent passes the k8s service account from a workload and the CSR to a Vault endpoint; the Vault owner decides the authentication and authorization methods, as well as certificate issurance policies. 
For example, one Vault owner may use Vault k8s auth method, whereas another Vault owner may use a proprietary auth method (e.g., through a proprietary Vault auth plugin) that for the k8s service account A, a certificate of identity B is issued.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
 After completing this tutorial, you may delete the testing cluster created
 at the beginning of this tutorial.
 
+## Configuring a basic Vault server
+
+The Vault server owners are responsible of hosting and managing the server. When signing a CSR, Istio Citadel Agent
```
myidpt:
```suggestion
The Vault server administrator is responsible of hosting and managing the server. When signing a CSR, Istio Citadel Agent
```

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+passes the Kubernetes service account from the workload and the CSR to your Vault server,
+which authenticate and authorize the request (e.g., based on the
+[Vault Kubernetes `auth` method](https://www.vaultproject.io/docs/auth/kubernetes.html)) and
+returns the signed certificate to Istio Citadel Agent, if the request is authorized.
+Based on their security requirements, owners of Vault servers may configure various authentication,
```
myidpt:
```suggestion
Based on their security requirements, owners of Vault servers may configure various
```

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+authorization, and certificate issuance policies for Kubernetes service accounts and certificate
```
myidpt:
```suggestion
certificate issuance policies for Kubernetes service accounts and certificate
```

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+signing requests.
+
+The following instructions configure an example basic Vault to authenticate and authorize a CSR
```
myidpt:
```suggestion
The following instructions configure an example basic Vault server to authenticate and authorize a CSR
```

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+based on the Vault Kubernetes auth method.
+**The instructions here are for illustrative purposes only. Please consult with security experts
+on the security configuration and certificate issuance policies of your Vault servers.** The instructions are based on the
```
myidpt:
```suggestion
to set up the security configuration and certificate issuance policies of your Vault servers.** The instructions are based on the
```

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+post [1](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) and
```
myidpt:
```suggestion
posts [1](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) and
```

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
 After completing this tutorial, you may delete the testing cluster created
 at the beginning of this tutorial.
 
+## Configuring a basic Vault server
+
+The Vault server owners are responsible of hosting and managing the server. When signing a CSR, Istio Citadel Agent
+passes the Kubernetes service account from the workload and the CSR to your Vault server,
```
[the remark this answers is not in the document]
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+which authenticate and authorize the request (e.g., based on the
```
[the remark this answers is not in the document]
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Done.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+**The instructions here are for illustrative purposes only. Please consult with security experts
+to set up the security configuration and certificate issuance policies of your Vault servers.** The instructions are based on the
+posts [1](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) and
+[2](https://coreos.com/tectonic/docs/latest/vault-operator/user/kubernetes-auth-backend.html).
+
+1.  Create a Kubernetes cluster to host an example basic Vault server.
+    In the Kubernetes cluster created, install, initialize, unseal, and login Vault.
+    Examples of install, initialize, unseal, and login Vault can be found in the post
+    [1](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/).
+    The example Vault server used in this guide is of 0.10.3 version.
```
tmshort: `0.10.3` i that the correct version? The latest version is 1.1.3.

content/docs/tasks/security/vault-ca/index.md
```diff
@@ -154,3 +158,123 @@ between workloads using the certificates the Vault CA issued.
...
+Based on their security requirements, owners of Vault servers may configure various
+certificate issuance policies for Kubernetes service accounts and certificate
+signing requests.
+
+The following instructions configure an example basic Vault server to authenticate and authorize a CSR
+based on the Vault Kubernetes auth method.
+**The instructions here are for illustrative purposes only. Please consult with security experts
+to set up the security configuration and certificate issuance policies of your Vault servers.** The instructions are based on the
+posts [1](https://evalle.xyz/posts/integration-kubernetes-with-vault-auth/) and
+[2](https://coreos.com/tectonic/docs/latest/vault-operator/user/kubernetes-auth-backend.html).
```
tmshort: This documentation is moving to redheat.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Yes, 0.10.3 is the version used in this example Vault server. The Vault API of version 0.10.3 is similar to that of the latest version.

tmshort: That should be clearer; this documentation mentions just "the example Vault server" without the term "API", whereas the Vault documentation generally refers to the product version. Specifically the [download page](https://www.vaultproject.io/downloads.html) references 1.1.3. Leaving only 0.10.3 in this documentation will confuse the user looking to download Vault version 0.10.3 which isn't available. I would suggesting stating the release version, and if you believe the API version  is important, *clearly* indicate that it's the API version.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Thanks! The links have been updated based on the deprecation notice on the original webpages.

content/docs/tasks/security/vault-ca/index.md
lei-tang: Revised to state the release version and added the link to the release.

tmshort:
Step 1 of this example has:
```
$ kubectl apply -f install/kubernetes/helm/istio-init/files/crd/
```

This step will fail since there is no `crd` directory. This should really be changed to:

```
$ kubectl apply -f install/kubernetes/helm/istio-init/files/crd*
```

lei-tang:
> Step 1 of this example has:
> 
> ```
> $ kubectl apply -f install/kubernetes/helm/istio-init/files/crd/
> ```
> 
> This step will fail since there is no `crd` directory. This should really be changed to:
> 
> ```
> $ kubectl apply -f install/kubernetes/helm/istio-init/files/crd*
> ```

@tmshort 
Done, thanks!

myidpt:
Lei, are you still working on this PR?

lei-tang:
@myidpt I have addressed all review comments. @rcaballeromx can you approve the PR?

costinm:
AFAIK the plan is to support only k8s trustworthy tokens - at least this is
what was agreed on last TOC.
The non-trustworthy tokens are not trust worthy - and results in install
complexity and risks.

This feature was never 'beta' - lack of testing, etc - so I don't think we
need a deprecation plan. If Vault integration
can be supported with the trusted tokens - we need to update the doc,
otherwise we need to hide the doc
until we have  an implementation.

geeknoid:
@lei-tang Do you intend to push this over the finish line?

lei-tang:
@geeknoid Yes, I intend to push this PR over the finish line. However, the Vault integration guide is temporarily removed from Istio 1.3 (due to the new k8s trustworthy token format). After we add back the Vault integration guide, I will push this PR into the Vault integration guide. Before that happens, users can still refer to this PR for the instructions of configuring a basic Vault server.

tmshort:
We shouldn't let this go stale.

lei-tang:
This PR is still needed for helping users on configuring a basic Vault server for illustration purpose.

sdake:
@lei-tang Is this PR relevant to February master? This PR has been open for about 8 months. I had thougt the Vault implementation had been deprecated. If this PR is no longer relevant please close it. If it is still relevant, please rebase as necessary and answer reviewer commentary so we can get this PR moving before the 1.5 release.

Cheers
-steve

lei-tang:
Close this PR for now and it will be reopened when the Vault support is added back to a newer Istio release.

PR #4433, opened 2019-06-14
Refine the health check doc.

incfly:
Fixes https://github.com/istio/istio/issues/13074

- Add annotation based for prober rewrite approach
- focusing on mtls turned on when http type health check is enabled. Remove the mtls if/else for command line option, it's not needed. (IMO, we don't even need to document cmd option, just mentioning cmd will be an alternative is good enough.
- move policy/destionatinrule/cleanup to the right place.


content/docs/ops/setup/app-health-check/index.md
```diff
@@ -8,15 +8,16 @@ aliases:
...
-There are three options for liveness and readiness probes in Kubernetes:
+[Kubernetes liveness and readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/)  offers three options for liveness and readiness probes:
 
 1. Command
-1. HTTP request
 1. TCP request
+1. HTTP request
 
-This task provides examples for the first two options with Istio mutual TLS enabled and disabled, respectively.
+This task explains how these approaches work when Istio mutual TLS is enabled.
```
rcaballeromx:
```suggestion
This task shows how to use these approaches in Istio with mutual TLS is enabled.
```

content/docs/ops/setup/app-health-check/index.md
```diff
@@ -8,15 +8,16 @@ aliases:
...
+
+Command and TCP type probes work with Istio regardless whether mutual TLS is enabled or not. HTTP request requires different Istio configuration when
```
rcaballeromx:
```suggestion
Command and TCP type probes work with Istio regardless of whether or not mutual TLS is enabled. The HTTP request approach requires different Istio configuration with
```

content/docs/ops/setup/app-health-check/index.md
```diff
@@ -27,27 +28,7 @@ This task provides examples for the first two options with Istio mutual TLS enab
...
-{{< text bash >}}
-$ kubectl get pod
-NAME                             READY     STATUS    RESTARTS   AGE
-liveness-6857c8775f-zdv9r        2/2       Running   0           1m
-{{< /text >}}
-
-The number '0' in the 'RESTARTS' column means liveness probes worked fine. Readiness probes work in the same way and you can modify `liveness-command.yaml` accordingly to try it yourself.
-
-### Mutual TLS enabled
+In this section, you first configure health checking when mutual TLS is enabled.
```
rcaballeromx: I fixed the sentence but the text makes no logical sense to readers. How they can do this is not discussed. Please revise.
```suggestion
First, you need to configure health checking with mutual TLS enabled.
```

content/docs/ops/setup/app-health-check/index.md
```diff
@@ -8,15 +8,16 @@ aliases:
...
 1. Command
-1. HTTP request
 1. TCP request
+1. HTTP request
 
-This task provides examples for the first two options with Istio mutual TLS enabled and disabled, respectively.
+This task explains how these approaches work when Istio mutual TLS is enabled.
+
+Command and TCP type probes work with Istio regardless whether mutual TLS is enabled or not. HTTP request requires different Istio configuration when
+Istio mutual TLS is enabled.
```
rcaballeromx:
```suggestion
mutual TLS enabled.
```

content/docs/ops/setup/app-health-check/index.md
```diff
@@ -84,10 +65,9 @@ Follow these steps to complete the configuration:
     EOF
     {{< /text >}}
 
-Run this command to re-deploy the service:
+Run this command to deploy the service:
```
rcaballeromx:
```suggestion
Run the following command to deploy the service:
```

content/docs/ops/setup/app-health-check/index.md
```diff
@@ -99,55 +79,23 @@ NAME                             READY     STATUS    RESTARTS   AGE
...
-{{< /text >}}
-
-Wait for a minute and check the pod status to make sure the liveness probes work with '0' in the 'RESTARTS' column.
-
-{{< text bash >}}
-$ kubectl get pod
-NAME                             READY     STATUS    RESTARTS   AGE
-liveness-http-975595bb6-5b2z7c   2/2       Running   0           1m
-{{< /text >}}
+This section shows how to configure health checking with the HTTP request option when mutual TLS is enabled.
```
rcaballeromx: Why is this here if the readers need to do these steps first? Please ensure the order of the instructions works well after removing the content.

content/docs/ops/setup/app-health-check/index.md
```diff
@@ -166,7 +114,18 @@ spec by yourself.
 The configuration changes above (by Helm or by the configuration map) effect all Istio app deployments.
 {{< /warning >}}
 
-##### Re-deploy the liveness health check app
+#### Use Annotations on Pod
+
+<!-- Add samples YAML or kubectl patch? -->
```
incfly: I'm thinking to add sample YAML inlined here, as 

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: liveness-http
spec:
  template:
    metadata:
      labels:
        app: liveness-http
        version: v1
      annotations:
        sidecar.istio.io/rewriteAppHTTPProbers: "true"
    spec:
      containers:
      - name: liveness-http
        image: docker.io/istio/health:example
        ports:
        - containerPort: 8001
        livenessProbe:
          httpGet:
            path: /foo
            port: 8001
          initialDelaySeconds: 5
          periodSeconds: 5
```

And let people to do kubectl apply, as opposed to just mention the annotation. I think that makes things easier to follow? wtdy? @rcaballeromx

content/docs/ops/setup/app-health-check/index.md
incfly: You mean the steps to enable Istio installation with helm flag `sidecarInjectorWebhook.rewriteAppHTTPProbe=true`? Some apporoach like prober rewrite by annotation does not require this helm flag, same for the separate port.

What do you think? Maybe just mention this flag as an option in `Before you Begin`?

content/docs/ops/setup/app-health-check/index.md
linsun: code snippet to copy is best imho

content/docs/ops/setup/app-health-check/index.md
```diff
@@ -166,7 +115,18 @@ spec by yourself.
 The configuration changes above (by Helm or by the configuration map) effect all Istio app deployments.
 {{< /warning >}}
 
-##### Re-deploy the liveness health check app
+#### Use Annotations on Pod
+
```
linsun: I think this option should go first.  It is most likely users will use annotation before enable it globally.

content/docs/ops/setup/app-health-check/index.md
```diff
@@ -99,55 +80,23 @@ NAME                             READY     STATUS    RESTARTS   AGE
...
-This approach rewrites the application `PodSpec` liveness probe, such that the probe request will be sent to
+This approach rewrites the application `PodSpec` readiness/liveness probe, such that the probe request will be sent to
 [Pilot agent](/docs/reference/commands/pilot-agent/). Pilot agent then redirects the
 request to application, and strips the response body only returning the response code.
 
-To use this approach, you need to configure Istio to rewrite the liveness HTTP probes.
+You have two ways to enable Istio to rewrite the liveness HTTP probes.
 
-##### Configure Istio to rewrite liveness HTTP probes
+#### Enable via Helm Option
```
linsun: add globally ?

content/docs/ops/setup/app-health-check/index.md
incfly: Maybe we should even focus on document this approach and then in the end mention the Helm flag can turn it on globally as a side note? That makes the doc even simpler.

rcaballeromx:
@linsun I approved the PR and filed issue #4478 to follow up after release and revisit the doc more holistically. That way we can figure out the right order and the content flow post release.
@incfly Please remove the trailing whitespaces for the test to pass.

content/docs/ops/setup/app-health-check/index.md
incfly: This might suits well in the follow-up improvements.

content/docs/ops/setup/app-health-check/index.md
incfly: Added.

content/docs/ops/setup/app-health-check/index.md
incfly: Added.

PR #4446, opened 2019-06-17
Added docs for imagePullSecret

brian-avery:
Noticed that this was in the helm charts but never documented.

### DIFF
--- a/content/docs/reference/config/installation-options/index.md
+++ b/content/docs/reference/config/installation-options/index.md
@@ -137,7 +137,7 @@ To customize Istio install using Helm, use the `--set <key>=<value>` option in H
 | --- | --- | --- |
 | `global.hub` | `gcr.io/istio-release` | `Default hub for Istio images. Releases are published to docker hub under 'istio' project. Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly` |
 | `global.tag` | `release-1.1-latest-daily` | `Default tag for Istio images.` |
-| `global.imagePullSecrets` | `` | `List of secrets to use when pulling images in the case of a secure registry.` |
+| `global.imagePullSecrets` | `` | `List of secrets to use when pulling Istio images from a secure registry.` |
 | `global.monitoringPort` | `15014` | `monitoring port used by mixer, pilot, galley` |
 | `global.k8sIngress.enabled` | `false` |  |
 | `global.k8sIngress.gatewayName` | `ingressgateway` | `Gateway used for k8s Ingress resources. By default it is using 'istio:ingressgateway' that will be installed by setting 'gateways.enabled' and 'gateways.istio-ingressgateway.enabled' flags to true.` |
--- a/content/docs/reference/config/installation-options/index.md
+++ b/content/docs/reference/config/installation-options/index.md
@@ -137,6 +137,7 @@ To customize Istio install using Helm, use the `--set <key>=<value>` option in H
 | --- | --- | --- |
 | `global.hub` | `gcr.io/istio-release` | `Default hub for Istio images. Releases are published to docker hub under 'istio' project. Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly` |
 | `global.tag` | `release-1.1-latest-daily` | `Default tag for Istio images.` |
+| `global.imagePullSecrets` | `` | `List of secrets to use when pulling images in the case of a secure registry.` |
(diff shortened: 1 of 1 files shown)


linsun:
cc @johnma14 

@brian-avery we don't normally update the page manually.  @johnma14 updates it via script.  (she has a PR out for it).

brian-avery:
@linsun @johnma14  Thanks. Oddly this functionality came out last year (August?) but never got added to the docs.

brian-avery:
https://github.com/istio/istio/commit/7759947ccd866a22bf87823bcda205f1b09b0b7a

content/docs/reference/config/installation-options/index.md
```diff
@@ -137,6 +137,7 @@ To customize Istio install using Helm, use the `--set <key>=<value>` option in H
 | --- | --- | --- |
 | `global.hub` | `gcr.io/istio-release` | `Default hub for Istio images. Releases are published to docker hub under 'istio' project. Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly` |
 | `global.tag` | `release-1.1-latest-daily` | `Default tag for Istio images.` |
+| `global.imagePullSecrets` | `` | `List of secrets to use when pulling Istio images from a secure registry.` |
```
rcaballeromx:
```suggestion
| `global.imagePullSecrets` | `` | `Lists the secrets you need to use to pull Istio images from a secure registry.` |
```

johnma14:
@brian-avery @linsun is this something we need to include in 1.2 as well. I will go ahead and verify whether these changes can be captured.

linsun:
@johnma14 yes!  it is part of 1.2, probably introduced in 1.0 or 1.1.

johnma14:
@brian-avery @rcaballeromx I have couple of questions regarding the imagePullSecrets configuration option:

- If we want this option to show in the installation-options file, the configuration option should be of the format: <configuration-option>: <value>. If empty, you can use "" or if this is a list, it could be []. The way it is listed in the istio/values.yaml file, it is skipping this option :

> # ImagePullSecrets for all ServiceAccount, list of secrets in the same namespace
> # to use for pulling any images in pods that reference this ServiceAccount.
> # For components that don't use ServiceAccounts (i.e. grafana, servicegraph, tracing)
> # ImagePullSecrets will be added to the corresponding Deployment(StatefulSet) objects.
> # Must be set for any cluster configured with private docker registry.
`imagePullSecrets: 
    # - private-registry-key`

1) Can I go ahead and make the value an empty list. For example:
` imagePullSecrets: []`
Will that break anything or is that fine.

2) The way we fill out descriptions for a configuration option is by using the description provided above the configuration option in the yaml file. So in this case, it would be the description right above this option in the values.yaml file. If we have to use the description you have provided in the PR, we would have to replace the description in the values.yaml file with the one you provided. Otherwise, every time we generate the installation-options file, your description will be replaced. 

Thanks @brian-avery @rcaballeromx. Appreciate your feedback.

johnma14:
@linsun a question for you - Once we get this option in the installation-option file, do we want this to show up in the installation-option-changes file. By default, it will show up since this option was not there in the 1.1 version of the installation-option file. Appreciate your feedback.

linsun:
@johnma14 - I think it is okay to show up in installation-option-changes file, because this is a new thing added and won't break others during upgrade.

linsun:
@johnma14 we may have an issue if we change it to [].  I noted it is used mostly in checking if the value exists (this may not work when it is an array), then go through it.

```
{{- if $.Values.global.imagePullSecrets }}
imagePullSecrets:
{{- range $.Values.global.imagePullSecrets }}
```

content/docs/reference/config/installation-options/index.md
```diff
@@ -137,6 +137,7 @@ To customize Istio install using Helm, use the `--set <key>=<value>` option in H
 | --- | --- | --- |
 | `global.hub` | `gcr.io/istio-release` | `Default hub for Istio images. Releases are published to docker hub under 'istio' project. Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly` |
 | `global.tag` | `release-1.1-latest-daily` | `Default tag for Istio images.` |
+| `global.imagePullSecrets` | `` | `Lists the secrets you need to use to pull Istio images from a secure registry.` |
```
sdake: this is the wrong way to document this option.  Please modify values.yaml for the relevant component as this documentation is autogenerated.

brian-avery:
This should be implemented in values.yaml instead. Closing this PR.

PR #4457, opened 2019-06-17
Remove "sidecars" to troubleshoot Hugo issue.

rcaballeromx:


geeknoid:
Rendered HTML isn't right, this isn't the fix.

PR #4461, opened 2019-06-17
Remove "service-entries" to troubleshoot Hugo.

rcaballeromx:


geeknoid:
CircleCi shows the bad errors, so this isn't the solution.

PR #4462, opened 2019-06-17
Remove "routing" to troubleshoot Hugo.

rcaballeromx:


geeknoid:
Rendered site is missing the content, so this isn't the fix.

PR #4459, opened 2019-06-17
Remove "destination-rules" to troubleshoot Hugo.

rcaballeromx:
Signed-off-by: rcaballeromx <grca@google.com>

PR #4463, opened 2019-06-17
Remove "network" to troubleshoot Hugo.

rcaballeromx:


geeknoid:
This one passed CircleCI and Netlify, a potential winner.

rcaballeromx:
The thing is that the error cannot be in either, because this contains the files that #4459 removes and vice versa.

geeknoid:
Rerunning CI, I now see a bunch of errors that shouldn't be there. So this isn't the solution.

PR #4469, opened 2019-06-18
Release note 1.2

linsun:
based on draft that team has populated: https://github.com/istio/istio/wiki/Release-Note-1.2-Draft

PR #4474, opened 2019-06-18
add ipv6 support to sync with release note

linsun:


linsun:
IPv6 in .spelling is fixed in #4469

PR #4482, opened 2019-06-18
Document commands for running Kiali on OpenShift

mwringe:
Kiali installed on OpenShift using the Istio installers will cause Kiali to not function properly. This is due to Kiali missing permission to access some OpenShift specific resources.

This PR adds to the documentation what commands are needed in order to grant Kiali these permissions.


jmazzitelli:
This looks good to me.

Issue #4400, opened 2019-06-13
[1.2 RC1] Kiali Comments


brian-avery:
From: https://preliminary.istio.io/docs/tasks/telemetry/kiali/

My sidebar for the screenshots was different than the ones shown in the screenshots. 
![image](https://user-images.githubusercontent.com/8484260/59463379-d9134980-8df3-11e9-8cba-83cc30ef19e4.png)

For step 8: _To view a namespace graph, click on the bookinfo namespace in the Bookinfo namespace card. The page looks similar to:_

* Bookinfo deploys into the default namespace on Istio by default, not the bookinfo namespace
* The instructions talk about clicking on the bookinfo namespace in the Bookinfo namespace card, but this doesn't really tell you what to click on. The card itself isn't clickable and there are a couple of buttons at the bottom, but none of them say "namespace". Would it be better to say graph?

brian-avery:
The login page also doesn't render properly on Firefox.

jshaughn:
This is fixed in the latest release of kiali.  This issue can be closed.

PR #4485, opened 2019-06-18
Update description of Kiali graph icon

mwringe:
Update to make it more clear what should be clicked on in the Kiali overview page to get to the graph view.

Should help solve: https://github.com/istio/istio.io/issues/4400


jmazzitelli:
while you are here, fix the "public API" link at the bottom:)

PR #4489, opened 2019-06-19
Clarify sds install profile

### LINKED ISSUE #4416: Documentation on SDS installation profile is unclear
A few issues with the doco around the SDS installation profile:

1. There must be a better way of explaining that there is no `sds` profile, only an `sds-auth` profile. Changing the text in the numbered list of profiles to be `sds-auth` instead of just `sds` would be a start, perhaps with something to explain further.
2. The table that explains which additional features are enabled with the `-auth` suffix further confuses the issue, because it's not so much that these are 'enabled' for `sds` when you add the suffix, it's that they're there by default with no option to not have them. Again, confusing, a note explaining that they are enabled by default would help.
3. Why is control plane auth disabled for the SDS profile? My (naive) take is that SDS is the maximally secure deployment strategy, so seeing control plane auth disabled is very puzzling. Is it incompatible? Is it not-required? Will everything break if I create a custom profile that enables both SDS and control plan auth? I actually don't know the answer to this one, would appreciate input.

Happy to submit a PR over the weekend.

Niksko:
Closes #4416 

This adds clarity around the SDS install profile
* Clarifies that there is no non-auth variant by changing the name to `sds-auth`
* Adds clarity around additional auth features being enabled by default, instead of vague 'only comes with authentication'
* Adds tip around why control plane auth is disabled in the sds profile

### DIFF
--- a/content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
@@ -53,16 +53,15 @@ The components marked as **X** are installed within each profile:
 Some profiles have an authentication variant, with `-auth` appended to the name, which adds the following
 security features to the profile:
 
+{{< tip >}}
+Control plane security with SDS is planned for an upcoming release.
+{{< /tip >}}
+
 | | default | demo | minimal | sds-auth |
 | --- | --- | --- | --- | --- |
 | Control Plane Security | | X | | |
 | Strict Mutual TLS | | X | | X |
 | SDS | | | | X |
-
-{{< tip >}}
-Control plan security still requires the use of secret volume mounts, and is therefore
-incompatible with SDS.
-{{< /tip >}}
 
 To further customize Istio and install addons, you can add one or more `--set <key>=<value>` options in the `helm template` or `helm install` command that you use when installing Istio. The [Installation Options](/docs/reference/config/installation-options/) lists the complete set of supported installation key and value pairs.
 
--- a/content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
@@ -57,11 +57,11 @@ security features to the profile:
 Control plane security with SDS is planned for an upcoming release.
 {{< /tip >}}
 
-| | default | demo | minimal | sds-auth |
-| --- | --- | --- | --- | --- |
-| Control Plane Security | | X | | |
-| Strict Mutual TLS | | X | | X |
-| SDS | | | | X |
+| Security feature | demo-auth | sds-auth |
+| --- | --- | --- |
+| Control Plane Security | X | |
+| Strict Mutual TLS | X | X |
+| SDS | | X |
 
 To further customize Istio and install addons, you can add one or more `--set <key>=<value>` options in the `helm template` or `helm install` command that you use when installing Istio. The [Installation Options](/docs/reference/config/installation-options/) lists the complete set of supported installation key and value pairs.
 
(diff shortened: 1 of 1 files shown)


content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
```diff
@@ -26,8 +26,8 @@ your specific needs. The following built-in configuration profiles are currently
 
 1. **minimal**: the minimal set of components necessary to use Istio's [traffic management](/docs/tasks/traffic-management/) features.
 
-1. **sds**: similar to the **default** profile, but also enables Istio's [SDS (secret discovery service)](/docs/tasks/security/auth-sds).
-    This profile comes only with authentication enabled.
+1. **sds-auth**: similar to the **default** profile, but also enables Istio's [SDS (secret discovery service)](/docs/tasks/security/auth-sds).
+    This profile comes with additional authentication features enabled by default (see table below).
```
Niksko: The '(see table below)' is probably not ideal. Is this necessary, or should I just omit it and leave the user to naturally find out later when they read further?

content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
```diff
@@ -53,12 +53,17 @@ The components marked as **X** are installed within each profile:
...
 
-| | default | demo | minimal | sds |
+| | default | demo | minimal | sds-auth |
 | --- | --- | --- | --- | --- |
 | Control Plane Security | | X | | |
 | Strict Mutual TLS | | X | | X |
 | SDS | | | | X |
 
+{{< tip >}}
+Control plan security still requires the use of secret volume mounts, and is therefore
```
sdake: plan/plane

content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
```diff
@@ -53,12 +53,17 @@ The components marked as **X** are installed within each profile:
 Some profiles have an authentication variant, with `-auth` appended to the name, which adds the following
 security features to the profile:
 
-| | default | demo | minimal | sds |
+| | default | demo | minimal | sds-auth |
```
sdake: for the first field, you might indicate `Security Model` or something equivalent.

content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
sdake: agreed.  I'd remove the ` (see table below)`

content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
```diff
@@ -53,12 +53,17 @@ The components marked as **X** are installed within each profile:
...
 | --- | --- | --- | --- | --- |
 | Control Plane Security | | X | | |
 | Strict Mutual TLS | | X | | X |
 | SDS | | | | X |
 
+{{< tip >}}
```
sdake: preceeed the table with the tip please.

geeknoid:
I added the label to auto-merge this to 1.2 once it is checked in to master.

content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
frankbu: This table is a little confusing. Would it be better to just have 2 columns:
```suggestion
| | demo-auth | sds-auth |
```

content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
```diff
@@ -53,12 +53,17 @@ The components marked as **X** are installed within each profile:
...
+Control plan security still requires the use of secret volume mounts, and is therefore
+incompatible with SDS.
```
quanjielin: controlplane sds is WIP (there isn't technical blocker but just timeline) 
probably tweak somethinlg like ```enable sds on controlplane is work in progress``` ?

content/docs/setup/kubernetes/additional-setup/config-profiles/index.md
Niksko: Putting the tip before the table looks a little strange to my eye in the preview. Anything we can do about this? Or does it look as expected?

PR #4493, opened 2019-06-19
Apply crds separately to avoid race

### LINKED ISSUE #4421: crd files are catted into a yaml instead of installed independently
The following files are using a workflow that works most of the time but not all of the time.  The kubeapi races the application of the CRDS with the consumption of the CRDs (by CRs) in the rest of the manifest.

```
tasks/security/vault-ca/index.md:    $ cat install/kubernetes/helm/istio-init/files/crd-* >> istio-auth.yaml
tasks/security/auth-sds/index.md:    $ cat install/kubernetes/helm/istio-init/files/crd-* >> istio-auth-sds.yaml
tasks/multicluster/gke/index.md:$ cat install/kubernetes/helm/istio-init/files/crd-* > $HOME/istio_master.yaml
```

Instead, the workflow should look more like what is documented here:

https://preliminary.istio.io/docs/setup/kubernetes/install/multicluster/gateways/#deploy-the-istio-control-plane-in-each-cluster

Niksko:
Closes #4421 

Cat-ing the crds into a single file along with the istio mainfests
leads to a race to install the crds. This applies the crds as a
separate step to avoid this.

### DIFF
--- a/content/docs/tasks/multicluster/gke/index.md
+++ b/content/docs/tasks/multicluster/gke/index.md
@@ -112,8 +112,8 @@ the `default` namespace:
 
 {{< text bash >}}
 $ kubectl config use-context "gke_${proj}_${zone}_cluster-1"
-$ cat install/kubernetes/helm/istio-init/files/crd-* > $HOME/istio_master.yaml
-$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system >> $HOME/istio_master.yaml
+$ kubectl apply -f install/kubernetes/helm/istio-init/files/crd/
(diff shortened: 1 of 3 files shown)


sdake:
@Niksko - I think in the runup to 1.2 I missed this PR.  My apologies...  Our workflow for the docs repo is a little different then the rest of the repositories, as we want live publishing.  Can you send this PR to the release-1.2 branch?

A forward port will happen automatically in the future.

Thanks again for the contribution and this PR lgtm - but wrong branch (my error).

Cheers
-steve

geeknoid:
CI failures were transient, so forcing in.

PR #4494, opened 2019-06-19
Trying to restore SkyWalking back online.

wu-sheng:
Since SkyWalking graduated as Top Level Project in Apache, the `incubator-` prefix removed from all repos. 

This PR is trying to change the repo name and make it back online.

But from the changelogs, I am not sure whether it is the reason.

### DIFF
--- a/content/docs/reference/config/policy-and-telemetry/adapters/apache-skywalking/index.html
+++ b/content/docs/reference/config/policy-and-telemetry/adapters/apache-skywalking/index.html
@@ -12,8 +12,8 @@ See the License for the specific language governing permissions and
 limitations under the License. -->
  
 ---
-WARNING: THIS IS AN AUTO-GENERATED FILE, DO NOT EDIT. PLEASE MODIFY THE ORIGINAL SOURCE IN THE https://github.com/apache/incubator-skywalking-data-collect-protocol REPO
-source_repo: https://github.com/apache/incubator-skywalking-data-collect-protocol
+WARNING: THIS IS AN AUTO-GENERATED FILE, DO NOT EDIT. PLEASE MODIFY THE ORIGINAL SOURCE IN THE https://github.com/apache/skywalking-data-collect-protocol REPO
+source_repo: https://github.com/apache/skywalking-data-collect-protocol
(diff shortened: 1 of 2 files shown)


wu-sheng:
Looking at the preview page, verified my guess, still missed there. https://deploy-preview-4494--preliminary-istio.netlify.com/docs/reference/config/policy-and-telemetry/adapters/

The first one should be SkyWalking, and displayed as a blank area. @geeknoid Could you help me find out why?

The `adapters/apache-skywalking/index.html` looks like still there.

geeknoid:
I took care of this when doing a full refresh of the reference docs. Thanks for pointing it out.

wu-sheng:
@geeknoid From I checked, still no there, is it my browser issue? 
https://istio.io/docs/reference/config/policy-and-telemetry/adapters/

geeknoid:
It should be here: https://preliminary.istio.io/docs/reference/config/policy-and-telemetry/adapters/

This will be published as istio.io when Istio 1.3 is released. If this needs to go on istio.io sooner, then we'll need to push a change to the release-1.2 branch separately.

wu-sheng:
@geeknoid Here is I saw at this page. Apache SkyWalking based on alphabeta order should be the first one, but I can only see it is blank zone.
image

Are you seeing as same as I do? Or I have some network or cache issue?

geeknoid:
Yes, I see the same thing. I just looked into it and the problem is that the HTML file we're pulling from https://github.com/apache/skywalking-data-collect-protocol.git starts with a copyright notice, which completely throws off our doc system. 

Are you in control of this repo? If so, please take off the copyright notice, or at least move it below the front-matter section of the file. The front-matter is the stuff between the two --- sequences.

Thanks.

wu-sheng:
@geeknoid I just update the repo file, https://github.com/apache/skywalking-data-collect-protocol/blob/master/service-mesh-probe/istio/skywalking.config.pb.html

Is this good enough? If yes, could you help me update 1.3 and 1.2 branches, please?

geeknoid:
Cool. I'll update this in 1.3 and merge to 1.2.

PR #4495, opened 2019-06-19
Fix list not displaying correctly

frankbu:


geeknoid:
Could you rebase please? Seems to have picked up some unexpected changes in the PR.

PR #4500, opened 2019-06-19
Install istio-cni into kube-system to use system-node-critical priorityClass

rlenglet:
Fixes https://github.com/istio/istio/issues/14977
Fixes https://github.com/istio/cni/pull/126
Related to https://github.com/kubernetes/kubernetes/pull/65593

### DIFF
--- a/content/docs/setup/kubernetes/additional-setup/cni/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/cni/index.md
@@ -82,7 +82,7 @@ Use the following command to render and apply Istio CNI components and override 
 `istio-cni` Helm chart's `logLevel` and `excludeNamespaces` parameters:
 
 {{< text bash >}}
-$ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=istio-system \
+$ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=kube-system \
     --set logLevel=info \
     --set excludeNamespaces={"istio-system,kube-system,foo_ns,bar_ns"} | kubectl apply -f -
 {{< /text >}}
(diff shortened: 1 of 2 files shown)


rlenglet:
This will need to be backported to `release-1.2`.

linsun:
@rlenglet why we need this change? is it because a newer version of k8s?  I'm sure we tested the instruction in 1.1.

frankbu:
If this change is required in the current 1.2 docs, you should send the PR to the `release-1.2` branch, instead of master. The commit will be cherry-picked to master later, hopefully by a bot.

rlenglet:
> If this change is required in the current 1.2 docs, you should send the PR to the release-1.2 branch, instead of master. The commit will be cherry-picked to master later, hopefully by a bot.

Right. I'll change the target branch to `release-1.2`. Thanks!

> why we need this change? is it because a newer version of k8s? I'm sure we tested the instruction in 1.1.

Adding the `priorityClass` is a change I made late in 1.2, after I tested the 1.2 instructions. I missed the bug because our e2e tests do install `istio-cni` into `kube-system` (cf. https://github.com/istio/cni/blob/df641f973b6c7f0c69912e6bf25fc969b12d400a/.circleci/run_e2e.sh#L27).

Since this is the only tested namespace, and the only working configuration, this should be reflected in the doc.

linsun:
@rlenglet thanks for the explanation.  Do you foresee any issue with this when upgrading from 1.1 while using CNI and switched from istio-system ns to kube-system ns?  I think users could end up running CNI in both nses.

sdake:
@frankbu I thought the workflow was merge to master first, then cherrypick?

cheers
-steve

frankbu:
@sdake, it turns out that the merge-to-release-branch bot that @geeknoid added wasn't working reliably (sometimes picking up extra commits), so he had to disable it. So for now we are back to the old process of just sending PRs that belong in 1.2 to the release-1.2 branch and then in a few weeks from now we will do a bulk cherry-pick. Martin is also going to experiment with changing the merge bot to work the other way, i.e., send PRs to release branch with a `merge-to-master` label to do the automatic cherry-pick back to master.

PR #4502, opened 2019-06-19
Add new Traffic management concept.

rcaballeromx:
Improved based on review comments.

Fix Hugo front matter.

Fix false positives in links test.

Remove terms from exceptions file.

corrections

review comments


bep:
What you see is most likely a data race due to parallel processing. I would guess that `hugo --path-warnings` should help you here, but in short:

```
content/docs/concepts/traffic-management/_index.md 
content/docs/concepts/traffic-management/index.md 
```

The above does not work.

geeknoid:
Thanks for looking into this.

This started out with index.md, but that file was deleted in the PR and is
replaced by _index.md as seen in the commit:
https://github.com/istio/istio.io/tree/f60562d2855320f38adbda908ce03eca74520f8d/content/docs/concepts/traffic-management

I'll see about --path-warnings, maybe that'll output something useful.

PR #4518, opened 2019-06-20
Document CNI upgrade instructions

rlenglet:
Fixes https://github.com/istio/istio/issues/14977

### DIFF
--- a/content/docs/setup/kubernetes/upgrade/steps/index.md
+++ b/content/docs/setup/kubernetes/upgrade/steps/index.md
@@ -36,11 +36,87 @@ are supported. If you are on an older version, please upgrade to 1.1 first.
 [Download the new Istio release](/docs/setup/kubernetes/#downloading-the-release)
 and change directory to the new release directory.
 
+### Istio CNI upgrade
+
+If you have installed or are planning to install [Istio CNI](/docs/setup/kubernetes/additional-setup/cni/),
+choose one of the following **mutually exclusive** options to check whether
+Istio CNI is already installed and to upgrade it:
+
+{{< tabset cookie-name="controlplaneupdate" >}}
+{{< tab name="Kubernetes rolling update" cookie-value="k8supdate" >}}
+
+You can use Kubernetes’ rolling update mechanism to upgrade the Istio CNI components.
+This is suitable for cases where `kubectl apply` was used to deploy Istio CNI.
+
+1. To check whether `istio-cni` is installed, search for `istio-cni-node` pods
+   and in which namespace they are running (typically, `kube-system` or `istio-system`):
+
+    {{< text bash >}}
+    $ kubectl get pods -l k8s-app=istio-cni-node --all-namespaces
+    $ NAMESPACE=$(kubectl get pods -l k8s-app=istio-cni-node --all-namespaces --output='jsonpath={.items[0].metadata.namespace}')
+    {{< /text >}}
+
+1. If `istio-cni` is currently installed in a namespace other than `kube-system`
+   (for example, `istio-system`), delete `istio-cni`:
+
+    {{< text bash >}}
+    $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=$NAMESPACE | kubectl delete -f -
+    {{< /text >}}
+
+1. Install or upgrade `istio-cni` in the `kube-system` namespace:
+
+    {{< text bash >}}
+    $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=kube-system | kubectl apply -f -
+    {{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="Helm upgrade" cookie-value="helmupgrade" >}}
+
+If you installed Istio CNI using [Helm and Tiller](/docs/setup/kubernetes/install/helm/#option-2-install-with-helm-and-tiller-via-helm-install),
+the preferred upgrade option is to let Helm take care of the upgrade.
+
+1. Check whether `istio-cni` is installed, and in which namespace:
+
+    {{< text bash >}}
+    $ helm status istio-cni
+    {{< /text >}}
+
+1. (Re-)install or upgrade `istio-cni` depending on the status:
+
+    * If `istio-cni` is not currently installed and you decide to install it:
+
+        {{< text bash >}}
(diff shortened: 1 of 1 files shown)


content/docs/setup/kubernetes/upgrade/steps/index.md
```diff
@@ -36,11 +36,86 @@ are supported. If you are on an older version, please upgrade to 1.1 first.
...
+{{< tab name="Kubernetes rolling update" cookie-value="k8supdate" >}}
+
+You can use Kubernetes’ rolling update mechanism to upgrade the Istio CNI components.
+This is suitable for cases where `kubectl apply` was used to deploy Istio CNI.
+
+1. To check whether `istio-cni` is installed, search for `istio-cni-node` pods
+   and in which namespace they are running (typically, `kube-system` or `istio-system`):
+
+    {{< text bash >}}
+    $ kubectl get pods -l k8s-app=istio-cni-node --all-namespaces
```
linsun: can you confirm in 1.1, the pod is labelled as `k8s-app=istio-cni-node`?  my env doesn't have CNI, and I know this is not typically how we label istio control plane components.

```
$ kubectl get pods -l app=pilot --all-namespaces       
NAMESPACE      NAME                           READY   STATUS    RESTARTS   AGE
istio-system   istio-pilot-5dd7b5f674-zrlvm   2/2     Running   0          25h
```

content/docs/setup/kubernetes/upgrade/steps/index.md
```diff
@@ -36,11 +36,86 @@ are supported. If you are on an older version, please upgrade to 1.1 first.
...
+
+    {{< text bash >}}
+    $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=$NAMESPACE | kubectl delete -f -
+    {{< /text >}}
+
+1. Install or upgrade `istio-cni` in the `kube-system` namespace:
+
+    {{< text bash >}}
+    $ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=kube-system | kubectl apply -f -
+    {{< /text >}}
```
linsun: Should we make a note that no new deployment or existing deployment recover/austoscale can work during this window?

linsun: or this is only impacting the node that istio-cni is being updated?  Should we use some rolling strategy?

content/docs/setup/kubernetes/upgrade/steps/index.md
```diff
@@ -36,11 +36,86 @@ are supported. If you are on an older version, please upgrade to 1.1 first.
 [Download the new Istio release](/docs/setup/kubernetes/#downloading-the-release)
 and change directory to the new release directory.
 
+### Istio CNI upgrade
+
+Choose one of the following **mutually exclusive** options to check whether
```
linsun: should we say if you are using CNI, then... we need to make sure it is noted as optional component thus this step is optional and only needed for some folks.

linsun: note it is also possible user was not using CNI in 1.1 but want to switch to use CNI in 1.2

content/docs/setup/kubernetes/upgrade/steps/index.md
rlenglet: The Helm chart has always used `k8s-app: istio-cni-node`, cf. its creation in https://github.com/istio/cni/commit/0f8609fa8c8647aa3234eb801e498f60129f7986 on 2019-10-04.

The only question mark I have is that the Helm chart was copied and modified from an older [.../install/manifests/istio-cni.yaml](https://github.com/istio/cni/blob/master/deployments/kubernetes/install/manifests/istio-cni.yaml), which uses `istio-node` instead of `istio-cni-node` as the label value and the name of the daemonset.

This one is referenced in istio/cni's `README.md`, but not in the Istio docs.

rlenglet: I'll just remove [.../install/manifests/istio-cni.yaml](https://github.com/istio/cni/blob/master/deployments/kubernetes/install/manifests/istio-cni.yaml) and update README.md to avoid any confusion.

content/docs/setup/kubernetes/upgrade/steps/index.md
rlenglet: > Should we make a note that no new deployment or existing deployment recover/austoscale can work during this window?

Well, the real issues are:
* There is at least one race condition, which I'm fixing in https://github.com/istio/cni/pull/118.
* New deployments *will* work, but they may silently not have any iptables redirect configured if they start while the CNI plugin is not installed, i.e. the sidecar will be silently bypassed in this case. And that will not be reported anywhere.

I've been thinking about recommending users to cordon all the nodes where the daemonset is installed, before upgrading. But that might be too much, and it might actually prevent upgrading the daemonset itself.

> or this is only impacting the node that istio-cni is being updated? Should we use some rolling strategy?

`istio-cni-node` being a daemonset, it needs to be upgraded on all nodes. The rolling strategy doesn't matter in this case. When the daemonset pod gets restarted on a given node during rollout, it impacts the creation of pods on that node during that restart. It doesn't matter whether we restart all the daemonset's pods at once or in sequence, since the impact is only local.

rlenglet: A better solution I've been thinking about would be to have the Istio CNI scan all pods on the node when it starts, and install or udpate the iptables chains if they are missing or outdated. This will take care of the case when pods are started while the CNI is uninstalled during upgrade.

rlenglet: There's nothing to do about this in this particular doc, since the last step is to restart all injected pods, after all Istio components have been upgraded, incl. the Istio CNI.

content/docs/setup/kubernetes/upgrade/steps/index.md
rlenglet: Done.

content/docs/setup/kubernetes/upgrade/steps/index.md
rlenglet: FYI https://github.com/istio/cni/pull/130

PR #4519, opened 2019-06-21
Mention istioctl authn tls-check only check conflict states between c…

diemtvu:
Fix https://github.com/istio/istio/issues/14937

### DIFF
--- a/content/docs/tasks/security/mutual-tls/index.md
+++ b/content/docs/tasks/security/mutual-tls/index.md
@@ -74,7 +74,7 @@ You can also provide the destination service to filter the status to that servic
 
 {{< tip >}}
 This tool only check the TLS setting consistency between destination rules and authentication policies. It does not take into account whether the
-workloads have sidecar or not (i.e, whether those policy/destination rule can be enforced). In other words, status `CONFLICT` doesn't alway mean traffic is broken.
+workloads have sidecar or not (i.e, whether those policy/destination rule can be enforced). In other words, status `CONFLICT` doesn't always mean traffic is broken.
 {{< /tip >}}
 
 The following commands identify the authentication policy for the `httpbin.default.svc.cluster.local` service and identify the destination rules for the service as seen from the same pod of the `sleep` app:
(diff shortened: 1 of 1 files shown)


frankbu:
This doesn't seem like a very good solution to https://github.com/istio/istio/issues/14937.
Why don't we figure out how to make istioctl take into account whether or not a service is istio-enabled or not, or at a minimum ignore control plane services that we know don't have a sidecar. Otherwise this seems like a terrible UX.

diemtvu:
The tool was initially designed for finding possible conflicted policies. Afaik, there are other efforts to have more robust debugging tools (issue https://github.com/istio/istio/issues/11091). This PR only tries to make the existing documentation clearer.

content/docs/tasks/security/mutual-tls/index.md
```diff
@@ -72,6 +72,11 @@ Please check [Istio identity](/docs/concepts/security/#istio-identity) for more
 Use the `istioctl` tool to check if the mutual TLS settings are in effect. The `istioctl` command needs the client's pod because the destination rule depends on the client's namespace.
 You can also provide the destination service to filter the status to that service only.
 
+{{< tip >}}
+This tool only check the TLS setting consistency between destination rules and authentication policies. It does not take into account whether the
```
frankbu:
```suggestion
This tool only checks the consistency of the static TLS settings between destination rules and authentication policies. It doesn't take into account whether or not the
```

content/docs/tasks/security/mutual-tls/index.md
```diff
@@ -72,6 +72,11 @@ Please check [Istio identity](/docs/concepts/security/#istio-identity) for more
...
+workloads have sidecar or not (i.e, whether those policy/destination rule can be enforced). In other words, status `CONFLICT` doesn't always mean traffic is broken.
```
frankbu:
```suggestion
corresponding workloads have sidecars or not. When they don't, the policy and destination rules are not enforced, so note that status `CONFLICT` doesn't always mean that traffic is broken.
```

PR #4525, opened 2019-06-21
Add new Traffic management concept as a single page

rcaballeromx:
This PR is a workaround until the following Hugo issue is handled.
 https://github.com/gohugoio/hugo/issues/6050


geeknoid:
That's a great idea.

PR #4526, opened 2019-06-21
fix path

quanjielin:
all crd in 1.2 build is moved to ```install/kubernetes/helm/istio-init/files```


quanjielin:
close this as https://github.com/istio/istio.io/pull/4527 fix same issue

PR #4558, opened 2019-06-26
Remove reference for name since it fails validation

brian-avery:
Name is still documented in destination rules for the port, but the field is not documented in portSelector. Attempts to use the field hit (https://github.com/istio/istio/blob/870877a1e3d01f6a2c285f67ed1f709a9f993095/pilot/pkg/model/validation.go#L2257) "Port.name HTTP is no longer supported for destination". 

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


frankbu:
This file is generated. Please update the original source in istio/api.
https://github.com/istio/istio.io/blob/7e7afd55e1c00bb65956526f3bab76027088ff6e/content/docs/reference/config/networking/v1alpha3/destination-rule/index.html#L2

brian-avery:
Thanks @frankbu

PR #4559, opened 2019-06-26
Added comment regarding merging to port names

brian-avery:

This was called out in the code (https://github.com/istio/istio/blob/master/pilot/pkg/model/gateway.go#L105), but never in docs. 

This is related to #7325 

[ x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/docs/setup/kubernetes/additional-setup/requirements/index.md
+++ b/content/docs/setup/kubernetes/additional-setup/requirements/index.md
@@ -34,6 +34,10 @@ cluster must satisfy the following requirements:
     plain TCP traffic unless the port [explicitly](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service)
     uses `Protocol: UDP` to signify a UDP port.
 
+    {{< warning >}}
+    To avoid conflicts, use different port names when multiple TLS or HTTPS servers use the same port in a gateway.
+    {{< /warning >}}
+
 - **Pod ports**: Pods must include an explicit list of the ports each
   container listens on. Use a `containerPort` configuration in the container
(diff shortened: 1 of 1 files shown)


content/docs/setup/kubernetes/additional-setup/requirements/index.md
```diff
@@ -34,6 +34,10 @@ cluster must satisfy the following requirements:
     plain TCP traffic unless the port [explicitly](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service)
     uses `Protocol: UDP` to signify a UDP port.
 
+    {{< warning >}}
```
frankbu: I'm not sure this warning is appropriate here because this section is talking about port names in a Kubernetes `Service` while your warning is about port names in a `Gateway` config.

Isn't this sufifcient? https://istio.io/docs/ops/traffic-management/deploy-guidelines/#configuring-multiple-tls-hosts-in-a-gateway

brian-avery: Thanks. I looked but didn't see this section before. 

Is there a way to highlight this better? I know of at least two separate reported issues where users have encountered issues while using the same port names for HTTPS. One on Istio Github, which is linked, and one on the maistra Jira, so I wonder if we can highlight this a bit more.

frankbu:
I agree we should try to come up with ways to better highlight the op guide entries. Thanks @brian-avery for opening #4560 to track this issue. Closing this PR since this warning is not in a suitable location.

PR #4639, opened 2019-07-05
Show the URL for the Mixer self-monitoring endpoint

jwendell:
So that the user does not have to guess.

### DIFF
--- a/content/docs/ops/telemetry/missing-metrics/index.md
+++ b/content/docs/ops/telemetry/missing-metrics/index.md
@@ -33,7 +33,7 @@ Mixer generates metrics to monitor its own behavior. The first step is to check 
     $ kubectl -n istio-system port-forward <istio-telemetry pod> 15014 &
     {{< /text >}}
 
-1. Verify successful report calls. On the Mixer self-monitoring endpoint, search for `grpc_io_server_completed_rpcs`. You should see something like:
+1. Verify successful report calls. On the [Mixer self-monitoring endpoint](http://localhost:15014/metrics), search for `grpc_io_server_completed_rpcs`. You should see something like:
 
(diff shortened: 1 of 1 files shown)


content/docs/ops/telemetry/missing-metrics/index.md
```diff
@@ -33,7 +33,7 @@ Mixer generates metrics to monitor its own behavior. The first step is to check
     $ kubectl -n istio-system port-forward <istio-telemetry pod> 15014 &
     {{< /text >}}
 
-1. Verify successful report calls. On the Mixer self-monitoring endpoint, search for `grpc_io_server_completed_rpcs`. You should see something like:
+1. Verify successful report calls. On the [Mixer self-monitoring endpoint](http://localhost:15014/metrics), search for `grpc_io_server_completed_rpcs`. You should see something like:
```
frankbu: This isn't really a clickable link, so I would suggest just printing the URL.
```suggestion
1. Verify successful report calls. On the Mixer self-monitoring endpoint (`http://localhost:15014/metrics`), search for `grpc_io_server_completed_rpcs`. You should see something like:
```

jwendell: If the user followed the instructions above (port-forward) then it is indeed clickable. I'm happy with any approach actually, as long as the user sees the correct endpoint.

frankbu: That's true, but often people just read the doc and don't necessarily follow the instructions step by step, in which case the link does not work. It just seems like a bad practice to have a link in a doc that isn't clickable at any time. I'm actually surprised that the linter isn't complaining about this one.

PR #4640, opened 2019-07-08
fix error in the jsonpath selector

hacktron95:
error in `jsonpath={.items..metadata.name}` the json selector produces the pod's name value to be incorrect
the fix `jsonpath='{.items[0].metadata.name}'` produces the right pod's name


hacktron95:
I signed it!

PR #4653, opened 2019-07-10
Dest rule evaluation subtlety

frankbu:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/docs/concepts/traffic-management/index.md
```diff
@@ -798,6 +799,32 @@ to:
 
 - Encode subsets in Server Name Indication (SNI) headers.
 
+### Service subsets
```
rcaballeromx: Please do not move this section. The concept has a deliberate content structure. Each section starts by describing the use cases or features before moving to the implementation information. That's why "Load balancing 3 subsets" came before "Service subsets". We can discuss the merits of this approach and make modifications accordingly but that would need to happen on a separate PR after we reach agreement of what would work for all sections.

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+the activation of subset-specific policies depends on route rule evaluation.
+
+The first step in Envoy's service routing process is to evaluate the route rules in
+a virtual service corresponding to the requested host, if there are any,
+to determine if a subset (that is, specific version) of the destination is being routed to.
+If so, the policies corresponding to the subset, if any, are then evaluated.
+Consequently, policies that are defined for specific subsets are only applied if
+the corresponding subset is explicitly routed to.
+
+For example, consider the following destination rule as the one and only configuration defined for the
```
rcaballeromx: This text and the text after the example should be combined to give readers the full context they need to understand the example. My suggestion does this but it is still unclear what applies the round-robin routing to the traffic.
```suggestion
When there are no routing rules in a corresponding virtual service and the following destination rule is the only configuration defined, the default round-robin routing applies.
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+*reviews* service, that is, there are no route rules in a corresponding virtual service definition:
```
rcaballeromx:
```suggestion
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+    labels:
+      version: v1
+    trafficPolicy:
+      connectionPool:
+        tcp:
+          maxConnections: 100
+{{< /text >}}
+
+Since there is no specific route rule defined for the *reviews*
+service, default round-robin routing behavior will apply, which will
```
rcaballeromx: Try to avoid giving a description of round-robin balancing here. It doesn't really help illustrate the message you are trying to convey. Maybe using {{< gloss >}}round-robin{{< /gloss >}} combined with an adequate glossary entry?

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+presumably call "v1" instances on occasion, maybe even always if "v1" is
+the only running version. Nevertheless, the above policy will never be
```
rcaballeromx: It's not clear what you mean with "done at a lower level". Readers don't know what is doing it based on the text.

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+invoked since the default routing is done at a lower level. The rule
+evaluation engine will be unaware of the final destination and therefore
```
rcaballeromx: Whose rule evaluation engine isn't aware? Why isn't it aware?

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+unable to match the subset policy to the request.
```
rcaballeromx: The phrase "subset policy" is not clear to readers at this point. Do you mean the subset definition in the destination rule, the subset in the routing rule, or the route established by the combination of both?

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+
+You can fix the above example in one of two ways. You can either move the
```
rcaballeromx: Use a bullet list to introduce the two possible solutions:
```suggestion
To ensure that the routing and destination rules work as intended, there are two possible solutions:
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+traffic policy up a level in the destination rule to make it apply to any subset:
```
rcaballeromx:
```suggestion
- Move the traffic policy in the destination rule up a level to make the traffic policy apply to any subset, for example:
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+    connectionPool:
+      tcp:
+        maxConnections: 100
+  subsets:
+  - name: v1
+    labels:
+      version: v1
+{{< /text >}}
+
+Or, better yet, define proper route rules for the service in the virtual service definition.
```
rcaballeromx:
```suggestion
Define proper route rules for the service using a virtual service.
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+For example, add a simple route rule for "reviews:v1":
```
rcaballeromx:
```suggestion
For our example, you must add a simple route rule for the `v1` subset of the `reviews` destination:
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+unable to match the subset policy to the request.
+
+You can fix the above example in one of two ways. You can either move the
+traffic policy up a level in the destination rule to make it apply to any subset:
+
+{{< text yaml >}}
+apiVersion: networking.istio.io/v1alpha3
+kind: DestinationRule
+metadata:
+  name: reviews
```
rcaballeromx: Does the name of the destination rule need to be the same as the service? The two being the same can confuse readers particularly when the service name is uses as the value of `host`.
```suggestion
  name: reviews-dr
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+  subsets:
+  - name: v1
+    labels:
+      version: v1
+{{< /text >}}
+
+Or, better yet, define proper route rules for the service in the virtual service definition.
+For example, add a simple route rule for "reviews:v1":
+
+{{< text yaml >}}
```
rcaballeromx: This virtual service is a perfect example of what I meant regarding the name of the destination rule. The same value, `reviews` is used in this example for the name of the virtual service, the service, and the host name for destination rule. It would be incredibly helpful if we used different names for each in our examples. We can make a separate recommendation for production environments where I think there might be some ease-of-use value in using the same name for everything. Please change the name of the virtual service, the destination rule, and the corresponding values of `host` and `hosts` for clarity.

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+  hosts:
+  - reviews
+  http:
+  - route:
+    - destination:
+        host: reviews
+        subset: v1
+{{< /text >}}
+
+Although the default Istio behavior conveniently sends traffic from any
```
rcaballeromx: This paragraph reads strangely, tone-wise. Certain words make the tone positive, like "conveniently" but others like "Although" shift the tone towards the negative. To make the tone consistently positive, I suggest:
```suggestion
The default Istio behavior conveniently sends traffic from any source
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+source to all versions of a destination service
```
rcaballeromx:
```suggestion
all versions of the destined service without you setting any rules.
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+without any rules being set, as soon as version discrimination is desired
```
rcaballeromx:
```suggestion
As soon as you need to differentiate between the versions of a service, you need to define routing rules.
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+rules are going to be needed.
+Therefore, setting a default rule for every service, right from the
```
rcaballeromx:
```suggestion
Due to this fact, we consider a best practice to set a default routing rule 
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+start, is generally considered a best practice in Istio.
```
rcaballeromx:
```suggestion
for every service from the start.
```

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+For example, consider the following destination rule as the one and only configuration defined for the
+*reviews* service, that is, there are no route rules in a corresponding virtual service definition:
+
+{{< text yaml >}}
+apiVersion: networking.istio.io/v1alpha3
+kind: DestinationRule
+metadata:
+  name: reviews
+spec:
+  host: reviews
```
rcaballeromx: Having the value of `name` and `host` in an example destination rule leads to a lot of confusion. Please see my other comments on this regard.

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
 propagation will take longer and there may be a lag time on the
 order of seconds.
 
+## Destination rule policy not activated
+
+Although destination rules are associated with a particular destination host,
+the activation of subset-specific policies depends on route rule evaluation.
```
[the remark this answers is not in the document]
frankbu: the reference doc says `Traffic policies to apply ...`, so I could say `the application of subset-specific ...` here, but I prefer `the activation of ...`.

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+
+The first step in Envoy's service routing process is to evaluate the route rules in
+a virtual service corresponding to the requested host, if there are any,
+to determine if a subset (that is, specific version) of the destination is being routed to.
```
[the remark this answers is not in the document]
frankbu: blue/green deployments are two version deployments of the same service, whether the label identifying them is `version:` or something else.

content/docs/ops/traffic-management/troubleshooting/index.md
```diff
@@ -62,6 +62,91 @@ to propagate to all the sidecars.  With large deployments the
...
+If so, the policies corresponding to the subset, if any, are then evaluated.
```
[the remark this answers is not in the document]
frankbu: it's Envoy, I'll clarify the wording

content/docs/ops/traffic-management/troubleshooting/index.md
frankbu: In this case, the name must be the same because the example is from our bookinfo sample, but in general, even if we want to use unique names, using short forms like `-dr`, `-svc`, etc., is generally not popular in any programming language naming conventions, so I wouldn't want to encourage it in Istio.

content/docs/ops/traffic-management/troubleshooting/index.md
frankbu: First of all, this example is from our Bookinfo sample application, so using a different naming convention is not an option. Second, we had a discussion about naming conventions for virtual service and destination rules quite a while ago and decided this is a nice simple pattern, so we used it in Bookinfo. Third, this is the operations guide, not the Istio 101 concepts, so readers are already using Istio and should not be so easily confused.

content/docs/ops/traffic-management/troubleshooting/index.md
frankbu: I think I'll remove this part. It's getting into more detail than needed.

content/docs/ops/traffic-management/troubleshooting/index.md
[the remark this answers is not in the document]
rcaballeromx: We can leave activation for now. I've added an issue to include it in the glossary. #4662

PR #4677, opened 2019-07-17
Document preference to using an NLB with gateways

sdake:
Gateways multicluster runs best with NLBs. Share this arcane bit of
information in the documentation.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/docs/setup/kubernetes/install/multicluster/gateways/index.md
```diff
@@ -45,6 +45,15 @@ on **each** Kubernetes cluster.
 
 ## Deploy the Istio control plane in each cluster {#deploy-istio}
 
+{{< tip >}}
```
frankbu: Wouldn't this be better in the prerequisites section, part of the `The IP address of the istio-ingressgateway service in each cluster must be accessible from every other cluster.` bullet? It should also be regular text. I've been trying to clean up some of the tips and warnings in several of our documents, since they are supposed to be used sparingly as mentioned in the last paragraph [here](https://preliminary.istio.io/about/contribute/creating-and-editing-pages/#callouts).

sdake: Using an NLB is not a prerequisite.  It is more like a recommendation, but I'll add the text there as a new bullet.

Cheers
-steve

content/docs/setup/kubernetes/install/multicluster/gateways/index.md
```diff
@@ -43,6 +43,15 @@ on **each** Kubernetes cluster.
   sample root CA certificate available in the Istio installation
   under the `samples/certs` directory.
 
+* This mode of multicluster works best with L4 network load balancers (NLB).
```
frankbu: This is better, but I was thinking that this could be combined with the earlier bullet about accessing the ingressgateway, since they are related. Would a single bullet like this work?
```
* The IP address of the `istio-ingressgateway` service in each cluster must be accessible
  from every other cluster, ideally using L4 network load balancers (NLB).
  Not all cloud providers support NLBs and some require special annotations to use them,
  so please consult your cloud provider’s documentation for enabling NLBs for
  service object type load balancers. When deploying on platforms without
  NLB support, it may be necessary to modify the healthchecks for the load
  balancer to register the ingress gateway.
```

sdake: good idea - will make it so.

Cheers
-steve

PR #4695, opened 2019-07-23
Fix kubectl used instead of istioctl

florianbegusch:
There is no support for the flag '-i' in vanilla kubectl, istioctl introduces this feature.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/blog/2018/soft-multitenancy/index.md
+++ b/content/blog/2018/soft-multitenancy/index.md
@@ -215,13 +215,13 @@ For example, the following command would be required to add a route rule to the 
 namespace:
 
 {{< text bash >}}
-$ kubectl –i istio-system1 apply -n ns-1 -f route_rule_v2.yaml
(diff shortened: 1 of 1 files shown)


florianbegusch:
> I signed it!

frankbu:
This blog looks to be outdated. RouteRule and DestinationPolicy resources don't exist anymore, they are replace by VirtualService and DestinationRule. Also, istioctl is no longer used to create configuration resources, kubectl is used now.

@john-a-joyce Should this blog be updated or marked as obsolete?

geeknoid:
Let's not update old blog entries, they are what  they are. Instead, we can mark it as deprecated and created a replacement post if need be.

PR #4699, opened 2019-07-23
capitalizes "pod"

ryanmcginnis:
[x] Docs

capitalizes instances of "pod" in this file. see the [Kubernetes style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-camel-case-for-api-objects) for guidance.

### DIFF
--- a/content/docs/concepts/what-is-istio/index.md
+++ b/content/docs/concepts/what-is-istio/index.md
@@ -68,7 +68,7 @@ manages authentication, authorization, and encryption of service communication a
 letting you enforce policies consistently across diverse protocols and runtimes -- all with little or no application changes.
 
 While Istio is platform independent, using it with Kubernetes (or infrastructure) network policies, the benefits are even greater, including the ability to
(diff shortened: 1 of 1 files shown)


geeknoid:
Regardless of what the k8s style guide says, I don't believe this is proper usage in this context.

When referring to a resource, then capitalization is appropriate. And in fact, the word should be put into backticks then as well. So `Pod` is appropriate, just like `VirtualService` is appropriate.

But when discussing the general concepts, then the words should not be capitalized. "When a pod crashes" or "when the virtual service  behaves as expected".

See: https://istio.io/about/contribute/style-guide/#use-consistent-capitalization

CC @rcaballeromx our professional tech writer.

sdake:
tend to agree with @rcaballeromx 's view.

sdake:
also - if you continue any work here, please submit to master - we have bots and whatnot to backport to 1.2 automagically.

geeknoid:
Thanks, but I'm going to close this one.

PR #4702, opened 2019-07-23
Point /charts/ at GCS directly instead of gcsweb.

Katharine:
gcsweb has not been considered a critical service and has no uptime or reliability guarantees, which can lead to surprising user-visible outages (e.g. istio/istio#15699).

This has the downside that users cannot see a file browser, but the upside of likely substantially better reliability, which may well be more important if people are expecting to be able to install from here.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/layouts/index.redir
+++ b/layouts/index.redir
@@ -19,7 +19,7 @@ https://istio.netlify.com/* https://istio.io/:splat 301!
 /  /zh   302  Language=zh
 
 # Redirect for the helm charts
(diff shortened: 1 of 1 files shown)


Katharine:
I don't think those commands do anything on this repo.

PR #4710, opened 2019-07-24
add a task about Kubernetes Services for egress

vadimeisenbergibm:
Replaces https://github.com/istio/istio.io/pull/3899

[ ] Docs
[ ] Installation
[X ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/docs/tasks/traffic-management/egress/egress-kubernetes-services/index.md
```diff
@@ -0,0 +1,247 @@
...
+1.  Disable mutual TLS for the calls to the external service:
+
+    {{< text bash >}}
+    $ kubectl apply -f - <<EOF
+    apiVersion: networking.istio.io/v1alpha3
+    kind: DestinationRule
+    metadata:
+      name: my-httpbin
+    spec:
+      host: my-httpbin.default.svc.cluster.local
```
frankbu: Maybe it would be better to include the explicit `tlsSettings` with mode set to `DISABLE`. The docs aren't really clear what the default is, and I think there's been some discussion about possibly honoring the `tlsSettings` of any other existing `DestinationRule` on the lookup path if the selected `DestinationRule` has other things (like subsets) but no explicit `tlsSettings` defined.

vadimeisenbergibm: Agree, good point! Handled in https://github.com/istio/istio.io/pull/4710/commits/5df8d209be2b4a05790bcee5cd23db071ccf3f9c

.spelling
```diff
@@ -384,6 +384,7 @@ pluggable
 png
 pprof
 pre-connected
+pre-Istio
```
rcaballeromx: Please reword to avoid adding this entry to the spelling exceptions.

content/docs/tasks/traffic-management/egress/egress-kubernetes-services/index.md
```diff
@@ -0,0 +1,253 @@
+---
+title: Kubernetes Services for Egress Traffic
+description: Shows how to configure Istio  Kubernetes External Services.
+keywords: [traffic-management,egress]
+weight: 60
+---
+
+Sometimes, for various reasons, for example in pre-Istio scenarios,
```
rcaballeromx: Please be clear and explicit about the scenarios where the k8s `ExternalName` or `Endpoints are used to access external services. As a reader, I need to be sure that this is the task I need before I begin going through the steps.

rcaballeromx:
I need to take a closer look at the content. I've added the the do-not-merge label for me to be able to complete the review.

content/docs/tasks/traffic-management/egress/egress-kubernetes-services/index.md
```diff
@@ -0,0 +1,253 @@
...
+Kubernetes [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname)
+services or Kubernetes services with
+[Endpoints](https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors) are used to
+access external services.
+This task shows that these Kubernetes mechanisms for accessing external services continue to work with Istio.
+The only configuration step you must perform is to disable Istio's
+[mutual TLS](/docs/concepts/security/#mutual-tls-authentication).
```
rcaballeromx: Requiring folks to disable mTLS to get common K8s features is not the right approach, IMO. @geeknoid WDYT?
At the very least, readers will want to know why that is the case.

vadimeisenbergibm: @rcaballeromx You are right, other TLS mode, for example SIMPLE, should be used instead of Istio mutual TLS.

frankbu: I agree with @rcaballeromx, you should add a couple of sentences to explain what is needed and why.

vadimeisenbergibm:
@rcaballeromx @frankbu I addressed your comments, thanks a lot for your feedback.

I thought about the motivation for this task and it became clear for me that it is not related to pre-Istio scenarios. Kubernetes external services and endpoints provide location transparency and can be used with or without Istio. So this task is even more relevant than I thought initially.

I also addressed the TLS mode. For HTTP cases, I recommend to use Egress TLS origination, explained in a separate task. For HTTPS cases, the traffic is already encrypted by the workload and Istio should not perform additional encryption (otherwise the external service will not be able to decrypt the traffic twice and will fail).

geeknoid:
Could you update this PR? The files needs to move into the content/en directory.

Thanks.

vadimeisenbergibm:
@geeknoid

>  Could you update this PR? The files needs to move into the content/en directory.

Done.

PR #4717, opened 2019-07-25
Rename multiple control planes multi cluster pattern

vadimeisenbergibm:
Following the discussion at https://docs.google.com/document/d/1ZwZ4KrE2pvcrdE1FVLke6ZMIAfC3y8_SYTzk6K2U9aQ/edit?ts=5d38f558#, remove multiple vs single control plane distinction, use dedicated vs shared control planes instead.

[ ] Docs
[X ] Installation
[X ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/docs/concepts/multicluster-deployments/index.md
+++ b/content/docs/concepts/multicluster-deployments/index.md
@@ -50,9 +50,9 @@ two clusters might share a control plane while a third has its own.
 Which approach to use and how to configure it depends on the requirements of the application
 and on the features and limitations of the underlying cloud deployment platform.
 
-### Multiple control plane topology
+### Dedicated control plane topology
 
-In a multiple control plane topology, each cluster has an identical Istio control plane
+In a dedicated control plane topology, each cluster has an identical Istio control plane
 installation and each control plane manages its own endpoints.
 Using Istio gateways, a common root Certificate Authority (CA), and service entries,
 you can configure a single logical service mesh that is composed from the participating clusters.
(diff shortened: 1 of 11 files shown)


content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -80,15 +80,15 @@ configuration. You configure service discovery of `foo.ns.global` by creating an
...
 
 To configure this type of multicluster topology, visit our
-[multiple control planes instructions](/docs/setup/kubernetes/install/multicluster/gateways/).
+[dedicated control planes instructions](/docs/setup/kubernetes/install/multicluster/gateways/).
 
 ### Shared control plane topology
 
-This multicluster configuration uses a single Istio control plane running on one of the clusters.
-The control plane's Pilot manages services on the local and remote clusters and configures the
+This multicluster configuration uses one or several Istio control planes running on some of the clusters.
```
frankbu: I agree that the current `a single` is causing problems, but we need to be careful about mixing up the HA discussion with the multicluster pattern. In the `or several` case, how is this different from a hybrid configuration of dedicated and shared clusters? If cluster A and B both have control planes, which one is managing the sidecars for a third cluster C, that is using the shared topology? I had previously thought that the fact that for HA there are multiple replicas of the single control plane running in multiple zones, for example, but I wouldn't complicate the description of the shared pattern by getting into that. If for HA, we need to have 2 separate control plane installations, then the question is how do we synchronize all the config/rules between them. Is that different from how synchronization would be done in the dedicated control plane case?

This is what I think is a big part of the disconnect in the discussion of MC with Costin. We need to understand how multiple control planes really mean and explain it in the doc.

PR #4725, opened 2019-07-26
Explain criteria for selecting multicluster topology

vadimeisenbergibm:


content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+However, for many reasons such as scale, redundancy, workload isolation, compliance, integration with legacy services,
+etc., most applications will eventually need to be distributed and have some of their services running elsewhere.
 
 Istio supports many possible topologies for distributing the services of an
 application beyond a single cluster, for example:
 
-* Services within the mesh can use [service entries](/docs/concepts/traffic-management/#service-entries)
-  to access standalone external services or to access services exposed by another loosely-coupled service mesh.
+* You can use [egress traffic control](/docs/tasks/traffic-management/egress/) to allow workloads inside the mesh
+  access standalone external services or to access services exposed by another loosely-coupled service mesh.
```
frankbu:
```suggestion
  to access standalone external services or to access services exposed by another loosely-coupled service mesh.
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
 * You can [expand the service mesh](/docs/setup/kubernetes/additional-setup/mesh-expansion/) to include
   services running on VMs or bare metal hosts.
 * You can combine the services from more than one cluster into a single composite service mesh.
+* You can create ad-hoc, loosely coupled _federation_ of multiple service meshes.
```
frankbu:
```suggestion
* You can create ad-hoc, loosely coupled [federation of multiple service meshes](#mesh-federation).
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+
+## Criteria for selecting the right topology
+
+When you plan your multi-mesh topology, ask yourself the following questions:
+
+**Uniform naming**: Will your meshes have _unoform naming_. In case of multiple Kubernetes clusters, _uniform naming_
+means that a service named `foo` in namespace `ns1` of cluster 1 has the same API and semantics as a service named `foo`
+in `ns1` of cluster 2.
+
+**Exposing behavior**: Will your combined meshes expose all their services one to another, or selective exposure is
```
frankbu:
```suggestion
**Exposing behavior**: Will your combined clusters expose all their services one to another, or selective exposure is
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+required? _Selective exposure_ means that the owner of mesh 1 may want to control which services they expose to the
```
frankbu:
```suggestion
required? _Selective exposure_ means that the owner of cluster 1 may want to control which services they expose to the
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+workloads in mesh 2. Selective exposure may be required for security or compliance reasons. Some security standards
```
frankbu:
```suggestion
workloads in cluster 2. Selective exposure may be required for security or compliance reasons. Some security standards
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+require exposure of services on the need-to-know basis.
+
+**Isolation for security or compliance**: Are you required to control all the traffic that enters your mesh **at the
```
frankbu:
```suggestion
**Isolation for security or compliance**: Are you required to control all the traffic that enters your cluster **at the
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+point of entrance**, for example at the Ingres Gateway?
+
+**Common trust**: Can you setup a common CA root for your meshes? Will you perform authentication by Istio
```
frankbu:
```suggestion
**Common trust**: Can you setup a common CA root for your clusters? Will you perform authentication by Istio
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+[mutual TLS](/docs/concepts/security/#mutual-tls-authentication) or will you use access tokens like
+[JWT](https://istio.io/docs/concepts/security/#origin-authentication)?
+
+**Operating multiple control planes**: Do you have infrastructure capabilities to manage multiple Istio control planes
+or do you prefer to manage a single control plane for multiple clusters?
+
+**Providing access to Kubernetes APIs**: Are you allowed to provide access to Kubernetes APIs of one cluster to
+other clusters?
+
+**Sending telemetry to external meshes**: Are you allowed to send telemetry data from your mesh to other meshes, from
```
frankbu:
```suggestion
**Sending telemetry to external meshes**: Are you allowed to send telemetry data from your cluster to other clusters, from
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+the security or compliance point of view? In particular, are your pods allowed to communicate with external meshes
```
frankbu:
```suggestion
the security or compliance point of view? In particular, are your pods allowed to communicate with external clusters
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,66 @@ a single controllable composite application.
...
+* You can use [egress traffic control](/docs/tasks/traffic-management/egress/) to allow workloads inside the mesh
+  access standalone external services or to access services exposed by another loosely-coupled service mesh.
 * You can [expand the service mesh](/docs/setup/kubernetes/additional-setup/mesh-expansion/) to include
   services running on VMs or bare metal hosts.
 * You can combine the services from more than one cluster into a single composite service mesh.
+* You can create ad-hoc, loosely coupled _federation_ of multiple service meshes.
+
+## Criteria for selecting the right topology
+
+When you plan your multi-mesh topology, ask yourself the following questions:
```
frankbu:
```suggestion
When you plan your multicluster topology, you first need to decide if you are going to combine the clusters into a
single multicluster mesh or create a federation of multiple meshes. Ask yourself the following questions:
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -128,3 +168,19 @@ different cluster, the remote cluster Gateway IP is used to connect to the servi
 
 To configure this type of multicluster topology, visit our
 [multi-network shared control plane instructions](/docs/setup/kubernetes/install/multicluster/shared-gateways/).
+
+## Mesh Federation
+
+Mesh federation is ad-hoc, loosely coupled composition of services running in multiple meshes.
```
frankbu:
```suggestion
Mesh federation is an ad-hoc, loosely coupled composition of services running in multiple meshes.
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -128,3 +169,19 @@ different cluster, the remote cluster Gateway IP is used to connect to the servi
...
+It exhibits the following characteristics
+(see [Criteria for selecting the right topology](#criteria-for-selecting-the-right-topology)):
+
+1. No uniform naming is assumed.
+1. Selective exposure: the mesh operators decide which services to expose to other meshes. By default, no service is
+exposed.
+1. Ingress Gateways control the incoming traffic from the coupled meshes.
+1. Common CA root between Citadel instances in the federated meshes is optional.
+
+You can achieve mesh federation by applying Istio [Ingress](/docs/tasks/traffic-management/ingress/) and
```
rcaballeromx:
```suggestion
To create a mesh federation, configure traffic control for [ingress](/docs/tasks/traffic-management/ingress/) and
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -128,3 +169,19 @@ different cluster, the remote cluster Gateway IP is used to connect to the servi
...
+[Egress](/docs/tasks/traffic-management/egress/) traffic control in your service meshes and by specifying the services
```
rcaballeromx:
```suggestion
[egress](/docs/tasks/traffic-management/egress/) traffic in your service meshes and specify which services
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -128,3 +169,19 @@ different cluster, the remote cluster Gateway IP is used to connect to the servi
...
+you expose in one mesh and consume in other meshes.
```
rcaballeromx:
```suggestion
to expose in one mesh and which to consume in the other meshes.
```

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,67 @@ a single controllable composite application.
 
 For applications up to a certain size, all of the microservices comprising the
 application can be running on a single orchestration platform (e.g., Kubernetes cluster).
-However, for many reasons such as scale, redundancy, compliance, etc., most applications
-will eventually need to be distributed and have some of their services running elsewhere.
+However, for many reasons such as scale, redundancy, workload isolation, compliance, integration with legacy services,
```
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx I am OK with the proposed change, however, please note that this is an advanced section. In general, people who work with Istio should understand what are distributed applications and microservices. Even more so, the people who want to distribute their applications among multiple clusters.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,67 @@ a single controllable composite application.
...
+etc., most applications will eventually need to be distributed and have some of their services running elsewhere.
 
 Istio supports many possible topologies for distributing the services of an
 application beyond a single cluster, for example:
 
-* Services within the mesh can use [service entries](/docs/concepts/traffic-management/#service-entries)
+* You can use [egress traffic control](/docs/tasks/traffic-management/egress/) to allow workloads inside the mesh
```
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx Service entries are configuration artifacts that you apply as part of egress traffic control. It is a low-level mechanics how to achieve egress traffic control. Even more so, for **secure** control of egress traffic, service entries are not enough - you have to use an egress gateway, a virtual service and destination rules.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,67 @@ a single controllable composite application.
...
+**Uniform naming**: Will your meshes have _uniform naming_. In case of multiple Kubernetes clusters, _uniform naming_
+means that a service named `foo` in namespace `ns1` of cluster 1 has the same API and semantics as a service named `foo`
+in `ns1` of cluster 2.
+
+**Exposing behavior**: Will your combined clusters expose all their services one to another, or selective exposure is
+required? _Selective exposure_ means that the owner of cluster 1 may want to control which services are exposed to the
+workloads in cluster 2. Selective exposure may be required for security or compliance reasons. Some security standards
+require exposure of services on the need-to-know basis.
+
+**Isolation for security or compliance**: Are you required to control all the traffic that enters your cluster **at the
```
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx This part is tricky. Consider this scenario: service A in cluster 1 calls service B in cluster 2. The policy of the organization is that only service A may call service B.

You can control which services may call service B and to enforce the policy in three ways:

1. Let service B check who calls it and enforce the policy. The ingress gateway just passes through the requests to services in cluster 2 without any checks.
1. Let an ingress gateway of cluster 2 check who calls a service in cluster 2 and enforce the policy.
1. Use a combination of 1 and 2, and perform double enforcement (the defense- in-depth principle).

Isolation means 2 or 3 - this is what the security standards require. 1 means regular policy enforcement, the same as within a single cluster.

Please note that this is an advanced concept, which is important for advanced users. If we leave it out, the advanced users will miss this security aspect.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,67 @@ a single controllable composite application.
...
+point of entrance**, for example at the Ingres Gateway?
+
+**Common trust**: Can you setup a common CA root for your clusters? Will you perform authentication by Istio
```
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx There are two questions:
1. Will you perform authentication by Istio mutual TLS or by access tokens?
1. If you will perform authentication by Istio mutual TLS, will you be able to share the certificate authority (CA) root for Citadels in different clusters? 

This question is advanced and requires understanding of PKI - how certificate authorities work and what is common CA root. See https://en.wikipedia.org/wiki/Root_certificate

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,67 @@ a single controllable composite application.
...
+[mutual TLS](/docs/concepts/security/#mutual-tls-authentication) or will you use access tokens like
+[JWT](https://istio.io/docs/concepts/security/#origin-authentication)?
+
+**Operating multiple control planes**: Do you have infrastructure capabilities to manage multiple Istio control planes
+or do you prefer to manage a single control plane for multiple clusters?
+
+**Providing access to Kubernetes APIs**: Are you allowed to provide access to Kubernetes APIs of one cluster to
```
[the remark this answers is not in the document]
vadimeisenbergibm: _ *Does a cluster in your deployment access the Kubernetes APIs of the other clusters?**_ this is not the question. Shared Control Plane Mesh in Istio requires access from the cluster with the control plane to Kubernetes APIs of other clusters. The question is: are you OK to provide such access? From the security, compliance, organization procedures point of view.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,67 @@ a single controllable composite application.
...
+other clusters?
+
+**Sending telemetry to external meshes**: Are you allowed to send telemetry data from your cluster to other clusters, from
```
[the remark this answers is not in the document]
vadimeisenbergibm: @frankbu proposed to stick to clusters instead of meshes. The question is not "does your cluster send telemetry data to another cluster?". The Shared Control Plane implementation of Istio requires this sending. The question is: "are you OK with that"? From security, compliance, procedures point of view.

The second question is are you OK that your pods access the other cluster directly, without going through some egress gateways in your clusters? In some environments, the outgoing traffic is allowed only through specific pods.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -40,7 +81,7 @@ To achieve this behavior, a single logical control plane needs to manage all ser
 however, the single logical control plane doesn't necessarily need to be a single physical
 Istio control plane. There are two possible deployment approaches:
 
-1. Multiple Istio control planes that have replicated service and routing configurations.
+1. Multiple dedicated Istio control planes that have replicated service and routing configurations.
```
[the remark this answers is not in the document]
vadimeisenbergibm: @frankbu and I introduced this term into this document, see Dedicated Control planes in this document, merged in preliminary.istio.io. See  https://preliminary.istio.io/docs/concepts/multicluster-deployments/#dedicated-control-plane-topology.

Dedicated control plane means: each cluster uses its own control plane. Each control plane manages only the cluster it runs in. This is the opposite of shared control plane: clusters can use control planes in other clusters. A control plane can manage multiple clusters.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,26 +12,67 @@ a single controllable composite application.
...
+
+**Sending telemetry to external meshes**: Are you allowed to send telemetry data from your cluster to other clusters, from
+the security or compliance point of view? In particular, are your pods allowed to communicate with external clusters
+directly?
+
+The following decision diagram instructs you on choosing the correct topology based on the answers to the questions
+above:
+
+{{< image width="80%"
+    link="./multicluster-patterns.png"
```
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx I agree regarding your remarks about the decision tree, however, from what I saw the convention is to use the same shape and color for `yes` and `no` arrows to show that both alternatives are equal. Sometimes, the selected decision is marked by bold arrows. See https://chunml.github.io/ChunML.github.io/tutorial/Decision-Tree/, the second and third decision trees.

content/docs/concepts/multicluster-deployments/index.md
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx I removed the link you provided for "distribute". Note that there are two kinds of distribution:
1. Partition one application into microservices inside a single cluster. This is the basic use of Istio.
2. Distributing the microservices from the previous partition among multiple clusters. This is the advanced use case, the document talks about this kind of distribution.

vadimeisenbergibm:
@rcaballeromx I have addressed your comments. Several points:

* Note that in selecting the right topology the question are not "what do you have today?" but "what do you want to achieve?" and "what are you able and allowed to configure?".
* To simplify the decisions, I split the decision tree and the related questions into two parts: single mesh or mesh federation, and selecting one of the three topologies of the single mesh.
* I simplified the questions.
* Regarding the different colors for yes and no arrows, the practice seems to show them in the same color and form, to show that these are equal choices.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,39 +12,94 @@ a single controllable composite application.
 
 For applications up to a certain size, all of the microservices comprising the
 application can be running on a single orchestration platform (e.g., Kubernetes cluster).
-However, for many reasons such as scale, redundancy, compliance, etc., most applications
-will eventually need to be distributed and have some of their services running elsewhere.
+Eventually, you will need to distribute most applications and run some or all of their services elsewhere due to one or more of the following reasons:
```
sdake: you may need to

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,39 +12,94 @@ a single controllable composite application.
...
+1. The same as the previous approach, but with services reachable through connected gateways.
 
-Even with these two approaches, there is more than one way to configure a multicluster service mesh.
+To select the correct topology, ask yourself the following questions:
+
+1. Do you require **a single control plane?**. Managing multiple Istio control planes
+  requires different infrastructure capabilities to those needed to manage a single
+  control plane. Be certain of your preference when planning your topology.
+
+1. Are you allowed by the security procedures in your organization to provide **access to Kubernetes APIs** of one
```
sdake: allowed/permitted throughout document

sdake:
added hold until agreement on https://github.com/istio/istio.io/pull/4725#pullrequestreview-267473137 is provided.

vadimeisenbergibm:
@sdake thank you for review. Everything is subject to change since the implementations themselves will change. I propose to merge it to reflect the current state to the users. Without this PR, it is just not clear to the users how to select the topology.

Waiting for more comments from @rcaballeromx.

content/docs/concepts/multicluster-deployments/index.md
vadimeisenbergibm: fixed in https://github.com/istio/istio.io/pull/4725/commits/773418edeb80c4779cfacc6df551cb540119560d

content/docs/concepts/multicluster-deployments/index.md
vadimeisenbergibm: Why is permitted better than allowed? I checked PCI DSS v3.2.1 compliance document, allow appears on 40 pages, permit on 10 pages.

Some examples:
Allow:
_1.3.4 Do not allow unauthorized outbound traffic from the cardholder data environment to the Internet._
_1.1.6.b Identify insecure services, protocols, and ports allowed; and verify that security features are documented for each service._

Permit:
_1.3.5 Permit only “established” connections into the network._
_User interfaces that do not permit access to unauthorized functions._

sdake:
> @sdake thank you for review. Everything is subject to change since the implementations themselves will change. I propose to merge it to reflect the current state to the users. Without this PR, it is just not clear to the users how to select the topology.
> 
> Waiting for more comments from @rcaballeromx.

I agree `how to select a topology` is a problem.  However, currently its a second-order problem.  Currently, the overall community is bogged down is basic terminology definitions that describe "What is a topology" that were introduced 50 years ago with the invention of ARPAnet.  In any regard, removing hold, although my hold was also strongly directed at the "glossary" section.

vadimeisenbergibm:
@sdake
>  my hold was also strongly directed at the "glossary" section.

I would be happy to hear better names. I based my proposals on https://docs.google.com/document/d/1g5vQXBuDZFDFBQStOjw4GNO5PJdBwgvCZwlz-VxoyYA:
    
> The goal of federation is to support relationships between organizations that don’t have **uniform naming**, identity and policy. There are now multiple independent operators of distinct meshes who collaborate in an ad-hoc manner to consume services they expose to each other. **It is not assumed that operators would want to expose the majority of their services and it’s perhaps more likely that they are only exposing a small subset.** If that assertion is true then operators are likely to prefer that exposure is a very explicit gesture done on a service by service basis. That exposure gesture also gives them precise control over who they expose a service to and how.

content/docs/concepts/multicluster-deployments/index.md
sdake: ya - makes sense.  thanks.

frankbu:
Note that this is only to be merged to preliminary.istio.io at this time. As Vadim mentioned, this is just trying to put something concrete in place, from which we will further iterate. It is absolutely expected to change before it's published to istio.io, esp. to align current documented patterns with the agreed terminology we are converging on in @ayj's doc.

content/docs/concepts/multicluster-deployments/index.md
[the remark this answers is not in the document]
frankbu: Dedicated == not shared. This is terminology that will potentially change before we release this to istio,io, but I agree it ultimately needs to be clearly defined if we stick with the term.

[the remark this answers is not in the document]
smawson: I don't love the "dedicated" terminology either, let's try to come up with something better if we can.

I also don't like the binary choice here (although there is a third choice which seems unrelated).

I like the way the federation section is written, which is basically "do you need multiple meshes". So let's think about how we can phrase this into a "how isolated do you need your control planes to be" question.

To me this is a continuum, at one end of the spectrum, every cluster has its own control plane. At the other end of the spectrum, there is a single *logical* control plane for the entire mesh. Each control plane itself may be replicated, which doesn't make sense at the "every cluster has its own control plane" end of the spectrum but makes sense everywhere else.

If we want to think about this like the federation question, either you need a "federation of control planes" or you have a single shared *logical* control plane.

So series of questions:
- Do you need more than one logical control plane? Here are questions to ask yourself to answer that.
- If you have fewer control planes than clusters, do you need them to be HA?

Also networking should be its own section, we shouldn't throw in the "oh do you need multi network too" here.

From top down then the questionnaire is:
- How many meshes do you need?
- How many *logical* control planes do you need?
  - For each *logical* control plane, how many replicas do you need?
- How many networks do you need?

In my mind the "dedicated" moniker is just at one extreme of the question of how many *logical* control planes. It doesn't deserve special status, just like we don't have special status for a federation of single cluster meshes (dedicated meshes?).

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,39 +12,94 @@ a single controllable composite application.
...
+   cluster to other clusters?
+
+1. Are you allowed by the security procedures in your organization to **send telemetry data from one cluster to other
```
smawson: Can we drop the "by the security procedures in your organization" part? Seems unnecessary.

Potential rewording:

In your organization, is one cluster allowed to **provide access to Kubernetes APIs** to other clusters?
In your organization, is one cluster allowed to **send telemetry data to other clusters**?

Seems less accusatory than the current language.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,39 +12,94 @@ a single controllable composite application.
...
+   clusters**?
+
+1. Are the services in your clusters **directly reachable** from all the workloads in all other the clusters?
```
smawson: This seems irrelevant? Network reachability is unrelated to how many *logical* control planes there should be. We're conflating networking and isolation.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,39 +12,94 @@ a single controllable composite application.
...
+
+## Single multicluster service mesh or mesh federation?
+
+As you plan your multicluster topology, you need to determine if you should combine the clusters into
+[a single multicluster mesh](#multicluster-service-mesh) or create [a federation of multiple meshes](#mesh-federation).
+To make that decision, ask yourself the following questions:
+
+1. Do you enforce **{{< gloss >}}uniform naming{{< /gloss >}}** in your clusters?
+
+1. Do you want your combined clusters to **expose** all their services to each other? Due to compliance or security
```
smawson: With the new Sidecar API it is possible to restrict service exposure using namespaces, rather than meshes. So this may no longer be a decision point for whether you need multiple meshes.

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,39 +12,94 @@ a single controllable composite application.
...
+   reasons, you might need {{< gloss >}}selective service exposure{{< /gloss >}}.
+
+1. Are you required to **control all the traffic that enters your cluster**?
```
smawson: Can we clarify this? I'm not entirely sure what control means here, maybe examples or some more specific wording?

content/docs/concepts/multicluster-deployments/index.md
```diff
@@ -12,39 +12,94 @@ a single controllable composite application.
...
+
+1. Will you be able to set **common CA root** for the Citadels running in your clusters?
```
smawson: This restriction will go away soon (1.3 or 1.4), we'll support multiple roots in a single mesh with a single trust bundle for the entire mesh.

The more specific question is whether the clusters trust each others' roots. We can reword once that stuff is in place though, but maybe put something in that we're working on lifting this restriction?

vadimeisenbergibm:
@smawson Regarding:
> How many meshes do you need?
How many logical control planes do you need?
For each logical control plane, how many replicas do you need?
How many networks do you need?

That's the point, the user cannot answer the question "how many meshes do you need? how many logical planes do you need". The user can answer questions about their organization, their security procedures, other requirements. "How many meshes/control planes do you need" is the answer, not the questions.

content/docs/concepts/multicluster-deployments/index.md
vadimeisenbergibm: This is a defense-in-depth/compliance requirement: you have to put a "firewall" at the entrance to your cluster and explicitly control who can call specific services in your cluster. Ingress gateway can function as such a firewall, in my opinion. This is in addition to the RBAC policies inside the cluster. In the current, single mesh implementations, the passthrough gateways allow all the traffic in. This cannot satisfy, for example, PCI DSS compliance, which requires network segmentation + a firewall between Cardholder Data Environment and other environments, in particular between Cardholder Data Environment and DMZ network.

content/docs/concepts/multicluster-deployments/index.md
vadimeisenbergibm: I do not think Sidecar API can serve as a security measure. The security requirement, I guess, is that all the incoming traffic to the cluster is denied by default. You have to specify explicitly which services you expose. With Sidecar API - everything is exposed by default.

content/docs/concepts/multicluster-deployments/index.md
vadimeisenbergibm: This question is to select between single network shared control plane topology and multiple networks shared control plane topology. It is a question of networking to choose between the two, isolation is the same in both cases.

vadimeisenbergibm:
@smawson The intention here is to capture the current situation AS IS, and then to iterate to improve it. HA is currently not documented at istio.io, so I propose not to mention it before it is demonstrated.

content/docs/concepts/multicluster-deployments/index.md
vadimeisenbergibm: @smawson Agreed, I can make the language less accusatory.

content/docs/concepts/multicluster-deployments/index.md
smawson: Exactly, we should separate the question of isolation from the question of networking. This section should be about how many logical control planes you need, and separately there should be a section on how many networks you have/need.

vadimeisenbergibm:
> In my mind the "dedicated" moniker is just at one extreme of the question of how many logical control planes.
@smawson Note that today it is a distinct pattern on istio.io. The goal of this PR is to capture the situation on istio.io and to explain it to the users in the simplest way possible. The user should be able to answer the question - what is good in this pattern https://istio.io/docs/setup/kubernetes/install/multicluster/gateways/ ? When should I use it?

content/docs/concepts/multicluster-deployments/index.md
smawson: That seems like the same question as the control all traffic that enters question then?

All I'm pointing out is that it isn't a requirement of a mesh that all services are exposed to each other, that is just the default (and a default we can support changing).

The primary issue is actually more about namespace vs. cluster isolation. If namespace isolation is good enough, a single mesh will work, if you need cluster isolation, you need multiple meshes.

content/docs/concepts/multicluster-deployments/index.md
vadimeisenbergibm: @smawson I propose to capture the current situation as it is on istio.io. Then to iterate on it, once new features/patterns are introduced/removed/simplified.

vadimeisenbergibm:
@smawson So which name is better than "dedicated"? I thought about "Nonshared". This document contains 17 pairs of names for the current single/multiple control plane topologies https://docs.google.com/document/d/13i_LgdcOeM6COX1uTN8wKU12sljhj5ODGCi_UNt5BrU/edit#heading=h.qex63c29z2to, please take a look at the end of the document.

content/docs/concepts/multicluster-deployments/index.md
smawson: OK, I think I'm a bit thrown off by the word cluster.

In my model if we think about groups of clusters that need to be isolated, each of those are a mesh, so this is about controlling all traffic that enters some subset of clusters, not necessarily each cluster. Maybe we can reword to be about groups and use an example of requiring firewalling between environments, and each environment should be its own mesh.

content/docs/concepts/multicluster-deployments/index.md
smawson: Fair enough, would still like some forward-looking verbiage since we're working on it and this isn't inherent in "what is a mesh".

vadimeisenbergibm:
@smawson 
> So the headers would be something like:
Single multicluster service mesh or mesh federation?
Single shared control plane or multiple control planes?
Single shared network or multiple networks?

I agree.

vadimeisenbergibm:
> In my model if we think about groups of clusters that need to be isolated, each of those are a mesh, so this is about controlling all traffic that enters some subset of clusters, not necessarily each cluster. Maybe we can reword to be about groups and use an example of requiring firewalling between environments, and each environment should be its own mesh.

@smawson I agree with this model. I started with environment=single cluster. Each isolated environment should be a separate logical mesh.

vadimeisenbergibm:
@smawson Regarding the CA roots, I see, so the question is whether the clusters can trust each other CA root, not necessarily to set a common CA root. Agreed.

nmittler:
Added a hold on this. Let's wait for @smawson to review next week.

content/docs/concepts/multicluster-deployments/index.md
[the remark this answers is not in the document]
frankbu: @smawson, "dedicated" means more than just one logical control plane per cluster. It's a fundamentally different question from "how many control planes". Consider the case of a 2 cluster multicluster mesh. If I want to have 2 logical control planes, I could choose to configure it with either 2 dedicated control planes or 2 shared control planes.

With 2 dedicated control planes, each cluster has its own control plane dedicated to the services running in that cluster. In the 2 shared control planes topology, each cluster has a local control plane but the applications in each cluster can also be controlled by the shared control plane in the other cluster.

Both approaches are HA in the sense that there are multiple clusters capable of running the app. I suppose the shared control plane topology has more HA in that if one of the control planes goes down, application services in the same cluster can still function using the remote control plane.

So, in addition to "how many control planes", I think there needs to be some kind of question about which clusters are sharing each control plane (0 -> dedicated). Maybe this is an "HA control plane", as opposed to "HA mesh", question?

vadimeisenbergibm:
Hi @ayj, sorry for not answering your questions before, I was busy with https://github.com/istio/istio.io/pull/4776, will answer your questions today.

 
> Isn't this somewhat analogous to k8s NetworkPolicy? One can use a single k8s cluster for multiple services and selectively expose namespaces/pods to one another with the NetworkPolicy API. Likewise, one can use a single Istio mesh and selectively expose namespaces/services with the Sidecar API.

Boundary protection, in particular, means the following:
- put an access control mechanism at the boundary (firewall, gateway, etc.)
- monitor the incoming/outgoing traffic at the boundary
- all the access control mechanisms must be _deny-all_ by default
- do not expose private IP addresses from the boundary
- do not let components from outside the boundary to impact security inside the boundary

Network policies can be considered as the access control mechanism at the boundary of a namespace. Implementing multiple networks within the same cluster. For example, the [Google Cloud PCI Data Security Standard compliance](https://cloud.google.com/solutions/pci-dss-compliance-in-gcp) document recommends:
>  Implement network policies in the GKE cluster when your in-scope projects contain different types of pods.

In some organisations, though, isolation by network policies within the same cluster considered to be insufficient. Even if the Network Policies can be considered sufficient for isolation, boundary protection would require that you use a separate control plane within the environment isolated by the Network policies. This is because it is not recommended to send private IP data outside of the protected boundary and to let an external system (external Istio control plane) to impact the security of the system inside the boundary.

In short, if in Kubernetes you enable Network policies, let's say on a namespace `mynamespace` and install a separate Istio control plane into that namespace, `mynamespace` could be considered as isolated in some organisations. Some other organisations with more strict security requirements would demand isolation by networks and clusters.

vadimeisenbergibm:
@ayj
> What if this was reworded to "Will you be able to establish common trust between across clusters. In the near term this means using a common CA root for Citadels in the mesh. In the future this may also include trusting multiple roots as part of a single trust bundle (insert link to revised Istio concept doc about identity and trust bundles)"

You are right, "common trust" is more appropriate.

vadimeisenbergibm:
@ayj 

> Does the same restriction exist if cross-cluster service discovery is exposed through MCP?

While MCP would alleviate the issue somewhat, the core problem is exposing private IP addresses outside of the protected boundary, which is not recommended by compliance standards. When you have an Istio control plane outside of your boundary, you have to send it your endpoints (by MCP) and also your telemetry data, and also it impacts the security of your boundary, which breaks isolation (according to PCI DSS).

vadimeisenbergibm:
@ayj 

> "Production systems should minimally deploy one control plane per isolated failure domain. Failure domain could be a region (lowest availability), zone, or cluster (highest availability) as defined by the user. A single global control plane is not recommended for larger multi-cluster production meshes."

> We could also be even more prescriptive and say something like "Deploy one control plane per-cluster. If this is cost/operational prohibitive deploy one control plane per region".

I agree with your statements. Having answered your questions, I will close this PR, since currently the rewrite of the whole multi-cluster/multi-mesh area is in place and discussed by a special environments sub-group.

PR #4759, opened 2019-08-05
Document how and why to use distroless variant

kramerul:
Please provide a description for what this PR is for.

istio will provide an additional set of docker images with 1.3. We added some notes on how to install these images.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[X] Docs
[X] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[X] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #4746, opened 2019-08-01
Add docs for updated Citadel namespace targeting mechanism

Monkeyanator:
[https://github.com/istio/istio/pull/15503](https://github.com/istio/istio/pull/15503) will change the way that Citadel decides which namespaces it should create `ServiceAccount` secrets for, and once merged, will need to be documented.

This PR documents the new `security.enableNamespacesByDefault` flag in the `Installation Options` page, and adds a section to the `Keys and Certificates` page surrounding this new selection mechanism, along with a few brief usage examples. Please let me know if this section would be better split up / moved elsewhere in the docs 

/cc @incfly 

[ ] Configuration Infrastructure
[X] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #4765, opened 2019-08-05
Move content into a new Troubleshooting section

adammil2000:
The biggest change was to identify and split the content under the Operations node into troubleshooting and conceptual content, as it was heavily intermingled.
Created a dedicated Troubleshooting node and reorganized the troubleshooting content there into sub-nodes based on the emergent subject matter areas.
Promoted the handful of remaining conceptual articles to the top level inside Operations, generally ordered top-down according to their popularity from our Analytics data. For example, “Health Checking of Istio Services” ranked near the top of all Operations content, so now it has a more prominent position in the Operations node.
Removed the nodes (and moved their contents accordingly) under which the Operations conceptual articles were categorized (e.g. Traffic Management, Security, etc.) because those nodes didn’t make sense anymore after the changes.
Combined some topics together as I identified complementary or overlapping content.
Moved an article to the Tasks section.
Fixed broken links in many places that were caused by moving so much content.
Moved “Introduction to Network Operations” to be the top conceptual Operations article as Operations doesn’t yet have any other content suitable to introduce the section.
Added many aliases to ensure that peoples’ bookmarks to old content locations still work.

Re-added subcategories to the Operations node based on feedback. Cleaned up titles for many troubleshooting articles to be more obvious. Moved some articles based on popularity indicated by Analytics.

Reintroduced the Traffic Management, Security, and Telemetry nodes after content was moved to Troubleshooting in order to illustrate the conceptual and procedural content gaps. Also fixed more broken links.

Fixed missing link

Implement phase 1 of the Operations docs reorganization by starting with a new Troubleshooting section and moving troubleshooting-specific content beneath it, then moving the remaining non-troubleshooting content into its respective locations.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


frankbu:
Is [this entry](https://github.com/istio/istio.io/pull/4759) in the right location, after your reorg?

adammil2000:
> Is [this entry](https://github.com/istio/istio.io/pull/4759) in the right location, after your reorg?

Yes it will be good.

Thanks!

frankbu:
Overall, I like the reorg, but I wonder if it would make sense to merge security-related entries into a single doc, similar to the "Network Problems" one. For example, how about creating a "Security Problems" doc with sections covering:

https://deploy-preview-4765--preliminary-istio.netlify.com/docs/ops/troubleshooting/end-user-auth/
https://deploy-preview-4765--preliminary-istio.netlify.com/docs/ops/troubleshooting/authorization-problems/
https://deploy-preview-4765--preliminary-istio.netlify.com/docs/ops/troubleshooting/keys-and-certs/
https://deploy-preview-4765--preliminary-istio.netlify.com/docs/ops/troubleshooting/mutual-tls/

Maybe some of the other entries should be merged somehow as well?

adammil2000:
> Overall, I like the reorg, but I wonder if it would make sense to merge security-related entries into a single doc, similar to the "Network Problems" one...

I can do that, seems like a good idea to me. 

But consolidating even further is tricky because the remaining nodes don't relate much to each other. Maybe just do a Security Problems node for now and see if other categories naturally form over time as we add more troubleshooting content there?

frankbu:
Btw, here's another PR that is updating an entry that you moved to troubleshooting: https://github.com/istio/istio.io/pull/4746

The added content, in this case, looks like it's concept material (not troubleshooting) so the material would probably need to be split in your new structure.

adammil2000:
Frank's suggestion is implemented in the latest commit. Thanks!

PR #4788, opened 2019-08-14
Multicluster mesh and mesh federation glossary terms

frankbu:
Let's try to crisply define some basic multi-Istio terms.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/docs/reference/glossary/mesh-federation.md
+++ b/content/docs/reference/glossary/mesh-federation.md
@@ -1,0 +1,7 @@
+---
+title: Mesh Federation
+---
+A loosely-coupled set of autonomous [service meshes](#service-mesh) collectively providing the services
+for an application. Two service meshes in a federation are isolated from each other and have different service
+definitions and independent administrative control. A mesh contributing to the federation may itself be a single
+or [multicluster mesh](multicluster-mesh).
--- a/content/docs/reference/glossary/multicluster-mesh.md
+++ b/content/docs/reference/glossary/multicluster-mesh.md
@@ -1,0 +1,6 @@
+---
+title: Multicluster Mesh
+---
+A [service mesh](#service-mesh) composed of services running within more than one underlying cluster.
+All clusters in a multicluster mesh are under the same administrative control and share the same service definitions.
+A service named `foo` in namespace `ns1` of cluster 1 represents the same service as `foo` in `ns1` of cluster 2.


content/en/docs/reference/glossary/mesh-federation.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Mesh Federation
+---
+A loosely-coupled set of autonomous [service meshes](#service-mesh) collectively providing the services
+for an application. Two service meshes in a federation are isolated from each other and have different service
```
nmittler: I think we may need to work on this definition a bit...

- what does `isolated` mean here?
- does mesh federation require independent administrative control? Or simply different domains of trust?
- it seems that we need to talk more directly about the two meshes selectively exposing services to each other for consumption. And to do that, they have to share identities and trust so that workloads across both meshes can mutually authenticate with each other.

Also, have we officially accepted the definition of `Mesh Federation`? I feel like there is still some internal debate, but I could be wrong.

content/en/docs/reference/glossary/multicluster-mesh.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Multicluster Mesh
```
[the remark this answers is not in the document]
frankbu: Interesting take, but it seems to me the term multicluster is thrown around and used so frequently that we need to concisely define what that is.

content/en/docs/reference/glossary/mesh-federation.md
frankbu: I'm trying to use this definition as a starting point to get something officially accepted :)

Maybe saying `isolated` is being too restrictive? Actually, I think mesh federation doesn't "require" independent administrative control or even different domains of trust. Instead I think it should just say "clusters can have different administrative control / domains of trust".

I'm just trying to give the high level (what is it) definition here. Things like how the federation is achieved (expose-consume) would be described in the concepts (not the glossary). It's a baby step.

content/en/docs/reference/glossary/multicluster-mesh.md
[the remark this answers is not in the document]
nmittler: I think the term `Multicluster` really only applies to a single mesh which spans multiple clusters. So `Multicluster` and `Multicluster Mesh` are the same thing.

If you have two or more meshes, the cluster topology is irrelevant. That's `Mesh Federation` (or something mesh-focused, rather than cluster-focused).

content/en/docs/reference/glossary/mesh-federation.md
nmittler: Understood ... really what you're grappling with here is that age old question: "what is a mesh?". I had to go through that same exercise when writing the concept docs.

I think we should drop the concept of administrative domain completely and focus on service naming and trust domain, and leave it at that. BTW, if we introduce the term "trust domain", we'll need to define that too :).

content/en/docs/reference/glossary/multicluster-mesh.md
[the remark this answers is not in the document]
frankbu: You've hit the nail right on the head. I'm trying to differentiate between 1) multiple clusters that are contributing to a single mesh vs. 2) multiple clusters that are each configured as single meshes and then federated. Both of these involve multiple clusters, so I think we need to be clear that when we use the term multicluster we mean case 1) and I think that consistently using the explicit term "multicluster mesh" in the docs is best.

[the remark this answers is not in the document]
nmittler: Ok that's fair. But it still doesn't solve the problem of folks using the term `multicluster`. I think if we define this, we should also define `multicluster` and reference `multicluster mesh`. This makes it clear that WRT Istio, they're the same thing.

content/en/docs/reference/glossary/mesh-federation.md
shalako: I find the term "administrative domain" as a synonym for mesh to be quite useful because it approaches the definition of mesh boundary from a users perspective; mesh federation is the mechanism by which services in meshes under the administration of distinct entities can interoperate. 

What CA cert the administrator chooses to use for a mesh (the trust domain) is just one of many choices an administrator would make for a mesh. I do not see these terms as synonymous.

content/en/docs/reference/glossary/multicluster-mesh.md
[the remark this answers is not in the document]
shalako: +1 to documenting a definition for this term; it may or may not help minimize the times we have to ask "by multi-cluster, do you mean one mesh or multiple?", but it can't hurt.

content/en/docs/reference/glossary/mesh-federation.md
nmittler: @shalako

I'm not against adding the term "administrative domain" so long as we have a clear, concise definition for it. In fact we probably need it.

Assuming its inclusion, a mesh would consist of:

- exactly 1 administrative domain
- exactly 1 trust domain
- common naming

The administrative domain is a little different since it's more of an abstract concept and doesn't have any concrete representation per se. However, I think it's true to say that the boundary of a mesh represents a single administrative domain. In that sense, an administrative domain's concrete representation is the mesh itself.

@smawson @louiscryan thoughts?

frankbu: I tried to explain the single administrative domain concept in application terms. Not sure it was clear enough but I was trying say that it essentially combines a set of services into a composite application under a single control. Take a look at the first paragraph [here](https://istio.io/docs/concepts/multicluster-deployments/). Does that seem reasonable?

nmittler: @frankbu A few things...

- We probably want to steer away from the term `microservices` and instead just use `services`, which is something we have a definition for.
- I don't think it's necessarily true to say that a mesh is a single composite application. It could be multiple. If a single mesh is used by a company, for example, they may offer multiple composite application within the same mesh, using namespace isolation between teams.

shalako:
Agreed, a single mesh may provide value for many applications. In addition, if the mesh is offered as a managed multi tenant service, it may provide value for the applications of multiple tenants or organizations.

content/en/docs/reference/glossary/mesh-federation.md
frankbu: Good point about the single application wording. Combining services into logical applications is the real value of service mesh, even if it's not limited to just one. I still wonder when there is more than one logical application in the same mesh, if it's still sort of one big application from the administrators point of view, even if the administration of different tenants, for example, are delegated to others. Any idea what would be a better word to use to describe this centrally administered set of one or more logical applications?

nmittler: I think there are a couple distinct administration roles here:

- mesh administrator: responsible for the installation and management of the control plane throughout the mesh.
- tenant administrator: responsible for the deployment of services to a single namespace within the mesh.

These may be the same person/group or they may not. In large orgs, they will likely be different.

It *may* be true that each tenant deploys a single composite application. I guess I'm not sure I see the need for calling out (and defining) composite applications at all.

content/en/docs/reference/glossary/mesh-federation.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Mesh Federation
+---
+A loosely-coupled set of autonomous [service meshes](#service-mesh) collectively providing the services
```
smawson: An application probably cannot span multiple meshes (for security and isolation reasons), and that's definitely not the key principle behind a federation of meshes.

Here's a possible replacement first sentence:

A loosely-coupled set of [service meshes](#service-mesh) that supports inter-mesh communication.

BTW I saw our current doc says something about a mesh being a way to create a composite application, that's incorrect as well, hopefully we can fix that as part of updating the docs.

content/en/docs/reference/glossary/mesh-federation.md
```diff
@@ -0,0 +1,7 @@
...
+for an application. Two service meshes in a federation are isolated from each other and have different service
+definitions and independent administrative control. A mesh contributing to the federation may itself be a single
```
smawson: Drop the sentence about single or multicluster mesh.

content/en/docs/reference/glossary/multicluster-mesh.md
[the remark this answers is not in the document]
smawson: Maybe we should add a glossary entry for multicluster by itself, and not for "multicluster mesh"?

Then under that we can say that multiple clusters can be part of a single mesh, or they can be part of multiple meshes and communicate using federation.

content/en/docs/reference/glossary/multicluster-mesh.md
```diff
@@ -0,0 +1,6 @@
...
+---
+A [service mesh](#service-mesh) composed of services running within more than one underlying cluster.
+All clusters in a multicluster mesh are under the same administrative control and share the same service definitions.
```
smawson: That's the definition of a service mesh, and should go there.

content/en/docs/reference/glossary/multicluster-mesh.md
```diff
@@ -0,0 +1,6 @@
...
+A service named `foo` in namespace `ns1` of cluster 1 represents the same service as `foo` in `ns1` of cluster 2.
```
smawson: Should be part of the "service mesh" glossary entry also.

content/en/docs/reference/glossary/mesh-federation.md
frankbu: I tend to agree with you. The only thing I'm still struggling with is whether, and how, we should associate applications with the general term service mesh. It seems to me that the main reason service meshes have come about is to tie back together the services that implement an application after moving from a monolith to microservices-based architecture. That's the purpose of a service mesh that I believe most people understand, so it seems that we should somehow be mentioning applications in the definition of service mesh, even if they aren't necessarily 1-1.

content/en/docs/reference/glossary/mesh-federation.md
frankbu: I would say Vadim's example is exactly that, an application that requires some of its services (e.g., the payment card processing) to be isolated from the rest of the application. You could argue they are separate applications, but I would say these are two components of a single application, separated for regulatory compliance reasons.

Yes, that doc is to be replaced by Nathan and Rigs' Istio Topology PR, which is supposed to be ready for review any time now. So here we just need to agree on a correct definitions for the glossary terms.

content/en/docs/reference/glossary/multicluster-mesh.md
frankbu: We could have a glossary entry for "Multicluster" only which just means more than one cluster running Istio, and explain that it can be modeled as a single mesh or multiple meshes that are federated. But, if we also remove the entry for "service mesh", we have no concrete term to use when talking about the single mesh composed from multiple clusters topology. Nathan and I have been leaning in the opposite direction, i.e., define the term "Multicluster" as the single mesh case only. Mesh Federation is then defined as the multiple mesh case.

Maybe defining all three terms in the glossary would be best: Multicluster Mesh (single mesh), "Mesh Federation" (multiple mesh), and just "Multicluster" (defined as can be either a "Multicluster Mesh" or a "Mesh Federation".

content/en/docs/reference/glossary/multicluster-mesh.md
frankbu: This is specifically talking explaining it in terms of two clusters, cluster 1 and cluster 2. Service mesh may or may not have more than one cluster.

content/en/docs/reference/glossary/multicluster-mesh.md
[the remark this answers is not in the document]
frankbu: We could have a glossary entry for "Multicluster" only which just means more than one cluster running Istio, and explain that it can be modeled as a single mesh or multiple meshes that are federated. But, if we also remove the entry for "service mesh", we have no concrete term to use when talking about the single mesh composed from multiple clusters topology. Nathan and I have been leaning in the opposite direction, i.e., define the term "Multicluster" as the single mesh case only. Mesh Federation is then defined as the multiple mesh case.

Maybe defining all three terms in the glossary would be best: Multicluster Mesh (single mesh), "Mesh Federation" (multiple mesh), and just "Multicluster" (defined as can be either a "Multicluster Mesh" or a "Mesh Federation".

content/en/docs/reference/glossary/mesh-federation.md
rcaballeromx: In an attempt to help craft these definitions, I would like to ask that we write them following the following established pattern:
Start with the term (word or phrase) to be defined.
Followed by the class of object or concept to which the term belongs.
Finish with the differentiating characteristics that distinguish it from all others of its class.

https://owl.purdue.edu/owl/general_writing/common_writing_assignments/definitions.html

Applying the pattern above to the proposed definition I come up with the following suggestion:

```
A mesh federation is a set of autonomous meshes that are loosely-coupled to collectively provide the services of an application. An autonomous service mesh is a mesh that fulfills the following requirements:

- Isolated from other meshes
- Has different service definitions than other meshes
- Has an independent administrative control
```

Missing from my suggestion is the explanation for `loosely-coupled` and it should be added IMO.

content/en/docs/reference/glossary/multicluster-mesh.md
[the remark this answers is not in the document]
rcaballeromx: PTAL at PR #4813 
IMO, the content there could help move this conversation forward.

PR #4799, opened 2019-08-15
Add a lint rule to prevent us forgetting to update the site version.

geeknoid:


duderino:
Can this be restricted to run only on the release branch?

Though I guess it'd just work on master because that would be a whole minor version ahead

Issue #4760, opened 2019-08-05
Misleading TLS example Policy name


esnible:
The [lockdown to mutual TLS](https://preliminary.istio.io/docs/tasks/security/mtls-migration/#lock-down-to-mutual-tls-optional) instructions include an example:

```
apiVersion: "authentication.istio.io/v1alpha1"
kind: "Policy"
metadata:
  name: "example-httpbin-permissive"
  namespace: foo
spec:
  targets:
  - name: httpbin
  peers:
  - mtls:
      mode: STRICT
```

The name of this Policy, `example-httpbin-permissive`, is extremely misleading.  Let's call it `example-httpbin-strict`.

PR #4800, opened 2019-08-15
Rename example YAML manifests to match intent

esnible:
Resolves the documentation component of https://github.com/istio/istio.io/issues/4760

Also related: https://github.com/istio/istio/blob/master/pkg/test/istio.io/tasks/security/mtls-migration/httpbin-foo-mtls-only.sh .  This file should also be changed.


esnible:
@frankbu What is needed to approve this PR?

PR #4804, opened 2019-08-16
Get istio.io working with BUILD_WITH_CONTAINER=1

geeknoid:
The default for this repo is to build with the container.


Makefile
```diff
@@ -19,31 +19,69 @@
...
+
 RUN =
 
-# Set the enviornment variable BUILD_WITH_CONTAINER to use a container
+# Set the environment variable BUILD_WITH_CONTAINER to use a container
 # to build the repo. The only dependencies in this mode are to have make and
 # docker. If you'd rather build with a local tool chain instead, you'll need to
 # figure out all the tools you need in your environment to make that work.
+
+ifeq ($(JOB_NAME),istio.io-presubmit)
```
sdake: this seems a little suspect.  Is this just temporary?

Makefile
```diff
@@ -19,31 +19,69 @@
...
 	-v /etc/passwd:/etc/passwd:ro \
 	-v $(readlink /etc/localtime):/etc/localtime:ro \
+	$(CONTAINER_OPTIONS) \
 	--mount type=bind,source="$(PWD)",destination="/work" \
 	--mount type=volume,source=istio-go-mod,destination="/go/pkg/mod" \
 	--mount type=volume,source=istio-go-cache,destination="/gocache" \
 	--mount type=bind,source="$(GOBIN_SOURCE)",destination="/go/out/bin" \
 	-w /work $(IMG)
+else
+export GOBIN ?= ./out/bin
```
sdake: I think GOBIN needs to be sanitized in common - this is an area for further refinement in common-files.

Makefile
geeknoid: Yes, until my PR in test-infra lands.

PR #4806, opened 2019-08-16
Make sure we spell check all .md files, we have been skipping some.

geeknoid:
This temporarily disables spell checking over the Chinese content, since making this change triggered >125 errors to surface.


geeknoid:
Unrelated link timeout error, so pushing...

PR #4808, opened 2019-08-16
Avoid redundant module downloads when running nodejs stuff.

geeknoid:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


Makefile.core.mk
```diff
@@ -29,21 +29,23 @@ lint: clean_public gen_nominify
...
+	@npm install --production --global \
+	    sass@v1.22.10 \
+	    typescript@v3.5.3 \
+	    svgstore-cli@v1.3.1 \
+		@babel/core@v7.5.5 \
+		@babel/cli@v7.5.5 \
+		@babel/preset-env@v7.5.5
+	@npm install --production --save-dev \
+		babel-preset-minify@v0.5.1
+	@npm install --save-dev \
```
sdake: this is inconsistent with the container implementation(the --save-dev which is probably needed according to the documentation) (and I htink also the version which IIRC is 7.5.0 for polyfill.)

scripts/build_site.sh
```diff
@@ -3,10 +3,10 @@ set -e
...
-npx sass src/sass/_all.scss all.css -s compressed --no-source-map
+sass src/sass/_all.scss all.css -s compressed --no-source-map
 mv all.css* generated/css
-npx tsc
+tsc
 
-npx babel-cli --source-maps --minified --no-comments --presets minify tmp/js/constants.js tmp/js/utils.js tmp/js/kbdnav.js tmp/js/themes.js tmp/js/menu.js tmp/js/header.js tmp/js/sidebar.js tmp/js/tabset.js tmp/js/prism.js tmp/js/codeBlocks.js tmp/js/links.js tmp/js/scroll.js tmp/js/overlays.js tmp/js/lang.js tmp/js/callToAction.js --out-file generated/js/all.min.js
-npx babel-cli --source-maps --minified --no-comments --presets minify tmp/js/themes_init.js --out-file generated/js/themes_init.min.js
-npx svgstore -o generated/img/icons.svg src/icons/**/*.svg
+babel --source-maps --minified --no-comments --presets minify tmp/js/constants.js tmp/js/utils.js tmp/js/kbdnav.js tmp/js/themes.js tmp/js/menu.js tmp/js/header.js tmp/js/sidebar.js tmp/js/tabset.js tmp/js/prism.js tmp/js/codeBlocks.js tmp/js/links.js tmp/js/scroll.js tmp/js/overlays.js tmp/js/lang.js tmp/js/callToAction.js --out-file generated/js/all.min.js
```
sdake: surprised this is in the path?

Makefile.core.mk
geeknoid: I don't see the inconsistencies. See https://github.com/istio/tools/blob/8a495d7e5de49ee394a33970492ca876a05ade77/docker/build-tools/Dockerfile#L141

Am I missing something?

geeknoid: BTW, When you install the polyfill stuff, it prints a message saying it is deprecated. But when I remove it, nothing works...

scripts/build_site.sh
geeknoid: Switching the install location to be --global instead of --save-dev is what makes this work. It installs a .js file in /usr/local/bin for the different tools. Using npx ended up being redundant and was causing all the stuff to be redownloaded upon use. Now, it just works :-)

Makefile.core.mk
sdake: right - I reviewed quickly and got version numbers mixed up in my intermediate-term memory (which is unreliable;) . I saw that deprecation problem with polyfill scroll by.  I'll see if we can sort out the replacement.  NODEJS FTW.

PR #4812, opened 2019-08-20
Add glossary entries for needed terms.

rcaballeromx:
The terms involved are:

    - Cluster
    - Identity
    - Trust domain

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/reference/glossary/cluster.md
+++ b/content/en/docs/reference/glossary/cluster.md
@@ -1,11 +1,7 @@
 ---
 title: Cluster
 ---
-A set of machines, called nodes, that run containerized applications managed by
-Kubernetes. A cluster has at least one worker node and at least one master
-node. The worker node(s) host the pods that are the components of the
-application. The master node(s) manages the worker nodes and the pods in the
-cluster. Multiple master nodes are used to provide a cluster with failover and
-high availability.
+A cluster is set of compute nodes within physical proximity that run containerized
+applications.
 
-[Source](https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-cluster)
+Typically, the compute nodes comprising a cluster can reach each other directly.
--- a/content/en/docs/reference/glossary/cluster.md
+++ b/content/en/docs/reference/glossary/cluster.md
@@ -1,0 +1,11 @@
+---
+title: Cluster
+---
+A set of machines, called nodes, that run containerized applications managed by
+Kubernetes. A cluster has at least one worker node and at least one master
+node. The worker node(s) host the pods that are the components of the
+application. The master node(s) manages the worker nodes and the pods in the
+cluster. Multiple master nodes are used to provide a cluster with failover and
+high availability.
+
+[Source](https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-cluster)
--- a/content/en/docs/reference/glossary/identity.md
+++ b/content/en/docs/reference/glossary/identity.md
@@ -2,7 +2,7 @@
 title: Identity
 ---
 
-An identity is a unique name backed by a key material, for
+An identity is a unique name backed by key material, for
 example:
 
 - [X.509 certificate](https://en.wikipedia.org/wiki/X.509)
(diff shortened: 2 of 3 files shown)


content/en/docs/reference/glossary/identity.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Identity
+---
+
+An identity is a unique name backed by a key material, for
```
geeknoid:
```suggestion
An identity is a unique name backed by key material, for
```

content/en/docs/reference/glossary/cluster.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Cluster
```
nmittler: I'm assuming you borrowed liberally from Kubernetes here. I don't think we should use their definition. We need to be more generic ... Istio is not k8s-specific, and neither is the concept of a cluster.

content/en/docs/reference/glossary/identity.md
```diff
@@ -0,0 +1,18 @@
...
+example:
+
+- [X.509 certificate](https://en.wikipedia.org/wiki/X.509)
+- [JWT](https://en.wikipedia.org/wiki/JSON_Web_Token)
+
+Each identity Istio assigns is understood throughout the mesh and you can use it
+to enable mutual authentication and enforce policies.
+
+An identity contains a substring identifying the mesh that created it. For
```
nmittler: You're duplicating the definition of trust domain here.  We should probably remove.

content/en/docs/reference/glossary/trust-domain.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Trust Domain
+---
+
+An identity contains a substring identifying the mesh that created it. For
```
nmittler: should you have identity be a link?

content/en/docs/reference/glossary/identity.md
frankbu: Or replace it with a link to the Trust Domain entry.

content/en/docs/reference/glossary/cluster.md
frankbu: I agree, this needs to be a more generic definition of cluster, not the K8s one.

rcaballeromx: I'm replacing the definitions with the following platform agnostic version: 
```
A set of compute nodes within physical proximity that run containerized
applications.

Typically, the compute nodes comprising a cluster can reach each other directly.
```
WDYT?

nmittler: I think that's a lot better ... let's switch to this and let folks throw darts at it.

content/en/docs/reference/glossary/trust-domain.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Trust Domain
+---
+
+An [identity](/docs/reference/glossary/#identity) contains a substring
```
nmittler: We should probably lead with the definition of trust domain.  Maybe something like:

```
A trust domain is a globally unique name used to create identities. Every Istio mesh has exclusive ownership of a single trust domain, which is used to create all identities within the mesh. 
```

I think that should be sufficient.  WDYT?

frankbu: @nmittler Yes, I like your wording but would still add a link:
```
A trust domain is a globally unique name used to create
[identities](/docs/reference/glossary/#identity). Every Istio
mesh has exclusive ownership of a single trust domain, which is used to create
all identities within the mesh.
```

rcaballeromx: Maybe a simplified version:

```
A trust domain is a globally unique name that Istio uses to create all identities within a mesh. Every mesh has an exclusive trust domain. 
```
WDYT

rcaballeromx: And yes definitely with a link.

nmittler: Hmm ... yours is more brief, for sure. But I feel it loses the explicit statement that all identities in the mesh are created with the same trust domain. Yes it's implied, but you have to parse it in your head a little to get there.

nmittler: Actually, looking at it again, I think your definition is fine. 

@frankbu WDYT?

frankbu: Yep, the simpler version looks fine to me as long as it has the link from "identities".

nmittler:
@smawson @myidpt PTAL

rcaballeromx:
I'm closing this PR to and including the entry glossaries in #4813

PR #4813, opened 2019-08-20
Add Istio Deployment Models concept.

rcaballeromx:

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


rcaballeromx:
The build fails due to the missing glossary entries.
See https://github.com/istio/istio.io/pull/4812

content/en/blog/2019/announcing-1.1/index.md
```diff
@@ -36,15 +36,15 @@ We’ve done work around namespace isolation as well. This lets you use
 Kubernetes namespaces to enforce boundaries of control, and ensures that your
 teams cannot interfere with each other.
 
-We have also improved the [multicluster capabilities and usability](/docs/concepts/multicluster-deployments/).
+We have also improved the [multicluster capabilities and usability]/docs/concepts/deployment-models/).
```
geeknoid: Missing a (

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+    link="./single-cluster.svg"
+    alt="A service mesh with a single cluster"
+    title="Single cluster"
+    caption="A service mesh with a single cluster"
+    >}}
+
+Single cluster deployments offer simplicity, but lack fault isolation and
+failover. If you need higher availability, you should use multiple clusters.
+
+### Multiple clusters
```
frankbu: Where would Kubernetes cluster federation fit into the picture? A kube-specific multi-cluster platform? Do we need to mention this pattern?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+Istio supports two mesh models:
+
+- Single mesh
+- Multiple meshes, also known as **multi-mesh**
+
+### Single mesh
+
+The simplest Istio deployment is a single mesh. Within a mesh, service names are
+unique. For example, only one service can have the name `foo/mysvc`.
```
frankbu: `foo/mysvc` is not a valid service name in Istio. Service names don't have `/` in the syntax. I think you're trying to say that only one service can be have the name `mysvc` in namespace `foo`.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+Multiple meshes afford the following capabilities beyond that of a single mesh:
+
+- Organizational boundaries: lines of business
+- Service name or namespace collisions: multiple distinct uses of the `default`
+  namespace
+- Stronger isolation: isolating test workloads from production workloads
+
+In the most basic multiple mesh model, each mesh forms a strict
+boundary and there is no communication between meshes. However, you can enable
+inter-mesh communication with **mesh peering**. When
```
frankbu: "mesh peering" needs to be in the glossary.

frankbu:
@rcaballeromx This is pretty hard to review without a preview. Would the build work if you move the new glossary terms into this PR, instead of having a separate PR for them? I think it would make more sense to iterate on the concepts and related glossary terms in a single PR anyway.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+    caption="A service mesh with a single cluster"
+    >}}
+
+Single cluster deployments offer simplicity, but lack fault isolation and
+failover. If you need higher availability, you should use multiple clusters.
+
+### Multiple clusters
+
+You can configure your mesh to include multiple clusters; such configurations
+are known as **multi-cluster**. Using multiple clusters
```
frankbu: We need to decide if the term "multi-cluster" is defined to be a single mesh or does it also cover the case of multiple meshes that are federated. This definition seems to be the former, but @smawson is suggesting the latter: https://github.com/istio/istio.io/pull/4788#discussion_r315451541.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+the following tasks:
+
+- Span [one or more clusters](#cluster-models)
+- Span [one or more networks](#network-models)
+- Use [namespaces for tenancy](#namespace-tenancy)
+
+### Multiple Meshes
+
+You can deploy multiple meshes; such a configuration is known as
+**multi-mesh**.
```
frankbu: So this is what I was proposing to call a mesh federation in https://github.com/istio/istio.io/pull/4788.
Notice that I had been defining it as a noun, "a mesh federation" similar to a "federation of states"? I see that below federation is used as a verb.

So, in retrospect I think I like this better. Using the name "multi-mesh" for the noun lets us then define "mesh federation" to be the act of federating two or more meshes in a multi-mesh deployment. I think that would be less ambiguous.

We need to add glossary entries for "Multi-mesh Deployment" and "Mesh Federation" to make sure it's totally clear how they relate.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+    caption="Multiple service meshes"
+    >}}
+
+To avoid service naming collisions, you can give each mesh a globally unique
+**mesh ID**, to ensure that the fully qualified domain
+name (FQDN) for each service is distinct.
+
+When peering two meshes that do not share the same
+{{< gloss >}}trust domain{{< /gloss >}}, {{< gloss >}}identity{{< /gloss >}} and
+**trust bundles** must be federated between them. See the section
```
frankbu: "federated" should point to the glossary entry for mesh federation

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+instances with identical configuration.
+
+Depending on your overall system architecture and cloud provider, the following
+Istio control plane models are possible:
+
+- [single control plane](#single-control-plane)
+- [multiple control planes](#multiple-control-planes)
+- [managed control planes](#managed-control-plane)
+
+### Single control plane
```
frankbu: As @costinm has pointed out several times, a single control plane is not a realistic deployment. The topology that we're interested in describing is a "shared" control plane, i.e., a control plane that is shared by more than one cluster. Even in this case, it would be likely that we would have multiple shared control planes for HA.

rcaballeromx:
@frankbu I can certainly update this PR to include the glossary entries.
My thought in having a separate PR was to be able to have two separate conversations.
If it is going to make reviewing difficult, I have no problem closing the other PR and include the glossary entries.
@geeknoid WDYT?

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @frankbu What would be an appropriate definition of `mesh federation` in your opinion?

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @frankbu What would be an appropriate definition of `mesh peering` in your opinion?

rcaballeromx:
This PR now includes the relevant glossary entries originally in #4812 
@smawson @myidpt PTAL

content/en/docs/concepts/deployment-models/index.md
frankbu: I don't know, that's why I think it needs to be in the glossary :)
Where did the `mesh peering` term come from? @nmittler? I assume the definition would be something analogous to `network peering`: https://en.wikipedia.org/wiki/Peering

content/en/docs/concepts/deployment-models/index.md
frankbu: Like I said in my [other comment](https://github.com/istio/istio.io/pull/4813#discussion_r316310594) I think it's the verb form of what I was proposing in https://github.com/istio/istio.io/pull/4788. That is, setting up the loose coupling between the isolated service meshes in a multi-mesh deployment.

It does make we wonder if "peering" and "federating" are the same thing.

@nmittler, WDYT?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+Each identity Istio assigns is understood throughout the mesh and you can use it
+to enable mutual authentication and enforce policies.
+
+An identity contains a substring identifying the mesh that created it. For
+example in `spiffe://mytrustdomain.com/ns/default/sa/myname` the substring
+identifying the mesh is: `mytrustdomain.com`. We call this substring the {{<
+gloss>}}trust domain{{< /gloss >}} of the mesh. Every Istio mesh has a globally
+unique trust domain used to create identity names.
+
+The Certificate Authority (CA) creates the identities and their key material.
```
smawson: The CA signs the identities, it doesn't create them. The identities are created through k8s api server today (by creating a service account).

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+You can verify the message sender's identity with the public key of the CA that
+created the identity. A **trust bundle** is the set of
```
smawson: s/created/signed/

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+all CA public keys used by an Istio mesh. With a mesh's trust bundle, anyone can
+verify the sender of any message coming from that mesh.
+
+### Trust within a mesh
+
+Within a single Istio mesh, a CA creates identities and ensures that each
```
smawson: Istio is ensuring this property, not the CA. The CA is just signing the identities. Istio ensures each workload instance has an appropriate certificate representing its own identity, and the trust bundle necessary to recognize all identities in the mesh and in federated meshes.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+    link="./single-trust.svg"
+    alt="A service mesh with a certificate authority"
+    title="Trust within a mesh"
+    caption="A service mesh with a certificate authority"
+    >}}
+
+### Trust between meshes
+
+If a service in a mesh requires a service in another, you must federate identity
+and trust between the two meshes. To federate identity and trust,, you must
```
smawson: s/,,/,/

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+exchange the trust bundles of the meshes. You can exchange the trust bundles
```
smawson: for a multi-mesh scenario, we want to actually automate this I think. But we have nothing so far. Maybe we can mention that some of this is coming soon?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+{{< image width="50%"
+    link="./exp-ns.svg"
+    alt="A service mesh with two namespaces and an exposed service"
+    title="Namespaces with an exposed service"
+    caption="A service mesh with two namespaces and an exposed service"
+    >}}
+
+When using [multiple clusters](#multiple-clusters), the
+namespaces in each cluster sharing the same name are considered as the
```
smawson: s/considered as the same/considered to be the same/

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+same namespace. For example, `foo/B` in `cluster-1` and `foo/B` in
```
smawson: Maybe break these apart:

service B in namespace foo in cluster-1 and service B in namespace foo in cluster-2 refer to the same service, ...

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+Istio supports using clusters as a unit of tenancy. In this case, you can give
+each team a dedicated cluster or set of clusters to deploy their
+workloads. Permissions for a cluster are usually limited to the members of the
+team that owns it. You can set various roles for finer grained control, for
+example:
+
+- Cluster administrator
+- Developer
+
+With cluster tenancy, Istio considers each cluster as an independent mesh.
```
smawson: Istio treats each cluster as an independent mesh.

I think we should reword this though, so we should say:

To use cluster tenancy with Istio, each cluster should be configured as an independent mesh.

Not to add too much complexity, but we should be clear that Istio can be used where a group of clusters is a tenant, not just a single cluster. So each team may own one or more clusters, and those are configured as a single mesh. Then mesh federation / multi-mesh is used to connect the meshes of various teams together.

content/en/docs/reference/glossary/identity.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Identity
+---
+
+An identity is a unique name backed by key material, for
```
smawson: Same comment about key material, that isn't the important part necessarily.

Maybe if we want to call that out we can say:

An identity is a unique service account name within a namespace, typically available through a cryptographically signed mechanism such as:

content/en/docs/reference/glossary/trust-domain.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Trust Domain
+---
+
+A trust domain is a globally unique name that Istio uses to create all
+[identities](/docs/reference/glossary/#identity) within a mesh. Every mesh has
+an exclusive trust domain.
```
smawson: Not true today...

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+- Scaling of network addresses
+
+In this model, the workload instances in different networks can only reach each
+other through one or more [Istio gateways](/docs/concepts/traffic-management/#gateways).
+Istio uses **partitioned service discovery** to provide consumers a different
+view of {{< gloss >}}service endpoint{{< /gloss >}}s. The view depends on the
+network of the consumers.
+
+{{< image width="50%"
+    link="./multi-net.svg"
```
smawson: Should we have a gateway on each side rather than one in the middle?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+- Stronger isolation: isolating test workloads from production workloads
+
+In the most basic multiple mesh model, each mesh forms a strict
+boundary and there is no communication between meshes. However, you can enable
+inter-mesh communication with **mesh peering**. When
+peering, each mesh can export a set of services and identities, which all
+participating meshes can recognize.
+
+{{< image width="50%"
+    link="./multi-mesh.svg"
```
smawson: Should we show only certain services being exported? It is an important part of mesh federation that by default nothing is exposed, each service must explicitly choose to export itself to another mesh.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+- Greatly reduce the complexity of user deployments
+- Greatly reduce system administration costs
+- Effectively eliminate the usage costs associated with the control plane:
+    - CPU
+    - Memory
+    - Network
+
+{{< image width="50%"
+    link="./managed-control.svg"
```
smawson: Should we have the managed control plane outside the mesh rather than inside?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+### Trust within a mesh
+
+Within a single Istio mesh, a CA creates identities and ensures that each
+workload instance has the key material needed to recognize all identities in the
+mesh. This model allows workload instances in the mesh to authenticate each
+other when communicating.
+
+{{< image width="50%"
+    link="./single-trust.svg"
```
smawson: I think we should still have the istio control plane interposing between the CA and the proxies, since it is managing lifecycle of the certs, the CA is just issuing them.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+If a service in a mesh requires a service in another, you must federate identity
+and trust between the two meshes. To federate identity and trust,, you must
+exchange the trust bundles of the meshes. You can exchange the trust bundles
+either manually or automatically using a protocol such as [SPIFFE Trust Domain Federation](https://docs.google.com/document/d/1OC9nI2W04oghhbEDJpKdIUIw-G23YzWeHZxwGLIkB8k/edit).
+Once you import a trust bundle to a mesh, you can configure local policies for
+those identities.
+
+{{< image width="50%"
+    link="./multi-cluster.svg"
```
smawson: Wrong image

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+    >}}
+
+When using [multiple clusters](#multiple-clusters), the
+namespaces in each cluster sharing the same name are considered as the
+same namespace. For example, `foo/B` in `cluster-1` and `foo/B` in
+`cluster-2` refer to the same service, and Istio merges their endpoints for
+service discovery and load balancing.
+
+{{< image width="50%"
+    link="./cluster-ns.svg"
```
smawson: This picture is a little confusing. I wonder if we should show two namespaces, each spanning across the clusters? Or have the namespace span over both clusters rather than being in each?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+team that owns it. You can set various roles for finer grained control, for
+example:
+
+- Cluster administrator
+- Developer
+
+With cluster tenancy, Istio considers each cluster as an independent mesh.
+
+{{< image width="50%"
+    link="./cluster-iso.svg"
```
smawson: Should have Business Unit 1 and Business Unit 2 (missing the 2).

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
+---
+title: Istio Deployment Models
```
frankbu:
```suggestion
title: Deployment Models
```

content/en/docs/concepts/deployment-models/index.md
nmittler: +1 to `multi-cluster` and `multi-mesh`

content/en/docs/concepts/deployment-models/index.md
frankbu: @geeknoid has always argued (and I agree) that using Istio in any titles is redundant.

content/en/docs/concepts/deployment-models/index.md
nmittler: @frankbu yes, that's exactly my thinking as well!

content/en/docs/concepts/deployment-models/index.md
nmittler: Yup ... peering and federation are the same thing. Let's replace `peering` with `federation` throughout.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+{{< image width="50%"
+    link="./multi-net.svg"
+    alt="A service mesh with multiple networks"
+    title="Multi-network deployment"
+    caption="A service mesh with multiple networks"
+    >}}
+
+## Mesh models
+
+Istio supports two mesh models:
```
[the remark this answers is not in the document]
nmittler: Agreed the wording does feel strange

+1 to @smawson's suggestion

content/en/docs/reference/glossary/trust-domain.md
nmittler: yeah, but that *is* the definition ... should we leave it for now? Maybe adding a note "coming soon"? :)

smawson: change to "should have"?

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: Yes, I'll add the missing no.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+network topologies.
+
+### Single network
+
+In the simplest case, a service mesh operates over a single fully
+connected network. A single network model satisfies the following conditions:
+
+- The network consists of a number of connected subnets without any
+  overlap in IP or VIP ranges.
+- All {{< gloss >}}workload instance{{< /gloss >}}s can reach each other directly
```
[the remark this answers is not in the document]
rcaballeromx: The issue is that the glossary entry is in singular, if I add the `s` then the glossary markup breaks.
The render does show the term correctly even if the `s` is left without an underline.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: +1 In the text, we tried to be very consistent and used the two very distinctly. 
@smawson  Once this is approved, we will need to invest some time making the existing content consistent with the definitions and models we are introducing here.

content/en/docs/concepts/deployment-models/index.md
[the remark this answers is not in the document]
smawson: Ah hah, I found the solution:

{{< gloss workload instance >}}workload instances{{< /gloss >}}

See if that works :)

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: I tried to keep the diagrams as simple as possible. For that reason, I place the gateway in the middle.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+Istio supports two mesh models:
+
+- Single mesh
+- Multiple meshes, also known as **multi-mesh**
+
+### Single mesh
+
+The simplest Istio deployment is a single mesh. Within a mesh, service names are
+unique. For example, only one service can have the name `foo/mysvc`.
+Additionally, workload instances all share a common identity and authentication
```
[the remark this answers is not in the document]
rcaballeromx: @smawson Would this wording work?

```
Workload instances can all share a common identity and authentication infrastructure and can authenticate each other. Istio also supports authentication that only uses shared trust. However, workload instances do have a common identity since service account names are unique within a namespace, just like service names.
```

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+infrastructure and can authenticate each other. With a single mesh you can do
+the following tasks:
```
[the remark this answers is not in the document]
rcaballeromx: It is considered poor style to force readers to complete the ideas presented on a list or to break out the ideas.
@smawson Would the following sentence work better? 

```
A single mesh can span [one or more clusters](#cluster-models), span [one or more networks](#network-models), and use [namespaces for tenancy](#namespace-tenancy).
```

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: I also had concerns showing all the services connecting with each other. That's why I added the trust bundles to the diagram. 
@smawson Since the default is to have nothing exposed, I would suggest I change the diagram to only show a connection between `Service A` and `Service B` and a connection between `Service C` and `Service D`.
Would that change address your concerns?

content/en/docs/concepts/deployment-models/index.md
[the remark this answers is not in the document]
smawson: I would just drop the part about common authentication infrastructure or shared trust:

Additionally, workload instances all share a common identity as service accounts names are unique within a namespace, just like service names.

content/en/docs/concepts/deployment-models/index.md
smawson: I'm worried it may be confusing, since there isn't actually a shared gateway. Each network needs its own gateway.

content/en/boilerplates/notes/1.1.md
```diff
@@ -51,7 +51,7 @@ concise list of things you should know before upgrading your deployment to Istio
 - **Namespace Scoping**. When referring to a `VirtualService` in a Gateway we use DNS based name matching
   in our configuration model. This can be ambiguous when more than one namespace defines a virtual service
   for the same host name. To resolve ambiguity it is now possible to explicitly scope these references
-  by namespace using a syntax of the form **`[{namespace-name}]/{hostname-match}`** in the `hosts` field.
+  by namespace using a syntax of the form **`[{namespace-name}](/{hostname-match}`** in the `hosts` field.
```
myidpt:
```suggestion
  by namespace using a syntax of the form **`[{namespace-name}]/{hostname-match}`** in the `hosts` field.
```

content/en/docs/concepts/deployment-models/index.md
smawson: I was thinking more like A or B can talk to C but they can't talk to D.

content/en/docs/reference/glossary/identity.md
```diff
@@ -0,0 +1,15 @@
...
+example:
+
+- [X.509 certificate](https://en.wikipedia.org/wiki/X.509)
+- [JWT](https://en.wikipedia.org/wiki/JSON_Web_Token)
+
+Each identity Istio assigns is understood throughout the mesh and you can use it
+to enable mutual authentication and enforce policies.
+
+An identity contains a [trust domain](/docs/reference/glossary/#trust-domain)
+identifying the mesh that created it.
```
myidpt:
```suggestion
identifying the mesh it belongs to.
```

content/en/docs/concepts/deployment-models/index.md
frankbu: I agree with @smawson, gateways are not bi-directional so the diagram is confusing.

Btw, the fonts in all the diagrams are too small in the preview. I also see that the fonts in the traffic management concepts diagrams are also still not right. What's the plan for fixing them?

content/en/docs/reference/glossary/trust-domain.md
nmittler: So `Every mesh should have an exclusive trust domain.`?

Yeah I think that works for now.

content/en/docs/concepts/deployment-models/index.md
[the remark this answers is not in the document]
nmittler: Could we blend the two? Maybe:

```
A single mesh can span [one or more clusters](#cluster-models) and
[one or more networks](#network-models).

Within a mesh, [namespaces](#namespace-tenancy) are used for [tenancy](#tenancy-models).
```

content/en/docs/concepts/deployment-models/index.md
nmittler: +1 to replacing peering with federation

content/en/docs/concepts/deployment-models/index.md
nmittler: Yeah should be an easy fix .. we already do something similar in another diagram.

content/en/docs/concepts/deployment-models/index.md
nmittler: Oof yes ... I didn't notice that when I reviewed previously.

@rcaballeromx maybe we can move the managed control plane to the top or bottom of the diagram to keep it outside the mesh?

frankbu: I think it's reasonable to show it inside the mesh. Even though the control plane components are not really "in the mesh", even in the other control plane diagrams, they're usually running in the same cluster as the mesh services, so you can sort of think of it as part of the mesh for those clusters. So, in the same way, there's a conceptual managed control plane that is dedicated to each mesh, so the diagram seems OK to me.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+    link="./managed-control.svg"
+    alt="A service mesh with a managed control plane"
+    title="Managed control plane"
+    caption="A service mesh with a managed control plane"
+    >}}
+
+## Identity and trust models
+
+When a workload instance is created within a service mesh, Istio assigns the
+workload an identity. An identity is a unique name backed by a key material, for
```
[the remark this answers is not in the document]
nmittler: @smawson if the name, itself, is the identity then it's not cryptographically signed, is it? Rather, it's backed by something that is signed (e.g. cert). With SPIFFE, for example, the identity would be the SPIFFE ID (which is just the sa name). The SVID, however, (which contains the ID) is the signed and verifiable "thing". So which is the identity in this case? The SVID or just the ID? 

I think we need to get this definition right here ... otherwise it will be a source of confusion.

content/en/docs/concepts/deployment-models/index.md
nmittler: > Also note that this section is specifically about single cluster meshes, in which case a single control plane is fine. But of course that is also a control plane per cluster...

That's right ... with a single cluster, you would only have a single control plane deployment. And that's what this section calls out. 

> Ideally we can describe that you can have one or more logical control planes, each of which is typically replicated into one or more clusters. Some extremes are control plane per cluster, or single logical control plane with multiple replicas (which is very similar to a managed control plane, just a self-managed one).

@smawson let's chat today so I have a better understanding of the various control plane models you're referring to. I'd like to condense this down to something that's reasonably abstract here.

content/en/docs/reference/glossary/trust-domain.md
frankbu: I think we should try to avoid talking about things "coming soon" in the glossary (and in the docs in general). I like the "should have ..." suggestion.

content/en/docs/concepts/deployment-models/index.md
nmittler: @rcaballeromx looks like you're using the multi-cluster diagram here ... should be using mesh-federation.

content/en/docs/concepts/deployment-models/index.md
frankbu: I agree. How about we start by changing this section to "Shared control plane" and change the diagram to be similar to the multiple control planes diagram, bit with a control plane in only one of the clusters.

We should also talk about how HA would be configured with this shared control plane setup.

content/en/docs/concepts/deployment-models/index.md
frankbu: I think it's fine as is, but that's just my opinion :)

content/en/docs/setup/install/multicluster/shared-gateways/index.md
```diff
@@ -10,7 +10,7 @@ aliases:
 ---
 
 Follow this guide to configure a multicluster mesh using a
-[shared control plane topology](/docs/concepts/multicluster-deployments/#shared-control-plane-topology)
+[single control plane deployment](/docs/concepts/deployment-models/#single-control-plane)
```
vadimeisenbergibm: Sorry to be late to the discussion, but are we back with single control plane deployment instead of shared control plane? I thought we agreed once that in this topology we can have multiple control planes shared, for example, two clusters with a control plane and two clusters without a control plane. cc @costinm

frankbu: Oh yes, I missed this one. There's currently a mixture of the term "shared" and "single" in this PR, in several places. This example should definitely be called "shared control plane deployment", but see my other comment about the need to describe shared control plane in the concepts.

content/en/docs/reference/glossary/identity.md
nmittler: Good call, @vadimeisenbergibm 

@rcaballeromx I think this (or some form of it) could probably just be used as our glossary definition for identity.  WDYT?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+{{< image width="50%"
+    link="./single-cluster.svg"
+    alt="A service mesh with a single cluster"
+    title="Single cluster"
+    caption="A service mesh with a single cluster"
+    >}}
+
+Single cluster deployments offer simplicity, but lack fault isolation and
+failover. If you need higher availability, you should use multiple clusters.
```
vadimeisenbergibm: FWIW: here is a list of reasons I found to use multiple clusters: http://vadimeisenberg.blogspot.com/2019/03/multicluster-pros-and-cons.html

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+### Multiple clusters
+
+You can configure your mesh to include multiple clusters; such configurations
+are known as **multi-cluster**. Using multiple clusters
+affords the following capabilities beyond that of a single cluster model:
+
+- Fault isolation and failover: `cluster-1` goes down, failover to `cluster-2`.
+- Location-aware routing and failover: Send requests to the nearest service.
+- Configuration isolation: A wrong configuration only affects a single cluster.
```
vadimeisenbergibm: > A wrong configuration only affects a single cluster.

Is it correct in the case of single/shared control plane? A wrong Istio configuration can affect multiple clusters.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+Many production systems require multiple networks or subnets for isolation
+and high availability. Istio supports spanning a service mesh over a variety of
+network topologies.
+
+### Single network
+
+In the simplest case, a service mesh operates over a single fully
+connected network. A single network model satisfies the following conditions:
+
+- The network consists of a number of connected subnets without any
```
vadimeisenbergibm: > The network consists of a number of connected subnets without any overlap in IP or VIP ranges.

Isn't it a technical detail, irrelevant to the users who are not interested to know how the next item is achieved? Suppose I am a user and I order several clusters from my cloud provider. I am interested to know if workload instances from one cluster can reach workload instances in other clusters. Why should I understand subnets, IP and VIP ranges? I am sure this is an interesting information for networking people, but is it necessary for all the Istio users?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+    >}}
+
+### Multiple networks
+
+You can span a single service mesh across multiple networks; such a
+configuration is known as **multi-network**.
+
+Multiple networks afford the following capabilities beyond that of single networks:
+
+- An existing network topology
```
vadimeisenbergibm: Why "an existing network topology" is a capability?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+- Overlapping IP or VIP ranges for **service endpoints**
+- Crossing of administrative boundaries
+- Fault tolerance
+- Scaling of network addresses
```
vadimeisenbergibm: Similarly to my comment about subnets and IP ranges - the list above is interesting to networking people or cloud providers or infrastructure providers. As a user of multiple clusters, why should I be interested in this list? I just order my clusters from my cloud provider(s) or my infrastructure team, and they setup the networks for the clusters for me. Why should a typical Istio user bother themselves with the details above?

content/en/docs/concepts/deployment-models/index.md
nmittler: Agreed.

Let's drop the discussion of subnets altogether ... it really is an implementation detail. Let's just focus on a "network" in terms of reachability.

content/en/docs/reference/glossary/identity.md
rcaballeromx: @nmittler @vadimeisenbergibm Agreed. I'll use the content in the security concept to craft a definition for the next update. Thanks.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+You can deploy multiple meshes; such a configuration is known as
+**multi-mesh**.
+
+Multiple meshes afford the following capabilities beyond that of a single mesh:
+
+- Organizational boundaries: lines of business
+- Service name or namespace collisions: multiple distinct uses of the `default`
+  namespace
+- Stronger isolation: isolating test workloads from production workloads
```
vadimeisenbergibm: Does the colon here mean "for example"? There are multiple reasons for isolation.

content/en/docs/reference/glossary/trust-domain.md
rcaballeromx: I agree with @frankbu regarding forward looking statements. If it is not true now, we can leave it out until it is and then make a PR to add it.

rcaballeromx: @nmittler @smawson Would leaving out the sentence for now work?

content/en/docs/concepts/deployment-models/index.md
nmittler: Agreed ... I think we can just drop this from the list entirely. It's not really adding any value.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+{{< image width="50%"
+    link="./multi-mesh.svg"
+    alt="Multiple service meshes"
+    title="Multi-mesh"
+    caption="Multiple service meshes"
+    >}}
+
+To avoid service naming collisions, you can give each mesh a globally unique
+**mesh ID**, to ensure that the fully qualified domain
+name (FQDN) for each service is distinct.
```
vadimeisenbergibm: This paragraph mentions some implementation detail, it is not clear how it is related to the text. Which FQDN is distinct? Isn't it always `<service>.<namespace>.svc.cluster.local`?

content/en/docs/setup/install/multicluster/shared-gateways/index.md
rcaballeromx: @frankbu @vadimeisenbergibm  The content added by this PR does not mix `shared` and `single`. Quite the contrary, I made an additional effort to be consistent in the PR and use only `single` even when updating the links. It is beyond the scope of this PR to make all content outside of the concept consistent. That would need to be its own effort post 1.3.

content/en/docs/concepts/deployment-models/index.md
nmittler: I think the list is needed. We need to define why you would use one model vs the other, and this is why. Yes, perhaps it a bit into details, but that's also where the differentiation lies. I don't think this list is pulling down the abstract nature of this document much, if any ... it's just giving a brief introduction into why multiple networks is a thing.

content/en/docs/concepts/deployment-models/index.md
vadimeisenbergibm: @nmittler
> with a single cluster, you would only have a single control plane deployment

Please note that you can have multiple control planes per single cluster, see this blog post https://istio.io/blog/2018/soft-multitenancy/

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+You can deploy multiple control planes across the mesh; such a configuration is
+known as a **multi-control plane**.
+
+Multiple control planes afford improved performance and availability for meshes
+that span [multiple clusters](#multiple-clusters) across multiple zones and
+regions.
+
+Istio supports control plane deployments at various levels of availability:
+
+- Control plane per region: *lowest availability*
```
vadimeisenbergibm: So in this model, the control planes are shared between the clusters? There is a single control plane per region that is shared between the clusters in the region?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+    >}}
+
+### Managed control plane
+
+Many cloud providers offer a control plane, which they manage on behalf of their
+customer.
+
+Typically, managed control planes guarantee some level of performance and
+availability, which you can assume when building your system. Managed control
+planes affords the following benefits:
```
vadimeisenbergibm: affords -> afford

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+
+- Greatly reduce the complexity of user deployments
```
vadimeisenbergibm: Is "greatly" necessary here?

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+- Performance
+
+Istio supports two types of tenancy models:
+
+- [Namespace tenancy](#namespace-tenancy)
+- [Cluster tenancy](#cluster-tenancy)
+
+### Namespace tenancy
+
+In Istio, a [namespace](https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-namespace)
```
vadimeisenbergibm: Is the term `namespace`  and `namespace tenancy` supported in non-Kubernetes environments?

content/en/docs/concepts/deployment-models/index.md
vadimeisenbergibm: The question who is the audience. Will the readers setup their networks or will they receive clusters from their cloud providers/infrastructure teams with given networking. My guess is that many readers will not understand these networking concepts.

content/en/docs/setup/install/multicluster/shared-gateways/index.md
frankbu: @rcaballeromx, I think you still have at least one link to `single` that is titled `shared`.

That said, those links (from the install section) should be titled `shared` and we need to either add a section in the concepts to describe `shared` or change the `single control plane` section to explain what `shared` means. 

We've had many discussions about this and concluded that `single control plane` is not a realistic Istio production configuration (even for single cluster), so we shouldn't describe it as a proper multicluster installation. That's why we updated the setup instructions to not say `single`. This PR should not undo that decision.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @vadimeisenbergibm There can be many reasons, the concept scope is limited to just the key reason. IMO, we should add a link to your blog post at the beginning of the multi-cluster installation but that would need to happen in a different PR.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @john-a-joyce had strong feeling about it if I remember correctly. Would it be ok to drop the subnets as suggested?

content/en/docs/concepts/deployment-models/index.md
nmittler: I don't think we can make assumptions either way. We can't make it a gate that every possible reader of this document will understand every word. If you're in a networking section of a document, I think it's fair that you'll see some terminology related to networking. The level of depth here is not a significant hurdle, and there is nothing preventing the user from just reading on.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @nmittler @vadimeisenbergibm I'm not sure, readers need to know that if they have an existing network topology with multiple networks this is the model for them. WDYT?

nmittler: @rcaballeromx the problem is that an existing network topology may very well be "single network", so it's out of place in either section. Perhaps you could just add a blurb to the introduction to  "network models" indicating that Istio supports various models, allowing them to span a single mesh over an existing topology if they have one.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @smawson @frankbu I'm changing the diagram based on your comments. Regarding the fonts in diagrams, that is an issue that is caused by the available fonts in your local system. I can convert the text to outlines but then it stops being editable and accessible to screen readers, which is terrible for accessibility reasons. How would you recommend I proceed?

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: I'll introduce new glossary entries for both `multi-mesh` used as a noun, and `mesh federation` as the following verb:
```
Mesh federation is the act of federating two or more meshes in a multi-mesh deployment.
```
@smawson @nmittler @frankbu Would that definition work?

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: We don't want to list all reasons for isolation, just give a common example. Just like in the other bullet points.

content/en/docs/concepts/deployment-models/index.md
nmittler: I think we should leave "federating" out of the definition.  Maybe:

```
Mesh federation is the act of exposing services between meshes in a multi-mesh
deployment. In a federation, each mesh may expose a subset of their services
for one or more other meshes to consume. This affords service communication
across mesh boundaries.
```

frankbu: That's a circular definition of federate :)

I think something along the lines of what @smawson said [here](https://github.com/istio/istio.io/pull/4813#discussion_r316395847) would work.
```
Mesh federation is the act of connecting two meshes together by sharing trust and identity information as well as service discovery information.
```

content/en/docs/concepts/deployment-models/index.md
frankbu: @rcaballeromx, are you saying the small fonts are only a problem for a small number of users, which I'm one of? For me, much of the text in diagrams is not really legible, without clicking to enlarge the diagram. If that's not the case for most users, then I think its OK, but if it's not just me, it seems like it needs to be fixed.

content/en/docs/concepts/deployment-models/index.md
nmittler: @frankbu 

> That's a circular definition of federate :)

Which definition are you referring to? [Rigs's](https://github.com/istio/istio.io/pull/4813#discussion_r316810773) or [mine](https://github.com/istio/istio.io/pull/4813#discussion_r316815058)?

> I think something along the lines of what @smawson said here would work.

The problem with that definition in my mind is that it doesn't really paint the picture of the objective of mesh federation. The point is that they can share services. The phrases `connecting two meshes together` and `sharing trust and identity... as well as service discovery information` doesn't really make it explicit. In addition, saying that they share `service discovery information` sounds like they expose everything, which very likely won't be the case.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+## Identity and trust models
+
+When a workload instance is created within a service mesh, Istio assigns the
+workload an identity. An identity is a unique name backed by a key material, for
+example:
+
+- [X.509 certificate](https://en.wikipedia.org/wiki/X.509)
+- [JWT](https://en.wikipedia.org/wiki/JSON_Web_Token)
+
+Each identity Istio assigns is understood throughout the mesh and you can use it
```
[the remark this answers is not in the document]
rcaballeromx: @smawson `can be used` is considered poor style. It is better to address reader directly. Can you help me understand why you would like to replace `you can use it`? Maybe I can address your concern with a different version of the sentence.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+to enable mutual authentication and enforce policies.
+
+An identity contains a substring identifying the mesh that created it. For
```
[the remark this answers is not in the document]
rcaballeromx: @smawson I agree that we should mention that these apply only if readers want to federate meshes. That said, we need to be careful not to introduce redundancies. I'll rewrite the text to make it clearer.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @smawson I'll change the text to reflect that nuance. Do you think I should include the fact about how the identities are created of would that be a k8s-specific implementation detail?

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @smawson I'll modify the text accordingly.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @smawson Would `considered the same` work? It is not good to introduce auxiliary verbs since they complicate grammar making the text harder for non-English native speakers.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: If the concern is the size of the text, I can always enlarge the diagrams. I can also enlarge the text but that could make the diagrams appear too loaded. @frankbu I suggest I make some adjustments to the diagrams for the next update based on your feedback. We can revisit after that, would that work?

content/en/docs/concepts/deployment-models/index.md
nmittler: @frankbu The point of this section is to abstract how the managed control plane is implemented ... it's not a concern of the user.

A managed control plane is inherently *not* part of the mesh and it should be represented as such in the diagram. The cloud provider will likely deploy an instance of their managed control plane per user mesh ... but again, it's an implementation detail. I think if we consistently picture the managed control plane outside of the mesh we avoid confusing the reader.

content/en/docs/concepts/deployment-models/index.md
vadimeisenbergibm: I just meant that your sentence can make impression that the only reason for multiple clusters is high availability, while there are many reasons for multiple clusters.

content/en/docs/concepts/deployment-models/index.md
frankbu: @nmittler, I meant Rigs' definition was circular `federation is the act of federating` :)

I see what you mean about @smawson's definition. Your suggestions works for me, although maybe we should avoid using the word federation as both a verb and a noun.

Should we also mention that the meshes being federated are isolated from each other and have different service definitions and independent administrative control, or maybe link `multi-mesh deployment` to the definition where that's all covered?
```
Mesh federation is the act of exposing services between meshes in
a [multi-mesh deployment](link). Each mesh may expose 
a subset of its services for one or more other meshes to consume.
This affords service communication across mesh boundaries.
```

nmittler: @frankbu 

> Should we also mention that the meshes being federated are isolated from each other and have different service definitions and independent administrative control, or maybe link multi-mesh deployment to the definition where that's all covered?

I don't think so. That's effectively the definition of a service mesh, which is defined elsewhere on istio.io. We should be able to assume that as read for the purposes of this document.

Also, I think you're rewording of my definition looks good!

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @smawson I'll introduce a control plane as you suggest.

content/en/docs/setup/install/multicluster/shared-gateways/index.md
rcaballeromx: @frankbu I'll wait until we reach consensus on the control plane section in the concept and implement the word choice we establish there. Does that SGTY?

frankbu: Makes sense to hold off, but I think it really needs to be changed before the doc goes live.

content/en/docs/concepts/deployment-models/index.md
frankbu: I think that at a minimum the text should be tweaked to not make it sound like failover is the only reason to move away from single cluster.

content/en/docs/concepts/deployment-models/index.md
frankbu: I still think that even in the non-managed control plane case the control plane is not part of the mesh, yet it is shown inside the mesh in those diagrams. So for consistency, I think it would be less confusing to leave it inside the mesh box in all the control plane diagrams. That said, I don't have a strong objection to changing it, so whatever you choose is fine.

nmittler: For non-managed control planes, it gets tricky ... particularly in the case of multi-cluster, since you have to show the control plane in each cluster.

For non-managed I think it's generally helpful to picture the control plane inside the mesh to show that it's dedicated to that mesh. This is especially helpful in multi-mesh scenarios. 

Managed is a different animal altogether as the topology is purposely left unknown, so outside of the mesh is the only thing that makes sense.

frankbu: Fine. But I think you're getting to the heart of the issue with control plane configuration: physical vs. logical, shared vs. dedicated, replicated and synchronized vs. federated. I think we need to explain these things.

content/en/docs/concepts/deployment-models/index.md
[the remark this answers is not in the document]
smawson: Righto, the identity is a trust domain, namespace, and service account. The SVID is a representation of that identity, and the cert is a signed attestation of that identity. But fundamentally the identity is just the tuple of those pieces, with lots of representations.

content/en/docs/concepts/deployment-models/index.md
smawson: @vadimeisenbergibm do you think we can skip going into that level of detail? Technically you could have a single cluster with many control planes, each in a different mesh :) Maybe we can just point at that blog and indicate that istio actually supports a wide range of control plane setups, but only a few standard ones?

content/en/docs/concepts/deployment-models/index.md
smawson: Yes that's right, and the single control plane per region may run in multiple clusters for HA. So its per-region shared control planes.

content/en/docs/concepts/deployment-models/index.md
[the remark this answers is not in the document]
smawson: Well it is because the reader is not the one using the identity. Istio uses that identity to enable mutual authentication and enforce policies.

Something like:

Each identity Istio assigns is understood throughout the mesh and Istio uses that identity to enable mutual authentication and enforce policies.

The one thing the user does do is define policies, so if we wanted to talk to the user we'd say that:

... and you can use it to define policies enforced by Istio.

but that's a bit weird.

content/en/docs/concepts/deployment-models/index.md
smawson: Yeah that's maybe overly K8s specific, although if we define an identity as a namespace/service account that's also k8s specific. We haven't really discussed non-k8s identities enough (even though we can in theory support them).

Maybe we can make this somewhat generic and just say:

The Certificate Authority (CA) is responsible for creating and signing the certificates used to attest identities used within the mesh.

Actually that's a bad sentence so please rewrite, but you get the idea :)

content/en/docs/concepts/deployment-models/index.md
smawson: They are for mesh expansion, although that is kind of cheating since it is based on a k8s foundation. But that's a really good point that I have no idea how to resolve, since a lot of how meshes work is (overly) tied to k8s. Maybe we can somehow mention that Istio also works in non-k8s environments but we don't have the tenancy models defined well there yet...

content/en/docs/concepts/deployment-models/index.md
smawson: SGTM

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,421 @@
...
+name (FQDN) for each service is distinct.
+
+When peering two meshes that do not share the same
+{{< gloss >}}trust domain{{< /gloss >}}, {{< gloss >}}identity{{< /gloss >}} and
+**trust bundles** must be federated between them. See the section
+on [Multiple Trust Domains](#trust-between-meshes) for an overview.
+
+## Control plane models
+
+The Istio [control plane](https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-control-plane)
```
frankbu: This link to the Kubernetes definition of control plane is not appropriate, since Istio's control plane is not a Kubernetes control plane. We should probably add a generic glossary definition for control plane, and link to it.

content/en/docs/concepts/deployment-models/index.md
vadimeisenbergibm: @smawson I did not mean to add the multiple control planes in a single cluster to the document, just to make sure @nmittler is aware of that pattern.

content/en/docs/concepts/deployment-models/index.md
vadimeisenbergibm: Should we add "some compliance standards require network segmentation"?

From PCI DSS:

>Network segmentation of, or isolating (segmenting), the cardholder data environment from the remainder of an entity’s network is not a PCI DSS requirement. However, it is strongly recommended as a method that may reduce: 
•The scope of the PCI DSS assessment
•The cost of the PCI DSS assessment
•The cost and difficulty of implementing and maintaining PCI DSS controls
•The risk to an organization (reduced by consolidating cardholder data into fewer, more controlled locations) 
Without adequate network segmentation (sometimes called a "flat network") the entire network is in scope of the PCI DSS assessment. Network segmentation can be achieved through a number of physical or logical means, such as properly configured internal network firewalls, routers with strong access control lists, or other technologies that restrict access to a particular segment of a network. To be considered out of scope for PCI DSS, a system component must be properly isolated (segmented) from the CDE, such that even if the out-of-scope system component was compromised it could not impact the security of the CDE.

content/en/docs/concepts/deployment-models/index.md
vadimeisenbergibm: I see. So we can have a single control plane, and multiple control planes, which can be shared or dedicated. cc @frankbu . But should we, if so, split the multiple control planes into shared/dedicated subsections?

frankbu: @nmittler @smawson @rcaballeromx @vadimeisenbergibm Can we please stop using the term `single control plane`? I've had several long discussion with @costin in various design docs to come to a common understanding that single control plane makes no sense in production so we should not describe it as a realistic configuration.

I think the control plane section intro should just say something like:

You can have one or more control planes (usually more for HA).

Then the following sections describe these three kinds of control plane configurations:

1. dedicated (manage workloads in only one cluster)
2. shared (manage workloads in more than one cluster)
3. managed

We would explain how using multiple control planes offer HA and that they would be synchronized (replicated types) if they are in a single multi-cluster mesh. Control planes in a multi-mesh configuration, on the other hand, would have different types in each mesh's control plane(s).

content/en/docs/concepts/deployment-models/index.md
rvennam: Managed control plane doesn't necessarily mean that the control plane is out of the mesh right? Cloud providers can manage the Istio control plane (installation, updates, health), but still have it residing in the cluster. This also means that usage costs concerns are not eliminated.

content/en/docs/concepts/deployment-models/index.md
rvennam: "collisions" seems negative. Maybe "reuse" would be a better word to convey that this is a capability?

content/en/docs/concepts/deployment-models/index.md
rvennam: How about something like "Advanced Deployment Models"? I'm concerned that naming this topic just "Deployment Models" will attract all audiences and introduce them to complexities they don't need to know yet 🤷‍♂ just an opinion

content/en/docs/concepts/deployment-models/index.md
nmittler: Of course, cloud providers implement it however they choose. The physical location of the managed control plane shouldn't generally be a concern for the end-user.

We probably shouldn't make too many assumptions about cloud provider pricing models. Maybe we can restate it this way:

```
Eliminate many of the usage costs associated with the control plane
```

content/en/docs/concepts/deployment-models/index.md
nmittler: Possibly, although the simple models are here too. I'm hoping that the content here isn't too advanced for any end-user to grok.

rcaballeromx:
Thank you all for the valuable feedback. In accordance with today's working group meeting, I've updated this PR. Keep in mind that further adjustments will be made in follow up PRs. 
@frankbu @costinm @nmittler @geeknoid @smawson @louiscryan PTAL

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,398 @@
...
+    alt="Multiple service meshes"
+    title="Multi-mesh"
+    caption="Multiple service meshes"
+    >}}
+
+To avoid service naming collisions, you can give each mesh a globally unique
+**mesh ID**, to ensure that the fully qualified domain
+name (FQDN) for each service is distinct.
+
+When peering two meshes that do not share the same
```
nmittler: Missed rename from `peering` to `federating.

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,398 @@
...
+- Performance
+
+Istio supports two types of tenancy models:
+
+- [Namespace tenancy](#namespace-tenancy)
+- [Cluster tenancy](#cluster-tenancy)
+
+### Namespace tenancy
+
+If you use Istio in a Kubernetes environment, a
```
nmittler: Where did this discussion of kubernetes come from? I'm not sure we need it ... Istio defines namespaces, so namespace is a unit of tenancy within Istio, regardless of the underlying platform.

content/en/docs/reference/glossary/service-mesh.md
```diff
@@ -1,5 +1,14 @@
 ---
 title: Service Mesh
 ---
-A shared set of names and identities that allows for common policy enforcement and telemetry collection.
-[Service names](#service-name) and [workload instance principals](#workload-instance-principal) are unique within a service mesh.
+A *service mesh* or simply *mesh* is an infrastructure layer that enables
+managed, observable and secure communication between
+[workload instances](/docs/reference/glossary/#workload-instance)
```
frankbu:
```suggestion
[workload instances](/docs/reference/glossary/#workload-instance).
```

content/en/docs/reference/glossary/multi-cluster.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Multi-Cluster
+---
+
+Multi-cluster is a deployment model that consists of a
+[mesh](/docs/reference/glossary/#service-mesh) with multiple
```
frankbu:
```suggestion
single [service mesh](/docs/reference/glossary/#service-mesh) with multiple
```

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,398 @@
...
+
+Important system models impact your overall Istio deployment model. This page
+discusses the options for each of these models and describes how you can
+configure Istio to address them.
+
+## Cluster models
+
+A cluster is a set of compute nodes that run containerized applications.
+Typically, the compute nodes comprising a cluster can reach each other directly
+and you can limit external access through rules or policies. For isolation,
```
frankbu: The first two sentences are the same text as the glossary, so would it make more sense to replace the them with something like:

```
Workloads of an Istio application run in one or more {{< gloss >}}clusters{{< /gloss >}}.
```

content/en/docs/concepts/deployment-models/index.md
```diff
@@ -0,0 +1,429 @@
...
+
+## Mesh models
+
+Istio supports having all of your services in a **single mesh**, or connecting
+multiple meshes together, which is also known as **multi-mesh**.
+
+### Single mesh
+
+The simplest Istio deployment is a single mesh. Within a mesh, service names are
+unique. For example, only one service can have the name `mysvc` in the `foo/`
```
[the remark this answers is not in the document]
frankbu: Don't forget to remove the trailing `/` here ^^^

content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -70,13 +70,13 @@ To further customize Istio and install addons, you can add one or more `--set <k
 ## Multicluster profiles
 
 Istio provides two additional built-in configuration profiles that are used exclusively for configuring a
-[multicluster service mesh](/docs/concepts/multicluster-deployments/#multicluster-service-mesh):
+[multi-cluster deployment](/docs/concepts/deployment-models/#multiple-clusters):
 
 1. **remote**: used for configuring remote clusters of a
-    multicluster mesh with a [shared control plane topology](/docs/concepts/multicluster-deployments/#shared-control-plane-topology).
+    multicluster mesh with a suitable [control plane model](/docs/concepts/deployment-models/#control-plane-models).
```
frankbu: Suitable is not a good here, but we can rethink these external references in a followup PR.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -16,5 +16,5 @@ a combination of the approaches can be used. For example,
 two clusters might share a control plane while a third has its own.
 {{< /tip >}}
 
-Refer to the [multicluster service mesh](/docs/concepts/multicluster-deployments/) concept documentation
+Refer to the [multicluster service mesh](/docs/concepts/deployment-models/) concept documentation
```
frankbu:
```suggestion
Refer to the [multicluster service mesh](/docs/concepts/deployment-models/#multiple-clusters) concept documentation
```

content/en/docs/setup/install/multicluster/shared-vpn/index.md
```diff
@@ -9,7 +9,7 @@ aliases:
     - /docs/setup/kubernetes/install/multicluster/shared-vpn/
 ---
 
-Follow this guide to install an Istio [multicluster service mesh](/docs/concepts/multicluster-deployments/)
+Follow this guide to install an Istio [multicluster service mesh](/docs/concepts/deployment-models/)
```
frankbu:
```suggestion
Follow this guide to install an Istio [multicluster service mesh](/docs/concepts/deployment-models/#multiple-clusters)
```

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: Agreed, I am using the introduced terminology though:
```
The workload instances of your application run in one or more {{< gloss "cluster" >}}clusters{{< /gloss >}}. For ...
```

content/en/docs/concepts/deployment-models/index.md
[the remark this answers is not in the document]
rcaballeromx: @frankbu Sorry for missing it before.

content/en/docs/concepts/deployment-models/index.md
rcaballeromx: @smawson We will iterate on the content in the future and we can add to it when the automation is in place.

content/en/docs/reference/glossary/multi-cluster.md
rcaballeromx: `a mesh` is already singular. I avoided `service` to keep the definition focused on the cluster level.

content/en/docs/reference/glossary/identity.md
```diff
@@ -0,0 +1,43 @@
...
+- GCP: GCP service account
+
+- AWS: AWS IAM user/role account
+
+- On-premises (non-Kubernetes): user account, custom service account, service
+  name, Istio service account, or GCP service account. The custom service
+  account refers to the existing service account just like the identities that
+  the customer’s Identity Directory manages.
+
+Typically, the [trust domain](/docs/reference/glossary/#trust-domain) specifies
```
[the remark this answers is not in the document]
rcaballeromx: @see the linked definition of trust domain.

content/en/docs/reference/glossary/identity.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Identity
+---
+
+Identity is a fundamental security infrastructure concept. The Istio identity
+model is based on a first-class service identity. At the beginning of
```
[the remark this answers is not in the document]
rcaballeromx: @costinm I changed it to workload identity.

content/en/docs/reference/glossary/identity.md
```diff
@@ -0,0 +1,43 @@
...
+service-to-service communication, the two parties exchange credentials with
+their identity information for mutual authentication purposes.
```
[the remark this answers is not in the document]
rcaballeromx: Those are implementation details. We certainly need to communicate it to our readers when we discuss how to achieve mutual authentication but not as part of the definition.

content/en/docs/reference/glossary/trust-domain.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Trust Domain
+---
+
+A trust domain is a globally unique name that Istio uses to create all
```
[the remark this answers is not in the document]
rcaballeromx: Removed `globally` since the name is unique within a tenancy boundary, as you point out.

PR #4815, opened 2019-08-21
Update docs for locality lb changes

howardjohn:


content/en/docs/tasks/traffic-management/locality-load-balancing/index.md
```diff
@@ -21,8 +21,8 @@ the geographic location where requests are sent.
 
 ## Enabling Locality Load Balancing
 
-This feature is experimental and off by default in 1.1. To enable locality load balancing,
-set the `PILOT_ENABLE_LOCALITY_LOAD_BALANCING` environment variable in all Pilot instances.
+This feature is disabled by default. To disable locality load balancing,
```
frankbu: Is it disabled or enabled by default. If disabled, shouldn't the second sentence say "To enable ..."?

howardjohn: Oops.. Thanks. Should be fixed now, it is enabled by default

PR #4820, opened 2019-08-22
add ambassador API gateway

richarddli:
The Ambassador API Gateway has many customers who deploy it as a N/S solution, integrating with Istio. For details, see https://www.getambassador.io/user-guide/with-istio.


richarddli:
The build log seems to indicate this is a spurious error.

/retest

PR #4854, opened 2019-08-27
Make sure user understands Outlier Detection is req'd for locality LB

ericvn:
I'm not sure if it's being too forceful adding the _TIP_ as well as moving the _Outlier Detection_ rule to the requirements. I'm afraid people will expect traffic to remain local based on the description, and it won't be...


ericvn:
@rcaballeromx Should have the requested changes.

PR #4863, opened 2019-08-28
Update istio 1.2 performance benchmarking docs

carolynhu:
This PR is to change this blog post https://istio.io/blog/2019/performance-best-practices/ to reference the latest istio performance benchmark repo.

### DIFF
--- a/content/en/blog/2019/performance-best-practices/index.md
+++ b/content/en/blog/2019/performance-best-practices/index.md
@@ -2,6 +2,7 @@
 title: "Best Practices: Benchmarking Service Mesh Performance"
 description: "Tools and guidance for evaluating Istio's data plane performance."
 publishdate: 2019-07-09
+last_update: 2019-09-05
 subtitle:
 attribution: Megan O'Keefe (Google), John Howard (Google), Mandar Jog (Google)
 keywords: [performance,scalability,scale,benchmarks]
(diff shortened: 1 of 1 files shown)


carolynhu:
@geeknoid Need approval for this PR. Thanks.

content/en/blog/2019/performance-best-practices/index.md
```diff
@@ -2,6 +2,7 @@
 title: "Best Practices: Benchmarking Service Mesh Performance"
 description: "Tools and guidance for evaluating Istio's data plane performance."
 publishdate: 2019-07-09
+lastupdatedate: 2019-09-05
```
carolynhu: @geeknoid Do you mean add this? I'm not sure which key-name I should use here.

geeknoid: Nope, it's `last_update`.

See <https://preliminary.istio.io/about/contribute/creating-and-editing-pages/#updating-front-matter> for all the front-matter fields supported.

mandarjog:
@carolynhu was the previous commit pointed to be the blog post not working? 
Why do you want to update it ? As long as the blog post is consistent with the version numbers and commits we should not change it.

We are updating the actual data in the concepts/performance doc, that should be sufficient.

carolynhu:
@mandarjog Since recently we updated the README docs, the old README is not working. So I update the commit to the working one. But the results should not update right? Then we do not need the `last_update` label? Since in this PR I only want the reader to have the working README doc to follow up.

PR #4877, opened 2019-08-30
Glossary improvements

frankbu:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/reference/glossary/control-plane.md
```diff
@@ -2,6 +2,6 @@
 title: Control Plane
 ---
 
-A control plane is a set of one or more instances that share the
-same configuration source. The control plane configures the mesh or a subset of
-the mesh to manage the communication between the workload instances within.
+A control plane configures the mesh or a subset of the mesh to manage
```
rcaballeromx: This change tells readers what a control plane does but not what it is. Keep in mind that definitions must follow a specific logical structure:
1. The term (word or phrase) to be defined
2. The class of object or concept to which the term belongs
3. The differentiating characteristics that distinguish it from all others of its class
Source: https://owl.purdue.edu/owl/general_writing/common_writing_assignments/definitions.html

Note that the original text follows that pattern whereas your proposed change doesn't.

content/en/docs/reference/glossary/managed-control-plane.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Managed Control Plane
+---
+
+Many cloud providers offer a [control plane](/docs/reference/glossary/#control-plane),
```
rcaballeromx: Similarly here, please follow the logical pattern of definitions.
A managed control plane is a control plane that cloud ...

content/en/docs/reference/glossary/control-plane.md
frankbu: All I did was change the order of the original two sentences, because the first sentence did not say what it is (its purpose), but rather just one of it's characteristics. It just said `one or more instances ...`. What's an instance, and what does it do? The second sentence explained what its for.

How about this:
```suggestion
A control plane is a service that configures the mesh or a subset of the mesh to manage
```

rcaballeromx: I would strongly disagree with that statement because it's my understanding that the control plane is a `set of processes/instances` and that text would then equate to stating that a service is a `set of processes/instances`, which is highly contentious, AFAIK. @smawson WDYT?

content/en/docs/reference/glossary/service-mesh.md
```diff
@@ -6,6 +6,7 @@ managed, observable and secure communication between
 [workload instances](/docs/reference/glossary/#workload-instance).
 
 Service names combined with a namespace are unique within a mesh. For example,
+in a [multi-cluster](/docs/reference/glossary/#multi-cluster) mesh,
```
geeknoid: Instead, use the {{< gloss >}} feature which gives the user a nice popup instead of a link.

content/en/docs/reference/glossary/control-plane.md
frankbu: I'm very confused. I'm using the term service generically (no glossary reference to Istio Service). It does provide a service with no further implications.

That said, a control plane actually is `a set of services` (processes/instances are the instances).

Anyway, to avoid ambiguity, how about: `is a component that ...`.

Otherwise, what do you suggest? What is a control plane and what's it for?

Wasn't this one of Costin's requested changes too?

content/en/docs/reference/glossary/service-mesh.md
frankbu: I wondered about this actually. So far it seems that we use links from glossary entries to other glossary entries, instead of glossary entries with nested glossary entries. Isn't that better. What happens in the pop-up if a glossary entry has a nested gloss entry?

content/en/docs/reference/glossary/control-plane.md
frankbu: Another suggestion is `a set of control processes ...` or `a control service ...`.

content/en/docs/reference/glossary/service-mesh.md
geeknoid: Oh, I didn't even notice that this was another glossary entry. Yeah, entry-to-entry references should use normal links as popups-in-popups won't work right.

content/en/docs/reference/glossary/control-plane.md
frankbu: After thinking about it more, `a set of system services` would seem to be most correct.

content/en/docs/reference/glossary/managed-control-plane.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Managed Control Plane
+---
+
+A managed control plane is a [control plane](/docs/reference/glossary/#control-plane)
+that many cloud providers offer to manage for their customer.
```
rcaballeromx:
```suggestion
that cloud providers manage for their customers.
```

content/en/docs/reference/glossary/service-mesh.md
rcaballeromx:
```suggestion
In a [multi-cluster](/docs/reference/glossary/#multi-cluster) mesh, for example,
```

content/en/docs/reference/glossary/service-mesh.md
```diff
@@ -6,6 +6,7 @@ managed, observable and secure communication between
 [workload instances](/docs/reference/glossary/#workload-instance).
 
 Service names combined with a namespace are unique within a mesh. For example,
```
rcaballeromx:
```suggestion
Service names combined with a namespace are unique within a mesh.
```

PR #4878, opened 2019-08-31
fix notes

leowucn:
I think this might be an error


leowucn:
ping @sdake @geeknoid

PR #1076, opened 2018-03-15
[0.8] layout and initial content for perf and scalability section

ldemailly:
Fixes #911 

@suryadu 

~This is just the start with the outline, let me know if you think it makes senses and which pages you’d like to author~

PR #4880, opened 2019-09-02
refine NetworkEndpoints fromRegistry field description

styblope:
Description in [Network.NetworkEndpoints field fromRegistry](https://istio.io/docs/reference/config/istio.mesh.v1alpha1/#Network-NetworkEndpoints) is misleading by guiding to use a Kubernetes secret name rather than a kubeconfig filename used as a data key inside the secret.

[x] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


geeknoid:
Hi,

This is an auto-generated file. Please apply the same change in the source content in the https://github.com/istio/api repo instead, and it will get automatically reflected here. Thanks.

styblope:
Thanks for your comments! Re-submitted as #1076

Issue #4889, opened 2019-09-03
Issue with docs/setup/upgrade/steps/index.md


howardjohn:
This page has many issues

* References to upgrading 1.1 -> 1.2 should be 1.2 -> 1.3
* Citadel supports multiple replicas now, so we don't need to say it doesn't
* Should not use some random script to do a rolling restart
* Not sure why we need some special cases for mtls... just link to profiles
* kill all migrating notices

PR #4890, opened 2019-09-03
Update upgrade steps for Istio 1.3

howardjohn:
Fixes https://github.com/istio/istio.io/issues/4889

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/upgrade/steps/index.md
+++ b/content/en/docs/setup/upgrade/steps/index.md
@@ -49,14 +49,14 @@ This is suitable for cases where `kubectl apply` was used to deploy Istio CNI.
     $ NAMESPACE=$(kubectl get pods -l k8s-app=istio-cni-node --all-namespaces --output='jsonpath={.items[0].metadata.namespace}')
     {{< /text >}}
 
-2. If `istio-cni` is currently installed in a namespace other than `kube-system`
+1. If `istio-cni` is currently installed in a namespace other than `kube-system`
    (for example, `istio-system`), delete `istio-cni`:
 
     {{< text bash >}}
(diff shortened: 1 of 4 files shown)


content/en/docs/setup/upgrade/steps/index.md
```diff
@@ -58,14 +49,14 @@ This is suitable for cases where `kubectl apply` was used to deploy Istio CNI.
     $ NAMESPACE=$(kubectl get pods -l k8s-app=istio-cni-node --all-namespaces --output='jsonpath={.items[0].metadata.namespace}')
     {{< /text >}}
 
-1. If `istio-cni` is currently installed in a namespace other than `kube-system`
+2. If `istio-cni` is currently installed in a namespace other than `kube-system`
```
geeknoid: Please leave numbered lists with 1. The right numbers will be inserted automatically when the page is generated.

data/args.yml
```diff
@@ -4,6 +4,9 @@ version: "1.3"
 # The full Istio version identifier the docs describe
 full_version: "1.3.0"
 
+# The previous Istio version identifier the docs describe, used for upgrade documentation
+previous_version: "1.2"
```
geeknoid: In order to reference this symbol from markdown files, you need to add a shortcode definition in layouts/shortcodes.

geeknoid: You should also update the release instructions in README.md to indicate this field needs to be updated when creating a new release.

pitlv2109:
I think `kubectl rollout restart` only works from k8s 1.15 https://github.com/kubernetes/kubernetes/issues/13488#issuecomment-491883922, but I didn't see we mentioned it in the upgrade doc.

Issue #3576, opened 2019-03-08
investigate auto generate of the install option page


linsun:
as part of istio.io build so that we don't need to do it manually.

linsun:
for post 1.1 :)

johnma14:
@linsun Closing this issue for now since @ostromart is working on an alternative way to generate this page.

johnma14:
Reopening this issue since @geeknoid would like to run this script through the CI system.

frankbu:
This has been replaced by operator config.

PR #4913, opened 2019-09-04
Update tablegen.py to be able to be run as part of the CI system

johnma14:
Enable tablegen.py to download the Istio repo for the latest release instead of
relying on one being already installed. Also made the change to automatically
detect release versions in the header section of the installation option page
and update it to reflect the current and previous release versions.

Fixes Bug:#3576

### DIFF
--- a/scripts/tablegen.py
+++ b/scripts/tablegen.py
@@ -16,6 +16,7 @@
 
 import collections
 import linecache
+import requests
 import string
 import sys
 import os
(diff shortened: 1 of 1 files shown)


scripts/tablegen.py
```diff
@@ -36,12 +37,43 @@
...
+    curl_command = "git clone --depth=1 -q -b %s 'https://github.com/istio/istio.git' %s"
+    status = os.system(curl_command %(releaseName, ISTIO_REPO))
+    if status != 0:
+        print("An error occured trying to clone Istio repo for release: %s." %releaseName)
+        exit()
+
+def deleteIstioRepo():
+    os.system("rm -rf %s" %ISTIO_REPO)
+
+def getCurrentReleaseTag():
```
geeknoid: I think this function needs to be replaced with a hardcoded constants for the release identity. Otherwise, when we create a release-1.3 branch of istio.io and try to update the reference docs there, it'll pull the content from the latest release in istio/istio instead of being pinned to the release-1.3 branch.

Basically, we need to have explicit control in istio/istio.io of which branch of istio/istio we pull stuff from. This is why we hardcode the branch names in scripts/grab_reference_docs.sh and I think we need to do the same here.

johnma14: @geeknoid I was trying to automatically get the latest release number for Istio by using the release tag API request: api.github.com/repos/istio/istio/tags gives us the list of tags for istio and from this I try to get the highest value. So for example, from a list of  1.3.0-rc.0 , 1.2.0-rc.0 , 1.1.0-rc.0, etc pick 1.3.0-rc.0 and use 1.3 as the release: git clone --depth=1 -q -b istio-repo 'https://github.com/istio/istio.git' release-1.3. 

I can go ahead and hardcode it if this is not the right thing to do. I was trying to see if there was a way we could avoid hardcoding values. Thanks @geeknoid, appreciate your feedback.

geeknoid: Yeah, I saw that you were figuring out the version automatically, but the problem is basically that "latest" is often not the right choice. Within the release-1.3 branch of istio/istio.io, we want to get the reference state for the release-1.3 branch of istio/istio and not the latest from master.

So I think pinning to specific release versions is necessary.

johnma14: Fixed it so that it we now have a hardcoded release of Istio. Thanks @geeknoid.

geeknoid:
You can fix the linting errors with "make format-python".

geeknoid:
Hmmm. I grabbed this change, and then updated the update_ref_docs target in Makefile.core.mk to:

```
update_ref_docs:
	@scripts/tablegen.py
	@scripts/grab_reference_docs.sh
```

Triggering this target failed with this:

```
~/go/src/istio.io/istio.io$ make update_ref_docs
Traceback (most recent call last):
  File "scripts/tablegen.py", line 19, in <module>
    import requests
ImportError: No module named requests
make[1]: *** [update_ref_docs] Error 1
make: *** [update_ref_docs] Error 2
```

Any idea what's the problem?

PR #4916, opened 2019-09-05
Remove warning about obscure edge case in install

howardjohn:
I don't think we need to document every obscure edge case, especially in the very first page a user will look at. If the user follows our docs, this will never happen. If they *don't* follow the docs, they still will probably never hit this.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/helm/index.md
+++ b/content/en/docs/setup/install/helm/index.md
@@ -57,13 +57,6 @@ $ helm repo add istio.io https://storage.googleapis.com/istio-release/releases/{
 Change directory to the root of the release and then
 choose one of the following two **mutually exclusive** options:
 
-{{< warning >}}
-Istio's installation uses [Helm Package Manager](https://helm.sh) extensively for rendering Istio
-manifests. [Helm Issue 5863](https://github.com/helm/helm/issues/5863) contains details of a problem where
-extra white space in the command line is not properly handled resulting in a `helm template`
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/helm/index.md
```diff
@@ -57,13 +57,6 @@ $ helm repo add istio.io https://storage.googleapis.com/istio-release/releases/{
 Change directory to the root of the release and then
 choose one of the following two **mutually exclusive** options:
 
-{{< warning >}}
```
frankbu: I agree we should remove this from here, but we should add an "Installation Problems" entry in https://preliminary.istio.io/docs/ops/troubleshooting/ that explains this problem.

howardjohn: I disagree. I have never heard of anyone running into this problem except the person who added the note here. We shouldn't document every single problem someone could possibly run into.

To me this is equivalent to saying "if you run `hlem` instead of `helm` the install won't work, so don't make a typo.

If you disagree that's fine though.. not a docs expert

frankbu: Fair enough. If Steve is the one and only person to have ever had this problem, I agree we don't need to document it. We should only document commonly occuring problems.

PR #4917, opened 2019-09-05
Simplify CRD verification step

howardjohn:
I don't think there is a need to insert this edge case about certmanager
here. Lets just look for Istio CRDs, which we know will always be 23,
and not confuse 99% of users for an advanced feature.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/boilerplates/verify-crds.md
+++ b/content/en/boilerplates/verify-crds.md
@@ -1,10 +1,6 @@
 Verify that all `23` Istio CRDs were committed to the Kubernetes api-server using the following command:
 
-{{< warning >}}
-If cert-manager is enabled, then the CRD count will be `28` instead.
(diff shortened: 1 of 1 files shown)


frankbu:
@sdake you added this warning IIRC? You OK to remove it?

frankbu:
Maybe if we remove it, we should add an "Installation Problems" entry in https://preliminary.istio.io/docs/ops/troubleshooting/ that explains the issue?

sdake:
> @sdake you added this warning IIRC? You OK to remove it?

@frankbu looks good to me :)

PR #4928, opened 2019-09-06
Remove port name requirement

howardjohn:
We now do protocol sniffing.

Note - this is definitely not safe to merge. We still need docs explaining protocol sniffing, and how to select a port type explicitly (required for things other than tcp/http, and more performant if you know its tcp/http). Not sure the path forward for this

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/additional-setup/requirements/index.md
+++ b/content/en/docs/setup/additional-setup/requirements/index.md
@@ -13,28 +13,6 @@ keywords: [kubernetes,sidecar,sidecar-injection]
 To be a part of an Istio service mesh, pods and services in a Kubernetes
 cluster must satisfy the following requirements:
 
-- **Named service ports**: Service ports must be named. The port name key/value
-  pairs must have the following syntax: `name: <protocol>[-<suffix>]`. To take
-  advantage of Istio's routing features, replace `<protocol>` with one of the
-  following values:
-
-    - `grpc`
-    - `http`
-    - `http2`
-    - `https`
-    - `mongo`
-    - `mysql`
-    - `redis`
-    - `tcp`
-    - `tls`
-    - `udp`
-
-    For example, `name: http2-foo` or  `name: http` are valid port names, but
-    `name: http2foo` is not. If the port name does not begin with a recognized
-    prefix or if the port is unnamed, traffic on the port is treated as
-    plain TCP traffic unless the port [explicitly](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service)
-    uses `Protocol: UDP` to signify a UDP port.
-
 - **Service association**: A pod must belong to at least one Kubernetes
   service even if the pod does NOT expose any port.
   If a pod belongs to multiple [Kubernetes services](https://kubernetes.io/docs/concepts/services-networking/service/),
--- a/content/en/docs/ops/traffic-management/protocol-selection/index.md
+++ b/content/en/docs/ops/traffic-management/protocol-selection/index.md
@@ -1,0 +1,51 @@
+---
+title: Protocol Selection
+description: Information on how to specify protocols.
+weight: 98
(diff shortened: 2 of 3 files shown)


geeknoid:
Your description says "this is definitely not safe to merge". What do we need to make it safe to merge?

howardjohn:
we need to properly document how to configure port naming/protocol
selection somewhere . I don't think we have anywhere right now

howardjohn:
@geeknoid we should add the below (I just wrote this) to some page in the docs. I am not sure where is appropriate, can you point me in the right direction?



# Protocol Selection

Istio supports proxying all TCP traffic by default, but in order to provide additional capabilities, such as routing and rich metrics, the protocol must be determined. This can be done automatically or explicitly specified.

## Automatic Protocol Selection

By default, Istio will automatically detect HTTP and HTTP/2 traffic. If the protocol cannot automatically be determined, traffic will be treated as plain TCP traffic.

## Manual Protocol Selection

Other protocols must be specified manually by naming the Service port to the protocol. The port name should match either `protocol` or `protocol-suffix`.

The following protocols are supported
* `grpc`
* `http`
* `http2`
* `https`
* `mongo`
* `mysql`\*
* `redis`\*
* `tcp`
* `tls`
* `udp`

Below is an example of a Service that defines a `mysql` port and an `http` port:

```yaml
kind: Service
metadata:
  name: myservice
spec:
  ports:
  - number: 3306
    name: mysql
  - number: 80
    name: http-web
```

\* These protocols are disabled by default to avoid accidentally enabling experimental features. To enable them, Pilot [environment variables](docs/reference/commands/pilot-discovery/#envvars) must be configured

geeknoid:
@frankbu ?

frankbu:
@howardjohn I think your new doc should go in the operations guide [here](https://preliminary.istio.io/docs/ops/traffic-management/).

content/en/docs/setup/additional-setup/requirements/index.md
```diff
@@ -13,28 +13,6 @@ keywords: [kubernetes,sidecar,sidecar-injection]
 To be a part of an Istio service mesh, pods and services in a Kubernetes
 cluster must satisfy the following requirements:
 
-- **Named service ports**: Service ports must be named. The port name key/value
```
frankbu: Instead of just removing this section, maybe it would be good to leave a one liner with a reference to the op guide for more information. For example, something like this:
```
**Named service ports**: For some protocols, service ports must be named to enable certain Istio features. Refer to [Protocol Selection](/docs/ops/traffic-management/protocol-selection) for details.
```

howardjohn: I disagree. This page is for *requirements* for using Istio. The whole point of this feature was to remove one more item from this list.

Defining protocols is no longer a requirement for running Istio - it is merely a requirement to get some extra functionality in *rare* cases (mysql, redis, etc are not even out of Alpha - I would guess 99% of traffic is http or tcp which works out of the box now).

Just like we don't say `specify `--set grafana.enabled=true` to get dashboards, it is not appropriate to specify this here as well.

What do you think?

howardjohn:
> I think your new doc should go in the operations guide

Will send a PR soon

content/en/docs/setup/additional-setup/requirements/index.md
frankbu: Fair enough, although I still wonder if there should be a link from somewhere else. E.g., FAQ entry: "Do service ports need to be named?", with answer, only for certain protocols that are not automatically recognized, See <link>.

howardjohn:
@frankbu thanks for the suggestions. I replaced an outdated FAQ with an updated one

PR #4930, opened 2019-09-06
Error in Secure Gateways (SDS) extra <angle brackets>

esnible:
Fixes

```
mtls-go-example$ ./generate.sh httpbin.example.com <password>
-bash: syntax error near unexpected token `newline'
```

when following instructions blindly.

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/secure-ingress-sds/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/secure-ingress-sds/index.md
@@ -252,7 +252,7 @@ need to create secrets for multiple hosts and update the gateway definitions.
 
     {{< text bash >}}
     $ pushd mtls-go-example
-    $ ./generate.sh httpbin.example.com <password>
+    $ ./generate.sh httpbin.example.com password
     $ mkdir ../httpbin.new.example.com && mv 1_root 2_intermediate 3_application 4_client ../httpbin.new.example.com
     $ popd
     $ kubectl create -n istio-system secret generic httpbin-credential \


content/en/docs/tasks/traffic-management/ingress/secure-ingress-sds/index.md
```diff
@@ -252,7 +252,7 @@ need to create secrets for multiple hosts and update the gateway definitions.
 
     {{< text bash >}}
     $ pushd mtls-go-example
-    $ ./generate.sh httpbin.example.com <password>
+    $ ./generate.sh httpbin.example.com password
```
frankbu: Shouldn't it instead be `<password>` on line 65 and in both places the instructions should tell the user to change `<password>` to some appropriate password? That's similar to how it's done in lots of other docs.

esnible: @frankbu The results of `find . -name "*.md" -exec grep "./generate.sh" \{\} \; -print` are 4 examples with `<password>`, and 8 examples with `password`.

Shall I rework the PR to set them all to `<password>`?

geeknoid: Yes, I think all this needs to be `<password>`

content/en/docs/tasks/traffic-management/ingress/secure-ingress-mount/index.md
```diff
@@ -53,7 +53,7 @@ from the <https://github.com/nicholasjackson/mtls-go-example> repository.
 1.  Generate the certificates for `httpbin.example.com`. Change `password` to any value you like in the following command:
```
frankbu:
```suggestion
1.  Generate the certificates for `httpbin.example.com`. Change `<password>` to any value you like in the following command:
```

content/en/docs/tasks/traffic-management/ingress/secure-ingress-mount/index.md
```diff
@@ -317,7 +317,7 @@ only this time for host `bookinfo.com` instead of `httpbin.example.com`.
 1.  Generate the certificates for `bookinfo.com`. Change `password` to any value you like in the following command:
```
frankbu:
```suggestion
1.  Generate the certificates for `bookinfo.com`. Change `<password>` to any value you like in the following command:
```

Issue #4888, opened 2019-09-03
Issue with docs/setup/upgrade/notice/index.md


carolynhu:
This page https://preliminary.istio.io/docs/setup/upgrade/notice/ should be 1.3 upgrade notice, not 1.2.

rcaballeromx:
Addressed in:
#4931
Add the upgrade notice for Istio 1.3
Opened by rcaballeromx 4 days ago

PR #4908, opened 2019-09-04
add initial 1.3 upgrade notice

yangminzhu:
Please provide a description for what this PR is for.

This is only the initial version of the upgrade notice, more contents could be added later.

For https://github.com/istio/istio/issues/15631

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #4931, opened 2019-09-06
Add the upgrade notice for Istio 1.3

rcaballeromx:
Includes the information operators need to consider when upgrading from Istio 1.2 to 1.3.

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ X ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


rcaballeromx:
The linter fails because the release notes page for 1.3 doesn't exist yet.

rcaballeromx:
Fixes #4888

howardjohn:
See also https://github.com/istio/istio.io/pull/4908 which does the same page but with different content. Also comments on that PR regarding why we have an upgrade notice and release notes that contain overlapping info

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,40 @@
...
+certificates for workload instances more securely.
+
+Kubernetes 1.12 introduces `trustworthy` JWTs to solve these issues. However,
+support for the `aud` field to have a different value than the API server
+audience didn't become available until [Kubernetes 1.13](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.13.md).
+To better secure the mesh, Istio 1.3 only supports `trustworthy` JWTs and
+requires the value of the `aud` field to be `istio-ca` when you enable SDS.
+
+Before upgrading your Istio deployment to 1.3 with SDS enabled, verify that you
+use Kubernetes 1.13 or later.
```
pitlv2109: I wonder should we link the blog post here instead, since it has more information about this change? For example, simply upgrading the k8s cluster to 1.3 doesn't work (it works for GKE only). For OSS k8s, the users will need to do some extra steps. For manged k8s clusters other than GKE, the users need to check with their vendors.

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,40 @@
...
-`--set mixer.templates.useTemplateCRDs=true --set mixer.adapters.useAdapterCRDs=true`
+We simplified the configuration model for Mixer and removed support for
+adapter-specific and template-specific Custom Resource Definitions (CRDs)
+entirely in 1.3. Please move to the new configuration model.
+
+We removed the Mixer CRDs from the system to simplify the configuration
+model, improve Mixer's performance in Kubernetes deployments, and improve
+reliability in various Kubernetes environments.
+
+## Secret Service Discovery
```
howardjohn:
```suggestion
## Secret Discovery Service
```

rlenglet:
cc @sdake @brian-avery

rcaballeromx:
@howardjohn We created this page to help folks found the information that is only relevant for those that are upgrading from the previous release to the new one. The release notes include a lot more information than that and some overlap is inevitable. There is value in providing users that don't need or care about all the changes in the release but care deeply about the important information that they should consider when upgrading.

howardjohn:
After https://github.com/istio/istio.io/pull/4934/files removed all of the warnings for upgrade, we need to add more to this:

* Capture all ports by default. Implication is that services that previous bypassed envoy no longer will. If you are intentionally listening on `127.0.0.1` to avoid exposing your service, or intentionally bypassing Envoy for performance reasons, you must opt out with `traffic.sidecar.istio.io/excludeInboundPorts`
* Protocol sniffing is enable by default. If you do not want protocol sniffing you should disable it with `--set pilot.enableProtocolSniffing=false`

rcaballeromx:
@howardjohn I've added the information you provided. Thanks! I rewrote the text for clarity. PTAL

howardjohn:
Thank!

> Istio now captures all ports by default. Thus, services can't bypass the Envoy
proxies even if they previously did. If your services listen on `127.0.0.1` to
avoid exposure or to intentionally bypass Envoy to improve performance

there are two different cases - you don't specify container port because your application should not be exposed, or you don't specify container port to intentionally bypass for performance/other reasons (maybe don't say performance as it makes it sounds bad .. just "intentionally bypass envoy"?)

I'm adding some docs on the protocol sniffing as well so we can point to that here once it's merged

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,62 @@
...
+
+We simplified the configuration model for Mixer and removed support for
+adapter-specific and template-specific Custom Resource Definitions (CRDs)
+entirely in 1.3. Please move to the new configuration model.
+
+We removed the Mixer CRDs from the system to simplify the configuration
+model, improve Mixer's performance in Kubernetes deployments, and improve
+reliability in various Kubernetes environments.
+
+Istio now captures all ports by default. Thus, services can't bypass the Envoy
```
rlenglet: > services can't bypass the Envoy proxies even if they previously did

This statement is wrong. Containers can still bypass the Envoy proxies. We still can't prevent that.

rlenglet: I would just remove that statement. The rest of this paragraph stands on its own.

rcaballeromx:
@rlenglet @howardjohn I've modified the text based on your feedback, PTAL

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,66 @@
...
+
+We simplified the configuration model for Mixer and removed support for
+adapter-specific and template-specific Custom Resource Definitions (CRDs)
+entirely in 1.3. Please move to the new configuration model.
+
+We removed the Mixer CRDs from the system to simplify the configuration
+model, improve Mixer's performance in Kubernetes deployments, and improve
+reliability in various Kubernetes environments.
+
+Istio now captures all ports by default. If your services listen on `127.0.0.1`,
```
howardjohn: These should be under a different heading than installation and upgrade. L28 - L42. Not sure how we want to section them, could be Traffic Management or give each its own section?

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,66 @@
...
+you probably don't specify container ports due to the following reasons:
```
howardjohn: I still don't think this completely  captures the change. How about something like:

> Istio now captures all ports by default. If you intentionally did not specify container ports to bypass Envoy, you must opt out of port capturing with the `traffic.sidecar.istio.io/excludeInboundPorts` option. Note that this change means any applications listening on `127.0.0.1` will be exposed, as Envoy connects over localhost.

The reason I am being picky here is this could bite someone really hard if they accidentally expose a private endpoint to the public accidentally with this update, or if they are using some low-latency application

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,66 @@
...
+
+- To avoid exposing your service.
+- To intentionally bypass Envoy.
+
+If that is the case, you must opt out of port capturing with the
+`traffic.sidecar.istio.io/excludeInboundPorts` option.
+
+Protocol sniffing is now enabled by default. Disable port sniffing with the
+`--set pilot.enableProtocolSniffing=false` option when you upgrade to get the
```
howardjohn: todo - once https://github.com/istio/istio.io/pull/4928 is merged, link to the doc.

rcaballeromx:
@rcaballeromx @rlenglet I've addressed the comments and rebased the PR to prepare it for the release. I think we are good to go but PTAL

content/en/docs/setup/upgrade/notice/index.md
howardjohn: This is still not resolved. `Intentionally bypassing Envoy` and `listen on 127.0.0.1 to avoid exposing your service` are two distinct items.  

Suggest:

> Istio now captures all ports by default. If your don't specify container ports intentionally, to avoid exposing services listening on `127.0.0.1` or to intentionally bypass Envoy, you must ...

howardjohn: Actually:

> Istio now captures all ports by default. If your don't specify container ports to intentionally bypass Envoy, you must ...


The warning below captures the 127.0.0.1 stuff

content/en/docs/setup/upgrade/notice/index.md
howardjohn: Can you link to https://preliminary.istio.io/docs/ops/traffic-management/protocol-selection/

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,73 @@
...
+template before upgrading to Istio 1.3:
+`--set pilot.env.PILOT_SKIP_VALIDATE_TRUST_DOMAIN=true`
+
+## Secret Discovery Service
+
+In Istio 1.3, we are taking advantage of improvements in Kubernetes to issue
+certificates for workload instances more securely.
+
+Kubernetes 1.12 introduces `trustworthy` JWTs to solve these issues. However,
+support for the `aud` field to have a different value than the API server
```
sdake: this sentence reads poorly.  `aud` is introduced in 67 and on line 68 audience is introduced yet no real connection is m ade between the two.  some amount of brainpower is required to understand they are related. it really eels like a period is missing at the end of 67 and 68 should start with "The `aud` field represents the audience in Kubernetes and didn't become available...

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,73 @@
...
+
+If that is the case, you must opt out of port capturing with the
+`traffic.sidecar.istio.io/excludeInboundPorts` option.
+
+{{< warning >}}
+This change exposes any applications listening on `127.0.0.1`
+since Envoy connects over `localhost`. Opt out to avoid exposing such applications.
+{{< /warning >}}
+
+Protocol sniffing is now enabled by default. Disable port sniffing with the
```
sdake: Disable protocol sniffing...

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,73 @@
...
+Istio now captures all ports by default. If your services listen on `127.0.0.1`,
+you probably don't specify container ports due to the following reasons:
+
+- To avoid exposing your service.
+- To intentionally bypass Envoy.
+
+If that is the case, you must opt out of port capturing with the
+`traffic.sidecar.istio.io/excludeInboundPorts` option.
+
+{{< warning >}}
```
sdake: many applications bind to the wildcard `0.0.0.0` which means they are binding to localhost.  Some operators have no idea their applications are wildcard binding.

This warning leaves me scratching my head "Ok so no protocol sniffing for 90% of the applications that wildcard bind by default.  Some applications have no way *NOT* to wildcard bind".  I'd be a little more explicit in this fact up front to help clarify this.  The fact that Envoy connects over `localhost` is really irrelevant... :)

howardjohn: 1. This feature is completely different than protocol sniffing, it is about capturing all inbound traffic by default. Before only container Ports were captured.
2. Envoy connecting over localhost is not really irrelevant. If you bind to 127.0.0.1 and some other pod calls you, the connection is refused. Now you add Istio and all the sudden the calls work.

I really want to avoid someone accidentally exposing a sensitive endpoint so if it's not clear we should refine this

rcaballeromx: Ok, I'm struggling because the underlaying problem is the behavior. The current text at least warns folks they should be aware of the possibility that their endpoints could be exposed and how to address it. I'd suggest we keep the warning but maybe we can use some better wording. @nmittler Can you help us find the right way to phrase this?

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,75 @@
...
+
+## Secret Discovery Service
+
+In Istio 1.3, we are taking advantage of improvements in Kubernetes to issue
+certificates for workload instances more securely.
+
+Kubernetes 1.12 introduces `trustworthy` JWTs to solve these issues.
+[Kubernetes 1.13](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.13.md)
+introduced the ability to change the value of the `aud` field to a value other
+than the API server. The `aud` field represents the audience in Kubernetes. To
```
pitlv2109: This sentence "The `aud` field represents the audience in Kubernetes" is not technically correct, and I think you can remove it. If not, you can change it to something like "The `aud` field represents the audience of the trustworthy JWTs, which is the Istio CA"

content/en/docs/setup/upgrade/notice/index.md
nmittler: I chatted with @howardjohn, who has more background on the changes.  His suggested text:

```
Istio now captures all ports by default. If your don't specify container ports to intentionally
bypass Envoy, you must opt out of port capturing with the 
`traffic.sidecar.istio.io/excludeInboundPorts` option.

Warning: This change exposes any applications listening on `127.0.0.1` since Envoy 
connects over localhost. Opt out to avoid exposing such applications.
```

content/en/docs/setup/upgrade/notice/index.md
pitlv2109: cc @myidpt

content/en/docs/setup/upgrade/notice/index.md
```diff
@@ -1,44 +1,70 @@
...
+{{< /warning >}}
+
+Protocol sniffing is now enabled by default. Disable protocol sniffing with the
+`--set pilot.enableProtocolSniffing=false` option when you upgrade to get the
+previous behavior. To learn more see our [protocol selection page](/docs/ops/traffic-management/protocol-selection/).
+
+To specify a hostname in multiple namespaces, you must select a single host using
+a [`Sidecar` resource](/docs/reference/config/networking/v1alpha3/sidecar/).
+
+## Trust domain validation
```
yangminzhu: Thanks for integrating this, could also integrate the changes in `docs/tasks/security/authn-policy/index.md` from https://github.com/istio/istio.io/pull/4908 to this PR? I think we don't need a separate PR.

cc @myidpt

PR #4948, opened 2019-09-10
Refresh docs to include IstioControlPlane proto

ostromart:
Note that the proto was generated from master rather than 1.3 since https://github.com/istio/operator/pull/268 has not yet been cherry-picked into release-1.3 but this should happen in the next couple of days.
The proto is essentially the same with some comment decorations being the major difference.


ostromart:
I'm not sure what to do about the linter error blocking this @geeknoid:

- ./public/zh/docs/concepts/performance-and-scalability/index.html
  *  External link https://raw.githubusercontent.com/istio/istio/master/bin/perfcheck.sh failed: 404 No error
htmlproofer 3.9.2 | Error:  HTML-Proofer found 1 failure!
LINTING FAILED

geeknoid:
That link is fixed. Rebase and repush, it should go away...

PR #4949, opened 2019-09-10
Disable support for the Chinese translation since it has fallen out of date.

geeknoid:


linsun:
+1 it was last updated in April-ish 2019.

PR #4952, opened 2019-09-10
Add support for text block links that point to other repos.

geeknoid:


data/args.yml
```diff
@@ -25,7 +25,7 @@ archive_date: YYYY-MM-DD
 archive_search_refinement: "V1.1"
 
 # GitHub branch names used when the docs have links to GitHub
-source_branch_name: master
+source_branch_name: release-1.3
```
sdake: is this right for master?

geeknoid: Yes, I should've changed that the day the 1.3 branches were created in other repos. Once I create the 1.3 branch in istio.io, then this value will be returned to master in the master branch.

I do the same dance in the scripts/grab_reference_docs.sh script.

Issue #4915, opened 2019-09-04
Update with 1.3 results docs/concepts/performance-and-scalability/index.md


carolynhu:
Update this page with the latest 1.3 results and also update the corresponding CN page.

PR #4956, opened 2019-09-11
Update Performance and Scalability page for istio 1.3

carolynhu:
This PR is to update istio 1.3 docs for `Performance and Scalability` page. issue: https://github.com/istio/istio.io/issues/4915


carolynhu:
I‘m wondering if we can only rendering .svg file here? Should I convert .png to .svg?

content/en/docs/concepts/performance-and-scalability/index.md
```diff
@@ -101,20 +101,20 @@ immediately. This process adds to the queue wait time of the next request and af
...
 Inside the mesh, a request traverses the client-side proxy and then the server-side
-proxy. This two proxies on the data path add about 8ms to the 90th percentile latency at 1000 requests per second.
-The server-side proxy alone adds 2ms to the 90th percentile latency.
+proxy. This two proxies on the data path add about 7ms to the 90th percentile latency at 1000 requests per second.
+The server-side proxy alone adds 3ms to the 90th percentile latency.
 
 ### Latency for Istio {{< istio_release_name >}}
 
-The default configuration of Istio 1.1 adds 8ms to the 90th percentile latency of the data plane over the baseline.
+The default configuration of Istio 1.3 adds 7ms to the 90th percentile latency of the data plane over the baseline.
```
mandarjog: use `< istio_release_name >` here

content/en/docs/concepts/performance-and-scalability/index.md
```diff
@@ -101,20 +101,20 @@ immediately. This process adds to the queue wait time of the next request and af
 average and tail latencies. The actual tail latency depends on the traffic pattern.
 
 Inside the mesh, a request traverses the client-side proxy and then the server-side
-proxy. This two proxies on the data path add about 8ms to the 90th percentile latency at 1000 requests per second.
-The server-side proxy alone adds 2ms to the 90th percentile latency.
+proxy. This two proxies on the data path add about 7ms to the 90th percentile latency at 1000 requests per second.
```
rshriram: These two..

content/en/docs/concepts/performance-and-scalability/index.md
```diff
@@ -101,20 +101,20 @@ immediately. This process adds to the queue wait time of the next request and af
...
+The server-side proxy alone adds 3ms to the 90th percentile latency.
```
rshriram: these numbers seem suspicious. Are you sure this is not standard variance? 7ms for end to end, where 3ms goes to server, leaving 4ms for the client including the network.
compared to previous one of 8ms end to end, where 2ms goes to server, leaving 6ms for the client including the network.

carolynhu:
@rshriram here is the graph, I generated today:
latency_p90

content/en/docs/concepts/performance-and-scalability/index.md
carolynhu: Thanks for your comments, I need to confirm with @mandarjog about the testing data.

howardjohn:
@geeknoid any help on the lint error? I couldn't grok it

content/en/docs/concepts/performance-and-scalability/index.md
carolynhu: @rshriram Thanks for your concern, checked again with the data, this serverside proxy alone should adds around 2ms. which is calculated by 3.1-0.82.

geeknoid:
The error from deploy/netlify is:

10:03:55 AM: Error: Error building site: "/opt/build/repo/content/en/docs/concepts/performance-and-scalability/index.md:116:1": failed to render shortcode "image": failed to process shortcode: "/opt/build/repo/layouts/shortcodes/image.html:91:22": execute of template failed: template: shortcodes/image.html:91:22: executing "shortcodes/image.html" at <float (index $sizes 2)>: error calling float: unable to cast <nil> of type <nil> to float64

This is my Hugo shortcode crapping out while processing an image, it didn't find what is was needed for some reason. Back momentarily...

carolynhu:
@geeknoid @howardjohn When I change .svg to .png the lint check is passing. And it can successfully rendered in my local server.

geeknoid:
The linting error is due to the fact both SVG files don't contain a viewBox attribute in the initial XML element, which is highly unusual for an SVG file. The logic I have for this in istio.io/layouts/shortcodes/image.html is supposed to recover from not finding a viewBox element, but it's getting confused for some reason, thinks it found the attribute and then crashes.

The quick solution is to add a viewBox element. It can be mechanically derived from the width and height. See all the other SVGs in the site for an example. Meanwhile, I'll try to fix the shortcode.

geeknoid:
Please keep using an SVG, it's generally much better for our users than PNG.

carolynhu:
@geeknoid Now this svg works.

geeknoid:
@carolynhu I've pushed a fix to the repo for this problem with viewBox.

PR #4959, opened 2019-09-11
fix port naming section

rshriram:
We seemed to have removed the port naming completely. Its generally a good practice, atleast for non HTTP protocols.


frankbu:
This information has been moved to the [op guide](https://preliminary.istio.io/docs/ops/traffic-management/protocol-selection/). If we want to leave something in the "requirements guide" it should just be a short one-liner that points to the op guide. @howardjohn was opposed to leaving anything in the "requirements guide", when I suggested it previously: https://github.com/istio/istio.io/pull/4928#discussion_r322307690

rshriram:
Oh that op guide page is neat! I think this PR is then redundant

PR #5023, opened 2019-09-19
dark-mode-css-bug

rvennam:
Some characters ( <, | ) appear highlighted in dark mode:
image

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/src/sass/themes/_light-theme-syntax.scss
+++ b/src/sass/themes/_light-theme-syntax.scss
@@ -45,7 +45,6 @@
 .language-css .token.string,
 .style .token.string {
     color: #9a6e3a;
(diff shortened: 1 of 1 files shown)


geeknoid:
I've been meaning to track this down for a while. 

Instead of fixing the dark theme by changing the light theme, I think it would make sense to change that same property in the _dark-theme-syntax.css. WDYT?

rvennam:
I don't believe this property is needed in light theme because of background color already matches.

PR #5026, opened 2019-09-19
update doc for http prob annotation

linsun:

[x ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/ops/app-health-check/index.md
+++ b/content/en/docs/ops/app-health-check/index.md
@@ -121,7 +121,7 @@ The configuration changes above (by Helm or by the configuration map) effect all
 
 <!-- Add samples YAML or kubectl patch? -->
 
-Rather than install Istio with different Helm option, you can annotate Pod with `sidecar.istio.io/rewriteAppHTTPProbers: "true"` under `spec.template.metadata.annotations`.  If you annotate under `metadata.annotations`, the annotation is ignored.
+Rather than install Istio with different Helm option, you can annotate the [Pod object](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/) with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.  If you annotate the other [Kubernetes objects](https://kubernetes.io/docs/concepts/#kubernetes-objects), the annotation will be ignored.
 
 {{< text yaml >}}
 apiVersion: extensions/v1beta1
--- a/content/en/docs/ops/app-health-check/index.md
+++ b/content/en/docs/ops/app-health-check/index.md
@@ -121,7 +121,7 @@ The configuration changes above (by Helm or by the configuration map) effect all
 
 <!-- Add samples YAML or kubectl patch? -->
 
-Rather than install Istio with different Helm option, you can annotate Pod with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.
+Rather than install Istio with different Helm option, you can annotate Pod with `sidecar.istio.io/rewriteAppHTTPProbers: "true"` under `spec.template.metadata.annotations`.  If you annotate under `metadata.annotations`, the annotation is ignored.
 
 {{< text yaml >}}
 apiVersion: extensions/v1beta1
--- a/content/en/docs/ops/app-health-check/index.md
+++ b/content/en/docs/ops/app-health-check/index.md
@@ -121,7 +121,7 @@ The configuration changes above (by Helm or by the configuration map) effect all
 
 <!-- Add samples YAML or kubectl patch? -->
 
-Rather than install Istio with different Helm option, you can annotate the [Pod object](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/) with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.  If you annotate the other [Kubernetes objects](https://kubernetes.io/docs/concepts/#kubernetes-objects), the annotation will be ignored.
+Rather than install Istio with different Helm option, you can annotate the [Pod object](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/) with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.  If you annotate any other [Kubernetes objects](https://kubernetes.io/docs/concepts/#kubernetes-objects), the annotation will be ignored.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/app-health-check/index.md
```diff
@@ -121,7 +121,7 @@ The configuration changes above (by Helm or by the configuration map) effect all
 
 <!-- Add samples YAML or kubectl patch? -->
 
-Rather than install Istio with different Helm option, you can annotate Pod with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.
+Rather than install Istio with different Helm option, you can annotate Pod with `sidecar.istio.io/rewriteAppHTTPProbers: "true"` under `spec.template.metadata.annotations`.  If you annotate under `metadata.annotations`, the annotation is ignored.
```
incfly: This wording assumes our customers use `Deployment`, which maybe 90% true for most cases. However, as I pointed out in the issue, it's possible other resources built on top of the Pod, like `ReplicaSet` etc. So it's not 100% accurate.

sidecar injector guru, @liamawhite  @ayj thoughts?

incfly: Hmm, I think it'll be fine if you mention "For `Deployment`, annotations should be in ....`.

frankbu: This clarification and warning seems inappropriate here. It's a general comment about how to add pod annotations to a deployment. Maybe there's some Kubernetes doc that this could point to: `you can [annotate the pod](link) with ...`

linsun:
@incfly @frankbu thanks for the feedback, I tried to address it PTAL

content/en/docs/ops/app-health-check/index.md
linsun: k8s has docs for what is k8s objects (which includes pod), and annotate k8s objects so I tried to link those.

geeknoid:
Lin, can you address remaining comments so we can merge this?

Thanks.

frankbu:
@linsun, @geeknoid I'd prefer that we close this PR. It currently says that you should "annotate a pod", which seem sufficient. I don't think it's appropriate to explain how to annotate a pod or to say that if you annotate something else it, it won't work - that goes without saying.

linsun:
@geeknoid I've addressed your feedback and others.

I would prefer the clarification in as I have hit this issue.  i believe others have too.

Longer term, I'm also working with @esnible on how we can flag this earlier through istioctl or other validator.

content/en/docs/ops/app-health-check/index.md
```diff
@@ -121,7 +121,7 @@ The configuration changes above (by Helm or by the configuration map) effect all
 
 <!-- Add samples YAML or kubectl patch? -->
 
-Rather than install Istio with different Helm option, you can annotate Pod with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.
+Rather than install Istio with different Helm option, you can annotate the [Pod object](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/) with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.  If you annotate any other [Kubernetes objects](https://kubernetes.io/docs/concepts/#kubernetes-objects), the annotation will be ignored.
```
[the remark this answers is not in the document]
linsun: updated

frankbu:
If this is really something that is tripping people up, I think we should add a task or example that shows how to do it. Then, we could even add a link to the example from here if we want to make sure people find it.

frankbu:
Another possibility is to put more information and the warning about which resource the various annotations belong to on this page: https://preliminary.istio.io/docs/reference/config/annotations.html. Maybe even include a couple of examples? This would handle the possible confusion generically, instead of only explaining it for one particular case.

linsun:
@frankbu yes, I thought of that too (e.g. put in a generic place).  I can rework it... I think the list here (https://preliminary.istio.io/docs/reference/config/annotations.html) all applies to pod annotation.

frankbu:
@linsun, most but not all of the annotations in https://preliminary.istio.io/docs/reference/config/annotations.html are for pod. The first one is for `Ingress`, the second one `Service`. I think we need to change the generator to add another table column (e.g., `Resource`), or the descriptions of all the annotations should be very explicit about where they apply.

^^^ @geeknoid WDYT?

linsun:
@geeknoid +1 to what @frankbu said, we need a column to indicate scope of each annotation.

geeknoid:
Makes sense to add a column there. I'll work on that.

geeknoid:
Here's the generator update: https://github.com/istio/tools/pull/419

@nmittler will add the necessary metadata to annotations.yaml.

geeknoid:
@linsun @frankbu Is this ready to merge? Does it need to be cherry picked to master too?

content/en/docs/ops/app-health-check/index.md
```diff
@@ -121,7 +121,7 @@ The configuration changes above (by Helm or by the configuration map) effect all
 
 <!-- Add samples YAML or kubectl patch? -->
 
-Rather than install Istio with different Helm option, you can annotate Pod with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.
+Rather than install Istio with different Helm options, you can annotate the [Pod object](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/) with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`.  If you annotate any other [Kubernetes objects](https://kubernetes.io/docs/concepts/#kubernetes-objects), the annotation will be ignored.
```
frankbu:
```suggestion
Rather than install Istio with different Helm options, you can [annotate the pod](/docs/reference/config/annotations/) with `sidecar.istio.io/rewriteAppHTTPProbers: "true"`. Make sure you add the annotation to the [pod resource](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/) because it will be ignored anywhere else (for example, on an enclosing deployment resource).
```

linsun:
thank you @frankbu for the suggestion.  I've added label to merge to master @geeknoid

PR #5033, opened 2019-09-20
constraint key incorrect with destination.labels

kenthua:
The way the constraint is defined, it appears it should be `source.labels[app]` instead of `destination.labels[app]`.  Since the rule is defined as `services: ["*"]` the example works as is, but can be confusing.

[X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/security/authz-http/index.md
+++ b/content/en/docs/tasks/security/authz-http/index.md
@@ -84,7 +84,7 @@ the services must have one of the listed `app` labels.
       - services: ["*"]
         methods: ["GET"]
         constraints:
-        - key: "destination.labels[app]"
+        - key: "source.labels[app]"
           values: ["productpage", "details", "reviews", "ratings"]
     {{< /text >}}
 


incfly:
@liminw could you take a look?

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -84,7 +84,7 @@ the services must have one of the listed `app` labels.
       - services: ["*"]
         methods: ["GET"]
         constraints:
-        - key: "destination.labels[app]"
+        - key: "source.labels[app]"
```
yangminzhu: Could you clarify why make this change? This doesn't match the description above: 
```
allows read access to any service in the default namespace that has the app label set to one of the values productpage, details, reviews, or ratings.
```

kenthua: Isn't that what the constraint is defining?  If I have the constraint that says a pod with the source label app = productpage, details, reviews, or ratings, allow access to the the downstream service, in this case `[*]` in the `default` namespace.  

Unless I'm understanding the path flow wrong `services: [*]` defines the service I want to protect and the constraint helps me define which service I want to allow access.

I tried to re-create this for one my scenarios of protecting Service A from Service B and using destination label didn't work until I switched it to source label.

yangminzhu: the constraint is actually used to further filter the services you want to protect. For example, you have `services: [*]` and constraints `destination.labels[app]: [dev]`, that means only the service with label `app: dev` is to be protected by the RBAC, another service that doesn't have label `app: dev` will not be protected by RBAC in this case.

Also note that the `source.labels` should only be used in the Properties: https://istio.io/docs/reference/config/authorization/constraints-and-properties/#supported-properties

kenthua: In the `ServiceRole`, isn't for granting access.  i.e. I have `service: ["serviceb"]`, but only a service with label `app:servicea` would be able to invoke it.

The definition of what services to be protected are `ClusterRbacConfig`, with `ON_WITH_INCLUSION`.   `services: ["serviceb.default.svc.cluster.local"]`

Unless I'm misusing the purpose.  For instance I used `destination.labels` for my use case, but it didn't work until I switched to `source.labels`

geeknoid:
@yangminzhu @liminw What do you think? Should we merge this PR?

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: `ClusterRbacConfig` is used to enable RBAC on a set of namespaces or services and ServiceRole is used to define the detailed RBAC policy on the namespace/service.

let's say you have svcA, svcB and svcC, and you have ClusterRbacConfig on ["SvcA", "SvcB"], this means if you don't define any ServiceRole/Binding for SvcA/SvcB, the request will be rejected. Here SvcC will always be accessible.

Now let's say you have ServiceRole with `services: [*]` with constrainst `destination.labels[app]: A`, this means it will only affect SvcA as the constraint is only matched with SvcA, the results is the request to SvcB will always be rejected because there is no policy selecting it.

As for constraint `source.labels`, it's not an supporter attribute, see https://istio.io/docs/reference/config/authorization/constraints-and-properties/#supported-properties

There might be something else wrong that makes you think it somehow worked in your case, would you mind to provide more context of your use case?

yangminzhu:
@geeknoid I think we shouldn't merge this change, the `source.label` is even not supported by RBAC.

kenthua:
I understand it's not supported.  If I'm using or understanding it incorrectly, please correct me.  Otherwise it may be confusing for others.

Thanks!

frankbu:
@kenthua I think your confusion stems from a misinterpretation of this sentence describing the example:
```
allows read access to any service in the default namespace that has the app label set to one of
the values productpage, details, reviews, or ratings.
```
"to any service" is somewhat ambiguous here. I think you are reading it as "allows read access to any client service ...". What it is saying is "allows read access to any target service ...":

Maybe the wording should be clarified? For example:
```
allows any service in the default namespace that has the app label set to one of the values
productpage, details, reviews, or ratings, to be accessed with a GET request.
```

yangminzhu:
Thanks @frankbu for the explanation, @kenthua Please me know if that answers your question. Also please be noted that we're making a lot of changes to the documentation for the new v1beta1 authz policy introduced in 1.4, the current page will be rewritten completely in 1.4.

kenthua:
Yes, thanks @frankbu for the explanation.  As for my issue, I am able to get it work with `destination.labels[app]` if I specify `services: ["*"]` but not if I specify `services: ["productpage"]` or `services: ["productpage.default.svc.cluster.local"]` in the `ServiceRole`.  

I don't have issues with the `ServiceRole/ServiceRoleBinding` for the `productpage` itself.  Just `productpage` -> `details`

Essentially I want fine grained service to service rbac.

ex:

```
apiVersion: "rbac.istio.io/v1alpha1"
kind: ClusterRbacConfig
metadata:
  name: default
spec:
  mode: "ON_WITH_INCLUSION"
  inclusion:
    services: ["productpage.default.svc.cluster.local", "details.default.svc.cluster.local"]
```

This service role does not work.
```
apiVersion: "rbac.istio.io/v1alpha1"
kind: ServiceRole
metadata:
  name: details-viewer
spec:
  rules:
  - services: ["productpage"]
    methods: ["GET"]
    constraints:
    - key: "destination.labels[app]"
      values: ["details"]
```
```
apiVersion: "rbac.istio.io/v1alpha1"
kind: ServiceRoleBinding
metadata:
  name: details-viewer-binding
spec:
  subjects:
  - properties:
      source.namespace: "istio-system"
  - properties:
      source.namespace: "default"
  roleRef:
    kind: ServiceRole
    name: "details-viewer"
```

adammil2000:
@yangminzhu @kenthua Hi all, can you confirm if this PR is still factually accurate and ready for the documentation? Thanks!

yangminzhu:
The service role
```
apiVersion: "rbac.istio.io/v1alpha1"
kind: ServiceRole
metadata:
  name: details-viewer
spec:
  rules:
  - services: ["productpage"]
    methods: ["GET"]
    constraints:
    - key: "destination.labels[app]"
      values: ["details"]
```

has 2 issues:
1. It should use full name in the services field, i.e. `productpage.default.svc.cluster.local`
2. The policy seems wrong because the service is for productpage but at the same time it has  `destination.labels[app]` with value `details`. This means it matches nothing because the productpage service should only have label "app=productpage". Note, for a ServicRole, the `services` field is used first to filter out all the services and the `destination.labels` attribute is used against on the filtered services.

What access control effect you want to achieve here? Do you want to enforce access control on both `details` and `productpage` service? If so, you can use the following policies:
```
apiVersion: "rbac.istio.io/v1alpha1"
kind: ClusterRbacConfig
metadata:
  name: default
spec:
  mode: "ON_WITH_INCLUSION"
  inclusion:
    services: ["productpage.default.svc.cluster.local", "details.default.svc.cluster.local"]
---
apiVersion: "rbac.istio.io/v1alpha1"
kind: ServiceRole
metadata:
  name: details-viewer
spec:
  rules:
  - services: ["productpage.default.svc.cluster.local", "details.default.svc.cluster.local"]
    methods: ["GET"]
---
apiVersion: "rbac.istio.io/v1alpha1"
kind: ServiceRoleBinding
metadata:
  name: details-viewer-binding
spec:
  subjects:
  - properties:
      source.namespace: "istio-system"
  - properties:
      source.namespace: "default"
  roleRef:
    kind: ServiceRole
    name: "details-viewer"
```

kenthua:
@yangminzhu thanks for the reply. 

> Do you want to enforce access control on both details and productpage service?

Yes.  Specifically only allow productpage to call details, no other services.



We can close this with the introduction of [Authorization Policy](https://istio.io/docs/reference/config/security/authorization-policy/) and deprecation of RBAC

PR #5060, opened 2019-09-26
Introduce top-level News section.

geeknoid:
This introduces a top-level news section to the site. All timely blog posts are moved into this section, making the blog post a pure technical set of docs. Having a dedicated news section also means we don't really need a stand-alone release note section, so that's gone.


frankbu:
I love it.

Should posts like these also be moved to News?
https://deploy-preview-5060--preliminary-istio.netlify.com/blog/2019/incorrect-sidecar-image-1.2.4/
https://deploy-preview-5060--preliminary-istio.netlify.com/blog/2019/istio-security-003-004/

geeknoid:
Yeah, those posts should be moved to the news section too.

I forgot to mention that the site now exposes 3 RSS feeds:

/news/feed.xml which exposes only the news items
/blog/feed.xml which exposes only the blog items
/feed.xml which is the same content as before and exposes both blog and news items

geeknoid:
@frankbu OK, I'll try to finish this up tonight then.

howardjohn:
I like it, agree about moving the security stuff to news though.

Nice!

nmittler:
Looks great! Awesome work, @geeknoid!

geeknoid:
All right, all done.

PTAL.

frankbu:
This is a huge improvement, thanks for doing it @geeknoid !

The only thing that I'm wondering about is [this entry](https://deploy-preview-5060--preliminary-istio.netlify.com/blog/2019/sail-the-blog/). It seems like more of a news (announce) entry, then a blog post, but since it's not really even correct now, maybe we should just remove it. But, more importantly, I think we need to update [the guidelines](https://deploy-preview-5060--preliminary-istio.netlify.com/about/contribute/creating-and-editing-pages/#choosing-a-page-type) that it is referring to, to update the criteria for Blog Posts, and also add something for News Posts. WDYT?

geeknoid:
I've updated the blog post and content guidelines.

PTAL.

PR #5073, opened 2019-09-27
Remove usages of curl inside istio-proxy

howardjohn:
Distroless builds of Istio do not contain curl, so we should not tell
users to use it. Pilot-agent handles this functionality for us

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


geeknoid:
One CI error...

I presume this should be cherry picked to the 1.3 branch?

howardjohn:
It can go in 1.3, but more important for 1.4. We may make distroless the default in 1.4 though in which case all of this would be broken -- right now its only broken if they use distroless which is probably a small % of people

PR #5094, opened 2019-10-02
Use same bookinfo version on both clusters of MC example

### LINKED ISSUE #5093: Issue trying to use GKE multi-cluster example with shared control plane setup.
Setup the following example on two IKS clusters:

https://istio.io/docs/setup/install/multicluster/shared-gateways/

This works as expected.

Tried to use the example from:

https://istio.io/docs/examples/multicluster/gke/

on top of the prior install. It does look like the Istio install in the later example matches what is done in the shared-gateways page, with the exception that it calls out GKE clusters to high I don't have access.

The example does not work as expected. One should see the product page load balance through the 3 reviews yielding no/black/red stars, however the red stars never appear (reviews-v3 is running on the second cluster).

I'm not sure the second example would actually run on GKE.

I was able to get the red stars to show up if I did two things:

- Update the reviews-v3.yaml to be more like the bookinfo.yaml (I took the review and ratings sections from bookinfo, removed the ratings, reviews-v1, and reviews-v2 deployments).
- Force an update by applying the default destination rules and removing them. Will the rules were active the app failed to run.

It may be that the former is all that's needed for GKE, but have doubts. I opened https://github.com/istio/istio/issues/17562 to understand why the second bullet is needed.

A third problem is that applying the destination rules causes the app to fail. I'm not sure why that is at the moment.

frankbu:
Fixes #5093 

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/examples/multicluster/gke/index.md
+++ b/content/en/docs/examples/multicluster/gke/index.md
@@ -237,6 +237,7 @@ $ kubectl label secret ${CLUSTER_NAME} istio/multiCluster=true -n ${NAMESPACE}
     $ kubectl config use-context "gke_${proj}_${zone}_cluster-2"
     $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo.yaml@ -l service=ratings
     $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo.yaml@ -l service=reviews
+    $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo.yaml@ -l account=reviews
     $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo.yaml@ -l app=reviews,version=v3
     {{< /text >}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/examples/multicluster/gke/index.md
```diff
@@ -237,6 +237,7 @@ $ kubectl label secret ${CLUSTER_NAME} istio/multiCluster=true -n ${NAMESPACE}
     $ kubectl config use-context "gke_${proj}_${zone}_cluster-2"
     $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo.yaml@ -l service=ratings
     $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo.yaml@ -l service=reviews
+    $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo.yaml@ -l account=reviews
```
ericvn: This would require an additional two lines in the bookinfo.yaml:
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: bookinfo-reviews
  labels:    <===
    account: reviews <===
---
``

frankbu: https://github.com/istio/istio/pull/17577

PR #5096, opened 2019-10-04
Add support for config analysis messages.

geeknoid:
- This provides the place where we can pour config analysis message
descriptions.

### DIFF
--- a/content/en/docs/reference/config/analysis/_index.md
+++ b/content/en/docs/reference/config/analysis/_index.md
@@ -1,0 +1,10 @@
+---
+title: Configuration Analysis Messages
+description: Documents the individual error and warning messages produced during configurarion analysis.
+weight: 40
+layout: analysis-landing
+---
+
+[`istioctl`](/docs/reference/commands/istioctl/#istioctl-experimental-analyze) provides rich analysis of Istio configuration state in
+order to identity invalid or suboptimal configurations. Here's is a list of the distinct possible
+error or warning messages produced by this analysis.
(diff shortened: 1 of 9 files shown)


content/en/docs/reference/config/analysis/_index.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Configuration Analysis Messages
+description: Documents the individual error and warning messages produced during configurarion analysis.
+weight: 40
+layout: analysis-landing
+---
+
+`istioctl` provides rich analysis of Istio configuration state in order to identity
```
frankbu: Since these are istioctl specific messages, should we:

1. Put a link to the istioctl doc?
2. Organize the material differently, so the command and it's associated messages are co-located?

geeknoid: I added a link to istioctl.

I don't think we want to colocate the warning messages with istioctl, since it's likely these messages will come out in other places too, such as in CR status fields and potentially in stand-alone UIs. As we revise the docs for the istioctl command proper, we should add links back to this pages.

content/en/docs/reference/config/analysis/ist0001/index.md
```diff
@@ -0,0 +1,6 @@
+---
```
frankbu: Is this file supposed to be here? All the other errors don't have a corresponding index.md file, so I assume their contents is generated from analysis.yaml.

geeknoid: This is a placeholder to show how to add messages.

We'll need to add one .md file per message. This is work @sushicw and co will take on. I'm just laying the structure.

PR #5134, opened 2019-10-11
Automatically add a warning to older blog posts about them being potentially out of date

geeknoid:


geeknoid:
Note that since preliminary.istio.io is configured for 1.4 and all posts predate that release, it means all blog posts get the warning on preliminary.istio.io.

frankbu:
Is there a way we can mark posts that we know are not outdated, either because the content is not release-specific or we know it still works? Post like [this one](https://deploy-preview-5134--preliminary-istio.netlify.com/blog/2017/0.1-canary/) look strange because they have two warnings now, one that says it may be obsolete, followed by another that says it's been updated.

geeknoid:
OK, I'll add a flag to suppress the auto-warning

geeknoid:
Oh actually, for those blogs we can just skip the target_release field in front-matter. How about I go ahead with the PR as-is, and you (@frankbu) can submit another PR that removes the target_release front-matter from blogs you feel are "timeless"?

PR #5138, opened 2019-10-12
Consolidate release-specific material into the news section.

geeknoid:


geeknoid:
NOTE: something is not right with how this is being rendered. Something has changed from the previews. The intent is NOT to have notes for patch releases (like 1.3.3) be nested under the main release (1.3). I'm not sure what's causing this to happen, I'll hopefully have a fix soon.

PR #5146, opened 2019-10-14
A compromise PR of a long original work

sdake:
See PR: https://github.com/istio/istio.io/pull/5142

Pretty much everything about this PR is compromised...

Please provide a description for what this PR is for.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[X] Docs
[X] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #5183, opened 2019-10-18
modify watch namespace to only Istio-operator

linsun:
also start the eval with demo profile first.

[x ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/standalone-operator/index.md
+++ b/content/en/docs/setup/install/standalone-operator/index.md
@@ -30,14 +30,14 @@ instead, which is a stable feature.
     $ kubectl apply -f https://preliminary.istio.io/operator.yaml
     {{< /text >}}
 
-1. (Optionally) change profiles from the default profile to one of the following profiles:
+1. (Optionally) change profiles from the demo profile to one of the following profiles:
 
 {{< tabset cookie-name="profile" >}}
 
-{{< tab name="permissive mutual TLS" cookie-value="permissive" >}}
+{{< tab name="demo" cookie-value="permissive" >}}
 When using the [permissive mutual TLS mode](/docs/concepts/security/#permissive-mode), all services accept both plaintext and
 mutual TLS traffic. Clients send plaintext traffic unless configured for
-[mutual TLS migration](/docs/tasks/security/mtls-migration/).
+[mutual TLS migration](/docs/tasks/security/mtls-migration/). This profile is installed during the first step.
 
 Choose this profile for:
 
@@ -53,7 +53,7 @@ $ kubectl apply -f https://preliminary.istio.io/operator-profile-demo.yaml
 
 {{< /tab >}}
 
-{{< tab name="strict mutual TLS" cookie-value="strict" >}}
+{{< tab name="demo-auth" cookie-value="strict" >}}
 This profile will enforce
 [mutual TLS authentication](/docs/concepts/security/#mutual-tls-authentication) between all clients and servers.
 
@@ -85,9 +85,7 @@ $ kubectl apply -f https://preliminary.istio.io/operator-profile-sds.yaml
 {{< /tab >}}
 
 {{< tab name="default" cookie-value="default" >}}
-This profile enables Istio's default settings. This profile is installed during the first step.
-
-Run the following command to switch to this profile:
+This profile enables Istio's default settings. Run the following command to switch to this profile:
 
 {{< text bash >}}
 $ kubectl apply -f https://preliminary.istio.io/operator-profile-default.yaml
--- a/scripts/grab_operator_yamls.sh
+++ b/scripts/grab_operator_yamls.sh
@@ -29,12 +29,7 @@ tempdir_operator="$(mktemp -d)"
 # Upstream GIT tags or branches used for the operator repo. The operator is
 # currently only available as a master version. This will change when
 # 1.4 is tagged.
-operator_tag="master"
-
-# Great care should be taken when modifying the ordering of this list. This
-# script cats these files together in order with a yaml separator.
-
-operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" )
+operator_tag="release-1.4"
 
 # Add profiles here to have them automatically added to the website.
 # It is important to also document these files, which is not done automatically.
@@ -49,13 +44,6 @@ git clone -q --single-branch --branch "${operator_tag}" https://github.com/istio
 pushd operator >/dev/null || exit
 git checkout -q "${operator_tag}"
 
-# Generate the main manifest
-for manifest_file in "${operator_manifest_files[@]}"
-do
-	cat "${manifest_file}" >> "${output_dir}"/operator.yaml
-	echo "---" >> "${output_dir}"/operator.yaml
-done
-
 # Generate the profiles
 for profile in "${operator_profiles[@]}"
 do
@@ -63,6 +51,19 @@ do
 	echo "---" >> "${output_dir}"/operator-profile-"${profile}".yaml
 	sed -i "s/profile: default/profile: ${profile}/g" "${output_dir}"/operator-profile-"${profile}".yaml 
 done
+
+# Great care should be taken when modifying the ordering of this list. This
+# script cats these files together in order with a yaml separator.
+
+operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" ""${output_dir}"/operator-profile-demo.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" )
+
+# Generate the main manifest
+for manifest_file in "${operator_manifest_files[@]}"
+do
+	cat "${manifest_file}" >> "${output_dir}"/operator.yaml
+	echo "---" >> "${output_dir}"/operator.yaml
+done
+
 
 popd >/dev/null || exit
 rm -rf "${tempdir_operator}" > /dev/null 2>&1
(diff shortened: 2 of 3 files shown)


linsun:
@sdake PTAL.  The operator.yaml is now generated, with this PR content patched: https://github.com/istio/operator/pull/413

Side note: I found there is a gnu-sed i can install on mac so i can run the sed cmd.

content/en/docs/setup/install/standalone-operator/index.md
```diff
@@ -30,14 +30,14 @@ instead, which is a stable feature.
     $ kubectl apply -f https://preliminary.istio.io/operator.yaml
     {{< /text >}}
 
-1. (Optionally) change profiles from the default profile to one of the following profiles:
+1. (Optionally) change profiles from the demo profile to one of the following profiles:
```
frankbu: What profile is installed on line 30?

This second step doesn't really make sense, does it? First operator.yaml is used to install istio, and then here we optionally reinstall?

linsun: it is the demo profile when this PR gets in.

content/en/docs/setup/install/standalone-operator/index.md
```diff
@@ -30,14 +30,14 @@ instead, which is a stable feature.
...
 
 {{< tabset cookie-name="profile" >}}
 
-{{< tab name="permissive mutual TLS" cookie-value="permissive" >}}
+{{< tab name="demo" cookie-value="permissive" >}}
 When using the [permissive mutual TLS mode](/docs/concepts/security/#permissive-mode), all services accept both plaintext and
 mutual TLS traffic. Clients send plaintext traffic unless configured for
-[mutual TLS migration](/docs/tasks/security/mtls-migration/).
+[mutual TLS migration](/docs/tasks/security/mtls-migration/). This profile is installed during the first step.
```
frankbu: If it's already installed in the first step, why is it an option here?

content/en/docs/setup/install/standalone-operator/index.md
linsun: i kept this step in case folks want to switch to a non demo profile then back to demo profile.

content/en/docs/setup/install/standalone-operator/index.md
linsun: in case user wants to switch back to demo profile :) if you think it is cleaner, i can remove this demo tab, along with demo-auth tab (since that is gone too).

frankbu:
I'm still confused as to why we are publishing the yaml and sh files to istio.io, they are not doc.

I also think these instructions need to be combined with, or replace, the [operator CLI doc](https://preliminary.istio.io/docs/setup/install/operator/). We shouldn't have two experimental operator install docs.

I added this to the agenda for the next doc WG call: https://docs.google.com/document/d/1RGb0NOp0J9QSIrMrZ6wokl16RbSfKHy-6NyTpOQdPa8/edit#

content/en/docs/setup/install/standalone-operator/index.md
frankbu: Shouldn't line 30 then be operator-profile-demo.yaml? If not, what's the difference between them?

linsun:
@frankbu there is a long discussion in @sdake's PR.  for now, it is serving as a page for people to test operator upon upcoming community testings.  We want to see if this can replace our quick start guide for 1.4 (there is debate on stability) but we do like the fact it is 1 line to install istio with the operator.

The cli operator is a little different because it is on CLI side, and will likely be deprecate when the operator on server side is mature.

linsun:
#5146

content/en/docs/setup/install/standalone-operator/index.md
frankbu: Why shouldn't users be able to use the demo profile with strict auth enabled?

content/en/docs/setup/install/standalone-operator/index.md
sdake: @frankbu the demo profile (which is a CR) is bundled in the operator.yaml.  To switch profiles, simply changing the profile gets things moved around.

content/en/docs/setup/install/standalone-operator/index.md
sdake: @frankbu there are two profiles.  demo, and demo-auth.  When we spoke via slack, I had the intention of unifying the documention post 1.4 with the profiles we have, so people can easily switch profiles to execute different tasks.  This would replace the helm operations.

scripts/grab_operator_yamls.sh
```diff
@@ -64,5 +52,19 @@ do
 	sed -i "s/profile: default/profile: ${profile}/g" "${output_dir}"/operator-profile-"${profile}".yaml 
 done
 
+# Great care should be taken when modifying the ordering of this list. This
+# script cats these files together in order with a yaml separator.
+
+operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "${output_dir}/operator-profile-demo.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" )
```
sdake: this part looks good.

scripts/grab_operator_yamls.sh
```diff
@@ -64,5 +52,19 @@ do
...
+
+# Generate the main manifest
+for manifest_file in "${operator_manifest_files[@]}"
+do
+	echo "manifest_file is $manifest_file"
+	cat "${manifest_file}" >> "${output_dir}"/operator.yaml
+	echo "---" >> "${output_dir}"/operator.yaml
+done
```
sdake: this part looks good.

static/operator.yaml
```diff
@@ -14,6 +14,8 @@ spec:
     listKind: IstioControlPlaneList
     plural: istiocontrolplanes
     singular: istiocontrolplane
+    shortNames:
```
sdake: is this coming from the upstream istio/operator/deploy directory?

static/operator.yaml
```diff
@@ -47,7 +49,8 @@ metadata:
   namespace: istio-operator
   name: example-istiocontrolplane
 spec:
-  profile: default
+  profile: demo
+---
```
sdake: if the plan is to terminate each file in istio/operator/deploy with `---` you should remove line 65.  This has impact on the e2e tests too.  Please point out the PR where the deploy directory is changed so I can make appropriate changes in the e2e tests.

static/operator.yaml
```diff
@@ -142,18 +142,32 @@ rules:
...
+  - events
+  - namespaces
+  - pods
+  - persistentvolumeclaims
+  - secrets
+  - services
+  - serviceaccounts
   verbs:
   - '*'
 ---
```
sdake: might as well terminate every yaml in the istio/operator/deploy directory so this is consistently rendered.

static/operator.yaml
```diff
@@ -203,14 +217,14 @@ spec:
       serviceAccountName: istio-operator
       containers:
         - name: istio-operator
-          image: gcr.io/istio-testing/operator:master-latest-daily
+          image: gcr.io/istio-testing/operator:1.4-dev
```
sdake: is there a post install job for this that you added?  If not, lmk and I'll add one.

content/en/docs/setup/install/standalone-operator/index.md
frankbu: Something is not very clear here. Let me try to understand.

Step 1: User applies operator.yaml, which installs the demo profile.
Step 2 (optional): User applies one of the operator-profile-xxx.yaml files to install a different profile.

Question: when a user wants to install something different from the demo profile, why are two steps needed? Why can't the user just apply operator-profile-xxx.yaml, i.e., just one step?

Are the operator-profile-xxx.yaml files additive to the operator.yaml? If so, then how can operator.yaml be starting with the demo profile ... things would have to be removed (not added) to change to a different profile.

What am I missing here?

frankbu: OK, I looked at the yaml files and I understand what's going on now :-)

I think we need to add something in the doc to explain that operator.yaml is starting the operator and by default is configuring a control plane for profile: demo. Then the second step is just changing the profile name (operator then knows what to do).

This would be cleaner if there was a way to pass in the profile into the first apply command, selector label maybe? Then we wouldn't need two steps.

sdake: looking at the yamls was good step :)

selector sounds reasonable, although we are a little crunched for time working on the operator to fine tune how the deployment works.  What this would allow us to do though is bundle all of . the profiles into operator.yaml and select the one we want from one operator.yaml IIUC.

frankbu:
@linsun, @sdake I'm OK with merging this if the goal is to just get something in place for the community testing. I'm sure there will be several followup PRs to get this and other operator docs organized and clarified before 1.4 docs go live on istio.io. It would be good if either or both of you can join the doc WG call next week where I've added this to the agenda.

sdake:
> I'm still confused as to why we are publishing the yaml and sh files to istio.io, they are not doc.
> 
> I also think these instructions need to be combined with, or replace, the [operator CLI doc](https://preliminary.istio.io/docs/setup/install/operator/). We shouldn't have two experimental operator install docs.
> 
> I added this to the agenda for the next doc WG call: https://docs.google.com/document/d/1RGb0NOp0J9QSIrMrZ6wokl16RbSfKHy-6NyTpOQdPa8/edit#

Merging would be fine.  I guess I was lazy with the original PR - wasn't certain how to merge them effectively.  We really do have two models - CLI and standalone operator.  They are semi-orthogonal.

Cheers
-steve

sdake:
> @linsun, @sdake I'm OK with merging this if the goal is to just get something in place for the community testing. I'm sure there will be several followup PRs to get this and other operator docs organized and clarified before 1.4 docs go live on istio.io. It would be good if either or both of you can join the doc WG call next week where I've added this to the agenda.

@frankbu I'll be there.  I also will try to put together a proposal on how I see us evolving the documentation to match the profiles so that we can use template blocks in the documentation to switch profiles or config as needed using either the operator cli or the operator standalone.

Cheers
-steve

PR #5185, opened 2019-10-18
update authorization for HTTP task for v1beta1 policy

yangminzhu:
For istio/istio#12394

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/security/authz-http/index.md
+++ b/content/en/docs/tasks/security/authz-http/index.md
@@ -126,7 +126,7 @@ $ kubectl delete authorizationpolicy.security.istio.io/bookinfo-viewer
 
 This task shows you how to set up workload-level access control using Istio authorization.
 
-### Step 1. applying a default `deny-all` policy
+### applying a default `deny-all` policy
 
 In this step, we will create a default policy that denies all requests to workload in the default namespace.
 
@@ -153,7 +153,7 @@ workloads in the default namespace.
 There may be some delays due to caching and other propagation overhead.
 {{< /tip >}}
 
-### Step 2. allowing access to the `productpage` workload
+### allowing access to the `productpage` workload
 
 We will create a policy that allows external requests to access the `productpage` workload.
 
@@ -188,7 +188,7 @@ in the following steps.
 There may be some delays due to caching and other propagation overhead.
 {{< /tip >}}
 
-### Step 3. allowing access to the `details` and `reviews` workloads
+### allowing access to the `details` and `reviews` workloads
 
 We will create a policy to allow the `productpage` workload to access the `details` and `reviews` workloads. Note that in the
 [setup step](#before-you-begin), we created the `bookinfo-productpage` service account for the `productpage` service. This
@@ -252,7 +252,7 @@ We will show how to do that in the next step.
 There may be some delays due to caching and other propagation overhead.
 {{< /tip >}}
 
-### Step 4. allowing access to the `ratings` workload
+### allowing access to the `ratings` workload
 
 We will create a policy to allow the `reviews` workload to access the `ratings` workload. Note that in the
 [setup step](#before-you-begin), we created a `bookinfo-reviews` service account for the `reviews` workload. This
--- a/content/en/docs/tasks/security/authz-http/index.md
+++ b/content/en/docs/tasks/security/authz-http/index.md
@@ -31,7 +31,7 @@ the product page, you can see the following sections:
 When you refresh the page, the app shows different versions of reviews in the product page.
 The app presents the reviews in a round robin style: red stars, black stars, or no stars.
 
-## Enforcing Mesh-level access control
+## Enforcing mesh-level access control
 
 Using Istio authorization, you can easily setup mesh-level access control for all workloads in the mesh.
 
@@ -60,7 +60,7 @@ workloads in the default namespace.
 There may be some delays due to caching and other propagation overhead.
 {{< /tip >}}
 
-### Cleanup Mesh-level access control
+### Cleanup mesh-level access control
 
 Remove the following configuration before you proceed to the next task:
 
@@ -68,7 +68,7 @@ Remove the following configuration before you proceed to the next task:
 $ kubectl delete authorizationpolicy.security.istio.io/deny-all -n istio-system
 {{< /text >}}
 
-## Enforcing Namespace-level access control
+## Enforcing namespace-level access control
 
 Using Istio authorization, you can easily setup namespace-level access control by specifying all (or a collection of) workloads
 in a namespace are accessible by workloads from another namespace.
@@ -114,7 +114,7 @@ with the "Book Details" section in the lower left part and the "Book Reviews" se
 There may be some delays due to caching and other propagation overhead.
 {{< /tip >}}
 
-### Cleanup Namespace-level access control
+### Cleanup namespace-level access control
 
 Remove the following configuration before you proceed to the next task:
 
@@ -122,7 +122,7 @@ Remove the following configuration before you proceed to the next task:
 $ kubectl delete authorizationpolicy.security.istio.io/bookinfo-viewer
 {{< /text >}}
 
-## Enforcing Workload-level access control
+## Enforcing workload-level access control
 
 This task shows you how to set up workload-level access control using Istio authorization.
 
@@ -291,7 +291,7 @@ the "black" and "red" ratings in the "Book Reviews" section.
 There may be some delays due to caching and other propagation overhead.
 {{< /tip >}}
 
-### Cleanup Workload-level access control
+### Cleanup workload-level access control
 
 *   Remove Istio authorization policy configurations:
 
--- a/content/en/docs/tasks/security/authz-http/index.md
+++ b/content/en/docs/tasks/security/authz-http/index.md
@@ -40,8 +40,11 @@ as there may be some delays due to caching and other propagation overhead.
 Using Istio, you can easily setup mesh-level access control for all workloads in your mesh.
 
 Run the following command to create a mesh-level access control policy `deny-all` policy
-with no workload selector in the root namespace `istio-system`, the policy will select all
-workloads in the mesh:
+in the root namespace, A policy in the root namespace will select all workloads in the mesh.
+
+The root namespace is configurable in the [`MeshConfig`](/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig)
+and has the default value of `istio-system`. If you have changed it to a different value from the
+default `istio-system`, Please update the value accordingly in the following examples.
 
 {{< text bash >}}
 $ kubectl apply -f - <<EOF
@@ -55,12 +58,14 @@ spec:
 EOF
 {{< /text >}}
 
+The `{}` in the policy means `deny-all`, you can also set other mesh-level policy for your own needs.
+
 Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
 Now you should see `"RBAC: access denied"`. The error shows that the configured `deny-all`
 policy is working as intended, and Istio doesn't have any rules that allow any access to
 workloads in the mesh.
 
-### Disable mesh-level access control
+### Clean up mesh-level access control
 
 Disable the configured access control with the following command before continuing:
 
@@ -70,8 +75,8 @@ $ kubectl delete authorizationpolicy.security.istio.io/deny-all -n istio-system
 
 ## Enforce namespace-level access control
 
-With Istio, you can easily setup namespace-level access control. You can configure whether
-workloads from a namespace can access all or a collection of workloads in another namespace.
+With Istio, you can easily setup namespace-level access control. You can configure
+how could workloads from a namespace access workloads in another namespace.
 
 The Bookinfo sample deploys the `productpage`, `reviews`, `details`, `ratings` {{< gloss "workload" >}}workloads{{< /gloss >}}
 in the `default` namespace. Istio deploys its components, for example `istio-ingressgateway`,
@@ -108,7 +113,7 @@ Now if you point your browser at Bookinfo's `productpage` (`http://$GATEWAY_URL/
 You should see the "Bookinfo Sample" page, with the "Book Details" section in the lower left part
 and the "Book Reviews" section in the lower right part.
 
-### Disable namespace-level access control
+### Clean up namespace-level access control
 
 Disable the configured access control with the following command before continuing:
 
@@ -119,6 +124,8 @@ $ kubectl delete authorizationpolicy.security.istio.io/bookinfo-viewer
 ## Enforce workload-level access control
 
 This task shows you how to set up workload-level access control using Istio authorization.
+You will start with a simple `deny-all` policy that rejects all requests to the workload, and then
+grant the permission gradually and incrementally to allow more access to the workload.
 
 1. Apply a default `deny-all` policy for workloads in the default namespace
 
@@ -259,7 +266,7 @@ This task shows you how to set up workload-level access control using Istio auth
     Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see
     the "black" and "red" ratings in the "Book Reviews" section.
 
-### Disable workload-level access control
+### Clean up workload-level access control
 
 *   Remove all authorization policies from your configuration:
 
--- a/content/en/docs/tasks/security/authz-http/index.md
+++ b/content/en/docs/tasks/security/authz-http/index.md
@@ -35,109 +35,15 @@ If you don't see the expected output in the browser as you follow the task, retr
 because some delay is possible due to caching and other propagation overhead.
 {{< /tip >}}
 
-## Enforce mesh-level access control
+## Configure access control for workloads using HTTP traffic
 
-Using Istio, you can easily setup mesh-level access control for all workloads in your mesh.
+Using Istio, you can easily setup access control for {{< gloss "workload" >}}workloads{{< /gloss >}}
+in your mesh. This task shows you how to set up access control using Istio authorization.
+You will start with a simple `deny-all` policy that rejects all requests to the workload,
+and then grant more access to the workload gradually and incrementally.
 
-Run the following command to create a mesh-level access control policy `deny-all` policy
-in the root namespace, A policy in the root namespace will select all workloads in the mesh.
-
-The root namespace is configurable in the [`MeshConfig`](/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig)
-and has the default value of `istio-system`. If you have changed it to a different value from the
-default `istio-system`, Please update the value accordingly in the following examples.
-
-This following policy has no rule so it cannot grant permission for any traffic. In other words, it denies all.
-
-{{< text bash >}}
-$ kubectl apply -f - <<EOF
-apiVersion: security.istio.io/v1beta1
-kind: AuthorizationPolicy
-metadata:
-  name: deny-all
-  namespace: istio-system
-spec:
-  {}
-EOF
-{{< /text >}}
-
-Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
-You should see `"RBAC: access denied"`. The error shows that the configured `deny-all` policy
-is working as intended, and Istio doesn't have any rules that allow any access to
-workloads in the mesh.
-
-**Congratulations!** You successfully deployed applied a mesh-level authorization policy
-to enforce access control for the requests.
-
-### Clean up mesh-level access control
-
-Remove the authorization policy with the following command before continuing:
-
-{{< text bash >}}
-$ kubectl delete authorizationpolicy.security.istio.io/deny-all -n istio-system
-{{< /text >}}
-
-## Enforce namespace-level access control
-
-With Istio, you can easily setup namespace-level access control. You can configure
-how could workloads from a namespace access workloads in another namespace.
-
-The Bookinfo sample deploys the `productpage`, `reviews`, `details`, `ratings` {{< gloss "workload" >}}workloads{{< /gloss >}}
-in the `default` namespace. Istio deploys its components, for example `istio-ingressgateway`,
-in the `istio-system` namespace.
-
-The following command creates a `bookinfo-viewer` policy that allows workloads in the `default`
-and `istio-system` namespaces to access workloads in the `default` namespace with `GET` method:
-
-{{< text bash >}}
-$ kubectl apply -f - <<EOF
-apiVersion: "security.istio.io/v1beta1"
-kind: "AuthorizationPolicy"
-metadata:
-  name: "bookinfo-viewer"
-  namespace: default
-spec:
-  rules:
-  - from:
-    - source:
-        namespaces: ["default", "istio-system"]
-    to:
-    - operation:
-        methods: ["GET"]
-EOF
-{{< /text >}}
-
-You can expect to see output similar to the following:
-
-{{< text plain >}}
-authorizationpolicy.security.istio.io/bookinfo-viewer created
-{{< /text >}}
-
-Point your browser at Bookinfo's `productpage` (`http://$GATEWAY_URL/productpage`).
-You should see the "Bookinfo Sample" page, with the "Book Details" section in the lower left part
-and the "Book Reviews" section in the lower right part.
-
-**Congratulations!** You successfully deployed applied a namespace-level authorization policy
-to enforce access control for the requests.
-
-### Clean up namespace-level access control
-
-Remove the authorization policy with the following command before continuing:
-
-{{< text bash >}}
-$ kubectl delete authorizationpolicy.security.istio.io/bookinfo-viewer
-{{< /text >}}
-
-## Enforce workload-level access control
-
-This task shows you how to set up workload-level access control using Istio authorization.
-You will start with a simple `deny-all` policy that rejects all requests to the workload, and then
-grant the permission gradually and incrementally to allow more access to the workload.
-
-1. Apply a default `deny-all` policy for workloads in the default namespace
-
-    The following command creates a `deny-all` policy with no workload selector, the policy will
-    select all workloads in the default namespace. The policy will deny all access and we will
-    add more policies to grant the access incrementally in following steps.
+1. Run the following command to create a policy `deny-all`, the policy has no rule so it
+   cannot grant permission for any traffic. In other words, it denies all:
 
     {{< text bash >}}
     $ kubectl apply -f - <<EOF
@@ -145,20 +51,53 @@ grant the permission gradually and incrementally to allow more access to the wor
     kind: AuthorizationPolicy
     metadata:
       name: deny-all
+      namespace: default
     spec:
       {}
     EOF
     {{< /text >}}
 
     Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
-    Now you should see `"RBAC: access denied"`. The error shows that the configured `deny-all` policy
-    is working as intended, and Istio doesn't have any rules that allow any access
-    to workloads in the default namespace.
+    You should see `"RBAC: access denied"`. The error shows that the configured `deny-all` policy
+    is working as intended, and Istio doesn't have any rules that allow any access to
+    workloads in the mesh.
 
-1. Allow access to the `productpage` workload
+    With Istio, you can easily configure namespace-level access control. You can configure how could workloads
+    from a namespace access workloads in another namespace.
 
-    The following command creates a `productpage-viewer` policy which allows
-    access to the `productpage` workload with `GET` method for all users and workloads.
+1. Run the following command to create a `bookinfo-viewer` policy that allows workloads in the `default`
+   and `istio-system` namespaces to access workloads in the `default` namespace with `GET` method:
+
+    {{< text bash >}}
+    $ kubectl apply -f - <<EOF
+    apiVersion: "security.istio.io/v1beta1"
+    kind: "AuthorizationPolicy"
+    metadata:
+      name: "bookinfo-viewer"
+      namespace: default
+    spec:
+      rules:
+      - from:
+        - source:
+            namespaces: ["default", "istio-system"]
+        to:
+        - operation:
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+    Point your browser at Bookinfo's `productpage` (`http://$GATEWAY_URL/productpage`).
+    You should see the "Bookinfo Sample" page, with the "Book Details" section in the lower left part
+    and the "Book Reviews" section in the lower right part.
+
+1. Remove the policy with the following command before continuing:
+
+    {{< text bash >}}
+    $ kubectl delete authorizationpolicy.security.istio.io/bookinfo-viewer
+    {{< /text >}}
+
+1. Run the following command to create a `productpage-viewer` policy which allows
+   access to the `productpage` workload with `GET` method for all users and workloads:
 
     {{< text bash >}}
     $ kubectl apply -f - <<EOF
@@ -180,23 +119,15 @@ grant the permission gradually and incrementally to allow more access to the wor
 
     Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
     Now you should see the "Bookinfo Sample" page. But there are errors `Error fetching product details`
-    and `Error fetching product reviews` on the page. These errors are expected because
-    we have not granted the `productpage` workload access to the `details` and `reviews` workloads.
-    We will fix the errors in the following steps.
+    and `Error fetching product reviews` on the page.
 
-1. Allow access to the `details` and `reviews` workloads
+    These errors are expected because we have not granted the `productpage`
+    workload access to the `details` and `reviews` workloads. We will fix the
+    errors in the following steps.
 
-    The following command has the following effects:
-
-    * Creates a `details-viewer` policy which allows access with `GET` method to the `details` workload for
-      service account `cluster.local/ns/default/sa/bookinfo-productpage` (representing the `details` workload).
-
-    * Creates a `reviews-viewer` policy which allows access with `GET` method to the `reviews` workload for
-      service account `cluster.local/ns/default/sa/bookinfo-productpage` (representing the `reviews` workload).
-
-    Note that in the [setup step](#before-you-begin), we created the `bookinfo-productpage` service account
-    for the `productpage` workload. This `bookinfo-productpage` service account is the authenticated identify
-    for the `productpage` workload.
+1. Run the following command to create a policy `details-viewer` to allow the `productpage`
+   workload (represented by `principals: ["cluster.local/ns/default/sa/bookinfo-productpage"]`)
+   to access the `details` workload through `GET` methods:
 
     {{< text bash >}}
     $ kubectl apply -f - <<EOF
@@ -216,7 +147,15 @@ grant the permission gradually and incrementally to allow more access to the wor
         to:
         - operation:
             methods: ["GET"]
-    ---
+    EOF
+    {{< /text >}}
+
+1. Run the following command to create a policy `reviews-viewer` to allow the `productpage` workload
+   (represented by `principals: ["cluster.local/ns/default/sa/bookinfo-productpage"]`) to access the
+   `reviews` workload through `GET` methods:
+
+    {{< text bash >}}
+    $ kubectl apply -f - <<EOF
     apiVersion: "security.istio.io/v1beta1"
     kind: "AuthorizationPolicy"
     metadata:
@@ -238,15 +177,15 @@ grant the permission gradually and incrementally to allow more access to the wor
 
     Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see the "Bookinfo Sample"
     page with "Book Details" on the lower left part, and "Book Reviews" on the lower right part. However, in the "Book Reviews" section,
-    there is an error `Ratings service currently unavailable`. This is because "reviews" workload does not have permission to access
-    "ratings" workload. To fix this issue, you need to grant the `reviews` workload access to the `ratings` workload.
+    there is an error `Ratings service currently unavailable`.
+
+    This is because "reviews" workload does not have permission to access `ratings` workload.
+    To fix this issue, you need to grant the `reviews` workload access to the `ratings` workload.
     We will show how to do that in the next step.
 
-1. Allow access to the `ratings` workload
-
-    The following command creates a policy to allow the `reviews` workload to access the `ratings` workload. Note that in the
-    [setup step](#before-you-begin), we created a `bookinfo-reviews` service account for the `reviews` workload. This
-    service account is the authenticated identify for the `reviews` workload.
+1. Run the following command to create a policy `ratings-viewer` to allow the `reviews` workload
+   (represented by `principals: ["cluster.local/ns/default/sa/bookinfo-reviews"]`) to access the
+   `ratings` workload through `GET` methods:
 
     {{< text bash >}}
     $ kubectl apply -f - <<EOF
@@ -272,12 +211,12 @@ grant the permission gradually and incrementally to allow more access to the wor
     Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
     You should see the "black" and "red" ratings in the "Book Reviews" section.
 
-    **Congratulations!** You successfully deployed applied both a mesh-level and a workload-level authorization policy
-    to enforce access control for the requests.
+    **Congratulations!** You successfully applied authorization policy to enforce access
+    control for workloads using HTTP traffic.
 
-### Clean up workload-level access control
+## Clean up
 
-*   Remove all authorization policies from your configuration:
+1. Remove all authorization policies from your configuration:
 
     {{< text bash >}}
     $ kubectl delete authorizationpolicy.security.istio.io/deny-all
--- a/content/en/docs/tasks/security/authz-http/index.md
+++ b/content/en/docs/tasks/security/authz-http/index.md
@@ -1,15 +1,14 @@
 ---
-title: Authorization for HTTP protocol
-description: Shows how to set up role-based access control for HTTP protocol.
+title: Authorization for HTTP traffic
+description: Shows how to set up role-based access control for HTTP traffic.
 weight: 10
 keywords: [security,access-control,rbac,authorization]
 aliases:
     - /docs/tasks/security/role-based-access-control.html
 ---
 
-This task covers the activities you might need to perform to set up Istio authorization for
-HTTP protocol in an Istio mesh. You can read more in [authorization](/docs/concepts/security/#authorization)
-and get started with a basic tutorial in Istio Security Basics.
+This task shows you how to set up Istio authorization for HTTP traffic in an Istio mesh.
+Learn more in our [authorization concept page](/docs/concepts/security/#authorization).
 
 ## Before you begin
 
@@ -31,11 +30,18 @@ the product page, you can see the following sections:
 When you refresh the page, the app shows different versions of reviews in the product page.
 The app presents the reviews in a round robin style: red stars, black stars, or no stars.
 
-## Enforcing mesh-level access control
+{{< tip >}}
+If you don't see the expected output in the browser, retry in a few more seconds
+as there may be some delays due to caching and other propagation overhead.
+{{< /tip >}}
 
-Using Istio authorization, you can easily setup mesh-level access control for all workloads in the mesh.
+## Enforce mesh-level access control
 
-Run the following command to create a mesh-level access control policy:
+Using Istio, you can easily setup mesh-level access control for all workloads in your mesh.
+
+Run the following command to create a mesh-level access control policy `deny-all` policy
+with no workload selector in the root namespace `istio-system`, the policy will select all
+workloads in the mesh:
 
 {{< text bash >}}
 $ kubectl apply -f - <<EOF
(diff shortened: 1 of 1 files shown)


yangminzhu:
cc @liminw

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,86 +31,80 @@ the product page, you can see the following sections:
 When you refresh the page, the app shows different versions of reviews in the product page.
 The app presents the reviews in a round robin style: red stars, black stars, or no stars.
 
-## Enabling Istio authorization
+## Enforcing Mesh-level access control
```
geeknoid:
```suggestion
## Enforcing mesh-level access control
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,86 +31,80 @@ the product page, you can see the following sections:
...
+
+Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see `"RBAC: access denied"`.
+This is because Istio authorization is "deny by default" and the `deny-all` policy doesn't have any rules to allow any access to
+workloads in the default namespace.
 
 {{< tip >}}
 There may be some delays due to caching and other propagation overhead.
 {{< /tip >}}
 
+### Cleanup Mesh-level access control
```
geeknoid:
```suggestion
### Cleanup mesh-level access control
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,86 +31,80 @@ the product page, you can see the following sections:
...
+
+Remove the following configuration before you proceed to the next task:
+
+{{< text bash >}}
+$ kubectl delete authorizationpolicy.security.istio.io/deny-all -n istio-system
+{{< /text >}}
+
 ## Enforcing Namespace-level access control
```
geeknoid:
```suggestion
## Enforcing namespace-level access control
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,54 +31,80 @@ the product page, you can see the following sections:
...
+    - source:
+        namespaces: ["default", "istio-system"]
+    to:
+    - operation:
+        methods: ["GET"]
+EOF
+{{< /text >}}
 
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="enforcing_namespace_level_access_control_bind_service_viewer.yaml" >}}
+The command creates a `bookinfo-viewer` policy which allows read access to any workloads in the `default` namespace.
```
rcaballeromx:
```suggestion
The created `bookinfo-viewer` policy grants read access to the workloads in the `default` namespace.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
 {{< text bash >}}
-$ kubectl delete -f @samples/bookinfo/platform/kube/rbac/namespace-policy.yaml@
+$ kubectl delete authorizationpolicy.security.istio.io/bookinfo-viewer
 {{< /text >}}
 
-## Enforcing Service-level access control
+## Enforcing workload-level access control
 
-This task shows you how to set up service-level access control using Istio authorization. Before you start, please make sure that:
+This task shows you how to set up workload-level access control using Istio authorization.
```
rcaballeromx: The introduction needs some additional explanation. Why are you starting with a denying all policy? What is the **best** flow to enforce access control at this level?
Generally, the task does not let readers know why they should use each level of access control. What are the goals at each level? When do I apply restrictions on each level?

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
 
-* You have [enabled Istio authorization](#enabling-istio-authorization).
-* You have [removed namespace-level authorization policy](#cleanup-namespace-level-access-control).
+### applying a default `deny-all` policy
```
rcaballeromx: This should be a step, not a heading.
```suggestion
1. Apply a default `deny-all` policy for the `default` namespace with the following command:
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
 
-Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). You should see `"RBAC: access denied"`.
-We will incrementally add access permission to the services in the Bookinfo sample.
+In this step, we will create a default policy that denies all requests to workload in the default namespace.
```
rcaballeromx:
```suggestion
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
 
-### Step 1. allowing access to the `productpage` service
+Run the following command:
```
rcaballeromx:
```suggestion
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
+kind: AuthorizationPolicy
+metadata:
+  name: deny-all
+spec:
+  {}
+EOF
+{{< /text >}}
 
-Run the following command:
+The command creates a `deny-all` policy with no workload selector, the policy will select all workloads in the default namespace.
```
rcaballeromx: This information should come before the command. This comment applies throughout. Make sure to explain what the command does and how before asking a reader to run it. This empowers readers to run the command with certainty around what it will do. This is true for this step and the subsequent ones.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
+
+Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see `"RBAC: access denied"`.
+This is because Istio authorization is "deny by default" and the `deny-all` policy doesn't have any rules to allow any access to
+workloads in the default namespace.
 
-{{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="enforcing_service_level_access_control_step1_apply.sh" >}}
+{{< tip >}}
```
rcaballeromx: This tip should be served at the beginning of the task since readers need to be aware that after applying each configuration they need to wait for it to propagate.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
+  name: deny-all
+spec:
+  {}
+EOF
+{{< /text >}}
 
-Run the following command:
+The command creates a `deny-all` policy with no workload selector, the policy will select all workloads in the default namespace.
+
+Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see `"RBAC: access denied"`.
```
rcaballeromx:
```suggestion
Whey you go to the `productpage` of Bookinfo  at `http://$GATEWAY_URL/productpage`, you see the following error message: `"RBAC: access denied"`.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
+This is because Istio authorization is "deny by default" and the `deny-all` policy doesn't have any rules to allow any access to
```
rcaballeromx:
```suggestion
The error shows that the configured `deny-all` policy is working as intended, and Istio doesn't have any rules that allow any access to
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
+workloads in the default namespace.
 
-{{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="enforcing_service_level_access_control_step1_apply.sh" >}}
+{{< tip >}}
+There may be some delays due to caching and other propagation overhead.
+{{< /tip >}}
 
-Once applied, the policy has the following effects:
+### allowing access to the `productpage` workload
```
rcaballeromx: See my edits for the previous step. This should be a numbered step not a heading. Please rewrite the text in lines 156-160 into a single numbered step as shown above.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -94,98 +119,170 @@ There may be some delays due to caching and other propagation overhead.
...
+      app: productpage
+  rules:
+  - to:
+    - operation:
+        methods: ["GET"]
+EOF
+{{< /text >}}
 
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="enforcing_service_level_access_control_step1_bind_productpage_viewer.yaml" >}}
+The command creates a `productpage-viewer` policy which allows read access to the `productpage` workload for all users and workloads.
```
rcaballeromx: Please include this information in the text before the command.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -194,16 +291,14 @@ the "black" and "red" ratings in the "Book Reviews" section.
...
-
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="remove_istio_authorization_policy.sh" >}}
-
-    Alternatively, you can delete all `ServiceRole` and `ServiceRoleBinding` resources by running the following commands:
-
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="remove_istio_authorization_policy_alternative.sh" >}}
+### Cleanup workload-level access control
 
-*   Disable Istio authorization:
+*   Remove Istio authorization policy configurations:
```
rcaballeromx:
```suggestion
*   Remove all authorization policies from your configuration:
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -1,16 +1,15 @@
 ---
-title: Authorization for HTTP Services
-description: Shows how to set up role-based access control for HTTP services.
+title: Authorization for HTTP workloads
```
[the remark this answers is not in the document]
yangminzhu: Yes, that's the workload we want to refer to in this task.

Also updated to user `traffic` to replace the `protocoal` based on comments from @geeknoid in the other PR.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: Moved the tip to the beginning of the task.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: Are you sure we want to start the sentence with `When...`? I remember I received comments when previously updating this page that we should start with a verb to indicate that this is asking user to do something explicitly.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,54 +31,80 @@ the product page, you can see the following sections:
...
-Run the following command to enable Istio authorization for the `default` namespace:
+Using Istio authorization, you can easily setup mesh-level access control for all workloads in the mesh.
 
-{{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="enabling_istio_authorization.sh" >}}
+Run the following command to create a mesh-level access control policy:
 
-Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see
-`"RBAC: access denied"`. This is because Istio authorization is "deny by default", which means that you need to
-explicitly define access control policy to grant access to any service.
+{{< text bash >}}
```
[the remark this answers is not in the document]
yangminzhu: These are embedded right in the task page. We have tests for the old task page but we don't have one for this for now.

It's more convenient for users to try out the new policy, they don't need access to local files and just need to copy and run the command. This technique is also used in many other tasks, like authentication and networking.

The other benefit is it reduces the maintenance burden, we could only maintain one copy of the sample policy.

yangminzhu:
@rcaballeromx I addressed the comments so far, PTAL, thanks!

yangminzhu:
@rcaballeromx ping

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,243 @@ the product page, you can see the following sections:
...
-{{< tip >}}
-There may be some delays due to caching and other propagation overhead.
-{{< /tip >}}
+Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
+Now you should see `"RBAC: access denied"`. The error shows that the configured `deny-all`
+policy is working as intended, and Istio doesn't have any rules that allow any access to
+workloads in the mesh.
 
-### Cleanup namespace-level access control
+### Disable mesh-level access control
```
diemtvu: I think the previous phrase "clean up" is better. You don't need to "disable" if you haven't add it in the first place.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,243 @@ the product page, you can see the following sections:
...
 
-Remove the following configuration before you proceed to the next task:
+Disable the configured access control with the following command before continuing:
 
 {{< text bash >}}
-$ kubectl delete -f @samples/bookinfo/platform/kube/rbac/namespace-policy.yaml@
+$ kubectl delete authorizationpolicy.security.istio.io/deny-all -n istio-system
```
diemtvu: [nit] just `authorizationpolicy` may be sufficient.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,243 @@ the product page, you can see the following sections:
...
-### Step 1. allowing access to the `productpage` service
-
-In this step, we will create a policy that allows external requests to access the `productpage` service via Ingress.
-
-Run the following command:
+## Enforce namespace-level access control
 
-{{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="enforcing_service_level_access_control_step1_apply.sh" >}}
+With Istio, you can easily setup namespace-level access control. You can configure whether
+workloads from a namespace can access all or a collection of workloads in another namespace.
```
diemtvu: A bit hard to read. Can you rephrase?

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,243 @@ the product page, you can see the following sections:
...
-Once applied, the policy has the following effects:
+Using Istio, you can easily setup mesh-level access control for all workloads in your mesh.
 
-*   Creates a `ServiceRole` `service-viewer` which allows read access to any service in the `default` namespace that has
-the `app` label
-set to one of the values `productpage`, `details`, `reviews`, or `ratings`. Note that there is a
-constraint specifying that
-the services must have one of the listed `app` labels.
+Run the following command to create a mesh-level access control policy `deny-all` policy
+with no workload selector in the root namespace `istio-system`, the policy will select all
```
diemtvu: Should we add a link for root namespace reference (e.g this https://preliminary.istio.io/docs/reference/config/istio.mesh.v1alpha1/) ? It may also worth to mention `istio-system` is the value used in the "before you begin" section

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,243 @@ the product page, you can see the following sections:
...
-{{< tip >}}
-There may be some delays due to caching and other propagation overhead.
-{{< /tip >}}
+The Bookinfo sample deploys the `productpage`, `reviews`, `details`, `ratings` {{< gloss "workload" >}}workloads{{< /gloss >}}
+in the `default` namespace. Istio deploys its components, for example `istio-ingressgateway`,
+in the `istio-system` namespace.
 
-### Step 2. allowing access to the `details` and `reviews` services
+The following command creates a `bookinfo-viewer` policy that allows workloads in the `default`
+and `istio-system` namespaces to access workloads in the `default` namespace with `GET` method:
```
diemtvu: [nit] It may cause some confusion between the use of `istio-system` as root namespace, and as a namespace that some workload run. It would be best if the setup use different value for root namespace.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,243 @@ the product page, you can see the following sections:
...
-servicerolebinding "bind-service-viewer" created
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: deny-all
+  namespace: istio-system
+spec:
+  {}
```
diemtvu: I think you need to high light `{}` means deny-all. Also, add examples or explain that user can set other mesh-wide policy than deny-all (or not)

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,243 @@ the product page, you can see the following sections:
...
+authorizationpolicy.security.istio.io/bookinfo-viewer created
+{{< /text >}}
 
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="remove_istio_authorization_policy.sh" >}}
+Now if you point your browser at Bookinfo's `productpage` (`http://$GATEWAY_URL/productpage`).
+You should see the "Bookinfo Sample" page, with the "Book Details" section in the lower left part
+and the "Book Reviews" section in the lower right part.
 
-    Alternatively, you can delete all `ServiceRole` and `ServiceRoleBinding` resources by running the following commands:
+### Disable namespace-level access control
```
diemtvu: Same here, "clean up" make more sense to me.

content/en/docs/tasks/security/authz-http/index.md
diemtvu: +1. I think instead of explaining the 3 different scopes, it is better to show how the authz policies match (to workload) and combine.

content/en/docs/tasks/security/authz-http/index.md
liminw: +1

content/en/docs/tasks/security/authz-http/index.md
liminw: +2. We can start with some introduction explaining the steps. So readers will get a picture of what this task demonstrates before delving into each step.

content/en/docs/tasks/security/authz-http/index.md
liminw: +1. Also, root namespace is configurable. Readers need to make sure that root namespace is not set to any other value. Maybe we can have a step here to check the value of the root namespace.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: I made the change from @rcaballeromx's comment, I can revert it back but note we cannot have multiple `### Clean up`, we have to make the title different from each other.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: Updated to link to mesh config.

The task assumes istio is installed with the basic quick start as a demo purpose, we can mention the root namespace is configurable but I don't think we need to check it in this task.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: This is to avoid a tricky case where user is upgrading from 1.3 where there is a `authorizationpolicy.istio.io` resource for some legacy reason. That CRD is never used though.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: Done.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: I think the current installation uses `istio-system` as the default root namespace, should be updated to a separate value.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: I added a high-level summary for the big picture of the task.

@diemtvu For now I'm converting the old task that was focusing on the scope, I think we can rewrite to make this more useful by also talking explaining more about the selector part, but let's do it in a follow-up PR to avoid introducing too much new stuff in a single PR.

yangminzhu:
@liminw @diemtvu Comments addressed, PTAL, thanks.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,250 @@ the product page, you can see the following sections:
...
-{{< tip >}}
-There may be some delays due to caching and other propagation overhead.
-{{< /tip >}}
+Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
+Now you should see `"RBAC: access denied"`. The error shows that the configured `deny-all`
+policy is working as intended, and Istio doesn't have any rules that allow any access to
+workloads in the mesh.
 
-### Cleanup namespace-level access control
+### Clean up mesh-level access control
```
liminw: Clean up => Cleaning up

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,250 @@ the product page, you can see the following sections:
...
-Remove the following configuration before you proceed to the next task:
+Disable the configured access control with the following command before continuing:
 
 {{< text bash >}}
-$ kubectl delete -f @samples/bookinfo/platform/kube/rbac/namespace-policy.yaml@
+$ kubectl delete authorizationpolicy.security.istio.io/deny-all -n istio-system
 {{< /text >}}
 
-## Enforcing Service-level access control
+## Enforce namespace-level access control
```
liminw: Enforce => Enforcing

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,250 @@ the product page, you can see the following sections:
...
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="remove_istio_authorization_policy_alternative.sh" >}}
+Disable the configured access control with the following command before continuing:
 
-*   Disable Istio authorization:
+{{< text bash >}}
+$ kubectl delete authorizationpolicy.security.istio.io/bookinfo-viewer
+{{< /text >}}
 
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="disabling_istio_authorization.sh" >}}
+## Enforce workload-level access control
```
liminw: Enforcing

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,250 @@ the product page, you can see the following sections:
...
+        to:
+        - operation:
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see
+    the "black" and "red" ratings in the "Book Reviews" section.
+
+### Clean up workload-level access control
```
liminw: Cleaning up

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,250 @@ the product page, you can see the following sections:
...
+  name: deny-all
+  namespace: istio-system
+spec:
+  {}
+EOF
 {{< /text >}}
 
-Now if you point your browser at Bookinfo's `productpage` (`http://$GATEWAY_URL/productpage`). You should see the "Bookinfo Sample" page,
-with the "Book Details" section in the lower left part and the "Book Reviews" section in the lower right part.
+The `{}` in the policy means `deny-all`, you can also set other mesh-level policy for your own needs.
```
diemtvu: This phrase may confuse users that this is a deny rule. In fact, `{}` is just a general yaml syntax for empty. 
Suggest to reword to smt like:

```
This policy has no rule so it cannot grant permission for any traffic. In other words, it denies all.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,250 @@ the product page, you can see the following sections:
...
-In our Bookinfo sample, the `productpage`, `reviews`, `details`, `ratings` services are deployed in the `default` namespace.
-The Istio components like `istio-ingressgateway` service are deployed in the `istio-system` namespace. We can define a policy that
-any service in the `default` namespace that has the `app` label set to one of the values of
-`productpage`, `details`, `reviews`, or `ratings`
-is accessible by services in the same namespace (i.e., `default`) and services in the `istio-system` namespace.
-
-Run the following command to create a namespace-level access control policy:
-
-{{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="enforcing_namespace_level_access_control_apply.sh" >}}
+## Enforce mesh-level access control
```
liminw: Enforcing

content/en/docs/tasks/security/authz-http/index.md
diemtvu: I think you should keep this policy, as it's more natural to show the namespace or workload policy effects later (granting access).
Again, I think we should change the narrative of this tutorial. Putting in 3 distinguished scopes is hard for user to understand why. Rather, you should showing deny-all via global, then open per-namespace and/or workload, pointing out the permission are always _additive_ in all cases.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: @rcaballeromx is asking me to change this from "enforcing" to "enforce", can we reach agreements before making any further changes?

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: See my other comment.

content/en/docs/tasks/security/authz-http/index.md
diemtvu: You may avoid include it here for simplicity. 

I also recommend do the workload policy first to grand ingress to access product-page, and showing the result that request to ingress now get to the product page but no content (you still have global deny-all). Then next step is adding namespace policy for all bookinfo apps, and now every thing works again.

content/en/docs/tasks/security/authz-http/index.md
diemtvu: I think it's ok to rewrite the story in one PR.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: this policy is added again in the `Enforce workload-level access control` section where we explicitly talk about this process that grants more permission gradually. I was mainly trying to maintain the same structure but just switch to use the new policy in these tasks, I agree we probably need to rewrite the task completely, and I think it would be better to not to use bookinfo, but to use sleep/httpbin that gives us more flexibility to show different features of the API.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: In terms of rewrite, I actually want to use sleep/httpbin in multiple namespaces demonstrate more features of the API, but for now I do not have enough time for the complete new task, we still need to update the migration/troubleshooting/etc. docs.

@liminw thoughts?

content/en/docs/tasks/security/authz-http/index.md
liminw: @rcaballeromx can decide. I was looking at the current docs. But there is inconsistency here and there. Mostly it has "ing" after a verb in the title, but there are exceptions. For example: https://preliminary.istio.io/docs/tasks/security/health-check/#deploying-citadel-with-health-checking, https://preliminary.istio.io/docs/tasks/traffic-management/circuit-breaking/

yangminzhu:
@liminw Could you approve if no further change is needed?

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+
+    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
+    Now you should see the "Bookinfo Sample" page. But there are errors `Error fetching product details`
+    and `Error fetching product reviews` on the page. These errors are expected because
+    we have not granted the `productpage` workload access to the `details` and `reviews` workloads.
+    We will fix the errors in the following steps.
+
+1. Allow access to the `details` and `reviews` workloads
+
+    The following command has the following effects:
```
rcaballeromx: This context should come before the call to action.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+    {{< /text >}}
+
+    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
+    Now you should see `"RBAC: access denied"`. The error shows that the configured `deny-all` policy
+    is working as intended, and Istio doesn't have any rules that allow any access
+    to workloads in the default namespace.
+
+1. Allow access to the `productpage` workload
+
+    The following command creates a `productpage-viewer` policy which allows
```
rcaballeromx: Please make sure the explanation comes before the call to action.
Readers need the context before you ask them to perform the action, not after.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+
+    The following command has the following effects:
+
+    * Creates a `details-viewer` policy which allows access with `GET` method to the `details` workload for
+      service account `cluster.local/ns/default/sa/bookinfo-productpage` (representing the `details` workload).
+
+    * Creates a `reviews-viewer` policy which allows access with `GET` method to the `reviews` workload for
+      service account `cluster.local/ns/default/sa/bookinfo-productpage` (representing the `reviews` workload).
+
+    Note that in the [setup step](#before-you-begin), we created the `bookinfo-productpage` service account
```
rcaballeromx: The section you are referencing does not mention any of this. The linked instructions do. Please link to the actual Bookinfo deployment instructions instead.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+    for the `productpage` workload. This `bookinfo-productpage` service account is the authenticated identify
```
rcaballeromx: There appears to be a typo but I ask, do the reader need to know this to complete the task? If your goal is for readers to make the connection between the SA and the value in `-from: -source: principals:` in the YAML file, then use the full value. Don't assume readers know how to complete the path with just the SA name. Also, you would need to explicitly call out and explain the fields involved or at the very least point readers towards a page with the explanation.
IMO, you don't need to go beyond my suggestion below though.
```suggestion
    for the `productpage` workload. Once you apply the configuration, only the `cluster.local/ns/default/sa/bookinfo-productpage` service account is authorized to access the `details` and `reviews` workloads through `GET` methods.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+    for the `productpage` workload.
```
rcaballeromx:
```suggestion
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+        to:
+        - operation:
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see the "Bookinfo Sample"
+    page with "Book Details" on the lower left part, and "Book Reviews" on the lower right part. However, in the "Book Reviews" section,
+    there is an error `Ratings service currently unavailable`. This is because "reviews" workload does not have permission to access
+    "ratings" workload. To fix this issue, you need to grant the `reviews` workload access to the `ratings` workload.
```
rcaballeromx:
```suggestion
    `ratings` workload. To fix this issue, you need to grant the `reviews` workload access to the `ratings` workload.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+    We will show how to do that in the next step.
```
rcaballeromx:
```suggestion
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+
+1. Allow access to the `ratings` workload
+
+    The following command creates a policy to allow the `reviews` workload to access the `ratings` workload. Note that in the
```
rcaballeromx: Use the explanations to lead readers towards the next step. See a suggestion on how to do so below:
```suggestion
   Next, we need to create a policy that allows the `reviews` workload to access the `ratings` workload. 
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+    [setup step](#before-you-begin), we created a `bookinfo-reviews` service account for the `reviews` workload. This
```
rcaballeromx: As mentioned above, this entire text is not only inaccurate, since the before you begin section does not contain the information, but also confusing. Do readers need to know that level of detail? I would change the text to make it easier to understand:
```suggestion
    Once you apply the configuration, the `cluster.local/ns/default/sa/bookinfo-reviews` service account is authorized to access the `ratings` workload through `GET` methods.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,259 @@ the product page, you can see the following sections:
...
+    service account is the authenticated identify for the `reviews` workload.
```
rcaballeromx:
```suggestion
```

content/en/docs/tasks/security/authz-http/index.md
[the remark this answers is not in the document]
yangminzhu: I know we're trying to automate the test and I wrote the automatic test for the HTTP page in istio.

The command (kubectl apply) is used everywhere in istio.io and for any normal user, I don't think it's causing any problems for them to understand the command which is separate from the policy being applied.

> It is also a lot easier for them to copy paste a command that references a sample file and it is less error prone. Not all readers use the copy button.

You need to use the copy button even if it's referring to a sample file. Also it's very common for user to actually look at the contents of the yaml, if it's a link, the user have to do one more click to open the page and switch back to see the corresponding explanation of the yaml.

> Lastly and regarding maintaining the samples: it is better and easier to maintain the samples in the repository than trying to maintain the samples embedded in the content.

From my experience of maintaining the task and sample, it is harder to maintain them in separate repos because if you want to make some change to the sample, you usually need to do it separately in two PRs, one to update the sample and the other to update the doc, there is a gap here and could also introduce inconsistent between the doc and the sample because they are updated separately.

Also this is not about code freeze, the "code freeze" is actually one of the examples of how it is making the maintenance harder, we have to keep the two repo in sync which is just hard to maintain.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: The new application I'm thinking is `httpbin` and `sleep` which is currently used in many places in both networking tasks and the authentication task. The `httpbin` is much smaller and gives better deterministic response which helps a lot in automating the tests, it's not maintained by Istio and is a public tool used widely in many other projects.

I will not change this in this PR but I think it's a better replacement for the Bookinfo in the future.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: Reworded to simplify, we don't need to refer to the Bookinfo page at all.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: I have updated the doc with new structure and this is probably not needed.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="remove_istio_authorization_policy_alternative.sh" >}}
-
-*   Disable Istio authorization:
-
-    {{< text_import file="examples/TestAuthorizationForHTTPServices.txt" snippet="disabling_istio_authorization.sh" >}}
+## Configure access control for workloads using HTTP traffic
+
+Using Istio, you can easily setup access control for {{< gloss "workload" >}}workloads{{< /gloss >}}
+in your mesh. This task shows you how to set up access control using Istio authorization.
+You will start with a simple `deny-all` policy that rejects all requests to the workload,
```
rcaballeromx:
```suggestion
First, you configure a simple `deny-all` policy that rejects all requests to the workload,
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+and then grant more access to the workload gradually and incrementally.
+
+1. Run the following command to create a policy `deny-all`, the policy has no rule so it
```
rcaballeromx:
```suggestion
1. Run the following command to create a `deny-all` policy. The `spec:` field of the policy has the
empty value `{}`. That value means that no traffic is permitted, effectively denying all requests.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+   cannot grant permission for any traffic. In other words, it denies all:
```
rcaballeromx:
```suggestion
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+      {}
+    EOF
+    {{< /text >}}
+
+    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
+    You should see `"RBAC: access denied"`. The error shows that the configured `deny-all` policy
+    is working as intended, and Istio doesn't have any rules that allow any access to
+    workloads in the mesh.
+
+    With Istio, you can easily configure namespace-level access control. You can configure how could workloads
```
rcaballeromx:
```suggestion
    With Istio, you can easily configure namespace-level access control. This approach enables you to configure  access control for workloads
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+    from a namespace access workloads in another namespace.
```
rcaballeromx:
```suggestion
    in a namespace to access workloads in another namespace.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+
+1. Run the following command to create a `bookinfo-viewer` policy that allows workloads in the `default`
+   and `istio-system` namespaces to access workloads in the `default` namespace with `GET` method:
```
rcaballeromx:
```suggestion
   and `istio-system` namespaces to access workloads in the `default` namespace with the `GET` method:
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+        - operation:
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+    Point your browser at Bookinfo's `productpage` (`http://$GATEWAY_URL/productpage`).
+    You should see the "Bookinfo Sample" page, with the "Book Details" section in the lower left part
+    and the "Book Reviews" section in the lower right part.
+
+1. Remove the policy with the following command before continuing:
```
rcaballeromx: Why do you ask readers to remove the policy? Going one step further, why are you applying the policy in the first place? Are you just trying to showcase namespace-level access control?

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+
+    {{< text bash >}}
+    $ kubectl delete authorizationpolicy.security.istio.io/bookinfo-viewer
+    {{< /text >}}
+
+1. Run the following command to create a `productpage-viewer` policy which allows
+   access to the `productpage` workload with `GET` method for all users and workloads:
```
rcaballeromx: Where in the configuration can a reader configure access to "all users and workloads"? Based on what I see in the yaml, it only applies to requests with the `productpage` label.

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
+    Now you should see the "Bookinfo Sample" page. But there are errors `Error fetching product details`
+    and `Error fetching product reviews` on the page.
+
+    These errors are expected because we have not granted the `productpage`
+    workload access to the `details` and `reviews` workloads. We will fix the
```
rcaballeromx:
```suggestion
    workload access to the `details` and `reviews` workloads. Next, you need to configure a policy to
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+    errors in the following steps.
```
rcaballeromx:
```suggestion
    grant access to those workloads.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+
+1. Run the following command to create a policy `details-viewer` to allow the `productpage`
```
rcaballeromx: Always put the name of the component before the noun and use `the` not `a` for specific objects. See my suggestion below and please make sure that format is followed throughout. 
```suggestion
1. Run the following command to create the `details-viewer` policy  to allow the `productpage`
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+   workload (represented by `principals: ["cluster.local/ns/default/sa/bookinfo-productpage"]`)
```
rcaballeromx: `represented by` might be well known jargon within the security team but not for our users. Please use simpler wording.
```suggestion
   workload, which issues requests using the `cluster.local/ns/default/sa/bookinfo-productpage` service account,
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+        - source:
+            principals: ["cluster.local/ns/default/sa/bookinfo-productpage"]
+        to:
+        - operation:
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+1. Run the following command to create a policy `reviews-viewer` to allow the `productpage` workload
+   (represented by `principals: ["cluster.local/ns/default/sa/bookinfo-productpage"]`) to access the
```
rcaballeromx:
```suggestion
   which issues requests using the `cluster.local/ns/default/sa/bookinfo-productpage` service account,
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+      - from:
+        - source:
+            principals: ["cluster.local/ns/default/sa/bookinfo-productpage"]
+        to:
+        - operation:
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+1. Run the following command to create a policy `reviews-viewer` to allow the `productpage` workload
```
rcaballeromx:
```suggestion
1. Run the following command to create a policy `reviews-viewer` to allow the `productpage` workload,
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+      - from:
+        - source:
+            principals: ["cluster.local/ns/default/sa/bookinfo-productpage"]
+        to:
+        - operation:
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now you should see the "Bookinfo Sample"
```
rcaballeromx:
```suggestion
    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`). Now, you should see the "Bookinfo Sample"
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+    page with "Book Details" on the lower left part, and "Book Reviews" on the lower right part. However, in the "Book Reviews" section,
+    there is an error `Ratings service currently unavailable`.
+
+    This is because "reviews" workload does not have permission to access `ratings` workload.
```
rcaballeromx:
```suggestion
    This is because the `reviews` workload doesn't have permission to access the `ratings` workload.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+    To fix this issue, you need to grant the `reviews` workload access to the `ratings` workload.
```
rcaballeromx:
```suggestion
    To fix this issue, you need to grant the `reviews` workload access to the `ratings` workload.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+    We will show how to do that in the next step.
```
rcaballeromx:
```suggestion
Next, we configure a policy to grant the `reviews` workload that access.
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+
+1. Run the following command to create a policy `ratings-viewer` to allow the `reviews` workload
```
rcaballeromx:
```suggestion
1. Run the following command to create the `ratings-viewer` policy to allow the `reviews` workload,
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,198 @@ the product page, you can see the following sections:
...
+   (represented by `principals: ["cluster.local/ns/default/sa/bookinfo-reviews"]`) to access the
```
rcaballeromx:
```suggestion
  which issues requests using the `cluster.local/ns/default/sa/bookinfo-reviews` service account, to access the
```

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: This is because the policy doesn't set `from` field in the `rule`. Updated the doc to make this clear.

content/en/docs/tasks/security/authz-http/index.md
yangminzhu: Yes, this is to showcase the namespace-level access control. We need to remove it otherwise the following workload-level policies doesn't make sense because the permission had already been granted in the namespace-level policy.

rcaballeromx: I would suggest you don't show the namespace-level access control here if it is not needed to complete the flow.
Another option would be to complete the flow using namespace-level access control, would that be doable?

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,168 @@ the product page, you can see the following sections:
...
+          app: productpage
+      rules:
+      - to:
+        - operation:
+            methods: ["GET"]
+    EOF
+    {{< /text >}}
+
+    Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
+    Now you should see the "Bookinfo Sample" page. But there are errors `Error fetching product details`
```
rcaballeromx:
```suggestion
    Now you should see the "Bookinfo Sample" page. However, you can see the following errors on the page:
    
     - `Error fetching product details`
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -32,178 +30,168 @@ the product page, you can see the following sections:
...
+    and `Error fetching product reviews` on the page.
```
rcaballeromx:
```suggestion
    - `Error fetching product reviews`
```

content/en/docs/tasks/security/authz-http/index.md
```diff
@@ -88,8 +88,11 @@ and then grant more access to the workload gradually and incrementally.
     {{< /text >}}
 
     Point your browser at the Bookinfo `productpage` (`http://$GATEWAY_URL/productpage`).
-    Now you should see the "Bookinfo Sample" page. But there are errors `Error fetching product details`
-    and `Error fetching product reviews` on the page.
+    Now you should see the "Bookinfo Sample" page. Now you should see the "Bookinfo Sample" page.
```
rcaballeromx: Duplicate text.

PR #5165, opened 2019-10-16
update authz concept page for the new v1beta1 policy

yangminzhu:
For https://github.com/istio/istio/issues/12394

Please provide a description for what this PR is for.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[X] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #5191, opened 2019-10-18
remove deprecated authz permissive task

yangminzhu:
For istio/istio#12394

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


yangminzhu:
@geeknoid Could you merge this one? We just don't need this task in 1.4 with the new authz policy, this will also fix the lint error in https://github.com/istio/istio.io/pull/5185 and https://github.com/istio/istio.io/pull/5165

geeknoid:
If folks have a bookmark to this page, where should they be redirected instead? Could you add an aliases: directive into that destination page?

yangminzhu:
> If folks have a bookmark to this page, where should they be redirected instead? Could you add an aliases: directive into that destination page?

The problem is we don't have any alternative page in 1.4, we don't support the contents described in this page in 1.4. Any suggestions for this case? 

Maybe we can keep this page but remove all the contents and make it clear it's not supported any more?

geeknoid:
Just a redirect to the current authz task is fine, it's better than a 404.

PR #5204, opened 2019-10-19
Operator docs fixes

nrjpoddar:
* Remove reference to downloading the Istio release as pre-requisite.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/standalone-operator/index.md
+++ b/content/en/docs/setup/install/standalone-operator/index.md
@@ -29,7 +29,7 @@ instead, which is a stable feature.
     {{< /text >}}
 
 {{< warning >}}
-This is only for demo usage and should not be used in production.
+This profile is only for demo usage and should not be used in production.
 {{< /warning >}}
 
 1. (Optionally) change profiles from the demo profile to one of the following profiles:
@@ -87,7 +87,8 @@ $ kubectl apply -f https://preliminary.istio.io/operator-profile-sds.yaml
 {{< /tab >}}
 
 {{< tab name="default" cookie-value="default" >}}
-This profile enables Istio's default settings. Run the following command to switch to this profile:
+This profile enables Istio's default settings which contains recommended
+production settings. Run the following command to switch to this profile:
 
 {{< text bash >}}
 $ kubectl apply -f https://preliminary.istio.io/operator-profile-default.yaml
--- a/content/en/docs/setup/install/standalone-operator/index.md
+++ b/content/en/docs/setup/install/standalone-operator/index.md
@@ -16,8 +16,6 @@ instead, which is a stable feature.
 
 ## Prerequisites
 
-1. [Download the Istio release](/docs/setup/#downloading-the-release).
-
 1. Perform any necessary [platform-specific setup](/docs/setup/platform-setup/).
 
 1. Check the [Requirements for Pods and Services](/docs/setup/additional-setup/requirements/).
@@ -29,6 +27,10 @@ instead, which is a stable feature.
     {{< text bash >}}
     $ kubectl apply -f https://preliminary.istio.io/operator.yaml
     {{< /text >}}
+
+{{< warning >}}
+This is only for demo usage and should not be used in production.
+{{< /warning >}}
 
 1. (Optionally) change profiles from the demo profile to one of the following profiles:
 
--- a/content/en/docs/setup/install/standalone-operator/index.md
+++ b/content/en/docs/setup/install/standalone-operator/index.md
@@ -22,7 +22,7 @@ instead, which is a stable feature.
 
 ## Installation steps
 
-1. Install Istio using the operator with the default profile:
+1. Install Istio using the operator with the demo profile:
 
     {{< text bash >}}
     $ kubectl apply -f https://preliminary.istio.io/operator.yaml


nrjpoddar:
@ostromart I think we should also add the following to enhance the docs (I can do that as part of PR if we have an agreement):

* Add profile for demo installation and make that the profile in docs as the first one users should use. The "default" profile seems to have substantial resource requirements which will make it difficult for users to try this out.
* Add demo-auth profile
* Add links to what these profiles are, may be link to Github?

nrjpoddar:
It is also very unclear as to which version of Istio are you installing with the operator.

nrjpoddar:
Makes sense. How does a user easily sets /find out what version of Istio was installed by operator? Currently, I see this image tag for pilot:

`gcr.io/istio-testing/pilot:latest`

ostromart:
Sorry, didn't see your suggestions before making comment above. Yes, I think the first installation that users see should be a demo install. 
If you look at operator.yaml, this points to the default profile, which is the default prod settings. 
We could point it to "profile: demo" which also installs prom + grafana, as well as being lighter on resources (we may want to tweak these a bit). 
We could have demo-auth although is there value in doing strict auth for the demo? Maybe just an default-auth-strict profile, with is default with strict for prod use?
Also, we should probably get rid of the permissive-auth tab and just move the default tab there in its place. I don't think users care much about that this is permissive auth at this point in trying out istio.

content/en/docs/setup/install/standalone-operator/index.md
```diff
@@ -53,7 +52,19 @@ $ kubectl apply -f https://preliminary.istio.io/operator-profile-demo.yaml
...
+
+Run the following command to switch to this profile:
+
+{{< text bash >}}
+$ kubectl apply -f https://preliminary.istio.io/operator-profile-default.yaml
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="demo auth" cookie-value="demo-auth" >}}
```
ostromart: maybe call this default strict auth? that's the profile it's pointing to.

nrjpoddar:
@linsun should we keep the demo-auth profile for installer. I think we should keep it as long as we mention it in the non-operator install section. @ostromart and I weren't sure, so thought we would ask you.

content/en/docs/setup/install/standalone-operator/index.md
```diff
@@ -16,44 +16,36 @@ instead, which is a stable feature.
...
-
 1. Perform any necessary [platform-specific setup](/docs/setup/platform-setup/).
 
 1. Check the [Requirements for Pods and Services](/docs/setup/additional-setup/requirements/).
 
 ## Installation steps
 
-1. Install Istio using the operator with the default profile:
+1. Install Istio using the operator with the demo profile. This is only for demo
+   usage and **should not be used in production**.
```
sdake: I think a callout would be better than this bolded warning.

sdake:
> Sorry, didn't see your suggestions before making comment above. Yes, I think the first installation that users see should be a demo install.
> If you look at operator.yaml, this points to the default profile, which is the default prod settings.
> We could point it to "profile: demo" which also installs prom + grafana, as well as being lighter on resources (we may want to tweak these a bit).
> We could have demo-auth although is there value in doing strict auth for the demo? Maybe just an default-auth-strict profile, with is default with strict for prod use?
> Also, we should probably get rid of the permissive-auth tab and just move the default tab there in its place. I don't think users care much about that this is permissive auth at this point in trying out istio.

Before we completely restructure these docs, I'd like to hear @frankbu 's comments in the docs wg on this topic.  He has taken a look at a related PR, and now that he understands the goals of these pages, he may have some longer term thoughts on how to structure this work.

Cheers
-steve

sdake:
> @linsun should we keep the demo-auth profile for operator. I think we should keep it as long as we mention it in the non-operator install section. @ostromart and I weren't sure, so thought we would ask you.

Master currently has demo as the initially installed profile.  I think demo-auth adds value.  We should document every profile we implement, and longer term for 1.5, the profiles should be tied directly to tasks in the documentation.

cc / @rcaballeromx @frankbu 

@rcaballeromx I'll try to prepare a preliminary plan for installation in 1.5 documentation, but would prefer to defer to you for tidying up the plan and execution if you are willing.


Cheers
-steve

nrjpoddar:
> > @linsun should we keep the demo-auth profile for operator. I think we should keep it as long as we mention it in the non-operator install section. @ostromart and I weren't sure, so thought we would ask you.
> 
> Master currently has demo as the initially installed profile. I think demo-auth adds value. We should document every profile we implement, and longer term for 1.5, the profiles should be tied directly to tasks in the documentation.
> 
> cc / @rcaballeromx @frankbu
> 
> @rcaballeromx I'll try to prepare a preliminary plan for installation in 1.5 documentation, but would prefer to defer to you for tidying up the plan and execution if you are willing.
> 
> Cheers
> -steve

Thanks @sdake for your review. IIUC, I am going to rebase on top of latest master and hopefully most of the changes around demo profile won't be needed for this PR as @linsun already addresses them. I will keep the demo-auth profile and other changes if needed as part of this PR and send it your way again.

frankbu:
I added some questions/concerns about the operator install docs to the doc WG agenda.

https://docs.google.com/document/d/1RGb0NOp0J9QSIrMrZ6wokl16RbSfKHy-6NyTpOQdPa8/edit#

ostromart:
Thanks Frank. The operator docs do need some attention, we can discuss in
the Thu meeting. I read over the agenda and have some thoughts.

PR #5226, opened 2019-10-22
Fix protocol sniffing docs

howardjohn:
These were outdated

Since I marked automated detection as experimental (which it is), I also moved it to the bottom, since it seemed to make sense to have the stable docs on top


content/en/docs/ops/traffic-management/protocol-selection/index.md
```diff
@@ -52,3 +45,10 @@ spec:
   - number: 80
     name: http-web
 {{< /text >}}
+
+## Automatic Protocol Selection (Experimental)
```
geeknoid: Could you please correct all headings on this page to only have the first letter capitalized, which our standard in docs.

Thanks.

PR #5236, opened 2019-10-23
Remove one last mention of demo-auth

howardjohn:


pitlv2109:
Are we replacing it with something else? i.e. if a user wants to try mtls without sds?

pitlv2109:
Ok got it, just set the flag. Why are we removing the file? From a UX point of view, having the file is much easier to try it out, not all users know how to pass in helm flags.

PR #5253, opened 2019-10-24
Add `and`  to description

brian-avery:
This shows up as 

What is Istio?
Introduces Istio, the problems it solves, its high-level architecture and design goals.

on https://preliminary.istio.io/docs/concepts/

This PR fixes it to show up as:

Introduces Istio, the problems it solves, and its high-level architecture and design goals.

### DIFF
--- a/content/en/docs/concepts/what-is-istio/index.md
+++ b/content/en/docs/concepts/what-is-istio/index.md
@@ -1,6 +1,6 @@
 ---
 title: What is Istio?
-description: Introduces Istio, the problems it solves, and its high-level architecture and design goals.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/what-is-istio/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: What is Istio?
-description: Introduces Istio, the problems it solves, its high-level architecture and design goals.
+description: Introduces Istio, the problems it solves, and its high-level architecture and design goals.
```
geeknoid: There's already an "and" in there. I think maybe that should be instead:
```suggestion
description: Introduces Istio, the problems it solves, its high-level architecture, and its design goals.
```

PR #3184, opened 2019-01-31
Update monitor port translation

ericvn:
Documentation updates in translated files that coincide with istio/istio#11421 which changed the default monitoring port from 9093 to 15014.

The English files are updated in PR https://github.com/istio/istio.io/pull/3183

WIP until that PR is merged.

PR #3183, opened 2019-01-31
Change default monitoring port

ericvn:
Documentation updates that coincide with https://github.com/istio/istio/pull/11421 which changed the default monitoring port from 9093 to 15014.

WIP until that PR is merged.

PR #5260, opened 2019-10-26
s/10514/15014/

ericbuehl:
Admittedly this is a naive suggestion as I don't know the history of port 10514.  Perhaps it used to be used for something?  However, in all the instances I found it seems to just be a typo that should really refer to port 15014.


ericbuehl:
On closer investigation, this does indeed appear to be a typo introduced when the port changed from 9093 -> 15014 via https://github.com/istio/istio.io/pull/3183 and https://github.com/istio/istio.io/pull/3184

geeknoid:
Hi, could you please rebase this PR so we can merge it?

Thanks.

PR #5261, opened 2019-10-27
add port name "tls" to both Service and Endpoints in Kubernetes Services for Egress task

vadimeisenbergibm:
otherwise the service is treated as HTTP, causing an error

Fixes https://github.com/istio/istio/issues/18208.


ericvn:
Verified the page now works on 1.4-beta.0

PR #5281, opened 2019-10-29
Helm/215.warning

sdake:


frankbu:
Does it make sense to change anything in this doc, now that it's deprecated. I added a warning on the top that basically says don't use these instructions anymore: https://preliminary.istio.io/docs/setup/install/helm/

sdake:
@frankbu missed that - this was a pr sitting around as an outstanding PR for several weeks.

subsumed by work from @frankbu .

PR #5313, opened 2019-11-01
Auto mTLS user guide doc on istio.io

incfly:
Auto mTLS user guide doc on istio.io.

A separate task page under Istio security section to explain how to use auto mutual TLS, the impact, consequences, instructions.

https://github.com/istio/istio/issues/14524


incfly:
@rcaballeromx This is a new doc task. let me know if you have quick early high level feedback requiring big changes first, so that we can worry about fine grane polishing later. 

I should be able to get a review from @diemtvu soon on technical part.

Thanks.

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,276 @@
...
+## Before you begin
+
+* Understand Istio [authentication policy](/docs/concepts/security/#authentication-policies) and related
+[mutual TLS authentication](/docs/concepts/security/#mutual-tls-authentication) concepts.
+
+* Have a Kubernetes cluster with Istio installed (e.g use `install/kubernetes/istio-demo.yaml` as described in
+[installation steps](/docs/setup/install/kubernetes/#installation-steps), set `global.mtls.enabled` to false  and `global.mtls.auto` to true using [Helm](/docs/setup/install/helm/)). For example,
+
+{{< text bash >}}
+$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system  \
```
frankbu: Helm is deprecated. Need to use istioctl manifest.

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,276 @@
...
+  --set global.mtls.enabled=false  --set global.mtls.auto=true \
+  --values install/kubernetes/helm/istio/values-istio-demo.yaml | kubectl apply -f -
+{{< /text >}}
+
+## Instructions
+
+### Setup
+
+Our examples use three namespaces, `foo`, `mixed`, and `legacy`.
```
frankbu: Why do you need all these namespaces? Seems to me that this task doesn't do anything with namespaces, they are just confusing the message. All you need for this task is a client and two httpbin services (one with sidecar and one without)? Am I missing something?

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,276 @@
...
+
+{{< text bash >}}
+$ kubectl create ns mixed
+$ kubectl apply -f <(istioctl kube-inject -f @samples/httpbin/httpbin.yaml@) -n mixed
+$ kubectl apply -f <(istioctl kube-inject -f @samples/sleep/sleep.yaml@) -n mixed
+$ cat <<EOF | kubectl apply -n mixed -f -
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: httpbin
```
frankbu: This is the same deployment as on line 49, isn't it? Won't this replace the previous one, so the pod with sidecar will be deleted?

Why do we need both anyway? What special is it demonstrating?

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,276 @@
...
+{{< text bash >}}
+$ cat <<EOF | kubectl apply -n foo -f -
+apiVersion: "authentication.istio.io/v1alpha1"
+kind: "Policy"
+metadata:
+  name: "httpbin"
+spec:
+  targets:
+  - name: httpbin
+EOF
```
frankbu: Would be good to add the fields that explicitly disable, rather than relying on default.

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,276 @@
...
+  name: "httpbin"
+spec:
+  targets:
+  - name: httpbin
+  peers:
+  - mtls: {}
+EOF
+{{< /text >}}
+
+Now the requests from the clients without sidecar start to fail.
```
frankbu: You should explicitly say which case(s) fail and exactly why (what's happening now).

The important thing to show (and explain) in this task is what is different because we're using the new auto mTLS feature.

content/en/docs/tasks/security/automtls/index.md
incfly: You're right. I want to illustrate the idea that  auto mTLS works for all the cases when people adopt istio

- server all have sidecars
- server partially have sidecars
- strict or permissive or plain text for the authn policy

Maybe I should emulate a CUJ of migrating workloads to Istio? Or creating different deployments (inline yaml?) at the same time?

The only issue is that we already have a `mtls-migration` task page, which only covers `PERMISSIVE` case, not auto mTLS.

What do you think?

content/en/docs/tasks/security/automtls/index.md
incfly: This deployment is created inline, without sidecar injected, i made the changes on the label. This won't replace previous deployment (which has sidecar and different app labels).

But you're right, probably i need to modify the deployment name as well.

I'm trying to demonstrate customers don't have to worry about configuration with mTLS, even when their server workloads are partially sidecar injected. Compared with `foo` namespace

content/en/docs/tasks/security/automtls/index.md
incfly: We no longer have the plaintext option in authentication policy. This is the only way users can disable mTLS today.

content/en/docs/tasks/security/automtls/index.md
incfly: done.

content/en/docs/tasks/security/automtls/index.md
incfly: offline discussion, we still use namespace, but

- sleep only in legacy and full
- more explaination on delta behavior of the auto mtls.

content/en/docs/tasks/security/automtls/index.md
incfly: modified the name to be `httpbin-nosidecar` this should be more obvious now.

content/en/docs/tasks/security/automtls/index.md
incfly: done. explaining the delta with auto mtls client `sleep.full` as well.

incfly:
This is ready for another look, thanks! @frankbu

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,289 @@
...
+httpbin-dcd949489-5fsbs    2/2     Running   0          12s
+NAME                       READY   STATUS    RESTARTS   AGE
+httpbin-54f5bb4957-lzxlg   1/1     Running   0          6s
+sleep-74564b477b-vb6h4     1/1     Running   0          4s
+{{< /text >}}
+
+You should also verify that there is a default mesh authentication policy in the system, which you can do as follows:
+
+{{< text bash >}}
+$ kubectl get policies.authentication.istio.io --all-namespaces
```
diemtvu: Not sure if we need to mention this. It is internal policy for grafana and may confuse users.

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,289 @@
...
+sleep-58d6644d44-gb55j    2/2     Running   0          38s
+NAME                       READY   STATUS    RESTARTS   AGE
+httpbin-6f6fc94fb6-8d62h   1/1     Running   0          10s
+httpbin-dcd949489-5fsbs    2/2     Running   0          12s
+NAME                       READY   STATUS    RESTARTS   AGE
+httpbin-54f5bb4957-lzxlg   1/1     Running   0          6s
+sleep-74564b477b-vb6h4     1/1     Running   0          4s
+{{< /text >}}
+
+You should also verify that there is a default mesh authentication policy in the system, which you can do as follows:
```
diemtvu: Should we also mention the default policy is in `PERMISSIVE` mode. E.g change the command below to show the actual policy content:
`kubectl get meshpolicies -o yaml | grep mode`

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,289 @@
...
+default   3m
+{{< /text >}}
+
+Last but not least, verify that there are no destination rules that apply on the example services. You can do this by checking the `host:` value of
+ existing destination rules and make sure they do not match. For example:
+
+{{< text bash >}}
+$ kubectl get destinationrules.networking.istio.io --all-namespaces -o yaml | grep "host:"
+    host: kubernetes.default.svc.cluster.local
+    host: '*.local'
```
diemtvu: This actually match every service. Are you sure this shows up in the default setup?

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,289 @@
...
+1. The workload without sidecar for `httpbin.partial` and workloads of `httpbin.legacy` can only serve
+plain text traffic.
+
+Automatic mutual TLS configures the client, `sleep.full`, to send mutual TLS to the first type of
+workloads and plain text to the second type.
+
+You can verify the reachability as:
+
+{{< text bash >}}
+$ for from in "full" "legacy"; do for to in "full" "partial" "legacy"; do kubectl exec $(kubectl get pod -l app=sleep -n ${from} -o jsonpath={.items..metadata.name}) -c sleep -n ${from} -- curl "http://httpbin.${to}:8000/ip" -s -o /dev/null -w "sleep.${from} to httpbin.${to}: %{http_code}\n"; done; done
```
diemtvu: May be use '/get' or '/headers' that will show the identity in the header if mTLS is used.

content/en/docs/tasks/security/automtls/index.md
incfly: This is copied from old doc. I think it ensure that customer dont have some leftover of their own policy, right?

content/en/docs/tasks/security/automtls/index.md
incfly: Right. This is a regression in the new `istioctl manifest` compared with the old demo installer profile. I filed https://github.com/istio/istio/issues/18654 for tracking. But this is accurate at this moment.

*.local DR has a DISABLED TLS mode.

content/en/docs/tasks/security/automtls/index.md
incfly: good point. I didn't know this knob before.

content/en/docs/tasks/security/automtls/index.md
incfly: done.

content/en/docs/tasks/security/automtls/index.md
incfly: done. the oneline is really long line now, but at least we can show mtls from the http output.

incfly:
Per @diemtvu  suggestion, use new `httpbin:8000/headers` to get the client certificate illustrating mtls is used actually

frankbu:
@incfly Thanks for clarifying things. Technically, it looks much better now! There are still lots of English wording and grammar problems that should be fixed. Not sure if it would be easier to merge it now and then fix the grammar in a followup PR, instead of trying to suggest many more changes.

@rcaballeromx @geeknoid WDYT?

incfly:
@frankbu Thanks for the feedback!

Yes, to clarify, regarding the timeline, we definitely need this as *minimal viable docs* for this important security feature in 1.4 before release is cut.

Let me know if anything I can do on my side to expedite the doc process. @geeknoid @rcaballeromx , thanks.

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,375 @@
...
+Istio tracks the server workloads migrated to Istio sidecar, and configures client sidecar to send mutual TLS traffic to those workloads automatically, and send plain text traffic to workloads
+without sidecars. This allows you to adopt Istio mutual TLS incrementally with minimal manual configuration.
+
+## Before you begin
+
+* Understand Istio [authentication policy](/docs/concepts/security/#authentication-policies) and related
+[mutual TLS authentication](/docs/concepts/security/#mutual-tls-authentication) concepts.
+
+* Have a Kubernetes cluster with Istio installed (e.g use `install/kubernetes/istio-demo.yaml` as described in
+[installation steps](/docs/setup/install/kubernetes/#install-the-demo-profile), set `global.mtls.enabled` to false and `global.mtls.auto` to true. For example,
```
frankbu:
```suggestion
For example, using the `demo` configuration profile:
```

content/en/docs/tasks/security/automtls/index.md
```diff
@@ -0,0 +1,375 @@
...
+
+Istio tracks the server workloads migrated to Istio sidecar, and configures client sidecar to send mutual TLS traffic to those workloads automatically, and send plain text traffic to workloads
+without sidecars. This allows you to adopt Istio mutual TLS incrementally with minimal manual configuration.
+
+## Before you begin
+
+* Understand Istio [authentication policy](/docs/concepts/security/#authentication-policies) and related
+[mutual TLS authentication](/docs/concepts/security/#mutual-tls-authentication) concepts.
+
+* Have a Kubernetes cluster with Istio installed (e.g use `install/kubernetes/istio-demo.yaml` as described in
```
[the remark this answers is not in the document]
incfly: applied.

incfly:
Thanks a lot!

PR #5319, opened 2019-11-02
Address rendering glitch on the istioctl reference page.

geeknoid:


geeknoid:
Not perfect, but I think the best we can hope for given the constraint imposed by having a very long non-wrappable string in the tables. 

WDYT?

Issue #5343, opened 2019-11-04
tablegen.py needs a rerun


sdake:
Run tablegen.py for master

Related to this issue #5259

geeknoid:
tablegen-diff.py also needs a rerun...

howardjohn:
@geeknoid is this done by https://preliminary.istio.io/docs/reference/config/installation-options/?

geeknoid:
This work is still pending: https://github.com/istio/istio.io/pull/5344

sdake:
I am unable to make forward progress on this issue without a significant investment of time (writing a yaml parser... do we really need another one?). While it is unfortunate, any fields that people think are relevant for helm documentation should be manually updated.

Using the in-repo code that @johnma14 wrote should be workable for istiod. Please file any new bugs for istiod helm reference documentation and assign it to @johnma14.

PR #5344, opened 2019-11-04
Update tablegen documentation

sdake:
Closes https://github.com/istio/istio.io/issues/5343

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -63,8 +63,11 @@ update_operator_yamls:
 update_examples:
 	@scripts/grab_examples.sh $(SOURCE_BRANCH_NAME)
 
-update_all: update_ref_docs update_operator_yamls update_examples
+update_helm_table:
+	@python3 scripts/tablegen.py
+
+update_all: update_ref_docs update_operator_yamls update_examples update_helm_table
 
 include common/Makefile.common.mk
 
-.PHONY: gen build build_nominify opt clean_public clean lint serve netlify_install netlify netlify_archive archive update_ref_docs update_operator_yamls update_examples update_all
+.PHONY: gen build build_nominify opt clean_public clean lint serve netlify_install netlify netlify_archive archive update_ref_docs update_operator_yamls update_examples update_helm_tables update_all
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ ifeq ($(BUILD_WITH_CONTAINER),1)
 export TARGET_OUT = /work/out/$(TARGET_OS)_$(TARGET_ARCH)
 CONTAINER_CLI ?= docker
 DOCKER_SOCKET_MOUNT ?= -v /var/run/docker.sock:/var/run/docker.sock
-IMG ?= gcr.io/istio-testing/build-tools:2019-10-24T14-05-17
+IMG ?= docker.io/sdake/build-tools:master-2019-11-10T18-25-28
 UID = $(shell id -u)
 GID = `grep docker /etc/group | cut -f3 -d:`
 PWD = $(shell pwd)
--- a/content/en/docs/reference/config/installation-options/index.md
+++ b/content/en/docs/reference/config/installation-options/index.md
@@ -32,108 +32,6 @@ for a detailed summary of the option changes between release 1.2 and release 1.3
 <!-- Run `make update_helm_table` to generate this table -->	
 
 <!-- AUTO-GENERATED-START -->
-## `certmanager` options
-
-<table>
-    <thead>
-    <tr>
-        <th>Key</th>
-        <th>Default Value</th>
-        <th>Description</th>
-    </tr>
-    </thead>
-    <tbody>
-
-            <tr>
-                <td>certmanager.enabled</td>
-                <td>False</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>certmanager.replicaCount</td>
-                <td>1</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>certmanager.hub</td>
-                <td>quay.io/jetstack</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>certmanager.image</td>
-                <td>cert-manager-controller</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>certmanager.tag</td>
-                <td>v0.8.1</td>
-                <td></td>
-            </tr>
-
-    </tbody>
-</table>
-
-## `galley` options
-
-<table>
-    <thead>
-    <tr>
-        <th>Key</th>
-        <th>Default Value</th>
-        <th>Description</th>
-    </tr>
-    </thead>
-    <tbody>
-
-            <tr>
-                <td>galley.enabled</td>
-                <td>True</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>galley.replicaCount</td>
-                <td>1</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>galley.rollingMaxSurge</td>
-                <td>100%</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>galley.rollingMaxUnavailable</td>
-                <td>25%</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>galley.image</td>
-                <td>galley</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>galley.enableServiceDiscovery</td>
-                <td>False</td>
-                <td></td>
-            </tr>
-
-            <tr>
-                <td>galley.enableAnalysis</td>
-                <td>False</td>
-                <td></td>
-            </tr>
-
-    </tbody>
-</table>
-
 ## `gateways` options
 
 <table>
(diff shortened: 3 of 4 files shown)


geeknoid:
Do we also need to update the diff file? That is, content/en/news/2019/announcing-1.4/helm-changes/index.md?

sdake:
@geeknoid possibly - I am doing these one file at a time as the script is not well maintained.  Further complicating issues is a lack of cross repo gating between istio/istio/install and istio.github.io - permitting people to check in invalid text in values.yaml. I don't think it is worth our time to make this tool more resilient nor add cross-repo gating as hopefully helm will have one big git rm -rf from the repos.

Cheers
-steve

sdake:
lovely rendering: https://deploy-preview-5344--preliminary-istio.netlify.com/docs/reference/config/installation-options/

geeknoid:
Ouch. Table syntax in markdown is pretty strict. If it's not perfect, you get that kind of rendering.

Something I noticed is also this text:

> Refer to Installation Options Changes for a detailed summary of the option changes between release 1.2 and release 1.3.

In the page needs updated text and link. We could try to make this automatically updated, but it's probably not worth the work given Helm's ultimate fate.

nrjpoddar:
@sdake We need to regenerate Helm reference docs for release-1.3 branch too. I know there's one discrepancy with protocol sniffing i.e.

```diff
- | `pilot.enableProtocolSniffing` | `false` | `if protocol sniffing is enabled. Default to false.` |
+ | `pilot.enableProtocolSniffingForOutbound` | `true` | `if protocol sniffing is enabled for inbound` |
+ | `pilot.enableProtocolSniffingForInbound` | `false` | `Resources for a small pilot install` |
```

sdake:
> @sdake We need to regenerate Helm reference docs for release-1.3 branch too. I know there's one discrepancy with protocol sniffing i.e.
> 
> ```diff
> - | `pilot.enableProtocolSniffing` | `false` | `if protocol sniffing is enabled. Default to false.` |
> + | `pilot.enableProtocolSniffingForOutbound` | `true` | `if protocol sniffing is enabled for inbound` |
> + | `pilot.enableProtocolSniffingForInbound` | `false` | `Resources for a small pilot install` |
> ```

if by we you mean you, feel free :)

geeknoid:
@nrjpoddar The 1.3 docs will disappear on the 12th and move to the archive, so I don't think we want to spend any time fixing tooling in that side of things. If you want to submit a PR in the release-1.3 branch to manually update these files, that's fine, but I don't know whether it's worthwhile.

nrjpoddar:
Thanks @geeknoid & @sdake, your comments make sense. I will generate a manual diff, I was unclear if manual edits were acceptable or not.

nrjpoddar:
Added https://github.com/istio/istio.io/pull/5350

sdake:
`:` characters in comments are being processed incorrectly.  The result is commenting does not match with their contents.

Makefile.core.mk
```diff
@@ -65,6 +65,9 @@ update_examples:
 
 update_all: update_ref_docs update_operator_yaml update_examples
```
geeknoid:
```suggestion
update_all: update_ref_docs update_operator_yamls update_examples update_install_docs
```

Makefile.core.mk
```diff
@@ -65,6 +65,9 @@ update_examples:
...
 
+update_install_docs:
+	@python scripts/tablegen.py
+
 include common/Makefile.common.mk
 
 .PHONY: gen build build_nominify opt clean_public clean lint serve netlify_install netlify netlify_archive archive update_ref_docs update_operator_yamls update_examples update_all
```
geeknoid:
```suggestion
.PHONY: gen build build_nominify opt clean_public clean lint serve netlify_install netlify netlify_archive archive update_ref_docs update_operator_yamls update_examples update_helm_tables update_all
```

Makefile.core.mk
```diff
@@ -65,6 +65,9 @@ update_examples:
 
 update_all: update_ref_docs update_operator_yaml update_examples
 
+update_install_docs:
```
geeknoid:
```suggestion
update_helm_tables:
```

sdake:
On PTO until Nov 10, Traveling Nov 15 for KubeCon, this is at risk for not making the release.

geeknoid:
I might try and write a Go program to take care of this, and probably output HTML rather than markdown, which would make handling the descriptions a lot simpler.

sdake:
> I might try and write a Go program to take care of this, and probably output HTML rather than markdown, which would make handling the descriptions a lot simpler.

Prior to writing this the first time, I looked high and low for a golang library that would keep the comments intact while parsing YAML into data structures.  Unfortunately I could find no such library.  It was a stretch to find a python library at the time (a few years ago).  I did find ruamel which maintains comments, but was very rough around the edges.  Ruamel has matured and has subsumed PyYAML as the cononiical YAML parsing library.  The greatest feature of Ruamel is round tripping - which permits keeping track of comments.

This new version uses the parser objects to keep track of the comments directly instead of (as prior) parsing part of the yaml on its own.  Thesee parser objects are new additions to the yaml pasing in 0.15 of Ruamel.

I didn't expect to be home until this evening, but my Father decided to head home this morning so I do have some time to take a look at this work today.

Outputting directly to HTML is viable, but I'd like  to just be able to build a dictionary with a data structure of key->(value, description) for easy output into any DSL (be it YAML or HTML).

If you can let me know if you are able to find a newer golang yaml library that preserves comments, that would be a preferrable approach I think, so we have a smaller language footprint across our tools (Golang,and shell and a bit of C:).  Makes maintenance easier for the next poor individual that has to look at our code 😁 

Cheers
-steve

sdake:
Haven't tried this, but this might be an option: https://github.com/kubernetes/kubernetes/issues/15157#issuecomment-403151628

sdake:
@geeknoid here is a fork of go-yaml which may support comment tokens..

https://github.com/mozilla-services/yaml

Haven't tried using it though.

Cheers
-steve

howardjohn:
One thing to consider is with the operator we have proto so we don't need to do all this yaml stuff

geeknoid:
@sdake I'll take a look, that might just be enough.

@howardjohn Do you think we can trivially and mechanically start from the protos and generate this helm-friendly stuff? I wrote a Go-based values.yaml->md converter yesterday in less than an hour to replace the Python code, so it's pretty easy. I was just missing support for retrieving the comments from the YAML, which apparently Steve's link will let me do.

nrjpoddar:
We use `https://github.com/kubernetes/gengo` in Istio client-go to parse files and create custom generators based on specific tags. This tool automatically tracks comments and bunch of other Go related things. I'm not sure how generic it is to use for non-go related file parsing.

Here's an example of its usage: https://github.com/istio/tools/tree/master/cmd/kubetype-gen Might be an overkill for this effort but the k8s library is pretty powerful, so just wanted to mention it here.

geeknoid:
@sdake Nope, that update won't work. Most of the Golang YAML parsers insist on just unmarshaling to a struct. What I need is more of a DOM or AST-like model so I don't need to know the structure of the file and can just walk it dynamically. The only parser I found that does this for Go is go-gypsy, but it doesn't do comments.

@nrjpoddar I'll check it out, maybe it'll fit the bill.

howardjohn:
@geeknoid I am not 100% sure I understand your question, so I will just dump all context I have here

We have the two install methods, helm and operator, with corresponding doc pages on configuration options. This one for helm, and the operator one, which is generated from proto: https://preliminary.istio.io/docs/reference/config/istio.operator.v1alpha12.pb/.

These are fairly similar, although there are some differences, so its not just that one is yaml and one is proto -- they represent different objects.

One thing I have found is the operator proto docs are not great. The doc gen works well for our other CRDs since they are small, but there are like 100s of objects in the ICP proto. It would be nice to display them like the helm one, or my preference if there was a way to display it something like:

```
# doc for some field
someField:
  # doc for some int
  someInt: int
  someString: int
  someListOfStrings: [string]
```

Then its really easy to map that to your own config file since you can just copy the whole object over, replace what you want and delete what you don't? Maybe this is just reinventing a new schema though :slightly_smiling_face:

sdake:
about this PR, I have a comment parser using ruamel that works well.  The code generates HTML as to not have to deal with making the content MD compliant.

Our values.yaml files are not properly docified in my opinion.  They have stuff that is before and after comments that may or may not be related to the target key/value pair.

What I have is pretty close to working and far simpler then the last trainwreck of a home built yaml parser.  ruamel is well maintained as a dependency and is the default in python3 for any type of real development (In the python world, once a library hits the standard lib it's api is frozen, so there is no improvement on the feature set)..

I am not anHTML wizard so not sure which tags to add to the tables.  If someone can help out here, we can merge this work as is.  Overall it works pretty well.

Cheers
-steve

sdake:
comments aren't quite right - but pretty close (this is only the main values.yaml - although adding the others is trivial, I just wanted to see if my hand edited HTML rendered properly.  

https://deploy-preview-5344--preliminary-istio.netlify.com/docs/reference/config/installation-options/

sdake:
> One thing to consider is with the operator we have proto so we don't need to do all this yaml stuff

@howardjohn while I agree this yaml to html conversion is low value in the future, at present it seems very high value until a full transition is made.  I would havce liked to have dumped this code from the repo in 1.4, but it does need to be maintained until helm is gone.

Cheers
-steve

Makefile.core.mk
sdake: addressed.

Makefile.core.mk
sdake: addressed.

geeknoid:
This is close, but still needs some TLC:

- When processing the description, you need to remove leading # symbols.

- You went a little too extreme on the html front. Instead of producing a full HTML document, I wanted to just generate an HTML table and insert it into the same markdown document as we had before. So instead of inserting a markdown table into the markdown doc, insert an HTML table into the markdown doc.

sdake:
> This is close, but still needs some TLC:
> 
> * When processing the description, you need to remove leading # symbols.
> * You went a little too extreme on the html front. Instead of producing a full HTML document, I wanted to just generate an HTML table and insert it into the same markdown document as we had before. So instead of inserting a markdown table into the markdown doc, insert an HTML table into the markdown doc.

`#` comment stripping is done

How do I insert HTML into MD?  I know how to generate the html table and I know how to insert it, but not sure what tags in md are needed so MD understands it is reading HTML now instead of MD.  Any examples?

Cheers
-steve

sdake:
I'm not sure how to proceed with this work. The library this code is dependent on doesn't properly round trip comments, and there is no golang library to do the same. Given the istiod plans, and the fact istiod is auto-generated from protos, the most expedient process would be to manually update the comments and remove this automation that has never really worked well.

dgn:
@sdake this is marked as P0 for 1.5.0 - is this still required?

sdake:
@nrjpoddar manual edits are the path forward for release-1.5 changes. For release 1.6, I'll speak with @johnma14 about writing a simple golang generator that leaves out the description field for ... helmv3.

sdake:
@dgn This is no longer P0 for 1.5. For 1.6, we should start afresh with a golang generator integrated into the build tools container that @johnma14 suggested to me.

PR #5387, opened 2019-11-05
zh-translation:/content/zh/faq/security/check-policy.md

qunqiang:
ref: https://github.com/servicemesher/istio-official-translation/issues/1021


rootsongjc:
> /pushed

请不要在这里回复 pushed，而是到你认领 issue 的地方回复。

PR #5391, opened 2019-11-05
Fix references to samples/operator files

jwendell:
Closes https://github.com/istio/istio/issues/18441

### DIFF
--- a/content/en/docs/setup/install/istioctl/index.md
+++ b/content/en/docs/setup/install/istioctl/index.md
@@ -192,7 +192,7 @@ You can show the differences in the generated manifests between the default prof
 
 {{< text bash >}}
 $ istioctl manifest generate > 1.yaml
-$ istioctl manifest generate -f samples/pilot-k8s.yaml > 2.yaml
+$ istioctl manifest generate -f samples/operator/pilot-k8s.yaml > 2.yaml
 $ istioctl manifest diff 1.yam1 2.yaml
 {{< /text >}}
 
@@ -233,7 +233,7 @@ Alternatively, the `IstioControlPlane` configuration can be specified in a YAML 
 `istioctl` using the `-f` option:
 
 {{< text bash >}}
-$ istioctl manifest apply -f samples/pilot-k8s.yaml
+$ istioctl manifest apply -f samples/operator/pilot-k8s.yaml
 {{< /text >}}
 
 {{< tip >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/operator/index.md
```diff
@@ -192,7 +192,7 @@ You can show the differences in the generated manifests between the default prof
 
 {{< text bash >}}
 $ istioctl manifest generate > 1.yaml
-$ istioctl manifest generate -f samples/pilot-k8s.yaml > 2.yaml
+$ istioctl manifest generate -f samples/operator/pilot-k8s.yaml > 2.yaml
```
frankbu: Use @ so the files are clickable links in the doc.

```suggestion
$ istioctl manifest generate -f @samples/operator/pilot-k8s.yaml@ > 2.yaml
```

content/en/docs/setup/install/operator/index.md
```diff
@@ -233,7 +233,7 @@ Alternatively, the `IstioControlPlane` configuration can be specified in a YAML
 `istioctl` using the `-f` option:
 
 {{< text bash >}}
-$ istioctl manifest apply -f samples/pilot-k8s.yaml
+$ istioctl manifest apply -f samples/operator/pilot-k8s.yaml
```
frankbu:
```suggestion
$ istioctl manifest apply -f @samples/operator/pilot-k8s.yaml@
```

content/en/docs/setup/install/operator/index.md
```diff
@@ -381,12 +381,20 @@ spec:
...
+            effect: NoSchedule
+          - key: CriticalAddonsOnly
+            operator: Exists
 {{< /text >}}
 
 Use `manifest apply` to apply the modified settings to the cluster:
 
 {{< text syntax="bash" repo="operator" >}}
-$ istioctl manifest apply -f @samples/pilot-k8s.yaml@
+$ istioctl manifest apply -f samples/operator/pilot-k8s.yaml
```
frankbu:
```suggestion
$ istioctl manifest apply -f @samples/operator/pilot-k8s.yaml@
```

content/en/docs/setup/install/operator/index.md
jwendell: @frankbu  They are not clickable. Those files do not exist in istio repo, they are from the operator repo which are included in the release tarball.

brian-avery: This brings up an interesting question. You can reference other repos in the links as follows (https://istio.io/about/contribute/creating-and-editing-pages/#links-to-github-files):
```
{{< text syntax="bash" repo="operator" >}}
$ cat @README.md@
{{< /text >}}
```

But the challenge here is that the path in github.com/istio/operator is different than the path in the archive. Should these files be located in github.com/istio/operator/samples/operators for consistency? Seems hacky, but would fix the link. Would be nice if there was a way to specify a different GitHub path, but I don't think we have that functionality right now.

geeknoid: Correct, we don't have the ability to point to some random location in the target repo.

frankbu: Merging sample directories from multiple repos into one samples directory in the release image seems like a bad idea. I suppose if we only do it in very specific cases, there's no risk of collision, but in the general case it does seem risky.

PR #5397, opened 2019-11-06
zh-translation:/docs/setup/platform-setup/alicloud/index.md

WisWang:
ref: https://github.com/servicemesher/istio-official-translation/issues/1291


WisWang:
Will submit another PR later.

PR #5433, opened 2019-11-06
Standalone operator

frankbu:


content/en/docs/ops/setup/standalone-operator/index.md
```diff
@@ -1,196 +1,134 @@
...
-* Applications where services with an Istio sidecar need to be able to
-  communicate with other non-Istio Kubernetes services
-
-Run the following command to switch to this profile:
+To install the Istio `demo` [configuration profile](/docs/setup/additional-setup/config-profiles/)
+using the operator, run the following command:
 
 {{< text bash >}}
-$ kubectl apply -f https://preliminary.istio.io/operator-profile-demo.yaml
+$ kubectl apply -f https://preliminary.istio.io/operator.yaml
```
ostromart: I worry that this install is prone to races - at least one tester experienced that. @sdake could we change this back to using kubectl -k? Otherwise we need to install the CRD separately.

content/en/docs/ops/setup/standalone-operator/index.md
```diff
@@ -1,196 +1,134 @@
...
+istio-pilot              ClusterIP      172.21.105.205   <none>          15010/TCP,15011/TCP,8080/TCP,15014/TCP                                                                                                       2m
+istio-policy             ClusterIP      172.21.14.236    <none>          9091/TCP,15004/TCP,15014/TCP                                                                                                                 2m
+istio-sidecar-injector   ClusterIP      172.21.155.47    <none>          443/TCP,15014/TCP                                                                                                                            2m
+istio-telemetry          ClusterIP      172.21.196.79    <none>          9091/TCP,15004/TCP,15014/TCP,42422/TCP                                                                                                       2m
+jaeger-agent             ClusterIP      None             <none>          5775/UDP,6831/UDP,6832/UDP                                                                                                                   2m
```
ostromart: Is this copied from somewhere or actual output from trying it recently? The reason I ask is we don't have jaeger agent at the moment in istioctl, so this install seems a bit different.

frankbu: @sdake added this, but I think he just copied it from the quick start instructions. If it's wrong, I think it needs to be updated in both places.

content/en/docs/ops/setup/standalone-operator/index.md
frankbu: IMO it would be better if operator.yaml just deployed the controller. Then we apply the CR in a second command. The single command install is not helpful in this guide, the intent of which is to show people how to use the operator.

ostromart: Not CR, the CRD. If that is not created first, the manifest apply can error out. 
If we separate the CR, it would be three apply commands.

ostromart: We still would have to separate the CRD, so it would be three apply commands.

content/en/docs/ops/setup/standalone-operator/index.md
ostromart: The charts are still somewhat in flux - we should have a definitive output tomorrow sometime. Let's hold off merging this till then and update the other docs at the same time.

content/en/docs/ops/setup/standalone-operator/index.md
frankbu: Won't we avoid the race if we just move the CR to a second apply? Aren't the CRDs only needed when you try to apply the CR, so it won't matter what order they're created in operator.yaml?

ostromart: Right, but the CR can still race with the CRD.

ostromart: Per our chat, it's ok to have everything in one base file and a separate CR. I've changed things around to support that in this PR: 
https://github.com/istio/istio.io/pull/5436
We should also update the blog to reflect these changes.

content/en/blog/2019/introducing-istio-operator/index.md
```diff
@@ -114,7 +114,8 @@ For example, to install the controller and a default version of Istio into your
 run the following command:
 
 {{< text bash >}}
-$ kubectl apply -f https://<repo URL>/<version>/operator-profile-default.yaml
+$ kubectl apply -f https://<repo URL>/operator.yaml
+$ kubectl apply -f https://<repo URL>/<version>/default-cr.yaml
```
frankbu: @ostromart I'm not sure I understand what exactly this <repo URL> and <version> in the path are supposed to be. What repo should this be? Why is the version a path component like this?

content/en/blog/2019/introducing-istio-operator/index.md
```diff
@@ -126,7 +127,7 @@ $ kubectl edit istiocontrolplane example-istiocontrolplane -n istio-system
 To upgrade to a new version of Istio, run:
 
 {{< text bash >}}
-$ kubectl apply -f https://<repo URL>/<new version>/operator-profile-default.yaml
+$ kubectl apply -f https://<repo URL>/<new version>/default-cr.yaml
```
frankbu: @ostromart Same confusion about the version in the path name here? What is different about upgrade vs just applying or editing the CR to change the config?

linsun:
My vote is for the operator page to go under here - https://preliminary.istio.io/docs/setup/install/ as experimental:

For 2 reasons:
1. this is our strategic direction where we want folks to use this.
2. we don't currently  have a path between istioctl to controller operator so user has to uninstall istio then move to controller operator.  it would be good to allow users to choose up front.

content/en/docs/ops/setup/standalone-operator/index.md
```diff
@@ -1,196 +1,141 @@
...
-installation method lets you quickly evaluate Istio in a Kubernetes cluster on
-any platform using a variety of profiles.
+This guide installs Istio using the standalone Istio
+[operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/).
+The only dependencies required are a supported Kubernetes cluster and the `kubectl` command.
 
-To install Istio for production use, we recommend using the [Helm Installation guide](/docs/setup/install/helm/)
+{{< warning >}}
+To install Istio for production use, we recommend [installing with {{< istioctl >}}](/docs/setup/install/istioctl/)
 instead, which is a stable feature.
```
sdake: may consider removing the language around "which is a stable feature".

content/en/docs/ops/setup/standalone-operator/index.md
```diff
@@ -1,196 +1,141 @@
...
-
-1. (Optionally) change profiles from the demo profile to one of the following profiles:
-
-{{< tabset cookie-name="profile" >}}
-
-{{< tab name="demo" cookie-value="permissive" >}}
-When using the [permissive mutual TLS mode](/docs/concepts/security/#permissive-mode), all services accept both plaintext and
-mutual TLS traffic. Clients send plaintext traffic unless configured for
-[mutual TLS migration](/docs/tasks/security/mtls-migration/). This profile is installed during the first step.
+    This command creates the operator custom resource definition and deploys the controller.
```
sdake: line 30 borders on too much information imo.  Seems like this could be simplified as either
`This command deploys the Istio operator and a sample profile`

or if you want the detail:

```
This command creates the following resources to deploy the operator:
- The `istio-operator` namespace
- The Istio operator custom resource definition
- The istio operator custom resource which defines how the operator should be deployed
- The `istio-operator-metrics` service to access operator metrics
- The `istio-operator` deployment to start and maintain a running operator
- Necessary Istio operator RBAC rules
```

content/en/docs/ops/setup/standalone-operator/index.md
```diff
@@ -1,196 +1,141 @@
...
-* Applications where services with an Istio sidecar need to be able to
-  communicate with other non-Istio Kubernetes services
-
-Run the following command to switch to this profile:
+To install the Istio `demo` [configuration profile](/docs/setup/additional-setup/config-profiles/)
+using the operator, run the following command:
 
 {{< text bash >}}
-$ kubectl apply -f https://preliminary.istio.io/operator-profile-demo.yaml
+$ kubectl apply -f - <<EOF
```
sdake: LGTM - this will need to be reconciled with the grab_operator_yamls.sh script as well as blog post.

content/en/docs/ops/setup/standalone-operator/index.md
sdake: ya - this was followup work that needed to be done.  :)

content/en/docs/ops/setup/standalone-operator/index.md
```diff
@@ -1,196 +1,141 @@
...
+zipkin                   ClusterIP      172.21.182.245   <none>          9411/TCP                                                                                                                                     2m
 {{< /text >}}
 
-{{< /tab >}}
-
-{{< tab name="default" cookie-value="default" >}}
-This profile enables Istio's default settings which contains recommended
-production settings. Run the following command to switch to this profile:
-
 {{< text bash >}}
```
sdake: Why two bash sections if only the demo profile is represented?

content/en/docs/ops/setup/standalone-operator/index.md
```diff
@@ -1,196 +1,141 @@
...
+Refer to the [`IstioControlPlane` API](/docs/reference/config/istio.operator.v1alpha12.pb/)
+for the complete set of configuration settings.
 
 ## Uninstall
 
-Delete the Istio Operator and Istio deployment:
+Delete the Istio operator and Istio deployment:
 
 {{< text bash >}}
 $ kubectl -n istio-operator get IstioControlPlane example-istiocontrolplane -o=json | jq '.metadata.finalizers = null' | kubectl delete -f -
```
sdake: These uninstall instructions need a change.  They represent the early implementaton of the oprator which had a broken removal requiring a force and removal of the finalizers metadata.

content/en/blog/2019/introducing-istio-operator/index.md
ostromart: The URL is TBD which is why I left it like that - we also haven't started creating subdirs for releases. This is just my best guess how this would work. I think it may be better to just leave the upgrade section out, since this is still somewhat in flux.

frankbu:
@linsun, @nrjpoddar I'm OK with moving this to /setup/install.

That said, let me give you my thought on future 1.5+ direction. I think that installing with istioctl should be the recommended approach, not just secondary for people that don't want to run controller. To be clear, I think istioctl should be made to work seamlessly with or without controller. If controller is not running, istioctl will apply the CR and also go ahead and create the Istio components for the install. If the controller is not running, it will just apply the CR and let controller do the install. Using istioctl, you can also start of stop the controller at any time.

istioctl also has all kinds of cool extra commands to list, view, and diff profiles and configs before applying. IMO, it should really be the preferred tool to interact with the CR, even when the controller is doing the work.

nrjpoddar:
> istioctl also has all kinds of cool extra commands to list, view, and diff profiles and configs before applying. IMO, it should really be the preferred tool to interact with the CR, even when the controller is doing the work.

@frankbu agree about using `istioctl` tool as primary CLI vehicle, for me imperative operations don't go over well in CI/CD pipelines and `istioctl` is all imperative. We can always change things based on what emerges as the primary way to install and maintain Istio.

content/en/docs/ops/setup/standalone-operator/index.md
sdake: @ostromart the CRD need not be separated.  I'd separate thee CRs, not the CRDs.

content/en/blog/2019/introducing-istio-operator/index.md
sdake: @ostromart the queestion i hae is what is `<repo URL>` - Is that a repository we make specifically for deploying the operator?  If so, the grab script could go there, and with a little maekfile magic, we could move this work from the documentation repo - which was probably the wrong place for it originally.

frankbu:
I've made the changed as discussed in comments. There are still some pending changes to update the repo URL and uninstall command, but I suggest we merge this as is, and fix those later in a followup PR.

PR #5434, opened 2019-11-06
task for protocol sniffing

yxue:
add task for demonstrating protocol sniffing works


content/en/docs/tasks/traffic-management/protocol-sniffing/index.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Protocol Sniffing
+description: Shows you how to define the new service without specifying port protocol.
+weight: 30
+keywords: [traffic-management,protocol-sniffing]
+---
+
+This task shows you how to define the new service without specifying the port protocol. For example, you
```
frankbu: I don't think we need a task to show people how to not do something. This isn't really a task, but rather a demonstration that some of Istio's features work, even if the port naming requirement is not followed.

Why don't we just remove the port names from the services used in all the other tasks and just let that demonstrate that it works?

geeknoid: @frankbu That's definitely an excellent idea, it would slightly simplify everything.

Issue #5307, opened 2019-10-31
Kiali page needs to be updated


jwendell:
https://preliminary.istio.io/docs/tasks/observability/kiali/

It needs to cover the new `istioctl manifest` method of installation.

When using the `demo` profile, kiali comes pre-installed. Otherwise, the user needs to install it after installing the control plane.

cc @jmazzitelli

jmazzitelli:
https://github.com/istio/installer/pull/503 will upgrade Kiali. I will work on the docs page once I confirm everything is ready and working with istioctl

PR #5440, opened 2019-11-07
[kiali] update task doc

jmazzitelli:
closes https://github.com/istio/istio.io/issues/5307


jmazzitelli:
This is ready to merge. However, there is a bug in istioctl installer such that these docs won't work IF the user wants to create their own secret. The problem is in the istioctl installer, not these docs, so we can merge this PR with the expectation that this bug will be fixed: https://github.com/istio/istio/issues/18763

content/en/docs/tasks/observability/kiali/index.md
```diff
@@ -18,16 +18,16 @@ This task uses the [Bookinfo](/docs/examples/bookinfo/) sample application as th
...
+To install Kiali without `istioctl`, follow the [Kiali installation instructions](https://www.kiali.io/documentation/getting-started/).
+{{< /tip >}}
 
 ### Create a secret
 
-{{< idea >}}
+{{< tip >}}
 If you plan on installing Kiali using the Istio demo profile as described in the [Istio Quick Start Installation Steps](/docs/setup/install/kubernetes/) then a default secret will be created for you with a username of `admin` and passphrase of `admin`. You can therefore skip this section.
-{{< /idea >}}
+{{< /tip >}}
```
jwendell: I wonder if it's better to instruct user to create the secret using something like this:
```
kubectl create secret generic -n istio-system kiali --from-literal "username=<USERNAME_HERE>" --from-literal "passphrase=<PASSWORD_HERE>"
```
That would save a lot of space in this page

content/en/docs/tasks/observability/kiali/index.md
```diff
@@ -69,31 +69,27 @@ data:
 EOF
 {{< /text >}}
 
-### Install Via Helm
+### Install Via `Istioctl`
 
 Once you create the Kiali secret, follow
```
jwendell: What about something like
-- `Once you create the Kiali secret`
++ `You are now ready to install Kiali.`

content/en/docs/tasks/observability/kiali/index.md
```diff
@@ -140,12 +136,10 @@ $ oc patch clusterrole kiali -p '[{"op":"add", "path":"/rules/-", "value":{"apiG
...
 
     {{< text bash >}}
-    $ kubectl -n istio-system port-forward $(kubectl -n istio-system get pod -l app=kiali -o jsonpath='{.items[0].metadata.name}') 20001:20001
+    $ istioctl dashboard kiali
     {{< /text >}}
 
-1.  Visit <https://localhost:20001/kiali/console> in your web browser.
-
-1.  To log into the Kiali UI, go to the Kiali login screen and enter the username and passphrase stored in the Kiali secret.
+1.  To log into the Kiali UI, enter the username and passphrase stored in the Kiali secret.
```
jwendell: append `(demo/demo if you are using the demo profile)` ?

content/en/docs/tasks/observability/kiali/index.md
jmazzitelli: I thought about that. I wondered why we originally wrote it this way - but I think the reason why is to not put the actual credential text in the user's command history. As it is now, the actual username and password is not found in any command line history. If we tell the user to use that one liner, they would have to put their passphrase in the command line and thus putting it in the command line history (which isn't secure). I am guessing that is the reason why this was originally written that way.

content/en/docs/tasks/observability/kiali/index.md
jmazzitelli: It's `admin/admin` but I get the idea - I'll add that.

jmazzitelli: ok, that is in..

```
1.  To log into the Kiali UI, enter the username and passphrase stored in the Kiali secret (`admin/admin` if you are using the demo profile).
```

content/en/docs/tasks/observability/kiali/index.md
jmazzitelli: I am hesitant to alter sentence structures/verbiage for parts in this document that are unrelated to the changes required, unless it is necessary. It took **a lot** of back-and-forth with the Istio doc writers when we originally wrote this page. I left everything as it was unless I needed to explicitly change something.

content/en/docs/tasks/observability/kiali/index.md
jwendell: ok, sounds good, we can improve this later in a follow up.

content/en/docs/tasks/observability/kiali/index.md
jwendell: ok, sounds good, we can improve this later in a follow up.

jmazzitelli:
This is ready to be merged. I see this is "Waiting on code owner review from istio/wg-docs-maintainers-english." - how do we get that done?

PR #5460, opened 2019-11-07
Fix issue in mtls-migration/index.md

johnzheng1975:
Update mtls-migration/index.md, add policy creation, whose Policy mode is PERMISSIVE.
The run result is fail, if this is missing

### DIFF
--- a/content/en/docs/tasks/security/mtls-migration/index.md
+++ b/content/en/docs/tasks/security/mtls-migration/index.md
@@ -75,7 +75,7 @@ down once the migration is done.
 
 ## Configure clients to send mutual TLS traffic
 
-Configure Istio services to send mutual TLS traffic by setting `Policy`, whose mode is PERMISSIVE.
+Configure Istio services to send mutual TLS traffic by setting `Policy`, whose mode is `PERMISSIVE`.
 
 {{< text bash >}}
 $ cat <<EOF | kubectl apply -n foo -f -
--- a/content/en/docs/tasks/security/mtls-migration/index.md
+++ b/content/en/docs/tasks/security/mtls-migration/index.md
@@ -75,6 +75,24 @@ down once the migration is done.
 
 ## Configure clients to send mutual TLS traffic
 
+Configure Istio services to send mutual TLS traffic by setting `Policy`, whose mode is PERMISSIVE.
+
+{{< text bash >}}
+$ cat <<EOF | kubectl apply -n foo -f -
+apiVersion: "authentication.istio.io/v1alpha1"
+kind: "Policy"
+metadata:
+  name: "example-httpbin-strict"
+  namespace: foo
+spec:
+  targets:
+  - name: httpbin
+  peers:
+  - mtls:
+      mode: PERMISSIVE
+EOF
+{{< /text >}}
+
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/security/mtls-migration/index.md
```diff
@@ -75,6 +75,24 @@ down once the migration is done.
 
 ## Configure clients to send mutual TLS traffic
 
+Configure Istio services to send mutual TLS traffic by setting `Policy`, whose mode is PERMISSIVE.
```
geeknoid:
```suggestion
Configure Istio services to send mutual TLS traffic by setting `Policy`, whose mode is `PERMISSIVE`.
```

johnzheng1975:
@googlebot I signed it!

johnzheng1975:
In "before-you-begin", the document asked for "without global mutual TLS enabled". Mostly user will think "there should be NO meshpolicy/ policy pre-created for this demo".  If so, later run will fail definitely, until with my document fix.

With current installation, a default meshpolicy will be created with permessive mode.  Even based on this, my document still works.  

And, my document fix is useful for enhance user understanding with permessive mode.

That's why I think my document fix should be merge.

johnzheng1975:
Maybe there is another way, to make this document easier to understand.
Run "kubectl get meshpolicy -o yaml", to view default meshpolicy, which mode is "PERMISSIVE"
```
  spec:
    peers:
    - mtls:
        mode: PERMISSIVE
```
If you think this is better, I can cancel my current PR.

yangminzhu:
@incfly @diemtvu Could you take a look? I'm not familiar with this task. Is the task failing for 1.4 release?

geeknoid:
@istio/wg-security-maintainers PTAL

content/en/docs/tasks/security/mtls-migration/index.md
```diff
@@ -75,6 +75,24 @@ down once the migration is done.
...
+kind: "Policy"
+metadata:
+  name: "example-httpbin-strict"
+  namespace: foo
+spec:
+  targets:
+  - name: httpbin
+  peers:
+  - mtls:
+      mode: PERMISSIVE
```
incfly: you make the policy name as strict but the mode is permissive?

content/en/docs/tasks/security/mtls-migration/index.md
```diff
@@ -75,6 +75,24 @@ down once the migration is done.
 
 ## Configure clients to send mutual TLS traffic
 
+Configure Istio services to send mutual TLS traffic by setting `Policy`, whose mode is `PERMISSIVE`.
```
incfly: Are this needed? We ship with default installation including PERMISSIVE?

Can you describe your installation command line, along with the version you used?

johnzheng1975: You are right, this is not necessary since ship with installation including PERMISSIVE.
However, it will make user confuse, since the document did not mention this.
So, as I said, I can cancel this pull request, had better mentioned this in document in another new pull request.

johnzheng1975: Just FYI.
I will cancel this if you think the document is clear enough.

PR #5463, opened 2019-11-07
zh-translation:/docs/reference/glossary/multi-mesh.md

niceforbear:
ref: https://github.com/servicemesher/istio-official-translation/issues/1265


geeknoid:
This is inserting Chinese content in the English section, I think you edited the wrong file :-)

PR #5473, opened 2019-11-07
Update mongo egress blog

vadimeisenbergibm:
Fix https://github.com/istio/istio/issues/17163.

The helm template update of the egress gateway must include deployment in addition to the service, since it seems 
the deployment has to have container port specified so the endpoints of the service will be taken into account.


vadimeisenbergibm:
@geeknoid 

> Please add a last_update field to the front-matter.

The last_update field is already there. Is it updated automatically, or should I update it?

geeknoid:
@vadimeisenbergibm You need to update the last_update field by hand. It represents when there were major content changes in the page, as opposed to the "Last Edited" time which is computed automatically and gets updated whenever anything changes in the page (such as fixing links, etc)

vadimeisenbergibm:
@geeknoid @frankbu I have updated the last_update field by https://github.com/istio/istio.io/pull/5473/commits/6d303f6811cd6b07c71b81bae6450175aa0ebfb8.

PR #5474, opened 2019-11-07
Organize security tasks

frankbu:


content/en/docs/tasks/security/authorization/rbac-groups/index.md
```diff
@@ -3,6 +3,8 @@ title: Authorization for groups and list claims
 description: Tutorial on how to configure the groups-base authorization and configure the authorization of list-typed claims in Istio.
 weight: 4
```
yangminzhu: Could you change the weight to 30 so that this page is after the HTTP and TCP task?

rcaballeromx:
Could we use the following categories instead?

- Authentication
- Authorization
- Encryption
- Identity

frankbu:
@rcaballeromx Not really, because we have no Encryption or Identity tasks, so there would be nothing in them.

rcaballeromx:
> @rcaballeromx Not really, because we have no Encryption or Identity tasks, so there would be nothing in them.

@frankbu The content you have around configuring Citadel is to get those things in place. Or am I missing something?

frankbu:
@rcaballeromx The tasks are a mixed bag, all around configuring Citatel. I woudn't really think of them as encryption or identify tasks. I wouldn't really know what to expect in an encryption task section. A task that shows how to change encryption algorithms? I don't think that's even supported.

frankbu:
After 1.4, we can consider further refining the categories, but we'll need to rework some of the tasks at the same time.

I think we should merge this as is now, the sooner the better, because it will have merge conflicts with lots of other affected docs.

rcaballeromx:
@frankbu I think you are right, post release we will revisit. Encryption wouldn't be how to change encryption algorithms but rather how to get traffic in the mesh encrypted. 
Let's revisit for 1.5 then.

PR #5475, opened 2019-11-07
Remove jaeger-agent from the Quick Start.

ericvn:
This PR removes the jaeger-agent line and note from the quick start since the _current_ istioctl no longer creates the service. https://github.com/istio/istio/issues/18658is open whether this service should still be created or not. For now, this matches what is deployed. If there is a change to add the service in the other PR, this can be reverted.


ericvn:
Looks like we are adding `jaeger-agent` back in so closing this PR.

PR #5478, opened 2019-11-07
Revert "Split up install to separate CRs (#5436)"

sdake:
Please see the script in scripts/grab_operator_yamls.sh.
    
This reverts commit 81da298400ddef3c1c1f533c5ecb164f7f6622ee.

### DIFF
--- a/static/operator-profile-minimal.yaml
+++ b/static/operator-profile-minimal.yaml
@@ -5,6 +5,5 @@ metadata:
   namespace: istio-operator
   name: example-istiocontrolplane
 spec:
-  profile: sds
+  profile: demo
 ...
----
--- a/static/operator-profile-sds.yaml
+++ b/static/operator-profile-sds.yaml
@@ -5,6 +5,5 @@ metadata:
   namespace: istio-operator
   name: example-istiocontrolplane
 spec:
-  profile: default
+  profile: demo
 ...
----
--- a/static/operator.yaml
+++ b/static/operator.yaml
@@ -164,7 +164,7 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - configmaps
+  - configmaps  
   - endpoints
   - events
   - namespaces
@@ -172,7 +172,7 @@ rules:
   - persistentvolumeclaims
   - secrets
   - services
-  - serviceaccounts
+  - serviceaccounts  
   verbs:
   - '*'
 ...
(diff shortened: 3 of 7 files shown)


static/operator-profile-minimal.yaml
```diff
@@ -5,6 +5,5 @@ metadata:
   namespace: istio-operator
   name: example-istiocontrolplane
 spec:
-  profile: default
```
ostromart: Is this correct? shouldn't it be minimal?

static/operator-profile-sds.yaml
```diff
@@ -5,6 +5,5 @@ metadata:
   namespace: istio-operator
   name: example-istiocontrolplane
 spec:
-  profile: minimal
```
ostromart: This should be sds?

static/operator.yaml
```diff
@@ -252,3 +252,11 @@ spec:
               value: "istio-operator"
 ...
 ---
+apiVersion: install.istio.io/v1alpha2
```
ostromart: This should still be removed as it leads to races. The docs should split up operator.yaml and the profile CRs.

sdake: Some attention is needed here.  The first step is to revert the problems introduced by manual modification of these files.  There is automated tooling to do this work.

static/operator-profile-sds.yaml
sdake: the automated tooling does some seding here, but it appears I didn't revert enough PRs - will address.

static/operator-profile-minimal.yaml
sdake: the automated tooling does some seding here, but it appears I didn't revert enough PRs - will address.

sdake:
@ostromart if you look in detail at the change, this reverts the name changes - which reverts some of the content.  Please take a look at the generator script if you wish to follow up after this PR merges with corrected automation.

static/operator-profile-sds.yaml
sdake: this should not be sds, the name of the file is minimal?

ostromart: the old file name is minimal but it's being changed back to operator-profile-sds.yaml.

static/operator.yaml
ostromart: Ok, I'll put this in and redo through the script.

PR #5479, opened 2019-11-07
update doc

JimmyCYJ:
The root transition script is updated in https://github.com/istio/tools/pull/517.
This PR updates user guide.

### DIFF
--- a/content/en/docs/ops/security/root-transition/index.md
+++ b/content/en/docs/ops/security/root-transition/index.md
@@ -48,12 +48,32 @@ please follow the procedure and check whether you will be affected.
     {{< text bash>}}
     $ wget https://raw.githubusercontent.com/istio/tools/{{< source_branch_name >}}/bin/root-transition.sh
     $ chmod +x root-transition.sh
-    $ ./root-transition.sh check
+    $ ./root-transition.sh check-root
     ...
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/security/root-transition/index.md
```diff
@@ -48,13 +48,33 @@ please follow the procedure and check whether you will be affected.
...
     $ chmod +x root-transition.sh
-    $ ./root-transition.sh check
+    $ ./root-transition.sh check-root
     ...
     ===YOU HAVE 30 DAYS BEFORE THE ROOT CERT EXPIRES!=====
     {{< /text >}}
 
     Execute the remainder of the steps prior to root certificate expiration to avoid system outages.
 
+1. Upgrade to Istio 1.0.8, 1.1.8 or later:
```
geeknoid: I don't think we want/need these changes. The doc set is for Istio 1.4, we don't need to have specific items for Iatio 1.0 and 1.1 (neither of which we support anymore)

JimmyCYJ: There are istio-on-GKE customers which use other release versions. @myidpt

geeknoid: Then that info belongs on some GKE page somewhere, not in the mainline Istio 1.4 doc set.

Thanks.

PR #5482, opened 2019-11-07
Fix auth installation and its references.

myidpt:
In previous PRs, we removed "demo-auth" from the installation guide, but we didn't add general auth enablement instructions, so users won't be able to learn how to enable auth from the installation page. Also the security tasks are referring to the installation page which is missing the security enablement instructions. This PR fixes these issues.

### DIFF
--- a/content/en/docs/setup/install/helm/index.md
+++ b/content/en/docs/setup/install/helm/index.md
@@ -130,7 +130,7 @@ $ helm template install/kubernetes/helm/istio --name istio --namespace istio-sys
 
 {{< /tab >}}
 
-{{< tab name="mTLS enabled" cookie-value="mtls" >}}
+{{< tab name="Mutual TLS" cookie-value="mtls" >}}
 
 {{< text bash >}}
 $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
@@ -141,7 +141,7 @@ $ helm template install/kubernetes/helm/istio --name istio --namespace istio-sys
 
 {{< /tab >}}
 
-{{< tab name="mTLS (SDS) enabled" cookie-value="sds" >}}
+{{< tab name="Mutual TLS with SDS" cookie-value="sds" >}}
 
 {{< text bash >}}
 $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
@@ -236,7 +236,7 @@ $ helm install install/kubernetes/helm/istio --name istio --namespace istio-syst
 
 {{< /tab >}}
 
-{{< tab name="mTLS enabled" cookie-value="mtls" >}}
+{{< tab name="Mutual TLS" cookie-value="mtls" >}}
 
 {{< text bash >}}
 $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
@@ -247,7 +247,7 @@ $ helm template install/kubernetes/helm/istio --name istio --namespace istio-sys
 
 {{< /tab >}}
 
-{{< tab name="mTLS (SDS) enabled" cookie-value="sds" >}}
+{{< tab name="Mutual TLS with SDS" cookie-value="sds" >}}
 
 {{< text bash >}}
 $ helm install install/kubernetes/helm/istio --name istio --namespace istio-system \
@@ -326,7 +326,7 @@ $ kubectl delete namespace istio-system
 
 {{< /tab >}}
 
-{{< tab name="mTLS enabled" cookie-value="mtls" >}}
+{{< tab name="Mutual TLS" cookie-value="mtls" >}}
 
 {{< text bash >}}
 $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
@@ -337,7 +337,7 @@ $ helm template install/kubernetes/helm/istio --name istio --namespace istio-sys
 
 {{< /tab >}}
 
-{{< tab name="mTLS (SDS) enabled" cookie-value="sds" >}}
+{{< tab name="Mutual TLS with SDS" cookie-value="sds" >}}
 
 {{< text bash >}}
 $ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
--- a/content/en/docs/setup/install/helm/index.md
+++ b/content/en/docs/setup/install/helm/index.md
@@ -130,7 +130,18 @@ $ helm template install/kubernetes/helm/istio --name istio --namespace istio-sys
 
 {{< /tab >}}
 
-{{< tab name="sds" cookie-value="sds" >}}
+{{< tab name="mTLS enabled" cookie-value="mtls" >}}
+
+{{< text bash >}}
+$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
+    --values install/kubernetes/helm/istio/values-istio-demo.yaml \
+    --set global.controlPlaneSecurityEnabled=true \
+    --set global.mtls.enabled=true | kubectl apply -f -
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="mTLS (SDS) enabled" cookie-value="sds" >}}
 
 {{< text bash >}}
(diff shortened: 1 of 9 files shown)


content/en/docs/tasks/security/citadel-config/auth-sds/index.md
```diff
@@ -49,8 +49,7 @@ This approach has the following benefits:
 
 ## Before you begin
 
-* Follow the [install instructions](/docs/setup/install/istioctl/)
-  to set up Istio with SDS and global mutual TLS enabled.
+Follow the [Istio installation guide](/docs/setup/install/helm/) to set up Istio with SDS and global mutual TLS enabled.
```
frankbu: Can't we do this with istioctl?

myidpt: Yes, I need to test it. I will update to istioctl once I verified it's working.

content/en/docs/setup/install/helm/index.md
```diff
@@ -130,7 +130,18 @@ $ helm template install/kubernetes/helm/istio --name istio --namespace istio-sys
 
 {{< /tab >}}
 
-{{< tab name="sds" cookie-value="sds" >}}
+{{< tab name="mTLS enabled" cookie-value="mtls" >}}
```
geeknoid: <
```suggestion
{{< tab name="Mutual TLS" cookie-value="mtls" >}}
```

content/en/docs/setup/install/helm/index.md
```diff
@@ -130,7 +130,18 @@ $ helm template install/kubernetes/helm/istio --name istio --namespace istio-sys
...
+{{< text bash >}}
+$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
+    --values install/kubernetes/helm/istio/values-istio-demo.yaml \
+    --set global.controlPlaneSecurityEnabled=true \
+    --set global.mtls.enabled=true | kubectl apply -f -
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="mTLS (SDS) enabled" cookie-value="sds" >}}
```
geeknoid:
```suggestion
{{< tab name="Mutual TLS with SDS" cookie-value="sds" >}}
```

content/en/docs/setup/install/helm/index.md
```diff
@@ -225,7 +236,18 @@ $ helm install install/kubernetes/helm/istio --name istio --namespace istio-syst
 
 {{< /tab >}}
 
-{{< tab name="sds" cookie-value="sds" >}}
+{{< tab name="mTLS enabled" cookie-value="mtls" >}}
```
geeknoid:
```suggestion
{{< tab name="Mutual TLS" cookie-value="mtls" >}}
```

content/en/docs/setup/install/helm/index.md
```diff
@@ -225,7 +236,18 @@ $ helm install install/kubernetes/helm/istio --name istio --namespace istio-syst
...
+{{< text bash >}}
+$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
+    --values install/kubernetes/helm/istio/values-istio-demo.yaml \
+    --set global.controlPlaneSecurityEnabled=true \
+    --set global.mtls.enabled=true | kubectl apply -f -
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="mTLS (SDS) enabled" cookie-value="sds" >}}
```
geeknoid:
```suggestion
{{< tab name="Mutual TLS with SDS" cookie-value="sds" >}}
```

content/en/docs/setup/install/helm/index.md
frankbu:
```suggestion
{{< tab name="demo mTLS" cookie-value="mtls" >}}
```

content/en/docs/setup/install/helm/index.md
```diff
@@ -304,7 +326,18 @@ $ kubectl delete namespace istio-system
 
 {{< /tab >}}
 
-{{< tab name="sds" cookie-value="sds" >}}
+{{< tab name="mTLS enabled" cookie-value="mtls" >}}
```
geeknoid:
```suggestion
{{< tab name="Mutual TLS" cookie-value="mtls" >}}
```

content/en/docs/setup/install/helm/index.md
```diff
@@ -304,7 +326,18 @@ $ kubectl delete namespace istio-system
...
+{{< text bash >}}
+$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system \
+    --values install/kubernetes/helm/istio/values-istio-demo.yaml \
+    --set global.controlPlaneSecurityEnabled=true \
+    --set global.mtls.enabled=true | kubectl apply -f -
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="mTLS (SDS) enabled" cookie-value="sds" >}}
```
geeknoid:
```suggestion
{{< tab name="Mutual TLS with SDS" cookie-value="sds" >}}
```

content/en/docs/setup/install/helm/index.md
frankbu:
```suggestion
{{< tab name="sds mTLS" cookie-value="sds" >}}
```

content/en/docs/setup/install/helm/index.md
frankbu:
```suggestion
{{< tab name="demo mTLS" cookie-value="mtls" >}}
```

content/en/docs/setup/install/helm/index.md
frankbu:
```suggestion
{{< tab name="sds mTLS" cookie-value="sds" >}}
```

content/en/docs/setup/install/helm/index.md
frankbu:
```suggestion
{{< tab name="demo mTLS" cookie-value="mtls" >}}
```

content/en/docs/setup/install/helm/index.md
frankbu:
```suggestion
{{< tab name="sds mTLS" cookie-value="sds" >}}
```

frankbu:
@geeknoid, I think it's OK to use `mTLS`, instead of `mutual TLS` here, because they're tab labels that we want to keep short. They don't cause spelling errors in this context either.

content/en/docs/setup/install/helm/index.md
myidpt: Frank, let's not call it "demo" since we have decided to not have two demos :)

content/en/docs/tasks/security/citadel-config/auth-sds/index.md
myidpt: It's still not working. We will update this when we have fixed it.

myidpt:
Talked with @geeknoid, mTLS is not straight-forward for average readers. I agree we should use "Mutual TLS" here.

content/en/docs/setup/install/helm/index.md
frankbu: The instruction says to choose a configuration profile (https://deploy-preview-5482--preliminary-istio.netlify.com/docs/setup/additional-setup/config-profiles/), so you can't just say "Mutual TLS" on the label, since there's no such profile.

Currently your tab is installing the `demo` profile with mTLS enabled. If you don't want to label it as such, then I suggest changing the contents of that tab to say that you can turn on mTLS for any configuration profile and explain what options the user needs to set and then give the instruction as just an example. Take a look at the CNI tab, it is written that way.

I would also change the SDS tab back to just `sds`, like it was before, because the configuration profiles already says that the sds profile turns on mutual TLS.

I would also suggest putting the Mutual TLS tab second last, after `sds` but before `CNI`, so the first 4 are straight profile selections, while the last 2 are how to turn on a special feature with any profile.

PR #5489, opened 2019-11-08
Improve authn-policy/index.md

johnzheng1975:
Had better add "Test it again after you add destination rule", which make user easy to understand.

### DIFF
--- a/content/en/docs/tasks/security/authn-policy/index.md
+++ b/content/en/docs/tasks/security/authn-policy/index.md
@@ -212,7 +212,7 @@ spec:
 EOF
 {{< /text >}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/security/authn-policy/index.md
```diff
@@ -212,6 +212,14 @@ spec:
 EOF
 {{< /text >}}
 
+Test it again after you add destination rule, it is pass.
```
geeknoid:
```suggestion
Test it again after you add the destination rule to ensure it passes:
```

geeknoid:
/cherrypick/master

PR #5508, opened 2019-11-08
zh-translation:/faq/general/how-was-istio-started.md

5idu:
ref: https://github.com/servicemesher/istio-official-translation/issues/1402


5idu:
@googlebot I signed it!

PR #5523, opened 2019-11-08
Update the egrep and service accounts.

incfly:
Changed the service account output and the egrep format (not sure why the old one does not work)

Test against latest build of 1.4 release, `istio-1.4-alpha.5385e895f4e045981a4369826389ea0fdbb2b9e6`.

### DIFF
--- a/content/en/docs/tasks/security/authentication/auto-mtls/index.md
+++ b/content/en/docs/tasks/security/authentication/auto-mtls/index.md
@@ -131,7 +131,7 @@ URI=spiffe://cluster.local/ns/full/sa/sleep
 response 200
 {{< /text >}}
 
-"URI=spiffe://cluster.local/ns/full/sa/sleep" is the underlying X509 client certificate, which
+The SPIFFE URI shows the client identity from X509 certificate, which
 indicates the traffic is sent in mutual TLS. If the traffic is in plain text, no client certificate
 will be displayed.
 
--- a/content/en/docs/tasks/security/authentication/auto-mtls/index.md
+++ b/content/en/docs/tasks/security/authentication/auto-mtls/index.md
@@ -1,7 +1,7 @@
 ---
 title: Automatic mutual TLS
 description: A simplified workflow to adopt mutual TLS with minimal configuration overhead.
-weight: 50
+weight: 10
 keywords: [security,mtls,ux]
 ---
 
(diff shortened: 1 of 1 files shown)


incfly:
This is after I tried to install with latest 1.4 daily build and istioctl manifest installation. Seems like the grep regex format needs to be updated and the service account we use for the sleep changed as well.

content/en/docs/tasks/security/authentication/auto-mtls/index.md
```diff
@@ -126,13 +126,14 @@ You can verify setup by sending an HTTP request with `curl` from any `sleep` pod
 For example, here is a command to check `sleep.full` to `httpbin.full` reachability:
 
 {{< text bash >}}
-$ kubectl exec $(kubectl get pod -l app=sleep -n full -o jsonpath={.items..metadata.name}) -c sleep -n full -- curl http://httpbin.full:8000/headers  -s  -w "response %{http_code}\n" | egrep 'Client-Cert.*sa.default|response.*$' -o
-Client-Cert": "By=spiffe://cluster.local/ns/full/sa/default
+$ kubectl exec $(kubectl get pod -l app=sleep -n full -o jsonpath={.items..metadata.name}) -c sleep -n full -- curl http://httpbin.full:8000/headers  -s  -w "response %{http_code}\n" | egrep -o 'URI\=spiffe.*sa/[a-z]*|response.*$'
```
diemtvu: Or may be just grep "spiffe" to avoid other change in the future.

incfly: The actual header is 
```
    "X-Forwarded-Client-Cert": "By=spiffe://cluster.local/ns/full/sa/httpbin;Hash=22c9623e6aac2e31f8720678d0816c8f43be3534b3c0094a0f6c21ce3ec90020;Subject=\"\";URI=spiffe://cluster.local/ns/full/sa/sleep"
```

Both client and server certs, kinda of verbose and confusing.

diemtvu: Hmm, why it show the server SA. May be it's worth to check, outside the scope of this PR

incfly: The second part is the client SA.

PR #5617, opened 2019-11-12
update ingress-certmgr index.md

WisWang:
I think the words I delete has no strong relation with this article and the words is really hard to understand.


geeknoid:
Slight lint error, remove the extra space at the end of line 10.

WisWang:
> Slight lint error, remove the extra space at the end of line 10.
Sorry for the Slight lint error

PR #5647, opened 2019-11-13
Fix root transition file download path.

myidpt:
The root_transition.sh file doesn't change between different versions. So it doesn't make sense to use branch-specific files. Actually this is causing improvements to the file need to be back ported to previous versions, which increases our workload.

### DIFF
--- a/content/en/docs/ops/security/root-transition/index.md
+++ b/content/en/docs/ops/security/root-transition/index.md
@@ -42,11 +42,11 @@ please follow the procedure and check whether you will be affected.
 
 1. Check when the root certificate expires:
 
-    Download this [script](https://raw.githubusercontent.com/istio/tools/master/bin/root-transition.sh)
+    Download this [script](https://raw.githubusercontent.com/istio/tools/root-transition/bin/root-transition.sh)
     on a machine that has `kubectl` access to the cluster.
 
     {{< text bash>}}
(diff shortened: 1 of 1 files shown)


geeknoid:
I don't believe this is correct. We should never refer to stuff directly in master as that can change in incompatible ways, move, be temporarily broken, etc.

myidpt:
@geeknoid istio/tools doesn't have all the release branches other repos have. For example, it only has release-1.4 branch. Does it make sense to refer to a dedicated branch "root-transition"?

myidpt:
Ping @geeknoid Please see if referring to the "root-transition" branch makes sense.
@mandarjog Is the istio/tool repo going to have a branch for each release? If so, we can always point to the release-branch. My concern is that if the istio/tool repo doesn't always maintain these branches, this link may break in the future.

geeknoid:
This is the 1.4 documentation set, it doesn't have any impact on 1.3. Customers looking at this set of documentation will get a link that works for them, which is what counts. If we make changes to the script in master later on, it won't affect customers using istio.io that are using 1.4.

This is how all of our references to GitHub stuff works, there is nothing special about this case.

geeknoid:
And using a root-transition branch doesn't help the problem of future updates. If we end up making some incompatible change to that script, or temporarily breaking the script in between relesses, users of the web site will have problems.

Basically, it is never OK for istio.io to point to content on a development branch, it must point to a stable release-specific branch.

PR #5703, opened 2019-11-15
update hub and tag for operator

linsun:


linsun:
Agreed, currently it is too manual to get the istio.io content updated.

PR #5719, opened 2019-11-15
Site improvements.

geeknoid:
- Display the version number in the site header. The version number was only displayed
in tiny font in the footer, now it's quite a bit more obvious.

- Fix layout of the footer to avoid the icons wrapping on two lines when the browser was
specific widths.

- Clicking on a folder in the left navigation area no longer expands the
folder. It only brings up the page associated with the folder in the main
page area. This was always my intent, I never got around to it til now.


geeknoid:
/cherrypick/release-1.4

PR #5729, opened 2019-11-15
Add resolution and examples for analzyer.

nrjpoddar:


nrjpoddar:
@sushicw @davidebbo I'm proposing that we use this format or something similar to this for documenting all analyzers. Without examples and resolution steps, it might be difficult for users to consume analysis output.

content/en/docs/reference/config/analysis/ist0109/index.md
```diff
@@ -3,10 +3,90 @@ title: ConflictingMeshGatewayVirtualServiceHosts
...
+
+To resolve this issue, you can take one of the following actions:
+
+* Merge the conflicting virtual services into a single resource.
+* Make the hostnames unique across virtual services attached to mesh gateway.
+* Scope the resource to a specific namespace by setting the `exportTo` field.
+
+## Examples
+
+The following `VirtualServices`, `productpage` in namespaces `team1` and `team2`
```
sushicw: I found this sentence confusing to read at first. Here's a stab at an alternative:
```
The `productpage` virtual service in namespace `team1` conflicts with the `productpage` 
virtual service in namespace `team2` because both of the following are true:
  * they are attached to the default mesh gateway (no custom gateway is specified)
  * They have the same host `productpage.default.svc.cluster.local` defined.
```

It may also be wise to have a different name for each virtual service, so it's clearer that the conflict isn't that they share a name (which is fine) but that they share a host.

content/en/docs/reference/config/analysis/ist0109/index.md
```diff
@@ -3,10 +3,90 @@ title: ConflictingMeshGatewayVirtualServiceHosts
...
+  - productpage.default.svc.cluster.local
+  http:
+  - route:
+    - destination:
+        host: productpage.team-2.svc.cluster.local
+---
+{{< /text >}}
+
+Note that setting the `exportTo` field to `.` restricts the scope to the
+namespace where the virtual service resource exists.
```
sushicw: I'd move this up to where you explain the resolution. e.g. 

```
You can resolve this issue by setting the `exportTo` field to `.` so that each virtual service 
is scoped only to its own namespace. 
```

content/en/docs/reference/config/analysis/ist0109/index.md
```diff
@@ -3,10 +3,90 @@ title: ConflictingMeshGatewayVirtualServiceHosts
...
+
+{{< text yaml >}}
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: productpage
+  namespace: team-1
+spec:
+  exportTo:
+  - "."
```
sushicw: @geeknoid Is there any way to have YAML text in our docs, with highlighted lines? 

If not, do you think it's a feasible FR? This is an example of where such a thing could be really useful.

nrjpoddar: I was thinking about that, I know some Hugo themes support this, not sure if the one in Istio supports it.

content/en/docs/reference/config/analysis/ist0109/index.md
```diff
@@ -3,10 +3,90 @@ title: ConflictingMeshGatewayVirtualServiceHosts
...
 ---
 
-This message occurs when Istio detects an overlap between virtual service
-resources. For example, you have multiple virtual services defined to use the
-same hostname and gateway, where the gateway is set to `mesh` or no gateway
-specified.
+This message occurs when Istio detects an overlap between
+[virtual service](/docs/reference/config/networking/virtual-service)
+resources that conflict with one another. For example, multiple virtual
+services defined to use the same hostname and attached to mesh gateway
```
zjory:
```suggestion
services are defined to use the same hostname and attached to mesh gateway
```

content/en/docs/reference/config/analysis/ist0109/index.md
```diff
@@ -3,10 +3,90 @@ title: ConflictingMeshGatewayVirtualServiceHosts
...
+will generate this message. Note that Istio supports merging of virtual
```
zjory:
```suggestion
will generate an error message. Note that Istio supports merging of virtual
```

content/en/docs/reference/config/analysis/ist0109/index.md
```diff
@@ -3,10 +3,93 @@ title: ConflictingMeshGatewayVirtualServiceHosts
...
+  hosts:
+  - productpage.default.svc.cluster.local
+  http:
+  - route:
+    - destination:
+        host: productpage.team-2.svc.cluster.local
+---
+{{< /text >}}
+
+You can resolve this issue by setting the `exportTo` field is set to `.` so
```
geeknoid:
```suggestion
You can resolve this issue by setting the `exportTo` field to `.` so
```

content/en/docs/reference/config/analysis/ist0109/index.md
geeknoid: No, we don't have support to highlight individual lines in a text block.  It could probably be done with some annotation, similar to the @@ syntax we support for links. Perhaps %% can indicate a span of text that should be rendered distinctively.

Anyway, not at the moment...

content/en/docs/reference/config/analysis/ist0109/index.md
```diff
@@ -3,10 +3,93 @@ title: ConflictingMeshGatewayVirtualServiceHosts
...
+services defined to use the same hostname and attached to mesh gateway
+will generate an error message. Note that Istio supports merging of virtual
+services that are attached to the ingress gateways.
+
+## Resolution
+
+To resolve this issue, you can take one of the following actions:
+
+* Merge the conflicting virtual services into a single resource.
+* Make the hostnames unique across virtual services attached to mesh gateway.
```
geeknoid:
```suggestion
* Make the hostnames unique across virtual services attached to a mesh gateway.
```

content/en/docs/reference/config/analysis/ist0109/index.md
```diff
@@ -3,10 +3,93 @@ title: ConflictingMeshGatewayVirtualServiceHosts
 layout: analysis-message
 ---
 
-This message occurs when Istio detects an overlap between virtual service
-resources. For example, you have multiple virtual services defined to use the
-same hostname and gateway, where the gateway is set to `mesh` or no gateway
+This message occurs when Istio detects an overlap between
+[virtual service](/docs/reference/config/networking/virtual-service)
+resources that conflict with one another. For example, multiple virtual
+services defined to use the same hostname and attached to mesh gateway
```
geeknoid:
```suggestion
services defined to use the same hostname and attached to a mesh gateway
```

nrjpoddar:
@geeknoid all comments addresses. Thanks for the review!

PR #5751, opened 2019-11-16
Prevent translated content from referencing English content.

geeknoid:
- In order to allow each language's content to evolve independently,
introduce a new linter rule to prevent translated content from having
references to English content.

- Fix many such references in the existing translated content.

### DIFF
--- a/content/zh/docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md
+++ b/content/zh/docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md
@@ -1,41 +1,31 @@
 ---
-title: Egress using Wildcard Hosts
-description: Describes how to enable egress traffic for a set of hosts in a common domain, instead of configuring each and every host separately.
+title: 使用通配符主机配置 Egress 流量
+description: 介绍如何为公共域中的一组主机启用 Egress 流量，而不是单独配置每个主机。
 keywords: [traffic-management,egress]
 weight: 50
 aliases:
   - /zh/docs/examples/advanced-gateways/wildcard-egress-hosts/
 ---
 
-The [Control Egress Traffic](/docs/tasks/traffic-management/egress/) task and
-the [Configure an Egress Gateway](/docs/tasks/traffic-management/egress/egress-gateway/) example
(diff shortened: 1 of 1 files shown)


geeknoid:
@rootsongjc We discussed challenges English doc writers have to deal with at this week's TOC meeting around broken links in translation. It became clear that we need to keep the individual languages isolated from one another to avoid problems. This PR puts in a linter rule to prevent cross-linking, and fixes a bunch of links that were pointing from the Chinese content to the English content.

Unfortunately, this leads to a number of broken links in the Chinese content. That is, the links used to point to content/en/XXX, but there is no matching content/zh/XXX and so the link checker fails.

Could you please take over this PR and fix the missing links? I think it's mostly about:

- Copying a file from the english content into the Chinese content, to be translated later.

- Adding {#XXX} annotations to headers such that links with fragment references (foo/bar#XXX) work.

As this PR touches a lot of files, it'd be best to push it in ASAP to avoid constant merge conflicts.

Thanks.

rootsongjc:
@geeknoid We have noticed this problem when submitting Chinese documents, and we have ensured that all checks are passed without the problem of broken links.

geeknoid:
@rootsongjc The issue is that Chinese content currently links to English content. As English content evolves, English authors end up having to modify the Chinese content in order for the cross-links to not be broken. This means that English changes are blocked on Chinese contributors.

What this PR tries to do is to separate the two worlds completely. Chinese content only references Chinese content, and English content only references English content. The two can thus evolve independently without interfering with one another.

rootsongjc:
@istio/wg-docs-maintainers-chinese anyone who can deal with this issue?

tanjunchen:
@rootsongjc we can deal with this pr first. Tell the translator, pay attention to the translation rules.
@geeknoid  deal with conflicts.
Next,we will open an issue to deal with this problems. Thanks

tanjunchen:
@geeknoid: PR needs rebase.

geeknoid:
@tanjunchen I've updated the PR to adapt to latest changes.

To avoid this getting so out of date again, can you copy this PR and address the linter issues so that it can be merged?

Thanks.

tanjunchen:
@geeknoid sure. I will fix the same question after  your pr is merged

PR #5768, opened 2019-11-18
Add support for dots & pills for both news and blog posts.

geeknoid:
- If a returning user comes to the site, if there are unread
blog posts or news articles less than 15 quadrllion nanosecond
old will be treated as being unread. When there are unred articles,
the News or Blog link in the title bar will get a green dot indicating
articles are available. When clicking on News, then you'll get the
news categories with a pill showing how many articles are unread for
each category.

First-time visitors to the site will not get any dots or pills for
existing articles. These will only appear in subsequent visits for
new articles.

Due to the default behavior for new users, if you just look at the
preview, you will not see any pills or dots. To see what this actually
looks like, load up the preview, then go to the Chrome Developer Tools,
click on the Application tab, then on Local Storage, and then find the
visitedPages entry. Right click on the entry, select Edit Value,
and set the value to {}. Then refresh the page and you
should see some dots show up next to the Blog and News links in
the header.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -26,13 +26,13 @@ disableAliases = true
     latexDashes = true
     extensions = [""]
 
-[outputs]
-  home = ["HTML", "RSS", "REDIR"]
-  section = ["HTML", "RSS"]
-
 [mediaTypes]
     [mediaTypes."text/netlify"]
         delimiter = ""
+
+[outputs]
+  home = ["HTML", "RSS", "REDIR", "JSON"]
+  section = ["HTML", "RSS"]
 
 [outputFormats]
     [outputFormats.RSS]
@@ -42,6 +42,8 @@ disableAliases = true
         baseName = "_redirects"
         isPlainText = true
         notAlternative = true
+    [outputFormats.JSON]
+        baseName = "_tracked_pages"
 
 [related]
     # Only include matches with rank >= threshold. This is a normalized rank between 0 and 100.
--- a/content/en/blog/2017/_index.md
+++ b/content/en/blog/2017/_index.md
@@ -3,5 +3,6 @@ title: 2017 Posts
 description: Blog posts for 2017.
 weight: 20
 icon: blog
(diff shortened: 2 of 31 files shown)


geeknoid:
Screen Shot 2019-11-17 at 19 58 56

frankbu:
This looks great. In terms of usability, I'm wondering if there's a way to "catch up" without clicking on all the articles. If not, it will probably show some large number of unread articles all the time, since many people are not likely to follow changes in any release other than the one they are currently using.

geeknoid:
Yeah, I was thinking of a Mark All Read command. I wasn't sure where to put it (menu, left nav, in the main page, etc). How about I add a button within the main page area to mark everything you see at that level and below as read?

geeknoid:
Now with "Mark all as read" button which applies to all items on the current page and below in the hierarchy. The button is only displayed if there are in fact items to be marked as read.

Screen Shot 2019-11-19 at 06 32 00

frankbu:
For News, mark all as read works great! For blog, it's a little hard to get to the button, because when you click on the Blog menu, it goes to the latest article. So, the only way to see the mark all button is by clicking on the breadcrumb to move up a level. Any idea how this could be improved? Maybe the Blog menu should go to the top level, instead of the latest article?

geeknoid:
How about this:

- Have the Blog link in the header take you to the current year's blog post page.

- Add the green dot on the section headers in the lefthand sidenav (so on "2019 Posts").

frankbu:
If you add green dots on the left hand sidenav for Blog, I think you need to do the same with News, for consistency. That might start getting a little noisy. I personally think that just pointing the Blog menu to the [top level Blog](https://deploy-preview-5768--preliminary-istio.netlify.com/blog/) would be the simplest approach and would be fine IMO.

geeknoid:
OK, I made it so the Blog link sends you to the current year's page of posts.

PR #5795, opened 2019-11-19
Use subset in the destination rule of TLS origination

vadimeisenbergibm:
This way the TLS origination will only be applied for the traffic directed by the
virtual service for TLS origination. TLS origination will not be applied for the original TLS traffic.

The problem raised by @fcantournet (cc @howardjohn)


fcantournet:
Just tested on 1.3.4 :ok_hand:

PR #5799, opened 2019-11-19
[release-1.4] rename latency graph labels

carolynhu:
To be consistent with other labels that uses dash as separator between istiomode and testmode, like `mixer-both`, `nomixer-serveronly`.  I renamed telemetryv2 label. It should be `telemetryv2_nullvm-both` not `telemetryv2-nullvm_both`.


geeknoid:
In general, please try to push changes into master and only cherry pick them into the release branch, rather than making changes directly to the release branch.

Thanks.

PR #5812, opened 2019-11-20
Add support for user feedback on how useful the content is.

geeknoid:
Here's a quick working but incomplete experiment to ask for user feedback, mostly to agree on the aesthetics.  Note that the feedback UI shows up on more pages than it should, I'll clean that up later. So given your average content page:

- Do you like the Feedback section at the bottom like that? An alternative is to put it on the right-hand side at the bottom of the table of contents.

- Do you like the Yes/No buttons? These could be thumbs up / thumbs down, happy face/sad face instead.

- At the moment, Feedback shows up as a level 2 heading, but is not listed in the table of contents on the right side. Should it be listed? 

Any other ideas?

### DIFF
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -258,3 +258,12 @@ other = "CVSS Impact Score"
 
 [security_bulletin_affected_releases]
 other = "Affected Releases"
+
+[feedback_title]
+other = "Was this information useful?"
+
+[feedback_yes]
+other = "Yes"
+
+[feedback_no]
+other = "No"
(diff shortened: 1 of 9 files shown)


geeknoid:
Instead of making this look like its part of the normal page content, it could have a different look. Different coloring, more spacing, maybe centered in the page, etc?

geeknoid:
Probably after a user clicks Yes/No, I should grey out the buttons...

frankbu:
I like the feedback section as you have it.

I wouldn't put it on _index.md, just index.md files. Maybe it should be before "See Also"?

I think it wold be OK in the index, esp if you put it before See Also.

PR #5817, opened 2019-11-21
Update index.md

ASaiun:
changed the command when user install istio with the helm v3 (without Tiller)

### DIFF
--- a/content/en/docs/setup/install/helm/index.md
+++ b/content/en/docs/setup/install/helm/index.md
@@ -80,7 +80,7 @@ deployed and you don't want to install it.
     (CRDs) using `kubectl apply`:
 
     {{< text bash >}}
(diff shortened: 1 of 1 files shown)


ostromart:
All our testing so far has been with helm v2, we haven't tested at all with v3. First, we would need to change our tests to use v3. 
For the time being we should make it clear that the docs are for v2 and migrate to v3 once tests are running with that.

ericvn:
I haven't tried this, but if the change above works with both Helm v3 and Helm v2 what's the problem with making the update? We want to force not working with Helm 3 even though we have a warning? Note that we already have a warning for an intermediate PR. @howardjohn had a change requested so for sure need his feedback.

howardjohn:
@ericvn I am not sure this actually makes helm 3 work. There is no testing. Helm 3 is intentionally NOT backwards compatible, so pretending it is doesn't seem like a good idea

frankbu:
Closing because these nstructions have been removed.

PR #5822, opened 2019-11-22
Clarified istioctl local case

davidebbo:
As suggested in https://github.com/istio/istio/issues/18978

### DIFF
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -34,7 +34,7 @@ Note that `x` in the command is because this is currently an experimental featur
 
 The example above is doing analysis on a live cluster. But the tool also supports performing analysis of a set of local yaml configuration files, or on a combination of local files and a live cluster.
 
-Analyze a specific set of local files:
+Analyze a specific set of local Kubernetes resource yaml files:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -34,7 +34,7 @@ Note that `x` in the command is because this is currently an experimental featur
 
 The example above is doing analysis on a live cluster. But the tool also supports performing analysis of a set of local yaml configuration files, or on a combination of local files and a live cluster.
 
-Analyze a specific set of local files:
+Analyze a specific set of local Kubernetes resource yaml files:
```
frankbu: It still doesn't seem very clear what this does. I can see that with the `-k` option, it analyzes the effect of applying these yaml files in the context of the current cluster's existing config. But without `-k`, what's the context for the analysis? What would it do, if I just pass in a yaml file with a single Service definition, for example? Seems it's not particularly interesting unless you pass in the whole set of yaml files that you plan to use, or am I missing something? Regardless, I think it wouldn't hurt to add a sentence or two in this doc to explain better what it does.

davidebbo: Thanks for the feedback. I expanded the explanation a bit.

PR #5831, opened 2019-11-22
Fix problems expanding entries in the sidebar, especially on Firefox.

geeknoid:


frankbu:
Hmm, this works better in Chrome now (perfectly, actually), but Firefox is still not working. Clicking to expand sections in the outline does nothing.

PR #5836, opened 2019-11-23
Site improvements.

geeknoid:
- The Yes/No feedback buttons now produce analytics events that include the
page language so that we can key off of that in the reports.

- Improve the layout of directory pages (_index.md) by removing excessive
blank space around each entry, which makes better use of the display area.

- Hide the Yea/No feedback buttons when printing.

- Fix the superscript link endnotes used when printing. These were
being rendered as full-size text instead of superscripts. This has
been broken for a long time it seems.

### DIFF
--- a/layouts/partials/primary_bottom.html
+++ b/layouts/partials/primary_bottom.html
@@ -39,8 +39,8 @@
             <div id="feedback">
                 <div>
                     {{ i18n "feedback_title" }}<br>
-                    <button class="btn feedback" onclick="sendFeedback(1)">{{ i18n "feedback_yes" }}</button>
-                    <button class="btn feedback" onclick="sendFeedback(0)">{{ i18n "feedback_no" }}</button>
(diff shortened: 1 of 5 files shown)


adammil2000:
In a past role, I found a significant jump in engagement if you put the survey control at the top of the page instead of the bottom. It made enough difference that more articles met the threshold for statistical significance because of the higher volume, allowing us to then consider their rating.

geeknoid:
@adammil2000 That's interesting. Last week, I looked at probably 20 different sites and found that all but one at the feedback buttons at the bottom. The only one at the top I found was docs.microsoft.com.

I wanted to initially put it at the top too, but I couldn't find a clean unobtrusive place to render these. I was just going to use thumbs up/thumbs down iconography for this, but I couldn't figure out a nice place to put these that I didn't find distracting. So I followed the masses and put the buttons at the bottom, using plain text.

PR #5845, opened 2019-11-25
zh-translation:content/zh/faq/distributed-tracing/how-envoy-based-tra…

tanjunchen:
[ ] Docs


tanjunchen:
@rootsongjc  so fast! Thanks! CI seems good!

PR #5854, opened 2019-11-25
Add info on embroidering the logo.

geeknoid:


geeknoid:
The zip file has the same files, I just renamed them all to Logo.<ext>.

PR #5861, opened 2019-11-25
Add a web site change log page.

geeknoid:


geeknoid:
@adammil2000 @frankbu 

So I put this together after Adam mentioned this would be useful for customers. Now, I called if a Change Log, but that's not really what this is. This list is showing the last time any given page was changed. It doesn't list the same page N times, once per change. It only lists a given page once with its latest change date.

So given this, seems like the name Change Log isn't quite right. Can you guys think of a better name?

PS: I don't have the information necessary to turn this into an actual change log, I only have access to the last modified date of each file, not their history. If we really want a change log, I can probably figure out a way to hack that in. But having folks look at Git might be better for that kind of detail.

frankbu:
Maybe "Website Change Record" would be better? I'm not sure this information is really very helpful though. It seems pretty unwieldy. Lots of files that are changed are just to fix link to other files that have moved and since it doesn't show you what's change in these files, I don't think it's really very helpful.

geeknoid:
@adammil2000 So Adam, is this really what you've had folks ask for? Frank is right that this shows any change to the files, including trivial updates such as internal link updates because some other document has been moved. Also, a lot of the doc titles aren't great in this context (for example, several are called Change Notes). 

I imagine that what customers would really want is a list of semantically significant doc changes. To do that, we'd need to grovel through the change logs and use a heuristic to decide what constitutes a "significant doc change", probably by looking at the size of each diff. That would probably be several days' worth of work to put in place, which is probably not worth the effort.

sdake:
https://deploy-preview-5861--preliminary-istio.netlify.com/about/log/ LGTM - although you are right the terminology "Website Change Log" is misleading.  I'm not totally clear what value this brings without showing the actual change - which may be too difficult to actually be worth the trouble :)

geeknoid:
@adammil2000 @frankbu @sdake A few changes:

- Don't show directory pages.
- When you hover over a link, you get the page's description.
- Each entry now shows the commit subject line.
- Remove site test pages from the list.

Is this more useful?

adammil2000:
> Is this more useful?

Yes this fantastic! The customers I spoke to just want line of sight into the changes to the content, even if some are minor. It is especially useful in one of the cases where they do presentations to other customers and they can check the latest changes and tweak presentations accordingly if needed.

You could also call these "Content updates" if it's more clear.

frankbu:
@geeknoid That is a lot better. Still not sure how useful it is. Also, will spell checking PR titles cause problems?

geeknoid:
Another update: Instead of bulleted lists, the info is now rendered in tables. I also renamed the page to "Website Content Changes".

@frankbu This won't cause problems with spell checking. We spell check markdown, and this is not markdown content.

adammil2000:
This is also very helpful as an alternative when there isn't an RSS feed for monitoring content.

PR #5878, opened 2019-11-26
zh-translation:/docs/tasks/security/citadel-config/health-check/index.md

yifanyou:
ref: https://github.com/servicemesher/istio-official-translation/issues/1513


tanjunchen:
@yifanyou 
first:
image

second:
CI unhappy if you delete /zh in url such as Line 7

yifanyou:
@tanjunchen I have fixed it. But it looks like this doc had been translated before.

PR #5886, opened 2019-11-27
Fix cni doc

richardwxn:
the original doc doesn't work, related issue: https://github.com/istio/istio/issues/19230
we need to provide overlay by IstioControlPlane CR file instead of set overlay and also specifies force flag now.

### DIFF
--- a/content/en/docs/setup/additional-setup/cni/index.md
+++ b/content/en/docs/setup/additional-setup/cni/index.md
@@ -88,12 +88,26 @@ Refer to the [Customizable Install with Helm](/docs/setup/install/helm) for comp
 Use the following command to render and apply Istio CNI components and override the default configuration of the
 `logLevel` and `excludeNamespaces` parameters for `istio-cni`:
 
+Create a IstioControlPlane CR yaml locally with your override to install istio, e.g. cni.yaml
+
+{{< text yaml >}}
+apiVersion: install.istio.io/v1alpha2
+kind: IstioControlPlane
+spec:
+  cni:
+    enabled: true
+  values:
+    cni:
+      logLevel: info
+      excludeNamespaces:
+       - istio-system
+       - kube-system
+       - foo_ns
+       - bar_ns
+{{< /text >}}
+
 {{< text bash >}}
-$ istioctl manifest apply \
-    --set <flags you used to install Istio>
-    --set cni.enabled=true \
-    --set values.cni.logLevel=info \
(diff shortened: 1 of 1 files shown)


richardwxn:
@ostromart PTAL

thecodejunkie:
@richardwxn A bit further down you also have outdated instructions for installing CNI on GKE https://github.com/istio/istio.io/blob/1637aa930975faa32dcebaa278e9d276c4792bcc/content/en/docs/setup/additional-setup/cni/index.md#gke-setup this should probably be replaced by docs that shows how to set `cniBinDir` using the IstioControlPlane

content/en/docs/setup/additional-setup/cni/index.md
```diff
@@ -88,12 +88,26 @@ Refer to the [Customizable Install with Helm](/docs/setup/install/helm) for comp
...
 
+Create a IstioControlPlane CR yaml locally with your override to install istio, e.g. cni.yaml
+
+{{< text yaml >}}
+apiVersion: install.istio.io/v1alpha2
+kind: IstioControlPlane
+spec:
+  cni:
+    enabled: true
+  values:
```
ostromart: Instead of using force flag, you can put the values not in the schema in unvalidatedValues for now, and could you please create an issue to fix the schema? Thanks!

richardwxn: done

richardwxn:
@thecodejunkie The reason why in this case we have to use IstioControlPlane CR file is because we cannot do --set overlay for array values(e.g. excludeNamespaces), but for the other fields overlay, for example cniBinDir it is ok. Also you can use the file overlay and set overlay together as well.

thecodejunkie:
@richardwxn Sorry I should have been more clear. My remark wasn't about the `--set` flag per se, but rather that the GKE docs still use `helm install` and not `Istioctl` with enabling cni and setting `cniBinDir`

```
istioctl manifest install \
  --set cni.enabled=true \
  --set values.cni.cniBinDir=/home/kubernetes/bin
```

or by doing the same but using an IstioControlPlane file instead.

richardwxn:
@thecodejunkie oh ok, thx for point out, I am not familiar with this page. But based on what you said, I found out a couple more places with outdated description and include the update in this PR

richardwxn:
@geeknoid Can you help take a look?

richardwxn:
seems like the linter failed for a weird and unrelated reason

Issue #5915, opened 2019-11-28
'Deny all' authorization policy example doesn't work


rolandkool:
The deny-all example authorization policy as described on this page does not work: https://istio.io/docs/reference/config/security/authorization-policy/

When applying it, the following error is thrown:

> The AuthorizationPolicy "deny-all" is invalid: []: Invalid value: map[string]interface {}{"apiVersion":"security.istio.io/v1beta1", "kind":"AuthorizationPolicy", "metadata":map[string]interface {}{"annotations":map[string]interface {}{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"security.istio.io/v1beta1\",\"kind\":\"AuthorizationPolicy\",\"metadata\":{\"annotations\":{},\"name\":\"deny-all\",\"namespace\":\"rkl\"},\"spec\":null}\n"}, "creationTimestamp":"2019-11-28T13:10:18Z", "generation":1, "name":"deny-all", "namespace":"foo", "uid":"6cd403b8-11e0-11ea-b8c3-42010a210004"}, "spec":interface {}(nil)}: validation failure list:
spec in body must be of type object: "null"

PR #5916, opened 2019-11-28
fix deny all authorization policy examples

rolandkool:
To fix  https://github.com/istio/istio.io/issues/5915


geeknoid:
These are generated files. Please fix the original errors in the source material in the istio/api repo.

Thanks.

PR #5953, opened 2019-12-02
Add support for events.

geeknoid:
Events are used for special announcements. There are stickers and banners that can be
displayed to the user. These can be used to announce an imminent release with a
sticker and countdown clock, or can be used to invite users to a future
conference, or can be used to announce that a new release is available for download.

See the authoring guide for instructions on how to use these announcements.


content/en/events/index.md
```diff
@@ -0,0 +1,3 @@
+---
```
frankbu: What it is this for?

content/en/events/README.md
```diff
@@ -0,0 +1,34 @@
...
+then this event is considered to have happened at the release date of the latest Istio release. This is used to
+automatically display a banner that says "Istio X.Y.Z has just been released" whenever a new release is produced.
+
+- event_date. The date of the event in YYYY-MM-DD format.
+
+- countdown_days. How long before the event to display the item to the user.
+
+- event_days. How long after the event to display the item to the user.
+
+- visibility. How many times to show the content to the user, or the string `always` to display it on every page during
```
frankbu: I wonder if a boolean would be sufficient ... display once or always. Picking a number in between would seem hard, and might end up looking like random behavior. Anyway, if it's not hard to keep track, i guess it doesn't hurt to allow any number.

content/en/events/README.md
```diff
@@ -0,0 +1,34 @@
...
+the event's period. A value of 3 would mean that the first three pages visited by the user during the period will display
+the content, and the content will be hidden on subsequent page loads.
+
+- timeout. The amount of time the content is visible to the user on a given page. After that much time passes, the item will be removed from
```
frankbu: Maybe this is overkill too, unless it's easy to support?

content/en/events/index.md
geeknoid: Ignorable infra stuff. This just tells Hugo that this whole tree is not to be rendered as normal pages. There's the same kind of file at the top of the boilerplates directory for the same reason.

content/en/events/README.md
geeknoid: That's really easy to support, but also really frivolous...

PR #5965, opened 2019-12-03
Update index.html

sureshvis:
Even with permissive mode, the filter chain adds "require_client_certificate": true" and required you to pass client cert.

Even with permissive mode, the filter chain adds "require_client_certificate": true" and required you to pass client cert.

put an X in all the areas that this PR affects. @rshriram


sureshvis:
will resend

PR #5980, opened 2019-12-04
Now collect additional info when a user says a page is not helpful.

geeknoid:


i18n/en.toml
```diff
@@ -274,6 +274,15 @@ other = "Yes"
 [feedback_no]
 other = "No"
 
+[feedback_ask_for_comment]
+other = "What could make this information more useful?"
```
frankbu: Maybe this could be worded to sound more optional? For example:
```suggestion
other = "Do you have any suggestions for improvement?"
```

PR #5989, opened 2019-12-05
Remove incorrect escape character.

mt-inside:
Also add full syntax for the YAML value, to remove ambiguity.


geeknoid:
This is a generated files. Please reapply this fix in the original source in the istio/api repo.

Thanks.

PR #5994, opened 2019-12-05
Warn helm 3 is not supported

frankbu:
Ref: https://github.com/istio/istio.io/pull/5817, https://github.com/istio/istio/issues/18280


frankbu:
I believe archived docs are read-only. @geeknoid can confirm

PR #6012, opened 2019-12-06
add a note to known problems about node autoscaler

kyessenov:
/assign @howardjohn


frankbu:
@kyessenov I assume you want this in the current istio.io (1.4) docs, so it needs to go to the release-1.4 branch. I added a cherry-pick label to do that, but if you expect this will be fixed in 1.5, the PR itself should go to the release-1.4 branch so it's not also in master. Please advise.

kyessenov:
I don't have any expectation on when it'll be fixed. Let's send to both branches, and if it does get solved, the docs would need to be updated.

xref: https://github.com/istio/istio/issues/19395

PR #6028, opened 2019-12-08
Correct pilot metric description

johscheuer:
Correct Pilot metric description.


johscheuer:
See also: https://github.com/istio/istio/pull/19462

geeknoid:
This is a generated file. Please fix the original in the Pilot sources in istio/istio instead.

PR #6038, opened 2019-12-09
Update reference docs.

selmanj:
I ran `make update_ref_docs` in order to get up-to-date analysis messages. 

However I'm a bit worried about the changes where the `aliases` field is renamed. Seems like this could break some url paths - any idea why this was done?


geeknoid:
Changes to the aliases field are fine, no worries.

PR #6042, opened 2019-12-10
Fix broken link for Configure Citadel Service Account Secret Generation

shamsher31:
Link for secrete creation page was broken with extra `(` to generate link.

[X] Docs
[X] User Experience


shamsher31:
The failed test is related to zh docs. I'm not sure if this test is valid for en.
/retest

shamsher31:
Looks like @tanjunchen  #6041 will resolve failing test.

PR #6046, opened 2019-12-10
correct-the-timestamp-value-to-be-as-same-as-it-is-in-the-sample

vivian99-wu:
In sample, the timestamp value in the outputs of logs for "istio-telemetry" pods is "warn". While, in the section "Understanding the logs configuration", it is descripted as "warning". I fix this kind of expression errors.

### DIFF
--- a/content/en/docs/tasks/observability/logs/collecting-logs/index.md
+++ b/content/en/docs/tasks/observability/logs/collecting-logs/index.md
@@ -86,7 +86,7 @@ The `kind: instance` stanza of configuration defines a schema for generated log 
 to generate log entries for requests based on the attributes reported by Envoy.
 
 The `severity` parameter is used to indicate the log level for any generated
-`logentry`. In this example, a literal value of `"warning"` is used. This value will
+`logentry`. In this example, a literal value of `"warn"` is used. This value will
 be mapped to supported logging levels by a `logentry` *handler*.
 
 The `timestamp` parameter provides time information for all log entries. In this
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/observability/logs/collecting-logs/index.md
```diff
@@ -86,7 +86,7 @@ The `kind: instance` stanza of configuration defines a schema for generated log
 to generate log entries for requests based on the attributes reported by Envoy.
 
 The `severity` parameter is used to indicate the log level for any generated
-`logentry`. In this example, a literal value of `"warning"` is used. This value will
+`logentry`. In this example, a literal value of `"warn"` is used. This value will
```
frankbu: I'm not sure this change is correct. Isn't this referring to the value in https://raw.githubusercontent.com/istio/istio/master/samples/bookinfo/telemetry/log-entry-crd.yaml?

frankbu: @douglas-reid PATL

douglas-reid: Yeah, I feel like, if anything, the sample output needs to be updated, not the text here.

Can you run through the task and confirm and put in an updated output to match (or, show the output and confirm that the sample yaml and this text should be updated) ?

vivian99-wu: Okey，I'll give it a try.

PR #6093, opened 2019-12-14
zh-translation: docs/reference/glossary/multi-mesh.md

gauliang:
Ref: https://github.com/servicemesher/istio-official-translation/issues/1265


gauliang:
@rootsongjc

PR #6102, opened 2019-12-15
zh-translation: some references update from previous translation.

ilylia:
ref: https://github.com/servicemesher/istio-official-translation/issues/1629
ref: https://github.com/servicemesher/istio-official-translation/issues/1627
ref: https://github.com/servicemesher/istio-official-translation/issues/1626
ref: https://github.com/servicemesher/istio-official-translation/issues/1623
ref: https://github.com/servicemesher/istio-official-translation/issues/1619
ref: https://github.com/servicemesher/istio-official-translation/issues/1613
ref: https://github.com/servicemesher/istio-official-translation/issues/1283

PR #6113, opened 2019-12-16
zh-translation: apply some suggested change

ilylia:
ref: #6098 
ref: #6102


yuxiaobo96:
@rootsongjc 
是对之前合并的 pr 的完善修改，可以合并

PR #6115, opened 2019-12-16
Fix typo

Youmoo:
"less than" not "less that".


Youmoo:
@googlebot I signed it!

PR #6128, opened 2019-12-17
update the cmd to retrieve token correctly

linsun:
Before the PR:

$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d '\t')
kubectl exec $(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name}) -c sleep -n foo -- curl https://kubernetes.default/api --header "Authorization: Bearer $TOKEN" --insecure -s -o /dev/null -w "%{http_code}\n"

403

Because the token had started with a few empty strings.

After the PR:
$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -ds ' ' '\t')
(⎈ |linistio12/bnsdnfgw0ltk2au0ssr0:default)
$ kubectl exec $(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name}) -c sleep -n foo -- curl https://kubernetes.default/api --header "Authorization: Bearer $TOKEN" --insecure -s -o /dev/null -w "%{http_code}\n"
200

[ x ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/security/authentication/authn-policy/index.md
+++ b/content/en/docs/tasks/security/authentication/authn-policy/index.md
@@ -230,7 +230,7 @@ The Kubernetes API server doesn't have a sidecar, thus request from Istio servic
 requests to any non-Istio service.
 
 {{< text bash >}}
-$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d '\t')
+$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d ' \t')
 $ kubectl exec $(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name}) -c sleep -n foo -- curl https://kubernetes.default/api --header "Authorization: Bearer $TOKEN" --insecure -s -o /dev/null -w "%{http_code}\n"
 000
 command terminated with exit code 35
@@ -261,7 +261,7 @@ is automatically injected into the system when you install Istio with mutual TLS
 Re-run the testing command above to confirm that it returns 200 after the rule is added:
 
 {{< text bash >}}
-$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d ' ')
+$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d ' \n')
 $ kubectl exec $(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name}) -c sleep -n foo -- curl https://kubernetes.default/api --header "Authorization: Bearer $TOKEN" --insecure -s -o /dev/null -w "%{http_code}\n"
 200
 {{< /text >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/security/authentication/authn-policy/index.md
```diff
@@ -261,7 +261,7 @@ is automatically injected into the system when you install Istio with mutual TLS
 Re-run the testing command above to confirm that it returns 200 after the rule is added:
 
 {{< text bash >}}
-$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d '\t')
+$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -ds ' ' '\t')
```
frankbu: Doesn't `tr -ds ' ' '\t'` mean delete all `'  '` and truncate multiple `'\t'` into a single `\t`? Is that exactly what you want? The original command deleted `'\t'` chars.

ericvn: You could change to tr -d '\t' | tr -d ' ' keeping the original remove all tabs and then remove all spaces (including potentially those in the token). If you just want to remove the leading spaces, maybe tr -d '\t' | sed 's/^ *//g'

linsun: Thank you @frankbu @ericvn!  I misunderstood usage of -s, thought it is multiple strings :)

I removed \t as it is not needed and retested the cmd.

PTAL

frankbu: Are you sure `\t` is not needed? Maybe on some other platforms it will have a `\t`. Just to be safe you can remove both with `-d ' \t'`

ericvn: Agree with @frankbu's last comment if instead you want to remove all tabs (which was being done before and probably shouldn't be removed) and all spaces (I don't think the token will have an embedded space).

linsun: Ok, I've added it back.  Yes, it was not needed for me but perhaps needed for some other k8s versions.

content/en/docs/tasks/security/authentication/authn-policy/index.md
```diff
@@ -261,7 +261,7 @@ is automatically injected into the system when you install Istio with mutual TLS
 Re-run the testing command above to confirm that it returns 200 after the rule is added:
 
 {{< text bash >}}
-$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d '\t')
+$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d ' \n')
```
ericvn:
```suggestion
$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default-token | cut -f1 -d ' ' | head -1) | grep -E '^token' | cut -f2 -d':' | tr -d ' \t')
```

PR #6146, opened 2019-12-18
update authorization

linsun:
done in istio proxy now since istio 1.1 which is why mixer policy check is disabled by default: https://github.com/istio/istio/issues/11736

### DIFF
--- a/content/en/docs/concepts/security/index.md
+++ b/content/en/docs/concepts/security/index.md
@@ -46,12 +46,12 @@ Security in Istio involves multiple components:
 
 - **Citadel** for key and certificate management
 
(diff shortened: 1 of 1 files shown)


frankbu:
Not sure we need to update anything in this doc since it is in the process of being replaced by a totally new security concepts rewrite.

cc: @rcaballeromx

linsun:
cc @johnma14

linsun:
@frankbu I created a 1.4 PR - https://github.com/istio/istio.io/pull/6148.  I'm good to have this corrected in 1.5 doc in the future.

frankbu:
Fixed in #6148

PR #6154, opened 2019-12-19
remove bin reference to istioctl

linsun:
as all of our other tasks assume istioctl is on the path already.  Having it cause me an alert on my mac:

“istioctl” can’t be opened because Apple cannot check it for malicious software.


frankbu:
@linsun I noticed a similar issue in this file: https://github.com/istio/istio.io/blob/master/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
It seems to use `./istioctl` for the commands. Would you mind fixing those ones in this PR as well?

linsun:
@frankbu done!  PTAL - thanks!

linsun:
https://github.com/istio/istio.io/pull/6156

PR #6172, opened 2019-12-22
Fix protocol selection Helm values

nacx:
Fix the protocol selection page and use the new Helm values used by Pilot.

### DIFF
--- a/content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
+++ b/content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
@@ -48,11 +48,11 @@ spec:
     name: http-web
 {{< /text >}}
 
-## Automatic protocol selection (experimental)
+## Automatic protocol selection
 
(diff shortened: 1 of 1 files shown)


ericvn:
See #6278, but should this text show that this is enabled for both inbound and outbound by default in 1.5?

content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -52,4 +52,7 @@ spec:
 
 Istio can automatically detect HTTP and HTTP/2 traffic. If the protocol cannot automatically be determined, traffic will be treated as plain TCP traffic.
 
-This feature is experimental and off by default. It can be turned on by providing the install option `--set values.pilot.enableProtocolSniffing=true`.
+This feature is experimental and only enabled for outbound traffic by default. It can be turned on by providing the following install options:
```
yxue: On master, both inbound and outbound are enabled by default.

nacx:
Changed the text to indicate it is now enabled by default.

PR #6181, opened 2019-12-23
zh-translations: /docs/tasks/observability/metrics/tcp-metrics/index.md

ilylia:
ref: https://github.com/servicemesher/istio-official-translation/issues/1633


tanjunchen:
@rootsongjc
add translation/chinese lavel for calculating data.

Issue #4735, opened 2019-07-29
Clarify mesh expansion concept


linsun:
We are taking a pass of MC concept, would need to ensure our mesh expansion doc is current and concept is clear.

linsun:
#4736

linsun:
@irisdingbj @vadimeisenbergibm anything left here?

vadimeisenbergibm:
@linsun I will work on it next week.

howardjohn:
```
$ rg 'mesh expansion' -i conent/en
```
no results

PR #6209, opened 2019-12-25
add mesh expansion to the deployment models document

vadimeisenbergibm:
Issue https://github.com/istio/istio.io/issues/4735. 

Any comments are welcome. Let's work together on the definition and the explanations.

[X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -435,7 +435,7 @@ reasons for deployments on VMs/bare metal computers could be special hardware re
 
 Adding the workloads running on VMs or bare metal systems to a mesh, you can benefit from all the
 service mesh features: traffic management, security, observability, and others. You can deploy the Istio control plane(s)
-on some of your clusters and integrate your VMs/bare metal computers with the mesh.
+on your cluster or clusters, you can add your VMs or bare metal systems to the mesh.
 
 Adding a VM/bare metal computer to a mesh means installing an Istio sidecar and a Node agent on the machine and
 configuring them to communicate with the control plane(s) on some of the clusters. Once you add a VMs/bare metal
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -437,7 +437,9 @@ Adding the workloads running on VMs or bare metal systems to a mesh, you can ben
 service mesh features: traffic management, security, observability, and others. You can deploy the Istio control plane(s)
 on your cluster or clusters, you can add your VMs or bare metal systems to the mesh.
 
-Adding a VM/bare metal computer to a mesh means installing an Istio sidecar and a Node agent on the machine and
+Adding a VM or a bare metal system to a mesh requires the following steps:
+
+1. Install an Istio sidecar and a Node agent on the VM or bare metal system.
 configuring them to communicate with the control plane(s) on some of the clusters. Once you add a VMs/bare metal
 computers to a mesh, you can configure the control plane(s) of the mesh to control communication between services on the
 clusters and on the machines (from a cluster to a VM, from a VM to a cluster, from a VM to another VM in the mesh).
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -433,7 +433,7 @@ concepts described in the [multiple meshes](#multiple-meshes) model.
 In same cases you might have to deploy some of the services on virtual machines or bare metal computers. The
 reasons for deployments on VMs/bare metal computers could be special hardware requirements, tighter isolation, tighter integration with the host's OS, legacy deployments.
 
-Having some of the services on VMs/bare metal, you may want to connect the services into a mesh and benefit from all the
+Adding the workloads running on VMs or bare metal systems to a mesh, you can benefit from all the
 service mesh features: traffic management, security, observability, and others. You can deploy the Istio control plane(s)
 on some of your clusters and integrate your VMs/bare metal computers with the mesh.
 
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -428,7 +428,7 @@ namespace.
 When each team has their own mesh, cross-mesh communication follows the
 concepts described in the [multiple meshes](#multiple-meshes) model.
 
-## Virtual machines and bare metal
+## Virtual machines and bare metal systems
 
 In same cases you might have to deploy some of the services on virtual machines or bare metal computers. The
 reasons for deployments on VMs/bare metal computers could be special hardware requirements, tighter isolation, tighter integration with the host's OS, legacy deployments.
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -12,6 +12,7 @@ keywords:
   - trust
   - single-mesh
   - multiple-meshes
+  - mesh-expansion
 aliases:
   - /docs/concepts/multicluster-deployments
   - /docs/concepts/deployment-models
@@ -27,6 +28,7 @@ or are there multiple control planes deployed to ensure high availability (HA)?
 If there is more than one cluster being deployed, and more specifically in isolated networks,
 are they going to be connected into a single {{< gloss >}}multicluster{{< /gloss >}}
 service mesh or will they be federated into a {{< gloss >}}multi-mesh{{< /gloss >}} deployment?
+Will services on virtual machines or bare metal computers be connected to the mesh?
 
 All of these questions, among others, represent independent dimensions of configuration for an Istio deployment.
 
@@ -34,6 +36,7 @@ All of these questions, among others, represent independent dimensions of config
 1. single or multiple network
 1. single or multiple control plane
 1. single or multiple mesh
+1. mesh expansion (integration with VMs/bare metal)
 
 All combinations are possible, although some are more common than others and
 some are clearly not very interesting (for example, multiple mesh in a single cluster).
@@ -424,3 +427,21 @@ namespace.
 
 When each team has their own mesh, cross-mesh communication follows the
 concepts described in the [multiple meshes](#multiple-meshes) model.
+
+## Mesh expansion
+
+In same cases you might have to deploy some of the services on virtual machines or bare metal computers. The
+reasons for deployments on VMs/bare metal computers could be special hardware requirements, tighter isolation, tighter integration with the host's OS, legacy deployments.
+
+Having some of the services on VMs/bare metal, you may want to connect the services into a mesh and benefit from all the
+service mesh features: traffic management, security, observability, and others. You can deploy the Istio control plane(s)
+on some of your clusters and expand the mesh of the cluster(s) by adding VMs/bare metal computers into the mesh.
+
+The VMs/bare metal computers can share the same network with the clusters of the mesh, or they can be on
+different networks. In the latter case, [Istio gateways](/docs/concepts/traffic-management/#gateways) can be used to
+provide reachability between the VMs/bare metal and the clusters.
+
+Adding a VM/bare metal computer to a mesh means installing an Istio sidecar and a Node agent on the machine and
+configuring them to communicate with the control plane(s) on some of the clusters. Once you add a VMs/bare metal
+computers to a mesh, you can configure the control plane(s) of the mesh to control communication between services on the
+clusters and on the machines (from a cluster to a VM, from a VM to a cluster, from a VM to another VM in the mesh).
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -12,7 +12,7 @@ keywords:
   - trust
   - single-mesh
   - multiple-meshes
-  - mesh-expansion
+  - virtual-machines
 aliases:
   - /docs/concepts/multicluster-deployments
   - /docs/concepts/deployment-models
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -36,7 +36,7 @@ All of these questions, among others, represent independent dimensions of config
 1. single or multiple network
 1. single or multiple control plane
 1. single or multiple mesh
-1. mesh expansion (integration with VMs/bare metal)
+1. integration with VMs/bare metal
 
 All combinations are possible, although some are more common than others and
 some are clearly not very interesting (for example, multiple mesh in a single cluster).
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -36,7 +36,7 @@ All of these questions, among others, represent independent dimensions of config
 1. single or multiple network
 1. single or multiple control plane
 1. single or multiple mesh
-1. integration with VMs/bare metal
+1. integration of VMs or bare metal systems
 
 All combinations are possible, although some are more common than others and
 some are clearly not very interesting (for example, multiple mesh in a single cluster).
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -437,10 +437,6 @@ Having some of the services on VMs/bare metal, you may want to connect the servi
 service mesh features: traffic management, security, observability, and others. You can deploy the Istio control plane(s)
 on some of your clusters and integrate your VMs/bare metal computers with the mesh.
 
-The VMs/bare metal computers can share the same network with the clusters of the mesh, or they can be on
-different networks. In the latter case, [Istio gateways](/docs/concepts/traffic-management/#gateways) can be used to
-provide reachability between the VMs/bare metal and the clusters.
-
 Adding a VM/bare metal computer to a mesh means installing an Istio sidecar and a Node agent on the machine and
 configuring them to communicate with the control plane(s) on some of the clusters. Once you add a VMs/bare metal
 computers to a mesh, you can configure the control plane(s) of the mesh to control communication between services on the
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -28,7 +28,7 @@ or are there multiple control planes deployed to ensure high availability (HA)?
 If there is more than one cluster being deployed, and more specifically in isolated networks,
 are they going to be connected into a single {{< gloss >}}multicluster{{< /gloss >}}
 service mesh or will they be federated into a {{< gloss >}}multi-mesh{{< /gloss >}} deployment?
-Will services on virtual machines or bare metal computers be connected to the mesh?
+Will services running on virtual machines or bare metal computers be part of the mesh?
 
 All of these questions, among others, represent independent dimensions of configuration for an Istio deployment.
 
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -440,6 +440,6 @@ on your cluster or clusters, you can add your VMs or bare metal systems to the m
 Adding a VM or a bare metal system to a mesh requires the following steps:
 
 1. Install an Istio sidecar and a Node agent on the VM or bare metal system.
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,21 @@ namespace.
 
 When each team has their own mesh, cross-mesh communication follows the
 concepts described in the [multiple meshes](#multiple-meshes) model.
+
+## Mesh expansion
```
frankbu: I think we decided not to use the term "Mesh Expansion" because it's not a known industry term and any configuration that adds services to the mesh would be considered "expanding the mesh".
```suggestion
## Virtual machines and bare metal
```

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -27,13 +28,15 @@ or are there multiple control planes deployed to ensure high availability (HA)?
...
 service mesh or will they be federated into a {{< gloss >}}multi-mesh{{< /gloss >}} deployment?
+Will services on virtual machines or bare metal computers be connected to the mesh?
 
 All of these questions, among others, represent independent dimensions of configuration for an Istio deployment.
 
 1. single or multiple cluster
 1. single or multiple network
 1. single or multiple control plane
 1. single or multiple mesh
+1. mesh expansion (integration with VMs/bare metal)
```
frankbu:
```suggestion
1. Integration with VMs/bare metal
```

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -27,13 +28,15 @@ or are there multiple control planes deployed to ensure high availability (HA)?
 If there is more than one cluster being deployed, and more specifically in isolated networks,
 are they going to be connected into a single {{< gloss >}}multicluster{{< /gloss >}}
 service mesh or will they be federated into a {{< gloss >}}multi-mesh{{< /gloss >}} deployment?
+Will services on virtual machines or bare metal computers be connected to the mesh?
```
frankbu:
```suggestion
Will services running on virtual machines or bare metal computers be part of the mesh?
```

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -12,6 +12,7 @@ keywords:
   - trust
   - single-mesh
   - multiple-meshes
+  - mesh-expansion
```
frankbu:
```suggestion
  - virtual-machines
```

frankbu:
@nmittler @smawson WDYT?

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: I don't think "mesh expansion" is the right phrase ... nobody knows what that means initially :). We should say VMs, or something like that.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,21 @@ namespace.
...
 concepts described in the [multiple meshes](#multiple-meshes) model.
+
+## Mesh expansion
+
+In same cases you might have to deploy some of the services on virtual machines or bare metal computers. The
+reasons for deployments on VMs/bare metal computers could be special hardware requirements, tighter isolation, tighter integration with the host's OS, legacy deployments.
+
+Having some of the services on VMs/bare metal, you may want to connect the services into a mesh and benefit from all the
+service mesh features: traffic management, security, observability, and others. You can deploy the Istio control plane(s)
+on some of your clusters and expand the mesh of the cluster(s) by adding VMs/bare metal computers into the mesh.
```
nmittler: Again, I'd lose the phrase "expand the mesh". Let's go with Frank's phrasing of "integrate with...".

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,21 @@ namespace.
...
+
+The VMs/bare metal computers can share the same network with the clusters of the mesh, or they can be on
```
nmittler: This paragraph seems redundant with other concepts. I think we can lose it.

content/en/docs/ops/deployment/deployment-models/index.md
vadimeisenbergibm: Addressed in https://github.com/istio/istio.io/pull/6209/commits/26fc6d5dcbe19d51ecd80c46190beedf0aec6456.

content/en/docs/ops/deployment/deployment-models/index.md
vadimeisenbergibm: Done in https://github.com/istio/istio.io/pull/6209/commits/3f4ce88594fb82b137a62288eea4f7a30f72a0de.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,17 @@ namespace.
...
+
+In same cases you might have to deploy some of the services on virtual machines or bare metal computers. The
+reasons for deployments on VMs/bare metal computers could be special hardware requirements, tighter isolation, tighter integration with the host's OS, legacy deployments.
+
+Having some of the services on VMs/bare metal, you may want to connect the services into a mesh and benefit from all the
+service mesh features: traffic management, security, observability, and others. You can deploy the Istio control plane(s)
+on some of your clusters and integrate your VMs/bare metal computers with the mesh.
+
+Adding a VM/bare metal computer to a mesh means installing an Istio sidecar and a Node agent on the machine and
+configuring them to communicate with the control plane(s) on some of the clusters. Once you add a VMs/bare metal
```
rcaballeromx:
```suggestion
1. Configure the sidecar and the agent to communicate with the control plane or planes in your deployment.
```

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,17 @@ namespace.
...
+computers to a mesh, you can configure the control plane(s) of the mesh to control communication between services on the
```
rcaballeromx:
```suggestion
1. Configure the control plane or planes of your mesh to control communication between the workloads running on 
    the clusters, running on the VMs, and running on the bare metal systems, for example:
```

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,17 @@ namespace.
...
+clusters and on the machines (from a cluster to a VM, from a VM to a cluster, from a VM to another VM in the mesh).
```
rcaballeromx:
```suggestion
    - From a cluster to a VM
    - From a VM to a cluster
    - From a VM to another VM in the mesh
```

smawson:
I'm a little concerned that this doesn't quite fit into the deployment models doc as written. The rest of the doc is about describing various dimensions and the choices the platform operator should be thinking about. This VM stuff is just describing how VMs work. I think we need to rewrite it to focus on the operational and deployment concerns with VMs.

Specifically we need to talk about how VMs are deployed and attached to the mesh (and we should include a picture with an example). We need to talk about how the VM attaches to a control plane, which is typically running in a Kubernetes cluster. We could also talk about running the control plane on VMs, but maybe that is a separate section. Think about this more as an operational guide (since it is under operations): What does the operator need to think about with regard to VMs?

Ideally this would also be a "VM and Bare Metal models" section since every other section is a "X models." So how are VMs modeled, how are they operated, etc.

frankbu:
I had a similar thought to @smawson when I first read this. Even in the overview, the 5th question seems out of sync with the other four. How about adding a separate doc in the operations section something like: "Integrating VM or Bare Metal Services"?

vadimeisenbergibm:
@rcaballeromx @frankbu @smawson @nmittler 

My point was to add some high level description about integration with VMs, and not to provide any operational details. This PR basically tries to say three things: 

1. There are reasons you need to run some workloads on VMs (not everything runs on Kubernetes clusters).
1. You can integrate workloads on VMs with Istio mesh on Kubernetes clusters.
1. What you gain from such integration.

From the three points above follows that you have to take into account both workloads on your VMs and workloads on Kubernetes cluster when you plan you deployment (networks, control planes, etc.)

The operational parts ("the how") of integration of VMs with Istio are described in https://istio.io/docs/examples/virtual-machines/. The question is where to put the high level description of "the why".

I see three options:
1. Reduce this PR to the minimum that would fit the deployment models.
1. Reduce this PR and put it into the glossary item, "integration with VMs and bare metal"
1. Put this PR elsewhere as a separate article.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -27,13 +28,15 @@ or are there multiple control planes deployed to ensure high availability (HA)?
 If there is more than one cluster being deployed, and more specifically in isolated networks,
 are they going to be connected into a single {{< gloss >}}multicluster{{< /gloss >}}
 service mesh or will they be federated into a {{< gloss >}}multi-mesh{{< /gloss >}} deployment?
+Will services running on virtual machines or bare metal computers be part of the mesh?
```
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx the questions in the list above are in the future tense, the reader must decide how to create a mesh, for example: _...are they going to be connected into a single multicluster
service mesh or will they be federated into a multi-mesh..._.

So the question is: "do you want your mesh to include workloads running on VMs?" It is up to you, (the reader), you must decide.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,17 @@ namespace.
 
 When each team has their own mesh, cross-mesh communication follows the
 concepts described in the [multiple meshes](#multiple-meshes) model.
+
+## Virtual machines and bare metal
+
+In same cases you might have to deploy some of the services on virtual machines or bare metal computers. The
```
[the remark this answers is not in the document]
vadimeisenbergibm: This sentence says that sometimes you must deploy some workloads on VMs. This statement is not related to the Istio support of VMs.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,17 @@ namespace.
...
+reasons for deployments on VMs/bare metal computers could be special hardware requirements, tighter isolation, tighter integration with the host's OS, legacy deployments.
```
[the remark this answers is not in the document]
vadimeisenbergibm: See https://github.com/istio/istio.io/pull/6209#discussion_r364678463. I am OK with making the sentence as list.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -424,3 +427,17 @@ namespace.
...
+
+Having some of the services on VMs/bare metal, you may want to connect the services into a mesh and benefit from all the
+service mesh features: traffic management, security, observability, and others. You can deploy the Istio control plane(s)
```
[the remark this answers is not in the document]
vadimeisenbergibm: @rcaballeromx please note that the goal is to present Istio features in general, since this is a high-level statement. The features of Istio on general: traffic management, security, observability. The list of the Istio features is too detailed to present the reader in a high level description.

frankbu:
@vadimeisenbergibm Yes, that makes sense, but I still think it would be better as a separate article (option 3). It should be a sibling article to this one, still under operations/deployment, something like this:
```
Operations
    Deployment
        Deployment Models
        Integrating with VMs or Bare Metal
        ...
```

vadimeisenbergibm:
@frankbu OK, if we agree on that, I will move this PR to a separate article. Let's wait for the agreement from @rcaballeromx @smawson @nmittler

rcaballeromx:
@vadimeisenbergibm @frankbu @smawson @nmittler 
I tend to agree with Frank that the "why?" for VM integration should be in a different page. 
The goal of the deployment models document is to provide readers with the information they need to determine what deployment model suits their needs.
Those needs can include running workloads on VMs or bare metal systems and that dimension should be included in the document, IMO.
The piece of information that is still missing from my perspective is: What changes in the deployment if you add VMs or bare metal systems to the mesh? 
The text in the PR mentions that the Istio control plane or planes must be configured and the proxies installed, but, it is not clear, for example, how it is any different from adding workloads running on a new cluster from a deployment perspective?
If I add a cluster, the workloads running on said cluster also need to have proxies in front of them and the Istio control plane or planes need to be configured properly. 
Pointing out the differences more clearly would make the content fit better with the other sections of the document. 
Each dimension in the existing document provides the reader with new considerations for them to take into account. 
The question that needs an answer in the deployment models doc is: **What are the things readers need to consider or change in the way they deploy the mesh if the mesh will contain VMs and bare metal systems alongside the Kubernetes clusters?**

vadimeisenbergibm:
> The goal of the deployment models document is to provide readers with the information they need to determine what deployment model suits their needs. Those needs can include running workloads on VMs or bare metal systems and that dimension should be included in the document, IMO.

I agree with that. In my opinion, VMs are part of the whole deployment, together with Kubernetes clusters. So the operators of the whole infrastructure, should give thought to which workloads they deploy on Kubernetes clusters, which workloads on VMs, which workloads may run on both. They should configure the networks between the VMs and the clusters, decide which VM talks to which control plane on which cluster, and on which network it is deployed.

Having said that, I do not think we have experience and concrete guidelines how to do it, and there is a lot of multiple options. The reader should just be aware that VMs are part of the whole picture of the deployment.

Regarding: "What are the things readers need to consider or change in the way they deploy the mesh if the mesh will contain VMs and bare metal systems alongside the Kubernetes clusters?"

The readers should consider how to configure the networks between clusters and VMs, and which Istio control plane the VMs' sidecar proxies should talk to. They should also consider workloads on which clusters will talk to the workloads of the VMs and vice versa, and configure clusters/VMs accordingly - creating service entries, gateways, etc.

frankbu:
I agree VM integration is an important deployment consideration, but not an orthogonal deployment model/dimension. If we treat it like another "model" then what about external services in general? The VM integration is about bringing an external service into the mesh to get added advantages, which we should explain in the VM doc. Other than that, regardless of deployment model, users need to consider what and how services outside of the mesh are accessed. IMO, the whole topic fits awkwardly in the deployment models document, which focuses on Kubernetes deployment variations. The VM material is definitely worthy of a top level document in the operations->deployment folder, with links to the deployment models doc where appropriate. I would start with no link to the VM docs from the deployment models doc, but in the future maybe add a sentence to mention integration of external services into the mesh is another possibility with a link to the VM doc, if users aren't finding it themselves.

content/en/docs/ops/deployment/deployment-models/index.md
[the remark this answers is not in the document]
rcaballeromx: The problem is that your list is not complete. This is also not the place to present Istio features in general. Sometimes, less is more. As @frankbu and @smawson stated, this doc is not about showcasing why folks should add VMs to their deployments.

content/en/docs/ops/deployment/deployment-models/index.md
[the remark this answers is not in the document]
rcaballeromx: They ask themselves the questions and decide in their present @vadimeisenbergibm 
This is a question of style and I don't feel strongly either way. You can leave it in future tense but please avoid using the term service:
```suggestion
Will you add workloads running on virtual machines or bare metal systems to the mesh?
```

content/en/docs/ops/deployment/deployment-models/index.md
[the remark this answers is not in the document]
rcaballeromx: @vadimeisenbergibm That's then a redundant statement. What's missing here is what impact does adding VMs and bare metal systems have on how readers are deploying the mesh. Readers of this doc know whether or not they need to add VMs or bare metal systems to the mesh. What is your goal in stating vaguely that "sometimes" this is needed? What are those times? Under what circumstances? With what goals? The answers to those questions vary from reader to reader.  On this doc we are trying to focus on helping folks determine what deployment suits their needs, not on encouraging them to use any specific deployment model.

content/en/docs/ops/deployment/deployment-models/index.md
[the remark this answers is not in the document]
rcaballeromx: My edits go towards providing readers with the reasons they should consider a deployment model that includes VMs or bare metal systems. I would be ok if you want to change the introductory sentence of the list to be more explicit:
`Chose a deployment model that adds VMs or bare metal systems to the mesh if any of the following apply:`

content/en/docs/ops/deployment/deployment-models/index.md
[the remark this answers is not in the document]
frankbu: Re: services vs. workloads. This isn't just a case of some workloads implementing a service running on a VM. We are talking about integrating into the mesh a "service" that is fully deployed on a VM, so I think service is the right term to use here.

That said, as I said in my other comment, this whole thing should move to a separate doc.

content/en/docs/ops/deployment/deployment-models/index.md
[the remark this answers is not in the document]
vadimeisenbergibm: Note that we do not know if all the features of Istio work for VMs. So we cannot state that the users will benefit from all the Istio features when they integrate VMs with the mesh. However, we still want to provide users some motivation to integrate VMs with the mesh. So what should be stated here: "you will benefit from some of the features of Istio"? Or "you will benefit from most of the features"?

>  this doc is not about showcasing why folks should add VMs to their deployments

The reader could ask a reasonable question: what do I gain by integrating VMs with the mesh?

vadimeisenbergibm:
@frankbu Please note that the workloads that run on VMs are not external services, they are managed by the reader. To provide an example: suppose a reader has two clusters and two VMs and 50 workloads. The reader should decide where to deploy the workloads - which clusters, which VMs. The reader can decide to deploy some the workloads both on clusters and VMs. Then the reader should decide how to connect the clusters and the VMs network-wise. The reader should decide where to run control planes and which sidecars of which VM/cluster will listen to which control plane.

This is different from external services, e.g. googleapis, that the reader does not manage. The reader can control egress traffic to the external services, which is different from adding a workload on a VM to the mesh.

content/en/docs/ops/deployment/deployment-models/index.md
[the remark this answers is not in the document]
vadimeisenbergibm: I specify the reasons for running workloads on VMs below:
```
- Special hardware requirements
- Tighter isolation
- Tighter integration between the workload and the OS
- Legacy systems, workloads, or deployments
```

vadimeisenbergibm:
@rcaballeromx @frankbu @nmittler @smawson 

It seems this PR is more complicated than I initially thought. Let's agree on the high level structure first:

1. There are reasons you need to run some workloads on VMs (not everything runs on Kubernetes clusters). Some of the reasons are as follows...
1. You can integrate workloads on VMs with Istio mesh on Kubernetes clusters.
1. What you gain from such integration.
1. A short description of how this integration is achieved.

Feel free to point which parts of the above are redundant.

Second let's agree where such a document should be placed.

frankbu:
@vadimeisenbergibm Re: external services, yes, that's true, but there is also the case that the "reader" needs to run the service on a VM or bare metal for performance or other reasons. In that case there's also the question of whether to use service entries and just treat it like an external service, or integrate it into the mesh as a mesh-internal service. The reasons for going one way or the other in that scenario also needs to be covered in this document. In other words your point 3., above, should also explain what you gain vs. just calling the VM as an external service, and what's the cost.

vadimeisenbergibm:
@frankbu Agree with "In that case there's also the question of whether to use service entries and just treat it like an external service, or integrate it into the mesh as a mesh-internal service... explain what you gain vs. just calling the VM as an external service, and what's the cost.", but it is an advanced topic in my opinion.

It was covered partially in https://istio.io/blog/2018/egress-tcp/#relation-to-virtual-machines-support. The simpler way is to explain the VM integration first, then explain egress traffic control in some other document and then compare between the two in the document about egress traffic control. For the reader who does not know about Istio egress traffic control, talking about external services can be confusing. For you it is a natural question, since you know well both topics.

frankbu:
@vadimeisenbergibm The one thing about VM integration vs. external services that I don't think is such an advanced topic would be to point out that if the service is a leaf service (i.e., doesn't make outgoing requests back to the mesh), then it's probably easier to model it as an external service, instead of deploying envoy, nodeagent, etc. on the VM. So, I'm not sure we can really ignore the option of external services when talking about the advantages of VM integration.

vadimeisenbergibm:
Closing due to lack of community interest.

PR #6245, opened 2019-12-30
V1/2_POD_IP error when there are other pods labelled version=v1/2

SerenaFeng:
When there are multiple pods labelled version=v1, the V1_POD_IP is error, and so does V2_POD_IP. In my case for example, except httpbin, bookinfo example is also deployed in the cluster, which includes some pods labelled as version=v1, the result is like:

```
$ k g po
NAME                              READY   STATUS    RESTARTS   AGE
details-v1-78d78fbddf-fvgn5       2/2     Running   0          21h
httpbin-v1-6b4c749d6-cvk5r        2/2     Running   0          20h
httpbin-v2-5748c488bc-k7bwk       2/2     Running   0          20h
productpage-v1-596598f447-pmndz   2/2     Running   0          21h
ratings-v1-6c9dbf6b45-cchmb       2/2     Running   0          21h
reviews-v1-7bb8ffd9b6-2ljlx       2/2     Running   0          21h
reviews-v2-d7d75fff8-6p48p        2/2     Running   0          21h
reviews-v3-68964bc4c8-hxgsk       2/2     Running   0          21h
sleep-bb596f69d-wdlrj             2/2     Running   0          21h
$ export V1_POD_IP=$(kubectl get pod -l app=httpbin -l version=v1 -o jsonpath={.items..status.podIP})
$ echo $V1_POD_IP
10.244.50.74 10.244.193.205 10.244.50.78 10.244.50.75 10.244.193.204
```

After fixing it as the patch method, the result would be

```
$ export V1_POD_IP=$(kubectl get pod -l app=httpbin,version=v1 -o jsonpath={.items..status.podIP})
$ echo $V1_POD_IP
10.244.193.205
```

PR #6259, opened 2019-12-31
certificate decode error when decoding from bash pipe by openssl

SerenaFeng:
Current implementation can not decode the certificate but generate the following error:

```
$ kubectl get secret dns.istio-galley-service-account -n istio-system -o json | jq -r '.data["cert-chain.pem"]' | base64 --decode | openssl x509 -in - -text -noout
Error opening Certificate -
140406230840984:error:02001002:system library:fopen:No such file or directory:bss_file.c:398:fopen('-','r')
140406230840984:error:20074002:BIO routines:FILE_CTRL:system lib:bss_file.c:400:
unable to load certificate
```

### DIFF
--- a/content/en/docs/tasks/security/dns-cert/index.md
+++ b/content/en/docs/tasks/security/dns-cert/index.md
@@ -72,7 +72,7 @@ and that the certificate contains the configured DNS names, you need to get the 
 decode it, and view its text output with the following command:
 
 {{< text bash >}}
-$ kubectl get secret dns.istio-galley-service-account -n istio-system -o json | jq -r '.data["cert-chain.pem"]' | base64 --decode | openssl x509 -in - -text -noout
+$ kubectl get secret dns.istio-galley-service-account -n istio-system -o json | jq -r '.data["cert-chain.pem"]' | base64 --decode | openssl x509 -in /dev/stdin -text -noout
 {{< /text >}}
 
 The text output should include:
@@ -98,7 +98,7 @@ contains the configured DNS names, you need to get the secret from Kubernetes, p
 and view its text output with the following command:
 
     {{< text bash >}}
-    $ kubectl get secret dns.istio-galley-service-account -n istio-system -o json | jq -r '.data["cert-chain.pem"]' | base64 --decode | openssl x509 -in - -text -noout
+    $ kubectl get secret dns.istio-galley-service-account -n istio-system -o json | jq -r '.data["cert-chain.pem"]' | base64 --decode | openssl x509 -in /dev/stdin -text -noout
     {{< /text >}}
 
 The output should include:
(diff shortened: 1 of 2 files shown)


content/en/docs/tasks/security/dns-cert/index.md
```diff
@@ -72,7 +72,7 @@ and that the certificate contains the configured DNS names, you need to get the
 decode it, and view its text output with the following command:
 
 {{< text bash >}}
-$ kubectl get secret dns.istio-galley-service-account -n istio-system -o json | jq -r '.data["cert-chain.pem"]' | base64 --decode | openssl x509 -in - -text -noout
+$ kubectl get secret dns.istio-galley-service-account -n istio-system -o json | jq -r '.data["cert-chain.pem"]' | base64 --decode | openssl x509 -in /dev/stdin -text -noout
```
frankbu: Isn't this same change in your other PR? #6245

SerenaFeng: This change is mistakenly mingled with #6245 , I have submitted a new pull request (#6282 ) to substitute #6245 .

frankbu: @SerenaFeng I'm still not sure I understand why you can't use `-` for the `-in` argument, but others can. What platform are you using? Are you using a different version of `openssl` than others? If so, are you sure your fix works on all version?

frankbu: @lei-tang can you confirm that `-in -` works for you? What platform and version of openssl are you using?

lei-tang: "-in -" works for me. In my environment, the OS is debian and the openssl version is: OpenSSL 1.1.1d  10 Sep 2019.

SerenaFeng: In my environment, `-in /dev/stdin` works on`OpenSSL 1.0.2g  1 Mar 2016` ubuntu xenial and `OpenSSL 0.9.8zh 14 Jan 2016` macOS Sierra v10.12.6, while `-in -` not. 

For `OpenSSL 1.1.1d  10 Sep 2019` ubuntu xenial situation, both `-in /dev/stdin` and `-in -` work correctly.

frankbu: So it sounds like OpenSSL 1.1 supports `-in -`, but older versions do not. We could fix this either by changing to `-in /dev/stdin` or by specifying that OpenSLL >= 1.1 is needed for this task.

@lei-tang You added this task, so WDYT?

lei-tang: As "-in /dev/stdin" works for both older and newer versions of openssl, I think we can changing the command to use "-in /dev/stdin".

PR #6278, opened 2020-01-02
fix docs for protocol sniffing

yxue:


yxue:
> Is this same change needed in the master branch?

No. For master, it's different. Protocol sniffing in on by default on master.

ericvn:
@yxue It does look like the `master` branch doc is different than you state above as it has `This feature is experimental and off by default`. I think there needs to be a change in `master` as well (but it is different than this one as 1.4 is only partially enabled by default).

yxue:
I meant the change in the master branch is different from the change in this PR. The doc also needs to be fixed in the master.

ericvn:
#6172 seems to be a change for `master` whose text is similar to this PR (only enabled for outbound).

yxue:
> @yxue I assume you will send a separate PR for the changes needed in master.

I am going to comment on this [PR](https://github.com/istio/istio.io/pull/6172) instead of sending a new one.

PR #6310, opened 2020-01-08
pt-br-translation: Home Page

jhonmike:
PR with translated to home page to portuguese brazil 🤘 

local:
image

the diff was confusing because the version in master got the wrong indentation because of my code editor.

but I made the adjustment according to the English version, and added the correct settings in my editor to not happen again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .pydevproject
 *.iml
 .idea/
+.vscode
 tmp/
 *.bak
 .DS_Store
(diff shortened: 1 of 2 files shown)


jhonmike:
the diff was confusing because the version in master got the wrong indentation because of my code editor.

jhonmike:
I don't think the translation worked, is it because of the pt-br that it should be just pt?

https://preliminary.istio.io/pt-br

jhonmike:
Done 🤘 https://deploy-preview-6310--preliminary-istio.netlify.com/pt-br

PR #6354, opened 2020-01-21
Document suppression arguments for istioctl analyze

selmanj:
Describes suppressing messages via CLI or annotations in more detail.

### DIFF
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -213,6 +213,7 @@ $ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST
 {{< /text >}}
 
 To ignore multiple codes for a resource, separate each code with a comma:
+
 {{< text plain >}}
 $ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002
 {{< /text >}}
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -183,7 +183,7 @@ $ istioctl manifest apply --set values.galley.enableAnalysis=true
 
 Sometimes you might find it useful to hide or ignore analyzer messages in certain cases. For example, imagine a situation where a message is emitted about a resource you don't have permissions to update:
 
-{{< text bash >}}
+{{< text plain >}}
 $ istioctl analyze -k --all-namespaces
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -178,3 +178,41 @@ You can enable this feature with:
...
+{{< /text >}}
+
+Because you don't have permissions to update the namespace, you cannot resolve the message by annotating the namespace. Instead, you can direct `istioctl analyze` to suppress the above message on the resource:
+
+{{< text bash >}}
+$ istioctl analyze -k --all-namespaces --suppress "IST0102=Namespace frod"
+✔ No validation issues found.
+{{< /text >}}
+
+The syntax used for suppression is the same syntax used throughout `istioctl` when referring to resources: '\<kind> \<namespace>.\<name>', or just '\<kind> \<name>' for cluster-scoped resources like `Namespace`. If you want to suppress multiple objects, you can either repeat the `--suppress` argument or use wildcards:
```
sushicw: Isn't in `<name>.<namespace>`, not `<namespace>.<name>`?

Also, can your syntax examples use backticks, avoiding the need to escape `<`?

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -178,3 +178,41 @@ You can enable this feature with:
...
+### Ignoring specific analyzer messages via CLI
+
+Sometimes you might find it useful to hide or ignore analyzer messages in certain cases. For example, imagine a situation where a message is emitted about a resource you don't have permissions to update:
+
+{{< text bash >}}
+$ istioctl analyze -k --all-namespaces
+Warn [IST0102] (Namespace frod) The namespace is not enabled for Istio injection. Run 'kubectl label namespace frod istio-injection=enabled' to enable it, or 'kubectl label namespace frod istio-injection=disabled' to explicitly mark it as not needing injection
+Error: Analyzers found issues.
+See https://istio.io/docs/reference/config/analysis for more information about causes and resolutions.
+{{< /text >}}
```
sushicw: Maybe this should be plain rather than bash. Only the first line is syntactically bash, we don't want weird highlights showing up in the rest. (Maybe none do anyway?)

Same for the other usages below.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
selmanj: Yes, it is. 🤦‍♂  Fixed.

I can use backticks - originally that's what I used but I was worried it was confusing to describe syntax with that.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
selmanj: 👍 I marked them all as plain since there's no real bash going on here - just arguments.

PR #6356, opened 2020-01-24
Improve the description of the installation option global.enableTracing

ElliotPM:
The installation option global.enableTracing is poorly described with global.enableTracing and tracing.enable easily confused.

I personally do not think that global.enableTracing should be defaulted to true. This caused issues with our tracing generation and it was not expected that Istio would overwrite tracing headers by default.

### DIFF
--- a/content/en/docs/reference/config/installation-options/index.md
+++ b/content/en/docs/reference/config/installation-options/index.md
@@ -251,7 +251,7 @@ To get the exact set of supported options, please see the [Helm charts]({{< gith
 | `global.controlPlaneSecurityEnabled` | `false` | `controlPlaneSecurityEnabled enabled. Will result in delays starting the pods while secrets are propagated, not recommended for tests.` |
 | `global.disablePolicyChecks` | `true` | `disablePolicyChecks disables mixer policy checks. if mixer.policy.enabled==true then disablePolicyChecks has affect. Will set the value with same name in istio config map - pilot needs to be restarted to take effect.` |
 | `global.policyCheckFailOpen` | `false` | `policyCheckFailOpen allows traffic in cases when the mixer policy service cannot be reached. Default is false which means the traffic is denied when the client is unable to connect to Mixer.` |
-| `global.enableTracing` | `true` | `EnableTracing sets the value with same name in istio config map, requires pilot restart to take effect.` |
+| `global.enableTracing` | `true` | `enableTracing enables the generation of trace spans and request IDs. Requires a trace span collector defined in the proxy configuration. enableTracing sets the value with same name in the Istio config map, requires pilot restart to take effect.` |
 | `global.tracer.lightstep.address` | `""` | `example: lightstep-satellite:443` |
 | `global.tracer.lightstep.accessToken` | `""` | `example: abcdefg1234567` |
 | `global.tracer.lightstep.secure` | `true` | `example: true\|false` |


ElliotPM:
@googlebot I signed it! and we'll verify it.

content/en/docs/reference/config/installation-options/index.md
```diff
@@ -251,7 +251,7 @@ To get the exact set of supported options, please see the [Helm charts]({{< gith
 | `global.controlPlaneSecurityEnabled` | `false` | `controlPlaneSecurityEnabled enabled. Will result in delays starting the pods while secrets are propagated, not recommended for tests.` |
 | `global.disablePolicyChecks` | `true` | `disablePolicyChecks disables mixer policy checks. if mixer.policy.enabled==true then disablePolicyChecks has affect. Will set the value with same name in istio config map - pilot needs to be restarted to take effect.` |
 | `global.policyCheckFailOpen` | `false` | `policyCheckFailOpen allows traffic in cases when the mixer policy service cannot be reached. Default is false which means the traffic is denied when the client is unable to connect to Mixer.` |
-| `global.enableTracing` | `true` | `EnableTracing sets the value with same name in istio config map, requires pilot restart to take effect.` |
+| `global.enableTracing` | `true` | `enableTracing enables the generation of trace spans and request IDs. Requires a trace span collector defined in the proxy configuration. enableTracing sets the value with same name in the Istio config map, requires pilot restart to take effect.` |
```
ericvn: The enableTracing description was gotten from the comment in: https://github.com/istio/istio/blob/master/manifests/global.yaml#L252. I believe this was probably gotten from the comment in the Istio values.yaml file: https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/values.yaml#L329.

I'm not sure if the `global.yaml` and `values.yaml` file should also be updated with another PR to reflect this desired comment.

I'll let another maintainer, who is more familiar with the mapping between these files, do the approval for this PR.

ericvn: @sdake @howardjohn Do either of you have an idea if one or both of the Istio files should also be updated with the same text?

ericvn: Looking closer, I do see this change is within the generated portion of this file. They are after the:
```
<!-- AUTO-GENERATED-START -->
```

Today the python generation script looks at the "install/kubernetes/helm/istio" directory so this change would have to go in there. I did create #6532 where at some point, the location will need to change (as that directory no longer exists in istio/istio _master).

frankbu: I don't think we generate this file anymore, since the helm install is deprecated.

@ostromart what the latest status of this option? Is there an IstioOperator replacement?

ostromart: yes, it's still there: https://github.com/istio/istio/blob/7f633729c5c31bdeb92e91ec567c3d25b98b6ddc/manifests/global.yaml#L255
not a very good description of the parameter. 
We also have a schematized version: https://github.com/istio/istio/blob/7f633729c5c31bdeb92e91ec567c3d25b98b6ddc/operator/pkg/apis/istio/v1alpha1/values_types.proto#L460
The latter file is planned to become the source of truth for the API going forward. We should consider autogenerating docs from this proto file rather than scraping the values files comments...

adammil2000:
This PR attempts to modify the output of auto-generated docs. Please make the necessary change in description that resides in the source code instead.

PR #6363, opened 2020-01-24
Define and link SDS on first mention

adammil2000:
Solve the problem where the user gets confused because we talk about SDS all of a sudden as if we've already explained it to them.


ericvn:
Not a very exhaustive search (just the other platform setup pages), but I think the `docs/setup/platform-setup/minikube/` page could use the same change.

ericvn:
@adammil2000 Did you want to pick up the change in the minikube file I mentioned above as well?

adammil2000:
Fixed, thanks!

ericvn:
@adammil2000 I believe I got the spife changes in #6379 so if you remove the last commit from this PR, it should remove the additional approvers and this PR might merge?

DonaldSpencer:
Testing 123

PR #6368, opened 2020-01-26
update documentation for TCP traffic shifting: use a dedicated namesp…

su225:
Use dedicated namespace `istio-io-tcp-traffic-shifting` instead of `default` for running the test.
This is part of the changes - https://github.com/istio/istio/issues/18285

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[X] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
+++ b/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
@@ -58,13 +58,13 @@ weighted routing feature.
         $ kubectl apply -f @samples/tcp-echo/tcp-echo-services.yaml@ -n istio-io-tcp-traffic-shifting
         {{< /text >}}
 
-2.  Next, route all TCP traffic to the `v1` version of the `tcp-echo` microservice.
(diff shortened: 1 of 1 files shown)


ericvn:
@su225 Note that the lint test failed and requires some changes.

su225:
@ericvn - not sure why linting failed this time. I see that Html-proofer was fine. I observed similar behavior when I ran `make lint` on my machine.

ericvn:
@su225 Looks like 
```
./en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md:61: MD029 Ordered list item prefix
./en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md:67: MD029 Ordered list item prefix
```

You shouldn't need to update the number, it'll be figured out automatically?  You can look at the netlify to make sure of the numbering if you want.

su225:
@ericvn - Whoops! Thanks. I scrolled down and I could see that. Fixed it.
https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md029---ordered-list-item-prefix

PR #6377, opened 2020-01-30
Updated cert-manager version

blhagadorn:
As of v.10 of cert-manager the `apiVersion` has changed to `cert-manager.io/v1alpha2`

Update version of cert-manager to more recent version as of Oct of last year. If not this, then we should specify that the user must install an older version of cert-manager.

[ x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


blhagadorn:
@ericvn Did we fix the url issues that was causing this to be blocked from last friday?

ericvn:
Yes @blhagadorn we did fix the issues and the _lint_ job is passing.

ericvn:
Also from the referenced issue: https://cert-manager.io/docs/installation/upgrading/upgrading-0.10-0.11/#additional-annotation-changes

PR #6379, opened 2020-01-30
Fix URLs to fix lint link errors

ericvn:
Update links that are failing lint. 
 - LightSpeed: I'm not sure that they are exact replacements since I don't know what was at the old links.
 - Spiffe


ericvn:
LightStep may have forwarding in place now. Testing in other PRs. DNM for now.

PR #6386, opened 2020-02-01
toss in action

JacksChoco:


JacksChoco:
@ericvn Can you check this?

ericvn:
@JacksChoco I did approve it, but it also needs approval from several others because the PR has changes in the other languages as well.

JacksChoco:
Thanks for answering!

ericvn:
One bit of advice to simply getting your work on would be to remove the non-English language content. Our typical workflow is to have English language content submitted and then the various translation teams will translate to PT or CN. This would reduce the number of approvals needed to just the English team for this PR.

JacksChoco:
I switched with your advice.

PR #6376, opened 2020-01-30
adding layer5 adapter to the repos

girishranganathan:
This PR adds Layer5 adapter to the list of adapters.

The areas that this PR affects.

[ ] Configuration Infrastructure
[ X] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #6389, opened 2020-02-03
finish off adding layer5 adapter

girishranganathan:
This PR adds Layer5 adapter to the list of adapters.

The areas that this PR affects.


girishranganathan:
FYI: this PR ties off the last piece of adding the html. It is part of https://github.com/istio/istio.io/pull/6376.

PR #6398, opened 2020-02-05
Add option for Chinese

gorda:
- [x] Docs

Hi, The translation of Istio Chinese by ServiceMesher is almost complete. Now, we wanna to make Chinese as an optional, I released the annotates in the header (i.e. this pr changed content) according to the [documentation](https://github.com/istio/istio.io#multi-language-support). 

In addition, I have two questions:

1. After modifying the `layouts/index.redir` file, the automatic redirection is not in my native language as described by [document](https://github.com/istio/istio.io#multi-language-support). This is caused by the lag of the document, or because the construction process of [preliminary](https://preliminary.istio.io/) and [release 1.4](https://istio.io/) are different?

image

2. I noticed that [preliminary](https://preliminary.istio.io/) now has a new language switch button, similar to this:

image

But I think its user experience is not very good. It is not possible to directly select a language. You can only switch according to the list. You know, like polling, and I removed the annotates of the `layouts/partials/headers.html` file according to the [document](https://github.com/istio/istio.io#multi-language-support). It can select a language directly in the header like [Istio 1.2](https://archive.istio.io/v1.2/).

image

image

I'm not sure my approach is correct, please help me to confirm it, thank you very much!

/cc @geeknoid  @rcaballeromx  @rootsongjc

### DIFF
--- a/layouts/index.redir
+++ b/layouts/index.redir
@@ -19,7 +19,7 @@
 https://istio.netlify.com/* https://istio.io/:splat 301!
 
 # Redirect to translated sites
-# /  /zh   302  Language=zh
+/  /zh   302  Language=zh
 
 # Redirect for the helm charts
 /charts/ https://storage.googleapis.com/istio-release/releases/{{ .Site.Data.args.full_version }}/charts/ 301
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -76,12 +76,10 @@
                 </button>
 
                 <div id="gearDropdownContent" class="menu-content" aria-labelledby="gearDropdownButton" role="menu">
-                    <!--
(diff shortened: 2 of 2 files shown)


geeknoid:
This all looks good.

Yes, the language selection button only lets you iterate through languages. This button is only on preliminary.istio.io, not on the full istio.io, so I thought it was good enough that way. The menu is there if you want to switch directly.

As for why the redirect doesn't work, if I recall correctly the redirect only takes effect if you haven't explicitly selected a language in the menu. If you have selected a language in the menu, that'll take precedence. So to test this is working, you could try either an Incognito browser window, or you can clear the web site local storage using the browser's developer tools.

PR #6436, opened 2020-02-10
[release 1.4] Manual cherry pick of #6422 - Fix CHANGLOG-* locations being updated in Kubernetes.

ericvn:


ericvn:
I'm not seeing the links failing the lint checker so maybe not as urgent as I suspected. But the existing links won't go to the correct location so still makes sense to update them.

PR #6454, opened 2020-02-12
Update index.md

geeknoid:
Redundant text, the score is already rendered by the infra in the detail panel.


ericvn:
We should cherry pick this to 1.4, but I notice that this page isn't on 1.4. Maybe we missed an earlier cherry pick.

PR #6458, opened 2020-02-12
WIP Remove mentions of helm installation

howardjohn:
For https://github.com/istio/istio/issues/20441

This moves all mentions of helm installation to istioctl

The helm charts will be removed in 1.6. This is not meant to be backported to 1.5. I am realizing now that I send this PR master will become 1.5, and there is no way to work on 1.6 until then? So I will mark as WIP


ericvn:
So with the exception of the removal of the `setup/install/helm/index.md` and `setup/upgrade/index.md` files (which from the meetings last week will be rewritten for 1.6), the changes from this PR should be in.

PR #6459, opened 2020-02-12
Remove deprecated platform instructions

howardjohn:
For https://github.com/istio/istio/issues/20441

These are marked as "will be deleted in the next release"


ericvn:
Similar to #6468

PR #6467, opened 2020-02-13
remove references to experimental simplified multicluster setup scripts

ayj:
The simplified multicluster scripts was an experiment that has run its course. We may adopt some of the ideas from this approach, but the implementation would be different. 

fixes https://github.com/istio/istio/issues/21054


ayj:
cc @nmittler

ayj:
```
- ./public/pt-br/news/releases/0.x/announcing-0.1/index.html
  *  External link https://envoyproxy.github.io/envoy/ failed: got a time out (response code 301)
```

/retest

PR #6470, opened 2020-02-14
remove the mixer policy concept page

yangminzhu:
This PR removes the mixer policy from concept page to avoid confusions, also update the task page for the deprecate and recommend users to use authorization policy.


ericvn:
Do we want to remove the lines near the bottom of the concepts page:

```
While we strongly recommend using the Istio authorization mechanisms, Istio is flexible enough to allow you to plug in your own authentication and authorization mechanisms via the Mixer component. To use and configure plugins in Mixer, visit our policies and telemetry adapters docs.
```

yangminzhu:
> Do we want to remove the lines near the bottom of the concepts page:
> 
> ```
> While we strongly recommend using the Istio authorization mechanisms, Istio is flexible enough to allow you to plug in your own authentication and authorization mechanisms via the Mixer component. To use and configure plugins in Mixer, visit our policies and telemetry adapters docs.
> ```

Good catch, I think we should also remove it there, thanks.


@mandarjog could you take a look? thanks.

mandarjog:
@douglas-reid just confirming that this is consistent with your changes to docs.

mandarjog:
@gargnupur

gargnupur:
@mandarjog , @douglas-reid , @yangminzhu :
Should we add some comment on top of "ratelimiting" and "control headers and routing" tasks too? Currently if you land on them, they ask to navigate to "Enabling Policy Enforcement" page and that says policy is disabled and use Authorization policy, which doesn't apply to these tasks..

yangminzhu:
> @mandarjog , @douglas-reid , @yangminzhu :
> Should we add some comment on top of "ratelimiting" and "control headers and routing" tasks too? Currently if you land on them, they ask to navigate to "Enabling Policy Enforcement" page and that says policy is disabled and use Authorization policy, which doesn't apply to these tasks..

@gargnupur Just to confirm, the "ratelimiting" and "control headers and routing" are also deprecated as part of the mixer v1 deprecation, is that correct? If so, I can update the message a little bit to make it clear that all these are deprecated and redirect user to authorization policy for access-control related tasks.

gargnupur:
> > @mandarjog , @douglas-reid , @yangminzhu :
> > Should we add some comment on top of "ratelimiting" and "control headers and routing" tasks too? Currently if you land on them, they ask to navigate to "Enabling Policy Enforcement" page and that says policy is disabled and use Authorization policy, which doesn't apply to these tasks..
> 
> @gargnupur Just to confirm, the "ratelimiting" and "control headers and routing" are also deprecated as part of the mixer v1 deprecation, is that correct? If so, I can update the message a little bit to make it clear that all these are deprecated and redirect user to authorization policy for access-control related tasks.

I would let @mandarjog or @douglas-reid comment on that...

mandarjog:
"ratelimiting" and "control headers and routing"  need alternatives that are *not* `Authorization policy` 

We should say.
```
Ratelimiting:
Consider using Envoy native rate limiting instead of mixer rate limiting. 
Istio will add support for native ratelmiting API thru the Istio extensions API.

Control headers and routing:
Consider using Envoy Ext-Auth, lua filter, or write a filter using the Envoy-wasm sandbox https://github.com/envoyproxy/envoy-wasm/tree/master/test/extensions/filters/http/wasm/test_data  filters
```

yangminzhu:
> "ratelimiting" and "control headers and routing" need alternatives that are _not_ `Authorization policy`
> 
> We should say.
> 
> ```
> Ratelimiting:
> Consider using Envoy native rate limiting instead of mixer rate limiting. 
> Istio will add support for native ratelmiting API thru the Istio extensions API.
> 
> Control headers and routing:
> Consider using Envoy Ext-Auth, lua filter, or write a filter using the Envoy-wasm sandbox https://github.com/envoyproxy/envoy-wasm/tree/master/test/extensions/filters/http/wasm/test_data  filters
> ```

yeah, sorry for the confusion, I was trying to only recommend the authorization policy for access-control use case.

yangminzhu:
@mandarjog @gargnupur Updated, PTAL.

@istio/wg-docs-maintainers-english Could you review the PR? Thank you.

douglas-reid:
This doesn't quite align with the other PR.  In the other PR, we moved Mixer-specific sections into a `Using Mixer for Telemetry` section and did not mark the Tasks as "Deprecated".

The Policy section here is bit tricky, as there is no non-Mixer replacement.

Maybe we need an entire new section organized as follows: `Tasks > Using Mixer (Deprecated) > { Policy | Metrics | Logs }`.  Then we could remove `Policy` as a top-level Task category (unless we have pending replacements).

Thoughts?

frankbu:
@douglas-reid I like the way you restructured the observability tasks in #6533, i.e., leave the top level Observability folder unchanged, but put the deprecated mixer tasks into a "Using Mixer" subfolder). The only thing I would suggest is to explicitly label the Mixer section as deprecated (see my comment in #6533).

I would suggest using the same approach for the Policy tasks, that is, leave the top level Policy section unchanged, but move all the Mixer-based tasks under a subfolder "Using Mixer for Policy (deprecated)", even if there's currently nothing else under the Policy folder. Unless I'm missing something, I'm assuming that we will eventually be adding some other non-mixer-based policy tasks.

content/en/docs/concepts/what-is-istio/index.md
```diff
@@ -74,18 +74,6 @@ secure {{<gloss>}}pod{{</gloss>}}-to-pod or service-to-service communication at
 
 Refer to the [Security concepts guide](/docs/concepts/security/) for more details.
 
-### Policies
```
frankbu: Shoudn't we leave a high level Policies section on this page? It makes sense to remove the policy adapters reference and the link to more details, but even though Mixer is deprecated, there are other ways to configure policies and enforce rules in Istio, so it seems to me that we should at least leave some high level overview in the what-is-istio page about policies.

douglas-reid:
> @douglas-reid I like the way you restructured the observability tasks in #6533, i.e., leave the top level Observability folder unchanged, but put the deprecated mixer tasks into a "Using Mixer" subfolder). The only thing I would suggest is to explicitly label the Mixer section as deprecated (see my comment in #6533).
> 
> I would suggest using the same approach for the Policy tasks, that is, leave the top level Policy section unchanged, but move all the Mixer-based tasks under a subfolder "Using Mixer for Policy (deprecated)", even if there's currently nothing else under the Policy folder. Unless I'm missing something, I'm assuming that we will eventually be adding some other non-mixer-based policy tasks.

@frankbu I'm of the "radical" opinion that we might be better off just removing "Policies" (as a top-level concept) from the docs altogether. So, we have three pillars, if you will: `Traffic Mgmt`, `Security`, `Observability`. Any mention of rate-limits, deny-lists, etc., can be included in an `Authorization` subsection of `Security` (and for now, not really mentioned except in the deprecated section).

Thoughts?

frankbu:
@douglas-reid I suppose removing "Policies" makes sense if it's really going to be an empty folder eventually. If all the deprecated tasks are going to be subsumed by tasks in the Security section, maybe we should change the heading for it to "Security and Policies"? (setting rate-limits, for example, doesn't seem like a "Security" task). To some degree, the right approach depends on exactly what we say about Policies on the what-is-istio concepts page. I don't think completely removing all mention of policies is such a good idea (see my other comment, above), since it was hyped as one of the main features of Istio in the early days, so just silently removing it might seem like a regression and lead to some confusion.

yangminzhu:
> @douglas-reid I suppose removing "Policies" makes sense if it's really going to be an empty folder eventually. If all the deprecated tasks are going to be subsumed by tasks in the Security section, maybe we should change the heading for it to "Security and Policies"? (setting rate-limits, for example, doesn't seem like a "Security" task). To some degree, the right approach depends on exactly what we say about Policies on the what-is-istio concepts page. I don't think completely removing all mention of policies is such a good idea (see my other comment, above), since it was hyped as one of the main features of Istio in the early days, so just silently removing it might seem like a regression and lead to some confusion.

@frankbu Some policy tasks (rate limiting and control header in response) is not supported in security. I feel it's nice and cleaner to actually remove the "Policy" page in the Task. We just need to create a separate section in the upgrade notice to point out the alternative options as mentioned by @mandarjog (rate limiting -> use Envoy rate limiting, control header -> custom filter, lua, etc. and access control -> authorization policy)

content/en/docs/concepts/what-is-istio/index.md
yangminzhu: I think we should just remove it from the concept page if it's deprecated, the concept page should always be consistent with the current istio version. Users should refer to the older docs for more information about the mixer v1.

And I think we can add to the upgrade note to make it clear that the policy (which is equivalent to mixer v1) is deprecated and the alternative options as mentioned by @mandarjog

frankbu: I think the problem I see stems from the fact that you guys are equating the 4th piller with Mixer, which is true at the implementation level, but at the concepts (what is istio) level, Mixer is an implementation detail. If we completely remove Policies from the overview, and go with the three pillar doc, we are saying that Istio only provides 3 kinds of functionality. Users can:

1. Manage the traffic
2. Secure their services
3. Observe their services

This list is leaving out some functionality, like setting rate limits, that Istio used to provide. Since it can still be done (just not using Mixer now), I think we should still keep the 4th pillar, or as I suggested below, add Policies to the Security pillar. I don't think that completely removing any mention of policies in the overview doc and expecting people to find out about policies by digging into the details of deprecated docs and upgrade notes is a very nice UX.

Anyway, that's just my opinion. Maybe others feel a functionality regression at the high level overview isn't a problem.

douglas-reid:
@frankbu I agree that removing "Policies" as a separate feature _feels_ like a regression. But, I'm wondering if (a) it reflects the reality of the project and (b) eliminates some user confusion. I know that we often get questions about `Authentication Policy` (and related) on the #policies-telemetry channel, when those are better served in #security, for instance.

The reason we've equated Policy with Mixer is that Mixer was/is the only _Istio-specific_ way to configure and enforce policies. Using Envoy's rate-limiting, etc., feels like something that is maybe best documented as an Envoy feature, if Istio isn't providing any value-add (to me, at least). I can see, however, the argument that adopting Istio gives you an easy way to get Envoy in place to use for policy enforcement.

If we think we have some non-deprecated content for Policy Enforcement, and a clear way to distinguish it security concepts like Authentication, then it makes sense to me to keep it. I just don't know that that is the case right now. Maybe in the next few weeks, we'll have some content grow in to that space?

I think we need others to weigh-in here, as I may be well in the minority opinion.

mandarjog:
I think it is ok to remove policy and have it subsumed by authorization policy. That is what policy meant many times except for rate limiting and policy based header manipuation and routing.
It makes the current doc cleaner.

We can have those 2 small blurbs about rate limiting and control headers.

yangminzhu:
> I think it is ok to remove policy and have it subsumed by authorization policy. That is what policy meant many times except for rate limiting and policy based header manipuation and routing.
> It makes the current doc cleaner.
> 
> We can have those 2 small blurbs about rate limiting and control headers.

I'm a little bit lost for the changes needed for this PR.

@mandarjog Do you mean to remove the whole `Policy` task page or just remove the `Denials and White/Black Listing`? If we still keep the `Policy` page, as @douglas-reid mentioned it doesn't align with the other PR.

@frankbu If it's a big concern of removing the `Policy` page that could create an impression of functionality regression in Istio, I can send out a follow-up PR to make this change explicit and clear in the release note, that is to say, we're explicitly deprecating this feature but with some alternatives and more to come with the new wasm framework in the future. I think removing the `policy` reflects the reality of the current state of Istio in 1.5 and it actually helps to reduce a lot of confusions from users.

kyessenov:
Do you think we can give examples of translating mixer policies to authorization policy? Maybe IP whitelists to start with? Then we can place a banner on top of mixer pages

yangminzhu:
> Do you think we can give examples of translating mixer policies to authorization policy? Maybe IP whitelists to start with? Then we can place a banner on top of mixer pages

@kyessenov I prefer to add a new task under authorization for enforcing access control on ingress gateway where we can use the source IP in both allow policy (whitelist) and deny policy (blacklist).

kyessenov:
@yangminzhu I was just thinking something simpler: for each basic task in Mixer, put a banner saying this is how you do it in authorization policy. It's easier to learn by example how to drop Mixer policies.

content/en/docs/tasks/policy-enforcement/control-headers/index.md
```diff
@@ -1,10 +1,17 @@
...
 description: Shows how to modify request headers and routing using policy adapters.
 weight: 20
 keywords: [policies,routing]
 ---
 
+{{< warning >}}
+The mixer policy is deprecated in Istio 1.5 and not recommended for production usage.
+
+Consider using Envoy `Ext-Auth`, `lua` filter, or write a filter using the
+[`Envoy-wasm` sandbox](https://github.com/envoyproxy/envoy-wasm/tree/master/test/extensions/filters/http/wasm/test_data).
```
kyessenov: I would advise against using direct links to github master. They almost certainly will break at some point in the future.

content/en/docs/tasks/policy-enforcement/control-headers/index.md
```diff
@@ -1,10 +1,17 @@
...
+title: Control Headers and Routing (Deprecated)
 description: Shows how to modify request headers and routing using policy adapters.
 weight: 20
 keywords: [policies,routing]
 ---
 
+{{< warning >}}
+The mixer policy is deprecated in Istio 1.5 and not recommended for production usage.
+
+Consider using Envoy `Ext-Auth`, `lua` filter, or write a filter using the
```
kyessenov: Link to https://www.envoyproxy.io/docs/envoy/v1.13.0/intro/arch_overview/security/ext_authz_filter

content/en/docs/tasks/policy-enforcement/enabling-policy/index.md
```diff
@@ -1,10 +1,23 @@
...
+title: Enabling Policy Enforcement (Deprecated)
 description: This task shows you how to enable Istio policy enforcement.
 weight: 1
 keywords: [policies]
 ---
 
+{{< warning >}}
+The mixer policy is deprecated in Istio 1.5 and not recommended for production usage.
+
+* Rate limiting: Consider using Envoy native rate limiting instead of mixer rate limiting.
```
kyessenov: Link to https://www.envoyproxy.io/docs/envoy/v1.13.0/intro/arch_overview/other_features/global_rate_limiting

content/en/docs/concepts/what-is-istio/index.md
frankbu: If we remove the Policies section, how about changing the Security section to include the bullets about policy. Change the `### Security` section to this:
 ```
### Security and Policies

Istio’s security capabilities free developers to focus on security at the application level. Istio provides the underlying secure communication channel, and
manages authentication, authorization, and encryption of service communication at scale. With Istio, service communications are secured by default,
letting you enforce policies consistently across diverse protocols and runtimes -- all with little or no application changes.

While Istio is platform independent, using it with Kubernetes (or infrastructure) network policies, the benefits are even greater, including the ability to
secure {{<gloss>}}pod{{</gloss>}}-to-pod or service-to-service communication at the network and application layers.

Istio also lets you configure custom policies for your application to enforce rules at runtime such as:

* Rate limiting to dynamically limit the traffic to a service
* Denials, whitelists, and blacklists, to restrict access to services
* Header rewrites and redirects

Refer to the [Security concepts guide](/docs/concepts/security/) for more details.
```

content/en/docs/tasks/policy-enforcement/control-headers/index.md
yangminzhu: I copied the link from @mandarjog, there is no release-1.5 branch in that repo, do you have any alternative we can use here?

kyessenov: @jplevyak @PiotrSikora is there any public doc about Wasm in Envoy except some random github links?

content/en/docs/concepts/what-is-istio/index.md
yangminzhu: > I think the problem I see stems from the fact that you guys are equating the 4th piller with Mixer, which is true at the implementation level, but at the concepts (what is istio) level, Mixer is an implementation detail.

Without Mixer, I don't feel we even need the `policy` concept at all, the `policy` concept exists because we have some dedicated Mixer API to use it. With the deprecation of Mixer, the Mixer API is gone and we do not want anyone to use it anymore.

> This list is leaving out some functionality, like setting rate limits, that Istio used to provide. Since it can still be done (just not using Mixer now), I think we should still keep the 4th pillar,

I understand it is a functionality regression but this is what happening anyway in Istio 1.5, and the doc should reflect this change clearly.

On the other hand, yes, the feature (rate limiting, control header, etc.) could be done but that is out of the scope of `policy` concept. You have to use the `EnvoyFilter` to configure Envoy directly which is quite hard and breaking class change. We do not have any documentation on this at all.

Even if we have the documentation, it should really go to the `Manage the traffic`  section because it has nothing to do with the `Policy` concept as it's all about the configuring the Envoy API directly.


@louiscryan @smawson @linsun @rshriram, I'm wondering what does TOC think here? Should we remove the `policy` from the concept or should we keep it but merge it with other sections? Thanks.

frankbu: > Even if we have the documentation, it should really go to the Manage the traffic section

@yangminzhu That's a good point. Maybe the right way to document it is to remove the `### Policies` section, as you've done, but then add these 2 features to the `### Traffic Management` section:
* Rate limiting to dynamically limit the traffic to a service
* Header rewrites and redirects

and this one to the `### Security` section:
* Denials, whitelists, and blacklists, to restrict access to services

This way, the "what is Istio" doc still includes all the functionality that it previously had.

yangminzhu: For `Denials, whitelists, and blacklists, to restrict access to services`, I can add a task under security but probably with a different name because the term "denials", "whitelists" and "blacklists" doesn't exist in authorization policy.

For `Rate limiting to dynamically limit the traffic to a service` and `Header rewrites and redirects`, I don't know who would will own and work on it, is it the Network WG or the Policies & Telemetry WG? @mandarjog @duderino 

Beyond this, @frankbu do you think we can get this PR merged first? We probably want to split the task changes in a follow-up PR and follow the normal process by creating and reviewing in a google doc before sending out a PR.

smawson:
My .02, I'm in favor of the 3 pillars (networking, security, observability) and moving all current policy content to one of those. Seems reasonable to move rate limiting and header rewrites/redirects to networking and access control to security.

rshriram:
> My .02, I'm in favor of the 3 pillars (networking, security, observability) and moving all current policy content to one of those. Seems reasonable to move rate limiting and header rewrites/redirects to networking and access control to security.

I agree. That is the reality of the project as it stands today

content/en/docs/concepts/what-is-istio/index.md
frankbu: @yangminzhu Sounds good. We can do the task changes in a follow-up PR.

PR #6472, opened 2020-02-14
Update docs for installing the operator

ostromart:
Use new istioctl operator init/remove commands instead of pulling pre-rendered manifest.


ostromart:
Moved to https://github.com/istio/istio.io/pull/6473

PR #6473, opened 2020-02-14
Update docs for installing the operator

ostromart:
Use new istioctl operator init/remove commands instead of pulling pre-rendered manifest.


ericvn:
link error should be fixed.

ostromart:
The link is definitely bad but it comes from content under public which is .gitignored so I'm not sure what the right way to fix this. Hard to see what the error has with this PR's file or how lint can be passing in master...

ericvn:
Sorry, I meant the link error should be fixed by another PR I submitted earlier

ostromart:
Ah, ic, thanks! Just merged so I hope it will pass now.

ericvn:
```
    ./en/docs/setup/install/standalone-operator/index.md
      119 | For example, to enable the grafana component and increase pilot  
      141 | in the cluster in response to IstioOperator CR updates by 

>> 2 spelling errors found in 520 files
```

sdake:
@ostromart if you could do me a solid and write a bug and assign the bug to me in isito.io to track the standalone operator without using specific Istio-provided tooling, I would appreciate that.

We can't enforce a situation in which `istioctl` will be available or usable as a deployment tool. As we have long known, there are some workflows that can't consume `istioctl` directly but instead use <insert X> where X is not `istioctl`. We also know the opposite to be a valid use case, that is, `isttioctl` is perfectly functional for an installation workflow.

PR #6484, opened 2020-02-18
Bump supported versions

howardjohn:


ericvn:
I would approve based on the assumption we bump up one K8s release per our release, but I'm not sure where that's documented.

PR #6489, opened 2020-02-18
Fix istio.io integration tests

nmittler:


prow/config/trustworthy-jwt-12.yaml
```diff
@@ -0,0 +1,16 @@
+# This configs KinD to spin up a k8s cluster with trustworthy jwt (Service Account Token Volume Projection) feature.
+# This also enables serverside dry run support
+# This must be used for 1.12 k8s versions
```
howardjohn: probably not needed, we will only run this with newer k8s I assume?

prow/config/metrics/metrics-server-deployment.yaml
```diff
@@ -0,0 +1,37 @@
+---
```
howardjohn: I suspect we don't need this. Do we have tests that rely on metrics server? Not a big harm to keep it in though

prow/integ-suite-kind.sh
```diff
@@ -39,21 +39,18 @@ export TEST_ENV=kind
...
 
-make init
-
 if [[ -z "${SKIP_SETUP:-}" ]]; then
   time setup_kind_cluster "${NODE_IMAGE:-}"
 fi
 
 if [[ -z "${SKIP_BUILD:-}" ]]; then
-  time build_images
   time kind_load_images ""
```
howardjohn: Also not needed, this is only needed for local images which are not pushed. For this case, we can skip it and kubernetes will pull the images like normal

prow/integ-suite-kind.sh
```diff
@@ -39,24 +39,17 @@ export TEST_ENV=kind
...
-
 if [[ -z "${SKIP_SETUP:-}" ]]; then
   time setup_kind_cluster "${NODE_IMAGE:-}"
 fi
 
-if [[ -z "${SKIP_BUILD:-}" ]]; then
-  time build_images
-  time kind_load_images ""
-fi
-
```
howardjohn: I think you need to make l58 `make "${@}"`

nmittler:
@howardjohn PTAL

howardjohn:
you might want to setup owners for these as well. lgtm though!

nmittler:
@howardjohn I've removed the WIP label and added a hold. Feel free to remove the hold after you take another look

PR #6497, opened 2020-02-19
Make external links in Installation Options (istioctl) page clickable

### LINKED ISSUE #6496: External links in Installation Options (istioctl) page is not clickable.
Ref: https://istio.io/docs/reference/config/istio.operator.v1alpha12.pb/

External links to the site should be clickable and it should open in new tab on click. Currently, the user has to manually copy links, past it in a new tab for detail references.

shamsher31:
This fixes #6496


frankbu:
All of these links are pulled into the docs from generated doc in other projects, so we have no ability in the istio project to change the source.

PR #6499, opened 2020-02-19
Make external links in Global Mesh Options page clickable

### LINKED ISSUE #6498: External links in Global Mesh Options page is not clickable
Ref: https://preliminary.istio.io/docs/reference/config/istio.mesh.v1alpha1/

External links to the site should be clickable and it should open in new tab on click. Currently, the user has to manually copy links, past it in a new tab for detail references.

shamsher31:
This fixes #6498

### DIFF
--- a/content/en/docs/reference/config/istio.mesh.v1alpha1/index.html
+++ b/content/en/docs/reference/config/istio.mesh.v1alpha1/index.html
@@ -1002,7 +1002,8 @@ No
 <td><code><a href="#TypeMapStringInterface2">TypeMapStringInterface2</a></code></td>
 <td>
 <p>Overrides for default values.yaml. This is a validated pass-through to Helm templates.
-See the Helm installation options for schema details: https://istio.io/docs/reference/config/installation-options/.
+See the Helm installation options for schema details: <a href="https://istio.io/docs/reference/config/installation-options/"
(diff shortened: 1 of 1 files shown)


shamsher31:
Ahh, I first thought the same that why is the file extension `.html`, No problem, I will update in `istio/api`.

shamsher31:
@ericvn Is this the right file to update https://github.com/istio/api/blob/master/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html

Just want to avoid the same mistake again.

ericvn:
@shamsher31 I admit I'm not real familiar with how that repo works, but I think the .html files are maybe generated as part of the build process. I'm not sure exactly where the change would need to be made.

frankbu:
@shamsher31 All of these links (with the exception of the first one) are pulled into the docs from generated doc in other projects, so we have no ability in the istio project to change the source. The one you can change is in this proto file: https://github.com/istio/api/blob/7fd43ea7fc2b2392c9e02b41a71776c6938416b2/operator/v1alpha1/operator.proto#L173

You can change it to:
```
See the [Helm installation options](https://istio.io/docs/reference/config/installation-options/)  for schema details.
```
and then you need to run `make gen` to regen the doc.

shamsher31:
@frankbu @ericvn raised PR in `istio/api`
https://github.com/istio/api/pull/1293

PR #6515, opened 2020-02-20
Remove webhook task from Istio 1.5 documentation

lei-tang:
Please provide a description for what this PR is for: remove webhook task as Galley and Sidecar Injector have been linked into istiod with a DNS cert signed by k8s CA.

[X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/blog/2019/webhook/index.md
```diff
@@ -37,4 +37,5 @@ will not be able to alter the webhook configurations.
 and that the certificate chain used by the webhook server is valid. This reduces the errors
 that can occur before a server is ready or if a server has invalid certificates.
 
-To try this new feature, refer to the [Istio webhook management task](/docs/tasks/security/webhook).
+To try this Istio 1.4 feature, refer to the Istio webhook management task
```
frankbu: Since this blog is obsolete, I would suggest changing it as little as possible to avoid the broken link.
```suggestion
To try this new feature, refer to the [Istio webhook management task](https://archive.istio.io/1.4/docs/tasks/security/webhook).
```

lei-tang: Done, but since https://archive.istio.io/1.4 has not been created yet, the linter may fail due to the unreachable link.

frankbu: Yeah, maybe we should disable the lint checking for archive docs to avoid this problem. I think that if you add `https://archive.istio.io/` to the `--url-ignore` list [here](https://github.com/istio/istio.io/blob/8f55ddbc67b929f7b99c525fa874c4301e3697ca/scripts/lint_site.sh#L143), the PR can merge.

ericvn: I'd maybe limit it to ignore `https://archive.istio.io/1.4` if we still want the other links (if there are any) checked. Then remove the ignore (or bump to 1.5) at istio.io branch cut time. Then could be part of the process if needed going forward.

frankbu: I'm not sure it's worth adding anything for this to the process, since there really shouldn't be pointers to archive docs. The only examples are things like blogs that are obsolete, which we really shouldn't be worrying about whether the links are broken or not. Maybe we shouldn't really even do lint checking on references from old blog entries, though I'm Not sure if there's a way to selectively exclude the lint checking for some specific files.

ericvn: I did a quick look at the English portion of the site and I think I found 7 places where we actually point at something on the _archive_ site and 4 were in blogs. So maybe it would just be easier to ignore them all as you suggest Frank.

lei-tang: Done: added archive.istio.io/ to the --url-ignore list.

PR #6522, opened 2020-02-20
Revert/egress kubernetes services

sdake:


frankbu:
Removing this task without any explanation seems like a bad idea. Since the reason for removing it is that it's not a recommended approach, maybe a [best practices](https://istio.io/docs/ops/best-practices/traffic-management/) entry should be added: `Use ServiceEntry instead of Kubernetes ExternalName or Endpoint`. This could give a short summary of what the task was saying (i.e., that you can use K8s services as long as you disable mTLS), but using Istio ServiceEntry is the recommeded best practice.

sdake:
> Removing this task without any explanation seems like a bad idea. Since the reason for removing it is that it's not a recommended approach, maybe a [best practices](https://istio.io/docs/ops/best-practices/traffic-management/) entry should be added: `Use ServiceEntry instead of Kubernetes ExternalName or Endpoint`. This could give a short summary of what the task was saying (i.e., that you can use K8s services as long as you disable mTLS), but using Istio ServiceEntry is the recommeded best practice.

@frankbu I did leave a very detailed long rationale in the PR. Unfortunately, it appears that it did not transfer over during the PR process.  You can see the rationale here: https://github.com/sdake/istio.github.io/commit/2f2afc8dc2a588f77b328cfbc7b0cc0ce2af7a98

I will attempt to compress the (very long) task into a best practices document.

Cheers
-steve

frankbu:
Closing in favor of #6664

PR #6535, opened 2020-02-22
Change how stickers should be used

sdake:


sdake:
Depends-On: https://github.com/istio/istio.io/pull/6536

Issue #6335, opened 2020-01-10
Doc tests should use snippets generated from doc


frankbu:
The [istio.io test framework](https://github.com/istio/istio/blob/master/tests/integration/istioio/README.md) requires writing tests (for Tasks and Examples) that generate [doc snippets](https://istio.io/about/contribute/creating-and-editing-pages/#snippets) that are then supposed to be used to produce content in the corresponding docs. This approach has some undesirable features:

1. Writing and reviewing docs will be hard because the document text and commands/output are in separate files. The main document `.md` file cannot be reviewed on its own.
2. Rather than just writing or updating a doc, authors must also first write or update a test program. This will definitely be an obstacle for many potential doc contributors.
3. This approach is not really "testing the docs", but rather "documenting tests", not a natural way for an author to explain how to do something.

We should instead generate snippets from the doc, and use them to write the test, instead of the other way around.

**Proposal**: snippets should work in the opposite direction. Source of truth (snippets) should be the doc, not the test. We want to test the docs, not document the tests.

1. Doc author simply writes the doc with ordinary text sections for commands and output. Like today, more or less.
2. Test author (may or may not be the same person) annotates the doc to include annotations that identify text fragments as named snippets that can be used in a test program. From the annotations, we generate snippets of code to be used in the test program.
3. Test author writes a test that uses the generated snippets for the commands and output checking. In the future, this could possibly even be generated automatically from annotations added in step 2.

This approach leaves the commands in the doc and uses them in a “doc verification test” instead of trying to generate the docs from test programs written in advance. It also properly separates roles. Doc writers can spend their time doing what they do well, explaining how something works, not writing test programs.

A general outline of the process using the [fault injection task](https://istio.io/docs/tasks/traffic-management/fault-injection/) is described here: https://github.com/istio/istio/pull/19606#discussion_r362642945

PR #6536, opened 2020-02-23
Add mirror-licenses to `make gen`

sdake:
The llinter is failing because its understanding of what licenses should be present differ
from those in the repo itself.


frankbu:
This is adding 425 files to the repo. Are we sure we need them? Why does the linter think the project needs all the dependent project licenses?

ericvn:
And I'm wondering why the tests where actually moved which require this change. We had talked about changing the snippet direction, but not sure if that meant actually moving the test cases.

Also that work was supposed to be done post 1.5, but looks like it's being done as part of the 1.5 work.

Unless I'm missing some other discussion of moving tests.

sdake:
@frankbu We are required, as per advice from legal, to include licenses for every dependency used within the go toolchain. The alternative was to vendor the dependencies and their licenses. @geeknoid talked the attorney down from that requirement.

This is why `make mirror-licenses` was born. This PR simply runs that mirroring, and then commits the changes to the repository.

The introduction of these test changes introduced 425 dependences. I am unclear why the tests were moved, or what the planning was behind this move, and as @ericvn points out, these tests were to be moved post 1.5. @nmittler can you comment on this change (IIRC you moved the test files).

There are one of two corrective actions that I see:
- this PR can be merged, moving the gate forward
- @nmittler's work can be reverted and targeted after the 1.5 docs have branched.

I am not attached to either approach. I am simply trying to unblock the docs gate, which is blocked on `make lint-licenses`.

Cheers
-steve

ericvn:
The third option is to only do what Frank has mentioned in #6335 if indeed that's different than this work.

frankbu:
@ericvn The direction we generate snippets for the tests is independent of which repo we store the tests in. I think the decision to move them happened while I was on vacation, but since these are tests for istio.io docs, it seems like a good idea to keep them in the istio/istio.io repo, instead of istio/istio. That way, if we change a doc that requires a corresponding test change, it can all be done in one PR.

sdake:
> @ericvn The direction we generate snippets for the tests is independent of which repo we store the tests in. I think the decision to move them happened while I was on vacation, but since these are tests for istio.io docs, it seems like a good idea to keep them in the istio/istio.io repo, instead of istio/istio. That way, if we change a doc that requires a corresponding test change, it can all be done in one PR.

@frankbu that wfm. The historical intent of placing the tests in istio is that it would cause the istio gate to block, forcing developers to correct the documentation when they made breaking changes. In practice, I'm not sure what happened in this case.

Cheers
-steve

ericvn:
@frankbu If you are fine with the tests moving, then that WFM. They are  under `@istio/wg-docs-maintainers-infra @istio/wg-test-and-release-maintainers` control in this repo. I think that works as you mentioned in the original PR that doc writers shouldn't have to worry about the tests.

brian-avery:
Agree. Having the tests in the istio.io repo makes sense to me, but I think we still need to trigger the tests off of the istio repo as well. The istio.io tests are really testing to make sure that the docs function against the code on Istio so I think we still need to make sure that changes in Istio don't break the Istio.io tests. Also, does this increase coupling meaning that a doc writer (whom I thought we were assuming was not necessarily a test writer), now is expected to be able to debug any tests that don't pass on their docs change?

ericvn:
@brian-avery As it is now, if the doc writer change causes an error, they will need to enlist a test person to write/review (if the doc person updates the test) the changes.

frankbu:
@ericvn @brian-avery Doc changes that affect tests are of two types:
1. a doc changed (or added) to do something new, so a test should be changed (or added) to test that new function. The test isn't failing, but is out of sync and therefore should be updated sometime soon.
2. the doc needs to be changed to use new/different APIs, base on an Istio design change. This is the case that the test may be failing, as is. Both the doc and the test need to be updated, but this kind of PR will probably be submitted by a developer, not a doc writer.

nmittler:
@brian-avery I'm not sure if we really want every istio/istio change to run the istio.io tests. Integration tests in istio/istio verify the correctness of changes. If changes in istio/istio break istio.io, we'll find out pretty soon (at the next istio.io PR), since CI is always using the latest Istio images.

We will need to clarify responsibilities a bit, however. If tests are failing due to changes in istio/istio, a developer will have to fix istio.io accordingly. This is probably a good topic for the next TOC.

sdake:
@frankbu re this PR, one of three actions is needed to get the docs moving:

- revert istio.io tests and all test-infra changes enforcing them
- disable the gen-check job
- merge this PR

Merging this PR seems natural as the tests per debate in this PR are that istio.io tests are to land in this repository. Not merging this PR means that either of the options must be chosen.

We can take on a separate discussion of whether to disable istio.io tests for the moment to enable the 1.5 release in a separate PR to test-infra.

Cheers
-steve

frankbu:
Closing, not needed since linter is not failing.

Issue #6542, opened 2020-02-24
Add samples, out, install directories to .gitignore


su225:
When I tried running istio.io tests I found that these directories are generated. Ideally, they should be ignored, but they are not and show up in diffs. This should be fixed.

howardjohn:
fixed https://github.com/istio/istio.io/pull/6543

PR #6543, opened 2020-02-24
(#6542) - add install, samples and out directories to .gitignore

su225:
Output directories and directories created for the sake of running tests must be ignored. So add these to .gitignore directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,12 @@ public
 resources
 node_modules
 generated
-install
-samples
-out
+
+# install generated files
+install/**
+
+# contains samples to demo Istio 
+samples/**
+
+# contains the built artifacts
+out/**
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ public
 resources
 node_modules
 generated
+install
+samples
+out


ericvn:
While I agree `out` should maybe had been added with the build tooling update, the install and samples are part of moving tests that I'm wondering what the status will be (see https://github.com/istio/istio.io/pull/6536).

su225:
Sure. Let me wait.

.gitignore
```diff
@@ -13,3 +13,6 @@ public
 resources
 node_modules
 generated
+install
+samples
+out
```
ericvn: Can you add a comment: # Contains the built artifacts

.gitignore
```diff
@@ -13,3 +13,6 @@ public
 resources
 node_modules
 generated
+install
```
ericvn: Add a comment: # Install generated files

.gitignore
ericvn: Does this need to be out/?

.gitignore
ericvn: Based on istio/istio, does this need to be install/** if we are only adding one line?

.gitignore
su225: Changed.

.gitignore
su225: Done.

.gitignore
su225: Done.

PR #6547, opened 2020-02-25
Update for 1.5 profiles

### LINKED ISSUE #6502: In Istio 1.5 with istiod we need to update the documentation page to reflect the changes.
Currently the docs show all the services which are not a part of istiod. we need to update the docs to reflect the changes.

Screen Shot 2020-02-18 at 10 37 39 AM

### LINKED ISSUE #6645: [1.5 Community Testing]: listing available configuration profiles shows wrong results
When following the steps in https://preliminary.istio.io/docs/setup/install/istioctl/#display-the-configuration-of-a-profile to list the available configuration profiles, the output is different from what is mentioned in the docs. 

I verified the list of configuration profiles mentioned here: https://preliminary.istio.io/docs/setup/additional-setup/config-profiles/ and that matches what is mentioned in the output in the docs. 

Here is the expected output per the docs:
`minimal
  remote
  sds
  default
  demo`

whereas this is what is seen on release-1.5-beta4 build:
 `demo
  empty
  minimal
  remote
  seperate
  default`

Not sure if this is a doc bug or a feature bug.

howardjohn:
Fixes #6502
Fixes #6645

### DIFF
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -38,7 +38,7 @@ your specific needs. The following built-in configuration profiles are currently
 
 1. **empty**: deploys nothing. This can be useful as a base profile for custom configuration.
 
-1. **separate**: this enables the legacy microservices deployment model. This is not recommended, and will be removed in future releases.
+1. **separate**: deploys Istio following the legacy micro-services model. This profile is not recommended because it won't be supported in future releases.
 
 The components marked as **X** are installed within each profile:
 
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -34,6 +34,12 @@ your specific needs. The following built-in configuration profiles are currently
     [multicluster mesh](/docs/ops/deployment/deployment-models/#multiple-clusters) with a
     [shared control plane](/docs/setup/install/multicluster/shared-vpn/) configuration.
 
+1. **preview**: this profile enables experimental features that may not meet the same level of stability, performance, and security as the default profiles.
+
+1. **empty**: deploys nothing. This can be useful as a base profile for custom configuration.
+
+1. **separate**: this enables the legacy microservices deployment model. This is not recommended, and will be removed in future releases.
+
 The components marked as **X** are installed within each profile:
 
 |     | default | demo | minimal | remote |
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -34,7 +34,7 @@ your specific needs. The following built-in configuration profiles are currently
     [multicluster mesh](/docs/ops/deployment/deployment-models/#multiple-clusters) with a
     [shared control plane](/docs/setup/install/multicluster/shared-vpn/) configuration.
 
-1. **preview**: this profile enables experimental features that may not meet the same level of stability, performance, and security as the default profiles.
+1. **preview**: enables experimental features that lack the stability, performance, and security of features in the default profiles.
 
 1. **empty**: deploys nothing. This can be useful as a base profile for custom configuration.
 
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -30,32 +30,23 @@ your specific needs. The following built-in configuration profiles are currently
 
 1. **minimal**: the minimal set of components necessary to use Istio's [traffic management](/docs/tasks/traffic-management/) features.
 
-1. **sds**: similar to the **default** profile, but also enables Istio's [SDS (secret discovery service)](/docs/tasks/security/citadel-config/auth-sds).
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -30,32 +30,23 @@ your specific needs. The following built-in configuration profiles are currently
...
-| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`istio-pilot` | X | X | X | X | |
-| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`istio-policy` | X | X | | X | |
-| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`istio-sidecar-injector` | X | X | | X | X |
-| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`istio-telemetry` | X | X | | X | |
-| Addons | | | | | | |
-| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`grafana` | | X | | | |
-| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`istio-tracing` | | X | | | |
-| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`kiali` | | X | | | |
-| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`prometheus` | X | X | | X | |
+|     | default | demo | minimal | remote |
```
ericvn: Do we need to create content to mention `empty` and `separate`?

howardjohn:
I think we should not document then because we don't want people to use them
If that's not how the docs team sees things I can add it

frankbu:
@howardjohn I think we should at least mention `empty` and `separate` and say we don't expect people to use them. Otherwise people see them in `istioctl profile list` but there will be nothing in the doc that says what they're for.

content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -30,32 +30,29 @@ your specific needs. The following built-in configuration profiles are currently
...
 1. **minimal**: the minimal set of components necessary to use Istio's [traffic management](/docs/tasks/traffic-management/) features.
 
-1. **sds**: similar to the **default** profile, but also enables Istio's [SDS (secret discovery service)](/docs/tasks/security/citadel-config/auth-sds).
-    This profile comes with additional authentication features enabled by default (Strict Mutual TLS).
-
 1. **remote**: used for configuring remote clusters of a
     [multicluster mesh](/docs/ops/deployment/deployment-models/#multiple-clusters) with a
     [shared control plane](/docs/setup/install/multicluster/shared-vpn/) configuration.
 
+1. **preview**: this profile enables experimental features that may not meet the same level of stability, performance, and security as the default profiles.
```
rcaballeromx:
```suggestion
1. **preview**: enables experimental features that lack the stability, performance, and security of features in the default profiles.
```

howardjohn: nit: To me this makes it sound like they are all unstable, bad performance, and not secure. Do you agree? I think change `and` to `or` or `and/or` may make more sense. Let me know what you think

content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -30,32 +30,29 @@ your specific needs. The following built-in configuration profiles are currently
...
+
+1. **empty**: deploys nothing. This can be useful as a base profile for custom configuration.
```
rcaballeromx: The phrase `deploys nothing.` can be confusing for readers. They might ask: Will the command do *nothing*? Which is not what happens. This profile just means that no configuration profile is applied, correct? The text should let readers know that's the case.

content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -30,32 +30,29 @@ your specific needs. The following built-in configuration profiles are currently
...
+
+1. **separate**: this enables the legacy microservices deployment model. This is not recommended, and will be removed in future releases.
```
rcaballeromx:
```suggestion
1. **separate**: deploys Istio following the legacy micro-services model. This profile is not recommended because it won't be supported in future releases.
```

content/en/docs/setup/additional-setup/config-profiles/index.md
howardjohn: It literally does nothing. Its not the same as no profile, which would mean the default one. It was added so you can do things like `--set profile=empty --set something.enabled=true` instead of needing to go and disable all the default stuff.

I didn't want to document it since its confusing but I got recommended to add all profiles here

content/en/docs/setup/additional-setup/config-profiles/index.md
ericvn: We should probably get this one in sooner than later. Any thoughts on adding the additional columns?

content/en/docs/setup/additional-setup/config-profiles/index.md
shamsher31: I think instead of `deploys nothing`,  the following would be better.  

`The empty profile has everything disabled.`

Issue #6544, opened 2020-02-24
istio repository cloned should not start with a double quote directory


su225:
While running istioio tests I found that there is a step to clone istio repository. I also found that it runs `bin/init.sh` script in this repository. I found a weird output. Some lines are as follows
```bash
+ [[ ! -d "/go/src/istio.io/istio" ]]
+ export ISTIO_BRANCH=master
+ ISTIO_BRANCH=master
+ pushd '"/go/src/istio.io/istio"'
/work/"/go/src/istio.io/istio" /work
+ git checkout master
Already on 'master'
Your branch is up to date with 'origin/master'.
+ go install ./istioctl/cmd/istioctl
+ popd
/work
```
Note the double quotes. It created a directory with that name! This is definitely an annoying bug that must be fixed. I checked out `bin/init.sh` script and found the following lines
```bash
pushd "${ISTIO_GO}"
git checkout "${ISTIO_BRANCH}"
go install ./istioctl/cmd/istioctl
popd
```
I think the environment variable `ISTIO_GO` passed to the `gcr.io/istio-testing/build-tools` container (tag: `master-2020-02-14T13-09-14`) should be fixed to remove extra double quotes. Also the directory cloned should be added to `.gitignore` of the repository so that it does not show up in diffs.

su225:
@ericvn - could you please tell me where `Dockerfile` of `build-tools` live so that I can fix that?

ericvn:
@su225: https://github.com/istio/tools/blob/master/docker/build-tools/Dockerfile is the location for the Dockerfile.

su225:
I'm wrong. Sorry. I see where that is coming from. It is from `Makefile.core.mk` and before launching tests, `init` target is launched to run `bin/init.sh` script. Here it is
```bash
# Set the directory for Istio source.
ISTIO_GO ?= "${GOPATH}/src/istio.io/istio"
export ISTIO_GO
```
It seems easier now :)

su225:
@ericvn  - Based on your comment from https://github.com/istio/istio.io/pull/6543 which also indicates the possibility of moving istio.io tests back, I will wait for some time.

su225:
This issue is fixed now. So closing it.

PR #6555, opened 2020-02-25
(#6544) cloned istio repository path should not have double quotes

su225:
Cloned Istio repository during testing should not map to a directory with double-quote (") as its name. It is very annoying.


su225:
Also add cloned repository to `.gitignore`

PR #6587, opened 2020-02-25
[Release-1.5] Remove tasks referring to the deprecated tool

diemtvu:
`istioctl authn tls-check` is deprecated and no longer works with beta mTLS API.


content/en/docs/tasks/security/authentication/mutual-tls/index.md
```diff
@@ -69,84 +69,6 @@ $ kubectl exec $(kubectl get pod -l app=httpbin -o jsonpath={.items..metadata.na
 
 Please check [Istio identity](/docs/concepts/security/#istio-identity) for more information about  _service identity_ in Istio.
 
-## Verify mutual TLS configuration
-
-Use [`istioctl authn tls-check`](/docs/reference/commands/istioctl/#istioctl-authn-tls-check) to check if the mutual TLS settings are in effect. The `istioctl` command needs the client's pod because the destination rule depends on the client's namespace.
```
frankbu: Is there a replacement for this command? Don't we need to change the instructions in this section to tell people how to verify the mutual TLS configuration some other way, rather than just delete the whole doc?

diemtvu: We don't have a equivalent tool at this point. We will update the doc when we have one.

On the other hand, I'm also working on different tutorial that using telemetry to verify the actual traffic than the config itself. It could be a replacement for this whole doc, given citadel and filemount secret is also very much out-of-date.

diemtvu: @frankbu , is this ok to submit?

adammil2000:
I assume the error here is this PR needs a rebase?

PR #6599, opened 2020-02-25
Update circuit breaker task for 1.5

rlenglet:
Tested with [Istio 1.5.0-beta.3](https://github.com/istio/istio/releases/tag/1.5.0-beta.3).


ericvn:
Don't need the cherry-pick label as _master_ will be the 1.5 version of the doc when it is released.

PR #6600, opened 2020-02-25
update operator ref doc

richardwxn:
generate by grabbing the one in istio/api instead of the deprecated one istio/istio


richardwxn:
@jasonwzm Other than the location tag issue you have fixed, seems like there is another issue with the comments, as you can see from: https://github.com/istio/api/blob/master/operator/v1alpha1/istio.operator.v1alpha1.pb.html#L2. The generated html file is missing the top comments part from the proto file. I guess it is also related to position?

richardwxn:
Hope someone can help review it, since right now the reference page is still pointing to the html generated by old API: https://preliminary.istio.io/docs/reference/config/istio.operator.v1alpha12.pb/.
@frankbu @ericvn @adammil2000

content/en/blog/2019/introducing-istio-operator/index.md
```diff
@@ -39,7 +39,7 @@ is tightly integrated with the tools to prevent common errors and simplify troub
 Every operator implementation requires a
 [custom resource definition (CRD)](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)
 to define its custom resource, that is, its API. Istio's operator API is defined by the
-[`IstioControlPlane` CRD](/docs/reference/config/istio.operator.v1alpha12.pb/),
+[`IstioControlPlane` CRD](https://github.com/istio/istio.io/blob/release-1.4/content/en/docs/reference/config/istio.operator.v1alpha12.pb/index.html),
```
esnible: This link takes the user to some HTML wrapped in Github UI.  Is that intended?

frankbu: I think this should point to the 1.4 archive:
```suggestion
[`IstioControlPlane` CRD](https://archive.istio.io/v1.4/docs/reference/config/istio.operator.v1alpha12.pb/),
```

PR #6605, opened 2020-02-26
Add https overlay tests

jacob-delgado:
To Fix https://github.com/istio/istio.io/issues/6570

### DIFF
--- a/tests/security/https-overlay/mtls-enabled/https_overlay_test.go
+++ b/tests/security/https-overlay/mtls-enabled/https_overlay_test.go
@@ -1,0 +1,40 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


jacob-delgado:
The tests were showing they passed, but https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio.io/6605/k8s-tests_istio.io/228 showed otherwise. Is that a known issue?

tests/security/https-overlay/mtls-enabled/https_overlay_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"testing"
+
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/istioio"
+)
+
+// TestHTTPSOverlayMtlEnabled simulates the task in docs/tasks/security/authentication/https-overlay/
+// with a sidecar and global mTLS is enabled
+func TestHTTPSOverlayMtlEnabled(t *testing.T) {
```
jacob-delgado: I had to break this out into 2 separate tests due to the docs saying you had to do a re-install. Is there a simpler way?

https://istio.io/docs/tasks/security/authentication/https-overlay/#create-an-https-service-with-istio-sidecar-with-mutual-tls-enabled

jacob-delgado:
This issue was closed https://github.com/istio/istio.io/issues/6570. It was unfortunate that I worked on this one.

PR #6610, opened 2020-02-26
Remove warning about multiple replicas in citadel deployment

jwendell:


jwendell:
That feature is supported.
https://discuss.istio.io/t/citadel-support-multiple-replica-or-not/5472

PR #6612, opened 2020-02-26
More fixes and cleanup for makefiles

nmittler:


Makefile
```diff
@@ -133,7 +133,7 @@ default:
 
 .PHONY: default
 
-else
+else # !BUILD_WITH_CONTAINER
```
howardjohn: this is from common files

jwendell: This file should not be included in the changes

nmittler:
@jwendell @howardjohn PTAL

PR #6621, opened 2020-02-27
Updated pilot options

hustshawn:
pilot no longer use `enableProtocolSniffing` anymore, it has two separate settings for both inbound and outbound instead.


hustshawn:
Regarding to https://github.com/istio/istio/issues/18110

ericvn:
Looking at the updated file, the changes are after the
`<!-- AUTO-GENERATED-START -->`
so I believe these changes should come in as part of some dependency updates. That said, I'm not sure if//how that's done. https://github.com/istio/istio.io/pull/6419/files doesn't seem to have this update.

adammil2000:
This PR attempts to modify the output of autogenerated content, closing.

hustshawn:
> This PR attempts to modify the output of autogenerated content, closing.

So where is the right way to update?

Issue #6557, opened 2020-02-25
Getting started page has several issues


ericvn:
https://preliminary.istio.io/docs/setup/getting-started/

Heading 1: Download and install Istio
Heading 2: Install Istio

We should remove the install from the first heading.

FYI. Table pointing to these headings:
Download the release
Install Istio

Also there is a line which may not make sense with the new installation preferences
   `Installation YAML files for Kubernetes in install/kubernetes` '

ericvn:
FYI @adammil2000

adammil2000:
Thanks for the heads-up, there's a new Getting Started PR coming shortly that will fix this and other issues.

ericvn:
Hopefully, I can check in on the next testing day.

dgn:
Fixed by the PR

PR #6642, opened 2020-02-28
Improved Istio Getting Started

adammil2000:
This is an updated Istio Getting Started to address pain points identified by recent UX testing.


content/en/docs/setup/getting-started/index.md
```diff
@@ -9,46 +9,30 @@ aliases:
...
 
-This guide is intended for users who are new to Istio and lets you quickly evaluate Istio by installing the
-`demo` [configuration profile](/docs/setup/additional-setup/config-profiles/).
+This guide lets you quickly evaluate Istio.
 
-If you are already familiar with Istio or interested in installing other configuration profiles
-or a more advanced [deployment model](/docs/ops/deployment/deployment-models/),
-follow the [installing with {{< istioctl >}} instructions](/docs/setup/install/istioctl) instead.
+These steps assume you are using
+[Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) and have a
```
frankbu: Assuming minikube doesn't sound like a simplification. Minikube is not the quickest way to get started unless minikube is already installed. Creating a cluster using Docker Desktop or IBM Cloud (and I assume other clouds like GCE) is an easier way to start from scratch than installing and setting up minikube and minikube tunnel.

These instructions should just say you need to start with a Kubernetes cluster (it can mention minikube as one way to do it), but assuming minikube throughout the instructions is not a good idea IMO.

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,331 @@ Download the Istio release which includes installation files, samples, and the
...
-    - Sample applications in `samples/`
-    - The [`istioctl`](/docs/reference/commands/istioctl) client binary in the `bin/` directory. `istioctl` is
-      used when manually injecting Envoy as a sidecar proxy.
+    {{< text bash >}}
+    $ export PATH=$PWD/bin:$PATH
+    {{< /text >}}
 
-1.  Add the `istioctl` client to your path, on a macOS or
-    Linux system:
+1.  For this tutorial, we use the `demo` profile. It's selected to have a good set
```
frankbu: This isn't a "tutorial", it's the setup instructions.

content/en/docs/setup/getting-started/index.md
```diff
@@ -9,46 +9,30 @@ aliases:
...
-Some platforms provide a {{< gloss >}}managed control plane{{< /gloss >}} which you can use instead of
-installing Istio manually. If this is the case with your selected platform, and you choose to use it,
-you will be finished installing Istio after creating the cluster, so you can skip the following instructions.
-For more information, see your platform service provider's documentation.
-
-## Download and install Istio {#download}
-
-Download the Istio release which includes installation files, samples, and the
-[{{< istioctl >}}](/docs/reference/commands/istioctl/) command line utility.
+## Install Istio {#download}
```
frankbu: This section (#download) is referenced from all the other more advanced install instructions, so you can't change it to do install as well.

content/en/docs/setup/getting-started/index.md
ericvn: I don't know that IBM Cloud is unless they also have an account and the CLI installed. Similar for GKE and others. I'm not sure if they would already have Docker installed (we ask our developers to use it, but potential uses kicking the tires?).

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,331 @@ Download the Istio release which includes installation files, samples, and the
...
 - [Rate limiting](/docs/tasks/policy-enforcement/rate-limiting/)
-- [Ingress gateways](/docs/tasks/traffic-management/ingress/ingress-control/)
 - [Accessing external services](/docs/tasks/traffic-management/egress/egress-control/)
 - [Visualizing your mesh](/docs/tasks/observability/kiali/)
 
-## Prepare for production deployment
-
-Before you install and customize Istio for production use,
-see the following topics:
+Before you customize Istio to fit your platform and intended use,
```
brian-avery: This content seems out of place in getting started. I wonder if it should be promoted to its own page.

brian-avery: It feels like "what is Istio". Ok, you installed Istio, ready to promote to prod?

content/en/docs/setup/getting-started/index.md
frankbu: @ericvn Yes, that sounds good. In fact, that's similar to what linkerd does in their quick start. They just say that you need a kubernetes cluster and they point to minikube and docker desktop as a couple of easy choices if you're also new to Kubernetes.

Regarding IBM Cloud or Docker Desktop being easier than minikube, I think that a good percentage of the people that are looking at Istio for the first time are not also new to Kubernetes and Docker. I'm assuming that many of the people trying to get started with Istio already use Kubernetes, and/or Docker, so they probably already have something like IBM Cloud or Docker Desktop or Minikube configured. So we can't say that regardless of what may be the easiest way to set up a cluster in you case, these instructions are specifically for minikube, so if you want to get started quickly with Istio you need to install minikube first.

content/en/docs/setup/getting-started/index.md
frankbu:
```suggestion
## Download Istio {#download}
```

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,331 @@ Download the Istio release which includes installation files, samples, and the
     $ cd istio-{{< istio_full_version >}}
     {{< /text >}}
 
-    The installation directory contains:
+1.  Add the `istioctl` client to your path (Linux or macOS):
 
-    - Installation YAML files for Kubernetes in `install/kubernetes`
-    - Sample applications in `samples/`
-    - The [`istioctl`](/docs/reference/commands/istioctl) client binary in the `bin/` directory. `istioctl` is
-      used when manually injecting Envoy as a sidecar proxy.
```
frankbu: Like I said above, these instructions are pointed to by all the more advanced install options, so this is pretty useful information that we probably don't want to lose.

content/en/docs/setup/getting-started/index.md
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
frankbu: Also should we link to the configure profile page?
```suggestion
1.  For this tutorial, we use the `demo` [configuration profile](link). It's selected to have a good set
```

content/en/docs/setup/getting-started/index.md
adammil2000: Reworded this section to be more clear about what we're telling the user to do.

content/en/docs/setup/getting-started/index.md
```diff
@@ -9,46 +9,30 @@ aliases:
...
+To use a cluster on a different platform or in the cloud, follow the appropriate
+[platform-specific setup instructions](/docs/setup/platform-setup/) before you
+continue.
 
-Before you can install Istio, you need a {{< gloss >}}cluster{{< /gloss >}} running a compatible version of Kubernetes.
-Istio {{< istio_version >}} has been tested with Kubernetes releases {{< supported_kubernetes_versions >}}.
+Follow these steps to get started with Istio:
 
-Create a cluster by selecting the appropriate [platform-specific setup instructions](/docs/setup/platform-setup/).
+1. [Download and install Istio](#download)
```
ericvn: You have two sections, so not sure if combining the into one is beneficial or not. I feel breaking this into two items and pointing to each section makes the most sense.

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,333 @@ Download the Istio release which includes installation files, samples, and the
...
+    {{< /text >}}
+
+## Open the application to outside traffic {#ip}
+
+Your app is deployed but not accessible from the outside. To make it accessible,
+you need to create an
+[Istio Ingress Gateway](/docs/concepts/traffic-management/#gateways), which maps a path to a
+route at the edge of your mesh.
+
+1.  Connect this application to be published by the Istio gateway:
```
ericvn: Grammer..  c/by/to might be enough to make it sound OK.

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,333 @@ Download the Istio release which includes installation files, samples, and the
...
+
+Ensure an IP address was successfully assigned to the environment variable:
+
+{{< text bash >}}
+$ echo $INGRESS_HOST
+192.168.4.102
+{{< /text >}}
 
-A benefit of Istio is automatic sidecar injection, which allows your applications to work in the service mesh without modification. To take advantage of this feature, enable sidecar injection by adding the `istio-injection=enabled` label to the Kubernetes namespaces in which you plan to deploy your applications. For more information, see [Istio sidecar injector](/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection).
+Run this command in a new terminal window to start a Minikube tunnel that
```
ericvn: I'm still not sure this, the `minikube tunnel` is needed. It probably doesn't hurt.

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,333 @@ Download the Istio release which includes installation files, samples, and the
...
+    {{< text bash >}}
+    $ echo $GATEWAY_URL
+    192.168.99.100:32194
+    {{< /text >}}
+
+### Verify external access {#confirm}
 
-## Next evaluation steps
+Confirm that the Bookinfo application is accessible from outside the cluster by
+navigating in your browser to `http://<your $GATEWAY_URL value>/productpage`.
```
ericvn: Put this command in a text bash.

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,333 @@ Download the Istio release which includes installation files, samples, and the
...
 
-After installing Istio and deploying an application,
-see the following topics to explore more Istio features:
+## View the dashboard {#dashboard}
+
+Istio has several optional dashboards installed by the `demo` installation. The
```
ericvn: Is this saying the dashboards are installed by the installation?  I 'm not sure that's 100% accurate. On the flip side, if I install Kiali outside of Istio, can I still get at it's dashboard with the CLI? @esnible, do you know?

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,333 @@ Download the Istio release which includes installation files, samples, and the
...
+    checking for the page title in the response:
+
+    {{< text bash >}}
+    $ kubectl exec -it $(kubectl get pod -l app=ratings -o jsonpath='{.items[0].metadata.name}') -c ratings -- curl productpage:9080/productpage | grep -o "<title>.*</title>"
+    <title>Simple Bookstore App</title>
+    {{< /text >}}
+
+## Open the application to outside traffic {#ip}
+
+Your app is deployed but not accessible from the outside. To make it accessible,
```
frankbu:
```suggestion
The Bookinfo application is deployed but not accessible from the outside. To make it accessible,
```

content/en/docs/setup/getting-started/index.md
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
adammil2000: fixed - they are broken up in the content, but for this summary at the beginning we mention. them combined.

content/en/docs/setup/getting-started/index.md
adammil2000: I had to add it for my environment to work. It literally was dead for hours until specifically this step.

content/en/docs/setup/getting-started/index.md
adammil2000: This isn't a bash command. But is there a way to open the GUI browser to a specific URL using a bash command?

content/en/docs/setup/getting-started/index.md
adammil2000: The demo install definitely adds the Kiali dashboard per my testing. The others I'm not sure.

content/en/docs/setup/getting-started/index.md
adammil2000: Users were failing to successfully complete the old Getting Started and the feedback was asking us to ensure we had the exact successful steps documented for a particular install path, minikube being the one that's free and accessible to everyone as our common starting point.

But we add note at the beginning the platform setup link and provide the "Other" tab at the mid-point so someone can still be successful using IBM or GKE if they prefer. Maybe we should just rename the Other tab to be more specific about IBM cloud, GKE, etc?

content/en/docs/setup/getting-started/index.md
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
ericvn: I think you can just use text. I incorrectly had bash in my prior remark.

content/en/docs/setup/getting-started/index.md
```diff
@@ -52,6 +52,14 @@ Follow these steps to get started with Istio:
     $ cd istio-{{< istio_full_version >}}
     {{< /text >}}
 
+    The installation directory contains:
+
+    - Installation YAML files for Kubernetes in `install/kubernetes`
+    - Sample applications in `samples/`
+    - The [`istioctl`](/docs/reference/commands/istioctl) client binary in the
+      `bin/` directory. `istioctl` is used if you choose to manually inject
```
ericvn: `istioctl` is also used to get at the dashboard and install Istio. I would be inclined to simply remove the `istioctl is used ...` instead of adding a list of items.

content/en/docs/setup/getting-started/index.md
frankbu: I see, so the instructions don't really "assume you are using minikube". The instructions work for any platform, it's just that minikube is the fast path. I guess that's OK, but you should reword the above sentence to just say that the user needs a Kubernetes cluster to get started, and mention minikube as a good choice for somebody also new to k8s. Then you can make minikube the simplest option in the following instructions, just so long as the instructions also work well on the other platforms too

content/en/docs/setup/getting-started/index.md
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
```diff
@@ -68,146 +52,331 @@ Download the Istio release which includes installation files, samples, and the
...
-      used when manually injecting Envoy as a sidecar proxy.
+    {{< text bash >}}
+    $ export PATH=$PWD/bin:$PATH
+    {{< /text >}}
 
-1.  Add the `istioctl` client to your path, on a macOS or
-    Linux system:
+1.  For this tutorial, we use the `demo` profile. It's selected to have a good set
+    of defaults for testing, but there are other profiles for production or
+    performance testing.
```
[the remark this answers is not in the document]
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
adammil2000: fixed

content/en/docs/setup/getting-started/index.md
```diff
@@ -72,142 +52,341 @@ Download the Istio release which includes installation files, samples, and the
...
+    {{< /text >}}
 
-## Next evaluation steps
+### Verify external access {#confirm}
+
+Confirm that the Bookinfo application is accessible from outside the cluster by
+navigating in your browser to:
+
+{{< text plain >}}
+`http://<your $GATEWAY_URL value>/productpage`
```
frankbu: This URL needs to be wrapped in back-ticks or a text section, but not both. Pick one or the other. Personally, I think back-ticks are better since the features of a text section (e.g., copy button) aren't really that useful here, since most of the string (`<your $GATEWAY_URL value>`) needs to be replaced before using it anyway.

ericvn: And I had previously asked him to put in a text box, so I'll let Adam decide on which way to go with this. I also did see in the Learning series of pages, where they have an _echo_ in a text bash box to actually output the URL (it'll expand the envvars), and have you cut/paste from that output to the browser.

ericvn:
fixes #6557

content/en/docs/setup/getting-started/index.md
frankbu: Tough call whether or not the echo approach is too basic-tutorial like for the setup instructions, but it is a simple way to get the right value for the browser.

This does get me to wonder if starting the bookinfo app in the setup getting started page is messing up the flow for other docs. The bookinfo example points to the setup instructions as a before you begin step, and then proceeds to give the instructions to start the sample. If the user followed the get started setup instructions, instead of other more advanced install instructions, the following bookinfo instructions are not needed anymore, since bookinfo is already up and running. I think get started and the bookinfo example need to add some text to explain that the get started install instructions are a quick start including bookinfo, so the bookinfo sample instructions are only for users that didn't use the quick start. Otherwise people will be confused about what instructions they need to follow.

ericvn: I have that same concern and noted it somewhere earlier (can't find it right now). We point to getting started in some docs and I don't think they expect that Bookinfo is necessarily installed after that is complete.

adammil2000: Going with the echo approach as it is the most user friendly and most accurate. 

Regarding the other comments, the Getting Started is getting hard to optimize for its original purpose because so many other docs depend on it for its sub-sections, e.g. other docs linking to its downloading Istio section, and others. Also when another doc links to Getting Started for one of its sub-sections, readers can easily make the mistake of forgetting to go back to the original doc at the immediate conclusion of the given subsection.

A better way to handle this situation is to remake the truly shareable subsections as single-sourced boilerplates instead of so much linking back and forth. This lets you use the instructions in context everywhere it's needed without linking so much, maintaining single-sourced content re-use, but also letting you remove the boilerplate and customize something one-off as needed in the places where the info truly needs to be different.

frankbu: @adammil2000 I don't think there's really that much linking back and forth with getting started (downloading section is the main one I know about). The bigger problem is that lots of docs link generally back to the setup docs, not getting started specifically, to tell people they need to install istio before they can do pretty much anything else. Users can choose from several ways to install istio, but getting started is the doc that stands out as the doc in the setup section to click first. So, I think that what's missing in this new version is a short description at the top of the doc, similar to what was there before, that says that this is a quick start for new users, but if you are already familiar with Istio or interested in installing other configuration profiles or a more advanced deployment model, follow the installing with istioctl instructions instead. The bookinfo sample should also have a tip/warning in the before you begin section that says if you've installed Istio with the getting started (quick start) guide, you will already have bookinfo deployed and can ignore the following instructions.

adammil2000: fixed

content/en/docs/setup/getting-started/index.md
```diff
@@ -72,142 +56,341 @@ Download the Istio release which includes installation files, samples, and the
...
+    reviews-v3-67b4988599-7nhwz       0/2     PodInitializing   0          2m29s
     {{< /text >}}
 
     {{< tip >}}
-    If your cluster runs in an environment that does not
-    support an external load balancer (e.g., minikube), the
-    `EXTERNAL-IP` of `istio-ingressgateway` will display
-    `<pending>`. To access the gateway, use the service's
-    `NodePort`, or use port-forwarding instead.
+    Re-run the previous command and wait until all pods report READY 2 / 2 and
```
craigbox: @howardjohn had a good example in another repo. This might work instead:

kubectl wait --for=condition=Available deployment --all --timeout=120s -n default

content/en/docs/setup/getting-started/index.md
```diff
@@ -72,142 +56,341 @@ Download the Istio release which includes installation files, samples, and the
...
+    {{< /text >}}
+
+### Verify external access {#confirm}
+
+Confirm that the Bookinfo application is accessible from outside. Copy the
+output of this command and paste into your browser:
 
 {{< text bash >}}
-$ kubectl create -n <namespace> -f <your-app-spec>.yaml
+$ echo http://$GATEWAY_URL/productpage
```
pazkooda, line 334: Is this actually valid thing to put into browser?
`GATEWAY_URL` will have internal node port set (here: `32194`).
Yet output from:
```sh
kubectl get svc istio-ingressgateway -n istio-system
```
shows that this is **destination** port for request that will originate **to** `http://${INGRESS_HOST}:80` (!)
The former one (with port `80`) works for me while testing. The one with `nodePort` gives no responses at all - from local browser.

craigbox: Hi pazkooda! Please raise this as an issue, rather than a comment on this merged PR, so it can be triaged and looked at.  Thanks!

Issue #6647, opened 2020-02-28
Issue with docs/setup/install/helm/index.md


jpapejr:
On [this](https://preliminary.istio.io/docs/setup/install/helm/#option-2-install-with-helm-and-tiller-via-helm-install) page, under Option 2 (Helm install with Tiller), step 5, for the `Mutual TLS enabled` install option, the command is shown as 

```
helm install install/kubernetes/helm/istio --name istio --namespace istio-system \
    --values install/kubernetes/helm/istio/values-istio-demo.yaml \
    --set global.controlPlaneSecurityEnabled=true \
    --set global.mtls.enabled=true | kubectl apply -f -
```

However, the piped output to `kubectl apply -f -` is erroneous because the prior command does not output any YAML that the `kubectl` command would then read and apply from stdout of the previous `helm install` command. 

The `| kubectl apply -f -` should be removed from the docs.

jpapejr:
Resolved with #6648

PR #6648, opened 2020-02-28
Removed unneeded command from snippet

jpapejr:
> Should address https://github.com/istio/istio.io/issues/6647.. The PR removes an seemingly erroneously included command in the Helm install (with Tiller) docs for 1.5 beta.3 (specifically, but it also shows in 1.4 and earlier docs.


jpapejr:
@googlebot I signed it!

PR #6652, opened 2020-02-28
Fixed anchor links to point at glossary

### LINKED ISSUE #6651: Links in definitions don't work
In the documentation, you can click on phrases to see their definitions. Some of these definitions include links. The links all point to an anchor on the current page and don't work since they aren't defined on the current page.

For example, click on "workloads" on https://preliminary.istio.io/docs/tasks/security/authorization/authz-http/#configure-access-control-for-workloads-using-http-traffic. In the definition of workloads, there are links to operators, attributes, workload instance, and pod. None of these links work.

brian-avery:
This fixes #6651. In it, many pages, are local links (i.e. #envoy). This works when the user views the glossary page itself, but fails when the user takes advantage of the embedded definitions in other pages due to the fact that the anchors are now located on those other pages. 

[ x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/reference/glossary/adapters.md
+++ b/content/en/docs/reference/glossary/adapters.md
@@ -2,7 +2,7 @@
 title: Adapters
 ---
 
-Adapters are plug-ins to [Mixer](#mixer), Istio's [policy and telemetry](/docs/reference/config/policy-and-telemetry/) component, which enable it to interface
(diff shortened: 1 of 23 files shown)


frankbu:
I suppose the site generator could/should be doing this when transforming from md to html, but since it's not we need to make sure all the links are absolute.

PR #6673, opened 2020-03-02
Fix pilot -> istiod for config validation docs

howardjohn:


content/en/docs/ops/common-problems/validation/index.md
```diff
@@ -147,24 +147,24 @@ configuration. Verify that it is not empty (see [verify webhook
...
 
     {{< text bash >}}
-    $ for pod in $(kubectl -n istio-system get pod -lapp=pilot -o jsonpath='{.items[*].metadata.name}'); do \
+    $ for pod in $(kubectl -n istio-system get pod -lapp=istiod -o jsonpath='{.items[*].metadata.name}'); do \
         kubectl -n istio-system logs ${pod} \
     done
     {{< /text >}}
 
-1. If the patching failed, verify the RBAC configuration for Pilot:
+1. If the patching failed, verify the RBAC configuration for Istiod:
```
ericvn: Need to backtrack Istiod. Was there ever a capitalization determined for this?
```suggestion
1. If the patching failed, verify the RBAC configuration for `Istiod`:
```

content/en/docs/ops/common-problems/validation/index.md
```diff
@@ -147,24 +147,24 @@ configuration. Verify that it is not empty (see [verify webhook
...
 
     {{< text bash >}}
-    $ for pod in $(kubectl -n istio-system get pod -lapp=pilot -o jsonpath='{.items[*].metadata.name}'); do \
+    $ for pod in $(kubectl -n istio-system get pod -lapp=istiod -o jsonpath='{.items[*].metadata.name}'); do \
         kubectl -n istio-system logs ${pod} \
     done
     {{< /text >}}
 
-1. If the patching failed, verify the RBAC configuration for Pilot:
+1. If the patching failed, verify the RBAC configuration for `Istiod`:
```
ericvn: From looking at the release announcements PR (https://github.com/istio/istio.io/pull/6705), it seems we are using `istiod`. I'm wondering if this should be updated to match. I know I asked this previously, just didn't see any official response on what it should be at this point.

ericvn: Alright, per some discussion we should be uppercasing Istiod when talking about the component and leaving lower case when talking about the deployment. So, can we just remove that last commit where I asked for the backticks and then we'll wait for the PR updating the dictionary with Istiod and then this should get through the checks.

Thanks John.

PR #6687, opened 2020-03-02
Minor update to selecting upgrade guide text

### LINKED ISSUE #6650: Fix upgrade docs
The https://preliminary.istio.io/docs/setup/upgrade/ page talks about two upgrades paths, one using istioctl and the other helm. The description to pick which one is confusing. In try to provide a better description, see #6687, there is discussion that there are even more upgrade paths and this is all under documented. That PR was thought to provide more confusion, so closed that PR and updating this issue to be a bigger effort in the future to provide documentation on all the paths, etc.

ericvn:
Fixes #6650

### DIFF
--- a/content/en/docs/setup/upgrade/_index.md
+++ b/content/en/docs/setup/upgrade/_index.md
@@ -1,6 +1,6 @@
 ---
 title: Upgrade
-description: Choose the upgrade guide that corresponds to the approach you previously used to install Istio.
+description: Upgrades must be done with the same approach used to install Istio. Select the upgrade guide that corresponds to the approach you previously used to install Istio.
 weight: 25
 keywords: [kubernetes,upgrading]
 ---


content/en/docs/setup/upgrade/_index.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Upgrade
-description: Choose the upgrade guide that corresponds to the approach you previously used to install Istio.
+description: Upgrades must be done with the same approach used to install Istio. Select the upgrade guide that corresponds to the approach you previously used to install Istio.
```
frankbu: I think this might be adding to the confusion since there are 4 "approaches" under `Install`, but only 2 (istioctl or helm) under `Upgrade`. We also are still missing a way for people to move off of helm, and on to istioctl or operator. I think it might be best to leave this as is for now, and wait to get some better overall upgrade story in place.

frankbu: @ostromart WDTY?

ericvn:
Since two other maintainers feel that this change causes more confusion than it fixes, I'll close this PR. Hopefully, we can fix the whole upgrade process and docs at some point in the future.

PR #6706, opened 2020-03-03
Remove deep dive mtls taks.

incfly:
Most deep dive docs are obsolete, not replacement exists. Some individual task if really needed, can be added somewhere else, such as inspect sds related key/cert. 

- verify citadel running, no citadel.
- key/cert path, no keycert path
- authn-tls check, no longer valid in peer authn context.

thus, too thin to be assembed as a page.


incfly:
I'm working on addressing dead links, let me know if you think differently for this PR @istio/wg-docs-maintainers

PR #6705, opened 2020-03-03
[Release-1.5] announcement

oaktowner:
Release announcement for 1.5

[ ] Configuration Infrastructure
[X ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #6711, opened 2020-03-03
Update architecture diagram to show istiod.

rcaballeromx:

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


elfinhe (commented):
Between Proxy and Istiod, should that be double arrows?

rcaballeromx:
> Between Proxy and Istiod, should that be double arrows?
For the sake of simplicity, I only included configuration, discovery, and certificates. While it is true that there is some control plane traffic going in the other direction, for example the istio-agent sends CSRs to istiod, those don't need to be depicted in the high-level architecture diagram. Too much detail might overwhelm readers.

frankbu (changes requested):
We can't replace this diagram without also rewriting all the surrounding text in the architecture doc. It currently describes all the components that you've removed from the diagram. In fact, this diagram seems pretty uninteresting now, since it doesn't even show what's in the control plane - just one box istiod == controlplane. Shouldn't we show boxes inside a big istiod box (maybe dotted lines) representing the services that it provides.

rcaballeromx:
@frankbu Yes, I could include all or some of the following services within:

- Network configuration
- Certificate Authority
- Authentication policies
- Authorization policies
- Sidecar injector
- API server configuration
- Observability configuration

Which services would you like me to include?

frankbu:
@rcaballeromx The architecture doc currently proceeds to describe components Pilot, Citadel, Mixer, and Gally, so at a high level, these correspond to the Istio features in the concept section (Traffic Managment, Security, and Observability) plus configuration, so maybe if the istiod box showed those 4, then the overall flow of the architecture doc could stay about the same.

Btw, what's the status of the security concepts rewrite? That should be consistent with what we update the architecture doc to say.

rcaballeromx:
@frankbu The security concept update is here: https://github.com/istio/istio.io/pull/6727
I'll update the diagram based on your comment, thanks.

rcaballeromx:
@frankbu I updated the diagram but I didn't include Mixer because that functionality is now in the Envoy proxies, AFAIK.
PTAL

frankbu:
@rcaballeromx I think this diagram (and architecture doc in general) should be consistent (or the same as) the istiod blog that @oaktowner said will be posted next week: https://github.com/istio/istio.io/pull/6705#discussion_r388319687. Do you know who the post author is and have you been working together on this diagram?

rcaballeromx:
@frankbu I've reviewed that blog post and helped create the diagrams it contains. The diagram there has to be kept as simple as possible and is slightly different than the diagram in this PR now. The goal is to introduce technical detail gradually to readers. I am fine having the old components in the diagram in this PR because the operations guide has an audience that is more familiar with Istio. We wouldn't want to have them in the blog post, which is read by a broader audience that includes folks less familiar with the project, its history, and the old architecture.

frankbu:
@rcaballeromx I'm not sure what to do. Short term, I guess we can merge this diagram, but I think the architecture doc needs to be reworked asap.

frankbu (approved):

rcaballeromx:
@frankbu Thanks, I agree that the doc needs to be reworked holistically sooner rather than later.

PR #6718, opened 2020-03-04
Add lint-fast target for local development

howardjohn:
This takes 30s and runs the spellcheck and some other quick linters.

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -85,6 +85,9 @@ clean:
 lint: clean_public build_nominify lint-copyright-banner lint-python lint-yaml lint-dockerfiles lint-scripts lint-sass lint-typescript lint-go
 	@scripts/lint_site.sh
 
+lint-fast:
+	@SKIP_LINK_CHECK=true scripts/lint_site.sh
+
 serve: gen
 	@hugo serve --baseURL "http://${ISTIO_SERVE_DOMAIN}:1313/" --bind 0.0.0.0 --disableFastRender
 
--- a/scripts/lint_site.sh
+++ b/scripts/lint_site.sh
@@ -140,8 +140,10 @@ find ./content/zh -type f \( -name '*.html' -o -name '*.md' \) -print0 | while I
     fi
 done
 
-if ! htmlproofer ./public --assume-extension --http-status-ignore "0" --check-html --check-external-hash --check-opengraph --timeframe 2d --storage-dir .htmlproofer --url-ignore "/archive.istio.io/,/localhost/,/github.com/istio/istio.io/edit/,/github.com/istio/istio/issues/new/choose/,/groups.google.com/forum/,/www.trulia.com/,/apporbit.com/,/www.mysql.com/,/www.oreilly.com/"; then
-    FAILED=1
+if [[ ${SKIP_LINK_CHECK:-} == "" ]]; then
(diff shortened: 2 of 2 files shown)


howardjohn:
@fpesce @adammil2000 maybe this can help devs lint locally

Makefile.core.mk
```diff
@@ -85,6 +85,9 @@ clean:
 lint: clean_public build_nominify lint-copyright-banner lint-python lint-yaml lint-dockerfiles lint-scripts lint-sass lint-typescript lint-go
 	@scripts/lint_site.sh
 
+lint-fast:
```
ericvn: Can we change the name to `lint-no-link` or something similar?

Removing the link checker will make it faster, but it's still not real fast. And `fast` doesn't really tell me what is happening.

howardjohn: Its sort of declarative vs imperative. If we make it `lint-no-link` then dev has to wonder "why do I not want link checking", we cannot add more "fast linters" in the future, its not actually accurate since it skips other things as well, etc. I think its also a fairly standard practice to have "fast" tests and "full" tests (maybe just in Google).

I am not the target audience for this so I can change it to whatever is desired, just my thought process here

rcaballeromx:
@howardjohn I don't think getting rid of the checks is the right solution. The link performs the checks because we want to make sure no one merges content with broken links, both external and internal. IMO, we need to improve the linter. For example, it could only check the links in files changed instead of the entire site. Adding targets for specific lint action could also help. For example `make lint-md` could just run the Markdown linter.
WDYT?

howardjohn:
> @howardjohn I don't think getting rid of the checks is the right solution. The link performs the checks because we want to make sure no one merges content with broken links, both external and internal. IMO, we need to improve the linter. For example, it could only check the links in files changed instead of the entire site. Adding targets for specific lint action could also help. For example make lint-md could just run the Markdown linter.
I am not removing the check in CI, this is for local development. I suspect no one is running the 30 min linter on their local machine, because its takes so long. The CI can catch the link checks.

Outside of this PR I think we should move the link checker to postsubmit. Its not perfect, but its much better than waiting 20min for the linter every time. Once we have the bandwidth to optimize the link check we can move it to presubmit if we really want. If you are interested I can set this up, but I think its orthogonal to this PR which is just for local

adammil2000:
Oh yes I run the local linter and it's a major problem as mine is running
even longer at about 45 minutes now. But I am about to stop using it
altogether locally.

I agree the link checker should become to post submit warning, especially
if we are not able to scan only the changes for a given PR. Large sites
don't block check ins on bad external links for good reasons, the two main
ones being that the link might be available later as planned, and the
internet being a live place constantly changing.

We should be checking the format of links however and if they are malformed
that's a good thing to block on.

The way larger sites handle this is there are occasional spikes where they
do link clean up in batches.

We should get Frank's opinion on this change, however because he is the
only Docs WG lead.

adammil2000:
And I forgot to mention all the times we get false errors on link checker,
like this PR <https://github.com/istio/istio.io/pull/6740> for example
where it complains this link is bad thinking the #rate bookmark is missing,
but it's clearly fine:

https://prometheus.io/docs/prometheus/latest/querying/functions/#rate

It's not reliable enough to be in our critical path.

ericvn:
I asked around and for **local** development, it seems people want more configurability in what is run. Targets to check only internal cross-references, another for only external links, and another for both are examples.

I've also heard on calls people think this PR is part of the effort to possibly change testing so links are testing in every PR and possibly daily. Not sure if that is really the case of not. But I would think we want targets that better spell out what needs to be done if that's the case.

fpesce:
Is there an agreement from Doc WG that this is a good start to iterate faster?
Or are you waiting for other lint target, and not satisfied with just lint-fast?

frankbu:
I'm OK with merging this PR as a first step, since it is addressing a pain point.

That said, I agree the actual link checking, and other linting things, could use a lot of improvement going forward.

Issue #6671, opened 2020-03-02
a couple errors in OpenShift Platform Setup docs 


jmazzitelli:
This issue is reporting a couple errors in this doc page:

https://preliminary.istio.io/docs/setup/platform-setup/openshift

I see two problems.

FIRST:

The docs say:

```
When removing your application, remove the NetworkAttachmentDefinition as follows.
$ oc -n <target-namespace> delete NetworkAttachmentDefinition istio-cni
```
However, the name of the type is incorrect. I get this error when I try it:

```
$ oc -n foo delete NetworkAttachmentDefinition istio-cni
error: the server doesn't have a resource type "NetworkAttachmentDefinition"
```

Using the type name `network-attachment-definition` works (I don't have a istio-cni resource at the moment, but notice is just says that - it does NOT say the server doesn't have this type):

```
$ oc -n foo delete network-attachment-definition istio-cni
Error from server (NotFound): network-attachment-definitions.k8s.cni.cncf.io "istio-cni" not found
```

You can see this better here:

```
$ oc get NetworkAttachmentDefinition
error: the server doesn't have a resource type "NetworkAttachmentDefinition"
$ oc get network-attachment-definition
No resources found.
```

So change that delete command to use "network-attachment-definition".

SECOND:

A few people have seen problems with the OpenShift route not working. The docs say this:

```
After installation is complete, expose an OpenShift route for the ingress gateway.
$ oc -n istio-system expose svc/istio-ingressgateway --port=80
```

However, this doesn't work. It DOES work if we replace "80" with "http2". So I think the docs should say the command is:

```
$ oc -n istio-system expose svc/istio-ingressgateway --port=http2
```

dgn:
@jmazzitelli I filed a PR. I could not reproduce the issues you mentioned with `oc expose --port=80`, but I changed it anyway, as both work.

PR #6721, opened 2020-03-04
Fix issues related to OpenShift setup

dgn:
Fixes https://github.com/istio/istio.io/issues/6671


ericvn:
I would remove the non-english changes in this PR. The translation team will move them at some point. Otherwise, the PR will be held out waiting on review from the translation teams.

Issue #6713, opened 2020-03-03
[1.5 Community Testing] Configuration Profiles guide needs to be updated to reflect the latest set of profiles


johnma14:
In this section of the docs: https://preliminary.istio.io/docs/setup/additional-setup/config-profiles/, it still refers to 5 profiles:
- default
- demo
- minimal
- sds
- remote

whereas when we do a `istioctl profile list`, it displays 6 profiles:
- default
- demo
- minimal
- seperate
- remote
- empty

I am pretty sure there was a bug opened for this issue but I cannot find it anymore. Opening this in case that got closed as fixed, since I am seeing this issue using the beta-5 build. Thanks.

shamsher31:
Thanks @johnma14 for catching this.

frankbu:
Duplicate of #6645

johnma14:
Thanks @frankbu.

PR #6723, opened 2020-03-04
Add updated profile to the list of istioctl profile list

shamsher31:
This is with respect to #6713 . The same list should be reflected in all docs.


frankbu:
This fixed in #6547

shamsher31:
@frankbu This is for `zh` and `pt-br` docs. I can see the docs for `en` is already updated.  But #6547 does not consist updates for `zh` and `pt-br`.

frankbu:
@shamsher31, sorry, I didn't notice that. I think it might be better, anyway, to let the individual translation teams update their versions after the en version is final. The en version may still not be in its final state.

shamsher31:
Sounds good. Thanks @frankbu

PR #6724, opened 2020-03-04
Updating istio/istio dependency

nmittler:


bin/init.sh
```diff
@@ -21,6 +21,9 @@ set -o pipefail
 export ISTIO_REMOTE=${ISTIO_REMOTE:-origin}
 export ISTIO_BRANCH=${ISTIO_BRANCH:-master}
 
+echo "Updating the istio/istio dependency to the latest ${ISTIO_BRANCH} branch ..."
+go get istio.io/istio@${ISTIO_BRANCH}
```
howardjohn: I don't think we want to do this automatically? Since then what is checked in and what is tested are two different things?

ericvn: I suspect that we will need to add something in the branching instructions, but I would think that currently we would want to be testing against the `release-1.5` branch (set when the release branch is created for istio/istio). Once we created the 1.5 doc branch we would revert to testing against `master` (release-1.5 istio.io branch continues to get against 1.5).

nmittler:
@howardjohn @ericvn PTAL

The script now downloads the same commit of Istio used in the go dependency.

Makefile.core.mk
```diff
@@ -26,15 +24,26 @@ GO ?= go
 GOARCH_LOCAL := $(TARGET_ARCH)
 GOOS_LOCAL := $(TARGET_OS)
 
+# ISTIO_IMAGE_VERSION stores the prefix used by default for the Docker images for Istio.
+# For example, a value of 1.6-dev will assume a default TAG value of 1.6-dev.<SHA>
+ISTIO_IMAGE_VERSION ?= 1.6-dev
```
howardjohn: I think its 1.6-alpha

howardjohn: for example: gcr.io/istio-testing/pilot:1.6-alpha.a11761a194325cc1af23e9ab53110a0369e69fde

Makefile.core.mk
```diff
@@ -125,12 +132,15 @@ foo2:
...
 .PHONY: init
 init:
+	@echo "ISTIO_SHA = ${ISTIO_SHA}"
+	@echo "HUB = ${HUB}"
+	@echo "TAG = ${TAG}"
 	bin/init.sh
 
 include tests/tests.mk
 
 # make lint-yaml seems to fail with pipefail, so remove now.
```
sdake: please file a bug and assign to me. Not super satisfied with the workaround in: https://github.com/istio/common-files/pull/200

nmittler: I don't seem able to create issues on https://github.com/istio/common-files/ ... permissions thing?

howardjohn:
you need to do it in the main repo

Makefile.core.mk
nmittler: Sure thing, @sdake  ... https://github.com/istio/istio/issues/21854

nmittler:
@howardjohn ah thanks :)

PR #6725, opened 2020-03-04
remove helm upgrade

linsun:
Since helm upgrade doesn't work and we don't recommend users to use it due to tiller security concerns.

https://github.com/istio/istio/issues/21648

### DIFF
--- a/content/en/docs/setup/upgrade/cni-helm-upgrade/index.md
+++ b/content/en/docs/setup/upgrade/cni-helm-upgrade/index.md
@@ -36,8 +36,6 @@ If you have installed or are planning to install [Istio CNI](/docs/setup/additio
 choose one of the following **mutually exclusive** options to check whether
 Istio CNI is already installed and to upgrade it:
 
-{{< tabset category-name="controlplaneupdate" >}}
-
 You can use Kubernetes’ rolling update mechanism to upgrade the Istio CNI components.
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/upgrade/cni-helm-upgrade/index.md
```diff
@@ -36,9 +36,6 @@ If you have installed or are planning to install [Istio CNI](/docs/setup/additio
 choose one of the following **mutually exclusive** options to check whether
```
ericvn: I don't think there are two options to chose from anymore.

content/en/docs/setup/upgrade/cni-helm-upgrade/index.md
```diff
@@ -63,57 +60,12 @@ This is suitable for cases where `kubectl apply` was used to deploy Istio CNI.
...
-        $ helm upgrade istio-cni install/kubernetes/helm/istio-cni --namespace kube-system
-        {{< /text >}}
-
-{{< /tab >}}
-{{< /tabset >}}
-
 ### Control plane upgrade
 
 Pilot, Galley, Policy, Telemetry and Sidecar injector.
 Choose one of the following **mutually exclusive** options
```
ericvn: Similar to above.

linsun:
@ericvn can we get this merged first?  I prefer such a statement in release note or upgrade note.

PR #6727, opened 2020-03-04
[Release-1.5] Update security concept page for Istio 1.5

rcaballeromx:
- Updated content reflects the architectural changes in 1.5
- Removed outdated content
- Updated diagrams
- Added information for new security features

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/concepts/security/index.md
```diff
@@ -673,12 +743,13 @@ metadata:
...
   namespace: default
 spec:
+  action: ALLOW
   rules:
   - {}
 {{< /text >}}
 
-The example below shows a simple policy `deny-all` which denies access to all workloads
-in the `admin` namespace.
+The example following example shows a simple  `deny-all` policy that denies any
```
diemtvu: Given we have "DENY" semantic now, should we rewrite this sentence to avoid confusion. E.g "The example following example shows a policy that does not allow any access to all workloads in the `admin` namespace."

May also show the recommendation and example to use DENY policy for such use case

+ @yangminzhu

rcaballeromx: @diemtvu Thanks for pointing it out. I also fixed the duplicated example.

yangminzhu: Yes, I think that makes sense:

We can change the current one to:
```
The following example shows a policy that does not allow any access to all workloads in the admin namespace. You can add more allow policies to allow extra access later.

{{< text yaml >}}
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
  name: allow-nothing
  namespace: admin
spec:
  action: ALLOW
  rules:
  - {}
{{< /text >}}

The following example shows a policy that explicitly denies any access to all workloads in the `database` namespace. You can't add allow policies to allow extra access because the deny policy takes precedence over allow policy, meaning it will deny the request before any allow policies.

{{< text yaml >}}
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
  name: deny-all
  namespace: database
spec:
  action: DENY
  rules:
  - {}
{{< /text >}}

Comment (yangminzhu) [reply content/en/docs/concepts/security/index.md]: @rcaballeromx I just made the comment, could you update with the latest change? thanks.

File: content/en/docs/concepts/security/index.md [hunk 2]
```diff
@@ -12,565 +12,543 @@ aliases:
...
-1. The Istio Agent sends the certificate received from Citadel and the private key to Envoy via the Envoy SDS API.
+{{< image width="75%"
+    link="./id-prov.svg"
+    caption="Identity Provision"
+    >}}
 
+Istio provisions identities through the secret discovery service (SDS) using the
+following flow:
+
+1. The CA offers a gRPC service to take CSR requests.
```
frankbu:
```suggestion
1. The CA offers a gRPC service to take [certificate signing requests](https://en.wikipedia.org/wiki/Certificate_signing_request) (CSRs).
```

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,543 @@ aliases:
...
+1. Envoy sends a certificate and key request via Envoy secret discovery
+   service (SDS) API.
+1. Upon receiving the SDS request, the Istio agent creates the private key
+   and CSR before sending the CSR with its credentials to the Istio CA for signing.
+1. The CAt validates the credentials carried in the CSR and signs the CSR to
```
frankbu:
```suggestion
1. The CA validates the credentials carried in the CSR and signs the CSR to
```

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,543 @@ aliases:
...
+{{< image width="75%"
+    link="./id-prov.svg"
+    caption="Identity Provision"
+    >}}
 
+Istio provisions identities through the secret discovery service (SDS) using the
+following flow:
+
+1. The CA offers a gRPC service to take CSR requests.
+1. Envoy sends a certificate and key request via Envoy secret discovery
```
frankbu:
```suggestion
1. Envoy sends a certificate and key request via the Envoy secret discovery
```

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,543 @@ aliases:
...
+[Security Tasks](/docs/tasks/security/) for detailed
+instructions to use the security features.
 
 ## High-level architecture
 
 Security in Istio involves multiple components:
 
-- **Citadel** for key and certificate management
+- A Certificate Authority (CA) for key and certificate management
+- Istiod distributes to the proxies:
```
frankbu: Istiod needs to be introduced before you can describe it's functions wrt security. That said, isn't Istiod too much an implementation detail for the concepts section? I thought one of the goals of the security concepts rewrite was to remove the architecture stuff (move it to the op guide) and focus instead on the security features. Like we did with traffic management.

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,543 @@ aliases:
...
 
-- **Sidecar and perimeter proxies** to implement secure communication between clients and servers
+    - [authentication policies](/docs/concepts/security/#authentication-policies)
+    - [authorization policies](/docs/concepts/security/#authorization-policies)
+    - [secure naming information](/docs/concepts/security/#secure-naming)
 
-- **Pilot** to distribute [authentication policies](/docs/concepts/security/#authentication-policies)
-  and [secure naming information](/docs/concepts/security/#secure-naming) to the proxies
+- Sidecar and perimeter proxies work as Policy Enforcement Points (PEPs) to
```
frankbu:
```suggestion
- Sidecar and perimeter proxies work as [Policy Enforcement Points](link to what is it) (PEPs) to
```

content/en/docs/concepts/security/index.md
frankbu:
```suggestion
- The configuration API server distributes to the proxies:
```

content/en/docs/concepts/security/index.md
```diff
@@ -633,17 +639,28 @@ spec:
...
 #### Value matching
 
-Exact match, prefix match, suffix match, and presence match are supported for most
-of the field with a few exceptions (e.g., the `key` field under the `when` section,
-the `ipBlocks` under the `source` section and the `ports` field under the `to` section only support exact match).
+Most fields in authorization policies support all the following matching
+schemas:
+
+- Exact match: exact string match.
+- Prefix match: A string with an ending `"*"`. For example, `"test.abc.*"`
```
frankbu:
```suggestion
- Prefix match: a string with an ending `"*"`. For example, `"test.abc.*"`
```

content/en/docs/concepts/security/index.md
```diff
@@ -633,17 +639,28 @@ spec:
...
+   matches `"test.abc.com"`, `"test.abc.com.cn"`, `"test.abc.org"`, etc.
+- Suffix match: A string with a starting `"*"`. For example, `"*.abc.com"`
```
frankbu:
```suggestion
- Suffix match: a string with a starting `"*"`. For example, `"*.abc.com"`
```

content/en/docs/concepts/security/index.md
```diff
@@ -633,17 +639,28 @@ spec:
...
-This means that the field can match any value, but it cannot be empty. Note that it is different from leaving a field unspecified, which means anything including empty.
+There are a few exceptions. For example, the following fields only support exact
+match:
 
-The following example policy allows access at paths with prefix `"/test/"` or suffix `"/info"`.
+- The `key` field under the `when` section
+- The `ipBlocks` under the `source` section
+- The `ports` field under the `to` section
+
+The following example policy allows access at paths with the  `/test/*` prefix
```
frankbu:
```suggestion
The following example policy allows access at paths with the `/test/*` prefix
```

content/en/docs/concepts/security/index.md
justinpettit: Due to time constraints, we didn't think we'd have time to do a full rewrite.  Instead we focused on updating the documentation to correctly reflect the 1.5 architecture--and remove a few pieces of extraneous information.  We hope to have a rewrite of the Security Concepts page for 1.6.

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,543 @@ aliases:
...
-The Istio security features provide strong identity, powerful policy, transparent TLS encryption, and authentication, authorization
-and audit (AAA) tools to protect your services and data. The goals of Istio security are:
-
-- **Security by default**: no changes needed for application code and infrastructure
+Breaking down a monolithic application into atomic services offers various
+benefits, including better agility, better scalability and better ability to
+reuse services. However, microservices also have particular security needs:
+
+- To defend against man-in-the-middle attacks, they need traffic encryption.
+- To provide flexible service access control, they need mutual TLS and
```
frankbu: Should mutual TLS be in the glossary, or maybe a link here (first use)?

content/en/docs/concepts/security/index.md
rcaballeromx: @yangminzhu The first example yaml  has an `ALLOW` action. How can a reader know that it is denying access?

content/en/docs/concepts/security/index.md
frankbu: Fair enough, but I still think we can try to keep this a little higher level. For example, my suggestion to use the generic Istio function (e.g., the config API server) instead of the implementation component name (Istiod) wherever possible.

rcaballeromx: @frankbu In general I agree with you. The issue with using the generic function is that the audience won't necessarily know anything about it. Using Istiod in this case is better, because readers will recognize it from their deployment.

content/en/docs/concepts/security/index.md
rcaballeromx: @frankbu I could use the following link: https://www.jerichosystems.com/technology/glossaryterms/policy_enforcement_point.html
WDYT?

content/en/docs/concepts/security/index.md
rcaballeromx: @frankbu We should add mutual TLS to the glossary as a separate PR and add the shortcode to all pages that need it.

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,543 @@ aliases:
...
+- Kubernetes: Kubernetes service account
+- GKE/GCE: GCP service account
+- GCP: GCP service account
+- AWS: AWS IAM user/role account
+- On-premises (non-Kubernetes): user account, custom service account,
+   service name, Istio service account, or GCP service account. The custom
+   service account refers to the existing service account just like the
+   identities that the customer's Identity Directory manages.
 
 ## PKI
```
frankbu:
```suggestion
## PKI (Public Key Infrastructure)
```

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,543 @@ aliases:
...
-across heterogeneous environments.
-
-Istio and SPIFFE share the same identity document: [SVID](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md) (SPIFFE Verifiable Identity Document).
-For example, in Kubernetes, the X.509 certificate has the URI field in the format of
-`spiffe://<domain>/ns/<namespace>/sa/<serviceaccount>`.
-This enables Istio services to establish and accept connections with other SPIFFE-compliant systems.
-
-Istio security and [SPIRE](https://spiffe.io/spire/), which is the implementation of SPIFFE, differ in the PKI implementation details.
-Istio provides a more comprehensive security solution, including authentication, authorization, and auditing.
+Identity is a fundamental concept of any security infrastructure. At the
```
frankbu: This section is inconsistent with the similar material in the glossary. In the glossary it talks about Istio assigning identifies to workloads but here it says they are given to service identities.
I suggest fixing this material to be consistent with the glossary version, and then shortening the glossary version to just have a short summary and a link to this section in the concepts.
https://preliminary.istio.io/docs/reference/glossary/#identity

content/en/docs/concepts/security/index.md
yangminzhu: I just fixed a minor issue in my comment. The `allow-nothing` doesn't have `rules` which means it will never match, which means it allows nothing and any request will be denied.

content/en/docs/concepts/security/index.md
rcaballeromx: @frankbu I am changing the content in the section to use `workload` instead of `service`. That said, fully aligning all content to use the workload terminology is outside the scope of this PR. That also applies to reworking the identity glossary entry.

content/en/docs/concepts/security/index.md
rcaballeromx: @yangminzhu Thanks for the correction. It is still confusing that readers need to use an allow policy to deny any requests. It is also not clear what the difference between `allow-nothing` and `deny-all` is. I can see how they both could be used depending on preference, but do we really want to introduce folks to both on the concept page? I would rather only keep the `deny-all` policy and keep things simple. @justinpettit WDYT?

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,546 @@ aliases:
...
+- Kubernetes: Kubernetes service account
+- GKE/GCE: GCP service account
+- GCP: GCP service account
+- AWS: AWS IAM user/role account
+- On-premises (non-Kubernetes): user account, custom service account,
+   service name, Istio service account, or GCP service account. The custom
+   service account refers to the existing service account just like the
+   identities that the customer's Identity Directory manages.
+
+## Public Key Infrastructure
```
frankbu:
```suggestion
## Public Key Infrastructure (PKI) {#pki}
```

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,546 @@ aliases:
...
+uses `.yaml` files to specify the policies. The policies are saved in the Istio
+configuration storage once deployed. The Istio controller watches the
+configuration storage.
+
+Upon any policy changes, the new policy is translated to the appropriate
+configuration telling the PEP how to perform the required authentication
+mechanisms. The control plane may fetch the public key and attach it to the
+configuration for JWT validation. Alternatively, Istiod provides the path to the
+keys and certificates the Istio system manages and installs them to the
+application pod for mutual TLS. You can find more info in the [PKI section](/docs/concepts/security/#public-key-infrastructure).
```
frankbu:
```suggestion
application pod for mutual TLS. You can find more info in the [PKI section](/docs/concepts/security/#pki).
```

content/en/docs/concepts/security/index.md
```diff
@@ -12,565 +12,546 @@ aliases:
...
+- GCP: GCP service account
+- AWS: AWS IAM user/role account
+- On-premises (non-Kubernetes): user account, custom service account,
+   service name, Istio service account, or GCP service account. The custom
+   service account refers to the existing service account just like the
+   identities that the customer's Identity Directory manages.
+
+## Public Key Infrastructure
+
+The Istio Public Key Infrastructure (PKI) securely provisions strong identities
```
frankbu:
```suggestion
The Istio PKI securely provisions strong identities
```

content/en/docs/concepts/security/index.md
frankbu: I actually wonder if the parens should be reversed, since PKI is the commonly used term??
```suggestion
## PKI (Public Key Infrastructure) {#pki}
```

content/en/docs/concepts/security/index.md
justinpettit: I agree that we're adding complexity to a document that should probably be higher-level.  @yangminzhu, can we make a note of this and find a more appropriate location post-1.5?

yangminzhu: @rcaballeromx the difference between `allow-nothing` and `deny-all` is the interaction between other policies. 

- If you have `allow-nothing`, it first restricts the permission to zero and you can incrementally add more allow policies to grant more permission. 

- If you have `deny-all`, all requests will be denied even if you add other allow policies, you need to delete the `deny-all` policy to allow other access.

We can keep the `deny-all` but please replace the current example with the one from my comment. That's a better way to express `deny-all` semantics.

yangminzhu: @rcaballeromx This comment is not addressed in the merged PR, Rigs, would you mind to send out a follow PR to fix this? Thanks.

PR #6732, opened 2020-03-04
make update_ref_docs against release_1.5

ericvn:
Since the nightly job that does this work was running against _master_, as soon as the sidecar-injector command was removed, that job's output (#6419) failed the lint test and we have not seen any updates.

This is the output of what that the script should have generated (I did run locally) if it was run against _release-1.5_.


ericvn:
Checking with @richardwxn as I removed a file he added in #6600. If we automate going forward, we probably need to do something to get things in sync.

ericvn:
The reason the file is not picked up when doing the release-1.5 scrapping is that `https://github.com/istio/api/blob/release-1.5/operator/v1alpha1/istio.operator.v1alpha1.pb.html` is missing the location tag. It is in the _master_ version of that file.

ericvn:
Slacked with @richardwxn and appears a cherrypick was missing for the mentioned file, and also for the file I had added `content/en/docs/reference/config/istio.operator.v1alpha12.pb/index.html` to remove it, so will manually remove that from this PR as well (in anticipation of the two cherrypicks).

PR #6533, opened 2020-02-22
fix(observability docs): restructure docs for v2 telemetry (and remove problematic Mixer refs)

douglas-reid:
This PR is a proposal for removing Mixer-specific content from a couple of key places, while preserving the content where needed. This is part of the work involved in making the telemetry v2 work on-by-default in 1.5.

In particular, the concept guides for Observability and "What is Istio?" have mentions of Mixer removed. The Mixer-specific Tasks are moved into a sub-category of Observability Tasks, as Mixer is merely deprecated but not removed. The 1.3 Alpha Task on enabling v2 telemetry is also removed, as it is no longer relevant. Finally, some of the old links are updated, along with Mixer-focused bits of the microservices example.

Notably, new content is not added in this PR. This includes any work to "fix" the existing Mixer-specific Tasks by including steps to enable Mixer, as well as any new content to discuss the v2 architecture.

Also not addressed are any updates to the "Ops" guides on troubleshooting metrics and on deployment architecture.

Signed-off-by: Douglas Reid <dougreid@google.com>

PR #6742, opened 2020-03-05
Issue with blog/2020/tradewinds-2020/index.md: missing a markdown link symbol

zliu9125:
In ` Sleeker, smoother and faster` part of this blog,  an underling link misses `(` symbol.

### DIFF
--- a/content/en/blog/2020/tradewinds-2020/index.md
+++ b/content/en/blog/2020/tradewinds-2020/index.md
@@ -50,7 +50,7 @@ and [authorization](/docs/concepts/security/#authorization) policies, which
 allow you to control workload-to-workload and end-user-to-workload authorization
 directly in the proxy. Common monitoring use cases have already moved into the
 proxy too - we have
-[introduced in-proxy support]/docs/ops/configuration/telemetry/in-proxy-service-telemetry/)
+[introduced in-proxy support](/docs/ops/configuration/telemetry/in-proxy-service-telemetry/)
 for sending telemetry to Prometheus and Stackdriver.
 
 Our benchmarking shows that the new telemetry model reduces our latency


ericvn:
Not the failing lint is for a file (content/en/docs/ops/configuration/telemetry/in-proxy-service-telemetry/index.md) removed in https://github.com/istio/istio.io/pull/6533 by @douglas-reid

douglas-reid:
@ericvn that file was removed because in-proxy telemetry is now the default (meaning there is no need to enable it) and because that file was 1.4 specific. It has been overcome by events and would increase confusion by persisting.

douglas-reid:
@ericvn @mandarjog @howardjohn we need to also fix the paragraph that reads:

> Installing Istio 1.5 with the 'preview' configuration profile won't install
Mixer. If you upgrade from a previous release, or install the 'default' profile,
we still keep Mixer around, to be safe. When using Prometheus or Stackdriver for
metrics, we recommend you try out the new mode and see how much your performance
improves.

Here is the output of `istioctl profile dump default`:

```
apiVersion: operator.istio.io/v1alpha1
kind: IstioOperator
spec:
  addonComponents:
...
  components:
...
    policy:
      enabled: false
...
    telemetry:
      enabled: false
...
  profile: default
  tag: 1.5.0-beta.3
  values:
...
    global:
...
      disablePolicyChecks: true
...
...
    telemetry:
      enabled: true
      v1:
        enabled: false
      v2:
        enabled: true
...
```

As you can see, the `default` profile does not select for Mixer-based telemetry.

ericvn:
And there is mention of the `preview` profile which I had questioned in the original PR (got approved before I verified things were fixed)

richardwxn:
@douglas-reid Yep the default profile would remove mixer deployment as well, not just functionality, maybe we should link to the doc of how to turn mixer back on instead

douglas-reid:
@richardwxn i'd be fine with that. remove this link altogether and add a link elsewhere for the Mixer stuff.

frankbu:
Fixed in #6792

PR #6737, opened 2020-03-04
Early draft for VM updates

costinm:
Would appreciate if someone can take over the change and translate it to english...

PR #6743, opened 2020-03-05
[Release-1.5] Update VM expansion doc

irisdingbj:
Based on @costinm 's PR: #6737


irisdingbj:
Based on @costinm 's PR: https://github.com/istio/istio.io/pull/6737

adammil2000:
I see the gcloud CLI used in the commands. It would be good to add a note to the beginning of this procedure to indicate that these instructions assume you're using GKE.

sdake:
@irisdingbj great work on this PR! Would you mind generalizing this PR to any environment such that we don't end up with variants for each cloud? I can see where this type of workflow would be helpful for GCP users, but also users of other clouds such as https://istio.io/docs/setup/platform-setup/. This has happened before, and dealing with the divergence of the multicluster instructions into 3 different versions (as an example that occurred in the past) makes documentation maintenance extremely difficult.

irisdingbj:
> @irisdingbj great work on this PR! Would you mind generalizing this PR to any environment such that we don't end up with variants for each cloud? I can see where this type of workflow would be helpful for GCP users, but also users of other clouds such as https://istio.io/docs/setup/platform-setup/. This has happened before, and dealing with the divergence of the multicluster instructions into 3 different versions (as an example that occurred in the past) makes documentation maintenance extremely difficult.

sure.

frankbu:
Good spotting @sdake, this PR should not have been merged! None of the docs on istio.io can be platform specific, so this needs to be fixed or removed ASAP!

PR #6762, opened 2020-03-05
fix label name for telemetry v2

mandarjog:


douglas-reid:
Do we need to add `Proxy-wasm` to the dictionary?

mandarjog:
@frankbu @sdake simple change PTAL.

PR #6764, opened 2020-03-05
[Release-1.5] Add telemetry v2 with Wasm runtime as experiment feature

bianpengyuan:


content/en/docs/reference/config/telemetry/telemetry_v2_with_wasm/index.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Telemetry V2 with Wasm runtime (Experimental)
+description: How to enable Telemetry V2 with Wasm runtime (experimental).
+weight: 60
+---
+
+In Istio 1.5, by default Telemetry V2 is enabled as compiled in Istio proxy filters. The same filters are also compiled to WebAssembly (Wasm) modules and shipped with Istio proxy. To enable Telemetry V2 with Wasm runtime, apply the following configuration, which is a drop-in replacement for the default Telemetry V2 installation:
+
+{{< text bash >}}
+$ kubectl apply -f https://raw.githubusercontent.com/istio/proxy/release-1.5/extensions/stats/testdata/telemetry-v2-wasm.yaml
```
gargnupur: do we need to delete the existing one's that get installed by default?

bianpengyuan: No they are named the same, so this will just config the resource, not create them.

bianpengyuan:
`lint_istio.io` should not run for 1.5 PR. There is already a `lint_istio.io_release-1.5` which already passed. I think it is because I changed base of this PR from master to release-1.5 and somehow that job sticks...

ericvn:
PR's still need to go into _master_ first and then get cherry-picked to _release-1.5_.

bianpengyuan:
@ericvn Thanks! here is the master doc: https://github.com/istio/istio.io/pull/6793

ericvn:
If it's OK, I'll close this PR, and we'll label the other one to be cherry-picked as that follows the process. I did note the other PR adds a block of text as well.

bianpengyuan:
The problem is the istioctl installation mechanism does not exist in 1.5. It is only in master. That is why master PR is different from this one. So it cannot be manually cherry-picked.

ericvn:
OK. Thanks for the update.

PR #6766, opened 2020-03-05
Fix failing link in pipeline

ericvn:


ericvn:
Closing this PR for now, since the line/link I was removing was changed in another PR to a valid value. Will revisit later if we should remove the line.

PR #6768, opened 2020-03-06
fix typo in the path for redis config

adarshaj:


ericvn:
Pushing the change into all 3 language versions is not typically advised. The non-english versions will get updated by the translation team later. But you can leave all of them in, it just may take longer as you'll need additional approval from the language teams.

adarshaj:
> Pushing the change into all 3 language versions is not typically advised. The non-english versions will get updated by the translation team later. But you can leave all of them in, it just may take longer as you'll need additional approval from the language teams.

Ah I see, however, the change I have is in the code part, which is common (and currently wrong) across all the languages? Probably we should have a cleaner way to include code segments which are not subjected to human language translation in a separate file and then include the snippet. I would rather wait for all the language teams to approve rather than have wrong code in some languages.

@ericvn  - It would be great if you can @mention the language maintainers so they can approve this change (which isn't really specific to any language)

frankbu:
@adarshaj Using snippets just makes the document harder to write and work with. Also, expecting doc updates to be done in every translated version will not scale as we have more and more translations. So, I think the right way for this to work is for the English version to be the canonical version that gets updated and the translation teams should try to automate the synchronization as much as possible. For example, maybe they could get notifications whenever the English version is updated which includes a partial PR that already includes changes to text fragments. Then they only need to deal with diffs in the actual translatable parts, if any.

adarshaj:
Okay, I had assumed that code snippets don't need to be translated across languages. I guess the way I saw was analogous to seeing code snippets essentially similar to image, so one could just include link to snippet as they would for image. (ofcourse, this would have required some template processing language that actually substitutes snippet link to actual code)

Having said that, let me know if you prefer this pr to just include changes for English language?

frankbu:
@adarshaj I agree we need a better process for updating common content in all the translated docs, but I think it would be best for now if you just change the English version and then we can approve and merge it right away.

That said, we do already have a [snippet include mechanism](https://istio.io/about/contribute/code-blocks/#snippets) that we could use as you suggest, but the downside is that it makes the md document a lot harder to edit and review. The thing we need to remember is that the English version is the canonical version of the doc (the others are just conversions that could theoretically be partially, if not eventually fully, generated), so we don't want to do things that make the English version harder to write, if we can avoid it.

One idea that comes to mind is the possibility of leaving the text in the English documents and using snippet import only in the translated versions. This would require change in the snippet design to allow the English md file to be the input file from the snippets (instead of just txt files with special comments).

adarshaj:
@rootsongjc / @yuxiaobo96 - request to pls approve this pr for zh translation. sorry for tagging, its a trivial change and has been long since I opened.

ericvn:
@adarshaj Since the deploy/netlify test didn't pass, and I don't know of anyway to kick it again (retest doesn't do it), I think you'll have to push something to cause the test to be run again.

PR #6776, opened 2020-03-06
zh-translation: about contribute build

helight:
ref:https://github.com/servicemesher/istio-official-translation/issues/1767


rootsongjc:
请注意 issue 标题的名字前面应该写为 `zh-translations:`

PR #6785, opened 2020-03-06
Fix operator image

dgn:


dgn:
Note the failed tests there are from when I accidentally open a PR to master. they won't go away, but there are the release-1.5 equivalents (which are passing)

PR #6793, opened 2020-03-06
Add telemetry v2 with Wasm runtime as experiment feature

bianpengyuan:


ericvn:
Per @bianpengyuan in #6764, this PR has instructions (Alternatively, set the following...) which can't be cherry-picked so using other PR as a manual cherry-pick to 1.5 (this minus the section).

PR #6796, opened 2020-03-06
networking issue section update: disclaimer about auto mtls and remove authn check.

incfly:
I'm taking some balance.

- we want to promote beta policy for now on, which does not work with auth tls check.
- customers may still not enable auto mtls yet, so the entries can still live there for a while.

### DIFF
--- a/content/en/docs/ops/common-problems/network-issues/index.md
+++ b/content/en/docs/ops/common-problems/network-issues/index.md
@@ -65,6 +65,8 @@ order of seconds.
 
 ## 503 errors after setting destination rule
 
+You should only see this error when you disable [automatic mutual](docs/tasks/security/authentication/authn-policy/#auto-mutual-tls) TLS during install.
+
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/common-problems/network-issues/index.md
```diff
@@ -65,6 +65,8 @@ order of seconds.
 
 ## 503 errors after setting destination rule
 
+You should only see this error when you disable [automatic mutual](docs/tasks/security/authentication/authn-policy/#auto-mutual-tls) TLS during install.
```
frankbu:
```suggestion
{{< tip >}}
You should only see this error if you disabled [automatic mutual TLS](docs/tasks/security/authentication/authn-policy/#auto-mutual-tls) during install.
{{< /tip >}}
```

incfly: thanks! i was thinking there indeed some cutomized tag can be used for highlight, but can't rembmer the name!

incfly:
suggestion applied.

PR #6804, opened 2020-03-06
authz: use shorter title in tasks

yangminzhu:
@liminw @adammil2000  @rcaballeromx I feel the current task title is a little bit too long, probably remove the `authorization` from the title as it's already under the authorization section?


liminw:
I think the current title is fine. It give more clarity if a user just looks at a single page.

PR #6819, opened 2020-03-08
replace Citadel with Istio CA

tigran-a:
Since Citadel is now built in into istiod, the certificate authority should be called Istio CA


tigran-a:
|@googlebot I signed it!|

PR #6826, opened 2020-03-09
Fix links for removal of helm installation directory

ericvn:
This PR will build on the work in https://github.com/istio/istio.io/pull/6458


ericvn:
FYI. #6847 has some updates to the setup doc as well.

ericvn:
There was discussion in last week's TOC about @ostromart and @adammil2000 updating the helm install instructions for 1.6 to include using the new `manifests` files. @howardjohn had mentioned simply removing the files, which he did in #6458, but in light of the files being reworked, I will leave them. They aren't correct right now, but Martin and Adam will fix them in the future.

PR #6829, opened 2020-03-09
cacerts secret to replace with istio-ca-secret

tigran-a:
Currently, 1.5.0 istiod is looking for istio-ca-secret, while in 1.4.x it was cacerts.


ayj:
istiod is still looking for cacerts. istio-ca-secret is the auto-generated self-signed CA if cacerts is not present. See https://github.com/istio/istio/blob/6fab76ef7ba34a9e551e0313b67371064cc3a522/pilot/pkg/bootstrap/istio_ca.go#L51 and https://github.com/istio/istio/blob/2c30e28e7b0d9f29c1323ca4d271c7044ce07e70/manifests/istio-control/istio-discovery/templates/deployment.yaml#L283. This PR should be reverted.

ericvn:
Removed cherrypick label since this issuing reverted.

PR #6831, opened 2020-03-09
Add missing quote

rytswd:
Noticed a really minor error with missing quotes. There were some other parts that could perhaps benefit from having quotes, but kept this change as simple as it could be.


frankbu:
This is a generated file. It needs to be fixed in the original source in istio/api.

rytswd:
Ah thanks for the info, I found the same part in istio/api. It seems the number of PRs there on documentation change is fewer (perhaps due to the compilation requirement), so I may actually leave it as is. It's a single character difference anyways, and probably won't benefit the community much.

PR #6836, opened 2020-03-10
Revert "cacerts secret to replace with istio-ca-secret"

tigran-a:
https://github.com/istio/istio/blob/6fab76ef7ba34a9e551e0313b67371064cc3a522/pilot/pkg/bootstrap/istio_ca.go#L51


ericvn:
Since the original never was cherry-picked to 1.5, not adding the label here.

PR #6847, opened 2020-03-11
Restructured doc to remove Tiller references

jpapejr:
With the merge of PR #6725 we've removed the upgrade docs for those managing Istio via Helm with Tiller. This PR removes all reference to the Helm/Tiller install on the Helm install path so that all we provide is the Helm templating method for those users not using `istioctl` or the in-cluster controller.


content/en/docs/setup/install/helm/index.md
```diff
@@ -68,10 +63,9 @@ instead with the flags `--set gateways.istio-ingressgateway.type=NodePort`
 appended to the end of the Helm instructions in the installation steps below.
 {{< /tip >}}
 
-### Option 1: Install with Helm via `helm template`
-
-Choose this option if your cluster doesn't have [Tiller](https://helm.sh/docs/topics/architecture/#components)
-deployed and you don't want to install it.
+{{< tip >}}
+The Helm installation method previously detailed here that used the [Tiller](https://helm.sh/docs/topics/architecture/#components) component has been removed. The preferred install path is to use `istioctl` as documented in [Installing with {{< istioctl >}}](/docs/setup/install/istioctl/). If you still desire to use Helm, you'll need to leverage the `helm template` method detailed below.
```
elysdir:
```suggestion
This document used to also describe a Helm installation method that used the [Tiller](https://helm.sh/docs/topics/architecture/#components) component. That installation method is no longer available. Instead, we recommend using `istioctl`, as documented in [Installing with {{< istioctl >}}](/docs/setup/install/istioctl/). If you want to use Helm, then you need to use the `helm template` method described below.
```

...Or something like that. My concerns here are:

* The phrase "The Helm installation method previously detailed here" is slightly hard to parse. But only slightly--if you want to keep that phrase as-is, that's fine with me.

* The words "desire" and "leverage" are fancier words than are needed here--I recommend simplifying the language for ease of reading.

elysdir: If I'm understanding right, it sounds to me like this paragraph is saying that it's best to go use a different process than the one described on this page. True?

If so, then I would recommend pulling that recommendation out of this tip and putting it in a new paragraph right below the "Installation instructions" heading. People often don't read tips, so I suspect that many people who come to this section won't see the line that says to use istioctl.

content/en/docs/setup/install/helm/index.md
```diff
@@ -283,7 +175,7 @@ $ helm install install/kubernetes/helm/istio --name istio --namespace istio-syst
 
 ## Uninstall
 
-- If you installed Istio using the `helm template` command, uninstall with these commands:
+- The `helm template` command is also used in uninstalling Istio. Uninstall with these commands:
```
elysdir: We recommend avoiding passive voice ("is also used"). (https://developers.google.com/style/voice)

And it's unclear in this context what the word "also" refers to.

So I would say something like "You can use the `helm template` command to uninstall Istio."

content/en/docs/setup/install/helm/index.md
jpapejr: Resolved with https://github.com/istio/istio.io/pull/6847/commits/9381d044356a14493fa33949f50f5c90584581e0

content/en/docs/setup/install/helm/index.md
jpapejr: Resolved with https://github.com/istio/istio.io/pull/6847/commits/9381d044356a14493fa33949f50f5c90584581e0

jpapejr:
Sorry about the extra commits/merge artifacts.. I narfled the Garthak in my VSCode diff viewer. Seems good now.

jpapejr:
Closing this PR since it got messy. I'll open another for this change.

PR #6850, opened 2020-03-11
Fix exposing container port CONTAINER_OPTIONS on Docker when running on Mac

christian-posta:
export CONTAINER_OPTIONS make env variable so run.sh script can see it; running on Mac/Bash/Docker for Mac


jpapejr:
I think this was done yesterday via https://github.com/istio/istio.io/pull/6862

frankbu:
Closing because I believe the underlying problem has been fixed in #6862.
@christian-posta Please reopen if you believe this PR is also still needed.

PR #6871, opened 2020-03-13
Update Istio DNS certificate document for Istio 1.5

lei-tang:
Please provide a description for what this PR is for: update Istio DNS certificate document for Istio 1.5.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/security/dns-cert/index.md
+++ b/content/en/docs/tasks/security/dns-cert/index.md
@@ -5,23 +5,14 @@ weight: 90
 keywords: [security,certificate]
 ---
 
-In Istio versions before 1.5, by default, the DNS certificates used by the webhooks of Galley and the sidecar
-injector are by default provisioned and managed by Citadel, which is a large component
-that maintains its own signing key and also acts as a CA for Istio.
-In certain deployments of Istio versions before 1.5, you may want to use your own certificate authority
-instead of Citadel. In those cases, Citadel ends up being used strictly for
-its DNS certificate provisioning functionality. Rather than having to deploy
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/security/dns-cert/index.md
```diff
@@ -6,14 +6,13 @@ keywords: [security,certificate]
 ---
 
 In Istio versions before 1.5, by default, the DNS certificates used by the webhooks of Galley and the sidecar
```
frankbu: This whole overview doesn't make much sense in the 1.5 docs. If it only applies to "certain deployments of Istio before 1.5" it shoudn't be in 1.5 docs. Also, you can't say Chiron is lighter weight than Citadel, because there is no Citadel in 1.5. This needs to explain why this task is interesting in 1.5. 

Is there any reason to do this in 1.5? Maybe something like this???
```
This task shows how to provision and manage DNS certificates
using [Chiron](link TBD), a lightweight component linked with Istiod that signs certificates
using the Kubernetes CA APIs without maintaining its own private key.
Using this feature has the following advantages:

* Unlike Istiod, this feature doesn't require maintaining a private signing key, which enhances security.

* Simplified root certificate distribution to TLS clients. Clients no longer need to wait for Istiod to generate and distribute its CA certificate.
```

lei-tang: Thanks for the suggestion! The overview has been revised accordingly.

PR #6872, opened 2020-03-13
Update home page links

### LINKED ISSUE #6774: Deprecated docs linked from the home page
This is probably known, but it seems reasonable to call it out explicitly:

If you go to the home page, you see these panels as links:

Screen Shot 2020-03-05 at 8 15 01 PM

But if you click on either the "Control" or "Observe" links, you are send to a doc screen that has says the feature is deprecated - right in the titla:

Screen Shot 2020-03-05 at 8 17 31 PM

This seems a bit... problematic from a "truth in advertising" point of view.:)

shamsher31:
This is to fix #6774

### DIFF
--- a/content/en/_index.html
+++ b/content/en/_index.html
@@ -73,7 +73,7 @@ description: Connect, secure, control, and observe services.
         </div>
 
         <div id="panel3" class="panel">
-            <a href="/docs/tasks/policy-enforcement/enabling-policy/">
+            <a href="/docs/reference/config/">
                 <div class="panel-img-top">
                     {{< inline_image "landing/policy-enforcement.svg" >}}
                 </div>
--- a/content/en/_index.html
+++ b/content/en/_index.html
@@ -73,7 +73,7 @@ description: Connect, secure, control, and observe services.
         </div>
 
         <div id="panel3" class="panel">
-            <a href="/docs/reference/config/policy-and-telemetry/">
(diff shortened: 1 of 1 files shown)


content/en/_index.html
```diff
@@ -73,7 +73,7 @@ <h5 class="panel-title">Secure</h5>
         </div>
 
         <div id="panel3" class="panel">
-            <a href="/docs/reference/config/policy-and-telemetry/">
+            <a href="/docs/tasks/policy-enforcement/enabling-policy/">
```
shamsher31: @ericvn As per the current depreciation warning, 

> Mixer is deprecated. The functionality provided by Mixer is being moved into the Envoy proxies.

I can't find the appropriate link for the `Control` section of the home page. But as per [this page ](https://istio.io/docs/tasks/policy-enforcement/enabling-policy/), I can see the reference links in the warning section at the top. 

So I have added a link to [policy enforcement](https://istio.io/docs/tasks/policy-enforcement/enabling-policy/) page, but not sure if this is the right link.

frankbu: I think that the home page needs to be reworked, but for now I suggest pointing to the higher level config page, since the control functions are distributed into multiple different places now:
```suggestion
            <a href="/docs/reference/config/">
```

shamsher31: Thanks @frankbu

rvennam:
Can this be back-ported to 1.5 please?

PR #6882, opened 2020-03-14
Gke private firewall port

so-jelly:
For GKE Private clusters, update firewall rule to allow access to the Pilot discovery webhook.

### DIFF
--- a/content/en/docs/setup/platform-setup/gke/index.md
+++ b/content/en/docs/setup/platform-setup/gke/index.md
@@ -36,6 +36,27 @@ Follow these instructions to prepare a GKE cluster for Istio.
     the `gcloud container clusters create` command.
     {{< /warning >}}
 
+    {{< warning >}}
+    For private GKE clusters, the firewall rule that is created does not open the port (15017) needed by the validation webhook for master access.
+    
+    To review this firewall rule for master access:
+
+        {{< text bash >}}
+        $ gcloud compute firewall-rules list \
+          --filter="name~gke-<cluster-name>-[0-9a-z]\*-master" \
+          --format="value(name)"
+        {{< /text >}}
+    
+    To replace the existing rule and allow master access:
+
+        {{< text bash >}}
+        $ gcloud compute firewall-rules update \
+          <firewall-rule> \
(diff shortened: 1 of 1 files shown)


so-jelly:
@Rachael-Graham  @frankbu @howardjohn sorry for the spam... couldn't manage to rebase the previous branch..  I blame the apostrophe in my name, but it was more likely user error.    I spent some time making sure the gcloud command actual worked and it looked better as well.   Cheers!

content/en/docs/setup/platform-setup/gke/index.md
```diff
@@ -36,6 +36,24 @@ Follow these instructions to prepare a GKE cluster for Istio.
...
+
+        {{< text bash >}}
+        $ gcloud compute firewall-rules list --filter="name~gke-<cluster-name>-[0-9a-z]*-master"
+        {{< /text >}}
+
+    To replace the existing rule and allow master access:
+
+        {{< text bash >}}
+        $ gcloud compute firewall-rules update <firewall-rule-name> --allow tcp:10250,tcp:443,tcp:15017
+        {{< /text >}}
```
frankbu:
```suggestion
        {{< /text >}}

```

content/en/docs/setup/platform-setup/gke/index.md
```diff
@@ -36,6 +36,25 @@ Follow these instructions to prepare a GKE cluster for Istio.
     the `gcloud container clusters create` command.
     {{< /warning >}}
 
+    {{< warning >}}
```
frankbu: So it seems you are running into a generator bug. My guess is that the problem is caused by 3 levels of nesting: 1. numbered list item, 2. warning, 3. text section. If you have time and want to try to track down the problem, that would be great.

Otherwise, looking at your PR, I think the warning could be moved to a better place, anyway. You could probably work around the generator bug by moving the warning to the top or bottom of the doc, i.e., before of after the list of instructions (unindent to col 0), or just add it as another step (instead of a warning), something like:
```
1. **For private GKE clusters only**, an automatically created ...
```

PR #6886, opened 2020-03-16
Remove KubeCon EU event, fix KubeSphere.io links, ignore 429s.

ericvn:


ericvn:
Force another push since the netlify had passed but wasn't showing successfully in the tests.

PR #6919, opened 2020-03-19
Add instructions for gateway customization

ostromart:
Based on fixes going into 1.5.1.

### DIFF
--- a/content/en/docs/setup/install/istioctl/index.md
+++ b/content/en/docs/setup/install/istioctl/index.md
@@ -19,8 +19,7 @@ Using these instructions, you can select any one of Istio's built-in
 and then further customize the configuration for your specific needs.
 
 Full customization of the installation can be done through the
-[`IstioOperator` API](/docs/reference/config/istio.operator.v1alpha1/)
-(also see [protobuf format](https://github.com/istio/api/blob/release-1.5/operator/v1alpha1/operator.proto)).
+[`IstioOperator` API](/docs/reference/config/istio.operator.v1alpha1/).
 
 ## Prerequisites
 
--- a/content/en/docs/setup/install/istioctl/index.md
+++ b/content/en/docs/setup/install/istioctl/index.md
@@ -18,6 +18,10 @@ Using these instructions, you can select any one of Istio's built-in
 [configuration profiles](/docs/setup/additional-setup/config-profiles/)
 and then further customize the configuration for your specific needs.
 
+Full customization of the installation can be done through the
+[`IstioOperator` API](/docs/reference/config/istio.operator.v1alpha1/)
+(also see [protobuf format](https://github.com/istio/api/blob/release-1.5/operator/v1alpha1/operator.proto)).
+
 ## Prerequisites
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/install/istioctl/index.md
```diff
@@ -18,6 +18,10 @@ Using these instructions, you can select any one of Istio's built-in
 [configuration profiles](/docs/setup/additional-setup/config-profiles/)
 and then further customize the configuration for your specific needs.
 
+Full customization of the installation can be done through the
+[`IstioOperator` API](/docs/reference/config/istio.operator.v1alpha1/)
+(also see [protobuf format](https://github.com/istio/api/blob/release-1.5/operator/v1alpha1/operator.proto)).
```
frankbu: Why should users also be told to look at the proto? Is there stuff missing in the API doc?

content/en/docs/setup/install/istioctl/index.md
```diff
@@ -376,6 +381,68 @@ namespaces:
...
+            ports:
+            - port: 8060
+              targetPort: 8060
+              name: tcp-citadel-grpc-tls
+            - port: 5353
+              name: tcp-dns
+{{< /text >}}
+
+Note that Helm values (`spec.values.gateways.istio-ingressgateway/egressgateway`) are shared by all ingress/egress
+gateways. If these must be customized per gateway, it is recommended to use a separate IstioOperator CR to generate
```
frankbu:
```suggestion
gateways. If these must be customized per gateway, it is recommended to use a separate `IstioOperator` CR to generate
```

frankbu:
I'm assuming we don't want to merge this PR until next week when 1.5.1 is out. Is that right?

content/en/docs/setup/install/istioctl/index.md
ostromart: No but the API doc unhelpfully reorders the proto alphabetically which makes it much harder to understand.

ostromart:
Hi Frank, yes this should merge together with the 1.5.1 release.

content/en/docs/setup/install/istioctl/index.md
frankbu: John recently sent a PR to change the doc generator to use the proto order, so maybe it's not needed if that's fixed? https://github.com/istio/tools/pull/808

ostromart: It's still sorted in 1.5 docs but I assume this is temporary until the PR takes effect - I've removed the proto link.

frankbu: Actually, this brings up a good question. Are istio/tools from the master branch used for release-1.5 docs, or does the tools PR need to be cherrypicked?
^^^ @howardjohn ?

howardjohn: Its pinned per version. Lets get it end to end into the doc before we try to cherrypick though, to make sure it looks a reasonable way

PR #6931, opened 2020-03-20
Fix typo for pod

shamsher31:
`kubectl get pod` looks better than `kubectl get po`   😉


frankbu:
This document will be completely replaced by the following in Istio 1.5.1 next week.
https://preliminary.istio.io/docs/setup/install/multicluster/shared/

PR #6965, opened 2020-03-25
Remove sourceLabels match part

kebe7jun:
Fix sourceLabels example yaml error.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/concepts/traffic-management/index.md
+++ b/content/en/docs/concepts/traffic-management/index.md
@@ -303,7 +303,6 @@ spec:
     route:
     - destination:
         host: ratings
-...
 {{< /text >}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/traffic-management/index.md
```diff
@@ -307,7 +307,7 @@ spec:
 
   http:
   - match:
-      sourceLabels:
+    - sourceLabels:
         app: reviews
     route:
 ...
```
frankbu: Maybe it would be better to just remove this sourceLabels part, since it seems to have nothing to do with the example being described.

PR #6974, opened 2020-03-25
Fix check for link: latest_release in events

frankbu:


ericvn:
Looking, these changes might already be part of #6973.

frankbu:
Remove cherrypick label : already fixed in release-1.5

PR #6981, opened 2020-03-26
Remove static operator yaml files

frankbu:
Fixes https://github.com/istio/istio/issues/22485


content/en/docs/setup/install/standalone-operator/index.md
```diff
@@ -35,14 +35,7 @@ instead.
...
 
-    Alternatively, you can deploy using `kubectl` from a pre-rendered manifest, which will install the latest released
-    version of the operator:
-
-    {{< text bash >}}
-    $ kubectl apply -f https://istio.io/operator.yaml
-    {{< /text >}}
-
-    Another alternative method of deployment is using Helm:
+    Alternatively, you can deploy using Helm:
```
howardjohn: Is the helm one also hardcoded to 1.5.1? Shouldn't it be `{{ latest_version }}` or whatever?

frankbu: Good question. Also, do we really even need to document the helm alternative?

ostromart: Some users have asked for this. Yes, we should change the version in the docs here to be < latest > or something like that.

frankbu: Actually the helm chart is just a path in the release image, so I think it's good.

PR #6982, opened 2020-03-26
zh-translation:/docs/concepts/security/index.md

potterhe:
ref: https://github.com/servicemesher/istio-official-translation/issues/1828


content/en/docs/concepts/security/index.md
```diff
@@ -487,7 +487,7 @@ policies:
   workloads  switch successfully to the desired mode, you can apply the policy
   with the final mode. You can use Istio telemetry to verify that workloads
   have switched successfully.
-- When migrating peer authentication policies from one JWT to another, add
+- When migrating request authentication policies from one JWT to another, add
```
frankbu: This line is not zh translation, but instead an en doc change, so it should be made in a separate en PR.

potterhe: pr https://github.com/istio/istio.io/pull/7009

PR #6989, opened 2020-03-26
remove alpha security policy

yangminzhu:
https://github.com/istio/istio/issues/22060

Manually run `make update_ref_docs` to make sure the reference page is not showing on the site even the directory is still auto generated.


ericvn:
I suspect there was a failure in the `make update_ref_docs` when run locally since it deletes the files and then generates new ones, ie. the delete happened but the files weren't regenerated.

That said, an automated job runs the `make update_ref_docs` and I suspect that the desired changes will happen shortly. See https://github.com/istio/istio.io/pull/6976.

That said, that PR is having some issues since some referred to items were removed and some additional changes are required (and would be required in this PR as well).

yangminzhu:
@frankbu @ericvn could you take a look? I reverted the `make update_ref_docs` changes and only changed to remove the v1alpha1 RBAC policy, we should fix the authN part separately to unblock #6976. Thanks.

PR #6991, opened 2020-03-27
Update index.md

brant4test:
fix typo


frankbu:
Thanks for the suggestion, but it's also grammatically correct as is, not to mention it's the same wording used in previous release notes.

PR #6992, opened 2020-03-27
Fix privileged flag.

lordofthejars:


lordofthejars:
@googlebot I signed it!

ericvn:
Would you also update this same txt in `istio.io/content/en/faq/security/verify-mtls-encryption.md`?

ericvn:
Looks like this is already going in. I'll update both files in _master_

ericvn:
I created #6997 for master with both changes, else if you make the additional change here and I see Frank now has a cherry pick back to master we can close what I just did.

frankbu:
Fixed in #6997

PR #6994, opened 2020-03-27
Update base64 command with --debug flag instead of -d to make it compatible with other OS

fai555:
base64 takes -D flag to decode

### DIFF
--- a/content/en/docs/tasks/security/authorization/authz-jwt/index.md
+++ b/content/en/docs/tasks/security/authorization/authz-jwt/index.md
@@ -108,7 +108,7 @@ with a `/` separator as shown:
 This causes Istio to generate the attribute `requestPrincipal` with the value `testing@secure.istio.io/testing@secure.istio.io`:
 
     {{< text syntax="bash" expandlinks="false" >}}
-    $ TOKEN=$(curl {{< github_file >}}/security/tools/jwt/samples/demo.jwt -s) && echo $TOKEN | cut -d '.' -f2 - | base64 -D -
+    $ TOKEN=$(curl {{< github_file >}}/security/tools/jwt/samples/demo.jwt -s) && echo $TOKEN | cut -d '.' -f2 - | base64 --decode -
     {"exp":4685989700,"foo":"bar","iat":1532389700,"iss":"testing@secure.istio.io","sub":"testing@secure.istio.io"}
     {{< /text >}}
 
--- a/content/en/docs/tasks/security/authorization/authz-jwt/index.md
+++ b/content/en/docs/tasks/security/authorization/authz-jwt/index.md
@@ -154,7 +154,7 @@ the JWT to have a claim named `groups` containing the value `group1`:
 1. Get the JWT that sets the `groups` claim to a list of strings: `group1` and `group2`:
 
     {{< text syntax="bash" expandlinks="false" >}}
-    $ TOKEN_GROUP=$(curl {{< github_file >}}/security/tools/jwt/samples/groups-scope.jwt -s) && echo $TOKEN_GROUP | cut -d '.' -f2 - | base64 -D -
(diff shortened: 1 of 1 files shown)


ericvn:
Is there a particular OS where the `-d` doesn't work?

On MacOS a man base64 lists 3 different options:
```
     -d
     -D
     --decode             Decode incoming Base64 stream into binary data.
```

frankbu:
On Linux, only `-d` is allowed. On osx, `-d` is short for `--debug`.

So I think the only fix that works on all platforms is `--decode`.

fai555:
-d didn't work in Mac OSX. After putting -D it worked. As you said @ericvn, --debug may be a better option that can support other OS.

frankbu:
@fai555 Yes, `--debug` is the option that seems to work on all platforms. Please update the PR so we can merge it. Thanks.

sdake:
> @fai555 Yes, `--debug` is the option that seems to work on all platforms. Please update the PR so we can merge it. Thanks.

@frankbu you had it right in your first comment https://github.com/istio/istio.io/pull/6994#issuecomment-605015959.

the PR should use `--decode` because we don't wish to debug in this context.

ericvn:
Just looking at other files that use base64 -d (or another flavor). I see a number that use `--decode`. Some that use `-d` and we may want to change as well include:
- content/en/docs/setup/platform-setup/kind/index.md
- content/en/docs/tasks/security/authorization/authz-jwt/index.md

ericvn:
I also note that this PR was opened against `release-1.5` instead of `master` and may have been opened from the doc. We will want to cherry-pick this back to `master` as well.

content/en/docs/tasks/security/authorization/authz-jwt/index.md
```diff
@@ -108,7 +108,7 @@ with a `/` separator as shown:
 This causes Istio to generate the attribute `requestPrincipal` with the value `testing@secure.istio.io/testing@secure.istio.io`:
 
     {{< text syntax="bash" expandlinks="false" >}}
-    $ TOKEN=$(curl {{< github_file >}}/security/tools/jwt/samples/demo.jwt -s) && echo $TOKEN | cut -d '.' -f2 - | base64 -d -
+    $ TOKEN=$(curl {{< github_file >}}/security/tools/jwt/samples/demo.jwt -s) && echo $TOKEN | cut -d '.' -f2 - | base64 -D -
```
ericvn, line 111:
```suggestion
    $ TOKEN=$(curl {{< github_file >}}/security/tools/jwt/samples/demo.jwt -s) && echo $TOKEN | cut -d '.' -f2 - | base64 --debug -
```

content/en/docs/tasks/security/authorization/authz-jwt/index.md
```diff
@@ -154,7 +154,7 @@ the JWT to have a claim named `groups` containing the value `group1`:
 1. Get the JWT that sets the `groups` claim to a list of strings: `group1` and `group2`:
 
     {{< text syntax="bash" expandlinks="false" >}}
-    $ TOKEN_GROUP=$(curl {{< github_file >}}/security/tools/jwt/samples/groups-scope.jwt -s) && echo $TOKEN_GROUP | cut -d '.' -f2 - | base64 -d -
+    $ TOKEN_GROUP=$(curl {{< github_file >}}/security/tools/jwt/samples/groups-scope.jwt -s) && echo $TOKEN_GROUP | cut -d '.' -f2 - | base64 -D -
```
ericvn, line 157:
```suggestion
    $ TOKEN_GROUP=$(curl {{< github_file >}}/security/tools/jwt/samples/groups-scope.jwt -s) && echo $TOKEN_GROUP | cut -d '.' -f2 - | base64 --debug -
```

ericvn:
@fai555 Hello, I think you should be able to select to add the suggestions to a batch and commit them to make the change to `--debug`. Of you cannot batch, then it'll show as two commits. Then we can get this change in. Thank you.

frankbu:
Fixed already in #7039

PR #6996, opened 2020-03-27
Copy iop-integration-test-defaults file from Istio

bison:
I was looking into istio/istio#21634, and found that the tests are broken with the latest istio master. As of istio/istio#21144, the default configuration values for the tests are looked for in the istio source tree as determined by the `REPO_ROOT` environment variable. However, when running the istio.io tests, `REPO_ROOT` points to the istio.io repo and not the istio repo, so the defaults aren't found and test setup fails.

This copies the defaults files from the istio source into the istio.io tree in `bin/init.sh` the way some other things are. Some other potential solutions:

- Change the `REPO_ROOT` environment variable to point to the istio source instead of the istio.io source in `Makefile.core.mk` and `common/scripts/run.sh`, but that seems confusing. Also, it looks like `run.sh` is updated automatically from the `common-files` repo, which might be a problem.

- Change the istio test framework to build-in the defaults files with `go-bindata` and avoid looking for them at run time.

- Just copy the `iop-integration-test-defaults.yaml` in the istio.io repo manually and commit it.

I'm new to the project -- not really sure what's best. Any suggestions?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@ install/**
 # contains samples to demo Istio 
 samples/**
 
+# configuration copied from Istio
+tests/integration/**
+
 # contains the built artifacts
 out/** 
 
--- a/bin/init.sh
+++ b/bin/init.sh
@@ -57,6 +57,12 @@ rm -rf "${ISTIOIO_GO}/install" "${ISTIOIO_GO}/samples"
 cp -a "${ISTIO_GO}/install" "${ISTIOIO_GO}/install"
 cp -a "${ISTIO_GO}/samples" "${ISTIOIO_GO}/samples"
 
+# Copy the default configuration over from Istio.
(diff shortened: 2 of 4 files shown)


dgn:
@istio/wg-test-and-release-maintainers I think this is something for you guys

bison:
Looks like this is basically the same as #6999.

bison:
I guess this isn't needed since #6999 was merged instead, though it seems to me like `tests/integration` should still be ignored like it is here.

sdake:
@bison thank you for your PR. As a community, we are working to improve duplicated work that comes in after a PR has already been submitted.

In this case, your PR was first and should have been reviewed and merged first. This happens often and it is an unfortunate waste of someone's time (in this case, @nmittler wasted his time, as you already had a solution).

I would encourage you to continue to contribute to Istio, even if this was a negative experience.

Cheers
-steve

PR #6997, opened 2020-03-27
Fix privildeged

ericvn:


ericvn:
This is a dup of #6992 if that PR adds the second incorrect spelling and we cherry-pick back to _master_. Will reopen if needed.

PR #7010, opened 2020-03-31
Fix mTLS migration test script

bison:
The mTLS migration task in the docs was updated to use
PeerAuthentication instead of Policy and DestinationRules, but the
tests were never updated.  This updates the test script to be inline
with the docs.

I was going to update the docs to use the snippets generated here, but
it doesn't look like any of the docs actually import those, and I'm
not sure the `make update_examples` mentioned in the tests README
actually does anything.

### DIFF
--- a/tests/security/scripts/mtls_migration.txt
+++ b/tests/security/scripts/mtls_migration.txt
@@ -20,56 +20,38 @@ set -o pipefail
 
 # $snippet curl_foo_bar_legacy syntax="bash" outputis="text"
 #nolint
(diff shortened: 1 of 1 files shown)


frankbu:
@bison For now, we just want to try to keep the tests aligned with the docs manually, as you are doing. Don't worry about importing snippets into the docs.

To synchronize the tests with the contents of the docs, we are looking to change the direction of snippet generation. Instead of having the test-generated snippets get imported into the docs, we are looking to generate snippets (maybe env vars) containing the commands from the doc (`<text>` sections) and then use those in the test scripts.

See issue: https://github.com/istio/istio.io/issues/6335

bison:
@frankbu: Thanks, yeah, I found that after posting this. In any case, this fixes these tests for now. Would love to get this merged.

bison:
The only remaining test failure should be `TestCircuitBreaker` which is failing consistently in CI and for me locally. I was looking into that today.

PR #7014, opened 2020-03-31
Make site linter more configurable

frankbu:
This change makes doc linting more configurable:

1. User can pass one or more languages to check, e.g., `lint_site.sh en zh` will only run the linter over the en and zh content.
2. Don't check external links (i.e., only check internal links) unless `CHECK_EXTERNAL_LINKS` is set to `true`.
3. New make target, `make lint-en`, will only check `en` and only internal links.
4. `pt-br` checking disabled by default (including `make lint`).

### DIFF
--- a/scripts/lint_site.sh
+++ b/scripts/lint_site.sh
@@ -18,6 +18,12 @@ set -e
 
 FAILED=0
 
+if [[ "$#" -ne 0 ]]; then
+    LANGS="$*"
+else
+    LANGS="en zh"
+fi
+
 # This performs spell checking and style checking over markdown files in a content
 # directory. It transforms the shortcode sequences we use to annotate code blocks
 # into classic markdown ``` code blocks, so that the linters aren't confused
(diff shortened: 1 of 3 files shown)


frankbu:
This is just a temporary solution for some low hanging fruit. We still need to do a better job of separating the languages (maybe even into separate repos).

scripts/lint_site.sh
```diff
@@ -97,52 +103,78 @@ check_content() {
...
+fi
+while IFS= read -r -d '' f; do
     if grep -H -n -i -e blockquote "${f}"; then
         echo "Ensure content only uses {{< tip >}}, {{< warning >}}, {{< idea >}}, and {{< quote >}} instead of block quotes"
         FAILED=1
     fi
 
     if grep -H -n -e "\"https://github.*#L[0-9]*\"" "${f}"; then
-        echo "Ensure content doesn't use links to specific lines in GitHub files as those are too brittle"
-        FAILED=1
```
frankbu, line 122: The previous version of lint_site.sh had a bug causing some errors to not make the script fail. Specifically, this FAILED=1 was getting lost because the script was piping to a while loop and running into [this issue](https://github.com/koalaman/shellcheck/wiki/SC2031). Now that I fixed the bug, the build is failing because there are several occurrences of brittle github links in the current docs. So, for now I've just made this a warning, instead of an error, which is effectively how it was working previously.

scripts/lint_site.sh
```diff
@@ -167,7 +167,7 @@ if [[ ${SKIP_LINK_CHECK:-} != "true" ]]; then
     if [[ ${#SKIP_LANGS[@]} -ne 0 ]]; then
         printf -v ignore_files "/^.\/public\/%s/," "${SKIP_LANGS[@]}"; ignore_files="${ignore_files%,}"
     fi
-    if [[ ${SKIP_EXTERNAL_LINK_CHECK:-} == "true" ]]; then
+    if [[ ${SKIP_EXTERNAL_LINK_CHECK:-} != "false" ]]; then
```
frankbu, line 170: I've changed the default behaviior to not check external links (only internal ones). I think we spend too much time building and rebuilding because of temporary issues with external links. IMO, we should just do the full external link check periodically, and not with every build.

scripts/lint_site.sh
```diff
@@ -97,52 +103,78 @@ check_content() {
...
-    if grep -H -n -E -e '\((/docs|/about|/blog|/faq|/news)' "${f}"; then
-        echo "Ensure translated content doesn't include references to English content"
-        FAILED=1
-    fi
-done
-
-if [[ ${SKIP_LINK_CHECK:-} == "" ]]; then
-    if ! htmlproofer ./public --assume-extension --http-status-ignore "0,429" --check-html --check-external-hash --check-opengraph --timeframe 2d --storage-dir .htmlproofer --url-ignore "/archive.istio.io/,/localhost/,/github.com/istio/istio.io/edit/,/github.com/istio/istio/issues/new/choose/,/groups.google.com/forum/,/www.trulia.com/,/apporbit.com/,/www.mysql.com/,/www.oreilly.com/,/docs.okd.io/,/www.aporeto.com/"; then
-        FAILED=1
+    if [[ ${SKIP_EXTERNAL_LINK_CHECK:-} != "false" ]]; then
```
ericvn, line 170:
```suggestion
    if [[ ${SKIP_EXTERNAL_LINK_CHECK:-true} != "false" ]]; then
```

ericvn: Just to emphasis that you are setting it to true.

ericvn: Or as we discussed, change to `CHECK_EXTERNAL_LINK` and default to false.

PR #7016, opened 2020-04-01
Fix TCP Traffic Shifting test script

bison:
The `TestTCPTrafficShifting` test script relies on fetching the port
named "tcp" on the istio-ingressgateway service, but that port is not
present in the default installation profile -- only in the demo
profile.  This patches the service to add the port in the setup
script, and also fixes the usage of `nc` in the test script.


tests/trafficmanagement/scripts/tcp_trafficshifting_setup.txt
```diff
@@ -32,3 +32,12 @@ $ kubectl label namespace istio-io-tcp-traffic-shifting istio-injection=enabled
 # $snippet deploy_tcp_echo_v1.sh
 $ kubectl apply -f @samples/tcp-echo/tcp-echo-services.yaml@ -n istio-io-tcp-traffic-shifting
 # $endsnippet
+
+# Add the TCP port to the ingress-gateway
+kubectl -n istio-system patch service istio-ingressgateway --patch "
```
dgn, line 37: I haven't looked at the full test, but I'm not sure this is the right place to do this. These sorts of things should happen in the config step of the test, ie set the profile to demo there. But even so, do we really require this port? If that's the case, it must be mentioned on the docs page as well

PR #7031, opened 2020-04-03
Remove GKE v1.11.x instructions

adrianludwin:
See https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#iam-rolebinding-bootstrap. Assuming that these instructions were *only* required for GKE, they are no longer necessary.

### DIFF
--- a/content/en/docs/setup/install/multicluster/gateways/index.md
+++ b/content/en/docs/setup/install/multicluster/gateways/index.md
@@ -63,16 +63,6 @@ Cross-cluster communication occurs over the Istio gateways of the respective clu
 1. Run the following commands in **every cluster** to deploy an identical Istio control plane
     configuration in all of them.
 
-    {{< tip >}}
-    Make sure that the current user has cluster administrator (`cluster-admin`) permissions and grant them if not.
-    On the GKE platform, for example, the following command can be used:
-
-    {{< text bash >}}
-    $ kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user="$(gcloud config get-value core/account)"
-    {{< /text >}}
-
-    {{< /tip >}}
-
     * Create a Kubernetes secret for your generated CA certificates using a command similar to the following. See [Certificate Authority (CA) certificates](/docs/tasks/security/plugin-ca-cert/) for more details.
 
         {{< warning >}}


content/en/docs/setup/install/multicluster/gateways/index.md
```diff
@@ -63,16 +63,6 @@ Cross-cluster communication occurs over the Istio gateways of the respective clu
 1. Run the following commands in **every cluster** to deploy an identical Istio control plane
     configuration in all of them.
 
-    {{< tip >}}
-    Make sure that the current user has cluster administrator (`cluster-admin`) permissions and grant them if not.
```
frankbu, line 67: Are you sure that these instructions only apply to GKE. The GKE part looks like just an example. Maybe it's safer to leave the tip in, but remove the GKE part?
```
{{< tip >}}
Make sure that the current user has cluster administrator (`cluster-admin`) permissions.
{{< /tip >}}

```

adrianludwin: I'm happy to leave something in, but I'm _pretty_ sure even this tip was GKE-specific. I really doubt you need `cluster-admin` permissions, I'm _guessing_ that `admin` permissions are enough. But the standard GKE setup instructions were always "give yourself cluster-admin," which probably doesn't even make sense on other platforms.

I can't work on this right now but I'll try to get back to it later today.

content/en/docs/setup/install/multicluster/gateways/index.md
```diff
@@ -63,16 +63,6 @@ Cross-cluster communication occurs over the Istio gateways of the respective clu
...
-    On the GKE platform, for example, the following command can be used:
```
sdake, line 68: I never liked these platform specific instructions. This part can definitely go.

content/en/docs/setup/install/multicluster/gateways/index.md
sdake: @frankbu I can confirm `cluster-admin` is a GKE specific role. Something more generic may be:

`Ensure that the current user has cluster administrator privileges. Most clouds grant this by default. Please read your cloud provider's documentation if in doubt.`

adrianludwin: The K8s docs seem to suggest that `cluster-admin` is a [K8s builtin role](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) - is this incorrect? I doubt it's actually needed to install Istio though (but I haven't tried).

ericvn:
I assume these changes also pertain moving forward, so the normal process (see https://istio.io/about/contribute/github/#branching-strategy) would be to apply this change in the `master` branch and then cherry-pick back to the `release-1.5` branch.

adrianludwin:
@ericvn Are you asking me to make that change or suggesting that someone else should? I made this PR based on clicking the "edit" button on the website, it hadn't even noticed I was on a branch.

ericvn:
@adrianludwin OK. If you go about it creating it that way, I understand what the automation is doing. In that case, we should probably leave it on `release-1.5` and we can addd the label to cherry-pick it back to `master. Does that sound correct, @frankbu? Might also be why some other PRs are pointed at release-1.5 and we aren't really seeing responses from the authors.

content/en/docs/setup/install/multicluster/gateways/index.md
frankbu: OK, let's just go ahead and remove the tip completely.

The prereq section of this doc already says you need authority to deploy Istio on each cluster, anything more is just noise.

PR #7038, opened 2020-04-07
Update sed to include space before boolen value

### LINKED ISSUE #7034: Issue with docs/ops/configuration/mesh/app-health-check/index.md
Firstly, thank you for all the hard work thus far, Istio is an amazing tool. We believe it will unlock valuable information.

While going through the documentation our team noticed an error concerning sidecar auto injection. The regex is a bit off

Currently:
```
kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e 's/"rewriteAppHTTPProbe":false/"rewriteAppHTTPProbe":true/' | kubectl apply -f -
```
Should be:
```
kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e 's/"rewriteAppHTTPProbe":\ false/"rewriteAppHTTPProbe":\ true/' | kubectl apply -f -
```

Take not of the space between `rewriteAppHTTPProbe` and the boolean value. This is where the error occures.

Source Docs: https://github.com/istio/istio.io/blob/f198c7bc4e121b7debf0e2949eaf1fb9683e2fe4/content/en/docs/ops/configuration/mesh/app-health-check/index.md

We would also recommended updating the translated docs as well.

shamsher31:
To fix #7034 

[X] Docs
[X] User Experience

### DIFF
--- a/content/en/docs/ops/configuration/mesh/app-health-check/index.md
+++ b/content/en/docs/ops/configuration/mesh/app-health-check/index.md
@@ -113,7 +113,7 @@ You have two ways to enable Istio to rewrite the liveness HTTP probes.
 **Alternatively**, update the configuration map of Istio sidecar injection:
 
 {{< text bash >}}
-$ kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e 's/"rewriteAppHTTPProbe":false/"rewriteAppHTTPProbe":true/' | kubectl apply -f -
+$ kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e 's/"rewriteAppHTTPProbe":\ false/"rewriteAppHTTPProbe":\ true/' | kubectl apply -f -
 {{< /text >}}
 
 The above installation option and configuration map, each instruct the sidecar injection process to automatically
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/configuration/mesh/app-health-check/index.md
```diff
@@ -113,7 +113,7 @@ You have two ways to enable Istio to rewrite the liveness HTTP probes.
 **Alternatively**, update the configuration map of Istio sidecar injection:
 
 {{< text bash >}}
-$ kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e 's/"rewriteAppHTTPProbe":false/"rewriteAppHTTPProbe":true/' | kubectl apply -f -
+$ kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e 's/"rewriteAppHTTPProbe":\ false/"rewriteAppHTTPProbe":\ true/' | kubectl apply -f -
```
frankbu, line 116: Why the `\`?

shamsher31: This is to escape colon delimiter, as it is preceded by space and separate two strings `rewriteAppHTTPProbe` and `false`. Not sure if it works without `\`

frankbu: But the backslash is escaping the following space character, not the colon. Since the whole rule is inside single quotes (`'...'`), and `\ ` has no meaning in sed, as far as I know, so my guess is that the backslash does nothing.

shamsher31: Thanks @frankbu , I did try to run the command without `\` and it works perfectly fine and updates `rewriteAppHTTPProbe`.

PR #7039, opened 2020-04-07
Update base64 -d to --decode

ericvn:


frankbu:
How about also fixing the doc from #6994 and then we can close that one as a duplicate? We've been waiting long enough for the suggested changes.

ericvn:
Done.

PR #7043, opened 2020-04-08
Fix flaky circuit breaker test

bison:
The `TestCircuitBreaker` test is very flaky.  It relies on examining the percentage of requests that succeed or fail to a service with circuit breaking configured, and ensuring the numbers fall within a given range.  Unfortunately, the percentages tend to vary widely from run to run, so the test fails as often as it passes.

This simplifies the test by only verifying that both successes and failures occur instead of checking the percentage of each.

Is there a better solution here? Is there something more concrete to verify in the test, or some configuration for the circuit breaker that might yield more consistent results?

### DIFF
--- a/tests/trafficmanagement/circuitbreaker_test.go
+++ b/tests/trafficmanagement/circuitbreaker_test.go
@@ -27,22 +27,8 @@ func TestCircuitBreaker(t *testing.T) {
 			Add(
 				istioio.Script{Input: istioio.Path("scripts/circuitbreaker_test_setup.txt")},
 				istioio.MultiPodWait("istio-io-circuitbreaker"),
-				istioio.Script{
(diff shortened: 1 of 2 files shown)


bison:
@frankbu, yeah, I'm really not sure what the best solution is in the long run. Though, this test is new, and the tests wen't being run regularly until very recently. So, I'm not sure the strategy of checking percentages was ever working consistently, or is going to.

If you like, I can create an issue about trying to make it work, and link to it in the script. Does that work?

brian-avery:
I agree with Frank. If we can't find a range of percentages that this is regularly passing at, then that points at a larger issue to me. Lets comment out the code, get this passing for now, and open an issue to investigate it further.

bison:
@frankbu / @brian-avery: I opened #7074 about looking into making the original method work, and I added a note in the code with a link to that and the relevant PRs. Is that good? I can add back the code, commented out, if you prefer.

PR #7044, opened 2020-04-08
Generate test snippets from md files POC

### LINKED ISSUE #6335: Doc tests should use snippets generated from doc

frankbu:
Fixes #6335


content/en/docs/tasks/security/authentication/authn-policy/authn-policy_snip.txt
```diff
@@ -0,0 +1,451 @@
+####################################################################################################
+# WARNING: THIS IS AN AUTO-GENERATED FILE, DO NOT EDIT. PLEASE MODIFY THE ORIGINAL MARKDOWN FILE
+####################################################################################################
+
+snip_line_23() {
```
brian-avery, line 5: What if a doc gets an extra line? Does this break the test the next time the generator is run? Also, I feel like referring to everything as line_* might make it hard to maintain tests. I know we have the option of `snip_id` so the decision not to rely on the auto-generated name might rely on the test writer/reviewer.

tests/trafficmanagement/scripts/traffic_shifting.txt
```diff
@@ -18,6 +18,9 @@ set -e
 set -u
 set -o pipefail
 
+#TODO in what directory will the _snip.txt file be stored?
+source ${TEST_OUTPUT_DIR}/traffic-shifting_snip.txt
```
brian-avery, line 22: We discussed `_snip.txt` being in the same directory as the test, which I think is fine. However, other tests may rely on these as well. For example, many tests rely on Bookinfo being present.

content/en/docs/tasks/security/authentication/authn-policy/authn-policy_snip.txt
frankbu: Good point. Any change to an md file will trigger a regen of its snippet file, which may cause the test to fail. If using the default snippet names, any change that adds or removes lines will likely break the test, even if the snippet contents hasn't changed at all. So, using snip_ids should definitely be a recommended best practice. Only actual command changes will then break the test.

tests/trafficmanagement/scripts/traffic_shifting.txt
frankbu: I think Rigs suggested that the _snip file should be in the same directory as the md file. Since it's really just a tester's view of the doc, that seems like the best place to me too. We just need a variable that points to the doc root directory so that the test can then import something like this:
```
source ${DOC_ROOT_DIR}/docs/tasks/traffic-management/traffic-shifting/traffic-shifting_snip.txt
```

content/en/docs/tasks/security/authentication/authn-policy/authn-policy_snip.txt
rcaballeromx: A serialized default could also be useful. You can suffix an index instead of the line number and that would solve the issue. If commands change order, the test should be revisited, so that's shouldn't be an issue. We should avoid requiring manual suffix IDs for every snippet, IMO. That way we avoid making authoring content unnecessarily cumbersome.

frankbu: @rcaballeromx Do you mean something like `snip_<section heading>_<text section number>`?
For traffic-shifting.md, it would look something like this:
```
snip_apply_weight_based_routing_1()
snip_apply_weight_based_routing_2()
...
snip_cleanup_1()
```
That sounds reasonable to me. Still could break, but a lot less fragile than line numbers.

rcaballeromx: Yes, although maybe using the folder name will help keep them short and readable.
`snip_authn-policy_1()` for example.

frankbu: My suggestion is less fragile. Only the following snippets in the same section of the doc will break (change) if one is added or removed. With your version, all the following snippets will break.

PR #7053, opened 2020-04-10
Retire helm documentation as we use a protobuf

sdake:
The new rendered source of truth is:
https://preliminary.istio.io/docs/reference/config/istio.operator.v1alpha1/

This is rendered from the API repo protobuf which (may) need description fields
set. That protobuf is here:

https://github.com/istio/api/blob/master/operator/v1alpha1/operator.proto


ericvn:
Just a pointer as the lint links need fixing since I'm not sure if these are in the contribution guidelines yet:
```
From the doc WG notes: Propose process for fixing broken links to documents that we remove:
If the broken link is from a release announcement (e.g. announcing-1.3.8), change the link to point to the same document in the corresponding archive release (e.g., prepend https://archive.istio.io/1.3 to the link).
Else if there is a replacement doc (containing an alias for it), change the link to reference the new doc.
Else change the link to point the same doc in the latest archive release where it still exists (e.g., https://archive.istio.io/1.4).
``

sdake:
https://github.com/istio/istio.io/issues/7067

sdake:
Superceeded by https://github.com/istio/istio.io/pull/7106

PR #7055, opened 2020-04-10
shouldn't this be istioctl, not kubectl ?

gbucknel:
This looked like a typo to me ? Not 100% sure though .

### DIFF
--- a/content/en/docs/setup/install/istioctl/index.md
+++ b/content/en/docs/setup/install/istioctl/index.md
@@ -191,7 +191,7 @@ $ istioctl manifest generate > $HOME/generated-manifest.yaml
 Inspect the manifest as needed, then apply the manifest using this command:
 
 {{< text bash >}}
-$ kubectl apply -f $HOME/generated-manifest.yaml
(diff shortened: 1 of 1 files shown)


rokrokss:
It's kubectl.
`istioctl manifest generate ~` generates manifest.yaml from the overlay values(like vaules.yaml in helm) that can be applied by kubectl.
`istioctl manifest apply ~` applies with the overlay values, skipping above step.

ericvn:
Agree it's `kubectl`. `istio manifest apply` is basically `istioctl manifest generate ~ | kubectl apply -f -`, the inverse of the longer line to delete the configuration.

frankbu:
Closing because it's correct as is. The yaml file for the `-f` option in `istioctl manifest apply` would be an `IstioOperator` configuration, but in this case it contains a generated kubernetes configuration, therefore the required command is `kubectl apply -f`.

gbucknel:
Cool - thanks, sorry for the hassle. I tried the kubectl command and it didn't work, so assumed it was wrong.

PR #7059, opened 2020-04-10
Fixing incorrect quote type in istio-authorization blog post

mrf:
Blog post code sample for ServiceRoleBinding had trailing "fancy" quotes making the sample unusable.


mrf:
@googlebot I signed it!

PR #7078, opened 2020-04-14
Ignore copied tests/integration directory

bison:
The configuration for the Istio installation used in integration tests
is copied at run-time from the Istio repository by `bin/init.sh` as of
istio/istio.io#6999, but the directory is otherwise empty here.  This
adds it to `.gitignore` to avoid a dirty tree after every test run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@ install/**
 # contains samples to demo Istio 
 samples/**
 
+# copied from istio repository
+tests/integration/**
+
 # contains the built artifacts
 out/** 
 


ericvn:
I noticed this yesterday as well. I don't see this being cleaned up by `make clean`. My thought was that the code should be changed to use the `out` directory or someplace else to simplify cleaning.

ericvn:
I'll push the clean changes mentioned above in a new PR.

bison:
Sounds fine. I was just tired of seeing it in `git status`, though other things are copied by `bin/init.sh` as well, and they are in the `.gitignore` file. Seems a little a weird to use `out` for things that aren't actually products of the build.

I honestly think this file, `iop-integration-test-defaults.yaml`, should just be copied over and commited to the repo, or else added to the common-files repo.

ericvn:
_out_ is really supposed to be where various build artifacts go during builds, but that doesn't always happen. I would need to do some more investigation to see if we could do something to remove some of these things from the normal directory.

PR #7035, opened 2020-04-06
Initial documentation on Multiple Control Planes

howardjohn:
I suspect this will be improved as we get more user feedback and
istioctl integrations, but this is a reasonable start

This includes both a blog post and upgrade doc update, as they reference each other. I can split these out if needed, but since its the same topic I figured we can avoid duplicated comments here

PR #7082, opened 2020-04-15
Update future date to the date when blog was written

shamsher31:
The blog post date was `MAY 19, 2020`. Updated date to the date when the blog post [pull request was raised](https://github.com/istio/istio.io/pull/7035) which is `APRIL 7, 2020`.


frankbu:
@howardjohn because it's a blog post, which is a point in time (instead of a long living) doc, it seems a little weird. So, I'm not sure we should generally do this. If you want to let people preview a blog post in advance of its post date, you can always point them to a PR preview.

I'm OK with leaving this one as an experiment, now that it's there. Depending on if it causes more confusion or not, we can decide whether to do this again in the future or not.

PR #7084, opened 2020-04-15
Snip regen

frankbu:


Makefile.core.mk
```diff
@@ -73,13 +73,20 @@ endif
...
 
-gen:
+site:
 	@scripts/gen_site.sh
 
-build: gen
+snips:
+	@scripts/gen_snips.sh
+
+gen: snips
```
howardjohn, line 82: I am not familiar with this, does this not need `site` as well?

howardjohn:
Nevermind no need to hold, I cannot approve :slightly_smiling_face:

Makefile.core.mk
frankbu: I don't think so. `make site` generates and serves the site. That's why I'm renaming it from `gen` to `site` and just using `make gen` to generate the artifacts, so `gen-check` does the right thing.

PR #7085, opened 2020-04-16
User guide tests and remove manual steps for plugging in CA cert

lei-tang:
- Add user guide tests for plugging in CA cert.
- Remove the manual steps in the user guide of plugging in CA cert to make it easier for an user to try the guide.


lei-tang:
@frankbu Thanks! I have converted the tests to work the new way. Please take another look.

scripts/snip.py
```diff
@@ -91,7 +91,7 @@
             else:
                 id = "snip_%s_%d" % (section, snipnum)
             if kind == "bash":
-                script = "\n%s() {\n" % id
+                script = "\n# shellcheck disable=SC2046\n%s() {\n" % id
```
frankbu, line 94: This is wrong. We should not be disabling any lint checks in generated bash functions. If there are lint errors in the generated snip.sh functions, it means that the command in the original index.md file is not following bash best practice. This is another benefit of the new model: we get lint checking of the commands that we are publishing in the docs.

You should fix the lint problem by changing the command in index.md and then regen.

lei-tang: Done: the lint problem is fixed by changing the command in index.md and the disabling lint check has been removed.

PR #6941, opened 2020-03-21
[release-1.4] Manual cherry-pick of #6886 - Remove KubeCon EU event, ignore 429s.

ericvn:

PR #7095, opened 2020-04-16
Backport of openshift installation instructions

jacob-delgado:
Fixes https://github.com/istio/istio/issues/22449

Related to
https://github.com/istio/operator/pull/771
https://github.com/istio/installer/pull/668
https://github.com/istio/cni/pull/278

### DIFF
--- a/content/en/docs/setup/additional-setup/cni/index.md
+++ b/content/en/docs/setup/additional-setup/cni/index.md
@@ -127,7 +127,8 @@ The following table shows the required settings for many common Kubernetes envir
 | EKS (AWS) | _(none)_ | _(none)_ |
 | AKS (Azure) | _(none)_ | _(none)_ |
 | Red Hat OpenShift 3.10+ | _(none)_ | _(none)_ |
-| Red Hat OpenShift 4.2+ | `--set components.cni.namespace=kube-system --set values.cni.cniBinDir=/var/lib/cni/bin --set values.cni.cniConfDir=/etc/cni/multus/net.d --set values.cni.chained=false --set values.cni.cniConfFileName="istio-cni.conf"` | _(none)_ |
+| Red Hat OpenShift 4.2+ | `--set values.sidecarInjectorWebhook.injectedAnnotations."k8s\.v1\.cni\.cncf\.io/networks"=istio-cni --set components.cni.namespace=kube-system --set values.cni.cniBinDir=/var/lib/cni/bin --set values.cni.cniConfDir=/etc/cni/multus/net.d --set values.cni.chained=false --set values.cni.cniConfFileName="istio-cni.conf"` | _(none)_ |
+
 ### GKE setup
 
 1.  Refer to the procedure to [prepare a GKE cluster for Istio](/docs/setup/platform-setup/gke/) and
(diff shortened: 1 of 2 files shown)


ericvn:
Looks like the lint errors are do to **429** errors. This was fixed in 1.5 and above, but not 1.4.

ericvn:
And one other comment and don't know if this affects this whole solution: @frankbu made a comment in https://github.com/istio/istio.io/pull/6941 `Oh, just realized this is release-1.4. We can't update archived sites.`

I'm not sure why they can't be updated, but thoughts on if that should prevent us from making the change.

content/en/docs/setup/additional-setup/cni/index.md
```diff
@@ -126,8 +127,7 @@ The following table shows the required settings for many common Kubernetes envir
 | EKS (AWS) | _(none)_ | _(none)_ |
 | AKS (Azure) | _(none)_ | _(none)_ |
 | Red Hat OpenShift 3.10+ | _(none)_ | _(none)_ |
-| Red Hat OpenShift 4.2+ | `--set cni.components.cni.namespace=kube-system --set values.cni.cniBinDir=/var/lib/cni/bin --set values.cni.cniConfDir=/var/run/multus/cni/net.d` | _(none)_ |
-
+| Red Hat OpenShift 4.2+ | `--set components.cni.namespace=kube-system --set values.cni.cniBinDir=/var/lib/cni/bin --set values.cni.cniConfDir=/etc/cni/multus/net.d --set values.cni.chained=false --set values.cni.cniConfFileName="istio-cni.conf"` | _(none)_ |
```
jwendell, line 130: I think you need the annotations as well

ericvn:
As a comment, once can always `Details` for the **deploy/netlify** test to see what the resulting website would look like.

content/en/docs/setup/additional-setup/cni/index.md
```diff
@@ -126,7 +127,7 @@ The following table shows the required settings for many common Kubernetes envir
 | EKS (AWS) | _(none)_ | _(none)_ |
 | AKS (Azure) | _(none)_ | _(none)_ |
 | Red Hat OpenShift 3.10+ | _(none)_ | _(none)_ |
-| Red Hat OpenShift 4.2+ | `--set cni.components.cni.namespace=kube-system --set values.cni.cniBinDir=/var/lib/cni/bin --set values.cni.cniConfDir=/var/run/multus/cni/net.d` | _(none)_ |
+| Red Hat OpenShift 4.2+ | `--set values.sidecarInjectorWebhook.injectedAnnotations."k8s\.v1\.cni\.cncf\.io/networks"=istio-cni --set components.cni.namespace=kube-system --set values.cni.cniBinDir=/var/lib/cni/bin --set values.cni.cniConfDir=/etc/cni/multus/net.d --set values.cni.chained=false --set values.cni.cniConfFileName="istio-cni.conf"` | _(none)_ |
```
jwendell, line 130: We need to verify this actually works on 1.4. I remember I had issues with the dots within the annotation that needed to be fixed in operator code. Also I remember those fields were addressed differently, like `cni.components.cni.namespace` versus `components.cni.namespace`.

In short, we need to test these arguments are accurate and actually work before merging this PR.

frankbu:
As @ericvn mentioned, above, we are not able to update archived sites, they are frozen. Even if we push this change to the release-1.4 branch of the source, there is no automated rebuild, so the change won't be published to the site. We can only update docs for the current (1.5) and next (1.6) release.

Making it possible to rebuild archive sites would require significant change to the build infrastructure, so if we think we need it, somebody should open an issue (with label `site/infra`), but it won't be a simple fix, so likely won't be done anytime soon.

PR #7099, opened 2020-04-17
Use correct binary name when downloading istioctl

jbrook:


frankbu:
@jbrook We generally prefer to just fix the en version in the first PR and let the translation teams worry about syncing the change to the other languages. This way the PR can be merged without waiting for approval from the other language maintainers.

PR #7113, opened 2020-04-21
add proxy repo to scrap docs from

richardwxn:
related: https://github.com/istio/proxy/pull/2803

/cc @mandarjog @kyessenov @jasonwzm


ericvn:
If you run this command, does it pull in more docs? I would assume so. Would you add them here.

richardwxn:
@ericvn Yep I just added those docs, however shouldn't those be added by automator job?

ericvn:
@richardwxn They will get picked up, but would never merge until the lint errors are fixed.

PR #7114, opened 2020-04-21
Fix doc 'Traffic Management'

tomocy:
The external service `ServiceEntry` adds in this chapter is `ext-svc.example.com`, so fix it from `ext-resource` to `ext-svc.example.com`.

### DIFF
--- a/content/en/docs/concepts/traffic-management/index.md
+++ b/content/en/docs/concepts/traffic-management/index.md
@@ -533,7 +533,7 @@ to control the traffic to destinations that aren't registered in the mesh.
 
 ### Service entry example {#service-entry-example}
 
-The following example mesh-external service entry adds the `ext-resource`
+The following example mesh-external service entry adds the `ext-svc.example.com`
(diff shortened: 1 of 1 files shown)


ericvn:
Hello @tomocy. Normally a PR would be created against the `master` branch and then cherry-picked to the `release-1.5` branch. This makes sure that the change is carried forward into the set of docs being worked for the next release. If someone creates a change from the docs, which may be what was done here, then that gets created against `release-1.5` and we need to cherry-pick to `master`.

Not being sure what was done, I'll add the cherry-pick label to get this in `master`.

frankbu:
Good catch @ericvn! There were 3 other recent fixes from @tomocy that I didn't notice were sent to the `release-1.5` branch, and therefore would have gotten lost after the next release.

It's much safer to always send fixes to master first and cherrypick back to the release branch.

tomocy:
@ericvn @frankbu Thank you for your advice!!! I’ll keep it in mind.

PR #7126, opened 2020-04-22
Convert health check test to new model

frankbu:


tests/configuration/scripts/cleanup.txt
```diff
@@ -18,11 +18,8 @@ set -e
...
-$ kubectl delete ns health-global
-$ kubectl delete ns health-annotate
-$ kubectl delete ns health-sepport
-# $endsnippet
\ No newline at end of file
+source ${REPO_ROOT}/content/en/docs/ops/configuration/mesh/app-health-check/snips.sh
+
+snip_cleanup_1
+
+kubectl delete ns health-annotate
```
ericvn, line 25: This looks like a missing deletion of the namespace in the doc. Suspect it should be added there.

frankbu: I had to do this one a little special because on the doc page it doesn't really do the `health-annotate` test, it just shows what the yaml would be.

I think it may make sense to further improve this and some of the other tests (not to mention the corresponding doc pages themselves) in future PRs, but my first goal is just to get all the tests converted to the new model (more or less as they are), so I can remove all the snippet generation code from the framework.

PR #7127, opened 2020-04-22
Convert bookinfo test to new model

frankbu:


ericvn:
Unfortunately, the test failed.

frankbu:
@ericvn, the failure is a legitimate change needed to the doc. For the first time, it's actually testing the contents of the doc :-)

ericvn:
And we will have fixed it before community testing may have found it, which is great!

PR #7129, opened 2020-04-22
Galley cmd removed in 1.6.

therealmitchconnors:
This is blocking merges into istio/istio, see https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio/22873/update-ref-docs-dry-run_istio/10


therealmitchconnors:
On review the job is optional, so it's not blocking.  It's just irritating, and doesn't appear to have passed in it's short history.

ericvn:
Duplicate of https://github.com/istio/istio.io/pull/7132

PR #7131, opened 2020-04-22
Convert TCP traffic shifting test

frankbu:


brian-avery:
These health check failures look like legitimate failures due to: https://github.com/istio/istio/issues/22602

v1alpha1/policy no longer exists.

frankbu:
> v1alpha1/policy no longer exists.

@brian-avery yeah, that's hilarious, it was working earlier today when I converted the test. So now [this part of the document](https://preliminary.istio.io/docs/ops/configuration/mesh/app-health-check/#liveness-and-readiness-probes-with-command-option) is outdated and needs to be updated. I guess that's why we have these tests :)

PR #7133, opened 2020-04-22
Update lint_site to only check languages for files in current PR

ericvn:


ericvn:
One issue I haven't solved it having BRANCH needing to be set for the linter if no languages are defined (ie, someone running the script directly).

scripts/lint_site.sh
```diff
@@ -18,10 +18,18 @@ set -e
...
 else
-    LANGS="en zh"
+    files_changed="$(git diff --name-only --diff-filter=b "$(git merge-base HEAD "$BRANCH")")"
+    if (echo "${files_changed}" | grep -q /en/); then
+        LANGS+="en "
+    fi
+    if (echo "${files_changed}" | grep -q /zh/); then
+        LANGS+="zh"
+    fi
 fi
```
frankbu, line 33: This is very cool, might as well add pt-br here.
```suggestion
    if (echo "${files_changed}" | grep -q /pt-br/); then
        LANGS+="pt-br"
    fi
fi
```

scripts/lint_site.sh
```diff
@@ -18,10 +18,18 @@ set -e
 
 FAILED=0
 
+BRANCH="${BRANCH:-master}"
+
 if [[ "$#" -ne 0 ]]; then
     LANGS="$*"
 else
-    LANGS="en zh"
+    files_changed="$(git diff --name-only --diff-filter=b "$(git merge-base HEAD "$BRANCH")")"
```
brian-avery, line 26: Awesome :)

scripts/lint_site.sh
ericvn: Pushed commit separately since this was actually missing two spaces.

scripts/lint_site.sh
```diff
@@ -18,10 +18,22 @@ set -e
 
 FAILED=0
 
+BRANCH="${BRANCH:-master}"
+
```
frankbu, line 22: This doesn't look like it will do the right thing when running `make lint` locally.

```suggestion
```

scripts/lint_site.sh
```diff
@@ -18,10 +18,22 @@ set -e
...
 if [[ "$#" -ne 0 ]]; then
     LANGS="$*"
 else
```
frankbu, line 25:
```suggestion
elif [[ ! -z "${BRANCH} ]]; then
    LANGS="en zh"
else
```

Makefile.core.mk
```diff
@@ -96,13 +96,13 @@ clean:
...
 lint: clean_public build_nominify lint-copyright-banner lint-python lint-yaml lint-dockerfiles lint-scripts lint-sass lint-typescript lint-go
-	@scripts/lint_site.sh
+	@BRANCH=$(SOURCE_BRANCH_NAME) scripts/lint_site.sh
 
 lint-en: clean_public build_nominify lint-copyright-banner lint-python lint-yaml lint-dockerfiles lint-scripts lint-sass lint-typescript lint-go
 	@scripts/lint_site.sh en
 
 lint-fast:
-	@SKIP_LINK_CHECK=true scripts/lint_site.sh en
+	@BRANCH=$(SOURCE_BRANCH_NAME) SKIP_LINK_CHECK=true scripts/lint_site.sh
```
frankbu, line 105: This is only used for local checks, BRANCH won't be set. So best to leave it as an en-only function.
```suggestion 
	SKIP_LINK_CHECK=true scripts/lint_site.sh en
```

scripts/lint_site.sh
frankbu: Actually, on second thought, maybe the -z BRANCH thing should be the first check. If it's set, use it, otherwise same behavior as before:
```
if [[ ! -z "${BRANCH} ]]; then
    files_changed="$(git diff --name-only --diff-filter=b "$(git merge-base HEAD "$BRANCH")")"
    if (echo "${files_changed}" | grep -q /en/); then
        LANGS+="en "
    fi
    if (echo "${files_changed}" | grep -q /zh/); then
        LANGS+="zh "
    fi
    if (echo "${files_changed}" | grep -q /pt-br/); then
        LANGS+="pt-br "
    fi
elif [[ "$#" -ne 0 ]]; then
    LANGS="$*"
else
    LANGS="en zh"
fi
```
```

File: scripts/lint_site.sh (hunk 3 above)
Comment (ericvn) [reply scripts/lint_site.sh:22]: Correct, locally means that we would default to master which should be where the changes are being made against. But for simplicity I will remove this in favor of below.

File: scripts/lint_site.sh (hunk 4 above)
Comment (ericvn) [reply scripts/lint_site.sh:25]: I like this. I'll add an echo message if nothing is defined to mention one can set the BRANCH if desired.

File: Makefile.core.mk (hunk 5 above)
Comment (ericvn) [reply Makefile.core.mk:105]: But if they are using the makefile, BRANCH is set to the SOURCE_BRANCH_NAME so thinks should be fine. Especially for non-english devs. I would entertain an lint-fast-en if we wanted to force English.

**Comment by ericvn** [comment]
@frankbu Let me know if there's another scenario I am missing.

File: Makefile.core.mk (hunk 5 above)
Comment (frankbu) [reply Makefile.core.mk:105]: If users want to run `lint-fast` on local changes, I don't think SOURCE_BRANCH_NAME default of `master` will do the right thing. The correct branch to use is `my-branch`, not `master`. That's why I think we should just do `en` for this, not require any extra setting of SOURCE_BRANCH_NAME. We also don't need to make it more complicated with `lint-fast-<lang>` rules because `en` is the only one that actually does any real spell checking, and the snip.sh lint checks are also only needed in the en tree.

Comment (ericvn) [reply Makefile.core.mk:105]: @frankbu Ahh. Your scenario isn't correct. Let's say I have a branch with commits I have pushed to a PR or not. (ex: branch is lintByLanguage).

If I run the command against that branch with everything in a commit, I see:
```
git diff --name-only --diff-filter=b "$(git merge-base HEAD "lintByLanguage")"
```
If I instead run it against master, I get all the changes I have in the branch since it was created:
```
git diff --name-only --diff-filter=b "$(git merge-base HEAD "master")" 
Makefile.core.mk
scripts/lint_site.sh
```
I can get the current branch and try to use that in the script, but that never showed all my changes so didn't work. I've always run locally using `BRANCH=master ./scripts/lint_site.sh` to get the list of files. This is because my changes are typically in their own branch off of master created using something like `git checkout master` and then `git checkout -b lintByLanguage`

That said, if the person is doing some branching between PRs then they could see the list being generated incorrectly.

Comment (ericvn) [reply Makefile.core.mk:105]: The above is why I defaulted to `master` as that would be the typical branch that someone would branch off of for their PRs.

Comment (frankbu) [reply Makefile.core.mk:105]: I see, but defaulting to `master` won't do the right thing if my branch is off of branch `release-1.5`, for example. It won't fail, it just may not be doing the right thing. I still think that keeping `lint-fast` as a simple en-only function, for the reasons I mentioned above, is best. In fact, maybe `make lint` shouldn't default to master either, since that also may or may not be working properly in local.

Comment (ericvn) [reply Makefile.core.mk:105]: But we removed any defaulting at this point. If one runs the script locally directly, they will get both languages as they would before this change. They can set BRANCH if they want if they want a faster scan.

If they are running the make file locally, they will use whatever branch is specified . If they are typically doing that in master, that's fine. If they breached of of release-1.5, then it will use releases-1.5 (it's all defined in the Makefile.core.mk that they are using as SOURCE_BRANCH_NAME ?= ...).

File: scripts/lint_site.sh [hunk 6]
```diff
@@ -18,9 +18,21 @@ set -e
...
+    if (echo "${files_changed}" | grep -q /zh/); then
+        LANGS+="zh "
+    fi
+    if (echo "${files_changed}" | grep -q /pt-br/); then
+        LANGS+="pt-br "
+    fi
+elif [[ "$#" -ne 0 ]]; then
     LANGS="$*"
 else
+    echo "if BRANCH is set to the source branch, ie. master, the linter will only check languages within the PR."
```
frankbu, line 35: What PR? This message will make no sense in a local run of the script. It also wouldn't be clear to a user what "source branch" is. The message would need to be much more precise and explain something like the following: `Note: when running this script in the context of a branch from some "source branch", you can set the the BRANCH environment variable to the "source branch", which will instruct the linter to only run over the language files that include some modified files relative to the source branch`. Even that will confuse people, and given that we expect people to use make instead of calling the script directly, I suggest no message at all.
```suggestion
```

scripts/lint_site.sh
```diff
@@ -18,9 +18,21 @@ set -e
...
     LANGS="en zh"
```
frankbu, line 36: Maybe just fail in this case?
```suggestion
    echo "ERROR: no languages specified!"
    exit 1
```

Makefile.core.mk
```diff
@@ -96,13 +96,13 @@ clean:
...
 lint: clean_public build_nominify lint-copyright-banner lint-python lint-yaml lint-dockerfiles lint-scripts lint-sass lint-typescript lint-go
-	@scripts/lint_site.sh
+	@BRANCH=$(SOURCE_BRANCH_NAME) scripts/lint_site.sh
 
 lint-en: clean_public build_nominify lint-copyright-banner lint-python lint-yaml lint-dockerfiles lint-scripts lint-sass lint-typescript lint-go
 	@scripts/lint_site.sh en
 
 lint-fast:
-	@SKIP_LINK_CHECK=true scripts/lint_site.sh en
+	SKIP_LINK_CHECK=true scripts/lint_site.sh en
```
frankbu, line 105:
```suggestion
	@SKIP_LINK_CHECK=true scripts/lint_site.sh en
```

PR #7135, opened 2020-04-23
Fix usage of Policy to PeerAuthentication

jacob-delgado:
Fixes broken k8s-tests_istio.io preventing PRs to istio.io

v1alpha/Policy was removed from the api. It has been replaced with v1beta1/PeerAuthentication


frankbu:
Closing because I've incorporated this fix in #7131

PR #7154, opened 2020-04-24
Add version skew policy

howardjohn:
Please don't merge without approval from TOC or similar

### DIFF
--- a/content/en/about/version-policy/index.md
+++ b/content/en/about/version-policy/index.md
@@ -1,0 +1,40 @@
+---
+title: Versioning Policy
+description: Supported versions of various components and dependencies of Istio.
+weight: 15
+icon: cadence
+---
+
+## Istio Versioning
+
+Istio versions are expressed as **x**.**y**.**z**, where **x** is the major version, **y** is the minor version, and **z** is the patch version.
+
+Istio provides support for each minor release for 3 months after the next minor release. For example, Istio 1.0 support will be dropped 3 months after Istio 1.2 is released. See [Build & Release Cadence](/about/release-cadence/) for more details.
+
+## Component Version Skew
+
+Various components of Istio that interact with each other must be kept within supported version ranges. Below describes the required version skews. Where possible, we always recommend using the same version between all components.
+
+### Istioctl
+
+We currently recommend using `istioctl` within the same minor version as both the control plane and data plane.
+
+For example, `istioctl` version **1.0.1** could be used with control plane version **1.0.0** and data plane version **1.0.6**.
+
+### Control/Data Plane
+
+The data plane (both sidecars and gateways) must be either the same version as the control plane, or at most 1 minor version older.
+
+For example, the data plane version **1.0.0** could be used with control plane version **1.0.0** or **1.1.0**.
+
+This implies that the control plane must be upgraded first. For more information, see the [Upgrade](/docs/setup/upgrade/) documentation.
+
+## Kubernetes versions
+
+The following table indicates the versions of Kubernetes that are supported for various Istio versions. While Istio may work on versions outside of those listed below, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
+
+|Istio Version       | Kubernetes Version
+|--------------------|-----------------------
+|1.6                 | 1.15, 1.16, 1.17, 1.18
+|1.5                 | 1.14, 1.15, 1.16
+|1.4                 | 1.13, 1.14, 1.15
--- a/content/en/about/version-policy/index.md
+++ b/content/en/about/version-policy/index.md
@@ -31,4 +31,12 @@ This implies that the control plane must be upgraded first. For more information
 
 ## Kubernetes versions
 
-The versions of Kubernetes that are supported for an Istio version is shown on the [Platform Setup](/docs/setup/platform-setup/). While Istio may work on versions outside of those listed, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
+The versions of Kubernetes that are supported for an Istio version is shown below:
+
+|Istio Version       | Kubernetes Version
+|--------------------|-----------------------
+|1.6                 | 1.15, 1.16, 1.17, 1.18
+|1.5                 | 1.14, 1.15, 1.16
+|1.4                 | 1.13, 1.14, 1.15
+
+While Istio may work on versions outside of those listed, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
--- a/content/en/about/version-policy/index.md
+++ b/content/en/about/version-policy/index.md
@@ -31,10 +31,4 @@ This implies that the control plane must be upgraded first. For more information
 
 ## Kubernetes versions
 
-The following table indicates the versions of Kubernetes that are supported for various Istio versions. While Istio may work on versions outside of those listed below, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
-
-|Istio Version       | Kubernetes Version
-|--------------------|-----------------------
-|1.6                 | 1.15, 1.16, 1.17, 1.18
-|1.5                 | 1.14, 1.15, 1.16
-|1.4                 | 1.13, 1.14, 1.15
+The versions of Kubernetes that are supported for an Istio version is shown on the [Platform Setup](/docs/setup/platform-setup/). While Istio may work on versions outside of those listed, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
--- a/content/en/docs/setup/platform-setup/_index.md
+++ b/content/en/docs/setup/platform-setup/_index.md
@@ -7,12 +7,4 @@ aliases:
 keywords: [platform-setup]
 ---
 
-We recommend using a version of Kubernetes that has been tested with Istio:
-
-|Istio Version       | Kubernetes Version
-|--------------------|-----------------------
-|1.6                 | 1.15, 1.16, 1.17, 1.18
-|1.5                 | 1.14, 1.15, 1.16
-|1.4                 | 1.13, 1.14, 1.15
-
-For more information, see the [Versioning Policy](/about/version-policy/).
+For more information on support Kubernetes versions, see the [Versioning Policy](/about/version-policy/).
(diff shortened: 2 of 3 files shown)


content/en/about/version-policy/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+For example, the data plane version **1.0.0** could be used with control plane version **1.0.0** or **1.1.0**.
+
+This implies that the control plane must be upgraded first. For more information, see the [Upgrade](/docs/setup/upgrade/) documentation.
+
+## Kubernetes versions
+
+The following table indicates the versions of Kubernetes that are supported for various Istio versions. While Istio may work on versions outside of those listed below, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
+
+|Istio Version       | Kubernetes Version
```
frankbu, line 36: This table seems out of place in this doc. Everything else is just generic rules, but this table is listing release specific information. Do we even need such a historical summary? Even if we do, I'm not sure this doc is a good place for it.

content/en/about/version-policy/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+The data plane (both sidecars and gateways) must be either the same version as the control plane, or at most 1 minor version older.
+
+For example, the data plane version **1.0.0** could be used with control plane version **1.0.0** or **1.1.0**.
+
+This implies that the control plane must be upgraded first. For more information, see the [Upgrade](/docs/setup/upgrade/) documentation.
+
+## Kubernetes versions
+
+The following table indicates the versions of Kubernetes that are supported for various Istio versions. While Istio may work on versions outside of those listed below, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
```
frankbu, line 34: How about:
```suggestion
The versions of Kubernetes that are supported for an Istio version is shown on the [install page](https://istio.io/docs/setup/install/). While Istio may work on versions outside of those listed, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
```

sdake: @frankbu @howardjohn and I were working on the same topic and we weren't aware of each other's work. I have started an RFC to obtain some TOC agreement on our objectives. See: https://docs.google.com/document/d/14Xfe9V0vv8d7W75c21d-cY2XyK6xvRH0_Kg8j_zfh4Y/edit#. John and I were likely thinking alike that a table helps (us) map the Istio releases to Ks8 releases, so we can develop a holistic policy.

content/en/about/version-policy/index.md
sdake: wrong place: @frankbu @howardjohn and I were working on the same topic and we weren't aware of each other's work. I have started an RFC to obtain some TOC agreement on our objectives. See: https://docs.google.com/document/d/14Xfe9V0vv8d7W75c21d-cY2XyK6xvRH0_Kg8j_zfh4Y/edit#. John and I were likely thinking alike that a table helps (us) map the Istio releases to Ks8 releases, so we can develop a holistic policy. :)

sdake:
Related PR: https://github.com/istio/istio.io/pull/7160/

content/en/about/version-policy/index.md
```diff
@@ -0,0 +1,40 @@
...
+For example, `istioctl` version **1.0.1** could be used with control plane version **1.0.0** and data plane version **1.0.6**.
+
+### Control/Data Plane
+
+The data plane (both sidecars and gateways) must be either the same version as the control plane, or at most 1 minor version older.
+
+For example, the data plane version **1.0.0** could be used with control plane version **1.0.0** or **1.1.0**.
+
+This implies that the control plane must be upgraded first. For more information, see the [Upgrade](/docs/setup/upgrade/) documentation.
+
```
linsun, line 31: does this imply people can't upgrade from 1.4 to 1.6?  Also, does this imply multiple control planes should be used when skipping 1 minor version upgrade.

linsun:
@howardjohn this looks good from my perspective.  thank you for the PR!

cc @istio/technical-oversight-committee

content/en/about/version-policy/index.md
linsun: it would be good to provide some guidance msg around this.

howardjohn: Yes! I believe this has always been out policy, and it is currently documented as well: https://istio.io/docs/setup/upgrade/cni-helm-upgrade/ "Istio does NOT support skip level upgrades. Only upgrades from 1.4 to 1.5 are supported. If you are on an older version, please upgrade to 1.4 first."

This is explicitly not mentioning use of multiple control planes to skip minor versions, because (1) this PR is meant to capture our current policy, which is no skip level updates (2) I am not (yet) convinced its a good idea

content/en/about/version-policy/index.md
frankbu: If we do want a table like this, I think it should appear with the releases, not in this general policy doc. How about putting the table on this page: https://istio.io/news/releases/. You can also link to it from this doc.

content/en/about/version-policy/index.md
linsun: Ok so multiple cps are recommended with 1 minor version upgrade.  They are not recommended for skip level update till further testing proves it.

frankbu:
@howardjohn Is this PR ready to merge? I still think you should first move the k8s table out to the releases section as I suggested above: https://github.com/istio/istio.io/pull/7154#discussion_r415851266. Otherwise it LGTM.

howardjohn:
We are still debating in T&R/TOC, will report back soon

content/en/about/version-policy/index.md
howardjohn: The release landing page seems like an odd location for this to me. I don't think many go there? I can put it on the install page

frankbu: My thought is that it's a summary of the k8s versions that each release note says it supports, so that's really where the table belongs, and it keeps release specific info out of the mainline docs.

I agree that people don't go to the release landing page often, but that doesn't matter if we link it from the places where users will go, like from this page and possibly other places where it might make sense.

content/en/about/version-policy/index.md
```diff
@@ -0,0 +1,40 @@
...
+## Component Version Skew
+
+Various components of Istio that interact with each other must be kept within supported version ranges. Below describes the required version skews. Where possible, we always recommend using the same version between all components.
+
+### Istioctl
+
+We currently recommend using `istioctl` within the same minor version as both the control plane and data plane.
+
+For example, `istioctl` version **1.0.1** could be used with control plane version **1.0.0** and data plane version **1.0.6**.
```
rshriram, line 22: This gets rather confusing very soon imo. How about something like “ istioctl works with same release of istio as well all the patch releases associated with it” or something along those lines.

content/en/about/version-policy/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Control/Data Plane
+
+The data plane (both sidecars and gateways) must be either the same version as the control plane, or at most 1 minor version older.
+
```
rshriram, line 27: Same as above. A control plane of a release is compatible with only the data plane from the previous release or something along those lines

content/en/about/version-policy/index.md
```diff
@@ -0,0 +1,40 @@
...
+## Kubernetes versions
+
+The following table indicates the versions of Kubernetes that are supported for various Istio versions. While Istio may work on versions outside of those listed below, we do not test the stability or safety of these versions. Sticking within the supported version range is highly recommended.
+
+|Istio Version       | Kubernetes Version
+|--------------------|-----------------------
+|1.6                 | 1.15, 1.16, 1.17, 1.18
+|1.5                 | 1.14, 1.15, 1.16
+|1.4                 | 1.13, 1.14, 1.15
```
rshriram, line 40: As frank says, I think this deserves a doc on its own or a page on its own for historical compatibility list.

content/en/about/version-policy/index.md
```diff
@@ -0,0 +1,40 @@
...
+Istio provides support for each minor release for 3 months after the next minor release. For example, Istio 1.0 support will be dropped 3 months after Istio 1.2 is released. See [Build & Release Cadence](/about/release-cadence/) for more details.
+
+## Component Version Skew
+
+Various components of Istio that interact with each other must be kept within supported version ranges. Below describes the required version skews. Where possible, we always recommend using the same version between all components.
+
+### Istioctl
+
+We currently recommend using `istioctl` within the same minor version as both the control plane and data plane.
+
```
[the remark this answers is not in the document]
howardjohn: "same version" could mean 1.0.0 and 1.0.1 are not compatible though, which is not the case?

content/en/docs/setup/platform-setup/_index.md
```diff
@@ -7,6 +7,12 @@ aliases:
 keywords: [platform-setup]
 ---
 
-{{< tip >}}
-Istio {{< istio_version >}} has been tested with these Kubernetes releases: {{< supported_kubernetes_versions >}}.
-{{< /tip >}}
+We recommend using a version of Kubernetes that has been tested with Istio:
+
+|Istio Version       | Kubernetes Version
```
frankbu, line 12: I don't think this is a good place for this table because this is specifically the Istio 1.6 setup page, so users shouldn't need to look for it in a table of historical releases. All they need is the tip that says what versions of kube they need to be using with this release.

If you absolutely do not want to put it on the release landing page, as I suggested, then I prefer you keep it in the version-policy doc, where you had it, although I think we're more likely to forget to update it there when we add a new release. Putting it with all the release notes will make it easier to remember to add a new line at the same time when we add a new release.

smawson:
FWIW here's is kubernetes version skew policy: https://kubernetes.io/docs/setup/release/version-skew-policy/

They use major.minor.patch in the same way as we describe here.

sdake:
@smawson K8s's document looks pretty good, although I am not sure we can directly use their simple policy, as we have a slightly different policy with is more complex.

@howardjohn - I'd encourage finishing this document as a version skew doc - off the install page.

I do agree with @frankbu that Istio should continue to document the K8s versions it is compatible with, but also keep a historical table if possible. I know the history is in the archives, although the archives are not very discoverable. That may be a problem solvable in a different way.

Cheers,
-steve

kramvan1:
+1 Looks like a good start, needs the finishing touches.

PR #7155, opened 2020-04-24
Add Kind Example

shaneutt:
The purpose of this PR is to add an example of deploying Istio in [KIND](https://github.com/kubernetes-sigs/kind) for testing Istio. The example covers using a `LoadBalancer` type `Service` in this environment as well as setting up certificates and configuring Istio gateways with SSL.

### DIFF
--- a/content/en/docs/examples/local-testing-with-kind/index.md
+++ b/content/en/docs/examples/local-testing-with-kind/index.md
@@ -1,0 +1,389 @@
+---
+title: Testing Istio Locally With Kubernetes In Docker
+description: Example of deploying Istio locally using Kubernetes In Docker for testing and evaluation.
+weight: 10
+keywords:
+- kubernetes
+- kind
+- metallb
+- cert-manager
+- ingress
+---
+
+This example uses Kubernetes In Docker ([KIND][kind]) to deploy Istio locally. The cluster produced is intended to be used for further testing, evaluation and experimentation with Istio.
+
+## Before you begin
+
+Sections in this example provide a walk through of deploying a `kind` cluster similar to how the [Istio Kind Documentation](/docs/setup/platform-setup/kind) recommends, but with several important follow up steps to deploy certificate management and Istio gateways with TLS. Before proceeding, familiarize yourself with that documentation to make sure you have a general working knowledge of `kind`.
+
+Make sure you have `jq` [installed][jq] on your local system for `JSON` processing.
+
+## Deploying Kubernetes
+
+Create a default Kubernetes cluster using `kind` with a single node:
+
+{{< text bash >}}
+$ kind create cluster --name istio-testing
+{{< /text >}}
+
+{{< tip >}}
+If you're interested in building a multi-node cluster, or otherwise further customizing the cluster setup (e.g. different Kubernetes versions, more worker nodes) then be sure to check the [Kind Cluster Configuration Documentation](https://kind.sigs.k8s.io/docs/user/quick-start#configuring-your-kind-cluster) which will help you explore the available options.
+{{< /tip >}}
+
+Once the cluster has completed deployment switch your [kubectl][kubectl] context to start using it:
+
+{{< text bash >}}
+$ kubectl cluster-info --context kind-istio-testing
+{{< /text >}}
+
+And then test access to the cluster by retrieving resources to make sure everything is working (e.g. `kubectl get all -A`).
+
+## Enabling external load balancers
+
+By default, Istio creates a [Load Balancer Service][servicelb] which resolves with an external address. For the purposes of this example, use [MetalLB][metallb] to resolve the `LoadBalancer` type `Service` resources to an address available on the Docker network from the host machine.
+
+{{< tip >}}
+For this example assume that the default Docker network is used with an IP range of `172.17.255.1-172.17.255.250` (if you ran `kind` with no specific configuration, this should be the default). If however you are using an alternative network configuration, you'll need to make the appropriate updates to the `addresses` field in the below MetalLB configuration before deploying.
+{{< /tip >}}
+
+First pick the latest release and deploy the MetalLB components:
+
+{{< text bash >}}
+$ export LATEST_METALLB_RELEASE="$(curl -s https://api.github.com/repos/metallb/metallb/releases/latest | jq -r '.tag_name')"
+$ kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/namespace.yaml
+$ kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/metallb.yaml
+{{< /text >}}
+
+Next, generate a secret for MetalLB:
+
+{{< text bash >}}
+$ kubectl create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
+{{< /text >}}
+
+Then, add a configuration file for MetalLB so it knows what protocol to use and what address range is available to it:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config
+  namespace: metallb-system
+data:
+  config: |
+    address-pools:
+    - name: default
+      protocol: layer2
+      addresses:
+      - 172.17.255.1-172.17.255.250
+EOF
+{{< /text >}}
+
+Now the external IP addresses for `LoadBalancer` type `Service` resources can be provisioned.
+
+## Certificates
+
+A default deployment of Istio serves only HTTP traffic, which can be very limiting for testing. For this example use [Cert Manager][certmanager] to provide a `ClusterIssuer`, which can sign certificates for testing HTTPS ingress through gateways.
+
+{{< tip >}}
+This example uses a self-signed certificate for simplicity (all testing is done over the local network). If you're adapting the examples here to a different kind of environment keep in mind that you'll need to change what kind of issuer you use. You may want to review other documentation such as the [Ingress With Cert Manager Task](/docs/ops/integrations/certmanager/) for your own adaptations.
+{{< /tip >}}
+
+First deploy the Cert Manager components themselves:
+
+{{< text bash >}}
+$ export LATEST_CERTMANAGER_RELEASE="$(curl -s https://api.github.com/repos/jetstack/cert-manager/releases/latest | jq -r '.tag_name')"
+$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/${LATEST_CERTMANAGER_RELEASE}/cert-manager.yaml
+{{< /text >}}
+
+{{< tip >}}
+Make sure to wait for all `Deployments` in the `cert-manager` namespace to be `READY`. If you try to create a `ClusterIssuer` and the webhook manager is not `READY`, you may receive 500 errors from the API until it's resolved.
+{{< /tip >}}
+
+Now, provide a `ClusterIssuer` for the certs:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: cert-manager.io/v1alpha2
+kind: ClusterIssuer
+metadata:
+  name: selfsigned-issuer
+spec:
+  selfSigned: {}
+EOF
+{{< /text >}}
+
+All set! Now certificates for `Gateway` resources can be more easily managed.
+
+## Deploying Istio
+
+After all the setup is done, it's time to deploy Istio.
+
+{{< tip >}}
+Make sure you have `istoctl` [installed](/docs/ops/diagnostic-tools/istioctl/#before-you-begin)!
+{{< /tip >}}
+
+Start by using `istioctl` to deploy the base Istio installation with [SDS](/docs/tasks/traffic-management/ingress/secure-ingress/) and [Ingress with HTTPS](/docs/ops/integrations/certmanager/) enabled:
+
+{{< text bash >}}
+$ istioctl manifest apply \
+  --set values.gateways.istio-ingressgateway.sds.enabled=true \
+  --set values.global.k8sIngress.enabled=true \
+  --set values.global.k8sIngress.enableHttps=true \
+  --set values.global.k8sIngress.gatewayName=ingressgateway
+{{< /text >}}
+
+Since HTTPS is enabled for the ingress `Gateway` make sure to create a `Certificate` for it:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: ingress-cert
+  namespace: istio-system
+spec:
+  secretName: ingress-cert
+  issuerRef:
+    name: selfsigned-issuer
+    kind: ClusterIssuer
+  commonName: examples.istio.io
+  dnsNames:
+  - '*.istio.io'
+EOF
+{{< /text >}}
+
+And patch the ingress `Gateway` to configure it for the `Certificate`:
+
+{{< text bash >}}
+$ kubectl -n istio-system patch gateway istio-autogenerated-k8s-ingress --type=json \
+  -p='[{"op": "replace", "path": "/spec/servers/1/tls", "value": {"credentialName": "ingress-cert", "mode": "SIMPLE", "privateKey": "sds", "serverCertificate": "sds"}}]'
+{{< /text >}}
+
+At this point you should be able to reach the default `Gateway` over both HTTP and HTTPS (but note, that you'll receive a 404 from the `Gateway` at this point and that's expected):
+
+{{< text bash >}}
+$ export ISTIO_GATEWAY_IP=$(kubectl -n istio-system get svc istio-ingressgateway --template='{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}')
+$ curl -v http://${ISTIO_GATEWAY_IP}
+$ curl -k -v https://${ISTIO_GATEWAY_IP}
+{{< /text >}}
+
+Now testing can start!
+
+## Testing traffic management
+
+In this section there are a couple of small examples which can assist in testing and prove things are working with basic traffic management.
+
+### HTTP test
+
+For this test a simple [NGINX][nginx] web server is deployed to test HTTP traffic.
+
+You can easily deploy an NGINX pod and service with:
+
+{{< text bash >}}
+$ kubectl run nginx --image nginx --wait=true
+$ kubectl expose pod nginx --port 80 --target-port 80
+{{< /text >}}
+
+Create a `VirtualService` (that uses the default provided gateway) to access the server from outside the cluster:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: networking.istio.io/v1beta1
+kind: VirtualService
+metadata:
+  name: nginx
+spec:
+  gateways:
+  - istio-system/istio-ingressgateway
+  hosts:
+  - 'http.examples.istio.io'
+  http:
+  - route:
+    - destination:
+        host: nginx
+        port:
+          number: 80
+EOF
+{{< /text >}}
+
+That's it! You should be able to access the web server now with `curl`:
+
+{{< text bash >}}
+$ export ISTIO_GATEWAY_IP=$(kubectl -n istio-system get svc istio-ingressgateway --template='{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}')
+$ curl --fail -H 'Host: http.examples.istio.io' http://${ISTIO_GATEWAY_IP}
+{{< /text >}}
+
+{{< tip >}}
+It can take a few moments before the route works. Make sure to retry a few times if you can't reach the NGINX server immediately.
+{{< /tip >}}
+
+When things are working properly the response output will include: `<h1>Welcome to nginx!</h1>`
+
+### HTTPS test
+
+This example is a variant of the HTTP test with TLS termination at the ingress level.
+
+Create and expose a new [NGINX][nginx] web server to test:
+
+{{< text bash >}}
+$ kubectl create namespace nginx
+$ kubectl -n nginx run nginx --image nginx --wait=true
+$ kubectl -n nginx expose pod nginx --port 80 --target-port 80
+{{< /text >}}
+
+Create a `Gateway` and `VirtualService` to serve traffic to the web server:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: nginx
+  namespace: nginx
+spec:
+  selector:
+    istio: ingressgateway
+  servers:
+  - port:
+      number: 443
+      name: https
+      protocol: HTTPS
+    tls:
+      credentialName: ingress-cert
+      mode: SIMPLE
+      privateKey: sds
+      serverCertificate: sds
+    hosts:
+    - "https.example.istio.io"
+EOF
+{{< /text >}}
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: nginx
+  namespace: nginx
+spec:
+  hosts:
+  - "https.example.istio.io"
+  gateways:
+  - nginx
+  http:
+  - route:
+    - destination:
+        port:
+          number: 80
+        host: nginx
+EOF
+{{< /text >}}
+
+Now it should be possible to access your web server over the external IP using the `https.example.istio.io` domain:
+
+{{< text bash >}}
+$ export ISTIO_GATEWAY_IP=$(kubectl -n istio-system get svc istio-ingressgateway --template='{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}')
+$ curl -k --resolve https.example.istio.io:443:${ISTIO_GATEWAY_IP} https://https.example.istio.io
+{{< /text >}}
+
+If everything worked properly you should see `<h1>Welcome to nginx!</h1>` in the output.
+
+### TLS passthrough test
+
+A simple test of TLS passthrough can be done by patching through to the Kubernetes API itself.
+
+For this create a new `Gateway` with TLS mode set to `PASSTHROUGH`:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: kubernetes
+spec:
+  selector:
+    istio: ingressgateway
+  servers:
+  - port:
+      number: 443
+      name: https
+      protocol: HTTPS
+    tls:
+      mode: PASSTHROUGH
+    hosts:
+    - examples.istio.io
+EOF
+{{< /text >}}
+
+And then a `VirtualService` to attach to the Kubernetes API via the internal `Service`:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: networking.istio.io/v1beta1
+kind: VirtualService
+metadata:
+  name: kubernetes
+spec:
+  gateways:
+  - kubernetes
+  hosts:
+  - examples.istio.io
+  tls:
+  - match:
+    - port: 443
+      sni_hosts:
+      - examples.istio.io
+    route:
+    - destination:
+        host: kubernetes
+        port:
+          number: 443
+EOF
+{{< /text >}}
+
+Then you should be able to access the Kubernetes API through the `Gateway`:
+
+{{< text bash >}}
+$ export ISTIO_GATEWAY_IP=$(kubectl -n istio-system get svc istio-ingressgateway --template='{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}')
+$ curl -H 'Content-Type: application/json' -w '\n' -k --resolve examples.istio.io:443:${ISTIO_GATEWAY_IP} https://examples.istio.io/version
+{{< /text >}}
+
+If all went well the result of the above `curl` command should be the `/version` output from the Kubernetes API, similar to:
+
+{{< text json >}}
+{
+  "major": "1",
+  "minor": "17",
+  "gitVersion": "v1.17.2",
+  "gitTreeState": "clean",
+  "goVersion": "go1.13.5",
+  "compiler": "gc",
+  "platform": "linux/amd64"
+}
+{{< /text >}}
+
+## What's next
+
+In this example a simple and relatively complete Istio enabled Kubernetes cluster was created.
+
+From here you can look at some of the other [examples](/docs/examples) and [tasks](/docs/tasks) and try them on this cluster and otherwise test and experiment with Istio.
+
+If you have any need for an integration testing environment which includes Istio for your own tests these examples may provide a foundation which can be expanded upon.
+
+## Cleanup
+
+Cleanup is as simple as deleting the `kind` cluster:
+
+{{< text bash >}}
+$ kind delete cluster --name istio-testing
+{{< /text >}}
+
+[jq]:https://github.com/stedolan/jq
+[kind]:https://github.com/kubernetes-sigs/kind
+[docker]:https://docs.docker.com/
+[kubectl]:https://kubernetes.io/docs/reference/kubectl/overview/
+[servicelb]:https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/
+[metallb]:https://metallb.universe.tf/
+[certmanager]:https://cert-manager.io/
+[nginx]:https://nginx.com
--- a/.spelling
+++ b/.spelling
@@ -398,6 +398,8 @@ memcached-2's
 Mesos
 mesos-dns
 metadata
+metallb
+MetalLB
 MicroK8s
 microservice
 microservices


content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+Make sure that you have [Docker][docker] and [Kind][kindquickstart] installed and that you are able to create and use kind clusters and connect to them with [Kubectl][kubectl].
+
+## Deploying Kubernetes
+
+For this example we'll be creating a default Kubernetes cluster using [Kind][kind] with a single node:
+
+```shell
+kind create cluster --name istio-testing
+```
```
frankbu, line 28: This (and all other code blocks) needs to be changed to use `{{< text }}>`. See documentation [here](https://istio.io/about/contribute/code-blocks/).
```suggestion
{{< text bash >}}
$ kind create cluster --name istio-testing
{{< /text >}}
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+This example uses Kubernetes In Docker ([KIND][kind]) to deploy Istio locally. The cluster produced is intended to be a supplement for further testing, evaluation and experimentation with Istio.
+
+## Before you begin
+
+Make sure that you have [Docker][docker] and [Kind][kindquickstart] installed and that you are able to create and use kind clusters and connect to them with [Kubectl][kubectl].
+
+## Deploying Kubernetes
+
+For this example we'll be creating a default Kubernetes cluster using [Kind][kind] with a single node:
```
frankbu, line 24: Istio documentation avoids "we" terminology. Please change "we" to "you" or reword, here and throughout the doc.
```suggestion
Create a default Kubernetes cluster using [Kind][kind] with a single node:
```

shaneutt: Updated all the spots, thank you :)

content/en/docs/examples/local-testing-with-kind/index.md
shaneutt: Updated all these spots, thank you :)

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,388 @@
+---
+title: Testing Istio Locally With Kubernetes In Docker
+description: Example of deploying Istio locally using Kubernetes In Docker for testing and evaluation
```
frankbu, line 3: Missing period is preventing the preview (https://app.netlify.com/sites/preliminary-istio/deploys/5ea6e54d7b751300067079ce) from generating.
```suggestion
description: Example of deploying Istio locally using Kubernetes In Docker for testing and evaluation.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,388 @@
...
+For this example we assume that the default docker network is used with an IP range of `172.17.255.1-172.17.255.250` (if you ran `kind` with no specifical configuration, this should be the default). If however you are using an alternative network configuration, you'll need to make the appropriate updates to the `addresses` field in the below MetalLB configuration before deploying.
+{{< /tip >}}
+
+First we'll pick the latest release and deploy the MetalLB components:
+
+{{< text bash >}}
+export LATEST_METALLB_RELEASE="$(curl -s https://api.github.com/repos/metallb/metallb/releases/latest | jq -r '.tag_name')"
+kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/namespace.yaml
+kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/metallb.yaml
+{{< /text >}}
```
frankbu, line 56: Every line this is a command (vs. output or extended multi-line command) must start with `$ `.
```suggestion
{{< text bash >}}
$ export LATEST_METALLB_RELEASE="$(curl -s https://api.github.com/repos/metallb/metallb/releases/latest | jq -r '.tag_name')"
$ kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/namespace.yaml
$ kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/metallb.yaml
{{< /text >}}
```

content/en/docs/examples/local-testing-with-kind/index.md
shaneutt: :+1: updated

content/en/docs/examples/local-testing-with-kind/index.md
shaneutt: :+1:

content/en/docs/examples/local-testing-with-kind/index.md
shaneutt: :+1:

shaneutt:
I'm working right now on getting my google cla done, we have a corporate one, I'll get that set up asap.

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,391 @@
...
+- kind
+- metallb
+- cert-manager
+- ingress
+test: true
+---
+
+## About
+
+This example uses Kubernetes In Docker ([KIND][kind]) to deploy Istio locally. The cluster produced is intended to be a supplement for further testing, evaluation and experimentation with Istio.
```
frankbu, line 16:
```suggestion
This example uses Kubernetes In Docker ([KIND][kind]) to deploy Istio locally. The cluster produced is intended to be used for further testing, evaluation and experimentation with Istio.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,391 @@
...
+
+## Before you begin
+
+Make sure that you have [Docker][docker] and [Kind][kindquickstart] installed and that you are able to create and use kind clusters and connect to them with [Kubectl][kubectl].
+
+## Deploying Kubernetes
```
frankbu, line 22: This and the previous section seems to duplicate material in the platform setup instructions for `kind`: https://preliminary.istio.io/docs/setup/platform-setup/kind/. Should it reuse, or at least explain how it relates to, those instructions?

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,391 @@
...
+      addresses:
+      - 172.17.255.1-172.17.255.250
+EOF
+{{< /text >}}
+
+Now the external IP addresses for `LoadBalancer` type `Service` resources can be provisioned.
+
+## Certificates
+
+A default deployment if Istio will only serve HTTP traffic, which can be very limiting for testing. For this example use [CertManager][certmanager] to provide a `ClusterIssuer` which can sign certificates for us for testing HTTPS ingress through our gateways.
```
frankbu, line 87:
```suggestion
A default deployment of Istio will only serve HTTP traffic, which can be very limiting for testing. For this example use [CertManager][certmanager] to provide a `ClusterIssuer` which can sign certificates for us for testing HTTPS ingress through our gateways.
```

content/en/docs/examples/local-testing-with-kind/index.md
shaneutt: I've tried adding a reference below in a "tip" section that points to the document and recommends familiarizing yourself with it before proceeding. Let me know what you think.

shaneutt:
@frankbu I noticed that the lint test failed on `/docs` type links:

https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio.io/7155/lint_istio.io/6083

However I noticed other docs use these kinds of links. Is there something more I need to do in this case?

frankbu:
@shaneutt the `/doc/...` links are internal links within the istio.io site. The ones that are failing are files that have recently moved. Moved files have redirect aliases from the old to new URL, so they can still be accessed with the old URL, but the lint checker doesn't follow aliases and requires all internal links be kept up to date. To fix the broken ones in you PR, prepend `preliminary.istio.io` to the broken URL and then paste in a browser to see where it redirects to.

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,390 @@
...
+title: Testing Istio Locally With Kubernetes In Docker
+description: Example of deploying Istio locally using Kubernetes In Docker for testing and evaluation.
+weight: 10
+keywords:
+- kubernetes
+- kind
+- metallb
+- cert-manager
+- ingress
+test: true
```
frankbu, line 11: Remove this line, unless you are planning to write a test program for the page: https://github.com/istio/istio.io/tree/master/tests#testing-istioio-content.
If you remove this field, you can also remove the generated snip.sh file.

frankbu: Before you delete snip.sh, you might want to take a look at the reported lint errors, to see if they are legitimate bad things in your code blocks or not.
```
If there are any lint errors in the generated snip.sh file, it means that a command in the
index.md file is not following bash best practices. Because we are extracting the commands
from the markdown file into a script file, we get the added benefit of lint checking of the
commands that appear in the docs.
```

frankbu: Looking at the lint error in the snip.sh, the problem is that you are using `<< EOF` in some of the commands. The doc framework doesn't recognize this as a multi-line bash command, it only handles `<<EOF` (no whitespace). The generated snip.sh ends up being bad, but more importantly, the copy button on the generated web page will also not work properly. Please change them to `<<EOF`.

shaneutt: Changed, thank you :+1:

shaneutt:
> preliminary.istio.io

Thank you for your help, I've followed your instructions and made changes according to `make lint` and I believe these should be taken care of, I'll be on the look out for any more stragglers.

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,389 @@
...
+{{< /text >}}
+
+All set! Now certificates for `Gateway` resources can be more easily managed.
+
+## Deploying Istio
+
+All the setup is done, it's time to deploy Istio.
+
+{{< tip >}}
+Make sure you have `istoctl` [installed](/docs/setup/)!
```
frankbu, line 123: I think this is a better place to reference.
```suggestion
Make sure you have `istoctl` [installed](/docs/ops/diagnostic-tools/istioctl/#before-you-begin)!
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,389 @@
...
+---
+
+## About
+
+This example uses Kubernetes In Docker ([KIND][kind]) to deploy Istio locally. The cluster produced is intended to be used for further testing, evaluation and experimentation with Istio.
+
+## Deploying Kubernetes
+
+{{< tip >}}
+This section deploys a [Kind][kind] cluster similar to how the [Istio Kind Documentation](/docs/setup/platform-setup/kind) recommends, but with several important follow up steps to deploy certificate management and Istio gateways with TLS. Familiarize yourself with that documentation and [Kind][kind] itself to make sure you have a general working knowledge of `kind` before proceeding.
```
frankbu, line 20: If you check the [preview](https://deploy-preview-7155--preliminary-istio.netlify.app/docs/examples/local-testing-with-kind/) you'll see that some of the links (like this `[Kind][kind]` one) are not displaying correctly. I think the problem may be that the square bracket syntax doesn't work inside tips (although it does seem to work fine outside of them). So anyway, I think you'll need to use the expanded form `[title](target link)` in the places where it doesn't work.

That said, for this particular case, I don't think you should be making every use of kind a clickable link. Just the first occurrence would be enough. After that, just enclose it in backticks: `kind`.

shaneutt: Sounds good, I made those changes as suggested :+1:

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,388 @@
...
+weight: 10
+keywords:
+- kubernetes
+- kind
+- metallb
+- cert-manager
+- ingress
+---
+
+## About
```
frankbu, line 13: No need for a heading for the overview.
```suggestion
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+This example uses Kubernetes In Docker ([KIND][kind]) to deploy Istio locally. The cluster produced is intended to be used for further testing, evaluation and experimentation with Istio.
+
+## Deploying Kubernetes
+
+{{< tip >}}
```
frankbu, line 19: One more suggestion would be to not make this a tip, but rather move this paragraph to a `Before you begin` section and change the first sentence to: The following instructions deploy a `kind` cluster similar to ...

shaneutt: Sounds good to me, changed :+1:

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+- metallb
+- cert-manager
+- ingress
+---
+
+This example uses Kubernetes In Docker ([KIND][kind]) to deploy Istio locally. The cluster produced is intended to be used for further testing, evaluation and experimentation with Istio.
+
+## Before You Begin
+
+Sections in this example provide a walk through of deploying a `kind` cluster similar to how the [Istio Kind Documentation](/docs/setup/platform-setup/kind) recommends, but with several important follow up steps to deploy certificate management and Istio gateways with TLS. Familiarize yourself with that documentation and `kind` itself to make sure you have a general working knowledge of `kind` before proceeding.
```
Rachael-Graham, line 17:
```suggestion
Sections in this example provide a walk through of deploying a `kind` cluster similar to how the [Istio Kind Documentation](/docs/setup/platform-setup/kind) recommends, but with several important follow up steps to deploy certificate management and Istio gateways with TLS. Before proceeding, familiarize yourself with that documentation to make sure you have a general working knowledge of `kind`.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+- kubernetes
+- kind
+- metallb
+- cert-manager
+- ingress
+---
+
+This example uses Kubernetes In Docker ([KIND][kind]) to deploy Istio locally. The cluster produced is intended to be used for further testing, evaluation and experimentation with Istio.
+
+## Before You Begin
```
Rachael-Graham, line 15: sentence case for headings throughout
```suggestion
## Before you begin
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+Once the cluster has completed deployment switch your [kubectl][kubectl] context to start using it:
+
+{{< text bash >}}
+$ kubectl cluster-info --context kind-istio-testing
+{{< /text >}}
+
+And then test access to the cluster by retrieving resources to make sure everything is working (e.g. `kubectl get all -A`).
+
+## Enabling External Load Balancers
```
Rachael-Graham, line 39:
```suggestion
## Enabling external load balancers
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+  "minor": "17",
+  "gitVersion": "v1.17.2",
+  "gitTreeState": "clean",
+  "goVersion": "go1.13.5",
+  "compiler": "gc",
+  "platform": "linux/amd64"
+}
+{{< /text >}}
+
+## What's Next
```
Rachael-Graham, line 364:
```suggestion
## What's next
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+{{< text bash >}}
+$ kubectl cluster-info --context kind-istio-testing
+{{< /text >}}
+
+And then test access to the cluster by retrieving resources to make sure everything is working (e.g. `kubectl get all -A`).
+
+## Enabling External Load Balancers
+
+By default Istio will create a [Load Balancer Service][servicelb] which will be resolved with an external address. For the purposes of this example use [MetalLB][metallb] to resolve these to an address available on the docker network from the host machine.
```
Rachael-Graham, line 41:
```suggestion
By default, Istio creates a [Load Balancer Service][servicelb] which resolves with an external address. For the purposes of this example, use [MetalLB][metallb] to resolve the `LoadBalancer` type `Service` resources to an address available on the Docker network from the host machine.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+{{< tip >}}
+For this example assume that the default docker network is used with an IP range of `172.17.255.1-172.17.255.250` (if you ran `kind` with no specific configuration, this should be the default). If however you are using an alternative network configuration, you'll need to make the appropriate updates to the `addresses` field in the below MetalLB configuration before deploying.
```
Rachael-Graham, line 44:
```suggestion
For this example assume that the default Docker network is used with an IP range of `172.17.255.1-172.17.255.250` (if you ran `kind` with no specific configuration, this should be the default). If however you are using an alternative network configuration, you'll need to make the appropriate updates to the `addresses` field in the below MetalLB configuration before deploying.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+First pick the latest release and deploy the MetalLB components:
+
+{{< text bash >}}
+$ export LATEST_METALLB_RELEASE="$(curl -s https://api.github.com/repos/metallb/metallb/releases/latest | jq -r '.tag_name')"
+$ kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/namespace.yaml
+$ kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/metallb.yaml
+{{< /text >}}
+
+We need to generate a secret for MetalLB:
```
Rachael-Graham, line 55: avoid us/we/etc
```suggestion
Next, generate a secret for MetalLB:
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+      addresses:
+      - 172.17.255.1-172.17.255.250
+EOF
+{{< /text >}}
+
+Now the external IP addresses for `LoadBalancer` type `Service` resources can be provisioned.
+
+## Certificates
+
+A default deployment of Istio will only serve HTTP traffic, which can be very limiting for testing. For this example use [Cert Manager][certmanager] to provide a `ClusterIssuer` which can sign certificates for us for testing HTTPS ingress through our gateways.
```
Rachael-Graham, line 84:
```suggestion
A default deployment of Istio serves only HTTP traffic, which can be very limiting for testing. For this example use [Cert Manager][certmanager] to provide a `ClusterIssuer`, which can sign certificates for testing HTTPS ingress through gateways.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+First deploy the Cert Manager components themselves:
+
+{{< text bash >}}
+$ export LATEST_CERTMANAGER_RELEASE="$(curl -s https://api.github.com/repos/jetstack/cert-manager/releases/latest | jq -r '.tag_name')"
+$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/${LATEST_CERTMANAGER_RELEASE}/cert-manager.yaml
+{{< /text >}}
+
+{{< tip >}}
+Make sure to wait for all `Deployments` in the `cert-manager` namespace to be `READY`, if you try to create a `ClusterIssuer` and the webhook manager is not `READY`, you may receive 500 errors from the API until it's resolved.
```
Rachael-Graham, line 98:
```suggestion
Make sure to wait for all `Deployments` in the `cert-manager` namespace to be `READY`. If you try to create a `ClusterIssuer` and the webhook manager is not `READY`, you may receive 500 errors from the API until it's resolved.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+{{< /tip >}}
+
+Now let's provide a `ClusterIssuer` for our certs:
```
Rachael-Graham, line 101:
```suggestion
Now, provide a `ClusterIssuer` for the certs:
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+spec:
+  selfSigned: {}
+EOF
+{{< /text >}}
+
+All set! Now certificates for `Gateway` resources can be more easily managed.
+
+## Deploying Istio
+
+All the setup is done, it's time to deploy Istio.
```
Rachael-Graham, line 118:
```suggestion
After all the setup is done, it's time to deploy Istio.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+{{< tip >}}
+Make sure you have `istoctl` [installed](/docs/ops/diagnostic-tools/istioctl/#before-you-begin)!
+{{< /tip >}}
+
+We'll start by using `istioctl` to deploy the base Istio installation with [SDS](/docs/tasks/traffic-management/ingress/secure-ingress/), [Ingress with HTTPS](/docs/ops/integrations/certmanager/) enabled:
```
Rachael-Graham, line 124:
```suggestion
Start by using `istioctl` to deploy the base Istio installation with [SDS](/docs/tasks/traffic-management/ingress/secure-ingress/) and [Ingress with HTTPS](/docs/ops/integrations/certmanager/) enabled:
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+{{< /text >}}
+
+And patch the ingress `Gateway` to configure it for the `Certificate`:
+
+{{< text bash >}}
+$ kubectl -n istio-system patch gateway istio-autogenerated-k8s-ingress --type=json \
+  -p='[{"op": "replace", "path": "/spec/servers/1/tls", "value": {"credentialName": "ingress-cert", "mode": "SIMPLE", "privateKey": "sds", "serverCertificate": "sds"}}]'
+{{< /text >}}
+
+If all went well, you could be able to reach the default `Gateway` over both HTTP and HTTPS (you'll receive a 404):
```
Rachael-Graham, line 161: is this saying that at this point you SHOULD be able to ping the gateway, or you should NOT be able to ping the gateway?

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+{{< text bash >}}
+$ export ISTIO_GATEWAY_IP=$(kubectl -n istio-system get svc istio-ingressgateway --template='{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}')
+$ curl -v http://${ISTIO_GATEWAY_IP}
+$ curl -k -v https://${ISTIO_GATEWAY_IP}
+{{< /text >}}
+
+Now testing can start!
+
+## Testing Traffic Management
```
Rachael-Graham, line 171:
```suggestion
## Testing traffic management
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+In this section there are a couple of small examples which can assist in testing and prove things are working with basic traffic management.
+
+### HTTP Test
```
Rachael-Graham, line 175:
```suggestion
### HTTP test
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+$ curl --fail -H 'Host: http.examples.istio.io' http://${ISTIO_GATEWAY_IP}
+{{< /text >}}
+
+{{< tip >}}
+It can take a few moments before the route will work, make sure to retry a few times over the coming seconds if you can't reach the NGinx server immediately.
+{{< /tip >}}
+
+When things are working properly the response output will include: `<h1>Welcome to nginx!</h1>`
+
+### HTTPS Test
```
Rachael-Graham, line 221:
```suggestion
### HTTPS test
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+Now it should be possible to access your web server over the external IP using the `https.example.istio.io` domain:
+
+{{< text bash >}}
+$ export ISTIO_GATEWAY_IP=$(kubectl -n istio-system get svc istio-ingressgateway --template='{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}')
+$ curl -k --resolve https.example.istio.io:443:${ISTIO_GATEWAY_IP} https://https.example.istio.io
+{{< /text >}}
+
+If everything worked properly you should see `<h1>Welcome to nginx!</h1>` in the output.
+
+### TLS Passthrough Test
```
Rachael-Graham, line 290:
```suggestion
### TLS passthrough test
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+## Testing Traffic Management
+
+In this section there are a couple of small examples which can assist in testing and prove things are working with basic traffic management.
+
+### HTTP Test
+
+For this test a simple [NGinx][nginx] web server will be deployed to test HTTP traffic.
+
+You can easily deploy an NGinx pod and service with:
```
Rachael-Graham, line 179:
```suggestion
For this test a simple [NGINX][nginx] web server is deployed to test HTTP traffic.

You can easily deploy an NGINX pod and service with:
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+{{< text bash >}}
+$ kubectl run nginx --image nginx --wait=true
+$ kubectl expose pod nginx --port 80 --target-port 80
+{{< /text >}}
+
+We'll create a `VirtualService` (that uses the default provided gateway) to access the server from outside the cluster:
```
Rachael-Graham, line 186:
```suggestion
Create a `VirtualService` (that uses the default provided gateway) to access the server from outside the cluster:
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+
+That's it! It should be able to access the web server now with `curl`:
+
+{{< text bash >}}
+$ export ISTIO_GATEWAY_IP=$(kubectl -n istio-system get svc istio-ingressgateway --template='{{range .status.loadBalancer.ingress}}{{.ip}}{{end}}')
+$ curl --fail -H 'Host: http.examples.istio.io' http://${ISTIO_GATEWAY_IP}
+{{< /text >}}
+
+{{< tip >}}
+It can take a few moments before the route will work, make sure to retry a few times over the coming seconds if you can't reach the NGinx server immediately.
```
Rachael-Graham, line 216:
```suggestion
It can take a few moments before the route works. Make sure to retry a few times if you can't reach the NGINX server immediately.
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+{{< /tip >}}
+
+When things are working properly the response output will include: `<h1>Welcome to nginx!</h1>`
+
+### HTTPS Test
+
+This example is a variant of the HTTP test with TLS termination at the ingress level.
+
+Create and expose a new [NGinx][nginx] web server to test:
```
Rachael-Graham, line 225:
```suggestion
Create and expose a new [NGINX][nginx] web server to test:
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,386 @@
...
+  http:
+  - route:
+    - destination:
+        host: nginx
+        port:
+          number: 80
+EOF
+{{< /text >}}
+
+That's it! It should be able to access the web server now with `curl`:
```
Rachael-Graham, line 208:
```suggestion
That's it! You should be able to access the web server now with `curl`:
```

content/en/docs/examples/local-testing-with-kind/index.md
shaneutt: Hopefully this clarifies:
```suggestion
At this point you should be able to reach the default `Gateway` over both HTTP and HTTPS (but note, that you'll receive a 404 from the `Gateway` at this point, not a 200):
```
Let me know what you think?

shaneutt:
I believe I have addressed all your feedback @Rachael-Graham, thank you for it! Let me know what you think.

frankbu:
@shaneutt The doc LGTM now. Only thing missing is you still need to sign the CLA: https://github.com/istio/istio.io/pull/7155#issuecomment-619197168.

Also, before merging an example, we need confirmation that it has been tested by somebody manually or with a test program. Alternatively, we can move it to the blog, but personally I'd like to keep it in the examples section.

shaneutt:
> @shaneutt The doc LGTM now. Only thing missing is you still need to sign the CLA: [#7155 (comment)](https://github.com/istio/istio.io/pull/7155#issuecomment-619197168).

I'm working on the CLA right now, we have a corporate one I'm trying to get added to so I'll try to get that taken care of asap.

> Also, before merging an example, we need confirmation that it has been tested by somebody manually or with a test program. Alternatively, we can move it to the blog, but personally I'd like to keep it in the examples section.

I have tested this manually and ran it, I'll do another testing pass now that we've made some tweaks just to make sure everything is looking :100:

shaneutt:
@Rachael-Graham & @frankbu I re-ran my original tests and it appears that in the latest version of `istioctl` the name of the default gateway changed from `istio-system/ingressgateway` to `istio-system/istio-ingressgateway` since I last tested this. As such I've updated the documentation and tested this again and it's working properly. All other sections worked properly in my test.

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,389 @@
...
+$ kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/${LATEST_METALLB_RELEASE}/manifests/metallb.yaml
+{{< /text >}}
+
+Next, generate a secret for MetalLB:
+
+{{< text bash >}}
+$ kubectl create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
+{{< /text >}}
+
+And add a configuration file for MetalLB so it knows what protocol to use and what address range is available to it:
```
frankbu, line 63: Small grammatical tweak to retrigger netlify which failed because it's been flaky lately.
```suggestion
Then, add a configuration file for MetalLB so it knows what protocol to use and what address range is available to it:
```

content/en/docs/examples/local-testing-with-kind/index.md
```diff
@@ -0,0 +1,389 @@
...
+After all the setup is done, it's time to deploy Istio.
+
+{{< tip >}}
+Make sure you have `istoctl` [installed](/docs/ops/diagnostic-tools/istioctl/#before-you-begin)!
+{{< /tip >}}
+
+Start by using `istioctl` to deploy the base Istio installation with [SDS](/docs/tasks/traffic-management/ingress/secure-ingress/) and [Ingress with HTTPS](/docs/ops/integrations/certmanager/) enabled:
+
+{{< text bash >}}
+$ istioctl manifest apply \
```
frankbu, line 129: I just realized that this is outdated and fails to work on latest 1.6 builds. For one, sds is now enabled by default and not sure about the k8sIngress stuff (cc @howardjohn).

This would need to work with https://github.com/istio/istio/releases/tag/1.6.0-beta.0, before we can merge to master.

I think you might need to tweak some of it.

frankbu:
@howardjohn I see your point. What I like about this doc is that it's a nice quick-start type all-in-one setup guide that produces a more functional cluster than our quick start does. I think you do bring up a good point that this might not be appropriate for an Example, because of the platform and other component specifics. I had been thinking that maybe it would be better suited as a Blog, and now I'm thinking that even more so. WDYT?

howardjohn:
I have no problem with this as a blog, as its a point in time snapshot (ie no maintenance)

frankbu:
@shaneutt Could you please move this from Examples to Blog. The only significant difference is that you'll need to add your name as author.

The real difference for us, as John pointed out, is that a blog can be outdated and stop working in a future release. If we make it an example, somebody would now be responsible for testing and updating it in future releases of Istio. The 1.6 incompatibility that I noticed, above, is an example of what may happen next release again.

Btw, the more you can point to other docs to do certain steps (e.g., install Istio), the less likely your blog will fail to work in the future. But, I also realize that there needs to be a balance between giving the instructions and pointing the reader all over the place at each step.

shaneutt:
> I don't think this documentation is appropriate here. If there is additional context on why it is can you add it to the PR description?

This documentation was very intentionally created as an "example" as per the bar set for what examples seems to come from the tagline:

"A variety of fully working example uses for Istio that you can experiment with."

In this context "fully working" seems to suggest that it represents something with end-to-end functionality, and as you can see from the documentation provided the "that you can experiment with" was taken very obviously in spirit, but also practically exampled through some basic tests you can run on the deployment.

From my review the contents of this example do not seem to stray far from the scope established by other examples like `bookinfo`, wherein your ultimate goal is to access a web app (in this case more simple just an NGinx instance) and providing all the resources and configurations prescriptively along the way needed to reach that goal.

>     * We already document how to setup kind, and other platforms, at https://istio.io/docs/setup/platform-setup/kind/

Yes, but that documentation isn't exactly "fully working" in the sense that you can't route HTTPS traffic through Istio into the cluster. This documentation makes specific reference to that documentation to point out that we're using kind very pointedly for the purposes of this example, but suggests reviewing that documentation as well.

>     * MetalLB is completely orthogonal to Istio. At best we should link to metalLB's documentation. There are many applications that "extend" Istio, and we cannot maintain full documentation for all of them. Instead we should be documenting what Istio specific changes are needed to integrate with these products (in metal lb case - there is none). Please see https://preliminary.istio.io/docs/ops/integrations/ for some existing examples.

MetalLB (as stated in the document) is a means to an end. In other examples the provisioning of `LoadBalancer` type `Service` resources is a given, but in a kind cluster it is not. This was the chosen implemented to enable a "fully working" example, but unlike the other integrations you've linked this isn't related directly to Istio but instead to Kubernetes.

Should people writing examples not feel free to include setup for components that don't directly relate to Istio itself, but relate ultimately to the ecosystem that is being exampled that Istio is being used in? There seems to be precent with [MySQL](https://github.com/istio/istio.io/pull/7155#pullrequestreview-403760678) and some other components throughout examples.

>     * Same comment for cert-manager. We should be documenting Istio, not cert-manager. The documentation on how to integrate Istio with cert-manager can be found in the link above

The reason for not simply linking directly to the existing documentation is that without specifics this wont be a "fully working" example, in the sense that the user would need to figure out cert-manager from the cert-manager documentation, and make choices based on the options available there. This provides a solution that should work in any environment, and even provides tips about how one could experiment beyond what's provided. I would like to submit that the specificity provided here is key to making this example simple and quick and allows them to get the desired result quickly, and go back and investigate how the underlying bits work later if desired.

>     * We are re-documenting how to install Istio. We already have an install document, we don't need to duplicate this just because we are now documenting `kind` - it is the same steps for any platform. If there were kind specific settings, they would belong in the platform setup page

This makes specific choices relevant to the goal of having fully working HTTPS gateways upon completion. Is it not valid to provide something that can be learned about elsewhere in example documentation? Is there not some way I could improve this section such that it would be more prescriptive or informational?

>     * Next we go on to document a bunch of stuff about traffic routing, which is covered already in https://istio.io/docs/tasks/traffic-management/

That documentation is great in the case that an end-user is coming to the docs trying to very specifically understand how traffic routing is done. But in this case a more prescriptive example is provided. Is that not basically the spirit of "example" documentation, such as to provide a specific implementation for a higher-level goal?

Look at [this part of the bookinfo example](https://istio.io/docs/examples/bookinfo/#apply-default-destination-rules) for instance. In this example specific destination rules are provided could you not also argue that would be a place where we could have instead linked to existing documentation and/or expanded that documentation to cover the needs here? But then would it not be harder to follow for the person who wants to start with an end result (a working webserver) and then work back through the implementation as opposed to needing to know how to come up with these manifests themselves?

> Note the above is my opinion - others may disagree. I appreciate the work you have put in here but I feel like it would be better spent by enhancing existing documents rather than creating a brand new one.
>
> @shaneutt Could you please move this from Examples to Blog. The only significant difference is that you'll need to add your name as author.
>
> The real difference for us, as John pointed out, is that a blog can be outdated and stop working in a future release. If we make it an example, somebody would now be responsible for testing and updating it in future releases of Istio. The 1.6 incompatibility that I noticed, above, is an example of what may happen next release again.

I would submit that "example" documentation takes a high level goal and provides a very specific set of implementations to reach that goal. To that end, I believe what I have submitted qualifies as an example and if possible I would like to keep this as an example rather than a blog post.

If you are not persuaded by my explanations and comparisons above, could you please help me to better understand what qualifies as an example and under what conditions you would accept something as example documentation?

If there's a criteria I have not met for what an example ultimately is, and what it takes to get one accepted here, I would like to know what that is so that I can have the opportunity to contribute real examples, if somehow different from what I've provided so far here.

shaneutt:
> The real difference for us, as John pointed out, is that a blog can be outdated and stop working in a future release. If we make it an example, somebody would now be responsible for testing and updating it in future releases of Istio. The 1.6 incompatibility that I noticed, above, is an example of what may happen next release again.

Could this problem not be solved another way? Can the guarantee of backwards compatibility be made with the other examples that already exist? Could perhaps documentation have to be specifically tagged with the versions it's been tested with, and only display for the versions provided? To me it seems that you _would_ want to test these things continue to work in newer versions, no?

howardjohn:
> Could this problem not be solved another way? Can the guarantee of backwards compatibility be made with the other examples that already exist? Could perhaps documentation have to be specifically tagged with the versions it's been tested with, and only display for the versions provided? To me it seems that you would want to test these things continue to work in newer versions, no? And if they don't, make the necessary version-specific corrections for the docs?

Yes, we absolutely want to make sure that the Istio docs are accurate as Istio changes. With this PR, we need to do *far* more than that. Now when cert-manager updates, we need to update the doc. When metallb, nginx, kind, ... update, we need to update the docs. 

Even if it was purely Istio, we have a task that describes a large set of this already: https://preliminary.istio.io/docs/tasks/traffic-management/ingress/secure-ingress/. Anytime we change how this works, now we have 2 places to change instead of just one. Its also confusing to users in my opinion. If I want to set up https should I use this guide? What if I don't want to use kind or cert-manager, now the example only sort of works.

> If you are not persuaded by my explanations and comparisons above, could you please help me to better understand what qualifies as an example and under what conditions you would accept something as example documentation?

I am not a docs maintainer so my opinions don't hold any more weight than yours here, but personally the whole "examples" section seems like something we should get rid of for the reasons I mentioned above

howardjohn:
Put another way, why do we have "Examples" and "Tasks" when an example is "A variety of fully working example uses for Istio that you can experiment with." and a Task is also literally a step by step copy+pastable example of uses of Istio features you can experiment with

joejulian:
As a side note (I work with Shane and have been following along out of interest), I'd tried to experiment with Istio before and the documentation was so vague and hard to find or follow that I gave up. Shane didn't have that luxury and was required to get it working in order to write up an evaluation. This PR is a response to his similar frustration with the documentation for this project and an attempt to make it better for future users.

If this documentation had existed when I tried, I would have not given up and I would have been successful. If you have criteria for documentation, I suggest you re-evaluate those criteria because they're lacking from the perspective of a new user.

joejulian:
"I suggest you re-evaluate" ... sorry, that sounds way more prescriptive than I intended. I don't mean to be antagonistic.

shaneutt:
@googlebot I signed it!

shaneutt:
@googlebot I signed it!

shaneutt:
@frankbu & @howardjohn I have taken care of the CLA, this PR is ready for re-review.

I would appreciate if you could review my above comments and provide a path forward where I could make modifications to this documentation that for you would make it acceptable as example documentation as it was originally intended.

Thank you for your time and considerations.

frankbu:
@joejulian thanks for the feedback. You're definitely not the first one to tell us that our getting started documentation is bad.

We've been struggling all along with the exact purpose of the examples section of the docs. Over the last couple of years we've removed as many docs from that section as we have added new ones. We've had as many complaints about docs that our outdated and don't work as lack of docs. The problem, I suppose, is that Istio has had a lot of changes, some breaking changes (not so good), but mostly just a lot of improved and different ways to do things. As a result, maintaining the docs has been a bit of a nightmare.

To try and get this under control, we've recently started to put in place a framework for testing docs, to make sure they work as described (https://github.com/istio/istio.io/blob/master/test/README.md). @shaneutt, you ran into this, when you initially copied another example md file to use as a template and it had the `test=true` settings in its metadata, which resulted in a snip file being generated for your doc.

So, now the question of what to do with this doc, which personally I think a very useful getting started for evaluation setup. There are several ways we could approach it:

1. Turn it into a blog so there is no guarantee it is still working after release 1.6, although it can be updated and kept working and referenced from the mainline docs. We do have a few examples where we've done that, including [this one](https://istio.io/blog/2017/0.1-canary/) that I wrote myself, some time ago.
2. Generate the snips and write a test for it. I think this will be complicated by the fact that the framework is more tuned to testing istio features vs. installing components, but it may be possible.
3. Think about publishing it somewhere else, and adding a section in the Istio docs (e.g. community examples) that is just a bunch of links to helpful documents. I suppose some of these links could be to Istio.io blogs.

I'm thinking that maybe option 1 is the easiest, and we can look at also doing number 3, where this blog would then be one of the links. Let me know what you guys think.

Also, most importantly, the article still needs to be updated to make sure it works with Istio 1.6, which will be the target for publishing it.

shaneutt:
> I'm thinking that maybe option 1 is the easiest, and we can look at also doing number 3, where this blog would then be one of the links. Let me know what you guys think.
> 
> Also, most importantly, the article still needs to be updated to make sure it works with Istio 1.6, which will be the target for publishing it.

@frankbu Your comment seems to suggest that you would prefer 1 or 3, but personally I am in favor of number 2, updating this for 1.6 and then adding the necessary tests. What are your thoughts?

frankbu:
@shaneutt If you are willing to write a test, I'm OK with putting it in the examples section, at least initially. If we end up restructuring the examples section in the future, it might have to be moved, but having a test that confirms it's working as written will always be great, even if it moves to blog (which have no guarantees) or moves somewhere else entirely. Eventually, some kind of "certified working" stamp on documents, based on the `test: true` attribute is our goal for docs on the site, regardless of where they live.

shaneutt:
> @shaneutt If you are willing to write a test, I'm OK with putting it in the examples section, at least initially. If we end up restructuring the examples section in the future, it might have to be moved, but having a test that confirms it's working as written will always be great, even if it moves to blog (which have no guarantees) or moves somewhere else entirely. Eventually, some kind of "certified working" stamp on documents, based on the `test: true` attribute is our goal for docs on the site, regardless of where they live.

This sounds good to me, thank you @frankbu :+1: 

For the moment I'm going to close this as I don't have cycles to do this in the next several days, but the intention will be to open/re-open in the future with the tests.

Thanks again everyone involved for your time. :)

joejulian:
Shane's heading out for paternity leave. What if this iteration was merged with the agreement that when he gets back he would add the testing required? Would that be satisfactory?

PR #7160, opened 2020-04-26
Update supported releases

### LINKED ISSUE #7146: Update the release tested.
Currently the docs has: `Istio 1.6 has been tested with these Kubernetes releases: 1.14, 1.15, 1.16.`

I'm not sure if we are going to simply shift one to the right, or extend to include 1.18.

ericvn:
Fixes #7146

Values from https://github.com/istio/istio/issues/23184


linsun:
Have we done some minimal testing on 1.18?

PR #7162, opened 2020-04-27
Setup for GKE

gnarled-cipher:


ericvn:
@adammil2000 Looks like you approved this one and no on did an `ok to test` so it never merged. Not sure if this should still go in after all the community items or not.

adammil2000:
Adding label to let testing occur on this PR...

PR #7172, opened 2020-04-27
lint_site.sh: Don't print error if "public" directory doesn't exist.

justinpettit:
If the "public" directory doesn't exist, "lint_site.sh" prints an error when running the "find" command.

### DIFF
--- a/scripts/lint_site.sh
+++ b/scripts/lint_site.sh
@@ -148,32 +148,35 @@ for lang in $LANGS; do
     fi
 done
 
-if [[ ${#SKIP_LANGS[@]} -ne 0 ]]; then
-    printf -v find_exclude " -name %s -prune -o" "${SKIP_LANGS[@]}"; read -r -a find_exclude <<< "$find_exclude"
-fi
-while IFS= read -r -d '' f; do
-    if grep -H -n -i -e blockquote "${f}"; then
-        echo "Ensure content only uses {{< tip >}}, {{< warning >}}, {{< idea >}}, and {{< quote >}} instead of block quotes"
-        FAILED=1
-    fi
-
-    #if grep -H -n -e "\"https://github.*#L[0-9]*\"" "${f}"; then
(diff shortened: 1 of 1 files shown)


scripts/lint_site.sh
```diff
@@ -151,17 +151,20 @@ done
...
+            echo "Ensure content only uses {{< tip >}}, {{< warning >}}, {{< idea >}}, and {{< quote >}} instead of block quotes"
+            FAILED=1
+        fi
+
+        #if grep -H -n -e "\"https://github.*#L[0-9]*\"" "${f}"; then
+        #    echo "Ensure content doesn't use links to specific lines in GitHub files as those are too brittle"
+        #    FAILED=1
+        #fi
+    done < <(find ./public "${find_exclude[@]}" -type f -name '*.html' -print0)
+fi
```
frankbu, line 167: Shouldn't this `fi` move down to after line 182, since none of the link checking is possible if there's no ./public?

scripts/lint_site.sh
```diff
@@ -151,17 +151,20 @@ done
...
-        echo "Ensure content only uses {{< tip >}}, {{< warning >}}, {{< idea >}}, and {{< quote >}} instead of block quotes"
-        FAILED=1
-    fi
 
-    #if grep -H -n -e "\"https://github.*#L[0-9]*\"" "${f}"; then
-    #    echo "Ensure content doesn't use links to specific lines in GitHub files as those are too brittle"
-    #    FAILED=1
-    #fi
-done < <(find ./public "${find_exclude[@]}" -type f -name '*.html' -print0)
+if [ -d ./public ]; then
```
frankbu, line 155: Actually, this could move up a few lines too, because line 151 isn't needed either, if there's no ./public.

scripts/lint_site.sh
justinpettit: Good point.

scripts/lint_site.sh
justinpettit: Yes.  I'll send out a revised version.  Thanks!

PR #7176, opened 2020-04-28
Add authz-tcp user guide test.

justinpettit:
Add authz-tcp user guide test.


frankbu:
There are several commands in index.md that the bash linter doesn't like. They look something like this:
```
kubectl exec $(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name}) -c sleep ...
```
Even though it's fine, bash best practice is to put quotes around arguments who's value could potentially include whitespace. Although it's fine in this case, you'll need to add quotes to all of them to make the linter happy.
```
kubectl exec "$(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name})" -c sleep
```

justinpettit:
> There are several commands in index.md that the bash linter doesn't like. They look something like this:
> 
> ```
> kubectl exec $(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name}) -c sleep ...
> ```
> 
> Even though it's fine, bash best practice is to put quotes around arguments who's value could potentially include whitespace. Although it's fine in this case, you'll need to add quotes to all of them to make the linter happy.
> 
> ```
> kubectl exec "$(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name})" -c sleep
> ```

Thanks.  Still learning about all the checks I need to run before submitting a PR.  I pushed a new version that I think should be lint-clean.

justinpettit:
Most of the errors seem transient.  The only failure I see related to the test introduced is:
```
2020-04-28T19:11:21.817293Z	info	ci	Checking pods ready...
2020-04-28T19:11:21.817380Z	info	ci	Checking pods ready...
2020-04-28T19:11:21.822915Z	info	ci	  [ 0]                        sleep-857f5b4f44-4hfrl         Running (container not ready: 'istio-proxy')
2020-04-28T19:11:21.822964Z	info	ci	  [ 1]                     tcp-echo-7ddc95fc6b-pt54h         Running (container not ready: 'istio-proxy')
    TestAuthzTCP: builder.go:127: failed waiting for pods to start: %v timeout while waiting (last error: 2 errors occurred:
        	* foo/sleep-857f5b4f44-4hfrl: container not ready: 'istio-proxy'
        	* foo/tcp-echo-7ddc95fc6b-pt54h: container not ready: 'istio-proxy'
        
        )
```
It looks like one of the pods failed to start.  Is there something else I should be doing?  The test does pass on my local system.

frankbu:
@justinpettit The test framework has been very flaky for the last couple of days, so if you're test passes locally, you should probably just wait for the framework problems to be fixed, and assume there's nothing else you need to do.

tests/security/authz_tcp/authz_tcp_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"istio.io/istio/pkg/test/framework"
+
+	"istio.io/istio.io/pkg/test/istioio"
+)
+
+//https://istio.io/docs/tasks/security/authorization/authz-tcp/
+//https://github.com/istio/istio.io/blob/release-1.5/content/en/docs/tasks/security/authorization/authz-tcp/index.md
+func TestAuthzTCP(t *testing.T) {
+	framework.
```
frankbu, line 28: @justinpettit I think it would be a good idea to add the following line to disable the test and then we should merge this PR. Otherwise, you're going to need to regen the snip file whenever the generated output is tweaked (for example, when #7233 merges). We can get it working in a followup PR.

```suggestion
	t.Skip("Not working yet")
	framework.
```

justinpettit: Thanks, Frank.  We're doing a bit more research on the underlying issue.  Let me see how that works out.  If there doesn't seem to be an underlying issue that needs to be addressed, I'll do resubmit with your suggestion.

tests/trafficmanagement/ingress/secure_ingress_sds_test.go
```diff
@@ -42,6 +42,7 @@ func TestSecureIngressSDS(t *testing.T) {
 	// Check the version of curl. This test requires the --retry-connrefused arg.
 	curl.RequireMinVersionOrFail(t, semver.MustParse("7.52.0"))
 
+	t.Skip("Skipping mTLS")
```
frankbu, line 45: Why skip this one? It's a little flaky, but does pass most of the time.

justinpettit: I was testing a theory that some of the tests were enabling mTLS in strict mode, which was causing this test to fail because the configuration was sticking around between tests.  Based on my local testing and this run, it looks like that's the case:
```
2020-05-08T05:32:21.937644Z	info	ci	Running command script scripts/authz_tcp.txt
2020-05-08T05:32:36.257556Z	info	ci	Creating snippets for input: scripts/authz_tcp.txt
2020-05-08T05:32:36.257646Z	info	ci	Running command script cleanup.sh
2020-05-08T05:33:20.376007Z	info	ci	Creating snippets for input: cleanup.sh
2020-05-08T05:33:20.376104Z	info	ci	=== DONE (passed):  Test: 'authz_tcp[TestAuthzTCP] (2m9.343649647s)' ===
--- PASS: TestAuthzTCP (129.34s)
```
(This test run failing was from a different test, not TestAuthzTCP.)

I was under the impression that the environment got cleaned up between runs.  (I thought this was brought up as one of the reasons that running the tests were slow.)

There are a couple of issues that I see:
    1) Should we be cleaning up the state between these tests?  (I'd think "yes", since it could lead to subtle cascading failures like this one.  However, we'll need to keep runtime in mind.)
    2) Why is this test failing with mTLS enabled?  On my team, there seems to be debate about whether this is expected behavior.  If this is expected behavior, in my opinion, we should update the task to handle this situation, since even though the task doesn't say to enable mTLS, someone could have mTLS enabled and be playing with these tasks, which would lead to confusion.

I'm still doing some investigation on my end about why this doesn't work with mTLS strict enabled.

justinpettit: I'm not suggesting that we disable mTLS in the tests.  😄

frankbu: Thanks for the explanation.

1. Yes, the tests are supposed to cleanup when they're finished, but there may be tests that are forgetting something in their cleanup script. Maybe some test is just doing what the task document says, which has something missing. Tasks that create their own namespace, and then delete the namespace at cleanup time are less likely to have an issue than ones that use the default namespace, or if they set global config. It sounds like that might be what you've found here. Only question is which test is doing it?
2. I don't know either if the TCP echo is supposed to work the same with or without mTLS, but I would think it should. If not, I guess what you're suggesting is that there should be another step in the Before you begin section to disable mTLS? That would make sense to me too.

Anyway, thanks for digging into this.

frankbu:
@justinpettit thanks for getting to the bottom with this test interference problem.

We really need to figure out why the test cluster is not always being reset to a clean slate between tests. I ran into a similar issue with one of the other tests, which I [fixed in a similar way](https://github.com/istio/istio.io/blob/affe99f97abab317723d657df5ce87160f435350/tests/util/samples.sh#L39) to what you did for peerauthentication here. We need to try to figure out what other tasks are leaving things in a bad state. Every task, is supposed to undo everything it configured in its cleanup section, so these may actually be bugs in the doc. We should probably also try to add something to the framework to enforce cleanup of cluster between tests (i.e., maybe we can check that the cluster state is back to the starting state at the end of each test), so even if there are buggy docs, they won't affect other docs.

justinpettit:
Thanks for merging the changes, Frank.  The "fix" wasn't very satisfying, but hopefully we can come up with a better way to cleanup between runs.

PR #7188, opened 2020-04-29
account for istiod scenario when troubleshooting mutatingwebhooks

tariq1890:


content/en/docs/ops/common-problems/injection/index.md
```diff
@@ -141,23 +141,37 @@ Warning  FailedCreate  3m (x17 over 8m)  replicaset-controller  Error creating:
 caused by an empty `caBundle` in the webhook configuration.
 
 Verify the `caBundle` in the `mutatingwebhookconfiguration` matches the
-   root certificate mounted in the `istio-sidecar-injector` pod.
+   root certificate mounted in the `istio-sidecar-injector` or `istiod`(if you're using istiod) pod.
```
frankbu, line 144: Do we need to document this (and the other changes, below) two ways? Can't it just be switched to `istiod` now?
```suggestion
   root certificate mounted in the `iistiod` pod.
```

howardjohn: Istiod is the only option in 1.6

tariq1890: ok

tariq1890:
What up with the netlify?

tariq1890:
@frankbu Thanks for merging the PR. Why did netlify fail?

frankbu:
@tariq1890 Don't know why, but netlify just seems to randomly fail sometimes, but currently the only way to rerun it is by pushing another commit, which is not very good if there's nothing else that needs to change in the PR. https://istio.slack.com/archives/C50V5EATT/p1588173408150300

PR #7204, opened 2020-04-30
Add linkinator to replace htmlproofer and increase linting speed

davidhauck:
Replacing htmlproofer with linkinator. Linkinator runs much faster than htmlproofer. On my computer, htmlproofer took close to 10 minutes. linkinator takes around 30s to 1 minute. Htmlproofer is still used for html validation and a few other things, though that only takes a couple of seconds to run.

Linkinator currently does not have a way to only run against certain files. This means we can't only run against english files at the moment. This is a feature that can be added to linkinator in the future. I think the speed increase would be worth it for now, and then we can add the ability to only scan english files in the future.

The scripts/grab_reference_docs.sh script is also edited to have a consistent format for linking. Links now point to directories correctly as opposed to html files directly and relying on aliases / redirects.

There were also quite a few broken links that were found and fixed in this PR.


content/en/blog/2019/webhook/index.md
```diff
@@ -37,4 +37,4 @@ will not be able to alter the webhook configurations.
 and that the certificate chain used by the webhook server is valid. This reduces the errors
 that can occur before a server is ready or if a server has invalid certificates.
 
-To try this new feature, refer to the [Istio webhook management task](https://archive.istio.io/1.4/docs/tasks/security/webhook).
\ No newline at end of file
+To try this new feature, refer to the [Istio webhook management task](https://archive.istio.io/v1.4/docs/tasks/security/webhook).
```
frankbu, line 40: It's good that it caught broken links like this, but I'm wondering why it did. I think that the reason that htmlproofer missed it was because we have explicitly disabled checking of links to various sites, and most recently to external sites in general, because we have too many broken builds caused by temporary outages or other changes in sites we have no control over. I'm wondering if linkinator has an equivalent feature (ignore links to certain sites, or external sites in general). If not, we may end up being frustrated again by regular build failures.

frankbu:
I see this in the lint output:
```
Checking 3 external links...
Ran on 1360 files!

Running linkinator...
🤖 Successfully scanned 2839 links in 32.24 seconds.
```
32.24 seconds ... Nice!
The Checking 3 external links message seems strange though. Do you know what that's about?

davidhauck:
@frankbu HTML-Proofer seems to be looking at some external links for some reason. I'm not exactly sure why that would be happening since it's not running a link checker, so I'll look into that.

content/en/blog/2019/webhook/index.md
davidhauck: yes, if you look at lint_site.sh:183, we are ignoring the regex "^((?!localhost).)*$". This essentially ignores any site that isn't local. The new code is still respecting the CHECK_EXTERNAL_LINKS variable

davidhauck:
@frankbu Those three links are from the script checker in htmlproofer. That's making sure the external scripts we are referencing are valid. This is probably a good check to keep in place, as they are important scripts to run the site correctly.

PR #7206, opened 2020-04-30
update for istio users why gloo

christian-posta:
Clarify benefits to Istio users of using Gloo, hence the SDS integration

@dcberg


content/en/blog/2020/istio-sds-gloo/index.md
```diff
@@ -7,7 +7,7 @@ attribution: "Christian Posta (Solo.io)"
...
...nagement/request_processing/direct_response_action/), and [Open API Spec/Swagger and gRPC discovery](https://docs.solo.io/gloo/latest/installation/advanced_configuration/fds_mode/). This benefits an Istio user by having API Gateway capabilities at the edge of your mesh (or multiple clusters of a mesh) without having to try and hack this into the Istio ingress gateway with Lua, etc. [Gloo Enterprise](https://www.solo.io/products/gloo/) supports more sophisticated security edge requirements like [OIDC authentication](https://docs.solo.io/gloo/latest/guides/security/auth/oauth/), [OPA authorization](https://docs.solo.io/gloo/latest/guides/security/auth/opa/), [Web Application Fire walling (WAF)](https://docs.solo.io/gloo/latest/guides/security/waf/), rate limiting and others. There is also a custom-auth framework for building your own edge security protocols (common ones like non-standard security handshakes, HMAC, etc) into the ext-auth server that Gloo uses. A lot of Gloo users put Gloo at the edge for North/South concerns and [integrate with Istio for east-west traffic management](https://www.solo.io/blog/using-gloo-as-an-ingress-gateway-with-istio-and-mtls-updated-for-istio-1-1/).
```
elysdir, line 10: "to try and hack this into the Istio ingress gateway"

I would say that the phrase "try and" is more informal than we usually aim for; consider "try to" instead?

I might also suggest using a more specific term than "hack", but if you think the core audience for this will understand clearly what you mean by "hack this into", then fine to leave alone.

elysdir, line 10: I see that "North/South" here uses capital letters and a slash, while "east-west" uses lowercase letters and a hyphen. Is that difference intentional? If not, then I recommend using consistent capitalization and punctuation here. ...When I do a web search for related terms, I'm not seeing a lot of consistency across sites, but I think lowercase is more common than uppercase. I'm not sure which punctuation is best; up to you which to use, I think.

PR #7209, opened 2020-04-30
Restructure the cert management tasks.

myidpt:
Move the following two tasks under the `Security/Certificate Management` section:

* Plugging in existing CA Certificates
* Istio DNS Certificate Management


frankbu:
@myidpt You also need to fix the test programs for these moved docs.
```
    TestDNSCert: script.go:369: script scripts/dns_cert.txt returned an error: exit status 1. Output:
        bash: line 22: /home/prow/go/src/istio.io/istio.io/content/en/docs/tasks/security/dns-cert/snips.sh: No such file or directory
    TestDNSCert: script.go:369: script cleanup.sh returned an error: exit status 127. Output:
        bash: line 3: /home/prow/go/src/istio.io/istio.io/content/en/docs/tasks/security/dns-cert/snips.sh: No such file or directory

    TestPluginCACert: script.go:369: script create_ns_foo_with_httpbin_sleep.sh returned an error: exit status 1. Output:
        bash: line 6: /home/prow/go/src/istio.io/istio.io/content/en/docs/tasks/security/plugin-ca-cert/snips.sh: No such file or directory
    TestPluginCACert: script.go:369: script cleanup.sh returned an error: exit status 127. Output:
        bash: line 3: /home/prow/go/src/istio.io/istio.io/content/en/docs/tasks/security/plugin-ca-cert/snips.sh: No such file or 
```

frankbu:
@myidpt one more
```
    TestDNSCert: script.go:369: script scripts/dns_cert.txt returned an error: exit status 1. Output:
        bash: line 22: /home/prow/go/src/istio.io/istio.io/content/en/docs/tasks/security/dns-cert/snips.sh: No such file or directory
```

myidpt:
@frankbu Hi Frank, do you have any idea why I get this lint error?
```
- ./public/about/log/index.html
  *  2249:32: ERROR: ID id_2020_05 already defined (line 2249)
  *  2291:32: ERROR: ID id_2020_04 already defined (line 2291)
```
It seems the auto-generated log file contains errors. It's confusing to me.

ericvn:
You might want to do a `make shell` and within the shell do the `make lint` and then look at the file in question to see if that helps. I know when I need to figure out what line 72000 in the generated file points to, that's helped in the past.

content/en/docs/tasks/security/cert-management/plugin-ca-cert/index.md
```diff
@@ -4,7 +4,7 @@ description: Shows how system administrators can configure Istio's CA with an ex
 weight: 80
 keywords: [security,certificates]
 aliases:
-    - /docs/tasks/security/plugin-ca-cert/
+    - /docs/tasks/security/cert-management/plugin-ca-cert/
```
frankbu, line 7: This is wrong. An alias is the old location of a moved file.
```suggestion
    - /docs/tasks/security/plugin-ca-cert/
```

frankbu: Also note that you need to add an alias to the other doc you moved. Add the following a the top of `/docs/tasks/security/cert-management/dns-cert/index.md`.
```
aliases:
    - /docs/tasks/security/dns-cert/
```

frankbu:
@myidpt 
> It seems the auto-generated log file contains errors. It's confusing to me.

Yes, I agree this is confusing. It looks like the log generator is getting messed up, but I can't see anything wrong with your PR. I wonder if it's a bug in the generator, or maybe a branch incompatibility issue.

Either way, please make the alias changes, I suggested above, and then I will force merge your PR. Hopefully the log generator problem will go away after that, but if not, we can fix the generator bug in a separate PR.

content/en/docs/tasks/security/cert-management/plugin-ca-cert/index.md
myidpt: Sure. I'll do that.

frankbu:
Looks to me like there is a bug in the change log generator, causing the lint error. Seems like a timing issue where a commit for this PR is not in the same month as the latest merged commits. Notice in the preview, how the 04/30 (Fix references) commit is starting a new 04 table, above the 05 table that already exists for a different merged commit.

image

If so, I wonder if we're seeing this because @davidhauck changed the way htmlproofer is being invoked now? Maybe before it never spotted this issue?

@geeknoid WDYT?

frankbu:
I'm going to force merge this PR to see if the change log gens correctly after merging.

davidhauck:
I don't think this is caused by the changes to htmlproofer. The only thing I did is have htmlproofer skip the link checking step, everything else is working the same as before

frankbu:
So, after merging, the [change log](https://preliminary.istio.io/about/log/) is fine, so it really is just a problem with the generated preview. If it's a timing bug in the generator, it's strange that we've never noticed it before, esp. if htmlproofer was always doing these checks as @davidhauck confirmed. It would seem like this could happen early on in every new month ????

PR #7211, opened 2020-05-01
Update automated test TAG to match go.mod istio commit

ericvn:
Currently the tests are using the `latest` TAG which varies as builds are done, and was pointing to a 1.7 image earlier. This will get the test to use a known image that matches the go.mod commit for istio/istio, which will hopefully get test passage better (only a few flakey tests).


frankbu:
This one seems to work pretty well, but we need a policy about what builds the doc framework tests against. How and when do we change this, or can we come up with an automated way to change when appropriate? We need a way to make sure it's only using good (endorsed) builds. That said, we also need to make sure we catch istio/istio changes that break docs quickly.

PR #7212, opened 2020-05-01
Remove extraneous operator repo from grab_reference_docs.sh

ericvn:
I also ran the script manually to pick up recent changes. Changing the script did not change the manual changes.


ericvn:
@frankbu This one is also ready to go and should fix the failing automation task (fails since the operator directory was removed in an istio/istio PR).

PR #7214, opened 2020-05-01
Update test framework to use 1.6.0-beta.0

### LINKED ISSUE #7081: istio.io tests using sha for istio, but tests are using latest build images during run
If I unset the HUB and TAG and run the tests, the deployments try to use the gcr.io/istio-testing/?:1.6-alpha.{sha}. The deployment will fail because the images don't exist.

Looking at test-infra, the calls use TAG=latest.

ericvn:
Fixes #7081


ericvn:
@frankbu This one is ready for review. The
```
kubectl get peerauthentication --all-namespaces
NAMESPACE      NAME                          AGE
istio-system   grafana-ports-mtls-disabled   105s
```
is part of the demo profile (added a week or so ago).

Makefile.core.mk
```diff
@@ -33,12 +33,12 @@ export ISTIO_IMAGE_VERSION
 ISTIO_SHA ?= $(shell < ${ISTIOIO_GO}/go.mod grep 'istio.io/istio v' | cut -d'-' -f3)
 export ISTIO_SHA
 
-HUB ?= gcr.io/istio-testing
+HUB ?= gcr.io/istio-prerelease-testing
```
sdake, line 36: these should be overriden by prow and stay wit their current defaults.

Makefile.core.mk
```diff
@@ -33,12 +33,12 @@ export ISTIO_IMAGE_VERSION
...
 ifeq ($(HUB),)
   $(error "HUB cannot be empty")
 endif
 
-TAG ?= $(shell echo "${ISTIO_IMAGE_VERSION}.${ISTIO_SHA}")
+TAG ?= 1.6.0-beta.0
```
sdake, line 41: same with tag. What is `ISTIO_IMAGE_VERSION` used for? Is this the same thing as the container version?

prow/integ-suite-kind.sh
```diff
@@ -39,8 +39,8 @@ export TEST_ENV=kind
 # See https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster
 export PULL_POLICY=IfNotPresent
 
-export HUB=${HUB:-"gcr.io/istio-testing"}
-export TAG="${TAG:-"1.6-alpha.1d3b136325a5592511f72a9b041ea396bdbd5180"}"
+export HUB=${HUB:-"gcr.io/istio-prerelease-testing"}
```
sdake, line 42: these should be coming from prow.

Makefile.core.mk
```diff
@@ -33,12 +33,12 @@ export ISTIO_IMAGE_VERSION
 ISTIO_SHA ?= $(shell < ${ISTIOIO_GO}/go.mod grep 'istio.io/istio v' | cut -d'-' -f3)
```
howardjohn, line 33: We should be matching this, right? Why are we using the 1.6.0-beta.0 tag? I would expect this PR to do nothing more than update go.mod with the SHA of 1.6.0-beta.0

ericvn: @howardjohn If you know of a way to get from the sha in go.mod (short) to a valid image tag  long SHA, then we can do this on non-releases.

See my issue: https://github.com/istio/istio.io/issues/7081

If one unsets HUB and TAG one would hope things work, but the calculated image name does not, so all you get is a failure.  The CI/CD incorrectly passes in `latest` so the images can be found, but the tests are now running images from who knows where.

Makefile.core.mk
ericvn: The issue is when someone tries to run them locally, they don't know they need to override them, so things are broken. This was made so that tests run easily locally and in the CI/CD.

We should revisit this as part of the work for https://github.com/istio/istio.io/issues/7081.

Makefile.core.mk
ericvn: It was the old calculated name that never worked. The SHA was the short sha and they are named with the long sha.

Makefile.core.mk
howardjohn: @ericvn two possible options:

1. if we have istio/istio repo cloned, we can resolve short SHA -> long SHA
2. We can just have the full SHA defined in the makefile to match the go.mod. We can add a simple assertion they are kept in sync

prow/integ-suite-kind.sh
ericvn: comment above about running locally.

Makefile.core.mk
ericvn: @howardjohn Yes, I know how to get the long sha from the short. I can add code in the init.sh to do that. The main issue is that certain skas will have images from one image repository (those like 1.6-alpha.1d3b136325a5592511f72a9b041ea396bdbd5180) while some will have images from another repository (like 1.6.0-beta.0). I'm not sure that I can figure out which from the sha alone, other that to see if I can pull the image.

howardjohn: Every single SHA will have `gcr.io/istio-testing`. We shouldn't be using 1.6.0-beta.0, we should be using `f7c106d20f70a0f7ef9720c790cc5c20230c33fd`, which will exist as 1.6-alpha.f7c106d20f70a0f7ef9720c790cc5c20230c33fd. There is no need to special case beta/alpha releases

ericvn: OK. I looked and there is indeed a set of `1.6-alpha.f7c106d20f70a0f7ef9720c790cc5c20230c33fd` images which should match the `1.6.0-beta.0` images from the other repo. I'll make a change to calculate the longer SHA using the cloned repo and then we can remove that code. And it fixes my other issue.

ericvn: Drafting while I make those changes.

Makefile.core.mk
ericvn: Moved back to using daily build repo as it does contain the release builds as well.

Makefile.core.mk
ericvn: We now calculate the tag when run via the Makefile. There is an issue with the TAG coming in via the prow script so TAG is still hardcoded there.

prow/integ-suite-kind.sh
ericvn: Have gone back to the daily build as noted above.

Makefile.core.mk
ericvn: OK. If run locally without setting HUB or TAG, the make will use the daily build repo and calculates the long SHA from the short SHA in the go.mod file and will use that just fine.

For some reason, the values are not getting set when called via the prow script in the prow builds, so they are still hardcoded there.

ericvn: Actually, the `eval` is getting evaluated too early when using prow. Things work fine for me locally calling `prow/integ-suite-kind.sh` so not exactly sure what the problem is. More investigation coming. May have to find a makefile and prow expert.

sdake: Thankfully, you have both a Makefile expert and Prow expert in this review 👍
Insert <tab>`env` to display the environment between these. two lines: https://github.com/istio/istio.io/blob/master/tests/tests.mk#L37-L38.

Run through prow and capture the output.

Then please try adding a `-` preceeding this line: https://github.com/istio/istio.io/blob/master/Makefile.core.mk#L150

Run through prow and capture the output.

Diff the `env` output. This may yield the answer you are looking for.

Please note, there is a way to do this without prow which is a little difficult to use, but easier to capture: https://github.com/istio/istio/wiki/Working-with-Prow#testing-changes-locally I am not sure if this method is the same as the integration test you are running locally. The linked method mimics Istio's CI operation.
Cheers
-steve

ericvn: I will try that shortly. What I notice in the prow logs is the second line here:
```
+ make test.kube.presubmit
/bin/bash: line 0: cd: /home/prow/go/src/istio.io/istio: No such file or directory
ISTIO_SHA = f7c106d20f70
HUB = gcr.io/istio-testing
ISTIOIO_GO=/home/prow/go/src/istio.io/istio.io
ISTIO_GO=/home/prow/go/src/istio.io/istio
```
I don't see this when running locally. I believe that's the eval line failing when it's run early (before the init.sh has created the directory and cloned the repo there).

Local runs (where I see the cp: preserving times instead).
```
+ make test.kube.presubmit
cp: preserving times for '/home/.kube': Read-only file system
ISTIO_SHA = f7c106d20f70
HUB = gcr.io/istio-testing
ISTIOIO_GO=/work
```

ericvn: Thanks to @howardjohn and @sdake I was able to come up with a way to calculate the long SHA from the short sha in the go.mod file and use it within the Makefile. This allows the updating of the base framework from a new istio/istio commit by `go get istio.io/istio@<commit|laber>` followed by a `go mod tidy`. Local tests and prow will use the commit as part of the image name, so no additional changes should be required.

Makefile.core.mk
```diff
@@ -140,12 +135,26 @@ update_all: update_ref_docs update_examples
 foo2:
 	hugo version
 
-.PHONY: init
-init:
+.PHONY: init preinit
+init: preinit
```
sdake, line 139: just leaving a comment here for future reviewers digging in the history. Make is supposed to work LEFT->RIGHT, then TOP->BOTTOM. In prow, this is not the case. It appears prow is in some way running target operations out of roder. That is why this preinit step is here. This isn't a good long term solution but it will hold for now.

Makefile.core.mk
```diff
@@ -140,12 +135,26 @@ update_all: update_ref_docs update_examples
...
+	$(eval ISTIO_LONG_SHA := $(shell cd ${ISTIO_GO} && git rev-parse ${ISTIO_SHA}))
+	@ export ISTIO_LONG_SHA
```
sdake, line 141: extra space.

Makefile.core.mk
sdake: @ericvn - did you follow the top to bottom rule of make? preinit should proceed init.

Makefile.core.mk
ericvn: Fixed.

Makefile.core.mk
ericvn: Order fixed and added comment to Makefile as well.

PR #7220, opened 2020-05-03
Fix typo of PodDisruptionBudgets

GijsvanDulmen:
Fix a typo.


ericvn:
Please make the changes in the English branch only. The translation teams then pull the changes as they can. That allows PRs to get in quicker as the translation teams may take longer to do the approval.

I'll approve now, and the change will merge when a Chinese documentation approval is made. Or if you resubmit with only the English change, it will work right away.

PR #7232, opened 2020-05-06
Try removing helm entries from see also on getting started page

ericvn:
Fixes https://github.com/istio/istio/issues/23558


ericvn:
Looking at the netlify output, the removal of the keywords on the helm pages removed them from the see also on the getting-started page. Based on the process docs, they would have been removed from other pages Hugo might have determined would be a good fix.

PR #7233, opened 2020-05-06
Remove snippet gen (# $snippet) from all tests

frankbu:
This PR includes the following:

- Remove all `# $snippet` constructs from existing test scripts.
- Fixes and enables the `authz-http` flaky test, which was previously being skipped.
- Fixes a bug in the generated snip.sh files that previously didn't allow special characters in the generated bash variables.
- Add ? (wildcard) support to the `_verify_like` function.


frankbu:
Force merging to override netlify flaky failure

PR #7245, opened 2020-05-08
Remove step 3 GKE-ism for single network vm

sdake:


sdake:
The source for this one liner: https://stackoverflow.com/a/61685899/10350191

content/en/docs/examples/virtual-machines/single-network/index.md
```diff
@@ -99,12 +99,10 @@ following commands on a machine with cluster admin privileges:
...
 
 1. Generate a `cluster.env` configuration to deploy in the VMs. This file contains the Kubernetes cluster IP address ranges
-    to intercept and redirect via Envoy. You specify the CIDR range when you install Kubernetes as `servicesIpv4Cidr`.
-    Replace `$MY_ZONE` and `$MY_PROJECT` in the following example commands with the appropriate values to obtain the CIDR
-    after installation:
+    to intercept and redirect via Envoy.
 
     {{< text bash >}}
-    $ ISTIO_SERVICE_CIDR=$(gcloud container clusters describe $K8S_CLUSTER --zone $MY_ZONE --project $MY_PROJECT --format "value(servicesIpv4Cidr)")
+    $ ISTIO_SERVICE_CIDR=$(echo '{"apiVersion":"v1","kind":"Service","metadata":{"name":"tst"},"spec":{"clusterIP":"1.1.1.1","ports":[{"port":443}]}}' | kubectl apply -f - 2>&1 | sed 's/.*valid IPs is //')
```
rcaballeromx, line 105: Is there a platform agnostic command for readers to obtain the IP address and port for the environment variable instead of just setting the values manually? If there isn't a command to do so, please let readers know what values they need to modify explicitly in the text. Place the value they need to enter in `<>` to let readers know. For example, "Replace `<IP-Address>` and `<Port>` with the suitable values in the following command."

sdake: The IP address `1.1.1.1` is never to be used by any platform, so it is platform agnostic :). I could use `8.8.8.8` but I thought people might confuse that with DNS - which is again a routed address never to be used by another service.  The port `443` is just a random port I picked from thin air. It could be any port - but K8s requires a port to be defined, annoyingly.

sdake: For further detail, the way this works is if 1.1.1.1 is not in the service range (which is impossible) - then it is rejected. This rejection then displays the service CIDR which is seeded into a CIDR to be used later during istio's VM setup.

rcaballeromx: I'm not sure that using the rejection to get the value is the best way to go, but we can use it if it's the only platform-agnostic way for now.  @howardjohn WDYT?

howardjohn: https://1.1.1.1/dns/ :slightly_smiling_face: 

I couldn't find a better way to do it. this doesn't seem great but does seem to work

sdake: @howardjohn the other technique I attempted was using `nslookup kubernetes` from within a pod. This can be used to produce the network address prefix, but not the suffix for the CIDR. I was unable to determine a CIDR using that technique.

sdake:
Closing.

Superceeded by: https://github.com/istio/istio.io/pull/7267

PR #7257, opened 2020-05-11
Virtual machine integration

sdake:
We have a story coming together for virtual machine integration.


content/en/about/feature-stages/index.md
```diff
@@ -103,6 +103,7 @@ Below is our list of existing features and their current phases. This informatio
 | [Istio CNI plugin](/docs/setup/additional-setup/cni/) | Alpha
 | IPv6 support for Kubernetes | Alpha
 | [Distroless base images for Istio](/docs/ops/configuration/security/harden-docker-images/) | Alpha
+| [Virtual Machine Integration](/docs/docs/setup/install/virtual-machine/) | Alpha
```
rshriram, line 106: @liminw / @smawson

smawson: Do you think what we have is good enough to call alpha? Maybe worth just calling it alpha to force us to fix stuff...

PR #7258, opened 2020-05-11
Mark virtual machine credential distribution alpha

sdake:
The current virtual machine credential distribution code lacks
e2e testing, perfomrnace analysis, installation and task instructions.

As such it meets the requirement for alpha, but not beta as per our
feature status governance documentation;.


frankbu:
Although it's true that we've changed the implementation recently, it probably does not make sense for a feature that has been `Beta` since 1.0 to get demoted to `Alpha` in 1.6.

PR #7260, opened 2020-05-11
Label TCPTrafficShifting flaky (won't run)

ericvn:
This will remove the failing test for now. We can remove the label when the test is fixed.


frankbu:
I don't think we need an issue to track this because I'm planning to fix this test, and the other flaky circuit breaker test, later today or tomorrow at the latest.

frankbu:
Actually, let's not disable this test yet. It isn't blocking the merge, and it's good to be reminded what still needs fixing :-)

PR #7269, opened 2020-05-12
[release-1.6] Update galley to istiod for istioctl-analyze command

shamsher31:
To fix https://github.com/istio/istio/issues/23710


shamsher31:
@ericvn or @frankbu needs your approval.

PR #7295, opened 2020-05-13
Update security note to skip 1.5.3

mandarjog:


content/en/news/security/istio-security-2020-005/index.md
```diff
@@ -22,7 +22,7 @@ By sending a specially crafted packet, an attacker could trigger a Null Pointer
 ## Mitigation
 
 * For Istio 1.4.x deployments: update to [Istio 1.4.9](/news/releases/1.4.x/announcing-1.4.9) or later.
-* For Istio 1.5.x deployments: update to [Istio 1.5.3](/news/releases/1.5.x/announcing-1.5.3) or later.
+* For Istio 1.5.x deployments: update to [Istio 1.5.4](/news/releases/1.5.x/announcing-1.5.3) or later.
```
howardjohn, line 25: link is wrong but there is no correct link

brian-avery: Linter will fail on the bad link if we fix it. Until we can get the build out, then it's better to have a wrong link than to say that a fix is in a build it's not.

duderino, line 25:
```suggestion
* For Istio 1.5.x deployments: update to Istio 1.5.4 or later.
```

PR #7297, opened 2020-05-13
Make authn peer authn output matches impl

incfly:
fixes https://github.com/istio/istio/issues/23799

### DIFF
--- a/content/en/docs/tasks/security/authentication/authn-policy/index.md
+++ b/content/en/docs/tasks/security/authentication/authn-policy/index.md
@@ -66,11 +66,12 @@ sleep.legacy to httpbin.bar: 200
 sleep.legacy to httpbin.legacy: 200
 {{< /text >}}
 
-Verify there is no peer authentication policy in the system with the following command:
+Verify there is no peer authentication policy in the system except one for grafana dashboard with the following command:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/security/authentication/authn-policy/index.md
```diff
@@ -66,11 +66,12 @@ sleep.legacy to httpbin.bar: 200
 sleep.legacy to httpbin.legacy: 200
 {{< /text >}}
 
-Verify there is no peer authentication policy in the system with the following command:
+Verify there is no peer authentication policy in the system except one for grafana dashboard with the following command:
```
frankbu, line 69:
```suggestion
Verify there is no peer authentication policy in the system except one for Grafana dashboard with the following command:
```

ericvn:
I think there is an issue with this PR in that it talks about Grafana, however if one installs Istio as described at the top of the doc `istio manifest apply`, Grafana is not installed. It would be there if the demo profile was used.

incfly:
Thanks for catching out! Indeed i have my own bash shortcuts when install and remove istio and didn't realize this is the difference compared with `istioctl manifest apply`!

closing

PR #7312, opened 2020-05-14
Delete mixer filter stats as they no longer exist

nrjpoddar:


nrjpoddar:
@douglas-reid @mandarjog I'm assuming this is ok?

douglas-reid:
They might exist if you re-enable v1, but I don't think it matter much.  However, have the defaults enforced in `istio/istio` changed?

nrjpoddar:
@douglas-reid I followed default install instructions and didn’t see these stats. I think we should document the new default behavior.

nrjpoddar:
Can we get this merged?

douglas-reid:
My only concern is that the default collection keys have not actually changed. I'm fine with hiding them in the docs, but I don't think the actual enforced behavior will match.

douglas-reid:
To clarify: I'm not suggesting you'll see stats for these values, but that the stats inclusion list will still include them.

nrjpoddar:
Agree with that, I now understand what you mean. As a user if the stats don’t show up it felt odd to me as I don’t understand the underlying nuances of collection keys & inclusion list.

Your call, I’m fine either way.

douglas-reid:
Let's merge it so that we don't forget later. I don't think any use harm will come from this removal.

PR #7315, opened 2020-05-14
Authz deny

justinpettit:
This commit introduces new "_run_and_verify_*()" functions that run a function and verifies it against expected output.  An optional argument indicates the number of times a failed match should cause the function to be run again before giving up.  This is useful for tests that may take some time for the system to settle.  The second commit, which introduces an authz-deny test, makes use of this retry facility.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[X] Test and Release
[ ] User Experience
[X] Developer Infrastructure

### DIFF
--- a/tests/util/verify.sh
+++ b/tests/util/verify.sh
@@ -14,56 +14,60 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-_err_exit() {
+__err_exit() {
     echo "VERIFY FAILED $1: $2";
     exit 1
 }
 
-# Verify that $out is the same as $expected.
-_verify_same() {
+# Returns 0 if $out and $expected are the same.  Otherwise, returns 1.
+__cmp_same() {
     local out=$1
     local expected=$2
-    local msg=$3
 
     if [[ "$out" != "$expected" ]]; then
-        _err_exit "$msg" "$out"
+        return 1
     fi
+
+    return 0
 }
 
-# Verify that $out contains the substring $expected.
-_verify_contains() {
+# Returns 0 if $out contains the substring $expected.  Otherwise, returns 1.
+__cmp_contains() {
     local out=$1
     local expected=$2
-    local msg=$3
 
     if [[ "$out" != *"$expected"* ]]; then
-        _err_exit "$msg" "$out"
+        echo "false"
+        return 1
     fi
+
+    return 0
 }
 
-# Verify that $out does not contains the substring $expected.
-_verify_not_contains() {
+# Returns 0 if $out does not contain the substring $expected.  Otherwise,
+# returns 1.
+__cmp_not_contains() {
     local out=$1
     local expected=$2
-    local msg=$3
 
     if [[ "$out" == *"$expected"* ]]; then
-        _err_exit "$msg" "$out"
+        return 1
     fi
+
+    return 0
 }
 
-# Verify that $out contains the lines in $expected where "..." on a line
-# matches one or more lines containing any text.
-_verify_elided() {
+# Returns 0 if $out contains the lines in $expected where "..." on a line
+# matches one or more lines containing any text.  Otherwise, returns 1.
+__cmp_elided() {
     local out=$1
     local expected=$2
-    local msg=$3
-    
+
     local contains=""
     while IFS=$'\n' read -r line; do
         if [[ "$line" =~ ^[[:space:]]*\.\.\.[[:space:]]*$ ]]; then
             if [[ "$contains" != "" && "$out" != *"$contains"* ]]; then
-                _err_exit "$msg" "$out"
+                return 1
             fi
             contains=""
         else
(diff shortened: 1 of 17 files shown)


justinpettit:
I ran "make lint-fast" on my local machine, and it passed.  I did this because the doc describing writing the tests states that "make lint-fast" should be run to lint the shell scripts.  However, "lint-fast" seems to not lint the shell scripts anymore, and you need to run "make lint" to do that.

I'll clean up the "make lint" errors and resubmit.

tests/util/verify.sh
```diff
@@ -169,10 +177,124 @@ _verify_like() {
...
+
+    while true; do
+      out=$($cmd 2>&1)
+
+      if $cmp_func "$out" "$expected"; then
+          return
+      fi
+
+      if (( $attempt >= $max_attempts )); then
+          __err_exit "$cmd" "$out"
```
frankbu, line 208: Don't we need $msg here as well to indicate which snip in the test is failing. Sometimes different snips in the same doc use the same command more than once. It also seems that without the the snip function name in the message, it will be harder, in general, to track back to where in the test the failing call is coming from.

frankbu:
This looks great! It make me wonder if we even need the _verify_xxx functions. Maybe we should always use run_and_verify?

Regarding lint-fast, if it's not easy to fix it, should we change the README to use make lint? If it's not too slow, it's not a big deal.

tests/security/authz_deny/scripts/authz_deny.txt
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
frankbu, line 1: I'm in the process of trying to clean up all the current tests to put the entire script into a single .sh file (instead of the current one or more .txt files), and with one simple boilerplate .go file (which hopefully in the future will be generated).

It would be good if this new test follows that convention (i.e., rename authz_deny.txt to authz_deny.sh). Note that you will need to add a shellcheck comment on line 2 and put quotes around the snip.sh path, to make lint happy. Here's an example of the pattern I'm trying to make all tests follow: https://github.com/istio/istio.io/blob/master/tests/trafficmanagement/scripts/request_routing.sh

tests/util/verify.sh
justinpettit: Most of the $msg arguments that I saw were just the name of the function anyway.  Just to be clear, this is the name of the function from the snips.sh file.  For example, here's what it prints from that "No resources found" failure that was discussed in Slack:
```
VERIFY FAILED snip_set_up_the_cluster_5: No resources found
```
It's probably confusing since it uses the argument name "$cmd" instead of "$func".  I just wanted to separate it from the comparison function argument.  I'll look at finding a better name, since "command" is overloaded.

Does that make sense?  Let me know if I misunderstood what you suggested.

tests/security/authz_deny/scripts/authz_deny.txt
justinpettit: > This looks great! It make me wonder if we even need the _verify_xxx functions. Maybe we should always use run_and_verify?

That thought occurred to me, too, but I didn't want to be too presumptuous.  There were a couple places that seemed to use _verify functions that didn't invoke a command.  Now that I know that you like the approach, I'll look into that more option more seriously.
 
> Regarding lint-fast, if it's not easy to fix it, should we change the README to use make lint? If it's not too slow, it's not a big deal.

I'll see if I can track down what changed.

I'll take care of changing the name of the script to "authz_deny.sh".   I like that better, too.

tests/util/verify.sh
frankbu: Ah, that makes perfect sense. I didn't look closely enough and I thought it was going to display something like `VERIFY FAILED "curl bla bla bla": No resources found`.

tests/security/authz_deny/scripts/authz_deny.txt
frankbu: When you're looking at existing tests to see how they use _verify functions, keep in mind that only the tests that have been converted to .sh files are really what I think "correct" tests. The other ones are just a first pass to get rid of the old snippet gen stuff, so I could remove it from the underlying framework. I'm going to keep cleaning up tests, as much as I find time, so hopefully in a few days they will be all cleaned up single .sh tests. At that point, I doubt that there will be any verify calls that don't also need to call a function, but you never know

frankbu:
This is a big improvement. Thanks @justinpettit. Let's merge it now, and only use the `_run_and_verify` functions from now on.

Once all the existing tests are cleaned up, including also only using `_run_and_verify`, then we can remove the old `_verify` functions, but, I'm thinking we can then shorten the name of these functions to just be the old `_verify` names (i.e., remove the `_run_and_` part. Then we just need to change the doc for the `_verify_xxx` functions to say `Verify that the output from running $func is ...` instead of `Verify $out is ...`. WDYT @justinpettit?

PR #7323, opened 2020-05-15
A bit of cleanup on the arch doc

costinm:
Removing the use of 'citade' and some sections I have trouble understdanding.

### DIFF
--- a/content/en/docs/ops/deployment/architecture/index.md
+++ b/content/en/docs/ops/deployment/architecture/index.md
@@ -25,12 +25,6 @@ The following diagram shows the different components that make up each plane:
     caption="Istio Architecture"
     >}}
 
-Traffic in Istio is categorized as data plane traffic and control plane traffic.
-Data plane traffic refers to the messages that the business logic of the workloads
-send and receive. Control plane traffic refers to configuration and control messages sent
-between Istio components to program the behavior of the mesh. Traffic management
-in Istio refers exclusively to data plane traffic.
-
 ## Components
 
 The following sections provide a brief overview of each of Istio's core components.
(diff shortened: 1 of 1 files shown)


costinm:
Note - there is a SVG showing Pilot, Citadel and Galley all happy together - not sure how it was generated, but it would be great to reduce the confusion since Galley is not really running in istiod 
and we're trying to stop using 'pilot' and 'citadel' terms.

costinm:
I would also remove the 'goals' section - I can't really understand them completely and I doubt 
a user gains much by reading it. 

The first 'goal' seem to suggest users should not encrypt 
or implement its own security - so Istio can inspect traffic and apply its own policy. If that is the
intent - it should be explicitly stated maybe, I'm trying to read between the lines... 

Last 2 goals are unfortunately far from reality (or current priorities) - and the 'extensibility' probably needs to mention WASM and be simpler.

content/en/docs/ops/deployment/architecture/index.md
```diff
@@ -82,94 +76,41 @@ Some of the Istio features and tasks enabled by Envoy proxies include:
...
-  with it as needed.
+Istiod maintains a CA and generates certificates to allow secure mTLS communication
+in the data plane.
+
+## Application design
+
+A design goal for Istio is that the developer can do minimum work
+to get real value from the system. Istio can automatically inject itself into
+the network paths between services and program the networking layer to
+securely route traffic and extract telemetry. If the application code
```
frankbu, line 113: I'm not sure I understand what point the last sentence is trying to make `If the application ...`.
Seems like it should add something like `which will make Istio less functional so don't do those things, let Istio do it for you`, or something like that.

PR #7331, opened 2020-05-15
Misleading default namespace value 'istio'

fhaze:
When generating the istio manifest, we should stick to the default namespace istio's installation uses which is **istio-system**. The way the text is written can mislead the user to a error because the namespace **istio** was never created or stated as custom in this context.


fhaze:
@googlebot I signed it!

PR #7338, opened 2020-05-15
Authz jwt

justinpettit:


tests/security/authz_jwt/authz_jwt_test.go
```diff
@@ -0,0 +1,54 @@
...
+		Run(istioio.NewBuilder("tasks__security___authz_jwt").
+			Add(istioio.Script{
+				Input: istioio.Inline{
+					FileName: "create_ns_foo_with_httpbin_sleep.sh",
+					Value: `
+source ${REPO_ROOT}/content/en/docs/tasks/security/authorization/authz-jwt/snips.sh
+snip_before_you_begin_1`,
+				},
+			}).
+			Add(istioio.MultiPodWait("foo")).
```
frankbu, line 39: Just FYI, part of the cleanup exercise I'm doing is to try to make all tests put the entire test in one .sh file (excluding cleanup which needs to be deferred, although I have an idea for handling that too, to be tried in a future cleanup). Rather than relying on code in this go file to wait for pods to deploy, I'd like the script to just call a bash library function to do it, or similar wait. Once all the tests are completely described in one bash file, this go file becomes total boilerplate, so we should be able to either generate it, or even do it on-the-fly.

justinpettit: That sounds like a great plan to me.  I wanted to finish knocking out the rest of the security task tests, so I'll plan to write them in the current style.  I'm happy to convert them when you're ready, though.

Issue #7344, opened 2020-05-17
Issue with docs/ops/diagnostic-tools/proxy-cmd/index.md


hobbytp:
it still uses 15001 for both inbound and outbound direction traffic, in fact, this behavious is changed, 15006 for inbound and 15001 for outbound traffic now, please help to update it.

PR #7353, opened 2020-05-19
Update inbound and outbound ports

shamsher31:
To fix https://github.com/istio/istio.io/issues/7344

Ref: https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/listener_builder.go#L56


frankbu:
@shamsher31 Please fix the merge conflict.

PR #7364, opened 2020-05-20
Add authz-td-migration user guide test.

justinpettit:


tests/security/authz_td_migration/scripts/authz_td_migration.sh
```diff
@@ -0,0 +1,47 @@
...
+
+set -e
+set -u
+set -o pipefail
+
+source "${REPO_ROOT}/content/en/docs/tasks/security/authorization/authz-td-migration/snips.sh"
+
+max_attempts=10
+
+snip_before_you_begin_3
```
ericvn, line 27: Do we need to do snip_before_you_begin_2?

justinpettit: snip_before_you_begin_2 sets up the namespace which needs to be a blocking call, so it's called as part of the Go program itself.  This task is really slow to execute, so I think it's timing out.  I have a patch that will improve debugging, which I plan to send out shortly, and then I'll confirm that.

tests/security/authz_td_migration/main_test.go
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authztdmigration
+
+import (
+	"testing"
+
+	"istio.io/istio/pkg/test/framework"
+	"istio.io/istio/pkg/test/framework/components/istio"
```
frankbu, line 21:
```suggestion
```

tests/security/authz_td_migration/scripts/authz_td_migration.sh
```diff
@@ -0,0 +1,62 @@
...
+
+
+set -e
+set -u
+set -o pipefail
+
+source "${REPO_ROOT}/content/en/docs/tasks/security/authorization/authz-td-migration/snips.sh"
+source "${REPO_ROOT}/tests/util/samples.sh"
+
+max_attempts=10
```
frankbu, line 26: I'm removing the `max_attempts` argument from the `_run_and_verify` functions in #7440. You can override the default 5 attempts by setting the `VERIFY_RETRIES` environment variable.

justinpettit: Sounds good.  This test is particularly slow, so it needs a higher retry count.  Hopefully we can start rolling out tests with the new istioctl "wait" mechanism to improve this situation.

frankbu:
@justinpettit This needs to be updated to use the latest new-and-improved framework style (https://github.com/istio/istio.io/tree/master/tests#test-authoring-overview). If you aren't able to do it soon, please close this PR and I can take what you've done, convert to the new style, and finish it for you. Thanks.

frankbu:
Test added in #7653

Note that the last check (trustDomainAliases) in this test isn't working, so it's currently commented out.

PR #7367, opened 2020-05-20
Add TLS origination doc test

howardjohn:
For https://github.com/istio/istio/issues/23899

* Add a new test for the doc
* Make some changes to the doc to allow the tests to work. Should be minimal change experienced by readers
* Add some logic to the test framework to stream test output. This is at `debug` log level so no impact unless you are looking for it. We can continue to refine this in the future and possibly make it not debug level.


frankbu:
@howardjohn

> Add some logic to the test framework to stream test output. This is at debug log level so no impact unless you are looking for it.

Can you explain how to use this and how to turn it on? Is it just to be used when running the tests locally? Could this use a few words in the README?

howardjohn:
> Can you explain how to use this and how to turn it on? Is it just to be used when running the tests locally? Could this use a few words in the README?

`--log_output_level=script:debug`

We can make it default if desired - I just wasn't sure so bumped it down

frankbu:
> We can make it default if desired - I just wasn't sure so bumped it down

Seems good like you have it for now.

PR #7383, opened 2020-05-22
Added virtualservice delegation

hzxuzhonghu:
Add missing vs delegation.


hzxuzhonghu:
cc @ericvn @jacob-delgado

PR #7391, opened 2020-05-23
Added Envoy, Go and Istio matchings to Prometheus federation ServiceMonitors

fktkrt:
Wouldn't it make more sense to include cases for those metrics too? Or at least mention these somewhere.

### DIFF
--- a/content/en/docs/ops/best-practices/observability/index.md
+++ b/content/en/docs/ops/best-practices/observability/index.md
@@ -146,10 +146,6 @@ Add the following job to your configuration:
     'match[]':
     - '{__name__=~"workload:(.*)"}'
     - '{__name__=~"pilot(.*)"}'
-    - '{__name__=~"envoy(.*)"}'
-    - '{__name__=~"go(.*)"}'
-    - '{__name__=~"istio(.*)"}'
-
 {{< /text >}}
 
 If you are using the [Prometheus Operator](https://github.com/coreos/prometheus-operator), use the following configuration instead:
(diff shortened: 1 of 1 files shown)


fktkrt:
Should I include more details about those metrics, or do you think this is enough?
I think it's useful to mention that you can include these as well, to have a better federation experience.

content/en/docs/ops/best-practices/observability/index.md
```diff
@@ -146,6 +146,10 @@ Add the following job to your configuration:
     'match[]':
     - '{__name__=~"workload:(.*)"}'
     - '{__name__=~"pilot(.*)"}'
+    - '{__name__=~"envoy(.*)"}'
+    - '{__name__=~"go(.*)"}'
+    - '{__name__=~"istio(.*)"}'
```
douglas-reid, line 151: We don't want istio metrics directly. That is the point of the the `workload` prefix in this section of the document.

This is the same thing below.

fktkrt: Alright, removing istio below makes perfect sense.
I'll update the PR with the removal and mention these between {{< tip >}}  {{< /tip >}} tags. Thank you for picking this up!

content/en/docs/ops/best-practices/observability/index.md
```diff
@@ -187,6 +187,8 @@ The key to the federation configuration is matching on the job in the Istio-depl
 the prefix used in the workload-level recording rules (`workload:`). This will allow existing dashboards and
 queries to seamlessly continue working when pointed at the production Prometheus instance (and away from the Istio instance).
 
+You can also include additional metrics, e.g. envoy, go, etc. when setting up federation.
```
frankbu, line 190:
```suggestion
You can also include additional metrics (for example, envoy, go, etc.) when setting up federation.
```

PR #7413, opened 2020-05-27
Add test: field to all md files

frankbu:
All md files have `test:` attribute in their metadata with values of `yes`, `no`, or `n/a` (if doc has no commands to test). The following command can be used to get a summary of the testing status:
```
echo "Documents: \
   $(find content/en -name '*.md' -exec grep --quiet '^test: yes$' {} \; -print | wc -l) (tested) \
   $(find content/en -name '*.md' -exec grep --quiet '^test: no$' {} \; -print | wc -l) (untested) \
   $(find content/en -name '*.md' -exec grep --quiet '^test: n/a$' {} \; -print | wc -l) (n/a) \
   $(find content/en -name '*.md' -print | wc -l) (total)"
```


ericvn:
Nice. So developers can look for `test: no` pages to see what needs test cases written.

ericvn:
Would it be nice to add a Makefile target with your mentioned summary commands? Not that it would be hard to figure out/write, or try and look back for this PR. It would just make that easier.

frankbu:
> Would it be nice to add a Makefile target with your mentioned summary commands? 

Easy enough to add something basic for now, but we probably want something better (e.g., spreadsheet) really.

Let's merge this one first, since it touches so many files, and we can look at adding a script in a followup PR.

PR #7414, opened 2020-05-27
CORS Policy Example: Change allowOrigin block

dntosas:
### Description
In this PR, we change the block of CORS Policy definition from `allowOrigin` to `allowOrigins` so not to be misleading for people who take this example as guide to setup their meshes.

After the upgrade to Istio 1.6 the block with deprecated `allowOrigin` is not respected thus not working.

### Addresses
https://github.com/istio/istio/issues/24145

### Checklist


dntosas:
@googlebot I signed it!

PR #7415, opened 2020-05-27
Add tracing integration pages

howardjohn:
Part of https://github.com/istio/istio/issues/23583

These are meant to be fairly light and delegate to the existing tasks

### DIFF
--- a/content/en/docs/ops/integrations/zipkin/index.md
+++ b/content/en/docs/ops/integrations/zipkin/index.md
@@ -1,7 +1,7 @@
 ---
 title: Zipkin
 description: How to integrate with Zipkin.
-weight: 20
+weight: 30
 keywords: [integration,zipkin,tracing]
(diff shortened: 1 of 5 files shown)


content/en/docs/ops/integrations/zipkin/index.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Zipkin
+description: How to integrate with Zipkin.
+weight: 20
```
ericvn, line 4: I suspect you'll want different weights on the docs to maybe keep them in alphabetical order. I'm not sure if that's guaranteed in the case of same weights.

frankbu: If all the entries have the same weight, I think they will just be processed in the same order they appear in the folder, which is alphabetical order.

PR #7416, opened 2020-05-27
network-issues: Remove "Headless TCP services losing connection".

justinpettit:
This section is no longer correct, since we don't restart Envoy every
45 days, and we no longer recommend creating a separate "istio-citadel"
deployment.

### DIFF
--- a/content/en/docs/ops/common-problems/network-issues/index.md
+++ b/content/en/docs/ops/common-problems/network-issues/index.md
@@ -196,30 +196,6 @@ spec:
         subset: v1
 {{< /text >}}
 
-## Headless TCP services losing connection
-
-If `istio-citadel` is deployed, Envoy is restarted every 45 days to refresh certificates.
-This causes the disconnection of TCP streams or long-running connections between services.
-
-You should build resilience into your application for this type of
-disconnect, but if you still want to prevent the disconnects from
-happening, you will need to disable mutual TLS and the `istio-citadel` deployment.
-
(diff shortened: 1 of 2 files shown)


ericvn:
Comment:

You don't need to make changes in any of the non-english pages. A translation team will move things across. The Chinese pages are approved by a different team, so changes may be delayed as you will need multiple approvals. That said, I don't believe there is a current Portuguese team (although I think there was mention of one coming onboard in a WG meeting). @frankbu, status and does it make sense to keep the pt-br directory around? 

Question:

Does this need to be cherry-picked into release 1.6? I suspect it should, but I'm not sure. If it does,  the release-1.6 cherry pick label should set.

justinpettit:
> Comment:
> 
> You don't need to make changes in any of the non-english pages. A translation team will move things across. The Chinese pages are approved by a different team, so changes may be delayed as you will need multiple approvals.

Good to know.  Thanks.

> Question:
> 
> Does this need to be cherry-picked into release 1.6? I suspect it should, but I'm not sure. If it does, the release-1.6 cherry pick label should set.

I suspect so, too, but would like @myidpt to confirm.

justinpettit:
@ericvn Should I resubmit without the Portuguese version?  It's easy enough.

ericvn:
@justinpettit I'd just leave this one as it is, but just something to remember going forward. Once we hear back on if we should cherry-pick to 1.6, I can approve.

PR #7418, opened 2020-05-28
Fix sudo command on VM doc

howardjohn:
If you are not running as root already, this will fail. The redirect part is not run as root previously

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


sdake:
@howardjohn this should go to master instead - and then be cherry-picked unless our processes have changed.

Thanks for running through the instructions. General thoughts on improvements?

Cheers
-steve

howardjohn:
If we want PRs to go to `master` we should make the "Edit this page on
github" button go to master branch I think

PR #7433, opened 2020-05-28
Authn policy

justinpettit:
[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X] Security
[X] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/security/authentication/authn-policy/index.md
+++ b/content/en/docs/tasks/security/authentication/authn-policy/index.md
@@ -74,6 +74,7 @@ Last but not least, verify that there are no destination rules that apply on the
 
 {{< text bash >}}
 $ kubectl get destinationrules.networking.istio.io --all-namespaces -o yaml | grep "host:"
+
 {{< /text >}}
 
 {{< tip >}}
@@ -100,6 +101,7 @@ When the server doesn't have sidecar, the `X-Forwarded-Client-Cert` header is no
 
 {{< text bash >}}
 $ kubectl exec $(kubectl get pod -l app=sleep -n foo -o jsonpath={.items..metadata.name}) -c sleep -n foo -- curl http://httpbin.legacy:8000/headers -s | grep X-Forwarded-Client-Cert
+
 {{< /text >}}
 
 ## Globally enabling Istio mutual TLS in STRICT mode
@@ -228,6 +230,21 @@ EOF
 
 Again, run the probing command. As expected, request from `sleep.legacy` to `httpbin.bar` starts failing with the same reasons.
 
+{{< text bash >}}
+$ for from in "foo" "bar" "legacy"; do for to in "foo" "bar" "legacy"; do kubectl exec "$(kubectl get pod -l app=sleep -n ${from} -o jsonpath={.items..metadata.name})" -c sleep -n ${from} -- curl "http://httpbin.${to}:8000/ip" -s -o /dev/null -w "sleep.${from} to httpbin.${to}: %{http_code}\n"; done; done
+sleep.foo to httpbin.foo: 200
+sleep.foo to httpbin.bar: 200
+sleep.foo to httpbin.legacy: 200
+sleep.bar to httpbin.foo: 200
+sleep.bar to httpbin.bar: 200
+sleep.bar to httpbin.legacy: 200
+sleep.legacy to httpbin.foo: 000
+command terminated with exit code 56
+sleep.legacy to httpbin.bar: 000
+command terminated with exit code 56
+sleep.legacy to httpbin.legacy: 200
+{{< /text >}}
+
 {{< text plain >}}
 ...
 sleep.legacy to httpbin.bar: 000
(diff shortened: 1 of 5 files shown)


content/en/docs/tasks/security/authentication/authn-policy/index.md
```diff
@@ -388,16 +414,14 @@ spec:
...
-{{< /text >}}
+Follow the instructions in
+    [Determining the ingress IP and ports](/docs/tasks/traffic-management/ingress/ingress-control/#determining-the-ingress-ip-and-ports)
+to define the `INGRESS_HOST` and `INGRESS_PORT` environment variables.
 
 And run a test query
 
 {{< text bash >}}
-$ curl $INGRESS_HOST/headers -s -o /dev/null -w "%{http_code}\n"
+$ curl "$INGRESS_HOST":"$INGRESS_PORT"/headers -s -o /dev/null -w "%{http_code}\n"
```
frankbu, line 424: I think it's cleaner to just wrap the entire URL argument in quotes:
```suggestion
$ curl "$INGRESS_HOST:$INGRESS_PORT/headers" -s -o /dev/null -w "%{http_code}\n"
```

tests/security/authn_policy/scripts/authn_policy.sh
```diff
@@ -0,0 +1,97 @@
...
+snip_enduser_authentication_4
+
+_run_and_verify_same  snip_enduser_authentication_5 "$snip_enduser_authentication_5_out"
+_run_and_verify_same  snip_enduser_authentication_6 "$snip_enduser_authentication_6_out"
+_run_and_verify_same  snip_enduser_authentication_7 "$snip_enduser_authentication_7_out"
+
+snip_enduser_authentication_8
+snip_enduser_authentication_9
+
+_run_and_verify_same  snip_enduser_authentication_10 "$snip_enduser_authentication_10_out"
```
frankbu, line 88: If we make `__verify_with_retry` public, tests could pass in a custom cmp function. Then we could use the generated _out snippet here, otherwise I suppose using verify_contains is the best you can do.
```suggestion
expected="200
400"
_run_and_verify_contains snip_enduser_authentication_10 $expected
```

justinpettit: Thanks for the great suggestion for the fix.  You saved me some time noodling on how to address the issue--all my immediate solutions were needlessly complex.

content/en/docs/tasks/security/authentication/authn-policy/index.md
justinpettit: Agreed.  I pushed a version with that change.

content/en/docs/tasks/security/authentication/authn-policy/index.md
```diff
@@ -506,13 +529,13 @@ EOF
 Retry the request without a token. The request now fails with error code `403`:
 
 {{< text bash >}}
-$ curl $INGRESS_HOST/headers -s -o /dev/null -w "%{http_code}\n"
+$ curl "$INGRESS_HOST:$INGRESS_PORT"/headers -s -o /dev/null -w "%{http_code}\n"
```
frankbu, line 532: This better, but a common convention is to wrap the entire single bash arg in quotes. Also, then you don't need to show quotes when describe it in the text (see my other suggestion, below).
```suggestion
$ curl "$INGRESS_HOST:$INGRESS_PORT/headers" -s -o /dev/null -w "%{http_code}\n"
```
Also, don't forget to regen the snips.

content/en/docs/tasks/security/authentication/authn-policy/index.md
```diff
@@ -506,13 +529,13 @@ EOF
...
 403
 {{< /text >}}
 
 ### Require valid tokens per-path
 
-To refine authorization with a token requirement per host, path, or method, change the authorization policy to only require JWT on `/headers`. When this authorization rule takes effect, requests to `$INGRESS_HOST/headers` fail with the error code `403`. Requests to all other paths succeed, for example `$INGRESS_HOST/ip`.
+To refine authorization with a token requirement per host, path, or method, change the authorization policy to only require JWT on `/headers`. When this authorization rule takes effect, requests to `"$INGRESS_HOST:$INGRESS_PORT"/headers` fail with the error code `403`. Requests to all other paths succeed, for example `"$INGRESS_HOST:$INGRESS_PORT"/ip`.
```
frankbu, line 538:
```suggestion
To refine authorization with a token requirement per host, path, or method, change the authorization policy to only require JWT on `/headers`. When this authorization rule takes effect, requests to `$INGRESS_HOST:$INGRESS_PORT/headers` fail with the error code `403`. Requests to all other paths succeed, for example `$INGRESS_HOST:$INGRESS_PORT/ip`.
```

content/en/docs/tasks/security/authentication/authn-policy/index.md
justinpettit: That's better.  Should be fixed in the new version.

content/en/docs/tasks/security/authentication/authn-policy/index.md
justinpettit: Good catch.  I moved through the search and replace too fast.  Should be fixed in the new version.

PR #7446, opened 2020-06-01
init.sh: Add "make gen-charts".

justinpettit:
Fixes https://github.com/istio/istio/issues/24325

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[X] Test and Release
[ ] User Experience
[X] Developer Infrastructure

### DIFF
--- a/bin/init.sh
+++ b/bin/init.sh
@@ -48,6 +48,7 @@ git fetch "$ISTIO_REMOTE"
 git checkout "$ISTIO_SHA"
 
 # Build and install istioctl
+make gen-charts
 go install ./istioctl/cmd/istioctl
 
 popd > /dev/null


bin/init.sh
```diff
@@ -48,6 +48,7 @@ git fetch "$ISTIO_REMOTE"
 git checkout "$ISTIO_SHA"
 
 # Build and install istioctl
+make gen-charts
```
frankbu, line 51: Is this `gen-charts` rule available in istio/istio.io? I don't see it in the makefile.

justinpettit: I made this change based on a conversation with @ericvn in #test-and-release:

https://istio.slack.com/archives/C6FCV6WN4/p1590690191035700?thread_ts=1590687831.027200&cid=C6FCV6WN4

It looks like it's defined in `Makefile.core.mk` in istio/istio.  Should I add a similar definition to the Makefile in istio/istio.io.

frankbu: I'm not really very familiar with the operator stuff, but I don't think we really want to add Makefile rules in istio.io for the Istio test env. Maybe instead of calling `make gen-charts` in init.sh, you should call the underlying script directly (`operator/scripts/create_assets_gen.sh`). Maybe @ostromart or @nmittler have some advice, on what would be the right way to do this.

If I understand correctly, this is an important fix that will be needed before we can write tests for any of the setup/install instructions.

howardjohn: why run the script instead of the make target? they make target is much more stable and less likely to break

frankbu: @howardjohn Maybe I'm just confused. I thought init.sh is running in the context of the istio/istio.io Makefile, so there is no target for `make gen-charts`. The make target is in the istio/istio Makefile. Am I missing something?

frankbu: Oh, I see there is this line just before the istioctl install: `pushd "${ISTIO_GO}"`, so I guess `make gen-charts` will work.

PR #7396, opened 2020-05-26
update doc to use istioctl install

shamsher31:
Based on https://github.com/istio/istio.io/pull/7336 and https://github.com/istio/istio/pull/22563, keeping consistency across docs for `istioctl install` instead of `istioctl manifest apply`. 

The `istioctl install` is already available for [Customizable Install with Istioctl](https://istio.io/docs/setup/install/istioctl/). Not sure when is the plan to deprecate `istioctl manifest apply`.

PR #7449, opened 2020-06-02
update doc to use istioctl install (#7396)

shamsher31:
Manual cherry-pick of https://github.com/istio/istio.io/pull/7396

will close https://github.com/istio/istio.io/issues/7448

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure


content/en/docs/tasks/security/authentication/authn-policy/snips.sh
```diff
@@ -0,0 +1,506 @@
+#!/bin/bash
```
ericvn, line 1: I believe this file shouldn't be included. It might be part of your `master` change as master has a test for this page (https://github.com/istio/istio.io/pull/7396/files#diff-72f5910f0635b96687140aa324274234L9) while `release 1.6` does not.

shamsher31: Ahh, thanks for catching.

PR #7452, opened 2020-06-02
Rename _run_and_verify_xxx functions to _verify_xxx

frankbu:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


tests/configuration/scripts/liveness_and_readiness_probes.sh
```diff
@@ -53,10 +53,14 @@ echo "$snip_use_annotations_on_pod_1" | kubectl -n health-annotate apply -f -
 
 kubectl -n health-annotate rollout status deployment liveness-http --timeout 30s
 
-out=$(kubectl -n health-annotate get pod 2>&1)
+# helepr function
```
justinpettit, line 56: s/helepr/helper/

tests/README.md
```diff
@@ -113,51 +113,49 @@ snip_config_50_v3 # Step 3: switch 50% traffic to v3
...
+   contain the substring `expected`, exponentially back off and try again, 5 times
+   by default. The number of retries can be changed by setting the `VERIFY_RETRIES`
+   environment variable.
 
-   Verify that `out` does not contains the substring `expected`. Failure messages will include
-   the specified `msg`.
+1. **`_verify_not_contains`** `func` `expected`
 
-1. **`_verify_first_line`** `out` `expected` `msg`
+   Runs `func` and compares the output with `expected`. If the command execution fails
```
justinpettit, line 140: This is the only command that mentions that if command execution fails, it will try again.  This should be true for all of them, so I think it would be good to mention that either in each command or elsewhere once as general information.

frankbu: The reason I don't think it's necessary in the other functions is because if they fail to execute the output will NOT be as expected (because they are looking for something specific in the output), so they will not return success. verify_not_contains was the only tricky one that needed to check for failure, otherwise it was returning on failed commands with a successful compare (string is missing from output, as expected). Maybe mentioning command failure is just confusing things, and I shouldn't mention it at all?

tests/configuration/scripts/liveness_and_readiness_probes.sh
frankbu:
```suggestion
# helper function
```

tests/README.md
justinpettit: Now I understand your reasoning.  However, by calling out the behavior on command failure for just this case, it reads to me like it's implying different behavior in those other cases.  My two cents, anyway.  :-)

frankbu: That makes sense. I'll update the doc to make it clear that command failure -> retry in all verify functions.

frankbu: Turns out that we do need a different behavior for the other cases. There are examples where the actual command does fail but the expected output is what we're looking for from the snip. So, _verify_not_contains is the only one that we don't want to fail verify if the command fails, because that makes no sense.

PR #7453, opened 2020-06-03
Update the Makefile path for root cert generation.

myidpt:
Corresponds to https://github.com/istio/istio/pull/24343.

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -66,7 +66,7 @@ deployment you place into production.
 
     {{< text bash >}}
     $ cd "${WORK_DIR}"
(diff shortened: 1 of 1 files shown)


myidpt:
@frankbu That's a good point. Adding a comment in the istio/tools/certs/README.md may not be effective because 1.6.0 users won't be able to see it :)

I think we may want to add a note in the 1.6 doc to say the location is `istio/samples/certs` for 1.6.0.

frankbu:
@myidpt Oh right, the 1.6.0 image doesn't even have the tools README :) So I guess the best idea is to merge this PR and the cherrypick, and then send a followup PR to release-1.6 only, with the note.

PR #7455, opened 2020-06-03
Simplify passing --context to access different cluster

shamsher31:
As requested by @ostromart in https://github.com/istio/istio.io/pull/7396#discussion_r432602440.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure


ericvn:
@shamsher31 I assume we want this in 1.6 as well?

frankbu:
This seems like just a small syntactic change, and isn't this one of the docs that is going to be replaced soon?

cc @nmittler

nmittler:
@frankbu Yeah it's just a semantic change, but it's probably fine. It isn't yet clear if these docs will be entirely replaced for 1.6.1. We may only have the docs for single network in place by then, in which case we'll still refer to these docs for multi-network.

@rcaballeromx

PR #7491, opened 2020-06-05
[release-1.6] Move the release-1.6 branch to use release-1.6 common files.

ericvn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
Note that this puts local builds onto the same build images as prow is using:

https://github.com/istio/test-infra/blob/master/prow/config/jobs/istio.io-1.6.yaml#L2-L3

Today, they are different.

PR #7520, opened 2020-06-10
blog for setting up ingress with Istio

devadvocado:
Cross-posting a popular ingress+Istio blog I posted on the Kubernetes blog that I think would be of interest to the Istio community. 

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


frankbu:
@devadvocado This doesn't seem appropriate for the Istio blog. Other than using the Istio bookinfo services as the backend application, it doesn't demonstrate any interesting use of Istio. Also, it uses Kong as a k8s Ingress controller while Istio doesn't even recommend using k8s Ingress at all. Istio uses an Istio Gateway for edge routing instead.  A Kong/ingress article on the Istio blog would be expected to explain how it compares to the Istio Gateway it replaces (pros/cons). For example, with an Istio Gateway, a VirtualService, handles the edge routing, etc., so what functionality of that approach is available, lost, or replaced when using Kong for Ingress?

linsun:
@devadvocado I second Frank's concern.  Is there anything that is istio related other than bookinfo examples?

linsun:
cc @istio/steering-committee

devadvocado:
@frankbu @linsun thanks for the reviewing and feedback! Quite honestly, I was just reposting an Istio related blog post we published on the Kubernetes blog that got us some great feedback. But I will review your feedback, rewrite, and republish something that fits better for the Istio community. Thanks!

rshriram:
@devadvocado thanks for your contribution. I think the concern that @linsun / @frankbu have is about the lack of additional explanation in this article regarding the motivations for doing this kind of integration. It is true that in Istio, we allow legacy ingress controller implementations to co-exist with the Istio gateway, and that the technique you have followed is essentially what some of our users follow when deploying the nginx ingress controller with Istio.

My suggestions are as follows:

Explain the motivation upfront - e.g., “for those who are already using Kong for API management and want to continue using it while taking advantage of Istio for service mesh capabilities like advanced routing, built in mtls, authz, reliability knobs, uniform observability, etc., here is a demo explaining how you can do so”. or something along these lines..

Now, with regard to the technique itself, what you have written works but is not the best way to take full advantage of both solutions. Instead, I suggest you try this:
1. Deploy Istio gateway container and the Kong container in the same pod (you can either inject the kong container specs into the gateway container or vice versa, and have a link to this yaml)
2. Configure kong (as usual) but set all the destinations to 127.0.0.1:15080 or some port that is not exposed by the k8s service.
3. Configure Istio ingress gateway plaintext http servers with the same port 15080 (following the standard Istio gateway tasks).

Essentially, you are forwarding all traffic from Kong to the Istio gateway on the localhost port 15080. From there on, its business as usual, with Istio gateway doing its thing, forwarding traffic. The performance profile should be the same as what you currently have, except you wont need iptables traffic capture.

The benefit of this model is that users get the best of both systems. You can configure Kong natively using whatever APIs you have, instead of using the limited k8s ingress api filled with annotations. At the same time, you can configure/control the Istio gateways with the native APIs.

If you dont feel like trying out this technique, you can fallback to the existing model but make it a point to clearly explain how the features align.. Something like, "Kong ingress controller handles inbound traffic, performs api management tasks, and then hands it over to istio sidecars that in turn take care of automatic mutual tls, authorization, reliability (through virtual services, destination rules, etc.), metrics and tracing, etc.". We have refrained from publishing this way of integration as a dedicated task purely because its pretty clunky, hacky and has a string of issues we have seen in the past. But we still support it to interoperate with legacy ingress controller solutions.

PR #7525, opened 2020-06-10
refactor(telemetry): move wasm telemetry docs to proxy extensions

douglas-reid:
This PR moves the "Istio Standard Metrics" documentation up a level in the Reference section (replacing the "Telemetry V2" section) and moves the document on Wasm-based Telemetry into the "Proxy Extensions" section.

There was user-confusion reported for the existing documentation. Users were having a tough time sorting out Wasm vs. Telemetry v2, etc. This PR aims to eliminate some of the source of that confusion by removing focus from "v2" and clarifying where Wasm fits in, and how it may be used.

cc: @therealmitchconnors @mandarjog 

[ X ] Policies and Telemetry


ericvn:
Based on user confusion, I expect this needs to be put in the 1.6 docs.

PR #7501, opened 2020-06-08
Egress Gateway TLS Origination Test Suite 

nschhina:
Gateway TLS Origination Test Suite for
https://istio.io/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/

Autogenerated snips.sh for this file is not testable due to inclusion of password placeholder not being a valid bash command. This PR replaces to password for ease of use with snips.sh

Required for [istio/istio/issues/14039]
[ ] Configuration Infrastructure
[x] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #7545, opened 2020-06-12
fix bug where *_test.sh doesn't source snips.sh

niwatorrri:
Previously `a_test.sh` and `b_test.sh` would auto-source `a_snips.sh` and `b_snips.sh`, which is a bug and should only source `snips.sh`. Related comment [here](https://github.com/istio/istio.io/pull/7522#issuecomment-643354850).

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


frankbu:
@niwatorrri How about removing the temporary workarounds from @nschhina's scripts (see https://github.com/istio/istio.io/pull/7501). Even though it's a very simple fix, it won't hurt to have test scripts in the PR to confirm it's working.

niwatorrri:
@frankbu Yup it's working!

PR #7566, opened 2020-06-15
Add archive-version build target

davidhauck:
This build target builds a static version of the site to be used as an archive, and can be copied into the archive folder of the current release. This also includes the archive/ folder with static files so they can be included in the build for this version if needed.

[ X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


frankbu:
@davidhauk, is `release-1.5` the right branch for this? Shouldn't it be `release-1.6` (current release)?

davidhauck:
@frankbu yes this is the correct branch. Pretty much all branches from 1.0 to 1.5 need to be updated to build on /latest so they can be archived correctly. I've also added a new script to those releases called "make archive-version." This script will build the site and allow it to be directly copied into the archive/ folder. If anything needs to be changed in an old release, the branch can be changed, and then make archive-version can be run.

I'll be doing a PR for the other branches soon.

frankbu:
I doubt that we'll want to make any changes to anything older than 1.5 (maybe 1.4, but even that seems unlikely). In that case, do you still need to update them to build on /latest?

PR #7591, opened 2020-06-18
Add new entries to clarify multicluster terminology.

rcaballeromx:
These entries are part of a wider effort to improve
multicluster content.  Implementing the terms
across the current content will happen in a separate PR.

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/reference/glossary/primary-cluster.md
+++ b/content/en/docs/reference/glossary/primary-cluster.md
@@ -1,8 +1,9 @@
 ---
 title: Primary Cluster
-test: n/a
 ---
 
-A primary cluster is a cluster with a control plane, which observes all
-clusters in the mesh for endpoints and facilitates cross-cluster load balancing
-between clusters.
+A primary cluster is a [cluster](/docs/reference/glossary/#cluster) with a
+[control plane](/docs/reference/glossary/#control-plane). A single
+[mesh](/docs/reference/glossary/#service-mesh) can have more than
+one primary cluster for HA or to reduce latency. Primary clusters can act as the
+control plane for [remote clusters](/docs/reference/glossary/#remote-cluster).
--- a/content/en/docs/reference/glossary/primary-cluster.md
+++ b/content/en/docs/reference/glossary/primary-cluster.md
@@ -1,0 +1,8 @@
+---
+title: Primary Cluster
+test: n/a
+---
+
+A primary cluster is a cluster with a control plane, which observes all
+clusters in the mesh for endpoints and facilitates cross-cluster load balancing
+between clusters.
--- a/content/en/docs/reference/glossary/remote-cluster.md
+++ b/content/en/docs/reference/glossary/remote-cluster.md
@@ -1,9 +1,9 @@
 ---
 title: Remote Cluster
-test: n/a
 ---
 
-A remote cluster is a cluster that uses a remote profile to connect to a
-control plane residing outside of the cluster. The control plane can reside
-either in another cluster or external to the mesh as is the case with managed
-control planes.
+A remote cluster is a [cluster](/docs/reference/glossary/#cluster) that
+connects to a [control plane](/docs/reference/glossary/#control-plane)
+residing outside of the cluster. A remote cluster can connect to a control plane
+running in a [primary cluster](/docs/reference/glossary/#primary-cluster)
+or to a [managed control plane](/docs/reference/glossary/#managed-control-plane).
--- a/content/en/docs/reference/glossary/remote-cluster.md
+++ b/content/en/docs/reference/glossary/remote-cluster.md
@@ -1,0 +1,9 @@
+---
+title: Remote Cluster
+test: n/a
+---
+
+A remote cluster is a cluster that uses a remote profile to connect to a
+control plane residing outside of the cluster. The control plane can reside
+either in another cluster or external to the mesh as is the case with managed
+control planes.


content/en/docs/reference/glossary/primary-cluster.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Primary Cluster
+test: n/a
+---
+
+A primary cluster is a cluster with a control plane, which observes all
+clusters in the mesh for endpoints and facilitates cross-cluster load balancing
```
frankbu, line 7: Is this correct?
```suggestion
A primary cluster is a cluster with a control plane, which observes more than one
cluster in a multicluter mesh for endpoints and facilitates cross-cluster load balancing
```

rcaballeromx: @nmittler @smawson WDYT?

nmittler: I think the original text was correct. A single-cluster mesh consists of a single primary cluster.

frankbu: I was worried about the other end, "observes all clusters". Is that always true? Isn't it possible to have a primary cluster that only observes a subset of the clusters in the mesh, while another primary cluster handles others?

nmittler: Not in any documented configuration that I'm aware of. Unless you're talking about versioned control planes, which is slightly orthogonal.

content/en/docs/reference/glossary/primary-cluster.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Primary Cluster
+test: n/a
+---
+
+A primary cluster is a cluster with a control plane, which observes all
```
smawson, line 6: Can/should we link to the glossary entry for control plane here?

content/en/docs/reference/glossary/primary-cluster.md
```diff
@@ -0,0 +1,8 @@
...
+clusters in the mesh for endpoints and facilitates cross-cluster load balancing
+between clusters.
```
smawson, line 8: Things we should mention:
- Multiple primary clusters can be connected together in a single mesh (point to glossary entry for mesh)
- Primary clusters can act as the control plane for remote clusters (point to glossary entry for remote cluster).

I think the first one can replace the note about observing endpoints and facilitating cross-cluster LB, which seems a bit too much implementation detail of how they are connected, and not really part of the definition of a primary cluster.

content/en/docs/reference/glossary/remote-cluster.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Remote Cluster
+test: n/a
+---
+
+A remote cluster is a cluster that uses a remote profile to connect to a
+control plane residing outside of the cluster. The control plane can reside
+either in another cluster or external to the mesh as is the case with managed
```
smawson, line 8: We can link to the definition of primary cluster here right? A remote cluster can connect to a control plane running in a primary cluster or to a managed control plane.

We also have a glossary entry for managed control plane.

content/en/docs/reference/glossary/primary-cluster.md
dcberg: @frankbu is correct. A single istiod control plane can be configured to observe only a subset of a cluster.

Also the statement “which observes more than one cluster” is too strict. Is should be “which observes resources from one or more clusters...”.

content/en/docs/reference/glossary/primary-cluster.md
dcberg: I personally think the term “primary” is too vague. I prefer a name which is a bit more intention revealing such as “control cluster”.

content/en/docs/reference/glossary/primary-cluster.md
nmittler: @dcberg 

> @frankbu is correct. A single istiod control plane can be configured to observe only a subset of a cluster.

Just to be clear, are you referring to control plane revisions?

If we want to be more precise, maybe we can rephrase to something like:

```
A primary cluster is a cluster within a mesh that contains a control plane. The control plane observes configuration resources from a single source (e.g. Kubernetes API Server). It also observes service endpoints from one or more clusters in the mesh in order to facilitate cross-cluster load balancing between workloads.
```

content/en/docs/reference/glossary/primary-cluster.md
frankbu: I tend to agree "primary" is too vague, and so is "remote cluster" for that matter. Neither term really aligns very well with what they are. A primary cluster is a "self-controlled cluster" and a remote cluster is an "externally-controlled cluster" (the external control can come from the control plane in a self-controlled cluster or a control plane running outside of any cluster). I'm not sure that "self-controlled" and "externally-controlled" are very good names to use, but the point is that "primary" and "remote" don't really seem to be indicative of what they are.

content/en/docs/reference/glossary/primary-cluster.md
frankbu: > Multiple primary clusters can be connected together

I don't think what "connected together" means is clear. What's it trying to say? You can have more than one (synchronized) primary cluster in the same mesh for HA? I think it needs to say something more specific.

frankbu:
@nmittler @smawson @dcberg @rcaballeromx 

Here's a proposal for both entries. WDYT? (maybe overkill on the links?)

```
---
title: Control Cluster
---

A control cluster is a [cluster](/docs/reference/glossary/#cluster) with a
[control plane](/docs/reference/glossary/#control-plane). You can have more than
one (synchronized) control cluster in a single [mesh](/docs/reference/glossary/#service-mesh)
for HA or to reduce latency. Control clusters can act as the control plane for
[remote clusters](/docs/reference/glossary/#remote-cluster).


---
title: Remote Cluster
---

A remote cluster is a [cluster](/docs/reference/glossary/#cluster that uses a
remote profile to connect to a [control plane](/docs/reference/glossary/#control-plane)
residing outside of the cluster. A remote cluster can connect to a control plane
running in a [control cluster](/docs/reference/glossary/#control-cluster)
or to a [managed control plane](/docs/reference/glossary/#managed-control-plane).
```

smawson:
I don't like Control cluster and prefer Primary cluster.

frankbu:
@smawson Two reasons I don't think "primary" is a great name is because of the definition of the word. It means either first or most important. So if you create a mesh using a managed control plane, the "first" cluster is not a primary cluster. Also, since you can have multiple primary clusters, they can't both be the "most important" one.

smawson:
I had similar misgiving when it was initially proposed but I've mostly gotten past them because nothing better has been suggested.

Multi-primary is an established model in mysql at least: https://dev.mysql.com/doc/refman/8.0/en/group-replication-multi-primary-mode.html

The problem with Control cluster is that it has some negative connotations. Primary is also where K8s would like to head, and presumably they'll also run in multi-primary mode (for example with regional clusters).

For managed control planes there is still one cluster that has the config for the control plane, the control plane itself just doesn't run there. We can call that cluster the primary cluster (but not the control cluster!). Or we can call it the config cluster or something else.

nmittler:
```
A control cluster is a [cluster](/docs/reference/glossary/#cluster) with a
[control plane](/docs/reference/glossary/#control-plane). You can have more than
one (synchronized) control cluster in a single [mesh](/docs/reference/glossary/#service-mesh)
for HA or to reduce latency. Control clusters can act as the control plane for
[remote clusters](/docs/reference/glossary/#remote-cluster).
```

Let's get rid of `(synchronized)`, I'm not sure that term has been defined and the concept isn't necessary here ... it's fine to have eventually or non-consistent primaries in the same mesh.

Otherwise, LG.

```
A remote cluster is a [cluster](/docs/reference/glossary/#cluster) that uses a
remote profile to connect to a [control plane](/docs/reference/glossary/#control-plane)
residing outside of the cluster. A remote cluster can connect to a control plane
running in a [control cluster](/docs/reference/glossary/#control-cluster)
or to a [managed control plane](/docs/reference/glossary/#managed-control-plane).
```

Let's get rid of `remote profile` here, it's an implementation detail. The first sentence can be simplified to `A remote cluster is a cluster that connects to a control plane residing outside of the cluster`. Also, have we defined `managed control plane` previously? I don't think we did for the concepts page.

frankbu:
Updated proposal based on comments from @smawson and @nmittler 

@nmittler "managed control plane" is in the glossary, but I think you're right that it's not in the concepts.

```
---
title: Primary Cluster
---

A primary cluster is a [cluster](/docs/reference/glossary/#cluster) with a
[control plane](/docs/reference/glossary/#control-plane). You can have more than
one primary cluster in a single [mesh](/docs/reference/glossary/#service-mesh)
for HA or to reduce latency. Primary clusters can act as the control plane for
[remote clusters](/docs/reference/glossary/#remote-cluster).


---
title: Remote Cluster
---

A remote cluster is a [cluster](/docs/reference/glossary/#cluster that
connects to a [control plane](/docs/reference/glossary/#control-plane)
residing outside of the cluster. A remote cluster can connect to a control plane
running in a [primary cluster](/docs/reference/glossary/#primary-cluster)
or to a [managed control plane](/docs/reference/glossary/#managed-control-plane).
```

nmittler:
@frankbu https://github.com/istio/istio.io/pull/7591#issuecomment-647780899 LGTM

rcaballeromx:
@frankbu @nmittler @smawson I'll implement Frank's last suggestion then, agreed?

PR #7592, opened 2020-06-18
Update ref docs for 1.6.3

ericvn:
This updates the ref docs. It's a small change which is the sorting the output. Making this change now will set it to the same order going forward (and this change won't happen again).


shamsher31:
Please note:

> WARNING: THIS IS AN AUTO-GENERATED FILE, DO NOT EDIT. PLEASE MODIFY THE ORIGINAL SOURCE IN THE 'https://github.com/istio/istio' REPO
> source_repo: https://github.com/istio/istio

PR #7601, opened 2020-06-21
Automate release process

niwatorrri:
This PR is to automate the current release process by adding several make targets.

Quick summary of usage examples
- run `make prepare-1.7.0` when istio 1.7 source is branched
- run `make release-1.7.0` on the day of 1.7 release
- run `make release-1.7.0-dry-run` for a dry run of 1.7 release, which only creates a dry run release branch
- run `make build-old-archive-1.6.0` to generate the folder `archive/v1.6` and overwrite if it already exists

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


niwatorrri:
@davidhauck Please take a look. Part of README changes needs your input.

README.md
```diff
@@ -6,8 +6,8 @@
 
 ## istio.io
 
-This repository contains the source code for the [istio.io](https://istio.io),
-[preliminary.istio.io](https://preliminary.istio.io), and [archive.istio.io](https://archive.istio.io) sites.
+This repository contains the source code for the [istio.io](https://istio.io) and
```
ericvn, line 9: remove `the` before istio.io.

scripts/patch_release.sh
```diff
@@ -0,0 +1,61 @@
...
+---
+
+This release note describes what's different between Istio ${PREV_VERSION} and Istio ${NEW_VERSION}.
+" >> "${RELEASE_NOTE_PATH}/index.md"
+
+git add -A
+
+echo "
+[SUCCESS] Almost there! Now complete the release note, then commit and push all the changes.
+Release note is located at ${RELEASE_NOTE_PATH}/index.md"
```
ericvn, line 61: If you are making a template for the release doc, you might want to create one for the security doc if it's a security patch.  Inmost cases, these files will already be created so I'm not sure the template is all that necessary here.

README.md
```diff
@@ -72,59 +66,23 @@ version of Istio is 1.3 and you wish to introduce 1.4 which has been under devel
...
-1. Edit the file `Makefile.core.mk` and change the `SOURCE_BRANCH_NAME` variable to the
-name of the newly created source branches (in this case `release-1.4`).
-
-1. Edit the file `data/args.yml` and set the `source_branch_name` field to the name of the newly created source
-branches (in this case `release-1.4`).
-
-1. Run `make update_all` in order to retrieve the latest material from the source repositories.
-
-1. Commit the previous edits to your local git repo and push your **master** branch to GitHub.
+Run `make prepare-1.4.0`, and that's it. This will grab the latest material from the new istio source branch.
```
ericvn, line 69: Your usage examples use 1.7.x. I think the example of going from 1.3 to 1.4 is old. I might renumber them so it's going from 1.6.x to1.7.x to be a little more current.

nmittler:
@niwatorrri @davidhauck would it make sense to split this PR into individual targets so that we can focus the review on a single step? WDYT?

README.md
niwatorrri: Sounds good!

scripts/create_version.sh
```diff
@@ -0,0 +1,167 @@
...
+    fi
+
+    echo "Previous minor release: ${PREV_MINOR}"
+    echo "Upcoming minor release: ${NEXT_MINOR}"
+}
+
+archive_old_release() {
+    echo -e "\nStep 1: archive the old release branch"
+    if [ "${DRY_RUN}" == '1' ]; then
+        echo "Skipping step 1 in dry run"
```
davidhauck, line 67: This is probably something we should still do in the dry run, right? We should just be pushing the archived site to a different branch. We just shouldn't check in the changes to the archived branch

scripts/create_version.sh
```diff
@@ -0,0 +1,167 @@
...
+        return
+    fi
+
+    git checkout "release-${PREV_MINOR}"
+    git pull "${ISTIOIO_GIT_SOURCE}" "release-${PREV_MINOR}"
```
davidhauck, line 72: As above, let's try to avoid non ff merges

scripts/create_version.sh
```diff
@@ -0,0 +1,167 @@
...
+        git add -u
+        git commit -m "mark v${PREV_MINOR} as archived"
+        git push origin "release-${PREV_MINOR}"
+    fi
+
+    # complete the archive process in master
+    git checkout "${MASTER}"
+    git pull "${ISTIOIO_GIT_SOURCE}" "${MASTER}"
+
+    scripts/redo_archive.sh "redo-archive-${PREV_MINOR}.0"
```
davidhauck, line 92: feels weird to be calling "redo-archive" when we're doing the actual archive. Maybe rename that to something a bit different

scripts/create_version.sh
```diff
@@ -0,0 +1,167 @@
...
+
+    # major/minor release
+    CURR_MINOR="${MAJOR}.${MINOR}"     # current version
+    PREV_MINOR="${MAJOR}.$((MINOR-1))" # previous version
+    NEXT_MINOR="${MAJOR}.$((MINOR+1))" # next version
+
+    if [ "${DRY_RUN}" == '1' ]; then
+        CURR_MINOR="${CURR_MINOR}-dry-run"
+        git checkout "${MASTER}"
+        git pull "${ISTIOIO_GIT_SOURCE}" "${MASTER}"
```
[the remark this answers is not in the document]
niwatorrri: Is that the same as `git pull --ff-only`?

scripts/create_version.sh
niwatorrri: That would mean we'd create yet another branch like `master-dry-run` off master? Or are you suggesting only pushing a single folder `archive/v1.x` to a separate branch?

scripts/create_version.sh
niwatorrri: How about `fetch-archive`?

scripts/patch_release.sh
niwatorrri: I checked some of those notes and this template gives a different description for security patch. Maybe there's more that can be added to the template but if it's not necessary I shall delete this part.

ericvn: Yes, the files themselves are usually pretty different. And usually the person(s) doing the release will have them started before you would get to the point where the release would be tested (this script run). And most start with an earlier version of the file(s). I'd probably vote to remove the template creation anyway, and we'll want to merge the PRs at different times.

scripts/patch_release.sh
```diff
@@ -0,0 +1,61 @@
...
+PATCH=$3
+
+NEW_VERSION="${MAJOR}.${MINOR}.${PATCH}"
+PREV_VERSION="${MAJOR}.${MINOR}.$((PATCH-1))"
+RELEASE_BRANCH="release-${MAJOR}.${MINOR}"
+
+git checkout "${RELEASE_BRANCH}"
+git pull "${ISTIOIO_GIT_SOURCE}" "${RELEASE_BRANCH}"
+
+echo "Migrating to the new release ${NEW_VERSION}..."
```
ericvn, line 30: Also there would be two different security related processes. One where the changes are in the public, so in that case we can update ref docs if we wanted (I suspect there wouldn't be any) and update the release version for the tests. In the case where the security build is private, neither of the changes would be required.

I would change SECURITY_PATCH to PRIVATE_PATCH and move the go mod and go tidy into the check with the update docs.

davidhauck:
@nmittler Tbh, I'd be fine with it either way. It all does relate quite heavily on each other, but it's never bad to have smaller PRs that are more focused

scripts/create_version.sh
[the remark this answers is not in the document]
davidhauck: Yes it is

scripts/create_version.sh
davidhauck: Maybe "build-old-archive" or something along those lines.

scripts/create_version.sh
davidhauck: I mean that we should still go back to the old branch, archive it, and add it as a folder in the release branch

README.md
```diff
@@ -30,12 +30,10 @@ To learn how to edit and build this repo's content, please refer to
...
 ## Versions and releases
 
-Istio maintains three variations of its public site.
+Istio maintains two variations of its public site.
 
 - [istio.io](https://istio.io) is the main site, showing documentation for the current release of the product.
-
-- [archive.istio.io](https://archive.istio.io) contains snapshots of the documentation for previous releases of the product.
-This is useful for customers still using these older releases.
+[istio.io/archive](https://istio.io/archive) contains snapshots of the documentation for previous releases of the product. This is useful for customers still using these older releases.
```
ericvn, line 36: A few lines down (not exactly sure how to select it), it has `All three sites`. This should say `Both sites`?

README.md
```diff
@@ -68,134 +62,54 @@ then as soon as your PR is merged into master, it will be merged into the curren
...
-1. Create a PR with the go.* changes.
+tests to use the new release to verify automated doc testing passes. 
 
-Creating a new patch release involves modifying a few files:
+To create a new patch release, you need to
 
-1. Create the release note for the release by adding a markdown file in
-`content/en/news/<YEAR>/1.X.Y/index.md`, where 1.X.Y is the name of the release. This is where
-you describe the changes in the release.
+1. Run `make release-A.X.Y`, where `A.X.Y` is the name of the release. If it is a private security patch, run `PRIVATE_PATCH=true make release-A.X.Y`, which will prevent the branch from grabbing the latest docs.
```
ericvn, line 106: grabbing the latest docs and updating the istio reference for testing.

We probably still want a step to do that after the private release is ready and made public. So maybe we want to add some words around running again without the PRIVATE_PACTH=true when the release is made. Actually, since we have everything inside the PRIVATE_PATCH (we removed the templating), maybe we can change the code to remove PRIVATE_PATCH and just have the instructions say to run it when a public build qualification is started (as directed by the release managers) or after a private release is made.

niwatorrri: Not sure if I've understood it clearly... if PRIVATE_PATCH is removed then "after a private release is made," go mod and go tidy will still be run, which sounds like opposite of a previous [comment](https://github.com/istio/istio.io/pull/7601#discussion_r445184673).

ericvn: My thought was that we were left with only two basic items in this flow, the go.* update and the reference document update after removing the templates. I looked again today and there is actually a third part of the update, the data/args.yml update.

In your testing, do you know if we can update the args.yml before we add the files in the news directory and have a PR pass its tests (I know there is an issue if you add the files without updating the args.yml)? I suspect the answer is that we can't, and we need to keep the news files and args.yml change within the same PR.

If we have to merge the args.yml change with the news, then I think we can remove that piece from automation and just let the release managers worry about that. This leaves us with the two pieces: go.* and reference docs. It then makes sense to remove the PRIVATE_PATCH check and simply have the target run ~3 days before the release if it's built publicly or after the release is publish if built privately.

If we can merge the args.yml change early without breaking things, then that could help with cherrypicking the news files. @frankbu Feedback from you as well, but should the release managers create the release notes in _master_ and then cherrypick to the _release_ branch and update the args.yml in that PR, or should they make the change in the _release_ branch and then manually cherrypick the changes back to _master_ without the args.yml change? In the past, the former seemed to be common, but recently, the later has been more common (I think because that limit the bottleneck for PR merging to one release of istio.io on publishing day and less likely to need a force merge).

ericvn: Comment from @brian-avery on ordering: `The process we decided on in the PSC is release first and then master given that we need to make sure it works in the release branch and not block on that, and we can get master merged after that.`

niwatorrri: Per my test it seems that args.yml *can* be merged early without breaking things (#7634). So I guess we can run this as it is now. After the release note is completed in the release branch, both changes can then be cherry-picked to master.

ericvn: The problem I noted with merging early in your other PR is that it updates the latest version shown on the bottom of the page (which isn't released yet). So we shouldn't be updating that until the release is being published (at the same time as the release/security notes).

So I think we have a go.* update along with the reference doc update that runs ~3 days before the release for public releases (when asked by the person heading up the release), or after the release is published and public repos updated in the case of a private release.  In the case of the private built release we could ask the release them to do the run the automated job and then let that get cherrypicked on publish day. Actually, it might just be best to have the release team run and submit the PR to the public/private repo as needed for the updates.

There is another piece of work which is the doc updates at publish time (args.yml and notes). That just comes from the release side. First to the `release` branch and then manual cherrypick of the notes to `master`.

niwatorrri: Sounds good. I made changes to patch_release.sh and updated [README](https://github.com/istio/istio.io/pull/7601/files#diff-04c6e90faac2675aa89e2176d2eec7d8R104). Does this look good to you?

README.md
```diff
@@ -68,134 +62,58 @@ then as soon as your PR is merged into master, it will be merged into the curren
...
-1. Commit the previous edits to your local git repo and push your **master** branch to GitHub.
-
-1. Create a new release branch off of master, named as release-**major**.**minor** (in this case `release-1.4`). There is one
-such branch for every release.
-
-1. Edit the file `data/args.yml`. Set the `preliminary` field to `false`
-and the the `doc_branch_name` field to the name of the release branch (in this case `release-1.4`).
-
-1. Commit the previous edits to your local git repo and push your **release** branch to GitHub.
+1. Run `make release-1.7.0`. This make target will change some variables in `master` and `release-1.6` as needed, and create a new branch `release-1.7` for the new version.
```
nmittler, line 73: I believe that this is specifically for major/minor releases, correct? If so, we need to make that more clear.

README.md
```diff
@@ -68,134 +62,58 @@ then as soon as your PR is merged into master, it will be merged into the curren
 ### Creating a version
 
 Here are the steps necessary to create a new documentation version. Let's assume the current
-version of Istio is 1.3 and you wish to introduce 1.4 which has been under development.
+version of Istio is 1.6 and you wish to introduce 1.7 which has been under development.
```
nmittler, line 65: Does `creating a version` imply major/minor or does it also include patch releases?

README.md
```diff
@@ -68,134 +62,58 @@ then as soon as your PR is merged into master, it will be merged into the curren
...
 A few days before the patch release, the release managers should notify the Doc WG that the release
 is built and is starting it's long running qualification test. At this time, move the doc automation
-tests to use the new release to verify automated doc testing passes. To move to a new release
-(make sure you are in the patch's release branch):
-
-1. `go get istio.io/istio@1.X.Y`
+tests to use the new release to verify automated doc testing passes. 
 
-1. `go mod tidy`
+To create a new patch release, you need to
```
nmittler, line 104: How much of the above process applies to patch releases as well? None?  Either way, I think we should reorganize the content to be something like:

```
Creating a Major/Minor release
...
Creating a Patch release
...
```

WDYT?

README.md
niwatorrri: Correct. Let's change that.

README.md
niwatorrri: Sounds good. I'll change it right away.

niwatorrri:
After discussing further with @ericvn , there seem to be some complexities beyond what we had imagined, as there are images built in private repo but not synced back to the public repo. So let's say to leave the patch release part as it was and we won't automate that part for now.

scripts/prepare_release.sh
```diff
@@ -0,0 +1,37 @@
...
+
+[[ $1 =~ ^prepare-([0-9]\.[0-9]+)\.0$ ]] ||
+    { echo "Target format error: should be 'prepare-x.x.0', got '$1'"; exit 1; }
+
+VERSION="${BASH_REMATCH[1]}"
+
+git checkout "${MASTER}"
+git pull --ff-only "${ISTIOIO_GIT_SOURCE}" "${MASTER}"
+
+sed -i "s/^source_branch_name: .*$/source_branch_name: release-${VERSION}/" data/args.yml
```
nmittler, line 27: IIUC the purpose of this script is to update the istio/istio.io master branch to use a release branch of istio/istio?

Why would we not simply create a matching branch of istio/istio.io on the same day as the istio/istio branch is cut?  Then we would immediately have istio.io/release-x.y.z pointing to istio/release-x.y.z and istio.io/master could still be pointing at istio/master (i.e. the bleeding edge).

Otherwise, we lose track of istio/master WRT docs for the lifetime of the release. If someone were to break docs in master (not in the istio/release-x.y.z branch), we wouldn't know until the release is completed.

@davidhauck any thoughts?

README.md
```diff
@@ -50,152 +48,76 @@ are automatically reflected on preliminary.istio.io.
...
-are included on archive.istio.io is determined by the `TOBUILD` variable in this
-[script](https://github.com/istio/istio.io/blob/master/scripts/build_archive_site.sh).
-
 ### Publishing content immediately
 
 Checking in updates to the master branch will automatically update preliminary.istio.io, and will only be reflected on
 istio.io the next time a release is created, which can be several weeks in the future. If you'd like some changes to be
 immediately reflected on istio.io, you need to check your changes both to the master branch and to the
-current release branch (named release-XXX such as release-1.4).
+current release branch (named release-XXX such as release-1.7).
```
nmittler, line 56: Are releases always just named with 2 digits? If so maybe this should read `named release-<MAJOR>.<MINOR>`?

README.md
```diff
@@ -30,17 +30,15 @@ To learn how to edit and build this repo's content, please refer to
 
 ## Versions and releases
 
-Istio maintains three variations of its public site.
+Istio maintains two variations of its public site.
```
nmittler, line 33: Why two? There are 3 bullets here.

scripts/prepare_release.sh
nmittler: Ah I see. I guess our goal should be to ultimately change the process so that documentation work is done earlier in the release cycle. If that were the case, then there wouldn't be the mad flurry of docs changes at the end ... which would probably make everyone on the docs team a lot happier anyway :).

But since we're not there yet, I guess this is the lesser of evils.

README.md
niwatorrri:
```suggestion
current release branch (named `release-<MAJOR>.<MINOR>` such as `release-1.7`).
```
I believe so. Yeah, this should look clearer.

README.md
niwatorrri: There are only two now: istio.io and preliminary.istio.io. The archive.istio.io won't be used for archive any more and instead we'd use istio.io/archive.

nmittler: ah ok, thanks

README.md
```diff
@@ -50,152 +48,76 @@ are automatically reflected on preliminary.istio.io.
...
 
 Checking in updates to the master branch will automatically update preliminary.istio.io, and will only be reflected on
 istio.io the next time a release is created, which can be several weeks in the future. If you'd like some changes to be
 immediately reflected on istio.io, you need to check your changes both to the master branch and to the
-current release branch (named release-XXX such as release-1.4).
+current release branch (named `release-<MAJOR>.<MINOR>` such as `release-1.7`).
 
 This process can be taken care of automatically by our infrastructure. If you submit a PR
 to the master branch and annotate the PR with the `actions/merge-to-release-branch` label,
 then as soon as your PR is merged into master, it will be merged into the current release branch.
```
frankbu, line 60:
```suggestion
to the master branch and annotate the PR with the `cherrypick/release-<MAJOR>.<MINOR>` label,
then as soon as your PR is merged into master, it will be merged into the specified release branch.
```

README.md
```diff
@@ -50,152 +48,76 @@ are automatically reflected on preliminary.istio.io.
...
-1. Edit the file `Makefile.core.mk` and change the `SOURCE_BRANCH_NAME` variable to the
-name of the newly created source branches (in this case `release-1.4`).
-
-1. Edit the file `data/args.yml` and set the `source_branch_name` field to the name of the newly created source
-branches (in this case `release-1.4`).
-
-1. Run `make update_all` in order to retrieve the latest material from the source repositories.
-
-1. Commit the previous edits to your local git repo and push your **master** branch to GitHub.
+Run `make prepare-1.7.0`, and that's it. This will grab the latest material from the new istio source branch.
```
frankbu, line 69: This sounds pretty mysterious. You should say what "latest material" it is grabbing and where's it's putting it. Isn't it it just updating the istio/istio.io master branch to use the release branch of istio/istio for doc testing? Or is there more? Either way, you should say exactly what "preparing" entails.

README.md
```diff
@@ -50,152 +48,76 @@ are automatically reflected on preliminary.istio.io.
...
-such branch for every release.
-
-1. Edit the file `data/args.yml`. Set the `preliminary` field to `false`
-and the the `doc_branch_name` field to the name of the release branch (in this case `release-1.4`).
-
-1. Commit the previous edits to your local git repo and push your **release** branch to GitHub.
+1. Run `make release-1.7.0`. This make target will change some variables in `master` and `release-1.6` as needed, and create a new branch `release-1.7` for the new version.
 
-#### Updating istio.io
+    - For a dry run before official release, run `make release-1.7.0-dry-run`, which will only create a new branch `release-1.7-dry-run`, and not touch any other branches.
```
frankbu, line 75: This seems out of place in the "#### On the day of the release" section, since a dry run would likely be done a day or so ahead of the release day.

PR #7623, opened 2020-06-25
fix(customize-metrics): typo

scraly:
[x ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
@frankbu Looks like this has the label, but missed being cherrypicked as a FYI

PR #7634, opened 2020-06-26
[DO NOT REVIEW] Test if args.yml can be merged in advance

niwatorrri:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/data/args.yml
+++ b/data/args.yml
@@ -2,7 +2,7 @@
 version: "1.6"
 
 # The full Istio version identifier the docs describe
-full_version: "1.6.3"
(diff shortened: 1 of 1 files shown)


ericvn:
I did a test on my Mac on Friday, and I was able to serve/lint with an updated args.yml. The only noticeable thing was the new update was being displayed on the bottom of the page (which would be incorrect since it's not released). I didn't see any failures.

PR #7652, opened 2020-06-30
Updates post 1.6.4. Move tests and update ref docs.

ericvn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
I think we have an issue when we build in private, publish to public, and then try to run the tests on that build.

We don't ever create the build for the commit in public, so the images to go with the release don't exist Istio-testing repository like all other public builds:

      "message": "Back-off pulling image \"gcr.io/istio-testing/pilot:1.6-alpha.58f551e08f08addfd81783e1c2ed1eabeb836168\""

PR #7659, opened 2020-07-02
Make customising external charts command copyable

shamsher31:
Good to marge after https://github.com/istio/istio.io/pull/7655, because it has new `--manifests` flag.


frankbu:
Actually, this seems to be reverting #7655, i.e., going back to --charts.

That said, I don't think we want this change anyway. These aren't actual instructions that users need to copy, they're just examples to help explain the possible kinds of values accepted.

PR #7666, opened 2020-07-06
Update removing separate files (security-2020-007). Allows seeing content on page.

ericvn:
This PR was one I worked on that removed the separate files and updated the content into the document for easier viewing.

I'm creating this in case there is a desire to use some of that work. It's perfectly fine to leave the docs as they are.

@brian-avery @mandarjog Thoughts on this?

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


frankbu:
This does look like a better version, but do we really want to change an existing security bulletin after it's been published? Maybe we should just make sure to do it this way in the future.

brian-avery:
I like this implementation more because:
* it's more obvious what's being done
* it keeps all of the content in the docs rather than creating a new storage location for related yaml files

That said, I agree with Frank here. I don't think there's anything technically wrong with the documentation as it currently stands and we probably shouldn't be modifying security releases after the fact unless there is something incorrect about them. I see pros and cons and could go either way.

frankbu:
It sounds like we all agree that this is a better, so we should make sure to in-line the yaml this way in the future. It's not worth the risk, however, to now change the existing bulletin, which has been out for a while and working fine as is.

PR #7701, opened 2020-07-10
networking: add inbound tunnel port

lambdai:
Reserve the istio port 15008 for better transport security 

Part of https://github.com/istio/istio/issues/24540

Signed-off-by: Yuchen Dai <silentdai@gmail.com>


DamianFekete:
`Envoy Tunnel port (Inbound)` doesn't say that much.

It would be nice to have some more documentation about what that means (e.g. a link to a page describing the feature, or maybe even to https://docs.google.com/document/d/1TARjqHylJPEkEo8XCTGygHk-4PygN4KnTZRJy_HyTFc/edit# or https://github.com/istio/istio/issues/24540 until the feature is complete). 

I've spent some time just grepping the source code to see where port is used, and it is currently just used for the iptables filter.

PR #7706, opened 2020-07-10
Update open usage blog post to draw distinctions in the use of "we"

ssuchter:
In an Istio Steering Committee discussion, we decided to update this blog post to make it more clear which uses of "we" referred to Google, and which referred to the Istio Steering Committee. This is accomplished by only using the word "we" to refer to the Istio Steering Committee, and using "Google" when that is the relevant intent.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ssuchter:
@linsun @craigbox PTAL.

craigbox:
LGTM.

Lin, I trust there's no concern about the "we" in the last and third-to-last paragraphs of the blog? It might pay to say "the steering committee" instead of the first we, to set the context.

ssuchter:
Yes, I saw the "we" in the last couple paragraphs, but those seemed to be things that the Steering Committee would reasonably say. @linsun, if you feel differently, I can change them too.

linsun:
@ssuchter thanks for the update!  Changes Looks good overall.  I think it could be more accurate to just say Google in the last few "we" in the last paragraph given IBM's blog - https://developer.ibm.com/blogs/istio-google-open-usage-commons/

ssuchter:
Ok, fixed. PTAL. The only remaining use of "We" is in "we were proud to appoint Neeraj..." - I thought that should stay as "we", because it definitely wasn't Google who did that.

linsun:
BTW should be Podder -> Poddar

cc @nrjpoddar

nrjpoddar:
> BTW should be Podder -> Poddar
> 
> cc @nrjpoddar

Even I missed it..

frankbu:
@ssuchter Please fix the lint errror so this can merge.

ericvn:
@sshucker The lint error is actually that the incorrect spelling `Podder` is in the .spelling file. Will need the update there as well.

ssuchter:
And ironically, I'm @ssuchter, not @sshucker. Thanks for the pointer, @ericvn.

PR #7669, opened 2020-07-07
Update istio/istio ref and reenable tests

ericvn:
Update the istio/istio reference to most recent. This should fix the current issue with the failing tests where they are using the `latest` proxyv2 image which then fails to start.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #7714, opened 2020-07-11
Bump istio/istio dep and enable TLS origination tests

nschhina:
Disabled here due to a bug in istio/istio: https://github.com/istio/istio.io/pull/7669 

Please confirm why auth tests were disabled as they shouldn't be related to the bug fix I merged for tls origination


frankbu:
Strange that the checks summary says that the k8s-tests_istio.io succeeded, even though 4 of the tests actually failed.

Three of the failing tests are failing on calls to `istioctl install`, soit seems that either `install` is broken in this istio/istio build, or `istioctl` hasn't been updated properly.

The 4th failing test seems to be a different problem that the tls-origination that was fixed. It had worked previously, but it looks like maybe istioctl wait is unreliable again?

nschhina:
yeah none of the errors now seem to be from faulty TLS origination API/task, I don't know how to help here, maybe @howardjohn knows what's broken in istio/istio now

howardjohn:
Install is 100% broken by #25384. Opened https://github.com/istio/istio/issues/25447. Need to revert the PR or forward fix

howardjohn:
no clue why it fails but says it passes though

nschhina:
Okay we can hold off on mergining this, I'll bump istio/istio after that change is reverted

frankbu:
@nschhina Unless you have time to fix it, please disable the one remaining failed test, and we can then merge this PR and fix the mtls test later.

nschhina:
@frankbu I'm trying to figure out why this case is failing, I would say we only merge after this is resolved. Also I'm not sure if updating istio/istio dep to master is a good idea right now. Master currently is broken and prow is reporting PASS status for all test cases, merging to istio/istio is currently blocked

PR #7725, opened 2020-07-13
Fix flaky mTLS origination test

nschhina:
_wait_for_istio is perhaps buggy as locally running these tasks works just fine. Need better way to triage


frankbu:
Since we suspect that the problem is that `istioctl x wait` is flaky, we should really be trying to get that fixed, rather than removing the calls. We do want to put them back eventually. As a temporary fix, we should comment out the lines, and add a TODO comment to put them back later, rather than just removing them and forgetting about it.

There seem to be other tests that occasionally fail as well in similar calls to `istioctl x wait`. Seems like maybe some kind of timing issue where the status update get's lost sometimes? cc @therealmitchconnors 

Maybe a better temporary fix, instead of commenting out the 4 calls in mtls_test.sh, we should comment out the body of `_wait_for_istio` in tests/util/helpers.sh (make it a no-op) and just rely on the retry behavior of the _verify calls for now.

therealmitchconnors:
Is there somewhere I can see examples of these flakey test runs?

nschhina:
> Is there somewhere I can see examples of these flakey test runs?

https://prow.istio.io/view/gcs/istio-prow/logs/k8s-tests_istio.io_postsubmit/678

therealmitchconnors:
Looking at Testgrid, this test is not flakey.  It is consistently failing on the same distribution line since 7/13.  Is there an easy way to get the pilot logs from one of these tests?  How do we know what version of Istio the test is running against?

nschhina:
Post submit might not be flaky, but we observed flaky behavior during a PR

therealmitchconnors:
Already posted this to the tracking issue, but this is not a `wait` flake, but a true failure to propagate LDS config to egress.

https://prow.istio.io/view/gcs/istio-prow/logs/k8s-tests_istio.io_postsubmit/680

frankbu:
The current failure looks like it's caused by @nschhina commenting out the `istioctl x wait` call in helpers.sh, which is causing a problem with ingress task. To see the the wait problem, we need to undo that change.

nschhina:
If problem is in LDS, this won't fix it. I'll track that issue to see when it gets resolved and then upgrading istio/istio dep should fix this

PR #7731, opened 2020-07-15
update proxy-config output

shamsher31:
Few column updates from https://github.com/istio/istio/pull/24867#issuecomment-655211507

proxy-status = rename `PILOT` > `ISTIOD`
proxy-config = New column `DESTINATION RULE` for `cluster` and `listeners`

### DIFF
--- a/content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
+++ b/content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
@@ -33,15 +33,15 @@ receiving configuration or is out of sync then `proxy-status` will tell you this
 
 {{< text bash >}}
 $ istioctl proxy-status
-NAME                                                   CDS        LDS        EDS        RDS          PILOT                       VERSION
-details-v1-6fc55d65c9-x6qz8.default                    SYNCED     SYNCED     SYNCED     SYNCED       istiod-56b6dd58dd-fx5tx     1.6.0-rc.2
-istio-ingressgateway-74f5cf489f-xqcnh.istio-system     SYNCED     SYNCED     SYNCED     NOT SENT     istiod-56b6dd58dd-fx5tx     1.6.0-rc.2
-productpage-v1-7f44c4d57c-5f99x.default                SYNCED     SYNCED     SYNCED     SYNCED       istiod-56b6dd58dd-fx5tx     1.6.0-rc.2
-prometheus-68b46fc8bb-dc965.istio-system               SYNCED     SYNCED     SYNCED     SYNCED       istiod-56b6dd58dd-fx5tx     1.6.0-rc.2
-ratings-v1-6f855c5fff-mlqmv.default                    SYNCED     SYNCED     SYNCED     SYNCED       istiod-56b6dd58dd-fx5tx     1.6.0-rc.2
-reviews-v1-54b8794ddf-tc8hd.default                    SYNCED     SYNCED     SYNCED     SYNCED       istiod-56b6dd58dd-fx5tx     1.6.0-rc.2
-reviews-v2-c4d6568f9-4flql.default                     SYNCED     SYNCED     SYNCED     SYNCED       istiod-56b6dd58dd-fx5tx     1.6.0-rc.2
-reviews-v3-7f66977689-hwknn.default                    SYNCED     SYNCED     SYNCED     SYNCED       istiod-56b6dd58dd-fx5tx     1.6.0-rc.2
+NAME                                                   CDS        LDS        EDS        RDS          ISTIOD                      VERSION
+details-v1-558b8b4b76-qzqsg.default                    SYNCED     SYNCED     SYNCED     SYNCED       istiod-6cf8d4f9cb-wm7x6     1.6.5
+istio-ingressgateway-66c994c45c-cmb7x.istio-system     SYNCED     SYNCED     SYNCED     NOT SENT     istiod-6cf8d4f9cb-wm7x6     1.6.5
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
```diff
@@ -379,6 +393,6 @@ To find out the Envoy version used in deployment, you can `exec` into the contai
 {{< text bash >}}
 $ kubectl exec -it prometheus-68b46fc8bb-dc965 -c istio-proxy -n istio-system pilot-agent request GET server_info
 {
- "version": "12cfbda324320f99e0e39d7c393109fcd824591f/1.14.1/Clean/RELEASE/BoringSSL"
+ "version": "80ad06b26b3f97606143871e16268eb036ca7dcd/1.14.3-dev/Clean/RELEASE/BoringSSL"
```
ericvn, line 396: Minor nit. Do we want to have a `-dev` in the output? I notice that a `-rc.2` was removed earlier, but that was at least a release that would have been available on-line.

shamsher31: I agree `dev` doesn't look good. But this is the output of running following command.
```
kubectl exec -it prometheus-68b46fc8bb-dc965 -c istio-proxy -n istio-system pilot-agent request GET server_info
``` 

We can keep the old output and not update only this version info ?

content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
```diff
@@ -154,25 +154,39 @@ to send a request from the `productpage` pod to the `reviews` pod at `reviews:90
...
+    0.0.0.0       15010 ALL                                              PassthroughCluster
+    10.100.93.102 15012 ALL                                              Cluster: outbound|15012||istiod.istio-system.svc.cluster.local
+    0.0.0.0       15014 App: HTTP                                        Route: 15014
+    0.0.0.0       15014 ALL                                              PassthroughCluster
+    0.0.0.0       15021 ALL                                              Inline Route: /healthz/ready*
+    10.111.121.13 15021 App: HTTP                                        Route: istio-ingressgateway.istio-system.svc.cluster.local:15021
+    10.111.121.13 15021 ALL                                              Cluster: outbound|15021||istio-ingressgateway.istio-system.svc.cluster.local
+    0.0.0.0       15090 ALL                                              Inline Route: /stats/prometheus*
+    10.111.121.13 15443 ALL                                              Cluster: outbound|15443||istio-ingressgateway.istio-system.svc.cluster.local
     {{< /text >}}
```
frankbu, line 190: I don't think it's sufficient to just change this output. Seems to me the surrounding text will need to be updated, since it isn't a correct anymore.

shamsher31: I think the `MATCH` column is self-explanatory with more details as compare to previous version. It indicates `HTTP Plain`, `HTTP TLS`, `TCP TLS` and also displays complete address and ports. 

The surrounding text with custom markup will mess up with the indentation because of the new `DESTINATION` column which occupies most of the screen.

frankbu: Sorry, I meant generally, the surrounding text outside of the updated `{{< text >}}` sections. For example, just below it talks about `useOriginalDst`, which seems to be deprecated. It should probably try to be a little more clear about how to interpret the MATCH and DESTINATION rule fields, although in fairness, it wasn't very good at that before either.

shamsher31: Gotcha, I will update the text providing info related to what `proxy-config listeners` output.

shamsher31: Haven't found much info that can fit here. Only described what is `PassThroughCluster` and `Route`. @frankbu PTAL.

content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
```diff
@@ -154,30 +154,42 @@ to send a request from the `productpage` pod to the `reviews` pod at `reviews:90
...
 
-1. From the above summary you can see that every sidecar has a listener bound to `0.0.0.0:15006` which is where IP tables routes all inbound pod traffic to and a listener bound to `0.0.0.0:15001` which is where IP tables routes all outbound pod traffic to. This listener has `useOriginalDst` set to true which means
-it hands the request over to the listener that best matches the original destination of the request.
-If it can't find any matching virtual listeners it sends the request to the `PassthroughCluster` which connects to the destination directly.
+1. From the above summary you can see that every sidecar has a listener bound to `0.0.0.0:15006` which is where IP tables routes all inbound pod traffic to and a listener bound to `0.0.0.0:15001` which is where IP tables routes all outbound pod traffic to. The `PassThroughCluster` is the default cluster traffic is directed to if no listeners, routes, clusters are found during the lookup. A `Route` is both a specification of how to match a request as well as an indication of what to do next (e.g., redirect, forward, rewrite, etc.).
```
frankbu, line 192: Talking about `Route` seems out of place in the description of the following example.
```suggestion
1. From the above summary you can see that every sidecar has a listener bound to `0.0.0.0:15006` which is where IP tables routes all inbound pod traffic to and a listener bound to `0.0.0.0:15001` which is where IP tables routes all outbound pod traffic to. The `0.0.0.0:15001` listener hands the request over to the virtual listener that best matches the original destination of the request, if it can find a matching one. Otherwise, it sends the request to the `PassthroughCluster` which connects to the destination directly.
```

content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
shamsher31: Done

PR #7732, opened 2020-07-15
Fix for Egress Gateway TLS Origination Task

GregHanson:

Related issue: https://github.com/istio/istio/issues/23910
related istio PR: https://github.com/istio/istio/pull/25536

Steps in docs currently broken:
https://preliminary.istio.io/latest/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/

[ ] Docs
[ ] Installation
[ X ] Networking
[ ] Policies and Telemetry
[ X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


GregHanson:
PR with necessary bug fixes https://github.com/istio/istio/pull/25588

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
```diff
@@ -104,10 +104,12 @@ be done by the egress gateway, as opposed to by the sidecar in the previous exam
       servers:
       - port:
           number: 80
-          name: http-port-for-tls-origination
-          protocol: HTTP
+          name: https-port-for-tls-origination
+          protocol: HTTPS
```
nschhina, line 108: does HTTPS even work on port 80?

GregHanson: @nschhina from a conceptual networking standpoint or from Istio configuration point of view? There are almost no Egress Gateway listeners by default as they need to be explicitly defined by a Gateway resource, so creating a HTTPS listener on port 80 will not interfere with any default port 80 HTTP traffic on the egress gateway as there is none by default.

nschhina: I see, just don't think this is a good practice to listen for HTTPS on standard HTTP port even if there's no conflicts. Wouldn't the PR istio/istio#25588 make these changes you made useless anyways? Maybe waiting for that bug fixed to get merged is better

GregHanson:
>Another way of making this task work is to simply Disable TLS between sidecar and gateway. And originate the TLS at gateway alone, this also is a very good use case of TLS origination at Gateways by delegating the connection from plaintext to TLS

@nschhina This is what is in the docs currently for the Egress Gateway TLS Origination example.  Auto MTLS currently does not apply to sidecar-to-egressgateway since the egress pod does not have the `tlsMode=istio` label nor the default mTLS listener configured to accept that kind of traffic by default like proxies attached to apps within the mesh.  Most customers have a requirement that all traffic in mesh must be secure.  We can re-add the tabs for secure/insecure sidecar-to-egress gateway so people can choose between HTTP or mTLS traffic between these services (see Step 3 in old 1.4 docs [here](https://archive.istio.io/v1.4/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/#perform-tls-origination-with-an-egress-gateway)).

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
GregHanson: istio/istio#25588 is a bug fix so communication between sidecar's and egress gateway can communicate securely with manually configured Istio MTLS.  It will not resolve the problems that exist in the docs without the changes provided here - both are required, not one or the other.  Auto MTLS broke the security aspect because auto mtls traffic between egress and proxy is not configured by default.  SDS changes broke additional parts because the certs were physically removed from the pods and placed in k8s secrets instead

Egress Gateway TLS Origination steps work as is on istio.io, but traffic is not secure - so we need to provide an example for this since it is required for most customers

Egress Gateway with MTLS Origination steps do not work right now and as is, and need istio/istio#25588 in addition to the changes I have made to the snips

nschhina: Got it, can you also enable the last check in mtls_test.sh as well so we know for sure nothing is broken after istio/istio dep is updated with that fix. As for simple TLS example it 100% needs to be updated

sebastienrainville: Is there a valid reason why the example still does https over port 80, even as of today? As @nschhina also noted I don't think it's a good practice to listen for HTTPS on port 80.

On a side note, this is a bit scary:
> Auto MTLS broke the security aspect because auto mtls traffic between egress and proxy is not configured by default. SDS changes broke additional parts because the certs were physically removed from the pods and placed in k8s secrets instead

Have those issues been resolved?

Issue #7672, opened 2020-07-07
Display if a doc page has been automated on istio.io


linsun:
@frankbu @brian-avery and doc team has done a lot of good work on istio.io test automation and there are metadata today in each page to indicate if a doc page has been automated with our new test framework. It is desirable to reflect that in the UI itself so users would have the visibility into that info as they try each page.

@qwertyuiop888 has kindly offered to look into this and been prototyping this.

cc @davidhauck

qwertyuiop888:
This is a great idea. An indicator would help users judge if problems running commands are their mistake, or possibly a problem with documentation. This is a prototype indicator.

![Screen Shot 2020-07-10 at 12 20 33 PM](https://user-images.githubusercontent.com/20189462/87181398-e2e1d780-c2a7-11ea-832e-dacfdbd40904.png)

linsun:
@qwertyuiop888 LG!  What does the page that has automated testing look like?

linsun:
Since you are already working on it :)

qwertyuiop888:
It would look something like this. Right now it is using the checkmark icon, but it would be better with an icon more similar to the clock icon for reading time. 
![Screen Shot 2020-07-10 at 5 01 03 PM](https://user-images.githubusercontent.com/20189462/87206278-010efe00-c2cf-11ea-9030-f820aaaaf6a4.png)

linsun:
@qwertyuiop888 Thanks!  I was thinking the first page, it could show "Needs Automation" (which could link to the automation page @frankbu mentioned how to contribute to docs to help automate a page).

For the 2nd page, it could show "Commands Automated".

cc @brian-avery @ericvn for input

frankbu:
@qwertyuiop888 what do you mean by `but it would be better with an icon more similar to the clock icon for reading time`? I think a green check mark or red x next to word `TESTED` would be good (no time indicator), since it means that the doc is tested against the release of the docs indicated on top left corner of the page. The word `TESTED` should be a clickable link that takes users to the doc testing README so that users can find out exactly what it means, and also encourage them to help write tests for docs that are still untested.

linsun:
@frankbu thanks for the input!  I was debating on words `TESTED` @qwertyuiop888 used.  My only concern is we still do manual testing each release, certainly we don't have bandwidth to test each page with each RC or final release build, however many of these unautomated pages are still tested.  That is why I was hesitate to use words `NOT TESTED` on the page that is not automated yet.

[...]

PR #7734, opened 2020-07-15
Automation Indicator

qwertyuiop888:
fixes: #7672 

[ x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


qwertyuiop888:
Since we are just adding automation to the docs, it would be nice to have an easy way to view automation status on a website. This adds a small indicator next to the time, with a green check for automated tested and a grey check for tested by hand. 

It would be useful for users to see automation status, as they may want to contribute by writing a test after reading the page

Screen Shot 2020-07-14 at 2 56 01 PM

Screen Shot 2020-07-14 at 2 55 53 PM

layouts/partials/primary_top.html
```diff
@@ -104,6 +104,26 @@ <h1 id="title">
                                         <span>&nbsp;</span>
                                         {{- printf (i18n "page_reading_time") .Page.ReadingTime -}}
                                     </span>
+
+                                    <span>&nbsp;</span>
+
+                                    <span>
+                                        {{if isset .Params "test"}}
+                                            {{ if(.Params.test) }}
```
frankbu, line 112: There a 3 possible values of .Params.test:
1. "yes"
2. "no"
3. "n/a"

1 should display green check, 2 should display grey check, 3 should display nothing (same as not isset).

qwertyuiop888: Ah sorry, The isset check around the code makes sure it is not displayed if "n/a"

qwertyuiop888:
Im also pretty confused as to why reading the value yes / no as a boolean works. There does not seem to be any documentation or issues about it online.

i18n/en.toml
```diff
@@ -297,3 +297,13 @@ other = "Document"
 
 [log_change]
 other = "Change"
+
+[tested]
+other = "tested"
+
+[has_test]
+other = "An automated test is available for this document."
```
frankbu, line 305:
```suggestion
other = "An automated test is available for this page. Click for details."
```

i18n/en.toml
```diff
@@ -297,3 +297,13 @@ other = "Document"
...
+
+[needs_test]
+other = "This document has only been tested by hand. Help create an automated test!"
```
frankbu, line 308:
```suggestion
other = "No automated test is available for this page. Click for details or to help create one."
```

frankbu:
@qwertyuiop888 From the feedback we've heard so far, I think we need to be careful to not imply anything about how well a page is tested (either manual or auto). We should make it clear that this is just an indication of whether or not there is an automated test for the page or not. So, in that light, my recommendation is to change the tag from `TESTED` to just `TEST` or maybe (`PAGE TEST`). This is more clearly saying whether or not there is a test case for the doc, not whether or not the doc has been tested one way or another.

Using that tag, then I think we should use an x icon for the "no" case (instead of checkmark), but make the x grey, instead of red (because it's still true that red x generally implies something is broken).

- test: yes ->  (green check) PAGE TEST
- test: no -> (grey x) PAGE TEST

How about changing it like this (and the hover text as I suggested in in-line comments) and then we can merge it and see how it goes. If we get more feedback or better ideas later, we can change it in a future PR. WDYT?

qwertyuiop888:
That sounds good. It also solves the visibility issues with the same icons!

PR #7736, opened 2020-07-15
Print proxy status when debug fails for root cause

therealmitchconnors:
This will help us understand if the distribution really has failed, or if there is a bug in wait..


therealmitchconnors:
We could also run wait in verbose mode, but that would clutter successful output...

PR #6194, opened 2019-12-24
Simplify tls origination egress gateways

vadimeisenbergibm:
1. Use openssl to generate certificates and keys
2. Use client certificate/key
3. Use mutual TLS between the sidecar and the gateway, do not provide both options (with and without mutual TLS)
4. Access my-nginx.mesh-external.svc.cluster.local directly instead of using fake hostname nginx.com. No need for service entries/virtual services
    for nginx.com
5. Use predefined secret names for the istio egress gateway: egressgateway-certs and egressgateway-ca-certs
 6. Use ISTIO_MUTUAL at the egress gateway, for the server's TLS mode

[ ] Configuration Infrastructure
[X ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #7754, opened 2020-07-20
Simplify TLS Origination with Gateways Task

nschhina:
Clone of https://github.com/istio/istio.io/pull/6194 with a few changes.

1. Not sure why we need to originate direct TLS from sidecar in this example, this requires users to create client cert secrets and deploy a custom sleep example with file-mounts. This seems unnecessary in this example, correct me if I am wrong. We do this in Direct TLS user task anyways, if the point is to show how client certificates are required, I'm not sure it's worth as this user task is already so long.

2. While I understand the reason behind using egressgateway secret name matching so we don't have to patch gateway deployment and perform a restart with file paths, this kind of makes this scenario unrealistic as we are then assuming user is only performing TLS origination to one single upstream server. I added a warning stating that if more than 1 client certs need to be mounted, user is expected to patch gateway deployment.


vadimeisenbergibm:
@nschhina 

1. Performing direct TLS from the sleep container is performed for testing the mutual TLS server. It appears under section "Deploy a container to test the NGINX deployment".

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
```diff
@@ -398,248 +388,55 @@ to hold the configuration of the NGINX server:
...
+    
+        otherwise, you have to manually inject the sidecar before deploying the `sleep` application:
+    
+        {{< text bash >}}
+        $ kubectl apply -f <(istioctl kube-inject -f @samples/sleep/sleep.yaml@)
+        {{< /text >}}
+    
+        Note that any pod that you can `exec` and `curl` from would do.
 
-1.  Create Kubernetes [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to hold the client's and CA
```
vadimeisenbergibm, line 449: Here you remove the testing the NGINX server. I like to provide a test for each step to verify that each step works. This makes the tasks long though.

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
```diff
@@ -398,248 +388,55 @@ to hold the configuration of the NGINX server:
...
-    {{< /text >}}
-
-### Redeploy the egress gateway with the client certificates
+### Mount client certificates to the egress gateway
 
 1. Create Kubernetes [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to hold the client's and CA
    certificates.
 
     {{< text bash >}}
-    $ kubectl create -n istio-system secret tls nginx-client-certs --key nginx.example.com/4_client/private/nginx.example.com.key.pem --cert nginx.example.com/4_client/certs/nginx.example.com.cert.pem
```
vadimeisenbergibm, line 573: But this way the readers will not know how to patch their deployment once they will need to mount client certificates for different sites. So maybe, instead of warning them, it is better just to show them how to patch the deployment?

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
nschhina: I purpose a different approach to this then, how about we do a SIMPLE TLS origination with CaCert from sleep first, show that it fails, then update the DR to do a MUTUAL TLS and show it succeeds? I think this is more in line with what this user task is for

vadimeisenbergibm:
I am fine with the PR except for the comments I left. @frankbu what do you think regarding my comments?

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
nschhina: While I agree with how we should maybe add more information about patching, there is a new approach to this now. Which uses SDS with no file mounts required for certs and it's coming with 1.7 release. Instead of investing more time in this user doc which would eventually be deprecated in 1.9, we should focus on letting users migrate to SDS approach instead

vadimeisenbergibm: I mean to leave the example as it is now, with patching the deployment?

nschhina: No I mean leaving the example with this simplification + warning but I don't mind adding more context to the warning if need be

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
vadimeisenbergibm: That could be OK, it is an indirect test. I just like to show how you can perform the test directly from a container. Mounting certificates to a container and performing Simple/Mutual TLS is also useful for Istio users in general. Agree that it makes the task longer.

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
vadimeisenbergibm: Why this user doc will be deprecated in 1.9?

nschhina: There's plans to default to SDS instead of file-mounts for egress TLS. Same was done for ingress TLS with file-mount doc deprecated in 1.5 or 1.6

vadimeisenbergibm: > his user doc which would eventually be deprecated in 1.9
you probably meant the file-mounting aspect will be deprecated in 1.7, not the whole doc.

so don't you want to switch to SDS already in this task? to show SDS with client certificates per specific site?

nschhina: SDS is different task than this one for now. The doc will be deprecated in 1.9(hopefully), this approach is identical to the one taken for ingress SDS which worked out fine. This doc will be renamed with "file-mount" suffix. So the users can migrate to SDS over next two releases and then file-mount doc will be deprecated. The support for file-mounts will still exist but we'll not document it

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/snips.sh
```diff
@@ -300,212 +294,20 @@ EOF
...
-kubectl -n istio-system patch --type=json deploy istio-egressgateway -p "$(cat gateway-patch.json)"
+snip_mount_client_certificates_to_the_egress_gateway_1() {
+kubectl create secret tls istio-egressgateway-certs --key client.example.com.key --cert client.example.com.crt -n istio-system
+kubectl create secret generic istio-egressgateway-ca-certs --from-file=example.com.crt -n istio-system
 }
 
-snip_redeploy_the_egress_gateway_with_the_client_certificates_4() {
-kubectl exec -n istio-system "$(kubectl -n istio-system get pods -l istio=egressgateway -o jsonpath='{.items[0].metadata.name}')" -- ls -al /etc/istio/nginx-client-certs /etc/istio/nginx-ca-certs
+snip_mount_client_certificates_to_the_egress_gateway_2() {
+kubectl exec -n istio-system $(kubectl -n istio-system get pods -l istio=egressgateway -o jsonpath='{.items[0].metadata.name}') -- ls -al /etc/istio/egressgateway-certs /etc/istio/egressgateway-ca-certs
```
nschhina, line 310: Seems like some change was introduced in Istio against automatic secret watching for `istio-egressgateway-certs` and `istio-egressgateway-ca-certs`. Because I tried this example with latest build and I'm not getting any mounted certs. Maybe we should just keep the deployment patch example entirely.

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
vadimeisenbergibm: So I would continue showing deployment patching with file-mounting for this document, and write the doc with SDS, to perform migration as you describe.

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/snips.sh
vadimeisenbergibm: Yes, as I wrote in https://github.com/istio/istio.io/pull/7754#discussion_r457553335

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
nschhina: I've kept patching of gateway entirely, so I reverted those changes and I'll resolve this concern now

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
```diff
@@ -398,180 +388,31 @@ to hold the configuration of the NGINX server:
...
-              number: 443
-          weight: 100
-    EOF
-    {{< /text >}}
-
-#### Deploy a container to test the NGINX deployment
-
-1.  Create Kubernetes [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to hold the client's and CA
-    certificates:
+1.  Deploy the [sleep]({{< github_tree >}}/samples/sleep) sample to test sending requests to the NGINX server:
```
frankbu, line 391: Why did we delete the sleep sample at the end of part 1, when we need to restart it again here?

I suggest deleting this step and also removing the sleep sample from the above part 1 cleanup section.

vadimeisenbergibm: This part is to deploy sleep with client certificates to test the NGINX mutual TLS server.

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/mtls_test.sh
```diff
@@ -22,77 +22,63 @@ set -u
...
-_verify_elided snip_configure_mutual_tls_origination_for_egress_traffic_3 "$snip_configure_mutual_tls_origination_for_egress_traffic_3_out"
+# Originate TLS with destination rule
+snip_configure_mutual_tls_origination_for_egress_traffic_3
+
+_wait_for_istio destinationrule istio-system originate-mtls-for-nginx
+
+# Verify that mTLS connection is set up properly
+_verify_contains snip_configure_mutual_tls_origination_for_egress_traffic_4 "Welcome to nginx!"
 
 #TODO: verify request is actually being routed through egress gateway
```
frankbu, line 77: Why is this a TODO? Doesn't a check similar to the one in `tls_tests.sh` work here?

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
frankbu: I'm OK with shortening this doc because it is particularly long. If the ngnix server isn't working, the user will find out in the next step. It wouldn't be a bad idea to have another security task that shows how to configure a sidecar to directly call an mtls server. If that task existed, and used the same nginx example, we could add a `{{< tip >}}` here that says you can test that the nginx server is working by following those referenced instructions.

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/index.md
nschhina: Now without that cert mounted sleep, this deployment is the same as used in TLS mode originally. I would remove it from cleanup and remove this deployment

frankbu: I see, the part that @nschhina deleted. See my comment, above. Even so, shouldn't we use the sleep pod without certs for the later test? Currently, that one's deleted up in the part 1 cleanup. If we want a source pod with certs, just for the nginx test, shouldn't we just create a second one for that (in a different ns)?

vadimeisenbergibm: I am fine with removing the cert-mounted sleep and the test to make the document shorter.

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/mtls_test.sh
nschhina: Added this check

PR #7765, opened 2020-07-22
Remove manifest apply info for 1.7

shamsher31:
The simpler `install` command replaces `manifest apply`.

Related:
https://github.com/istio/istio/pull/25737


shamsher31:
@frankbu I think release-1.7 branch is not setup on istio.io docs repo?

PR #7773, opened 2020-07-22
Update grab_reference_docs to also strip `latest` from url

ericvn:

Remove the `latest` from the href if it exists. 

Not that the link is a correct link in istio.api: example: `https://istio.io/latest/docs/reference/config/networking/sidecar/` 

We currently strip the _https://istio.io_ off of it in our gathering process. It makes sense we should strip the _latest_ as well if it exists. This allows correct URLs and older now incorrect URLs which don’t have _latest_ in the gathered docs.

Also the one file change from running with the new script

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/scripts/grab_reference_docs.sh
+++ b/scripts/grab_reference_docs.sh
@@ -85,8 +85,7 @@ locate_file() {
     FN=${FN%.html}
     PP=$(echo "${FNP}" | rev | cut -d'/' -f2- | rev)
     mkdir -p "${ROOTDIR}/content/en/docs${PP}/${FN}"
-    sed -e 's/href="https:\/\/istio.io/href="/g' "${FILENAME}" >"${ROOTDIR}/content/en/docs${PP}/${FN}/index.html"
-    sed -E -e 's/(href="https:\/\/istio.io.*)\.html/\1\//' -e 's/href="https:\/\/istio.io/href="/g' "${FILENAME}" >"${ROOTDIR}/content/en/docs${PP}/${FN}/index.html"
+    sed -E -e 's/(href="https:\/\/istio.io.*)\.html/\1\//' -e 's/href="https:\/\/istio.io/href="/g' -e 's/href="https:\/\/istio.io/href="/g' -e 's/href="\/latest\//href="\//g' "${FILENAME}" >"${ROOTDIR}/content/en/docs${PP}/${FN}/index.html"
 
     LEN=${#WORK_DIR}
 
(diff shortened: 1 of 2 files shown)


scripts/grab_reference_docs.sh
```diff
@@ -85,8 +85,7 @@ locate_file() {
     FN=${FN%.html}
     PP=$(echo "${FNP}" | rev | cut -d'/' -f2- | rev)
     mkdir -p "${ROOTDIR}/content/en/docs${PP}/${FN}"
-    sed -e 's/href="https:\/\/istio.io/href="/g' "${FILENAME}" >"${ROOTDIR}/content/en/docs${PP}/${FN}/index.html"
-    sed -E -e 's/(href="https:\/\/istio.io.*)\.html/\1\//' -e 's/href="https:\/\/istio.io/href="/g' "${FILENAME}" >"${ROOTDIR}/content/en/docs${PP}/${FN}/index.html"
+    sed -E -e 's/(href="https:\/\/istio.io.*)\.html/\1\//' -e 's/href="https:\/\/istio.io/href="/g' -e 's/href="https:\/\/istio.io/href="/g' -e 's/href="\/latest\//href="\//g' "${FILENAME}" >"${ROOTDIR}/content/en/docs${PP}/${FN}/index.html"
```
frankbu, line 88:
```suggestion
    sed -E -e 's/(href="https:\/\/istio.io.*)\.html/\1\//' -e 's/href="https:\/\/istio.io/href="/g' -e 's/href="\/latest\//href="\//g' "${FILENAME}" >"${ROOTDIR}/content/en/docs${PP}/${FN}/index.html"
```

ericvn: I'm fine with that change. It is removing the first of the sed lines, which was getting override anyway by the second one.

frankbu: Not sure I understand? Your sed command had two copies of the exact same `-e 's/href="https:\/\/istio.io/href="/g'`, so I just removed the 2nd one, which does nothing. The old 2-line implementation looks buggy as well.

shamsher31:
As per @frankbu https://github.com/istio/api/pull/1550#issuecomment-662998650 I will have to cherry-pick workload selector fix in 1.6 and 1.7 manually.

@ericvn Based on that I think this also needs to be cherry-picked for 1.6 and 1.7 ?

PR #7774, opened 2020-07-23
Use full version for charts download URL

shamsher31:
Derived from https://github.com/istio/istio/pull/25783 to avoid updating these URL's with every major release.

### DIFF
--- a/content/en/docs/setup/getting-started/index.md
+++ b/content/en/docs/setup/getting-started/index.md
@@ -43,8 +43,8 @@ Follow these steps to get started with Istio:
     {{< tip >}}
     The command above downloads the latest release (numerically) of Istio.
     To download a specific version, you can add a variable on the command line.
-    For example to download Istio 1.4.3, you would run
-      `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.4.3 sh -`
+    For example to download Istio 1.6.3, you would run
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/getting-started/index.md
```diff
@@ -43,8 +43,8 @@ Follow these steps to get started with Istio:
     {{< tip >}}
     The command above downloads the latest release (numerically) of Istio.
     To download a specific version, you can add a variable on the command line.
-    For example to download Istio 1.4.3, you would run
-      `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.4.3 sh -`
+    For example to download Istio 1.6.3, you would run
```
ericvn, line 46: Why not use _istio_full_version_ here as well?

shamsher31: Because the instructions say

> To download a specific version, you can add a variable on the command line.

`istio_full_version` adds the latest release version. and as per the instruction, I think it means to download older version?

frankbu: It's probably fine to not update this at all, since it's just an example of some "other" release, so 1.4.3 is as good as any for the example.

shamsher31: Done

PR #7778, opened 2020-07-23
Tracing by default in the demo profile is 1%, not 100%

jgreens:
Noticed an inconsistency in the docs. Please close this PR if I am incorrect


douglas-reid:
That certainly is not the intention: https://github.com/istio/istio/blob/master/manifests/profiles/demo.yaml#L51-L52

frankbu:
Closing because this is not the intent. If it isn't working as documented, need to open an istio/istio issue.

PR #7781, opened 2020-07-23
Automated test case for "istioctl analyze"

esnible:
Automates https://preliminary.istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze testing

@frankbu This is my first automated test.  Please be extra strict so I learn how to do it properly.

### DIFF
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -91,12 +91,10 @@ kind: VirtualService
 ...
 status:
   validationMessages:
-...
   - code: IST0101
     documentation_url: https://istio.io/docs/reference/config/analysis/IST0101?ref=status-controller
     level: Error
     message: 'Referenced gateway not found: "bogus-gateway"'
-...
 {{< /text >}}
 
 `enableAnalysis` runs in the background, and will keep the status field of a resource up to date with its current validation status. Note that this isn't a replacement for `istioctl analyze`:
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -25,14 +25,18 @@ And that’s it! It’ll give you any recommendations that apply.
 For example, if you forgot to enable Istio injection (a very common issue), you would get the following warning:
 
 {{< text syntax=plain snip_id=analyze_all_namespace_sample_response >}}
-...
 Warn [IST0102] (Namespace default) The namespace is not enabled for Istio injection. Run 'kubectl label namespace default istio-injection=enabled' to enable it, or 'kubectl label namespace default istio-injection=disabled' to explicitly mark it as not needing injection
 {{< /text >}}
 
-Fix the issue and try again:
+Fix the issue:
 
 {{< text syntax=bash snip_id=fix_default_namespace >}}
-$ kubectl label namespace default istio-injection=enabled --overwrite
+$ kubectl label namespace default istio-injection=enabled
+{{< /text >}}
+
+Then try again:
+
+{{< text syntax=bash snip_id=try_with_fixed_namespace >}}
 $ istioctl analyze --namespace default
 ✔ No validation issues found when analyzing namespace: default.
 {{< /text >}}
@@ -84,16 +88,6 @@ For example. if you have a misconfigured gateway on your "ratings" virtual servi
 {{< text syntax=yaml snip_id=vs_yaml_with_status >}}
 apiVersion: networking.istio.io/v1beta1
 kind: VirtualService
-metadata:
-...
-  name: ratings
-  namespace: default
-...
-spec:
-  gateways:
-  - bogus-gateway
-  hosts:
-  - ratings
 ...
 status:
   validationMessages:
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -46,14 +46,14 @@ $ istioctl analyze --namespace default
 Analyze the current live cluster, simulating the effect of applying additional yaml files like `bookinfo-gateway.yaml` and `destination-rule-all.yaml` in the `samples/bookinfo/networking` directory:
 
 {{< text syntax=bash snip_id=analyze_sample_destrule >}}
-$ istioctl analyze samples/bookinfo/networking/bookinfo-gateway.yaml samples/bookinfo/networking/destination-rule-all.yaml
+$ istioctl analyze @samples/bookinfo/networking/bookinfo-gateway.yaml@ samples/bookinfo/networking/destination-rule-all.yaml
 Error [IST0101] (VirtualService bookinfo.default samples/bookinfo/networking/bookinfo-gateway.yaml:16) Referenced host not found: "productpage"
 {{< /text >}}
 
 Analyze the entire `networking` folder:
 
 {{< text syntax=bash snip_id=analyze_networking_directory >}}
-$ istioctl analyze samples/bookinfo/networking/
+$ istioctl analyze @samples/bookinfo/networking/@
 {{< /text >}}
 
 Analyze all yaml files in the `networking` folder:
@@ -88,6 +88,10 @@ For example. if you have a misconfigured gateway on your "ratings" virtual servi
 {{< text syntax=yaml snip_id=vs_yaml_with_status >}}
 apiVersion: networking.istio.io/v1beta1
 kind: VirtualService
+...
+spec:
+  gateways:
+  - bogus-gateway
 ...
 status:
   validationMessages:
@@ -143,7 +147,7 @@ $ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST
 To ignore multiple codes for a resource, separate each code with a comma:
 
 {{< text syntax=bash snip_id=annotate_for_deployment_suppression_107 >}}
-$ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002 --overwrite
+$ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002
 {{< /text >}}
 
 ## Helping us improve this tool
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -4,7 +4,7 @@ description: Shows you how to use istioctl analyze to identify potential issues 
 weight: 40
 keywords: [istioctl, debugging, kubernetes]
 owner: istio/wg-user-experience-maintainers
-test: no
+test: yes
 ---
 
 `istioctl analyze` is a diagnostic tool that can detect potential issues with your
@@ -16,7 +16,7 @@ apply changes to a cluster.
 
 You can analyze your current live Kubernetes cluster by running:
 
-{{< text bash >}}
+{{< text bash snip_id=analyze_all_namespaces >}}
 $ istioctl analyze --all-namespaces
 {{< /text >}}
 
@@ -24,7 +24,7 @@ And that’s it! It’ll give you any recommendations that apply.
 
 For example, if you forgot to enable Istio injection (a very common issue), you would get the following warning:
 
-{{< text plain >}}
+{{< text plain snip_id=analyze_all_namespaces_sample_response >}}
 Warn [IST0102](Namespace default) The namespace is not enabled for Istio injection. Run 'kubectl label namespace default istio-injection=enabled' to enable it, or 'kubectl label namespace default istio-injection=disabled' to explicitly mark it as not needing injection
 {{< /text >}}
 
@@ -32,19 +32,19 @@ Warn [IST0102](Namespace default) The namespace is not enabled for Istio injecti
 
 Analyze the current live cluster, simulating the effect of applying additional yaml files like `bookinfo-gateway.yaml` and `destination-rule-all.yaml` in the `samples/bookinfo/networking` directory:
 
-{{< text bash >}}
-$ istioctl analyze @samples/bookinfo/networking/bookinfo-gateway.yaml@ @samples/bookinfo/networking/destination-rule-all.yaml@
+{{< text bash snip_id=analyze_sample_destrule >}}
+$ istioctl analyze samples/bookinfo/networking/bookinfo-gateway.yaml samples/bookinfo/networking/destination-rule-all.yaml
 {{< /text >}}
 
 Analyze the entire `networking` folder:
 
-{{< text bash >}}
+{{< text bash snip_id=analyze_networking_directory >}}
 $ istioctl analyze samples/bookinfo/networking/
 {{< /text >}}
 
 Analyze all yaml files in the `networking` folder:
 
-{{< text bash >}}
+{{< text bash snip_id=analyze_all_networking_yaml >}}
 $ istioctl analyze samples/bookinfo/networking/*.yaml
 {{< /text >}}
 
@@ -54,7 +54,7 @@ Typically, this is used to analyze the entire set of configuration files that ar
 
 Analyze all yaml files in the `networking` folder:
 
-{{< text bash >}}
+{{< text bash snip_id=analyze_all_networking_yaml_no_kube >}}
 $ istioctl analyze --use-kube=false samples/bookinfo/networking/*.yaml
 {{< /text >}}
 
@@ -71,7 +71,7 @@ This analysis uses the same logic and error messages as when using `istioctl ana
 
 For example. if you have a misconfigured gateway on your "ratings" virtual service, running `kubectl get virtualservice ratings` would give you something like:
 
-{{< text yaml >}}
+{{< text yaml snip_id=vs_yaml_with_status >}}
 apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
 metadata:
@@ -110,7 +110,7 @@ status:
 
 You can enable this feature with:
 
-{{< text bash >}}
+{{< text bash snip_id=install_with_custom_config_analysis >}}
 $ istioctl install --set values.global.istiod.enableAnalysis=true
 {{< /text >}}
 
@@ -118,8 +118,8 @@ $ istioctl install --set values.global.istiod.enableAnalysis=true
 
 Sometimes you might find it useful to hide or ignore analyzer messages in certain cases. For example, imagine a situation where a message is emitted about a resource you don't have permissions to update:
 
-{{< text bash >}}
-$ istioctl analyze -k --all-namespaces
+{{< text bash snip_id=analyze_k_frod >}}
+$ istioctl analyze -k --namespace frod
 Warn [IST0102] (Namespace frod) The namespace is not enabled for Istio injection. Run 'kubectl label namespace frod istio-injection=enabled' to enable it, or 'kubectl label namespace frod istio-injection=disabled' to explicitly mark it as not needing injection
 Error: Analyzers found issues.
 See https://istio.io/docs/reference/config/analysis for more information about causes and resolutions.
@@ -127,14 +127,14 @@ See https://istio.io/docs/reference/config/analysis for more information about c
 
 Because you don't have permissions to update the namespace, you cannot resolve the message by annotating the namespace. Instead, you can direct `istioctl analyze` to suppress the above message on the resource:
 
-{{< text bash >}}
-$ istioctl analyze -k --all-namespaces --suppress "IST0102=Namespace frod"
-✔ No validation issues found.
+{{< text bash snip_id=analyze_suppress0102 >}}
+$ istioctl analyze -k --namespace frod --suppress "IST0102=Namespace frod"
+✔ No validation issues found when analyzing namespace: frod.
 {{< /text >}}
 
 The syntax used for suppression is the same syntax used throughout `istioctl` when referring to resources: `<kind> <name>.<namespace>`, or just `<kind> <name>` for cluster-scoped resources like `Namespace`. If you want to suppress multiple objects, you can either repeat the `--suppress` argument or use wildcards:
 
-{{< text bash >}}
+{{< text bash snip_id=analyze_suppress_frod_0107_baz >}}
 $ # Suppress code IST0102 on namespace frod and IST0107 on all pods in namespace baz
 $ istioctl analyze -k --all-namespaces --suppress "IST0102=Namespace frod" --suppress "IST0107=Pod *.baz"
 {{< /text >}}
@@ -143,14 +143,14 @@ $ istioctl analyze -k --all-namespaces --suppress "IST0102=Namespace frod" --sup
 
 You can also ignore specific analyzer messages using an annotation on the resource. For example, to ignore code IST0107 (`MisplacedAnnotation`) on resource `deployment/my-deployment`:
 
-{{< text bash >}}
+{{< text bash snip_id=annotate_for_deployment_suppression >}}
 $ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107
 {{< /text >}}
 
 To ignore multiple codes for a resource, separate each code with a comma:
 
-{{< text bash >}}
-$ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002
+{{< text bash snip_id=annotate_for_deployment_suppression_107 >}}
+$ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002 --overwrite
 {{< /text >}}
 
 ## Helping us improve this tool
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -37,7 +37,6 @@ $ istioctl analyze --namespace default
 ✔ No validation issues found when analyzing namespace: default.
 {{< /text >}}
 
-
 ## Analyzing live clusters, local files, or both
 
 Analyze the current live cluster, simulating the effect of applying additional yaml files like `bookinfo-gateway.yaml` and `destination-rule-all.yaml` in the `samples/bookinfo/networking` directory:
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -16,7 +16,7 @@ apply changes to a cluster.
 
 You can analyze your current live Kubernetes cluster by running:
 
-{{< text bash snip_id=analyze_all_namespaces >}}
+{{< text syntax=bash snip_id=analyze_all_namespaces >}}
 $ istioctl analyze --all-namespaces
 {{< /text >}}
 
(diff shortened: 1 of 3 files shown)


content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,94 @@
...
+set -u  # Unset is an error
+# There is no need to echo, output appears in ops_diagnostic-tools_istioctl-analyze_test_debug.txt
+set -o pipefail
+
+# Use 'at' setup profile=default to install default profile?
+# This script doesn't need a control plane, but we need base CRDs
+# @setup profile=none
+
+# This command is allowed to fail
+snip_analyze_all_namespaces || true
```
frankbu, line 28:
```suggestion
# This script doesn't need a control plane initially and will install Istio when needed
# @setup profile=none

# Test for forgot to enable Istio injection warning
kubectl label namespace default istio-injection-
_verify_contains snip_analyze_all_namespaces_sample_response "$snip_analyze_all_namespaces_sample_response"
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,94 @@
...
+
+# The following step is 'usually' done by cleanup.  Do it here, just in case
+kubectl delete vs ratings || true
+
+snip_analyze_sample_destrule
```
frankbu, line 33:
```suggestion
_verify_contains snip_analyze_sample_destrule "No validation issues found"
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,94 @@
...
+
+# TODO Check the output on failure for 'Referenced host+subset in destinationrule not found: "details+v2"' 
+_verify_failure snip_analyze_networking_directory
```
frankbu, line 36: Is this really expected to have issues?
```suggestion
_verify_contains snip_analyze_networking_directory 'Referenced host+subset in destinationrule not found: "details+v2"'
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,94 @@
...
+
+snip_analyze_all_networking_yaml
+
+snip_analyze_all_networking_yaml_no_kube
```
frankbu, line 40: It might be a good idea to put something in the cluster that shows the difference between these two commands, one has an error/warning, while the self contained one doesn't.
```suggestion
# TODO: put some config in the cluster to show a different output between the following two commands.

_verify_contains snip_analyze_all_networking_yaml "No validation issues found"

_verify_contains snip_analyze_all_networking_yaml_no_kube  "No validation issues found"
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,94 @@
...
+
+istioctl analyze --help
```
frankbu, line 42: Add check for some kind of expected output, using _verify_contains or _verify_lines, for example.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,94 @@
...
+
+# TODO Is it safe to install testing Istio with custom config here?
+# TODO Uncomment?
+# snip_install_with_custom_config_analysis
```
frankbu, line 46:
```suggestion
snip_install_with_custom_config_analysis
_wait_for_deployment istio-system istiod
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/snips.sh
```diff
@@ -0,0 +1,111 @@
...
+  http:
+  - route:
+    - destination:
+        host: ratings
+        subset: v1
+status:
+  validationMessages:
+  - code: IST0101
+    level: Error
+    message: 'Referenced gateway not found: "bogus-gateway"'
```
frankbu, line 75: A common approach to make this testable, and also more readable in the doc, is to change the "noise" to "...". You need to change it in the .md file and then regen the snip which will then look something like this:

```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
...
spec:
  gateways:
  - bogus-gateway
  hosts:
  - ratings
  http:
  - route:
    - destination:
        host: ratings
        subset: v1
status:
  validationMessages:
  - code: IST0101
    level: Error
    message: 'Referenced gateway not found: "bogus-gateway"'
```

Then, in the test you can do something like this:

```
get_ratings_virtual_service() {
kubectl get vs ratings -o yaml
}
_verify_elided get_ratings_virtual_service "$snip_vs_yaml_with_status"
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,94 @@
...
+snip_annotate_for_deployment_suppression
+
+kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress-
+snip_annotate_for_deployment_suppression_107
+
+# @cleanup
+set +e # ignore cleanup errors
+snip_cleanup_1
+kubectl delete ns frod || true
+kubectl delete deployment my-deployment || true
```
frankbu, line 94:
```suggestion
kubectl delete ns frod
kubectl delete deployment my-deployment
kubectl delete ns istio-system
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
esnible: @frankbu What I suspect is happening is that the analyzer sees _samples/bookinfo/networking/destination-rule-all.yaml_ but then sees _samples/bookinfo/networking/fault-injection-details-v1.yaml_.  Under ordinary circumstances a user would have one or the other.  The problem is that we may check the VirtualService, which refers to subset v2, against a Destination Rule, which only labels subset v1.  Suggestion?

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
esnible: Your comment makes sense from a testing standpoint.  From the standpoint of documentation, the point is to show that the `--use-kube` exists, not that it works.
(I could deploy _httpbin_, and show that the analysis of _samples/httpbin/httpbin-gateway.yaml_ is different if we look at httpbin on the cluster.  I am not saying the test won't work, just that it should be tested in integration rather than doc tests.)

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -16,35 +16,46 @@ apply changes to a cluster.
...
 
 And that’s it! It’ll give you any recommendations that apply.
 
 For example, if you forgot to enable Istio injection (a very common issue), you would get the following warning:
 
-{{< text plain >}}
-Warn [IST0102](Namespace default) The namespace is not enabled for Istio injection. Run 'kubectl label namespace default istio-injection=enabled' to enable it, or 'kubectl label namespace default istio-injection=disabled' to explicitly mark it as not needing injection
+{{< text syntax=plain snip_id=analyze_all_namespace_sample_response >}}
+...
+Warn [IST0102] (Namespace default) The namespace is not enabled for Istio injection. Run 'kubectl label namespace default istio-injection=enabled' to enable it, or 'kubectl label namespace default istio-injection=disabled' to explicitly mark it as not needing injection
```
frankbu, line 29: Using elision here has no advantage, because the text just says you'll see this message. Elision makes sense when you have a text block that shows the command and expected output, but you want to elide some of it.
```suggestion
Warn [IST0102] (Namespace default) The namespace is not enabled for Istio injection. Run 'kubectl label namespace default istio-injection=enabled' to enable it, or 'kubectl label namespace default istio-injection=disabled' to explicitly mark it as not needing injection
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,109 @@
...
+set -e  # Exit on failure
+set -u  # Unset is an error
+# There is no need to echo, output appears in ops_diagnostic-tools_istioctl-analyze_test_debug.txt
+set -o pipefail
+
+# This script doesn't need a control plane initially and will install Istio when needed
+# @setup profile=none
+
+echo '*** istioctl-analyze step 1 ***'
+_verify_elided snip_analyze_all_namespaces "$snip_analyze_all_namespace_sample_response"
```
frankbu, line 27: This check is currently failing because of interference from an earlier test (which we need to fix), but a good general rule is to explicitly make sure the condition you are testing for is true, removing the injection label in this case. Also, you should use `_verify_contains` here, instead of `_verify_elided`, see my related cmment in the md file.
```suggestion
kubectl label namespace default istio-injection-
_verify_contains snip_analyze_all_namespaces "$snip_analyze_all_namespace_sample_response" 
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -16,35 +16,46 @@ apply changes to a cluster.
...
 {{< /text >}}
 
+Fix the issue and try again:
+
+{{< text syntax=bash snip_id=fix_default_namespace >}}
+$ kubectl label namespace default istio-injection=enabled --overwrite
+$ istioctl analyze --namespace default
+✔ No validation issues found when analyzing namespace: default.
+{{< /text >}}
```
frankbu: The --overwrite is an example of where somebody reading the doc would say "why do I need to do that?". It doesn't matter if it fails or not if running it on the command line, it's only a problem for the test. So, rather than complicating the doc to accommodate the test, it's better to split this into two text blocks.
```suggestion
Fix the issue:

{{< text syntax=bash snip_id=fix_default_namespace >}}
$ kubectl label namespace default istio-injection=enabled --overwrite
{{< /text >}}

Then try again:

{{< text syntax=bash snip_id=try_with_fixed_namespace >}}
$ istioctl analyze --namespace default
✔ No validation issues found when analyzing namespace: default.
{{< /text >}}
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,109 @@
...
+set -o pipefail
+
+# This script doesn't need a control plane initially and will install Istio when needed
+# @setup profile=none
+
+echo '*** istioctl-analyze step 1 ***'
+_verify_elided snip_analyze_all_namespaces "$snip_analyze_all_namespace_sample_response" 
+
+echo '*** istioctl-analyze step 2 ***'
+_verify_contains snip_fix_default_namespace "$snip_fix_default_namespace_out"
```
frankbu, line 34: See comment in md file.
```suggestion
snip_fix_default_namespace
_verify_same try_with_fixed_namespace "$try_with_fixed_namespace_out"
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -16,35 +16,45 @@ apply changes to a cluster.
...
+Warn [IST0102] (Namespace default) The namespace is not enabled for Istio injection. Run 'kubectl label namespace default istio-injection=enabled' to enable it, or 'kubectl label namespace default istio-injection=disabled' to explicitly mark it as not needing injection
+{{< /text >}}
+
+Fix the issue and try again:
+
+{{< text syntax=bash snip_id=fix_default_namespace >}}
+$ kubectl label namespace default istio-injection=enabled --overwrite
+$ istioctl analyze --namespace default
+✔ No validation issues found when analyzing namespace: default.
 {{< /text >}}
```
frankbu, line 38: The --overwrite is an example of where somebody reading the doc would say "why do I need to do that?". It doesn't matter if it fails or not if running it on the command line, it's only a problem for the test. So, rather than complicating the doc to accommodate the test, it's better to split this into two text blocks.
```suggestion
Fix the issue:

{{< text syntax=bash snip_id=fix_default_namespace >}}
$ kubectl label namespace default istio-injection=enabled
{{< /text >}}

Then try again:

{{< text syntax=bash snip_id=try_with_fixed_namespace >}}
$ istioctl analyze --namespace default
✔ No validation issues found when analyzing namespace: default.
{{< /text >}}
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -71,35 +81,28 @@ This analysis uses the same logic and error messages as when using `istioctl ana
...
   gateways:
   - bogus-gateway
   hosts:
   - ratings
-  http:
-  - route:
-    - destination:
-        host: ratings
-        subset: v1
+...
```
frankbu, line 97: Do you really need this many `...` in the example? It looks kind of ugly in the doc. Can't you cut it down to a couple and still get the point across?

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
esnible: fixed

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
esnible: @frankbu fixed

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
esnible: But the rule output of `analyze all-namespaces` **MAY** include other messages, such as

```
Warn [IST0102] (Namespace ibm-system) The namespace is not enabled for Istio injection. Run 'kubectl label namespace ibm-system istio-injection=enabled' to enable it, or 'kubectl label namespace ibm-system istio-injection=disabled' to explicitly mark it as not needing injection
```

There is no way to predict what `--all-namespaces` will show for a given user.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
esnible: Maybe I don't understand `...` and elided.  For the user I thought it means "unimportant stuff" and for the recognizer I thought it means "*" (wildcard).  So for example I didn't want to show that there might be a second reported problem, so I used ... there.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
frankbu: That true, but doesn't matter. The text in the document isn't saying anything about whether or not there will be other output from the command, it's just saying that it there will be a warning like this if injection isn't enabled. For the test program, It's a perfect match for the `_verify_contains` function, which is just checking that the specified message is in the output. The elision is confusing, because the doc at this point doesn't (and shouldn't) mention anything about other possible errors that the command may find. To be perfectly safe for testing in the general case, you would actually need to add `...` both before and after the message of interest, but that's just unnecessary noise in the doc.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
frankbu: OK, that's fair enough.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
frankbu: Maybe this means that the bookinfo networking folder isn't a good example, because it has a mixed bag of configs that aren't intended to work together. Maybe there's another folder (e.g., httpbin sample maybe?) that would make a more realistic example of analyzing a folder?

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
frankbu: You're right that `...` is a wildcard for testing and that in docs it's used to elide "unimportant stuff", but what you are doing is complicating the doc by trying to make the example a fully correct regular expression, which isn't necessary or helpful. In docs, you really just want to show a simple specific example, and only elide things that are noise (e.g., metadata or run specific values like date/time). The "unimportant stuff" you should elide is stuff in the original example that isn't needed to support what the document is trying to explain with the example. The original example didn't need to indicate the possible presence of other unrelated stuff that isn't relevant to the discussion at hand, so the elided one shouldn't either. Elision in docs is used to simplify, not to produce a generally correct regular expression.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -16,35 +16,49 @@ apply changes to a cluster.
...
 {{< /text >}}
 
 ## Analyzing live clusters, local files, or both
 
 Analyze the current live cluster, simulating the effect of applying additional yaml files like `bookinfo-gateway.yaml` and `destination-rule-all.yaml` in the `samples/bookinfo/networking` directory:
 
-{{< text bash >}}
-$ istioctl analyze @samples/bookinfo/networking/bookinfo-gateway.yaml@ @samples/bookinfo/networking/destination-rule-all.yaml@
+{{< text syntax=bash snip_id=analyze_sample_destrule >}}
+$ istioctl analyze samples/bookinfo/networking/bookinfo-gateway.yaml samples/bookinfo/networking/destination-rule-all.yaml
```
frankbu, line 49: YAML files were clickable links.

```suggestion
$ istioctl analyze @samples/bookinfo/networking/bookinfo-gateway.yaml@ @samples/bookinfo/networking/destination-rule-all.yaml@
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,115 @@
...
+istioctl analyze --help
+
+echo '*** istioctl-analyze step 8 ***'
+snip_install_with_custom_config_analysis
+_wait_for_deployment istio-system istiod
+
+echo '*** istioctl-analyze step 9 ***'
+set +e  # Don't exit on failure
+kubectl apply -f - <<EOF
+apiVersion: networking.istio.io/v1alpha3
```
frankbu, line 61:
```suggestion
apiVersion: networking.istio.io/v1beta1
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,115 @@
...
+
+echo '*** istioctl-analyze step 10 ***'
+get_ratings_virtual_service() {
+kubectl get vs ratings -o yaml
+}
+_verify_elided get_ratings_virtual_service "$snip_vs_yaml_with_status"
+
+echo '*** istioctl-analyze step 11 ***'
+kubectl create ns frod
+_verify_failure analyze_k_frod
```
frankbu, line 86: Instead of just testing for failure, you should confirm that the output shown in the doc is correct, a more thorough "test of the doc".

```suggestion
_verify_same analyze_k_frod "$analyze_k_frod_out"
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
esnible: @frankbu They were!  I would like to bring that back.  I went through the _.md_ files and couldn't find any instances of `@samples` in `< text snip_id=... >` blocks.  Can you offer style guidance for this situation?

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,115 @@
...
+kubectl create deployment my-deployment --image=docker.io/kennethreitz/httpbin
+snip_annotate_for_deployment_suppression
+
+echo '*** istioctl-analyze step 15 ***'
+kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress-
+snip_annotate_for_deployment_suppression_107
+
+# @cleanup
+# Don't remove the injection label on default, as that was the original state of the test harness
+# kubectl label namespace default istio-injection-
```
frankbu, line 108: I think that's a bug in a previous test. The @setup profile=none should start clean, but until we have the between-test cleanup checking in the framework there's always the possibility of cleanup bugs in other tests :(

```suggestion
kubectl label namespace default istio-injection-
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,115 @@
...
+
+echo '*** istioctl-analyze step 14 ***'
+kubectl create deployment my-deployment --image=docker.io/kennethreitz/httpbin
+snip_annotate_for_deployment_suppression
+
+echo '*** istioctl-analyze step 15 ***'
+kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress-
+snip_annotate_for_deployment_suppression_107
+
+# @cleanup
```
frankbu, line 106: The cleanup commands needs to ignore errors, since cleanup may be called early (before some of the things are even created) if the test fails. If you look at other tests, you'll notice that every one has `set +e` as the first line. Maybe the framework should inject that automatically (like it does some of the include files?)

```suggestion
# @cleanup
set +e # ignore cleanup errors
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,115 @@
...
+kubectl get vs ratings -o yaml
+}
+_verify_elided get_ratings_virtual_service "$snip_vs_yaml_with_status"
+
+echo '*** istioctl-analyze step 11 ***'
+kubectl create ns frod
+_verify_failure analyze_k_frod
+# TODO Replace the above with _verify_failure_same when it is available
+# _verify_same analyze_k_frod "$snip_analyze_k_frod_out"
+
```
frankbu, line 89:
```suggestion

```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -143,14 +140,14 @@ $ istioctl analyze -k --all-namespaces --suppress "IST0102=Namespace frod" --sup
...
+{{< text syntax=bash snip_id=annotate_for_deployment_suppression >}}
 $ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107
 {{< /text >}}
 
 To ignore multiple codes for a resource, separate each code with a comma:
 
-{{< text bash >}}
-$ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002
+{{< text syntax=bash snip_id=annotate_for_deployment_suppression_107 >}}
+$ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002 --overwrite
```
frankbu, line 150: Instead of adding `--overwrite` in the doc, you should just remove the IST0107 suppress in the test.sh before calling this snip function. Since the doc describes this snip and the previous one as either or (as opposed to step by step run this, then run that), the reader might be confused why `--overwrite` is needed in the second snip but not the first. In general, you should try not to change the commands in the doc unless your fixing a bug or it makes something more clear for the reader.
```suggestion
$ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,115 @@
...
+echo '*** istioctl-analyze step 12 ***'
+_verify_same snip_analyze_suppress0102 "$snip_analyze_suppress0102_out"
+
+echo '*** istioctl-analyze step 13 ***'
+# TODO This will return errors on non-Istio namespaces, e.g. ibm-cert-store and kube-node-lease
+# It is hard to know what to supply to fix that problem for testing.
```
frankbu, line 95: You could create a broken deployment in ns baz and use `_verify_contains` to check for something expected in the ouput. Then the next 2 snips could use `_verify_not_contains` to confirm the same string is not in the output any more. The request routing task does something similar in its test: https://github.com/istio/istio.io/blob/54113c6864f28d4eaee6b58154ecb85f7faa1789/content/en/docs/tasks/traffic-management/request-routing/test.sh#L72

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
frankbu: Most docs containing commands that use yaml files in the github samples dir wrap them in `@` chars so that the file is clickable in the UI. That includes this command (Look at line 36 in the diff, above, which has the `@` chars that you removed). https://preliminary.istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze/#analyzing-live-clusters-local-files-or-both. I'm just saying leave it as it was, don't change this in the doc.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
esnible: @frankbu I'd like to keep the user commands the same and use `_verify_not_contains`.  I tried it, and it works, but I suspect `not_contains` will suffer because changes that affect CLI output will not be detected.

frankbu: Yes, the safest way to use `_verify_not_contains` is in conjunction with `_verify_contains`, i.e., check that some output is there, and then after changing something, check that it's not. Another useful `_verify_` function is `_verify_lines` which can check that 0 or more lines appear in the output AND 0 or more lines do not.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -0,0 +1,110 @@
...
+
+echo '*** istioctl-analyze step 11 ***'
+kubectl create ns frod
+_verify_contains snip_analyze_k_frod "$snip_analyze_k_frod_out"
+
+echo '*** istioctl-analyze step 12 ***'
+_verify_same snip_analyze_suppress0102 "$snip_analyze_suppress0102_out"
+
+echo '*** istioctl-analyze step 13 ***'
+_verify_lines snip_analyze_suppress_frod_0107_baz "-Warn [IST0102] (Namespace frod) The namespace is not enabled for Istio injection. Run 'kubectl label namespace frod istio-injection=enabled' to enable it, or 'kubectl label namespace frod istio-injection=disabled' to explicitly mark it as not needing injection"
```
frankbu, line 91: (from doc) For each line in `expected` with the prefix "- " there must be no line in the output containing the following string.
```suggestion
_verify_lines snip_analyze_suppress_frod_0107_baz "- Warn [IST0102] (Namespace frod) The namespace is not enabled for Istio injection. Run 'kubectl label namespace frod istio-injection=enabled' to enable it, or 'kubectl label namespace frod istio-injection=disabled' to explicitly mark it as not needing injection"
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
frankbu: I would leave the invalid specification in the doc, so it's clear what the error message is referring to.
```
kind: VirtualService
...
spec:
  gateways:
  - bogus-gateway
  hosts:
  - ratings
  http:
  - route:
    - destination:
        host: ratings
        subset: v1
status:
  validationMessages:
...
  - code: IST0101
    documentation_url: https://istio.io/docs/reference/config/analysis/IST0101?ref=status-controller
    level: Error
    message: 'Referenced gateway not found: "bogus-gateway"'
...
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -71,35 +85,18 @@ This analysis uses the same logic and error messages as when using `istioctl ana
...
-  - ratings
-  http:
-  - route:
-    - destination:
-        host: ratings
-        subset: v1
+...
 status:
   validationMessages:
+...
```
frankbu, line 94: Is this message coming out in the middle between other messages when you run the test? What other messages? If you can configure the test environment to only produce the one error, or to make it first or last, you could simplify the doc by removing one or both `...` around the messages.

PR #7787, opened 2020-07-24
Add new multicluster deployment guide for single networks

rcaballeromx:
The new deployment guide is a modular information experience that
is extensible, consistent, and clear.
This guide covers how to deploy a multicluster mesh on a single network,
and redirects to the old guides for multi-network scenarios.

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/boilerplates/mc-env-var.md
```diff
@@ -0,0 +1,54 @@
...
+         <td>The name of the first network. The guide uses <code>network1</code>.</td>
+      </tr>
+      <tr>
+         <td>MESH</td>
+         <td>The common mesh ID to use for all clusters in the mesh, for example
+         sample-mesh.</td>
+      </tr>
+      <tr>
+         <td>DISCOVERY_ADDRESS</td>
+         <td>Stores the IP address of the Istio ingress gateway of your first cluster,
```
howardjohn, line 50: How do I get this? This is the install guide but it assumes its already installed? The doc never says "Step 1: install a single cluster" from what I can see

content/en/boilerplates/mc-env-var.md
```diff
@@ -0,0 +1,54 @@
...
+         <td>CTX_1</td>
+         <td>The path to the Kubernetes context file for the first cluster.</td>
+      </tr>
+      <tr>
+         <td>CTX_2</td>
+         <td>The path to the Kubernetes context file for the second cluster.</td>
+      </tr>
+      <tr>
+         <td>NETWORK_1</td>
+         <td>The name of the first network. The guide uses <code>network1</code>.</td>
```
howardjohn, line 41: This and mesh id I expect to be very confusing. I would not know what to set here if I was not familiar already

content/en/boilerplates/mc-env-var.md
rcaballeromx: @howardjohn This is not a document for beginners and the text expressly asks readers to be familiar with single cluster deployments already. Additionally, the content you see here is part of a boilerplate that can be reused, that's why you don't see any further guidance.

content/en/boilerplates/mc-env-var.md
rcaballeromx: Familiarity with single cluster deployments is required for this content.

content/en/boilerplates/mc-env-var.md
howardjohn: Familiarity with single cluster != familiarity with knowing what a "network" is. I am fairly sure a large portion of Istio users would not understand what to set here.

And the doc says to *understand* single cluster, it doesn't say to deploy one.

frankbu:
@rcaballeromx I'm assuming this is post 1.7, so we should mark it as WIP do-not-merge, so we don't accidentally merge it before we cut the 1.7 branch.

content/en/docs/setup/install/multicluster/single-network/_index.md
```diff
@@ -0,0 +1,30 @@
...
+for more information.
+
+To start a multicluster deployment, go to the [initial configuration section.](/docs/setup/install/multicluster/single-network/initial-configuration)
+
+After completing the initial configuration, complete the instructions in the
+following sections as many times as needed to add all the
+{{< gloss "cluster" >}}clusters{{< /gloss >}} that you need.
+
+- To add a cluster with a {{< gloss >}}control plane{{< /gloss >}} that engages
+    in cross-cluster load balancing, go to the [add a primary cluster section.](/docs/setup/install/multicluster/single-network/primary)
```
craigbox, line 23: Instantly confusing to me: there are more than one primary cluster? Didn't I start this with a single cluster already? Am I load balancing my control plane or my workloads?  Why does it matter?

content/en/docs/setup/install/multicluster/single-network/_index.md
```diff
@@ -0,0 +1,30 @@
...
+
+- To add a cluster that connects to a control plane outside the cluster, go
+    to the [add a remote cluster section.](/docs/setup/install/multicluster/single-network/remote)
```
craigbox, line 26: It sounds like a sentence like 

"A cluster can either run a control plane replica, or connect to a control plane running elsewhere"

before these, would be useful

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+at least one primary cluster to complete the [different multicluster deployments.](/docs/ops/deployment/deployment-models/#multiple-clusters)
+
+Each cluster in a multicluster deployment needs a unique identifier. Using
+`CLUSTER_1` as a unique identifier for the first cluster, the following
+diagram shows a service mesh spanning a single primary cluster.
+
+{{< image width="75%"
+    link="init-conf.svg"
+    caption="A service mesh with a single primary cluster"
+    >}}
```
craigbox, line 24: The bottom box lists "Proxy" and a box with the Envoy logo as if they are different things.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
+---
+title: Initial Configuration
+description: Describes how to complete the initial configuration for a multicluster mesh on a flat network.
```
craigbox, line 3: This description isn't included on this page (good!)

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+keywords: [multicluster,first-cluster]
+owner: istio/wg-environments-maintainers
+test: n/a
+---
+
+To build a {{< gloss >}}multicluster{{< /gloss >}} deployment, configure a
+single cluster {{< gloss >}}service mesh{{< /gloss >}} as your starting point.
+
+The single cluster mesh you deploy in this section, serves as a model for the
+configuration of the first {{< gloss >}}primary cluster{{< /gloss >}}. You need
```
craigbox, line 14: Why doesn't it just serve as the first primary cluster?

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+    >}}
+
+To set up a cluster as the primary cluster of a multicluster service
+mesh, complete the following configuration steps:
+
+1. [Configure trust for the cluster](#configure-trust)
+1. [Deploy Istio to the cluster](#deploy-istio)
+
+Complete the [Before you begin instructions](/docs/setup/install/multicluster/#before-you-begin)
+before you continue.
```
craigbox, line 33: This order is odd.  You say "here are two things that are on this page, which you can see the first one of below the fold, but before any of that - go back to a page you've already been on, but listed as the third step".

(I wonder if there could be a much easier flow through all these pages.)

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+            <td><code>meshExpansion</code></td>
+            <td>Recommended: Exposes the Istio control plane through the ingress gateway
+            of the mesh. Enable this option to connect remote clusters to a control
+            plane.</td>
+            <td><code>enabled: true</code></td>
+        </tr>
+    </tbody>
+</table>
+
+Using the previously [set environment variables,](/docs/setup/install/multicluster/#env-var)
```
craigbox, line 148: "previously set" all in the link text?

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+            - fromRegistry: Kubernetes
+            gateways:
+            - registry_service_name: istio-ingressgateway.istio-system.svc.cluster.local
+                port: 443
+        # Expose the control plane through istio-ingressgateway.
+        meshExpansion:
+            enabled: true
+    EOF
+    {{< /text >}}
+
```
craigbox, line 182: Random thought: I hope this gets _substantially_ easier

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+current [deployment model](/docs/ops/deployment/deployment-models/).
+
+Add primary clusters to your deployment to improve the availability of your
+mesh. To keep multiple {{< gloss "control plane" >}}control planes{{< /gloss >}}
+in sync, apply your configuration to each cluster. Production systems employ
+continuous integration and continuous deployment (CI/CD) pipelines to apply
+configurations since they typically have at least one control plane per region.
+Administrators use these pipelines to manage configuration rollouts. Apply new
+configurations to one primary cluster at a time to help minimize
+troubleshooting.
```
craigbox, line 27: This sounds like a Markov chain generated it. 😁

What I think you're trying to get across is:

- a primary cluster is one with a control plane replica deployed in it
- all multi-cluster meshes need at least one primary cluster
- you can choose to make subsequent clusters primary clusters or remote clusters
- multiple control planes can be in different regions
- you should apply config one at a time (pref with a CI/CD system)

I'm a little surprised there is no global config, and you might want to explain why not?

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+
+The following diagram shows a multicluster deployment with two primary
+clusters:
+
+{{< image width="75%"
+    link="primary.svg"
+    caption="A multicluster deployment with two primary clusters"
+    >}}
```
craigbox, line 35: thought: you should use the Kubernetes logo for "API server", if that's what it is

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+Complete the [initial configuration instructions](/docs/setup/install/multicluster/single-network/initial-configuration)
+before you continue.
+
+## Configure Trust
+
+{{< boilerplate trust-config >}}
+
+## Deploy Istio
+
+Next, deploy a full Istio control plane on `Cluster_2`. The new control plane
```
craigbox, line 46: You just said that at the end of the last section.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+            <td><code>clusterName</code>
+            </td>
+            <td>Specifies a human-readable cluster name.</td>
+            <td><code>${CLUSTER_2}</code></td>
+        </tr>
+        <tr>
+            <td><code>network</code></td>
+            <td>Specifies a network ID as an arbitrary string. All clusters in your mesh
+            must be on the same network, and have the same network ID.</td>
+            <td><code>${NETWORK_1}</code></td>
```
craigbox, line 68: If there's only one network, how about ${NETWORK} throughout, without the _1?

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+    istiod-f756bbfc4-thkmk                  1/1     Running   0          136m
+    prometheus-b54c6f66b-q8hbt              2/2     Running   0          136m
+    {{< /text >}}
+
+1. After the status of all pods is `Running`, you can continue configuring
+   your deployment.
+
+**Congratulations!**
+
+You successfully added a control plane to your primary
```
craigbox, line 142: to "a" primary cluster; your second primary cluster if following along. (This is why the name feels odd to me)

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+cluster.
+
+Next, configure endpoint discovery to support cross-cluster load balancing.
```
craigbox, line 145: Thought: you talk about cross-cluster load balancing a lot.  Has it been defined anywhere?

You may also not want to do load balancing, but instead just have services in cluster A accessible from cluster B.  (This would be common in a regional situation with different deployments)

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Configure endpoint discovery
+
+To enable cross-cluster load balancing in your mesh, configure endpoint
+discovery. This feature requires that clusters share secrets between them. If
+the shared secrets provide the [needed trust](#heading=h.ww2golqkax84), each
```
craigbox, line 151: that's a Google Doc heading tag, not an istio.io page tag

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+
+You successfully added a control plane to your primary
+cluster.
+
+Next, configure endpoint discovery to support cross-cluster load balancing.
+
+## Configure endpoint discovery
+
+To enable cross-cluster load balancing in your mesh, configure endpoint
+discovery. This feature requires that clusters share secrets between them. If
```
craigbox, line 150: What secrets and for what purpose

craigbox: And are they rotated?

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+{{< /text >}}
+
+1. Share the secret of `Cluster_2` with `Cluster_1`:
+
+{{< text bash >}}
+$ istioctl x create-remote-secret \
+  --context=${CTX_2} \
+  --name=${CLUSTER_2} | \
+  kubectl apply -f - --context=${CTX_1}
+{{< /text >}}
```
craigbox, line 172: What do I do when I have more than 2 primary clusters?  Do I share the secrets with all of them?  Does this explode quickly?

content/en/docs/setup/install/multicluster/single-network/remote/index.md
```diff
@@ -0,0 +1,159 @@
...
+
+To build a {{< gloss >}}multicluster{{< /gloss >}} deployment, you can add a
+{{< gloss >}}remote cluster{{< /gloss >}} to your
+{{< gloss >}}service mesh{{< /gloss >}}.
+
+To give you clean instructions, this section starts right after you complete the
+initial configuration, and assumes that `Cluster_1` and `Cluster_2` reside on
+the same network. The new remote cluster, `Cluster_2` in this section, shares
+the {{< gloss >}}control plane{{< /gloss >}} of `CLUSTER_1`, and this guide
+assumes that `Cluster_1` and `Cluster_2` reside on the same network.
```
craigbox, line 18: I know this isn't a tutorial, but as it reads like one, why don't you call them CLUSTER_1, CLUSTER_2, CLUSTER_3 (or even better CLUSTER_PRIMARY, CLUSTER_SECONDARY, CLUSTER_REMOTE or such) and then you can run either of these two and have it all work?

content/en/docs/setup/install/multicluster/single-network/remote/index.md
```diff
@@ -0,0 +1,159 @@
...
+
+Production systems use this configuration when all clusters within a region
+share a common control plane. Any configuration with one control plane per zone
```
craigbox, line 21: unfinished thought.

(and why does it have to be within a region?)

content/en/docs/setup/install/multicluster/single-network/remote/index.md
```diff
@@ -0,0 +1,159 @@
...
+## Configure Trust
+
+{{< boilerplate trust-config >}}
+
+## Deploy Istio
+
+Next, deploy Istio in `Cluster_2` with the discovery address pointing at the
+ingress gateway of `Cluster_1`. Your [initial configuration](/docs/setup/install/multicluster/single-network/initial-configuration)
+enabled mesh expansion on `Cluster_1`, and now `Cluster_2` can access the
+discovery server in `Cluster_1` via the ingress gateway.
```
craigbox, line 43: This is all on the same network, so no ingress gateway needed, no?  It's not on your diagram

content/en/docs/setup/install/multicluster/single-network/remote/index.md
```diff
@@ -0,0 +1,159 @@
...
+        <tr>
+            <td><code>meshID</code></td>
+            <td>Specifies a mesh ID as an arbitrary string. All clusters in your mesh share
+            the same mesh ID.</td>
+            <td><code>${MESH}</code></td>
+        </tr>
+        <tr>
+            <td><code>remotePilotAddress</code></td>
+            <td>IP address of the Istio ingress gateway of `Cluster_1`. </td>
+            <td>${DISCOVERY_ADDRESS}</td>
```
craigbox, line 77: needs <code>

content/en/docs/setup/install/multicluster/single-network/remote/index.md
```diff
@@ -0,0 +1,159 @@
...
+      network: ${NETWORK_1}
+      # Access the control plane discovery server via the ingress
+      remotePilotAddress: ${DISCOVERY_ADDRESS}
+EOF
+{{< /text >}}
+
+1. Install Istio on `Cluster_2` with the following command:
+
+{{< text bash >}}
+$ istioctl --context=${CTX_2} manifest apply -f \
```
craigbox, line 119: we're using "install" instead of "manifest apply" now are we not?

content/en/docs/setup/install/multicluster/single-network/remote/index.md
```diff
@@ -0,0 +1,159 @@
...
+   command:
+
+{{< text bash >}}
+$ kubectl --context=${CTX_2} -n istio-system get pod
+NAME                                    READY   STATUS    RESTARTS   AGE
+istiod-f756bbfc4-thkmk                  1/1     Running   0          136m
+prometheus-b54c6f66b-q8hbt              2/2     Running   0          136m
+{{< /text >}}
+
+1. After the status of all pods is `Running`, you can continue configuring
```
craigbox, line 133: all these list items have the number 1. in the rendered output

content/en/docs/setup/install/multicluster/single-network/remote/index.md
```diff
@@ -0,0 +1,159 @@
...
+   your deployment.
+
+## Configure endpoint discovery
+
+To enable cross-cluster load balancing in your mesh, configure endpoint
+discovery. This feature requires that clusters share secrets between them. If
+the shared secrets provide the [needed trust](#configure-trust), each
```
craigbox, line 140: The link is right here.

It seems that the "needed trust" from the CA certs isn't enough, and you still need to set a secret up so that the control plane in CLUSTER_1 can query the API server in CLUSTER_2.

content/en/docs/setup/install/multicluster/single-network/verify/index.md
```diff
@@ -0,0 +1,318 @@
...
+1. Deploy the `HelloWorld` service to `Cluster_1` with the following command:
+
+    {{< text bash >}}
+    $ kubectl create --context=${CTX_1} \
+        -f ${ISTIO}/samples/helloworld/helloworld.yaml \
+        -l app=helloworld -n sample
+    {{< /text >}}
+
+Using the previously [set environment variables,](/docs/setup/install/multicluster/#env-var)
+deploy the HelloWorld service to `Cluster_2` with the following steps:
```
craigbox, line 118: `HelloWorld` 

(or just cut this as it's all the same as the above paragraph)

content/en/docs/setup/install/multicluster/single-network/verify/index.md
```diff
@@ -0,0 +1,318 @@
...
+    {{< /text >}}
+
+1. After the status of `helloworld-v2` is `Running`, your cluster is ready
+   to deploy a load-generating service.
+
+## Deploy a load-generating service
+
+To generate the traffic needed to verify your deployment, deploy the `sleep`
+service to your clusters. The `sleep` service is a simple load-generating
+service. If you are using your own service to verify your deployment, ensure
```
craigbox, line 194: No it's not, it's a simple sleeping service.  you're (ab)using it as a pod you can use to curl things.  Why not use a curl service, or fortio or some other load generator perhaps?

content/en/docs/setup/install/multicluster/single-network/verify/index.md
```diff
@@ -0,0 +1,318 @@
...
+    Hello version: v2, instance: helloworld-v2-758dd55874-6x4t8
+    Hello version: v1, instance: helloworld-v1-86f77cd7bd-cpxhv
+    ...
+    {{< /text >}}
+
+**Congratulations!**
+
+You successfully verified the deployment of your multicluster mesh.
+
+## Clean up
```
craigbox, line 292: Is this not a production installation guide?
Why end it with "And now delete Istio?"

craigbox (commented):
PTAL at a "user first" review.

content/en/docs/setup/install/multicluster/legacy/gateways/index.md
```diff
@@ -7,11 +7,18 @@ aliases:
...
+    - /docs/setup/install/multicluster/gateways
 keywords: [kubernetes,multicluster,gateway]
 owner: istio/wg-environments-maintainers
 test: no
 ---
 
+{{< warning >}}
+This is legacy content. If you are deploying a multicluster service mesh on a
+flat network, go to the [new multicluster guide.](/docs/setup/install/multicluster/single-network)
+Only use this guide if you are deploying a mesh spanning multiple networks.
```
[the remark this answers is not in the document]
nmittler: +1. Maybe also mention that new multi-network content is coming soon.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
+---
+title: Initial Configuration
```
nmittler, line 2: Should this be called something like `Install your first primary cluster`? Initial configuration isn't clear.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
nmittler: +1. This should probably either be done at the top of the page or removed.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
nmittler: Yeah, we're working on it. We're hoping to make MeshNetworks effectively go away in the future and make everything way more dynamic. But this is the unfortunate world we're in for now. The funny thing is that I'm not sure how many people knew how bad all of this was ... actually putting it down in docs makes it glaringly obvious :).

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
nmittler: +1 that's exactly what this should be. I think that we should probably rephrase things to make sure this is more clear. We can probably drop the concept of "single cluster service mesh" altogether. Maybe we can replace the first 3 paragraphs with something like:

```
To build a service mesh, we begin by installing our first primary cluster.
```

I purposely left "multicluster" out of this, since this page will eventually be the starting point for "install", not specifically "multicluster installation".

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+
+To build a {{< gloss >}}multicluster{{< /gloss >}} deployment, you can add
+{{< gloss "primary cluster" >}}primary clusters{{< /gloss >}} to your
+{{< gloss >}}service mesh{{< /gloss >}}.
+
+To give you clean instructions, this section starts right after you complete the
+[initial configuration](/docs/setup/install/multicluster/single-network/initial-configuration),
+and assumes that `Cluster_1` and `Cluster_2` reside on the same network.
+However, you can add primary clusters to your existing mesh regardless of your
+current [deployment model](/docs/ops/deployment/deployment-models/).
```
nmittler, line 18: I think we can leave deployment model out of this paragraph. It's not really adding much and is just adding to the link/definition overload of the reader.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
+---
+title: Add a Primary Cluster
+description: Describes how to add a primary cluster to your mesh on a flat network.
+weight: 20
+keywords: [multicluster, primary-cluster, deployment]
+owner: istio/wg-environments-maintainers
+test: n/a
+---
+
+To build a {{< gloss >}}multicluster{{< /gloss >}} deployment, you can add
```
nmittler, line 10: The opening paragraph should be more direct, explaining the task this document describes.  Maybe something like:

```
This guide expands your existing service mesh to include another a primary cluster.

Before you begin, you should have already installed your first primary cluster...
```

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+{{< gloss "primary cluster" >}}primary clusters{{< /gloss >}} to your
+{{< gloss >}}service mesh{{< /gloss >}}.
+
+To give you clean instructions, this section starts right after you complete the
```
nmittler, line 14: Probably shouldn't give our opinion as to the quality of our docs :).

content/en/docs/setup/install/multicluster/single-network/primary/index.md
```diff
@@ -0,0 +1,178 @@
...
+            <td><code>meshExpansion</code></td>
+            <td>[Recommended] Exposes the Istio control plane through the ingress gateway
+            of the mesh. Enable this option to connect remote clusters to a control
+            plane.</td>
+            <td><code>enabled: true</code></td>
+        </tr>
+    </tbody>
+</table>
+
+Using the previously [set environment variables,](/docs/setup/install/multicluster/#env-var)
```
nmittler, line 86: Let's leave this out. We already mentioned above that they should have done something before landing here.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
nmittler: Having services communicate across clusters is what the intent of "cross-cluster LB" is here. Or am I missing something?

We could drop that term altogether and just say:

```
Next, configure cross-cluster endpoint discovery.
```

content/en/docs/setup/install/multicluster/single-network/primary/index.md
nmittler: These are the secrets for accessing the remote API servers (they don't expire). We should make that more clear.  Maybe:

```
In order to enable endpoints to be discovered by control planes in other clusters,
we need to share secrets that allow each control plane to access the
Kubernetes API Server in each cluster. This allows each control plane
to discover all of the endpoints the mesh.
```

content/en/docs/setup/install/multicluster/single-network/primary/index.md
nmittler: Yes, we have to do that today. We should say that here.

content/en/docs/setup/install/multicluster/single-network/remote/index.md
nmittler: `CLUSTER_SECONDARY` wouldn't be a good name, since we have no concept of "secondary" in Istio. It could be `PRIMARY1`, `PRIMARY2`, and `REMOTE`, however.

content/en/docs/setup/install/multicluster/single-network/remote/index.md
nmittler: For remote, we're using ingress for LB. Otherwise, we have to use pod IP which is volatile. We should make that more clear.

content/en/docs/setup/install/multicluster/single-network/remote/index.md
nmittler: I believe so.  @ostromart can you confirm?

content/en/docs/setup/install/multicluster/single-network/verify/index.md
nmittler: It's a valid point. Do we use curl or fortio anywhere else currently?

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -1,7 +1,8 @@
 ---
 title: Multicluster Installation
-description: Configure an Istio mesh spanning multiple Kubernetes clusters.
+description: Configure an Istio mesh spanning multiple clusters.
```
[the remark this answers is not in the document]
rcaballeromx: Hi Craig,
Let me explain, the description you see is a Hugo "feature". To fix this, a field would need to be added to the front matter to show/hide the description and the template for bundle pages changed accordingly.
In the past, I've requested this issue be addressed and it has been consistently ignored.
Fixing this issue is outside the scope of this PR, and commenting about it on my contribution after my requests to address it have been ignored for months, seems inappropriate, even if it is addressed to the wider audience.
Please address this issue in a different PR for all content contributors moving forward.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -10,12 +11,58 @@ keywords: [kubernetes,multicluster]
...
 ---
 
-{{< tip >}}
-Note that these instructions are not mutually exclusive.
-In a large multicluster deployment, composed from more than two clusters,
-a combination of the approaches can be used. For example,
-two clusters might share a control plane while a third has its own.
-{{< /tip >}}
+This guide shows you how to deploy a multicluster
+{{< gloss >}}service mesh{{< /gloss >}}.
```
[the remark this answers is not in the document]
rcaballeromx: The glossary entry for multicluster is on the next sentence. That way the glossary mouseover is clearly different for each term.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -10,12 +11,58 @@ keywords: [kubernetes,multicluster]
...
 
-Refer to the [multicluster deployment model](/docs/ops/deployment/deployment-models/#multiple-clusters)
-concept documentation for more information.
+Familiarity with [single cluster deployments](/docs/ops/deployment/deployment-models/#single-cluster)
+is needed before attempting a {{< gloss >}}multicluster{{< /gloss >}} deployment.
+
+To deploy a service mesh across multiple networks, go to the following legacy guides:
```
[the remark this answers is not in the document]
rcaballeromx: Legacy means previous yet still valid, not deprecated. Technical writing best practice dictates that no time based information be included because it creates maintenance issues over time.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -10,12 +11,58 @@ keywords: [kubernetes,multicluster]
...
+
+- [Shared control planes](/docs/setup/install/multicluster/legacy/shared)
+    provides information on adding remote clusters on other networks.
+
+## Before you begin
+
+Choose a [deployment model](/docs/ops/deployment/deployment-models/) that suits
+your needs before you build your multicluster deployment. If you already have a
+single cluster service mesh, you can
+[add clusters that are within the same network to your mesh.](/docs/setup/install/multicluster/single-network/)
```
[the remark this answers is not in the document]
rcaballeromx: If they don't readers already left for the legacy guides above. The mechanism of adding new clusters is not part of the legacy content included underneath the hierarchy of this page. Thus, including the description here wouldn't work. I've included that description in the intro of the single network page.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -10,12 +11,58 @@ keywords: [kubernetes,multicluster]
...
+
+### Prerequisites
+
+To successfully deploy a multicluster service mesh, you need to meet the
+following prerequisites:
+
+- Your {{< gloss "pod" >}}pods{{< /gloss >}} and
+  {{< gloss "workload" >}}workloads{{< /gloss >}} meet [Istio's requirements.](/docs/ops/deployment/requirements/)
```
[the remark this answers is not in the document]
rcaballeromx: This was a direct request of the TOC. Including the requirements here allows readers to assess if they meet them without necessarily leaving the page.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -10,12 +11,58 @@ keywords: [kubernetes,multicluster]
...
+
+- Your {{< gloss "cluster" >}}clusters{{< /gloss >}} are using a [supported platform.](/docs/setup/platform-setup/)
+
+- Individual cluster pod [CIDR](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing)
+  ranges and service CIDR ranges that are unique across the network without any overlap.
+
+- All pod CIDRs in the same network can route to each other.
```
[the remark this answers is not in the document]
rcaballeromx: Similarly, for usability, I defined the requirement using the context already introduced to readers. Linking out our mentioning it as the pods being on the single network would force a context switch.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -10,12 +11,58 @@ keywords: [kubernetes,multicluster]
...
+  [default Kubernetes configuration file.](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/)
+  That default is either `~/.kube/config` or the value stored in the
+  `KUBECONFIG` environment variable. Relying on the default allows us to
+  easily switch between clusters.
+
+### Set environment variables {#env-var}
+
+{{< boilerplate mc-env-var >}}
+
+To continue, pick from the following options:
```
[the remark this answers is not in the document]
rcaballeromx: The complexity of the content here reflects the complexity of the feature. I kept the navigation as simple as possible. The bottom links present a nice decision tree for readers.

content/en/docs/setup/install/multicluster/legacy/gateways/index.md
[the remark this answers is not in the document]
rcaballeromx: Identifying the multi-network content on the legacy guides is outside the scope of this PR.

content/en/docs/setup/install/multicluster/single-network/_index.md
```diff
@@ -0,0 +1,30 @@
...
+owner: istio/wg-environments-maintainers
+test: n/a
+---
+
+This guide provides all the information needed to complete a {{< gloss >}}multicluster{{< /gloss >}} deployment on a
+[single network.](/docs/ops/deployment/deployment-models/#single-network)
+Go to the [multicluster deployment models page](/docs/ops/deployment/deployment-models/#multiple-clusters)
+for more information.
+
+To start a multicluster deployment, go to the [initial configuration section.](/docs/setup/install/multicluster/single-network/initial-configuration)
```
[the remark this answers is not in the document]
rcaballeromx: This is a landing page explaining the mechanics of the guide. This higher-level description came about in direct response to TOC feedback saying that the detailed flow being too complex.

content/en/docs/setup/install/multicluster/single-network/_index.md
rcaballeromx: This is a landing page explaining the mechanics of the guide. This higher-level description came about in direct response to TOC feedback saying that the detailed flow being too complex. The deployment models concept covers the answers to your questions.

content/en/docs/setup/install/multicluster/single-network/_index.md
rcaballeromx: That is conceptual information that readers don't need to know to move on to one of the pages. It is also clearly explained in the deployment models concept.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: The TOC pointed out that readers might come to the guide with a cluster already installed, or many other different states. Thus, the previous title you mention doesn't work well. `Initial Configuration` works better because these are the steps readers need to take to setup things initially for a multicluster deployment.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: This was feedback I got from the TOC. This shouldn't be anyone's "first primary cluster" since folks reading this document must be familiar with single cluster meshes. Those already have a primary cluster, and members of the TOC found it confusing to start this guide with a "first" cluster of any kind. I explicitly removed all mentions of installing Istio from the introduction of this page because of feedback from the TOC that the content should be accessible for folks with an existing single cluster mesh.

Calling this doc the initial configuration is accurate since it is the set of steps folks need to complete to configure a cluster to start their multicluster deployment. 

Currently, people starting out should go to the getting started and deploy a single cluster mesh. Multicluster is not for them at the moment.

@craigbox @nmittler The potential future transformation of this page into something more suitable for beginners should be handled separately at a later date.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: All proxies in all diagrams are depicted that way, so I'm keeping things consistently.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: The opening paragraph is consistent across all pages of the guide. The topic is not expanding the mesh but adding a primary cluster to it. The sentence is built to offer readers a ramp too.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: I'm not commenting on the quality of the docs, just letting readers know what the initial expected state is and why.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: Linking allows us to avoid information overload by not having to define all those concepts on this page. Readers don't need to click the link.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: When there's a global config, the paragraph can be revisited as a separate PR at a future date. Rewording the paragraph is possible but it was the result of multiple compromises already.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: I know but the sentence is in place for folks arriving to the section through links. It's an accessibility feature.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: Letting readers know the assumption for the following steps at this point, is considered technical writing best practice.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: Once the glossary entry for `load balancing` is introduced, a follow up PR can implement mouseovers for it here. An issue to add that entry has been filed already.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: This procedure doesn't require full understanding of secrets and we provide steps to generate and push the secrets to a cluster. A PR introducing better security conceptual content about secrets would allow us to provide a link for readers to learn more.

content/en/docs/setup/install/multicluster/single-network/remote/index.md
rcaballeromx: The naming schema is due to the content reusability and modularity requirements. Using the names you propose, would mean no longer being able to reuse content across docs causing an increase in maintenance overhead.

content/en/docs/setup/install/multicluster/single-network/remote/index.md
rcaballeromx: This is explained on the deployment models concept page explaining this in a lot of detail. That page is linked across the guide multiple times. Do you want me to add a link here too?

content/en/docs/setup/install/multicluster/single-network/remote/index.md
rcaballeromx: I could add the gateway to the diagram and further explanation to the paragraph but doing so adds a lot of technical nuance that readers don't need to know to complete the instructions. I can add a link to the gateway concept page for readers that need more information.

content/en/docs/setup/install/multicluster/single-network/remote/index.md
rcaballeromx: It needs the `<code>` tag, thanks.

content/en/docs/setup/install/multicluster/single-network/remote/index.md
rcaballeromx: That step is on line 146, thanks.

content/en/docs/setup/install/multicluster/single-network/verify/index.md
rcaballeromx: I can rewrite the sentence to `We use the sleep service as a simple load-generating service.`

content/en/docs/setup/install/multicluster/single-network/verify/index.md
rcaballeromx: We need to tell folks how to undo their deployment and reach the original state.

content/en/boilerplates/mc-env-var.md
```diff
@@ -0,0 +1,54 @@
+---
+---
+To simplify the deployment instructions and use the examples provided, set the
+following environment variables. If you use your own string values, ensure they
+are viable as filenames and compatible with `Makefiles`. Go to the [GNU guide](https://www.gnu.org/software/make/manual/html_node/Rule-Syntax.html)
```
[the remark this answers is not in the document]
rcaballeromx: I split it this way to avoid adding `Makefile` to the exceptions of the spell checker, and putting it in backticks makes the render to fail.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: Next update moves it a little further up to avoid the issue, thanks.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: The naming schema enables content reusability.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: We address users directly and they are not just adding a control plane to any cluster, but specifically to their cluster.

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: Thanks

content/en/docs/setup/install/multicluster/single-network/primary/index.md
rcaballeromx: Added a tip with that information.

content/en/docs/setup/install/multicluster/_index.md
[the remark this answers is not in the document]
craigbox: Do you have an issue link to this request?  I'm happy to pursue it.

content/en/docs/setup/install/multicluster/single-network/verify/index.md
craigbox: The bigger point is that _we should not be using a pod called `sleep` to do something other than sleep._ If you want me to raise an issue for this as a general rule, please let me know.

content/en/docs/setup/install/multicluster/single-network/remote/index.md
rcaballeromx: Thanks, the issue was the indentation.

content/en/docs/setup/install/multicluster/single-network/verify/index.md
frankbu: Maybe a bad name, but this [sleep service](https://github.com/istio/istio/blob/master/samples/sleep/README.md#simple-sleep-service) was designed for this purpose and is used pervasively in our task docs as a source service for demonstrating various Istio features. We use fortio, instead, for a source when we want to generate load.

frankbu: Just to be clear, the sleep service is not a "load-generating service". It's purpose to provide an "in-mesh pod" that users can exec into and use curl to make requests that are internal mesh (sidecar-to-sidecar) traffic.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+
+    {{< text bash >}}
+    $ cd ${WORK_DIR}
+    {{< /text >}}
+
+1. Generate the root CA and the intermediate CA for the cluster with the
+   following command:
+
+    {{< text bash >}}
+    $ make -f ${ISTIO}/install/certs/Makefile ${CLUSTER_1}-certs
```
nschhina, line 74: `make -f ${ISTIO}/tools/certs/Makefile`  1.6 and 1.7 don't have Makefile at this location.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+        meshExpansion:
+            enabled: true
+    EOF
+    {{< /text >}}
+
+1. Install Istio on the first cluster with the following command:
+
+    {{< text bash >}}
+    $ istioctl --context=${CTX_1} manifest apply -f \
+        {WORK_DIR}/${CLUSTER_1}/install.yaml
```
nschhina, line 187: missing `$` for WORK_DIR

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+1. To pass configuration values to the Istio operator for installation,
+   define a custom resource (CR). Define and save the `install.yaml` CR with
+   the following command:
+
+    {{< text bash >}}
+    $ cat <<EOF> ${WORK_DIR}/${CLUSTER_1}/install.yaml
+    apiVersion: install.istio.io/v1alpha1
+    kind: IstioOperator
+    spec:
+    values:
```
nschhina, line 160: The formatting for this yaml is completely off- values should exist in spec

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+        global:
+        meshID: ${MESH}
```
nschhina, line 162: everything below `global` falls under global requires tab for all lines after this

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+        meshNetworks:
+            ${NETWORK_1}:
+            endpoints:
+            # Always use ‘Kubernetes' as the registry name for
+            # endpoints that reside on the same network as this
+            # instance of the Istio control plane.
+            - fromRegistry: Kubernetes
+            gateways:
+            - registry_service_name: istio-ingressgateway.istio-system.svc.cluster.local
+                port: 443
```
nschhina, line 176: port should be at the same level as `registry_service_name`

nschhina (changes requested):
Main concern I have with the Makefile used in this task not existing in istio 1.6 and 1.7 and the istio operator yaml is not formatted correctly.

rcaballeromx:
@frankbu That assumption is incorrect, this is the final draft for publication as soon as possible, ideally as part of 1.7. I'm removing the label because of this.

content/en/docs/setup/install/multicluster/_index.md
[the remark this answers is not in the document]
frankbu: I think the more general problem is that this PR is using an `_index.md` file for doc, when it is really intended to just be a directory page (with little written material). We tried [doing this in the past](https://github.com/istio/istio.io/blob/release-1.1/content/docs/setup/kubernetes/_index.md), and it didn't work our well for several reasons, so we went back to keeping the documentation primarily in `index.md` files as Hugo expects.

frankbu:
@rcaballeromx Given the size of this PR and all the concerns from Craig , Nate, and  Navraj, I can't imagine how we could possibly get this into 1.7. It would also need testing, but nobody will want to write tests for content that is far from being ready to  freeze. I myself, agree with many of Craig's comments (re: continuity and flow) and also feel the overall structure using `_index.md` files needs to change. I also don't want to see the old docs moved to a  new (legacy) location, until there are replacements (btw, they need aliases, or your PR will be a breaking change) for them.

Maybe, if you restructure this to move it into an "Experimental Multicluster Install" section and leave the old docs where they are, we could merge it and continue to work on it from there, but I would still think that will need to be after the 1.7 branch is cut, so we can work on it at preliminary.istio.io only. I don't see this material as acceptable content in the mainline istio.io docs any this point in time.

frankbu (changes requested):
See previous comment.

craigbox:
It is my wish to see this in 1.7; most of my comments aren't related to this PR, but the way docs are put together in total, and they will be addressed in a different place.  Please don't block on anything I wrote. I'll find a more appropriate place.

frankbu:
@craigbox A couple of your comments are pretty significant issues IMO, and should be addressed before we merge this. That said, I think that the overall state of this PR, esp. it's untested and a major disruption very late in the release cycle, makes it seem like it will hurt the docs more than help, if we force it in as is. My concerns are certainly mitigated if @rcaballeromx changes it as I suggested, i.e., add it as new WIP docs, leaving the old docs in place for now, instead of moving the old docs to a legacy folder and putting this material front-and-center. I could see releasing that in 1.7, but even there it will need more testing to at least make sure it's hangs together and works for now. As is, it does not.

content/en/boilerplates/trust-config.md
```diff
@@ -0,0 +1,45 @@
...
+1. Go to the `${WORK_DIR}` folder with the following command:
+
+{{< text bash >}}
+$ cd ${WORK_DIR}
+{{< /text >}}
+
+1. Generate the intermediate CA files for `Cluster_2` with the following command:
+
+{{< text bash >}}
+$ make -f ${ISTIO}/samples/certs/Makefile ${CLUSTER_2}-certs
```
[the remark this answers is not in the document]
rcaballeromx: @nschhina Thanks but your suggestion used `Cluster_1`, I've modified it and applied it to the content.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: I've modified the command following your previous suggestion.

content/en/boilerplates/trust-config.md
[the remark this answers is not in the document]
nschhina: I'm actually not sure if  `${NAME}-cacerts-k8s` is the correct make command @nmittler can you confirm? I know there's no ${NAME}-certs make command anymore

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: I've addressed the indentation.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: I've addressed the indentation.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: I've addressed the indentation.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
rcaballeromx: Thanks, added.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
nschhina: same here @nmittler since you own the certs/Makefile you probably know better

nschhina:
I also had issues adding a remote cluster with Istio 1.7 alpha. Sidecar injection in the remote cluster was not working and pilot was not being discovered by the remote cluster. I don't know if there's some feature gap, some configuration error on my part or some configuration error in the docs. During the testing meeting please confirm if adding a remote cluster works fine.

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+    kind: IstioOperator
+    spec:
+        values:
+            global:
+                meshID: ${MESH}
+                multiCluster:
+                    clusterName: ${CLUSTER_1}
+                network: ${NETWORK_1}
+
+        meshNetworks:
```
nschhina, line 167:
```       
            meshNetworks:
            ${NETWORK_1}:
            endpoints:
            # Always use ‘Kubernetes' as the registry name for
            # endpoints that reside on the same network as this
            # instance of the Istio control plane.
            - fromRegistry: Kubernetes
            gateways:
            - registry_service_name: istio-ingressgateway.istio-system.svc.cluster.local
            port: 443
        # Expose the control plane through istio-ingressgateway.
        meshExpansion:
            enabled: true
```

You missed the indentation for these lines, all of them fall under global, so we need two more tabs until very the last line

nschhina (changes requested):

content/en/docs/setup/install/multicluster/single-network/initial-configuration/index.md
```diff
@@ -0,0 +1,205 @@
...
+is a simple tool that helps generate the root certificate and intermediate
+certificates, but do not use it as a production-ready root CA. The tool
+lacks the following critical features:
+
+- Private key protection
+- Access control
+- Auditing
+- Monitoring
+
+Use the `Makefile` to generate the common root and the intermediate certificates
```
nmittler, line 58: This is outdated. We should not be using this Makefile anymore. Instead, we should be using the same `samples/certs` for every cluster. We're no longer generating intermediate certs, etc.

nmittler:
@frankbu @craigbox @nschhina just trying to loop back on this ... have your comments been addressed? If not, can you list the issues that remain? I think it's safe to assume this will not land for 1.7.  So, given that, what are your remaining concerns/comments?

I know @stevenctl had temporarily [forked this PR](https://github.com/istio/istio.io/pull/7888) with some fixes to get things working. @rcaballeromx can you work with @stevenctl to incorporate those changes here?

frankbu:
Superseded by #8180

PR #7792, opened 2020-07-27
Fix missing timeout unit for istioctl experimental wait

nschhina:
Cherrypick to 1.6 and 1.7


frankbu:
No need for 1.7 cherrypick, since we haven't cut the 1.7 branch of istio.io yet.

PR #7798, opened 2020-07-28
fix kubectl patch instructions for meshConfig.outboundTrafficPolicy

vadimeisenbergibm:
take into account the case when meshConfig.outboundTrafficPolicy is not defined

[ X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


vadimeisenbergibm:
cc @GregHanson @frankbu

PR #7791, opened 2020-07-25
Add docs for new istioctl and operator change

richardwxn:
1. update uninstall step in istioctl installation guide and upgrade guide with new `istioctl x uinstall` command
2. update operator installation guide for more details about canary upgrade

PR #7811, opened 2020-07-29
Updates to the main istioctl install page

howardjohn:
* Favor passing --manifests (see
https://github.com/istio/istio/pull/25951)

* Restructure things a bit, to avoid a huge wall of top level headers.
For example, all "profile" stuff is under a profile section

* Remove some commands that I feel are too in depth/not used enough to
get this "front page" documentation space. I can revert this if
controversial

My primary goal is for deprecating the built in charts; I am fine with
removing the other changes

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/istioctl/index.md
+++ b/content/en/docs/setup/install/istioctl/index.md
@@ -45,13 +45,15 @@ The simplest option is to install the `default` Istio
 using the following command:
 
 {{< text bash >}}
-$ istioctl install
+$ istioctl install --manifests=manifests/
 {{< /text >}}
 
 This command installs the `default` profile on the cluster defined by your
 Kubernetes configuration. The `default` profile is a good starting point
 for establishing a production environment, unlike the larger `demo` profile that
 is intended for evaluating a broad set of Istio features.
+
+The `manifests` flag determines which installation manifests should be used for installation. The installation manifests are included in the release tar in the `manifests` directory.
 
 Various settings can be configured to modify the installations. For example, to enable access logs:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/istioctl/index.md
```diff
@@ -45,14 +45,16 @@ The simplest option is to install the `default` Istio
 using the following command:
 
 {{< text bash >}}
-$ istioctl install
+$ istioctl install --manifests=manifests/
```
frankbu, line 48: This breaks the flow of the doc. The sentence, just above, says this is the `simplest option`, but then passes the `--manifests` arg. The following example, installs the demo profile, without passing `--manifests`, so that seems simpler. 

Also, I think there are other docs that say or imply that all you need is to download `istioctl` to install Istio. Users don't NEED to download the release. This change makes downloading the release a prereq for using `istioctl install`, which sounds bad.

howardjohn: Ok, we need to pass `--manifests` everywhere is the primary goal of the PR. The agreement in Environments WG was we would remove the "compiled in" charts. Seems like that will have a bigger docs impact than I originally expected

frankbu: You should probably wait for @richardwxn's PR to merge before making this change because it sounds like the commands he's adding will also need to change: https://github.com/istio/istio.io/pull/7791

frankbu: Or maybe not? He's only doing `istioctl uninstall`

content/en/docs/setup/install/istioctl/index.md
```diff
@@ -45,14 +45,16 @@ The simplest option is to install the `default` Istio
...
 {{< /text >}}
 
 This command installs the `default` profile on the cluster defined by your
 Kubernetes configuration. The `default` profile is a good starting point
 for establishing a production environment, unlike the larger `demo` profile that
 is intended for evaluating a broad set of Istio features.
 
+The `manifests` flag determines which installation manifests should be used for installation. The installation manifests are included in the release tar in the `manifests` directory.
```
ostromart, line 56: we should mention something about version compatibility early on in the doc e.g. istioctl is guaranteed to be compatible with the same minor version of manifests e.g. istioctl v1.7.2 is compatible with all 1.7.x manifests.

content/en/docs/setup/install/istioctl/index.md
ostromart: uninstall will also need charts to be passed in most cases but that flag is already supported in the subcommand.

content/en/docs/setup/install/istioctl/index.md
```diff
@@ -223,40 +211,6 @@ resources must be removed manually.
 
 {{< /warning >}}
 
-## Show differences in manifests
```
ostromart, line 226: maybe move this somewhere in the advanced section rather than completely removing?

sdake:
See this thread for a  discussion John and I had via slack:

https://github.com/istio/istio/pull/25951#issuecomment-667472832

frankbu:
@sdake @howardjohn So is the conclusion that we should close this PR and wait until 1.8?

ostromart:
I propose a compromise - could we start with strongly recommending using external charts to lay the groundwork for removing compiled-in charts? I think everyone agrees this is good practice. Then we can roll out this change in a more gradual way (e.g. put in a deprecation message) to minimize disruption to users.

PR #7815, opened 2020-07-30
fix secret name

therealmitchconnors:
Currently, the docs tell you to create secret named `cacerts`, but the generate_cert tool looks for `istio-ca-secret`.  It is possible that the tool is the one that is wrong, in which case I can submit a second PR.

### DIFF
--- a/content/en/docs/examples/virtual-machines/single-network/index.md
+++ b/content/en/docs/examples/virtual-machines/single-network/index.md
@@ -60,7 +60,7 @@ following commands on a machine with cluster admin privileges:
 
     {{< text bash >}}
     $ kubectl create namespace istio-system
-    $ kubectl create secret generic cacerts -n istio-system \
+    $ kubectl create secret generic istio-ca-cert -n istio-system \
         --from-file=@samples/certs/ca-cert.pem@ \
         --from-file=@samples/certs/ca-key.pem@ \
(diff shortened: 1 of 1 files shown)


content/en/docs/examples/virtual-machines/single-network/index.md
```diff
@@ -60,7 +60,7 @@ following commands on a machine with cluster admin privileges:
 
     {{< text bash >}}
     $ kubectl create namespace istio-system
-    $ kubectl create secret generic cacerts -n istio-system \
+    $ kubectl create secret generic istio-ca-cert -n istio-system \
```
frankbu, line 63: Is this a design change that also needs more doc updates, for example here: https://preliminary.istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/#plugging-in-existing-certificates-and-key?

therealmitchconnors: This is hard coded in https://github.com/istio/istio/blob/master/security/tools/generate_cert/main.go#L98 as istio-ca-secret.  I don't know if the tool is wrong or the docs are wrong, but they should agree on the name.

frankbu:
IIUC, this change is incorrect and the PR is not needed. Therefore closing.

Please reopen if I'm missing something.

PR #7832, opened 2020-07-31
Egress Gateway Automated Doc Test

nschhina:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/tasks/traffic-management/egress/egress-gateway/index.md
```diff
@@ -37,7 +37,33 @@ on them cannot access the Internet. Defining an egress gateway, directing all th
 allocating public IPs to the egress gateway nodes allows the application nodes to access external services in a
 controlled way.
 
-{{< boilerplate before-you-begin-egress >}}
+## Before you begin
```
frankbu, line 40: This is an interesting change. I assume you did this because you noticed that the boilerplate snips don't get generated in the doc that uses them. What we've done in other docs with this issue, is:

1. Ignore the snips and just do the equivalent setup using the helper functions in `samples.sh`
2. Assume the boilerplate snips will be tested separately.

This approach avoids retesting the same boilerplate snips over and over in every doc that uses them, and avoids duplicating the text in the doc.

That said, if we don't actually have boilerplate tests, those snips might not actually be tested at all, but then it probably doesn't matter since the code in `samples.sh` is essentially the same simple startup anyway.

Another, maybe better, approach is to generate `snip.sh` for boilerplate files and then you can `source` it into the `test.sh` that uses it?

nschhina: I was trying to be consistent with other docs, I actually don't see any other traffic task use boilerplate? Doesn't make sense to just do it at one place and not all. You can decide what approach works best, I will make the changes if necessary

frankbu: Maybe the best thing, for now anyway, is to remove the duplicate text and use the samples function. Here's another test that is using the same boilerplate and doing it that way: https://github.com/istio/istio.io/blob/5905596d2532462a1e2aa9d11751e8145c520873/content/en/docs/tasks/observability/logs/access-log/test.sh#L37

nschhina: Done.

content/en/docs/tasks/traffic-management/egress/egress-gateway/test.sh
```diff
@@ -0,0 +1,124 @@
...
+set -u
+set -o pipefail
+
+source "tests/util/samples.sh"
+
+# Deploy sleep sample and set up variable pointing to it
+# Start the sleep sample
+startup_sleep_sample
+export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath='{.items[0].metadata.name}')
+_wait_for_deployment default sleep
```
frankbu, line 30: This line isn't needed because `startup_sleep_sample` conveniently does it already.

nschhina: gotchu

nschhina:
the test failure seems like a flake in a different test, but I'll re-run the CI for this

PR #7847, opened 2020-08-03
Do not explicitly set addon namespace

howardjohn:
its hardcoded in the manifest, and there are multiple namespaces used so
hardcoded it to istio-system will fail

Corresponding implementation: https://github.com/istio/istio/pull/26072

### DIFF
--- a/content/en/docs/ops/integrations/grafana/index.md
+++ b/content/en/docs/ops/integrations/grafana/index.md
@@ -26,7 +26,7 @@ There are a few ways to configure Grafana to use these dashboards:
 Istio provides a basic sample installation to quickly get Grafana up and running, bundled with all of the Istio dashboards already installed:
 
 {{< text bash >}}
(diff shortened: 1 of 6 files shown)


ericvn:
The last file uses the istio download (it's downloaded at the top of the page). The other pages don't have that restriction. I believe we should be OK with that.

howardjohn:
> All lines but in the last file are missing a -f

Thanks! fixed. That is what using bash aliases gets me..

ericvn:
Do we want/need to wait on https://github.com/istio/istio/issues/26037 and it's cherrypick before merging this? And do we want to make sure we update a test-ref to point to it as well. I know the getting started test (PR but waiting on a review to merge (I can't since I wrote it)) uses the command from the last file. Once we merge this, the snip will no longer have the namespace at it'll try loading from the current alpha.2 reference sample directory.

PR #7857, opened 2020-08-04
update proxy-config output

shamsher31:
Few updates to `proxy-status` and `proxy-config` output. Major update in `bootstrap` config output. 

Found during 1.7 testing.
```
$ curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.7.0-alpha.2 sh -
```

Note that I have updated version to `1.7.0` in the output instead of  `1.7.0-alpha.2`.

### DIFF
--- a/content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
+++ b/content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
@@ -35,14 +35,14 @@ receiving configuration or is out of sync then `proxy-status` will tell you this
 {{< text bash >}}
 $ istioctl proxy-status
 NAME                                                   CDS        LDS        EDS        RDS          ISTIOD                      VERSION
-details-v1-558b8b4b76-qzqsg.default                    SYNCED     SYNCED     SYNCED     SYNCED       istiod-6cf8d4f9cb-wm7x6     1.6.5
-istio-ingressgateway-66c994c45c-cmb7x.istio-system     SYNCED     SYNCED     SYNCED     NOT SENT     istiod-6cf8d4f9cb-wm7x6     1.6.5
-productpage-v1-6987489c74-nc7tj.default                SYNCED     SYNCED     SYNCED     SYNCED       istiod-6cf8d4f9cb-wm7x6     1.6.5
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/diagnostic-tools/proxy-cmd/index.md
```diff
@@ -61,48 +61,9 @@ where the issue may lie.
 
 {{< text bash json >}}
 $ istioctl proxy-status details-v1-6dcc6fbb9d-wsjz4.default
---- Istiod Clusters
```
ericvn, line 64: I'm a little concerned here since the text above says we are retrieving a diff, while the output being changed goes from a diff to things matching. It seems that there should be a diff in the output or maybe the section is no longer correct.

shamsher31: I think I noticed this in `1.6` also but was not sure whether it returns diffs. I haven't found any diffs during my testing. I'm I missing something?

ericvn: cc @esnible Do you know if this is supposed to show a diff as the docs suggest?

ericvn: Either the whole section needs to be removed (the title contains diff and the last sentence says to see the diffs but they are removed in this PR) or something is broken and we aren't seeing the diffs we should.

frankbu: Looks to me like there would only be a diff if the environment is set up to cause an update problem. If everything is working properly, there should be no diff.

So, I suggest leaving the Clusters part as it was, just update the Routes part to include the last update time.

esnible: If Istio is operating correctly, it will output that Envoy and Istio match.
If Envoy is rejecting the config supplied by Istiod then the command will show the difference between what Envoy believes and what Istiod believes.  Often this gives an advanced troubleshooter an idea about what is wrong.

shamsher31: Keeping diffs as it is. Thanks.

PR #7859, opened 2020-08-04
Update controlz interface img and output command

shamsher31:
Component Introspection:
Ref: https://preliminary.istio.io/latest/docs/ops/diagnostic-tools/controlz/


content/en/docs/ops/diagnostic-tools/controlz/index.md
```diff
@@ -14,23 +14,22 @@ Istio components are built with a flexible introspection framework which makes i
...
-{{< image width="80%" link="./ctrlz.png" caption="ControlZ User Interface" >}}
+{{< image width="90%" link="./ctrlz.png" caption="ControlZ User Interface" >}}
 
-To access the ControlZ page of deployed components (i.e. Mixer, Galley, Pilot), you can port-forward their ControlZ endpoints
+To access the ControlZ page of deployed component (i.e. istiod), you can port-forward their ControlZ endpoints
 locally and connect through your local browser:
 
 {{< text bash >}}
-$ istioctl dashboard controlz <podname>
+$ istioctl dashboard controlz istiod-954c974bb-qgb4h -n istio-system
```
ericvn, line 32: Not sure why the <podname> was changed, although I know when automating a page, the <> in the bash has lint issues. In this cases, the <>'s have just been removed so this line would have become `istioctl dashboard controlz podium`

shamsher31: This page was very old and haven't been updated for a long time. As all the functionality is combined into `istiod`, it's better to mention `podname` directly in the command. Also, the namespace flag was missing.

content/en/docs/ops/diagnostic-tools/controlz/index.md
```diff
@@ -14,23 +14,22 @@ Istio components are built with a flexible introspection framework which makes i
 of a running component. Components open a port which can be used from a web browser to get an interactive view into the state of the
 component, or via REST for access and control from external tools.
 
-Mixer, Pilot, and Galley all implement the ControlZ functionality. When these components start, a message is logged indicating the
+Istiod implement the ControlZ functionality. When the component start, a message is logged indicating the
```
frankbu, line 17:
```suggestion
Istiod implement the ControlZ functionality. When it starts, a message is logged indicating the
```

content/en/docs/ops/diagnostic-tools/controlz/index.md
```diff
@@ -14,23 +14,22 @@ Istio components are built with a flexible introspection framework which makes i
...
+2020-08-04T23:28:48.889370Z     info    ControlZ available at 100.76.122.230:9876
 {{< /text >}}
 
 Here's sample of the ControlZ interface:
 
-{{< image width="80%" link="./ctrlz.png" caption="ControlZ User Interface" >}}
+{{< image width="90%" link="./ctrlz.png" caption="ControlZ User Interface" >}}
 
-To access the ControlZ page of deployed components (i.e. Mixer, Galley, Pilot), you can port-forward their ControlZ endpoints
+To access the ControlZ page of deployed component (i.e. istiod), you can port-forward their ControlZ endpoints
```
frankbu, line 28:
```suggestion
To access the ControlZ page of istiod, you can port-forward its ControlZ endpoint
```

content/en/docs/ops/diagnostic-tools/controlz/index.md
frankbu, line 32:
```suggestion
$ istioctl dashboard controlz <istiod-podname> -n istio-system
```
@ericvn now that we support `snip_id=none`, we can use it for snippets with invalid bash syntax, e.g., `<podname>`, instead of removing the `<` and `>`. Since this kind of example is not a runnable command, as is, there's no need to generate the snip anyway.

shamsher31: @frankbu Is it ok to name it like `istiod-<podname>` as `istiod` will not change?

frankbu: @shamsher31 That's not quite right, because the `<...>` part would not be a podname, but rather a pod name suffix. Maybe a better (less confusing) syntax would be `<istiod pod name>`?

frankbu: Updated suggestion above

shamsher31: Thanks

content/en/docs/ops/diagnostic-tools/controlz/index.md
```diff
@@ -14,23 +14,22 @@ Istio components are built with a flexible introspection framework which makes i
 of a running component. Components open a port which can be used from a web browser to get an interactive view into the state of the
```
frankbu, line 14: Is istiod the only component that this applies to now? If so, should we change the title, description, and first paragraph to be more specific?
```
title: Istiod Introspection
description: Describes how to use ControlZ to get insight into a running istiod component.
...
Istiod is build with a flexible introspection framework, called ControlZ, which makes it easy to inspect and manipulate the internal state
of an istiod instance. Istiod opens a port which can be used from a web browser to get an interactive view into its state,
or via REST for access and control from external tools.

When Istiod starts, a message is logged indicating the IP address and port to connect to in order to interact with ControlZ.
```

shamsher31: AFAIK it's true. Thanks.

PR #7865, opened 2020-08-04
Redirect download scripts to master instead of stable branch

jwendell:
Fixes https://github.com/istio/istio/issues/26115

### DIFF
--- a/layouts/index.redir
+++ b/layouts/index.redir
@@ -35,8 +35,8 @@ https://istio.netlify.com/* https://istio.io/:splat 301!
 {{- end }}
 
 # Download scripts
-/downloadIstio https://raw.githubusercontent.com/istio/istio/release-{{ .Site.Data.args.version }}/release/downloadIstioCandidate.sh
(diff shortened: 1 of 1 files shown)


jwendell:
How can we use different scripts per release if the url redirects to only one place? We don't have one url per release. It's a single [short] url.

If we, for some reason, modify the script to work on 1.7 that somehow only works for 1.7 (hypothetically speaking), then the url is broken for all other versions. In practice, we should have only one script that works for every supported version.

That's why I see that script belonging to another [neutral] place, for example, test-infra, which only has master branch and contains everything to support all the supported releases.

shamsher31:
> If we, for some reason, modify the script to work on 1.7 that somehow only works for 1.7 (hypothetically speaking), then the url is broken for all other versions. In practice, we should have only one script that works for every supported version.

I agree with you @jwendell. The download URL should point to master.

ericvn:
Added to WG agenda. I'm going more off prior process and discussions I had with Martin when making changes to the script in the past.

ericvn:
Per WG discussion, we will go back to using the script in `master`

PR #7866, opened 2020-08-04
Fix broken circuit-breaking test

frankbu:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/circuit-breaking/index.md
+++ b/content/en/docs/tasks/traffic-management/circuit-breaking/index.md
@@ -105,7 +105,7 @@ governed by Istio.
     {{< /text >}}
 
 1. Log in to the client pod and use the fortio tool to call `httpbin`.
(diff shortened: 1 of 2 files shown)


ldemailly:
apologies looks like -quiet doesn't apply at the stage the dynamic config is checked, let me fix that bug but in the meantime maybe use `2> /dev/null` instead?

ldemailly:
fix incoming in https://github.com/fortio/fortio/pull/386/files

ldemailly:
I fixed the fortio image (`fortio/fortio:latest_release` is now 1.6.6) so let's use

`-- /usr/local/bin/fortio curl -quiet URL`

frankbu:
Force merging because lint failure is an unrelated problem.

PR #7869, opened 2020-08-04
Replace usage of meshExpansion flag with istiod gateway code sample

tbarrella:
As described in https://github.com/istio/istio/issues/25933, the meshExpansion flag is being deprecated in favor of user config that can be added post-installation. This updates docs accordingly

/hold
Note that this isn't ready to be merged for the 1.7 release because https://github.com/istio/istio/pull/26091 is past the cutoff

Also, for this to fully work, either I need to finish https://github.com/istio/istio/issues/25933 by adding more ports to the ingress Service by default, or these docs would need to be further modified to instruct users to patch in port 15012. I prefer the former, but first wanted feedback on this, in case it would be better to make a new Example or Task just for exposing Istiod and such a page should be linked in the release notes/deprecation warning

[ ] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/multicluster/shared/index.md
+++ b/content/en/docs/setup/install/multicluster/shared/index.md
@@ -212,7 +212,7 @@ Apply the primary cluster's configuration.
 $ istioctl install -f istio-main-cluster.yaml --context=${MAIN_CLUSTER_CTX}
 {{< /text >}}
 
-If you're using the `istio-ingressgateway` option, expose the control plane by using the provided sample configuration.
+If you selected the `istio-ingressgateway` option, expose the control plane using the provided sample configuration.
 
 {{< text bash >}}
 $ kubectl apply -f @samples/istiod-gateway/istiod-gateway.yaml@
--- a/content/en/docs/setup/install/multicluster/shared/index.md
+++ b/content/en/docs/setup/install/multicluster/shared/index.md
@@ -154,10 +154,6 @@ spec:
           gateways:
           - registry_service_name: istio-ingressgateway.istio-system.svc.cluster.local
             port: 443
-
-      # Use the existing istio-ingressgateway.
-      meshExpansion:
-        enabled: true
 EOF
 {{< /text >}}
 
@@ -214,6 +210,12 @@ Apply the primary cluster's configuration.
 
 {{< text bash >}}
 $ istioctl install -f istio-main-cluster.yaml --context=${MAIN_CLUSTER_CTX}
+{{< /text >}}
+
+If you're using the `istio-ingressgateway` option, expose the control plane by using the provided sample configuration.
+
+{{< text bash >}}
+$ kubectl apply -f @samples/istiod-gateway/istiod-gateway.yaml@
 {{< /text >}}
 
 Wait for the control plane to be ready before proceeding.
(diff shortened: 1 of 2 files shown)


tbarrella:
@adammil2000 would you have a chance to look at this?

content/en/docs/setup/install/multicluster/shared/index.md
```diff
@@ -216,6 +212,12 @@ Apply the primary cluster's configuration.
 $ istioctl install -f istio-main-cluster.yaml --context=${MAIN_CLUSTER_CTX}
 {{< /text >}}
 
+If using the `istio-ingressgateway` option, expose Istiod, for example by using the provided sample configuration.
```
nmittler, line 215: this sentence seems like it's missing something. Maybe `you can expose istiod...`?

Also, maybe we should call it `the control plane` instead of `istiod`?

content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -46,26 +46,18 @@ and `SERVICE_ACCOUNT`
...
-      values:
-        global:
-          meshExpansion:
-            enabled: true
-    EOF
+    $ istioctl install
     {{< /text >}}
 
-1. Install or upgrade Istio with virtual machine integration features enabled.
+1. Expose Istiod, for example by using the provided code sample.
```
nmittler, line 57: Can we get rid of `for example`?

content/en/docs/setup/install/multicluster/shared/index.md
```diff
@@ -216,6 +212,12 @@ Apply the primary cluster's configuration.
 $ istioctl install -f istio-main-cluster.yaml --context=${MAIN_CLUSTER_CTX}
 {{< /text >}}
 
+If you're using the `istio-ingressgateway` option, expose the control plane by using the provided sample configuration.
```
frankbu, line 215: What to do if you are NOT using the `istio-ingressgateway` option. I don't think that's possible. How about?
```suggestion
Expose the control plane by configuring the `istio-ingressgateway` using the provided sample configuration.
```

content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -46,26 +46,18 @@ and `SERVICE_ACCOUNT`
...
-      values:
-        global:
-          meshExpansion:
-            enabled: true
-    EOF
+    $ istioctl install
     {{< /text >}}
 
-1. Install or upgrade Istio with virtual machine integration features enabled.
+1. Expose the control plane by using the provided code sample.
```
frankbu, line 57:
```suggestion
1. Expose the control plane using the provided sample configuration.
```

content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -46,26 +46,18 @@ and `SERVICE_ACCOUNT`
 
 ## Install the Istio control plane
 
-Install Istio with the installation setting `values.global.meshExpansion.enabled: true`.
+Install Istio and expose the control plane so that virtual machines can reach it.
```
frankbu, line 49:
```suggestion
Install Istio and expose the control plane so that your virtual machine can access it.
```

content/en/docs/setup/install/multicluster/shared/index.md
tbarrella: In the other case (Internal Load Balancer), the control plane is already exposed by the `components` part of the install config

frankbu: I see. Maybe this would be a little more clear?
```suggestion
If you selected the `istio-ingressgateway` option, expose the control plane using the provided sample configuration.
```

Issue #7827, opened 2020-07-31
Install with Istioctl references citadel


ericvn:
https://preliminary.istio.io/latest/docs/setup/install/istioctl/

has a snippet:

```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
spec:
  components:
    citadel:
      namespace: istio-citadel

```

which yields an error: `Error: failed to install manifests: unknown field "citadel" in v1alpha1.IstioComponentSetSpec:`

PR #7872, opened 2020-08-05
Remove citadel from installation instructions

myidpt:

Fixes https://github.com/istio/istio.io/issues/7827.

[ ] Docs
[ X ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/istioctl/index.md
+++ b/content/en/docs/setup/install/istioctl/index.md
@@ -362,26 +362,6 @@ spec:
 $ istioctl install -f telemetry_off.yaml
 {{< /text >}}
 
-Another customization is to select different namespaces for features and components. The following is an example
-of installation namespace customization:
-
-{{< text yaml >}}
-apiVersion: install.istio.io/v1alpha1
-kind: IstioOperator
-metadata:
-  namespace: istio-system
-spec:
-  components:
-    citadel:
-      namespace: istio-citadel
-{{< /text >}}
-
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/istioctl/index.md
```diff
@@ -362,26 +362,6 @@ spec:
 $ istioctl install -f telemetry_off.yaml
 {{< /text >}}
 
-Another customization is to select different namespaces for features and components. The following is an example
```
frankbu, line 365: Instead of completely removing this section, should we change the example to something else that illustrates the same feature (components different namespaces), e.g., an ingress-gateway in a different namespace?

cc @ostromart

frankbu:
@myidpt Unless it's a critical issue, we don't cherrypick to archived releases, just the current release and master.

content/en/docs/setup/install/istioctl/index.md
ostromart: I'm actually in favor of removing. Although theoretically different namespaces can be selected, in practice using the same namespace is what's recommended and tested. In fact we may want to consider deprecating this part of the API and use only the global namespace (gateways are an exception).

ericvn: Wow, and I remember when we said it was better to have everything in its own namespace (and not that long ago).

PR #7888, opened 2020-08-07
new multicluster deployment guides

stevenctl:
Based on https://github.com/istio/istio.io/pull/7787 with fixes for #26017 to unblock 1.7 release/testing. 

closes https://github.com/istio/istio/issues/26017

[ ] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/boilerplates/trust-config.md
```diff
@@ -0,0 +1,45 @@
...
+1. Go to the `${WORK_DIR}` folder with the following command:
+
+{{< text bash >}}
+$ cd ${WORK_DIR}
+{{< /text >}}
+
+1. Generate the intermediate CA files for `Cluster_2` with the following command:
+
+{{< text bash >}}
+$ make -f ${ISTIO}/tools/certs/Makefile ${CLUSTER_2}-cacerts-k8s
```
stevenctl, line 18: This does not seem to work:

```
make -f $ISTIO/tools/certs/Makefile cluster1-cacerts-k8s    ✘ 130 
fetching root ca from k8s cluster: gke_landow-istio-dev_us-central1-c_cluster-1
Error from server (NotFound): secrets "istio-ca-secret" not found
Error from server (NotFound): secrets "istio-ca-secret" not found
generating cluster1/ca-key.pem
Generating RSA private key, 4096 bit long modulus (2 primes)
......................................................................++++
....................++++
e is 65537 (0x010001)
generating cluster1/cluster-ca.csr
generating cluster1/ca-cert.pem
Signature ok
subject=O = Istio, CN = Intermediate CA, L = cluster1
unable to load certificate
4687273408:error:0909006C:PEM routines:get_name:no start line:crypto/pem/pem_lib.c:745:Expecting: TRUSTED CERTIFICATE
make: *** [cluster1/ca-cert.pem] Error 1
rm cluster1/cluster-ca.csr cluster1/intermediate.conf
```

I noticed the makefile has some kubectl usages without context. Wondering if that's an issue when setting up additional clusters? 

The rest of these docs I've run through successfully while just using the pre-generated samples certs. 

cc @nmittler

ericvn:
Verified the current guide does not work for shared control plane multiple networks. Will try again with this new set of docs.

stevenctl:
@ericvn there is a code fix for the certs issue (we won't need to set caAddress). With that set, let me know if there are multinetwork issues.

ericvn:
Looks like I was running into Looks like I’m running into https://github.com/istio/istio/issues/23364.  So doc is still wrong. Hopefully just need to add `--set values.global.centralIstiod=true` to the instioctl installs for remote and main in docs/setup/install/multicluster/legacy/shared/.

Note that the prometheus listed on the pod outputs on that page is wrong as well.

stevenctl:
> Hopefully just need to add --set values.global.centralIstiod=true

centraliIstiod should be totally separate from our traditional multicluster deployments. If we require that for every multicluster installation that's a bug.

ericvn:
@irisdingbj @linsun I believe you found this setting was required even for the older multicluster installments? I needed it for the legacy shared control plane multiple network to get things to work. On my remote clusters, istio-proxy would not start and istiod had errors like `User "system:serviceaccount:istio-system:istio-reader-service-account" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope.`.

linsun:
--set values.global.centralIstiod=true should not be needed for our old multicluster install.  It is only needed when you don't have CA or istiod at all on the remote cluster.

nmittler:
> --set values.global.centralIstiod=true should not be needed for our old multicluster install. It is only needed when you don't have CA or istiod at all on the remote cluster.

It unfortunately is today. This was a regression introduced recently. I have a fix pending: https://github.com/istio/istio/pull/26334

nmittler:
All, given the state of this PR and the push-back, I suspect that we should just abandon it (and https://github.com/istio/istio.io/pull/7787) at this point.

I'm currently working on istio installation docs that will get rid of the multi-cluster distinction entirely, and will include multi-network support. I suggest that we just wait until post 1.7 and go with that. I'll be sharing this doc as soon as everything is tested (currently the only blocker is https://github.com/istio/istio/issues/26325).

WDYT @frankbu @craigbox @nschhina @rcaballeromx @linsun @stevenctl

ericvn:
So the shared multi network docs are broken today for IKS.... I don't know about GKE at all.

frankbu:
@nmittler, that makes sense but we need to make sure that what's documented for 1.7 works. It sounds like the old instructions are currently broken.

Can we discuss the 1.7 multicluster plan in the doc WG call tomorrow?

linsun:
+1 we need some resolution for this in 1.7.  I am curious if the mutli network docs work for GKE.  I don't expect us to use central istiod if the local istiod on the remote cluster is running.

frankbu:
>  I don't expect us to use central istiod if the local istiod on the remote cluster is running.

I don't understand this. A "remote cluster", by definition, does not have a "local istiod".

https://preliminary.istio.io/latest/docs/reference/glossary/#remote-cluster

linsun:
Based on this config - https://preliminary.istio.io/latest/docs/setup/install/multicluster/shared/#remote-cluster

We are not disabling istiod on the remote cluster, which means remote cluster will have istiod as well.

nmittler:
@frankbu @linsun if the old instructions are broken, that's a separate discussion (and probably a bug). We should fix that for 1.7.  Seems like a release blocker.

FYI, I've just fixed an issue with remote clusters where installation now required manually setting `caAddress`: https://github.com/istio/istio/pull/26334. I suspect that was at least of the issues with the old docs.

linsun:
@ericvn I can see #26334 impact you.  Would be good to retest with a build that has it on IKS if you have some bandwidth.  If it still doesn't work, I agree it would be a release blocker.

ericvn:
Assuming this is in rc.1, I can test when it's available. I just found another issue where another idea was mentioned: https://github.com/istio/istio/issues/26017#issuecomment-667315335

nmittler:
@ericvn 

> Assuming this is in rc.1, I can test when it's available. I just found another issue where another idea was mentioned: istio/istio#26017 (comment)

That comment is referring to the need for `caAddress`, which I fixed in istio/istio#26334. So I think it's the same issue, unless I'm missing something.

nmittler:
FYI I had to manually create a cherry-pick: https://github.com/istio/istio/pull/26421.

PR #7890, opened 2020-08-07
remove how to enable app http probe instruction

linsun:
as this is enabled by default for all profiles now.

[ x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/ops/configuration/mesh/app-health-check/index.md
+++ b/content/en/docs/ops/configuration/mesh/app-health-check/index.md
@@ -110,7 +110,7 @@ This feature is enabled by default when installing with any of our [profiles](/d
 
 ### Separate port
 
-Another alternative is to use separate port for health checking and regular traffic.  Seperate port is not recommended as it requires changing your application's health check on a seperate port. This option should only be explored when the `probe rewrite` option doesn't work.
+Another alternative is to use separate port for health checking and regular traffic.  Separate port is not recommended as it requires changing your application's health check on a separate port. This option should only be explored when the `probe rewrite` option doesn't work.
 
 ### Disable the probe rewrite option globally
 
@@ -120,7 +120,7 @@ Another alternative is to use separate port for health checking and regular traf
 $ kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e 's/"rewriteAppHTTPProbe": true/"rewriteAppHTTPProbe": false/' | kubectl apply -f -
 {{< /text >}}
 
-### Explore the seperate port option
+### Explore the separate port option
 
 Run these commands to re-deploy the service:
 
--- a/content/en/docs/ops/configuration/mesh/app-health-check/index.md
+++ b/content/en/docs/ops/configuration/mesh/app-health-check/index.md
@@ -106,84 +106,21 @@ This approach rewrites the application `PodSpec` readiness/liveness probe, such 
 [Pilot agent](/docs/reference/commands/pilot-agent/). Pilot agent then redirects the
 request to application, and strips the response body only returning the response code.
 
-This feature is enabled when installing with the `default` profile. If you find that the profile used to install Istio does not have it enabled, you have two ways to enable the rewrite of the liveness HTTP probes:
-
-#### Enable globally via install option
-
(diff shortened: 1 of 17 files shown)


frankbu:
Just removing these instructions leaves the whole document making little sense. It talks about options, and then one of the options just says this is always enabled, which makes the rest of the options irrelevant. The doc would need to be changed to explain the default behavior, and then explain the other options, why a user would want to change it, and how to disable the default.

linsun:
@frankbu Thanks for the feedback, I've reworked the instruction based on it, PTAL.  I 'll work on fixing make gen, tests etc.

content/en/docs/ops/configuration/mesh/app-health-check/index.md
```diff
@@ -106,84 +106,21 @@ This approach rewrites the application `PodSpec` readiness/liveness probe, such
...
-            port: 8001
-          initialDelaySeconds: 5
-          periodSeconds: 5
-{{< /text >}}
-
-This approach allows you to enable the health check prober rewrite gradually on each deployment without reinstalling Istio.
+### Separate port
 
-#### Re-deploy the liveness health check app
+Another alternative is to use separate port for health checking and regular traffic.  Separate port is not recommended as it requires changing your application's health check on a separate port. This option should only be explored when the `probe rewrite` option doesn't work.
```
howardjohn, line 113: this option doesn't even work today

content/en/docs/ops/configuration/mesh/app-health-check/index.md
```diff
@@ -106,84 +106,21 @@ This approach rewrites the application `PodSpec` readiness/liveness probe, such
...
-$ kubectl -n istio-same-port get pod
-NAME                             READY     STATUS    RESTARTS   AGE
-liveness-http-975595bb6-5b2z7c   2/2       Running   0           1m
+$ kubectl get cm istio-sidecar-injector -n istio-system -o yaml | sed -e 's/"rewriteAppHTTPProbe": true/"rewriteAppHTTPProbe": false/' | kubectl apply -f -
 {{< /text >}}
 
-### Separate port
-
-Another alternative is to use separate port for health checking and regular traffic.
+### Explore the separate port option
```
howardjohn, line 123: this should be linked to the Separate port option above - but both should be removed

content/en/docs/ops/configuration/mesh/app-health-check/index.md
```diff
@@ -106,84 +106,21 @@ This approach rewrites the application `PodSpec` readiness/liveness probe, such
...
-    matchLabels:
-      app: liveness-http
-      version: v1
-  template:
-    metadata:
-      labels:
-        app: liveness-http
-        version: v1
-      annotations:
-        sidecar.istio.io/rewriteAppHTTPProbers: "true"
```
howardjohn, line 151: should document using this to *disable*

content/en/docs/ops/configuration/mesh/app-health-check/index.md
linsun: hmm why would separate port won't work?

howardjohn: we capture all ports. It would work if you use a separate port AND use excludeInboundPorts I guess, but no one will do that I guess. A readiness port on a different port is useless too.. the whole point is its testing the app, if you test another port it loses value

linsun: That is basically what the sample does.  It exposes the health check on a separate port than the main app.  We had it as before we had probes rewrite.   Now the reason i didn't delete yet it is if it can be a workaround when probes rewrite is disabled.   @myidpt @frankbu thoughts?

howardjohn:
If doesn't work - we used to only capture some ports, now we capture them
all

linsun:
Okay removed the separate port instruction.  I agree it is too hard to use that option.

@frankbu @howardjohn PTAL.

linsun:
our users have reported this doc inconsistent issue on 1.6

PR #7891, opened 2020-08-07
Move back doc links to v1beta1 since we are still supporting 1.15

ericvn:
See https://github.com/istio/istio/pull/26285


frankbu:
The platform setup page says that Istio 1.7 supports Kubernetes releases: 1.16, 1.17, 1.18, buy not 1.15. Is that wrong?

ericvn:
@howardjohn I'm not sure if the 1.15 is a full support statement or more of a we won't break people who use 1.15. I suspect it's the former, but haven't seen an official statement (after the doc we had approved where 1.15 wasn't supported).

howardjohn:
I think we will unofficially support 1.15 *maybe*. We might not even. the webhook stuff is only semi-related, we had to revert because of upgrade issues

Issue #7853, opened 2020-08-04
Correct hanging reference in VM Docs


therealmitchconnors:
https://github.com/istio/istio.io/blob/master/content/en/docs/examples/virtual-machines/multi-network/index.md#preparing-the-kubernetes-cluster-for-vms

In Step 1, the grammar indicates that there will be instructions following the step, but there are no instructions.  Instructions should be added, or the wording changed.

jasonwzm:
We were discussing this before. Assigning to myself, Limin and John to work on the current VM docs.

duderino:
Looks like https://github.com/istio/istio.io/commit/479b4b161cec6427d37aab3cc642f2795b9a1d3c#diff-66df94b4108652e624a42b21d8bc3394 unintentionally deleted the example?  @adammil2000 was this intentional?

duderino:
Since this is a release blocker, let's have a single owner so we don't have a diffusion of responsibility problem.  Jason, I think you can satisfy this blocker by just restoring the missing example in my prior comment.  A general revamp of the VM docs would be welcome, but it wouldn't block the 1.7 release.

Also, if you don't think you're a good fit for this please let me know so I can find another owner.

jasonwzm:
@duderino sounds good. I think in TOC meeting we also mentioned the single network setup instruction is not working. Is that a release blocker as well?

liminw:
Sounds good. Jason, let me know if you need help on security setup. /cc @williamaronli also has context since he worked on VM identity bootstrapping for 1.7.

PR #7894, opened 2020-08-07
Fix multi network VM doc

jasonwzm:
Fixes https://github.com/istio/istio.io/issues/7853

Created https://github.com/istio/istio/issues/26294 for further fixes.

/cc @williamaronli @liminw PTAL

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


sdake:
@jasonwzm Thank you for the work here, but this change is not suitable. Instead the makefile needs to be fixed. I think this PR is in need of a revert. There was significant debate around creating certs instead of using sample certs in the original PR, and the decision was that certs **MUST** be generated.

Cheers,
-steve

frankbu:
Reverting this PR wouldn't change anything, since the instructions before and after are the same, just direct instead of indirect. If there's a need to change the instructions, they need to be changed in both this and the single network doc in a followup PR, but I don't see anything wrong with using the sample certs in these docs, since they are just examples and both warn against using them in production.

PR #7898, opened 2020-08-10
Remove redundant warning for VM installation.

shamsher31:
The warning was added in https://github.com/istio/istio.io/pull/7461 but the `Makefile` was moved into `tools/certs` for 1.6 in https://github.com/istio/istio/pull/23897


frankbu:
The warning is needed because it is specifically about `1.6.0`. The Makefile was moved in `1.6.1`.

PR #7903, opened 2020-08-10
Show examples in `Check what's installed` section

jwendell:


content/en/docs/setup/install/istioctl/index.md
```diff
@@ -94,10 +94,25 @@ $ istioctl install --set profile=demo
 ## Check what's installed
 
 The `istioctl` command saves the `IstioOperator` CR that was used to install Istio in a copy of the CR named `installed-state`.
-You can inspect this CR if you lose track of what is installed in a cluster.
+You can inspect this CR if you lose track of what is installed in a cluster. For example, use the command below to dump its content into the file `installed-state.yaml` for further inspection:
```
frankbu, line 97:
```suggestion
If you lose track of what is installed in a cluster, you can inspect this CR.
For example, dump its content into a YAML file using the following command:
```

content/en/docs/setup/install/istioctl/index.md
```diff
@@ -94,10 +94,25 @@ $ istioctl install --set profile=demo
...
+
+{{< text bash >}}
+$ kubectl -n istio-system get deploy
+NAME                   READY   UP-TO-DATE   AVAILABLE   AGE
+istio-ingressgateway   1/1     1            1           49m
+istiod                 1/1     1            1           49m
+{{< /text >}}
+
+The results may vary according to the profile or options you used to install Istio.
+
```
frankbu, line 116: This is obvious. It's also not really a practical way of figuring out the "installed state" of the cluster, so I suggest removing.
```suggestion
```

jwendell: I agree that's obvious for us. Is it for people new to Istio? My intention was just give some pointers for users to know what to look for, or how a default installation should look like. As a user, that's what I expect in a section called `Check what’s installed`.

frankbu: In that case it should really be in the getting started doc (https://preliminary.istio.io/latest/docs/setup/getting-started/#install), this doc is for users with more experience. But, even there, we removed commands like that, which we had used previously used to help people understand what's under the covers.

jwendell: Yeah, I remember seeing such snippets in the past. They were useful to me. Anyway, I respect your judgment, you're the expert here, not me :)

content/en/docs/setup/install/istioctl/index.md
```diff
@@ -94,7 +94,12 @@ $ istioctl install --set profile=demo
 ## Check what's installed
 
 The `istioctl` command saves the `IstioOperator` CR that was used to install Istio in a copy of the CR named `installed-state`.
-You can inspect this CR if you lose track of what is installed in a cluster.
+If you lose track of what is installed in a cluster, you can inspect this CR.
+For example, dump its content into a YAML file using the following command:
```
frankbu, line 98: Maybe a more subtle way to mention getting the resources would be ok?

```suggestion
Instead of inspecting the deployments, pods, services and other resources that were installed by Istio, for example:

{{< text bash >}}
$ kubectl -n istio-system get deploy
NAME                   READY   UP-TO-DATE   AVAILABLE   AGE
istio-ingressgateway   1/1     1            1           49m
istiod                 1/1     1            1           49m
{{< /text >}}

You can inspect the `installed-state` CR, to see what is installed in the cluster, as well as all custom settings.
For example, dump its content into a YAML file using the following command:
```

PR #7912, opened 2020-08-11
Remove deprecated addonComponents

shamsher31:
Based on https://github.com/istio/istio/pull/24567

`Grafana` is now available as an addon and can be [installed independently](https://preliminary.istio.io/latest/docs/ops/integrations/grafana/).

### DIFF
--- a/content/en/docs/setup/install/operator/index.md
+++ b/content/en/docs/setup/install/operator/index.md
@@ -138,8 +138,8 @@ spec:
 EOF
 {{< /text >}}
 
-You can also enable or disable components and modify resource settings.
-For example, to enable the `Grafana` component and increase pilot memory requests:
+You can also modify resource settings.
+For example, to increase pilot memory requests:
 
 {{< text bash >}}
 $ kubectl apply -f - <<EOF
@@ -156,9 +156,6 @@ spec:
         resources:
           requests:
             memory: 3072Mi
-  addonComponents:
-    grafana:
-      enabled: true
 EOF
 {{< /text >}}
 
--- a/content/en/docs/setup/install/operator/index.md
+++ b/content/en/docs/setup/install/operator/index.md
@@ -138,8 +138,8 @@ spec:
 EOF
 {{< /text >}}
 
-You can also modify resource settings.
-For example, to increase pilot memory requests:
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/operator/index.md
```diff
@@ -138,8 +138,8 @@ spec:
 EOF
 {{< /text >}}
 
-You can also enable or disable components and modify resource settings.
-For example, to enable the `Grafana` component and increase pilot memory requests:
```
frankbu, line 142: Rather than removing the part about enable/disable components, why not change the example to egress-gateway instead?

frankbu: Also, shouln't `pilot` -> `istiod`?

howardjohn: +1

thanks for doing this!

shamsher31: > Rather than removing the part about enable/disable components, why not change the example to egress-gateway instead?

Done

> Also, shouln't pilot -> istiod?

The [default](https://github.com/istio/istio/blob/master/manifests/profiles/default.yaml#L22) and [other](https://github.com/istio/istio/blob/master/manifests/profiles/demo.yaml#L47) profiles still use `pilot`.

shamsher31: > Rather than removing the part about enable/disable components, why not change the example to egress-gateway instead?

Done

> Also, shouln't pilot -> istiod?

The [default](https://github.com/istio/istio/blob/master/manifests/profiles/default.yaml#L22) and [other](https://github.com/istio/istio/blob/master/manifests/profiles/demo.yaml#L47) profiles still use `pilot`.

content/en/docs/setup/install/operator/index.md
```diff
@@ -139,7 +139,7 @@ EOF
 {{< /text >}}
 
 You can also enable or disable components and modify resource settings.
-For example, to enable the `Grafana` component and increase pilot memory requests:
+For example, to enable the `egressGateways` component and increase pilot memory requests:
```
frankbu, line 142:
```suggestion
For example, to enable the `istio-egressgateway` component and increase pilot memory requests:
```

Issue #7915, opened 2020-08-11
Improve references to enabling mutual TLS in BookInfo install page


rvennam:
https://istio.io/latest/docs/examples/bookinfo/#apply-default-destination-rules
<img width="901" alt="image" src="https://user-images.githubusercontent.com/5502967/89932823-35cae980-dbdc-11ea-9459-66f6464607ec.png">

If users installed the default profile, what should they run? What if the user doesn't know what profile they have (managed istio users). With auto-mTLS, does it matter?

@GregHanson

PR #7921, opened 2020-08-11
update wording on bookinfo destination rules differences

GregHanson:
Some missed conditional logic that was required prior to Istio auto mTLS

Resolves https://github.com/istio/istio.io/issues/7915

[ X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/examples/bookinfo/index.md
+++ b/content/en/docs/examples/bookinfo/index.md
@@ -192,22 +192,16 @@ versions, called *subsets*, in [destination rules](/docs/concepts/traffic-manage
 
 Run the following command to create default destination rules for the Bookinfo services:
 
-* If you did **not** enable mutual TLS, execute this command:
+{{< text bash >}}
+$ kubectl apply -f @samples/bookinfo/networking/destination-rule-all.yaml@
+{{< /text >}}
 
-    {{< tip >}}
-    Choose this option if you are new to Istio and are using the `demo`
-    [configuration profile](/docs/setup/additional-setup/config-profiles/).
-    {{< /tip >}}
-
-    {{< text bash >}}
-    $ kubectl apply -f @samples/bookinfo/networking/destination-rule-all.yaml@
-    {{< /text >}}
-
-* If you **did** enable mutual TLS, execute this command:
-
-    {{< text bash >}}
-    $ kubectl apply -f @samples/bookinfo/networking/destination-rule-all-mtls.yaml@
-    {{< /text >}}
+{{< tip >}}
+You can enforce mutual TLS using the `samples/bookinfo/networking/destination-rule-all-mtls.yaml`
+destination rule.  However, the `default` and `demo` [configuration profiles](/docs/setup/additional-setup/config-profiles/) have [auto mutual TLS](/docs/tasks/security/authentication/authn-policy/#auto-mutual-tls) enabled by default.  If a TLS context is not provided in a
+destination rule, Istio configures client proxies to send mutual TLS traffic to injected proxies
+automatically.
+{{< /tip >}}
 
 Wait a few seconds for the destination rules to propagate.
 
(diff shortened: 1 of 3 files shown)


content/en/docs/examples/bookinfo/index.md
```diff
@@ -192,22 +192,16 @@ versions, called *subsets*, in [destination rules](/docs/concepts/traffic-manage
...
+{{< /text >}}
 
-    {{< text bash >}}
-    $ kubectl apply -f @samples/bookinfo/networking/destination-rule-all-mtls.yaml@
-    {{< /text >}}
+{{< tip >}}
+You can enforce mutual TLS using the `samples/bookinfo/networking/destination-rule-all-mtls.yaml`
+destination rule.  However, the `default` and `demo` [configuration profiles](/docs/setup/additional-setup/config-profiles/) have [auto mutual TLS](/docs/tasks/security/authentication/authn-policy/#auto-mutual-tls) enabled by default.  If a TLS context is not provided in a
+destination rule, Istio configures client proxies to send mutual TLS traffic to injected proxies
+automatically.
```
frankbu, line 203: The last sentence seem unnecessary given the link to the auto-mutual-tls doc in the previous sentence. Also, maybe reversing the first two sentences would flow better?
```suggestion
The `default` and `demo` [configuration profiles](/docs/setup/additional-setup/config-profiles/) have [auto mutual TLS](/docs/tasks/security/authentication/authn-policy/#auto-mutual-tls) enabled by default.
To enforce mutual TLS, use the destination rules in `samples/bookinfo/networking/destination-rule-all-mtls.yaml`.
```

rvennam: In an effort to simplify our documentation, I wonder if we really need this tip. Shouldn't users just use AuthenticationPolicy to enforce mtls in their entire namespace/mesh instead of using DR to enforce tls per service?
I recommend removing this tip. I also think we should take it a step further and just remove this whole section and then move the DR apply command as a step in the previous section. Or just move it to `Request Routing` task

content/en/docs/examples/bookinfo/index.md
```diff
@@ -192,22 +192,16 @@ versions, called *subsets*, in [destination rules](/docs/concepts/traffic-manage
...
+{{< /tip >}}
 
 Wait a few seconds for the destination rules to propagate.
```
rvennam, line 206: Do you really need to wait a few seconds?

content/en/docs/examples/bookinfo/index.md
rvennam: I'm trying to avoid introducing users to Custom Resources they might not need know unless they are doing the task that needs them.

content/en/docs/examples/bookinfo/index.md
frankbu: Not really, since it only matters when route rules are applied, which isn't being done immediately afterwords. That said, the general rule is to wait for config to propagate before proceeding to whatever is done next.

content/en/docs/examples/bookinfo/index.md
frankbu: Removing the section entirely isn't a good idea IMO, because this app is deploying several versions of the review service, so defining their subsets should really be a prereq before moving on.

I think the tip is also fine since this app is used in a lot more places than just the getting started demo. I think it's much better as Greg made it a tip (easy to ignore when getting started) instead of inline noise, like it was before.

content/en/docs/examples/bookinfo/index.md
GregHanson: Most cases, in the time it takes for a user to copy and past the next set of instructions, things will have propagated by then.  However, I think it is good practice to leave comments like this in place to reiterate that it may take a few seconds for istio routing/authentication rules to go into effect.

PR #7922, opened 2020-08-11
Add docs for target_arch

brian-avery:
The TARGET_ARCH option for downloading Istio is currently not documented. This updates the example to point at a more recent version of Istio and documents the TARGET_ARCH.

ref: https://github.com/istio/istio/issues/24577

### DIFF
--- a/content/en/docs/setup/getting-started/index.md
+++ b/content/en/docs/setup/getting-started/index.md
@@ -44,7 +44,11 @@ Follow these steps to get started with Istio:
     The command above downloads the latest release (numerically) of Istio.
     To download a specific version, you can add a variable on the command line.
     For example to download Istio 1.4.3, you would run
-      `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.4.3 sh -`
+      `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.6.8 sh -`. 
+      
+    The processor architecture will be detected automatically, but can alternatively
+    be overridden with the `TARGET_ARCH` environment variable. For example: 
+    `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.6.8 TARGET_ARCH=x86_64 sh -`. 
     {{< /tip >}}
 
 1.  Move to the Istio package directory. For example, if the package is
--- a/content/en/docs/setup/getting-started/index.md
+++ b/content/en/docs/setup/getting-started/index.md
@@ -44,11 +44,10 @@ Follow these steps to get started with Istio:
     The command above downloads the latest release (numerically) of Istio.
     To download a specific version, you can add a variable on the command line.
     For example to download Istio 1.4.3, you would run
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/getting-started/index.md
```diff
@@ -44,7 +44,10 @@ Follow these steps to get started with Istio:
     The command above downloads the latest release (numerically) of Istio.
     To download a specific version, you can add a variable on the command line.
     For example to download Istio 1.4.3, you would run
-      `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.4.3 sh -`
+    `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.6.8 sh -`. The
```
ericvn, line 47: You need to update the 1.4.3 in the prior sentence.

brian-avery: Thanks. Fixed.

content/en/docs/setup/getting-started/index.md
```diff
@@ -43,8 +43,11 @@ Follow these steps to get started with Istio:
     {{< tip >}}
     The command above downloads the latest release (numerically) of Istio.
     To download a specific version, you can add a variable on the command line.
-    For example to download Istio 1.4.3, you would run
-      `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.4.3 sh -`
+    For example to download Istio 1.6.8, you would run
+    `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.6.8 sh -`. The
+    processor architecture can be overridden with the `TARGET_ARCH` environment
+    variable. For example, to download the x86_64, architecture, you would run
+    `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.6.8 TARGET_ARCH=x86_64 sh -`.
```
frankbu, line 50: This seems like TMI for the simple "Getting Started" doc. I was already wondering about the value of providing instructions for downloading a specific version, given that this doc is trying to be as simple as possible for new users just wanting to kick the tires. Maybe we could provide more "advanced" download instructions like this somewhere else, and just put a link to it in the tip in this doc?

brian-avery: That would work for me. Issue here was that people were opening issues because the TARGET_ARCH wasn't documented. As long as we put this/the ISTIO_VERSION somewhere I think it's fine.

frankbu: Since I can't see a good place to add the extra info, let's leave it here, but make it a little less verbose.
```
    {{< tip >}}
    The command above downloads the latest release (numerically) of Istio.
    You can pass variables on the command line to download a specific version
    or to override the processor architecture.
    For example, to download Istio 1.6.8 for the x86_64 architecture,
    run `curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.6.8 TARGET_ARCH=x86_64 sh -`.
    {{< /tip >}}
```

brian-avery: Added. Thanks :)

PR #7923, opened 2020-08-11
Explain pod annotation overrides for tracing

nrjpoddar:
[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/observability/distributed-tracing/configurability/index.md
+++ b/content/en/docs/tasks/observability/distributed-tracing/configurability/index.md
@@ -39,7 +39,13 @@ There are two ways you can configure tracing options:
 
 1.  Globally via `MeshConfig` options.
 
-1.  Per pod annotations for workload specific customization.
+1.  Per pod annotations for workload specific customization. Note that any pod
+    annotations added for tracing configuration overrides global settings. In order
+    to preserve any global settings you should copy them from
+    global mesh config to pod annotations along with workload specific
+    customization. In particular, make sure that the tracing backend address is
+    always provided in the annotations to ensure that the traces are reported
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/observability/distributed-tracing/configurability/index.md
```diff
@@ -39,7 +39,13 @@ There are two ways you can configure tracing options:
...
 1.  Globally via `MeshConfig` options.
 
-1.  Per pod annotations for workload specific customization.
+1.  Per pod annotations for workload specific customization. Note that any pod
+    annotations added for tracing configuration overrides global settings. In order
+    to preserve any global settings you should copy them from
+    global mesh config to pod annotations along with workload specific
+    customization. In particular, make sure that the tracing backend address is
+    always provided in the annotations to ensure that the traces are reported
+    correctly for the workload.
```
frankbu, line 48: I would suggest moving the warning, below, before the new text.

```suggestion
1.  Per pod annotations for workload specific customization.

{{< warning >}}
In order for the new tracing configuration to take effect for either of these
options you need to restart pods injected with Istio proxies.
{{< /warning >}}

Note that any pod annotations added for tracing configuration overrides
global settings. In order to preserve any global settings you should copy them
from global mesh config to pod annotations along with workload specific
customization. In particular, make sure that the tracing backend address is
always provided in the annotations to ensure that the traces are reported
correctly for the workload.
```

nrjpoddar:
@frankbu PTAL

PR #7924, opened 2020-08-12
small update to canary upgrade doc

linsun:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/upgrade/index.md
+++ b/content/en/docs/setup/upgrade/index.md
@@ -56,7 +56,7 @@ istio-sidecar-injector-canary   2020-04-28T19:03:26Z
 
 ### Data plane
 
-Installing the new revision has in place upgraded the istio ingress gateway, which is automatically configured to point to the 'istiod-canary' control plane.
+Installing the new revision has in place upgraded the istio-ingress gateway pod and service, which are automatically configured to point to the 'istiod-canary' control plane.
 
 {{< text bash >}}
 $ kubectl get pods -n istio-system -l app=istio-ingressgateway
--- a/content/en/docs/setup/upgrade/index.md
+++ b/content/en/docs/setup/upgrade/index.md
@@ -56,10 +56,10 @@ istio-sidecar-injector-canary   2020-04-28T19:03:26Z
 
 ### Data plane
 
-Unlike istiod, installing the new revision has in place upgraded all the istio gateway pods and services, which are automatically configured to point to the 'istiod-canary' control plane.  To verify that the `istio-ingress` gateway pod are using the `istiod-canary` service corresponding to the `canary` revision, select one newly created `istio-ingress` pod and use the `istio-ingress-pod_name` in the following command:
+Unlike istiod, installing the new revision has in place upgraded all the istio gateway pods and services, which are automatically configured to point to the 'istiod-canary' control plane.  To verify that the `istio-ingress` gateway pod are using the `istiod-canary` service corresponding to the `canary` revision, select one newly created `istio-ingress` pod and use the following command:
 
-{{< text bash >}}
-$ istioctl proxy-config endpoints ${istio-ingress-pod_name}.istio-system --cluster xds-grpc -ojson | grep hostname
+{{< text bash >}} 
+$ istioctl proxy-config endpoints $(kubectl -n istio-system get  pod -l app=istio-ingressgateway -o jsonpath='{.items..metadata.name}').istio-system --cluster xds-grpc -ojson | grep hostname
 "hostname": "istiod-canary.istio-system.svc"
 {{< /text >}}
 
--- a/content/en/docs/setup/upgrade/index.md
+++ b/content/en/docs/setup/upgrade/index.md
@@ -23,10 +23,10 @@ and [upgrading from 1.4.x](#upgrading-from-1.4).
 
 ### Control plane
 
-To install a new revision called `canary`, you would set the `revision` field as follows:
+To install a new revision called `canary` or the istio version you would set the `revision` field as follows:
 
 {{< text bash >}}
-$ istioctl install --set revision=canary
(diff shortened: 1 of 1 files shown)


linsun:
https://github.com/istio/istio.io/issues/7927

content/en/docs/setup/upgrade/index.md
```diff
@@ -56,18 +56,24 @@ istio-sidecar-injector-canary   2020-04-28T19:03:26Z
...
-Simply installing the new revision has no impact on the existing proxies. To upgrade these,
-you must configure them to point to the new control plane. This is controlled during sidecar injection
+Installing the new revision has in place upgraded the istio-ingress gateway pod and service, which are automatically configured to point to the 'istiod-canary' control plane.
+
+{{< text bash >}}
+$ kubectl get pods -n istio-system -l app=istio-ingressgateway
+NAME                                    READY   STATUS    RESTARTS   AGE
+istio-ingressgateway-7bf8767888-9cchz   1/1     Running   0          8m30s
+{{< /text >}}
+
```
frankbu, line 66: What's the purpose of this command? It doesn't seem to show anything to indicate "in place upgrade" has or will happen.
```suggestion
```

content/en/docs/setup/upgrade/index.md
```diff
@@ -56,18 +56,24 @@ istio-sidecar-injector-canary   2020-04-28T19:03:26Z
 
 ### Data plane
 
-Simply installing the new revision has no impact on the existing proxies. To upgrade these,
-you must configure them to point to the new control plane. This is controlled during sidecar injection
+Installing the new revision has in place upgraded the istio-ingress gateway pod and service, which are automatically configured to point to the 'istiod-canary' control plane.
```
frankbu, line 59: Is this only the case for the istio-ingress gateway? What about egress gateways, or even other ingress-gateways that have been installed? I think this needs more explanation.

frankbu: cc @howardjohn

howardjohn: All gateways are the same

linsun: I only used istio-ingress gateway because that is the one enabled by default for default profile.  I'll update to include all gateways.

content/en/docs/setup/upgrade/index.md
linsun: The purpose is to show it doesn't have a canary next to it, e.g. old one is removed and this new one is deployed.

linsun: Let me know - happy to remove it if there is no need to show it.

frankbu: Yes, I would suggest removing, but I think you need to add some text to explain this better. Something that explains that unlike istiod, gateways do not have revisions, but are updated to the latest revision, immediately when a new istiod revision is deployed. How does it know what is the latest revision tag to use?

content/en/docs/setup/upgrade/index.md
linsun: reworded @frankbu PTAL and let me know

content/en/docs/setup/upgrade/index.md
```diff
@@ -23,7 +23,7 @@ and [upgrading from 1.4.x](#upgrading-from-1.4).
 
 ### Control plane
 
-To install a new revision called `canary`, you would set the `revision` field as follows:
+To install a new revision called `canary` or the Istio version you would set the `revision` field as follows. If you use your Istio version, replace `.` in the version with `-` as `.` is not currently supported as a valid revision value.
```
frankbu, line 26:
```suggestion
To install a new revision called `canary`, you would set the `revision` field as follows:

<tip>
In a production environment, a better revision name would correspond to the Istio version.
However, you must replace `.` characters in the revision name, for example, `revision=1-6-8` for Istio `1.6.8`,
because `.` is not a valid revision name character.
</tip>
```

content/en/docs/setup/upgrade/index.md
```diff
@@ -56,18 +56,23 @@ istio-sidecar-injector-canary   2020-04-28T19:03:26Z
 
 ### Data plane
 
-Simply installing the new revision has no impact on the existing proxies. To upgrade these,
-you must configure them to point to the new control plane. This is controlled during sidecar injection
+Unlike Istiod, installing the new revision has in place upgraded all the Istio gateway pods and services, which are automatically configured to point to the 'istiod-canary' control plane.  To verify that the `istio-ingress` gateway pod are using the `istiod-canary` service corresponding to the `canary` revision, select one newly created `istio-ingress` pod and use the following command:
```
frankbu, line 59:
```suggestion
Unlike Istiod, Istio gateways do not run revision-specific instances, but are instead in-place upgraded to use the new control plane revision.
You can verify that the `istio-ingress` gateway is using the `canary` revision by running the following command:
```

content/en/docs/setup/upgrade/index.md
```diff
@@ -56,18 +56,23 @@ istio-sidecar-injector-canary   2020-04-28T19:03:26Z
...
+
+{{< text bash >}}
+$ istioctl proxy-config endpoints $(kubectl -n istio-system get  pod -l app=istio-ingressgateway -o jsonpath='{.items..metadata.name}').istio-system --cluster xds-grpc -ojson | grep hostname
```
frankbu, line 62:
```suggestion
$ istioctl proxy-config endpoints $(kubectl -n istio-system get pod -l app=istio-ingressgateway -o jsonpath='{.items..metadata.name}').istio-system --cluster xds-grpc -ojson | grep hostname
```

linsun:
@frankbu thank you for the nice suggestions!

Issue #7927, opened 2020-08-12
canary upgrade removal old cp command is not working


linsun:
when there is no revision.

```
$ istioctl x uninstall --revision default                                                                  
The following resources will be pruned from the cluster: 
Proceed? (y/N) y
✔ Uninstall complete                                                                                                                   (⎈ |mycluster-wdc06-b3c.4x16/bs76p72w0qc3sa93oti0:default)
~/Downloads/istio-1.7.0-beta.2/bin ⌚ 21:33:37
$ k get pods -n istio-system -l app=istiod                      
NAME                             READY   STATUS    RESTARTS   AGE
istiod-9f9cb4674-jl8gr           1/1     Running   0          12d
istiod-canary-7445496947-sbv48   1/1     Running   0          22m
```

```
$ istioctl x uninstall -f manifests/profiles/default.yaml 
There are still 5 proxies pointing to the control plane revision :
helloworld-v1-7d568fc748-7gd5s.default
helloworld-v2-5c55cbd644-n7j4n.default
httpbin-5b87f575f4-tmf72.default
istio-ingressgateway-7bf8767888-9cchz.istio-system
sleep-579db58c8c-fwrmf.default
If you proceed with the uninstall, these proxies will become detached from any control plane and will not function correctly. Proceed? (y/N) 
```

linsun:
cc @richardwxn @ostromart

linsun:
```
$ istioctl version                                       
client version: 1.7.0-beta.2
pilot version: 1.8-alpha.d5ab2ebfa13107099a6fed596b5201f88ad28d24-d5ab2ebfa13107099a6fed596b5201f88ad28d24
istiod version: 1.7.0-beta.2
data plane version: 1.7.0-beta.2 (5 proxies)
(⎈ |mycluster-wdc06-b3c.4x16/bs76p72w0qc3sa93oti0:default)
~/Downloads/istio-1.7.0-beta.2 ⌚ 10:30:13
$ k get pod helloworld-v1-7d568fc748-7gd5s -o yaml | grep canary                                                    
    istio.io/rev: canary
        {"discoveryAddress":"istiod-canary.istio-system.svc:15012","proxyMetadata":{"DNS_AGENT":""}}
```

linsun:
note: steps in #7930 need to fixed else users would end up w/ a non-working cluster.

richardwxn:
[...]

PR #7943, opened 2020-08-13
add warning about uninstall with filename

richardwxn:
This is applicable to `istioctl x uninstall` with filename flag. Resolves: https://github.com/istio/istio.io/issues/7927. The actual fix is: https://github.com/istio/istio/pull/26449

### DIFF
--- a/content/en/docs/setup/upgrade/index.md
+++ b/content/en/docs/setup/upgrade/index.md
@@ -102,15 +102,15 @@ To uninstall a control plane of revision `istio-1-6-5`, run:
 $ istioctl x uninstall --revision istio-1-6-5
 {{< /text >}}
 
+{{< warning >}}
+This command will also remove gateway deployments for specified revision. If you want to remove the canary revision, please reinstall the gateway(s) with previous revision first to avoid down time.
+{{< /warning >}}
+
 If the old control plane does not have a revision label, uninstall it using its original installation options, for example:
 
 {{< text bash >}}
 $ istioctl x uninstall -f manifests/profiles/default.yaml
 {{< /text >}}
-
-{{< warning >}}
-Note that this command would also remove gateway deployments, use it with caution. The issue would be fixed in release `1.7.1`
-{{< /warning >}}
 
 Confirm that the old control plane has been removed and only the new one still exists in the cluster:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/upgrade/index.md
```diff
@@ -108,6 +108,10 @@ If the old control plane does not have a revision label, uninstall it using its
 $ istioctl x uninstall -f manifests/profiles/default.yaml
 {{< /text >}}
 
+{{< warning >}}
+Note that this command would also remove gateway deployments, use it with caution. The issue would be fixed in release `1.7.1`
```
frankbu, line 112:
```suggestion
This command will also remove gateway deployments, which will then need to be redeployed. The issue will be fixed in release `1.7.1`
```

linsun: @richardwxn if the fix is NOT in 1.7.0, can we just recommend users how to clean up manually without uninstall cmd?

Using uninstall cmd would create a non working cluster for users without the fix.

richardwxn: @linsun the use cases of --revision and --purge works, which I expect most people would use, I can remove the part of -f, wdyt?

linsun: Looks like the fix will land in 1.7.0 afterall.  If --revision works for default, I agree with removing the -f portion.

richardwxn: updated the description

content/en/docs/setup/upgrade/index.md
```diff
@@ -102,6 +102,10 @@ To uninstall a control plane of revision `istio-1-6-5`, run:
 $ istioctl x uninstall --revision istio-1-6-5
 {{< /text >}}
 
+{{< warning >}}
+This command will also remove gateway deployments for specified revision. If you want to remove the canary revision, please reinstall the gateway(s) with previous revision first to avoid down time.
```
linsun, line 106: Is this correct all the time?  I think it won't remove if I'm deleting the old revision (1-6-5) but would if i'm deleting the new revision (1-7-0).

linsun: I propose to have 2 sections, 1 for remove old cp, 1 for remove the new version (canary).  This warning only applied to the 2nd scenario.

frankbu: I added some text for this along with another fix in #7957. If that looks alright, then we can close this PR.

richardwxn: @linsun yes, that's why I emphasized `canary` one here. But if that's not clear enough, then different section is ok

frankbu:
Closing, because this is addressed in #7957. Please reopen or send a followup PR if more changes are needed.

PR #7944, opened 2020-08-13
Automated test for customizing TCP metrics

su225:
Automated test for verifying example given in https://preliminary.istio.io/latest/docs/tasks/observability/metrics/customize-metrics/

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
```diff
@@ -0,0 +1,62 @@
...
+for i in {1..100}; do
+  curl http://$gateway_url/productpage > /dev/null
+done
+
+# Finally verify results
+# I cannot use the command in the snippet because it has pod-name, but I need an actual pod
+productpage_pod=$(kubectl get pod -l app=productpage -o jsonpath='{.items[0].metadata.name}')
+metric_count=$(kubectl exec $productpage_pod -c istio-proxy -- curl 'localhost:15000/stats/prometheus' | grep 'istio_requests_total' | wc -l)
+echo "obtained count: $metric_count"
+__cmp_at_least $metric_count 1
```
frankbu, line 58: You should call an `_verify_at_least` function (instead of `__cmp_at_least` directly) so that you get the automatic retry which is needed in case the metrics are slow to appear.
```suggestion
get_productpage_metrics() {
    kubectl exec $productpage_pod -c istio-proxy -- curl 'localhost:15000/stats/prometheus' | grep 'istio_requests_total' | wc -l
}
_verify_at_least get_productpage_metrics 1
```

frankbu: Actually, is this verify correct? It seems to be just confirming that there are 1 or more `istio_reqests_total` fields in the output, but what you want to confirm is that the added dimensions are there. Can't you just use `_verify_contains` to check that the expected dimensions are in the output?

ericvn: cc @ericvn Pointer to another way of getting stats (instead of using istioctl dashboard) in case that behaves code behaves flakey.

frankbu:
@su225 The changes that @bianpengyuan is making in #7952 should allow you to remove your python patch code and instead use `istioctl install` to add the dimensions.

su225:
@frankbu - Great to hear that. I have to admit that the script is ugly.

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
su225: @frankbu - Agreed. The verification is incorrect. I'll fix that.

su225: fixed verification and also removed ugly python script for patching.

su225:
@frankbu - I tried with default instead of demo profile and it passed all tests! Is it good to merge? Or should it be re-run just to make sure that it is not too flaky? I tried with `profile=none`, but cleanup takes a lot of time when I use `istioctl x uninstall --purge -y` - not the command itself, but the one which checks cluster snapshot after test - I see it going till `sleep 512` and `sleep 1024` which is long.

Also one more thing based on your suggestions on the other PR - can we edit this slightly to say "deploy bookinfo" and check metrics on productpage (instead of saying some `pod-name`)? This way, we can use the generated snippets in the test.

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "$snip_enable_custom_metrics_2" | istioctl install --set tag="$TAG" --set hub="$HUB" -f -
+
+kubectl get istiooperator installed-state -n istio-system -o yaml
+_wait_for_istio envoyfilter istio-system stats-filter-1.6
+_wait_for_istio envoyfilter istio-system stats-filter-1.7
+
+## Verify if patching works correctly
+send_productpage_requests
+_verify_contains snip_verify_the_results_2 "destination_port"
+_verify_contains snip_verify_the_results_2 "request_host"
```
frankbu, line 57:
```suggestion
_verify_lines snip_verify_the_results_2 "
+ destination_port
+ request_host
"
```

frankbu:
> checks cluster snapshot after test - I see it going till sleep 512 and sleep 1024 which is long

If it's taking that long it means that something has probably not been cleaned up properly. It will take about 17min total before it gives up, but then it will print out what is different.

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
```diff
@@ -0,0 +1,61 @@
...
+  for _ in {1..100}; do
+    snip_verify_the_results_1 > /dev/null
+  done
+}
+
+send_productpage_requests
+_verify_not_contains snip_verify_the_results_2 "destination_port"
+_verify_not_contains snip_verify_the_results_2 "request_host"
+
+echo "$snip_enable_custom_metrics_2" | istioctl install --set tag="$TAG" --set hub="$HUB" -f -
```
frankbu, line 48: Why is --set tag and hub needed?

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
```diff
@@ -0,0 +1,61 @@
...
+_wait_for_istio envoyfilter istio-system stats-filter-1.7
+
+## Verify if patching works correctly
+send_productpage_requests
+_verify_contains snip_verify_the_results_2 "destination_port"
+_verify_contains snip_verify_the_results_2 "request_host"
+
+# @cleanup
+set +e # ignore cleanup errors
+cleanup_bookinfo_sample
```
frankbu, line 61: If you don't use profile=none, you need to reverse the configuration change made on line 48. That's why profile=none would probably be easier to cleanup.

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
su225: I assume it would be set in Prow. Right? 

The reason why I added them - I work with my private registry (which also resides in my local machine) where I push Istio images (created from `make docker` in istio). Pulling over the internet (from gcr) is a pain. Main reason is bandwidth constraints, internet connection issues and timeouts over here. As I saw the test results, it does not cause issues in prow as well[?]  In my local environment I use `IfNotPresent` pull policy to reduce time while running `make docs.test`.

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
su225: Ok. Deleting the namespace works. But I had to clean up CRDs myself. There are others like istio CA root certificate (per namespace config map) that are not wiped out.

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
su225: Nice trick. Thanks :)

su225:
> If it's taking that long it means that something has probably not been cleaned up properly.

It seems that it happens when Istio CRDs are wiped out on cleanup. I know that there are calls to `__cluster_state` after clean up. However I'm not sure what is going on. Trace shows that it is not erroring out either.

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
su225: Done

PR #7928, opened 2020-08-12
Fix Virtual Machine Installation Doc

williamaronli:
Depends-On: https://github.com/istio/istio.io/pull/7945

Please provide a description for what this PR is for.
Fix the vm installation broken docs

ref: https://github.com/istio/istio.io/issues/7855
ref: https://github.com/istio/istio/issues/26294
And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[X] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #7945, opened 2020-08-13
Workaround code regression

sdake:
The project does not want to rebuild Istio with a change but also
does not want to ship a regression. This small workaround provides
a suitable solution until the next-generation documentation is ready
or [Ed's PR](https://github.com/istio/istio/pull/26452) merges.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -67,15 +67,12 @@ deployment you place into production.
 
     {{< text bash >}}
     $ cd "${WORK_DIR}"
-    $ make -f "${ISTIO_DIR}"/tools/certs/Makefile NAME="${CLUSTER_NAME}" NAMESPACE="${SERVICE_NAMESPACE}" "${CLUSTER_NAME}"-cacerts-selfSigned
+    $ make -f "${ISTIO_DIR}"/tools/certs/Makefile.vm NAME="${CLUSTER_NAME}" NAMESPACE="${SERVICE_NAMESPACE}" "${CLUSTER_NAME}"-certs-wl
     {{< /text >}}
 
-1. Execute the following commands to create certificates for use on the virtual machine.
-
-    {{< text bash >}}
-    $ cd "${WORK_DIR}"
-    $ make -f "${ISTIO_DIR}"/tools/certs/Makefile NAME="${CLUSTER_NAME}" NAMESPACE="${SERVICE_NAMESPACE}" "${NAMESPACE}"-certs-selfSigned
-    {{< /text >}}
+    {{< warning >}}
+    In Istio 1.6.0, the above Makefile is located in `"${ISTIO_DIR}"/samples/certs/Makefile`.
+    {{< /warning >}}
 
 ## Install the Istio control plane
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -67,15 +89,12 @@ deployment you place into production.
...
 
-1. Execute the following commands to create certificates for use on the virtual machine.
-
-    {{< text bash >}}
-    $ cd "${WORK_DIR}"
-    $ make -f "${ISTIO_DIR}"/tools/certs/Makefile NAME="${CLUSTER_NAME}" NAMESPACE="${SERVICE_NAMESPACE}" "${NAMESPACE}"-certs-selfSigned
-    {{< /text >}}
+    {{< warning >}}
+    In Istio 1.6.0, the above Makefile is located in `"${ISTIO_DIR}"/samples/certs/Makefile`.
+    {{< /warning >}}
```
frankbu, line 98: This isn't needed in 1.7 docs.

content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -49,6 +49,28 @@ but not production. Like all alpha features, this guide is subject to change.
...
+    $ tar -zxf istio-1.6.8-linux-amd64.tar.gz
+    $ popd
+    {{< /text >}}
+
+1. Set the Istio variable used within this document.
+
+    {{< text bash >}}
+    $ ISTIO_DIR="${WORK_DIR}"/istio-1.6.8
+    {{< text >}}
+
```
frankbu, line 73: Other than this section, does the rest of the doc stay the same in 1.7.1+ (i.e., just skip this part)? If not, then this whole doc is just a 1.7.0 hack, so why not just add these commands into the mainline instructions. If the rest of the doc will stay the same, then I would move this section up and put it in a warning `{{< warning >}} For release 1.7.0 only, download the Makefile from release 1.6.8 before proceeding: <download commands> {{< /warning >}}`.

frankbu:
This and #7928 seem to be a mess atm. How about just adding a single warning at the top of the doc like this:
```
{{< warning >}}
This procedure does not work with Istio 1.7.0. It will be fixed in 1.7.1 and later versions.
{{< /warning >}}
```

sdake:
> This and #7928 seem to be a mess atm. How about just adding a single warning at the top of the doc like this:
> 
> ```
> {{< warning >}}
> This procedure does not work with Istio 1.7.0. It will be fixed in 1.7.1 and later versions.
> {{< /warning >}}
> ```

This is just the 1.6.8 doc copied to master, with a few hacks. Although @esnible has fixed the makefile. I have not yet had an opportunity to test the new makefile. I will do that today. If so, the 1.7.0 docs will be much like the 1.6 docs. Certainly, #7928 should not merge as is, although that doc - once rectified - will replace this document. It isn't clear to me when that will happen, although this doc is not a hack (minus the curl stuff to pull the makefile for 1.6.8).

Cheers,
-steve

content/en/docs/setup/install/virtual-machine/index.md
sdake: I think we have it fixed in 1.7.0 given the Tuesday slip. Although I have not had a chance to verify that.

content/en/docs/setup/install/virtual-machine/index.md
sdake: wil remove.

frankbu:
> Certainly, #7928 should not merge as is, although that doc - once rectified - will replace this document.

Seems fine if the 1.6 docs work for now. As long as #7928 also works with 1.7.0, then I think it can merge (cherry-pick) anytime, although if it only works with 1.7.1+, then we'll need to keep 2 sets of docs for the rest of the 1.7 life (one for 1.7.0, and the other for 1.7.1+). That would be a little ugly.

frankbu:
I suppose that if we merge #7928 after 1.7.1, the replacement doc can have a warning at the top that says `This procedure doesn't work with 1.7.0, please upgrade to 1.7.1+`.

content/en/docs/setup/install/virtual-machine/index.md
sdake: @frankbu as sloppy as this is, we are having two makefiles in the release - approved by T&R, env, release managers, and leads meeting :) https://github.com/istio/istio/pull/26513

sdake:
@frankbu the replacement doc will work with 1.7.0, although it will take weeks to get t hat replacement doc workflow into shape IMO.

PR #7950, opened 2020-08-13
remove helm instruction and upgrade notes from 1.4

linsun:
Users should not need these as they are documented in 1.6 doc.   1.7 upgrade requires istio 1.6.

[x ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


linsun:
Expect in place upgrade section to be removed as well, waiting for confirmation on that in #release-17

linsun:
https://github.com/istio/istio/issues/26475

PR #7957, opened 2020-08-14
Upgrade doc clarification

frankbu:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/upgrade/index.md
+++ b/content/en/docs/setup/upgrade/index.md
@@ -25,11 +25,11 @@ and [upgrading from 1.4.x](#upgrading-from-1.4).
 
 To install a new revision called `canary`, you would set the `revision` field as follows:
 
-<tip>
+{{< tip >}}
 In a production environment, a better revision name would correspond to the Istio version.
 However, you must replace `.` characters in the revision name, for example, `revision=1-6-8` for Istio `1.6.8`,
 because `.` is not a valid revision name character.
-</tip>
+{{< /tip >}}
 
 {{< text bash >}}
 $ istioctl install --set revision=canary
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/upgrade/index.md
```diff
@@ -128,6 +128,14 @@ istiod-canary-55887f699c-t8bh8   1/1     Running   0          27m
 
 Note that the above instructions only removed the resources for the specified control plane revision, but not cluster-scoped resources shared with other control planes. To uninstall Istio completely, refer to the [uninstall guide](/docs/setup/install/istioctl/#uninstall-istio).
 
+### Uninstall canary control plane
+
+If you decide to rollback to the old control plane, instead of completing the canary upgrade,
+you can uninstall the canary revision using `istioctl x uninstall --revsion=canary`.
+
+However, in this case you must first reinstall the gateway(s) for the previous revision manually,
+because the uninstall command will not automatically revert the previously in-place upgraded ones.
```
linsun, line 137: Good to add using the previsous version's istioctl to install the gateway.

linsun: Also, the uninstall cmd will delete the in place upgraded one so it is critical to have the old ingress gateway installed first before issue the uninstall cmd to avoid down time.

frankbu: This sounds ridiculously complicated for such a trivial operation (rollback a canary). Is this going to be fixed in the next release?

frankbu: Added extra tip, but if this is going to be fixed in the next release, we should probably say that here too.

linsun: definitely not in 1.7.x.... perhaps in 1.8 when we can introduce canary for the gateways.

PR #7961, opened 2020-08-14
Cleanup health checking doc

frankbu:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/ops/configuration/mesh/app-health-check/index.md
```diff
@@ -15,36 +15,36 @@ test: yes
...
+This feature is enabled by default in all built-in Istio [configuration profiles](/docs/setup/additional-setup/config-profiles/)
+but can be disabled as described below.
 
-* Have a Kubernetes cluster with Istio installed, without global mutual TLS enabled.
+## Liveness and readiness probes using the command approach
 
-## Liveness and readiness probes with command option
+Istio provides a [liveness sample]({{< github_file >}}/samples/health-check/liveness-command.yaml) that
+implements this approach. To demonstrate it working with mutual TLS enabled, first configure an authentication policy
+and a destination rule:
```
linsun, line 41: Think DR is not needed any more with auto mTLS.

linsun:
LG, thanks for updating it.

linsun:
@howardjohn @myidpt what is your thoughts on leaving command option in?  I feel no one would really use it and it was there in the first place before we had http probe rewrite.

content/en/docs/ops/configuration/mesh/app-health-check/index.md
frankbu: Good point. What is this DR doing, esp. given that it has `host: "*.default.svc.cluster.local"`, so it has nothing to do with the liveness sample, running in the istio-io-health ns. I think I'll just remove it.

frankbu:
@linsun, if we decide to not document the command option, then I'd say this whole doc is unnecessary and should be removed. But, that will need to be a different PR, this one is just trying to make the current doc correct, for now.

PR #7826, opened 2020-07-31
Collecting Metrics for TCP Services automated test

ericvn:
Please provide a description for what this PR is for.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #7966, opened 2020-08-17
Automated test for Querying metrics from prometheus

su225:
Automated test for querying metrics from prometheus
https://preliminary.istio.io/latest/docs/tasks/observability/metrics/querying-metrics/

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/observability/metrics/querying-metrics/index.md
+++ b/content/en/docs/tasks/observability/metrics/querying-metrics/index.md
@@ -7,7 +7,7 @@ aliases:
     - /docs/tasks/telemetry/querying-metrics/
     - /docs/tasks/telemetry/metrics/querying-metrics/
 owner: istio/wg-policies-and-telemetry-maintainers
-test: no
+test: yes
 ---
 
 This task shows you how to query for Istio Metrics using Prometheus. As part of
--- a/content/en/docs/tasks/observability/metrics/querying-metrics/index.md
+++ b/content/en/docs/tasks/observability/metrics/querying-metrics/index.md
@@ -40,7 +40,7 @@ the example application throughout this task.
     browser or issue the following command:
 
     {{< text bash >}}
-    $ curl http://$GATEWAY_URL/productpage
+    $ curl http://"$GATEWAY_URL"/productpage
     {{< /text >}}
 
     {{< tip >}}
--- a/content/en/docs/tasks/observability/metrics/querying-metrics/index.md
+++ b/content/en/docs/tasks/observability/metrics/querying-metrics/index.md
@@ -40,7 +40,7 @@ the example application throughout this task.
     browser or issue the following command:
 
     {{< text bash >}}
-    $ curl http://"$GATEWAY_URL"/productpage
+    $ curl "http://$GATEWAY_URL/productpage"
     {{< /text >}}
 
     {{< tip >}}
--- a/content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
+++ b/content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
@@ -31,12 +31,13 @@ startup_bookinfo_sample
 kubectl apply -f samples/addons/prometheus.yaml -n istio-system
 _wait_for_deployment istio-system prometheus
 
+_verify_like snip_querying_istio_metrics_1 "$snip_querying_istio_metrics_1_out"
+
 # Fire a couple of requests
 _set_ingress_environment_variables
-export INGRESS_URL="$INGRESS_HOST:$INGRESS_PORT"
-echo "$INGRESS_URL"
+export GATEWAY_URL="$INGRESS_HOST:$INGRESS_PORT"
 for _ in {1..50}; do
-    curl -s "http://$INGRESS_URL/productpage" > /dev/null
+    snip_querying_istio_metrics_2 > /dev/null
 done
 
 # Now check Prometheus dashboard for the metric. It should be present
@@ -68,7 +69,7 @@ function query_rate_of_requests_to_productpage_5m() {
 
 # Prometheus living inside cluster is not accessible from outside.
 # So we need some sort of port forwarding mechanism
-istioctl dashboard prometheus &
+snip_querying_istio_metrics_3 &
 
 _verify_contains query_total_requests '"istio_requests_total"'
 _verify_contains query_requests_to_productpage '"productpage.default.svc.cluster.local"'
--- a/content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
+++ b/content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
@@ -31,26 +31,59 @@ startup_bookinfo_sample
 kubectl apply -f samples/addons/prometheus.yaml -n istio-system
 _wait_for_deployment istio-system prometheus
 
-# Prometheus living inside cluster is not accessible from outside.
-# So we need some sort of port redirection mechanism
-istioctl dashboard prometheus &
-
 # Fire a couple of requests
 _set_ingress_environment_variables
-export INGRESS_URL=$INGRESS_HOST:$INGRESS_PORT
-for i in {1..50}; do
-    curl -s http://$INGRESS_URL/productpage > /dev/null
+export INGRESS_URL="$INGRESS_HOST:$INGRESS_PORT"
+echo "$INGRESS_URL"
+for _ in {1..50}; do
+    curl -s -m 3.0 http://"$INGRESS_URL"/productpage > /dev/null
 done
 
 # Now check Prometheus dashboard for the metric. It should be present
-function query_prometheus() {
-    curl -sg 'http://localhost:9090/api/v1/query?query=istio_requests_total' | jq .data.result[0].metric.__name__
+function urlencode() {
+    local value=$1
+    python3 -c "import urllib.parse; print(urllib.parse.quote('''$value'''))"
 }
 
-_verify_contains query_prometheus '"istio_requests_total"'
+function query_prometheus() {
+    local query_expr=$1
+    local prometheus_api_root='localhost:9090/api/v1'
+    local encoded_query=$(urlencode "$query_expr")
+    curl -sg -m 3.0 http://"$prometheus_api_root"/query?query="$encoded_query"
+}
+
+function query_total_requests() {
+    query_prometheus 'istio_requests_total' | jq .data.result[0].metric.__name__
+}
+
+function query_requests_to_productpage() {
+    local query='istio_requests_total{destination_service="productpage.default.svc.cluster.local"}'
+    query_prometheus "$query" | jq .data.result[0].metric.destination_service
+}
+
+function query_requests_to_reviews_v3() {
+    local query='istio_requests_total{destination_service="reviews.default.svc.cluster.local",destination_version="v3"}'
+    query_prometheus "$query" | jq '.data.result[0].metric.destination_service,.data.result[0].metric.destination_version'
+}
+
+function query_rate_of_requests_to_productpage_5m() {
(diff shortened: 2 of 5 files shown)


content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# Prometheus living inside cluster is not accessible from outside.
+# So we need some sort of port forwarding mechanism
+istioctl dashboard prometheus &
+
+_verify_contains query_total_requests '"istio_requests_total"'
+_verify_contains query_requests_to_productpage '"productpage.default.svc.cluster.local"'
+_verify_contains query_requests_to_reviews_v3 '"reviews.default.svc.cluster.local"'
+# _verify_contains query_requests_to_reviews_v3 '"v3"'
+# _verify_contains query_rate_of_requests_to_productpage_5m '"productpage.default.svc.cluster.local"'
```
frankbu, line 82: Why are these commented out?

content/en/docs/tasks/observability/metrics/querying-metrics/index.md
```diff
@@ -40,7 +40,7 @@ the example application throughout this task.
     browser or issue the following command:
 
     {{< text bash >}}
-    $ curl http://$GATEWAY_URL/productpage
+    $ curl http://"$GATEWAY_URL"/productpage
```
frankbu, line 43: This works, but a nicer bash convention, IMO, is to put the quotes around the entire arg.

```suggestion
    $ curl "http://$GATEWAY_URL/productpage"
```

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
su225: It is a mistake. Forgot that.

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
```diff
@@ -0,0 +1,89 @@
...
+    curl -sg -m 3.0 http://"$prometheus_api_root"/query?query="$encoded_query"
+}
+
+function query_total_requests() {
+    query_prometheus 'istio_requests_total' | jq .data.result[0].metric.__name__
+}
+
+function query_requests_to_productpage() {
+    local query='istio_requests_total{destination_service="productpage.default.svc.cluster.local"}'
+    query_prometheus "$query" | jq .data.result[0].metric.destination_service
```
frankbu, line 61: Why aren't you using the generated snips for this and the other queries?
```suggestion
    query_prometheus "$snip_querying_istio_metrics_5" | jq '.data.result[0].metric.destination_service,.data.result[0].metric.destination_version'
```

content/zh/docs/tasks/observability/metrics/querying-metrics/index.md
```diff
@@ -33,7 +33,7 @@ aliases:
     以 Bookinfo 为例，在 web 浏览器中访问 `http://$GATEWAY_URL/productpage` 或执行如下命令：
 
     {{< text bash >}}
-    $ curl http://$GATEWAY_URL/productpage
+    $ curl http://"$GATEWAY_URL"/productpage
```
frankbu, line 36: Don't modify zh files.

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# @setup profile=demo
+kubectl label namespace default istio-injection=enabled --overwrite
+
+# Install Bookinfo sample
+startup_bookinfo_sample
+
+# Install Prometheus
+kubectl apply -f samples/addons/prometheus.yaml -n istio-system
+_wait_for_deployment istio-system prometheus
```
frankbu, line 32: Remember that the main goal of this test is to confirm that the commands in the doc are working, so you should
try to test as many generated snips as possible.

```suggestion
_wait_for_deployment istio-system prometheus

_very_like snip_querying_istio_metrics_1 "$snip_querying_istio_metrics_1_out"
```

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Fire a couple of requests
+_set_ingress_environment_variables
+export INGRESS_URL="$INGRESS_HOST:$INGRESS_PORT"
+echo "$INGRESS_URL"
+for _ in {1..50}; do
+    curl -s "http://$INGRESS_URL/productpage" > /dev/null
```
frankbu, line 39:
```suggestion
export GATEWAY_URL="$INGRESS_HOST:$INGRESS_PORT"
echo "$GATEWAY_URL"
for _ in {1..50}; do
    snip_querying_istio_metrics_2 > /dev/null
```

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
su225: Ok. Besides that I'll add some prometheus related verification as well.

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
```diff
@@ -0,0 +1,84 @@
...
+    query_prometheus "$snip_querying_istio_metrics_6" | jq '.data.result[0].metric.destination_service,.data.result[0].metric.destination_version'
+}
+
+function query_rate_of_requests_to_productpage_5m() {
+    query_prometheus "$snip_querying_istio_metrics_7" | jq '.data.result[0].metric.destination_service'
+}
+
+# Prometheus living inside cluster is not accessible from outside.
+# So we need some sort of port forwarding mechanism
+istioctl dashboard prometheus &
```
frankbu, line 71:
```suggestion
snip_querying_istio_metrics_3 &
```

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
```diff
@@ -0,0 +1,84 @@
...
+_verify_contains query_total_requests '"istio_requests_total"'
+_verify_contains query_requests_to_productpage '"productpage.default.svc.cluster.local"'
+_verify_contains query_requests_to_reviews_v3 '"reviews.default.svc.cluster.local"'
+_verify_contains query_requests_to_reviews_v3 '"v3"'
+_verify_contains query_rate_of_requests_to_productpage_5m '"productpage.default.svc.cluster.local"'
+pgrep istioctl | xargs kill
+
+# @cleanup
+set +e
+pgrep istioctl | xargs kill
```
frankbu, line 82: Doesn't `snip_cleanup_1` work? Would be better if the test could run the commands as documented, if possible.

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
frankbu: Extra verification is probably overkill, since the following snips will fail if things aren't right. The main point is that this test should be trying to test that what's documented is correct (i.e., run this command and you should see this output). If the doc is outdated and the command or output snips are a little different from what you actually get when you run them, we are trying to detect that kind of error. A side effect is we get extra testing of the istio functionality itself, but that's not the main goal for these tests.

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
su225: No. It didn't work. That was the first one I tried. Originally it was only in cleanup section. Tests didn't terminate. While looking in some other place, I found similar pattern in `tcp-metrics` test and used it. Digging into commit https://github.com/istio/istio.io/pull/7826/files/63a826b94cbff6eec4975432e6f13faef81789e8#diff-fb7f37aa500c501bd8b4ec45f80023f2 it seems that it was an issue there as well. How about changing the docs (as this does the same thing) to match snips?

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
su225: Done.

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
su225: Oh. Yeah.  Sorry. I meant the prometheus related snippet in the same task.Added it.

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
su225: Done

content/en/docs/tasks/observability/metrics/querying-metrics/test.sh
ericvn: The cleanup scripts fail because we don't have `killall` installed in the test image. We could add it, and then use the cleanup script.

What we do have in the install image is psgrep, so in my tests I resorted to using that. We could change the docs to use that and then it would be in the script and we just call the script.

However, and I haven't figured out why, we need to make the call twice else the container will fail to stop. It's something weird with our image since I can run everything from the command line and it works just fine. I suspect it's something weird in that we are running a script and that's starting the istioctl...

su225: To be honest, when I tried running commands one by one, it worked fine. But not while running automated tests. (Timing issue probably?)

ericvn: If the code works that was in the other test such that we can hit the prometheus server to verify metrics from another container (maybe we need to have a sleep container installed), so we don't use portforwarding and have to try and kill that process, that might be better. However, that makes the doc tougher since we would have to have them not use the dashboard command (or divert from the code which @frankbu rightfullysuggests is not good).

ericvn: Final answer for me would be to update the docs to use psgrep instead of killall and then revert to calling the script function twice.

ericvn: Correct. A single psgrep works fine if you are running it from the command line, but fails in the test case (either in the pipeline or local). Someone who understands the subprocesses and running in a script within a container may be able to explain why we need the second invocation in the test.

su225: > Final answer for me would be to update the docs to use psgrep instead of killall and then revert to calling the script function twice.

Cool. I'll update that.

ericvn: I thought I tried it, but we might also try doing a single call and add a `-9` to the kill in case after the first call it's trying to do a nice shutdown and the second is causing that to end anyway.

su225: > Someone who understands the subprocesses and running in a script within a container may be able to explain why we need the second invocation in the test.
Wait... But both are running in the same container (means same PID namespace). Right? On a different note, I ran into this - https://www.kernel.org/doc/Documentation/cgroup-v1/pids.txt . 

> The process number controller is used to allow a cgroup hierarchy to stop any
new tasks from being fork()'d or clone()'d after a certain limit is reached.
Do we have any such cgroup settings?

frankbu: Because of the weird issue of needing to call it twice in the test, I wouldn't update the doc yet. If it's working now, I think we should merge with the different test cleanup vs doc cleanup. We can revisit the doc cleanup instructions in a followup PR.

su225: Sure. That followup PR could clean up both tasks..

PR #7978, opened 2020-08-19
blog post for central istiod

irisdingbj:
This PR tried to summarize the central istiod single cluster and multicluster work we did with impact from users perspective, under these two design docs:

Approved: https://docs.google.com/document/d/1r58al52otp1uvB7YFHh8ifA_78hYeyd0hNLq4qPDSUM/edit

Implemented for single cluster: https://docs.google.com/document/d/1EymnzMqa7hdmKn8sEoaf1elajyp4gXEDSUavxzXB3jw/edit#heading=h.xw1gqgyqs5b

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
+---
+title: New Deployment Models for Istio
```
frankbu, line 2: This title seems pretty confusing. It's really just a variation of the shared control plane multicluster model, where the primary cluster has no user services running, just the control plane.
```suggestion
title:  An Isolated Control Plane Deployment Model
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+>}}
+
+With istio 1.7, it is possible to run Istiod on a separate cluster like the diagram below.
+
+{{< image
+    link="./central-istiod-single-cluster.jpeg"
+    caption="New deployment model for one data plane cluster"
+    >}}
+
+In this setup, the `cluster1` should be set up first with `istiodRemote` and `base` components, then istiod on the control plane cluster is installed second, with its KUBECONFIG configured to `cluster1`. With this new deployment model, mesh operators can work purely on the control plane cluster when there is a need to upgrade or reconfig istio. Platform owners and service owners can instead work solely on cluster1 to focus on their istio configs or services.
```
frankbu, line 47: I think this article needs to explain how a `control plane cluster` is different from a [primary cluster](https://preliminary.istio.io/latest/docs/reference/glossary/#primary-cluster). It fits the definition of a primary cluster, just one that has nothing else other than the control plane running in it. To avoid confusion, I think this needs to explain what's different in the implementation for this case. If there is a fundamental difference, then we should consider adding another term to the glossary for `control plane cluster: a primary cluster that runs a control plane only`, or something like that.

irisdingbj: Thanks @frankbu for the review!  I am thinking we may need a new terminology in this blog or in https://preliminary.istio.io/latest/docs/reference/glossary/#primary-cluster.

@linsun WDYT?

irisdingbj:
@linsun @Rachael-Graham ^^

content/en/blog/2020/new-deployment-model/index.md
linsun: It is actually not just a variation of the shared control plane multicluster model.  This is a single cluster model but istiod runs outside of the cluster for easier management by another team/provider.

content/en/blog/2020/new-deployment-model/index.md
linsun: I would not call this primary cluster for multicluster.  Primary cluster is a concept used within multicluster.  Here, we start with single cluster.  The control plane cluster is the cluster that runs one or more Istiod control planes.

content/en/blog/2020/new-deployment-model/index.md
linsun: Another title suggestion would be "Introduce Central Istiod where Istiod is managed separately"

frankbu:
This blog relates to Istio deployment terminology and recommended patterns. Requesting signoff from @istio/steering-committee

content/en/blog/2020/new-deployment-model/index.md
frankbu: The problem is that it fits the current definition of a primary cluster "a primary cluster is a cluster with a control plane". We'll need to clarify that definition and this article needs to be clear how a "control plane cluster" is different. How, for example, do you achieve HA control planes? Where does all this fit into the [deployment guide](https://preliminary.istio.io/latest/docs/ops/deployment/deployment-models/)?

cc @smawson @nmittler

craigbox:
From a quick skim, I think this this makes Istio seem much more complicated, rather than "less complicated" as we're trying to make it seem.

I'd love to do a pass and see if I can help.  What is the source format for the images?  There's at least typos in them, and I'd propose some reorganising.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+subtitle: Clear Separation between Mesh Operator and Mesh Users
+description: A new deployment model for Istio.
+publishdate: 2020-08-19
+attribution: "Lin Sun (IBM), Iris Ding (IBM)"
+keywords: [multicluster,install,deploy,'1.7']
+---
+
+## Background
+
+In working with various service mesh users and vendors, we believe there are 3 key personas for a typical service mesh:
```
rvennam, line 12:
```suggestion
From experience working with various service mesh users and vendors, we believe there are 3 key personas for a typical service mesh:
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+* Mesh Operator
+* Platform Owner
```
rvennam, line 14: Unclear what Platform Owner is responsible for. The rest of the blog groups platform owner and service owner persons together. Suggest to clarify what the roles of each are, or just remove one of Platform or Service

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+It is common to have all these personas work on the same clusters without any clear separation.  For example, there are multiple ways to deploy Istio according to the docs, which all started with mesh operator, platform owner and service owner sharing the single cluster first and then gradually expanding the mesh to multiple clusters or VMs.  None of these provided a clear separation between mesh operator and platform/service owner at the boundary of a cluster.  You may be thinking you could set up [Kubernetes RBAC rules](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) and namespaces to control which personas can do what within the cluster, however, sometimes you need stronger separation between the istio control plane and the rest at the cluster level.  Introduce a new deployment model for Istio which enables mesh operators to install and manage mesh control plane on dedicated clusters, separated from the data plane clusters.  This new deployment model could easily enable Istio vendors to run Istio control plane for service mesh users while users can focus on their workloads and Istio resources without worrying about installing or managing the Istio control plane.
```
rvennam, line 17:
```suggestion
It is common to have all these personas work on the same clusters without any clear separation.  For example, there are multiple ways to deploy Istio according to the docs, which all start with mesh operator, platform owner and service owner sharing the single cluster first and then gradually expanding the mesh to multiple clusters or VMs.  None of these provided a clear separation between mesh operator and platform/service owner at the boundary of a cluster.  You may be thinking you could set up [Kubernetes RBAC rules](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) and namespaces to control which personas can do what within the cluster, however, sometimes you need stronger separation between the Istio control plane and the rest at the cluster level.  We're introducing a new deployment model for Istio which enables mesh operators to install and manage the mesh control plane on dedicated clusters, separated from the data plane clusters.  This new deployment model could easily enable Istio vendors to run Istio control plane for service mesh users while users can focus on their workloads and Istio resources without worrying about installing or managing the Istio control plane.
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+## New Deployment model
+
+Let’s take a close look of what has been deployed in your cluster if you install Istio following the default install:
```
rvennam, line 22:
```suggestion
Let’s take a close look of what gets deployed in your cluster if you install Istio following the default install:
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+    * Secrets: stores key and certs and credentials related to service account or Istiod.
+
+Below diagram shows how resources are deployed in the cluster after following the [default install](https://istio.io/latest/docs/setup/install/istioctl/#install-istio-using-the-default-profile).
+
+{{< image
+    link="./default-install.jpeg"
+    caption="default deployment model for istio"
+>}}
+
+With istio 1.7, it is possible to run Istiod on a separate cluster like the diagram below.
```
rvennam, line 40:
```suggestion
With the new `istiodRemote` component in Istio 1.7, it is possible to run Istiod on a separate cluster as shown in the diagram below.
```

content/en/blog/2020/new-deployment-model/index.md
rvennam, line 47:
```suggestion
In this setup, the `cluster1` should be set up first with `istiodRemote` and `base` components. Then, istiod on the control plane cluster is installed second, with its KUBECONFIG configured to `cluster1`. With this new deployment model, mesh operators can work purely on the control plane cluster when there is a need to upgrade or re-configure Istio. Platform owners and service owners can instead work solely on `cluster1` to focus on their Istio configs or services.  
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+
+{{< image
+    link="./central-istiod-single-cluster.jpeg"
+    caption="New deployment model for one data plane cluster"
+    >}}
+
+In this setup, the `cluster1` should be set up first with `istiodRemote` and `base` components, then istiod on the control plane cluster is installed second, with its KUBECONFIG configured to `cluster1`. With this new deployment model, mesh operators can work purely on the control plane cluster when there is a need to upgrade or reconfig istio. Platform owners and service owners can instead work solely on cluster1 to focus on their istio configs or services.  
+
+To support this, we introduced a new `istiodRemote` component in Isito 1.7. You can see now in cluster1, it only has istio configs and services that belong to data plane. Istiod runs in a separate cluster(control plane cluster) and is pointing its KUBECONFIG to cluster1.  Some sample snippet you can reference to install Istio on cluster1:
```
rvennam, line 49:
```suggestion
To support this, we introduced a new `istiodRemote` component in Isito 1.7. `cluster1`only has Istio configs and services that belong to data plane. Istiod runs in a separate cluster (control plane cluster) and is pointing its KUBECONFIG to `cluster1`.  Here is a sample snippet you can reference to install Istio on cluster1:
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+      enabled: true
+values:
+  global:
+    operatorManageWebhooks: true
+
+{{< /text >}}
+
+If you are interested in exploring this, you can follow the [central istiod single cluster step by step guide.](https://github.com/istio/istio/wiki/Central-Istiod-single-cluster-steps)
+
+The diagram above shows a single cluster as data plane for an Istio mesh, however, you may expand the setup to multiple clusters as data plane where these multiple clusters are managed by Istiod running on the control plane cluster and are getting configs from the config cluster per diagram below.
```
rvennam, line 91:
```suggestion
The diagram above shows a single cluster as the data plane for an Istio mesh. However, you may expand your data plane to multiple clusters, all managed by the same Istiod running on the control plane cluster. All of the data plane clusters will receive their configs from the config cluster per diagram below. 
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+
+{{< /text >}}
+
+If you are interested in exploring this, you can follow the [central istiod single cluster step by step guide.](https://github.com/istio/istio/wiki/Central-Istiod-single-cluster-steps)
+
+The diagram above shows a single cluster as data plane for an Istio mesh, however, you may expand the setup to multiple clusters as data plane where these multiple clusters are managed by Istiod running on the control plane cluster and are getting configs from the config cluster per diagram below. 
+
+{{< image width="100%" link="./central-istiod-multi-cluster.jpeg" caption="New deployment model for multi data plane clusters" >}}
+
+You can further expand this deployment model to manage multiple istio mesh from a centralized control plane cluster that runs multiple Istiods, per diagram below:
```
rvennam, line 95:
```suggestion
You can further expand this deployment model to manage multiple Istio meshes from a centralized control plane cluster that runs multiple Istiods, per diagram below:
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+
+{{< image width="100%" link="./central-istiod-multi-mesh.jpeg" caption="New deployment model for multi mesh" >}}
+
+Control plane cluster can be used to host multiple Istiods and each Istiod manages his own data plane remotely.  In this model we can install Istio into the control plane cluster and use Istio ingress gateway and virtual services to route traffic between different istiod instances.
```
rvennam, line 99:
```suggestion
Control plane cluster can be used to host multiple Istiods and each Istiod manages its own remote data plane.  In this model we can install Istio into the control plane cluster and use Istio ingress gateway and virtual services to route traffic between different Istiod instances.
```

content/en/blog/2020/new-deployment-model/index.md
rvennam: Another suggestion "A Deployment Model for Running the Istio Control Plane Outside of Your Workload Cluster"

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,103 @@
...
+
+{{< image
+    link="./default-install.jpeg"
+    caption="default deployment model for istio"
+>}}
+
+With istio 1.7, it is possible to run Istiod on a separate cluster like the diagram below.
+
+{{< image
+    link="./central-istiod-single-cluster.jpeg"
```
rvennam, line 43: I think this diagram exposes too much lower level complexity for this blog post. I suggest making a new simpler diagram - I'm happy to draw one.

content/en/blog/2020/new-deployment-model/index.md
frankbu: @rvennam That sounds more like a description than a title :) But it is getting to the point. All of the "workloads" are running in what Istio terminology calls [remote clusters](https://preliminary.istio.io/latest/docs/reference/glossary/#remote-cluster). The control plane is running somewhere else. I'm assuming that in such a configuration, istiod wouldn't need to be running in a kubernetes cluster, it just happens to be in this example.

I think that this article needs to align with Istio terminology (e.g., use, and reference, the glossary terms). I also think that we need to describe this model in the current docs by updating [this section of the deployment guide](https://preliminary.istio.io/latest/docs/ops/deployment/deployment-models/#control-plane-models). It should include a picture and description of this "all remote clusters" model. Then the blog can be presented as an example of this model, one where the control plane is running in another kubernetes cluster as opposed to in one of the mesh/workload clusters (aka, a primary cluster).

content/en/blog/2020/new-deployment-model/index.md
linsun: Platform owner is the overall owner of all services and their running platform, who controls gateway resources, host/domain/ports.  service owner is owner for individual service.

content/en/blog/2020/new-deployment-model/index.md
linsun: Agreed, I'll rework based on the diagrams from deployment model.

linsun:
@rvennam @frankbu @craigbox thank you for your valuable comments and contributions.  I agree with the diagram comments which I will rework based on existing diagrams from deployment models.

I'm thinking to change control plane cluster to istiod cluster to avoid confusion with existing multicluster concepts as this applies to single cluster as well.  WDYT?

linsun:
> From a quick skim, I think this this makes Istio seem much more complicated, rather than "less complicated" as we're trying to make it seem.
> 
> I'd love to do a pass and see if I can help. What is the source format for the images? There's at least typos in them, and I'd propose some reorganising.

@craigbox feel free to share your ideas or propose changes!

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+You may further expand your data plane to multiple clusters, all managed by the same Istiod running on the central control plane cluster. All of the data plane clusters will receive their configs from the primary cluster per diagram below. 
+{{< image width="100%" link="./central-istiod-multi-cluster.jpeg" caption="New deployment model for multi data plane clusters" >}}
+
+
+## Conclusion
+
+This new deployment model enables the Istio control plane to be managed by others who have operation expertise in Istio and are able to keep up with the constant innovation of the Istio project.  We believe this deployment model will provide the best experience to platform/service owners where they can focus on architecturing the deployment or configuration of their services or the core business logic of their services.  While Istio has been perceived as complex by some users, the Istio community will continue working on making Istio simpler so users can focus on their core business.
```
rvennam, line 54:
```suggestion
This new deployment model enables the Istio control plane to be run and managed by others who have the operational expertise in Istio.  This reduces the complexity and provides the best experience to platform/service owners by allowing them to focus on the core business logic of their applications and the configuration of their services. 
```

linsun:
FYI image sources are in https://drive.google.com/drive/folders/1jczscJueUBR3IOvH30q9HAZX0jC7GSyW, following our diagram guideline for istio.io.

linsun:
@rvennam @frankbu @craigbox I've reworked based on your comments, PTAL.  Key changes:
- Incorporated Ram's editing.
- Simplify the diagrams
- Simplify the contents of the blog by removing a lot of details.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Background
+
+From experience working with various service mesh users and vendors, we believe there are 3 key personas for a typical service mesh:
+
+* Mesh Operator, who manages the service mesh installation and upgrade.
+
+* Mesh Users:
+
+  1. Platform Owner, who owns the service mesh platform and defines the overall strategy and implementation for service owners to adopt service mesh.
```
craigbox, line 18: I'm not sure this breakdown is right. There are definitely "platform owners", and "service owners".  Is a "mesh operator" different to a platform owner?  I wouldn't call them a mesh user, I'd say they are the team that manages Kubernetes and makes it available to others.

I tend to break personas down as: platform owner, service operator, app developer.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
...rsonas work on the same clusters without any clear separation.  For example, there are multiple ways to deploy Istio according to the [docs](/docs/setup/install/), which all start with mesh operator, platform owner and service owner sharing the single cluster first and then gradually expanding the mesh to multiple clusters or VMs.  None of these provided a clear separation between mesh operator and platform/service owner at the boundary of a cluster.  You may be thinking you could set up [Kubernetes RBAC rules](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) and namespaces to control which personas can do what within the cluster, however, sometimes you need stronger separation between the Istio control plane and the rest at the cluster level.  We're introducing a new deployment model for Istiod which enables mesh operators to install and manage the mesh control plane on dedicated clusters, separated from the data plane clusters.  This new deployment model could easily enable Istio vendors to run Istio control plane for service mesh users while users can focus on their workloads and Istio resources without worrying about installing or managing the Istio control plane.
```
craigbox, line 21: Is the goal here the strength of the separation, or is it the administration moving from one place to another?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## New Deployment model
```
craigbox, line 23: Central cluster deployment model?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+{{< image width="100%"
+    link="single-cluster.svg"
+    alt="Istio mesh in a single cluster"
+    title="Single cluster Istio mesh"
+    caption="Istio mesh in a single cluster"
+    >}}
+
+In Istio 1.7, it is possible to run Istiod on a separate, dedicated cluster (central control plane cluster) as shown in the diagram below. With this new deployment model, mesh operators can work purely on the central control plane cluster when there is a need to upgrade or re-configure Istio. Platform/service owners will work solely on the data plane cluster to deploy their applications and apply any Istio resources or configurations. They don't have direct access to the central control plane cluster to deploy anything there, neither can the central control plane cluster process any Istio resources in that cluster. If you are interested in exploring this, you can follow the [central istiod single cluster step by step guide](https://github.com/istio/istio/wiki/Central-Istiod-single-cluster-steps).
```
craigbox, line 34:
```suggestion
From Istio 1.7, it is possible to run Istiod on a separate, dedicated cluster (the "central control plane cluster") as shown in the diagram below. With this new deployment model, mesh operators can work purely on the central control plane cluster when there is a need to upgrade or re-configure Istio. Platform/service owners work solely on the data plane cluster to deploy their applications and apply any Istio resources or configurations. They don't have direct access to the central control plane cluster to deploy anything there, neither can the central control plane cluster process any Istio resources in that cluster. If you are interested in exploring this, you can follow the [central istiod single cluster step by step guide](https://github.com/istio/istio/wiki/Central-Istiod-single-cluster-steps).
```

I'm also not sure what "neither can the central control plane cluster process any Istio resources in that cluster. " means.  Central cluster can't process resources in the data plane cluster?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+The diagram above shows a single cluster as the data plane for an Istio mesh. However, you can expand this deployment model to manage multiple Istio meshes from a central control plane cluster that runs multiple Istiod control planes:
+
+{{< image width="100%"
+    link="multiple-central-Istiods.svg"
+    alt="Istio meshes in single clusters with Istiod outside"
+    title="Multiple single clusters Istio meshes with Istiod outside"
+    caption="Multiple single clusters with multiple Istiod control planes in a central control plane cluster"
+    >}}
+
+Central control plane cluster can be used to host multiple Istiod control planes and each Istiod manages its own data plane clusters. In this model we can leverage Istio mesh in the central control plane cluster and use `istio-ingress` gateway to route traffic between different Istiod control planes.
```
craigbox, line 52: > In this model we can leverage Istio mesh in the central control plane cluster and use `istio-ingress` gateway to route traffic between different Istiod control planes.

I don't follow why you would do this? Aren't they multi-tenant, so there should be no connection between them? It's not reflected on the diagram.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+You may further expand your data plane to multiple clusters, which are managed by the same Istiod running on the central control plane cluster. All of the data plane clusters will receive their configurations from the primary cluster per the diagram below.
```
craigbox, line 54: we would normally say "expand your mesh to multiple clusters", I guess saying "expand data plane" makes sense when the mesh is already across multiple clusters in this case.

However, I don't think the primary cluster sentence is correct: all the clusters (at least according to the picture) receive their configurations from the central control plane cluster?

content/en/blog/2020/new-deployment-model/index.md
craigbox: (Also, we could strengthen the message about this being useful for vendors)

content/en/blog/2020/new-deployment-model/index.md
linsun: Central control plane cluster can't process resources in the central cluster as it processes resource in the data plane cluster.

content/en/blog/2020/new-deployment-model/index.md
linsun: Basically, leverage istio (another mesh not in the diagram to avoid confusion) to expose different istiod control planes to the outside of the cluster.

content/en/blog/2020/new-deployment-model/index.md
linsun: So the Istio configs do come up the primary cluster, as that is the cluster mesh users can deploy istio resources.  However, the configs are received from the central control plane cluster.

content/en/blog/2020/new-deployment-model/index.md
nmittler: @linsun @frankbu @smawson `primary cluster` is not specifically a multicluster concept, it's simply a cluster with a control plane. Adding another, similar term here will simply cause confusion.  I think before this change lands, we should first agree on a set up updated glossary terms.

@linsun this all goes back to a discussion that we had around central istiod. We need to break it down into specific concepts. To get the discussion rolling, I propose the following:

- Primary Cluster: a cluster with a control plane (no change from current definition).
- Config Cluster: a cluster that provides configuration for one or more Istio control planes.

WRT the use of `central istiod`, this seems to be in conflict with the existing term [Managed Control Plane](https://istio.io/latest/docs/reference/glossary/#managed-control-plane). I think we either need to abandon the use of `central istiod` or discuss its use and how it's different from `managed control plane` with the TOC.

content/en/blog/2020/new-deployment-model/index.md
linsun: The goal here is to separate control plane out and it will be upgraded by a different team/vendor so mesh users/admins don't have to worry about it.

content/en/blog/2020/new-deployment-model/index.md
linsun: I'll run this through the env WG and TOC if needed.  This blog can be easily updated by changing central to managed if we prefer.  In the last week's env wg mtg, seems most folks prefer to use central.

content/en/blog/2020/new-deployment-model/index.md
linsun: Updated, PTAL - hope it is more clear now.

content/en/blog/2020/new-deployment-model/index.md
nmittler: Sounds good, @linsun ... thanks!

The issue with `central` as I see it is that it's simply in conflict with `managed`, so it's less about individual opinions and more about how to reconcile the two.

What are your thoughts on using `primary cluster`?

linsun: Primary cluster in multicluster refers to the control plane cluster in general, however in central/managed istiod env, since control plane no longer runs in user's clusters, primary cluster refers to primary data plane cluster.

WDYT @nmittler ?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
...ing to the [docs](/docs/setup/install/), which all start with mesh operator, platform owner and service owner sharing the single cluster first and then gradually expanding the mesh to multiple clusters or VMs.  None of these provided a clear separation between mesh operator and platform/service owner at the boundary of a cluster.  You may be thinking you could set up [Kubernetes RBAC rules](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) and namespaces to control which personas can do what within the cluster, however, sometimes you need stronger separation between the Istio control plane and the rest at the cluster level.  We're introducing a new deployment model for Istiod which enables mesh operators to install and manage the mesh control plane on dedicated clusters, separated from the data plane clusters.  This new deployment model could easily enable Istio vendors to run Istio control plane for service mesh users while users can focus on their workloads and Istio resources without worrying about installing or managing the Istio control plane. This model enables vendors to build managed Istio offerings that automatically upgrade Istio control plane for mesh users.
```
smawson, line 21: I think you can reword this a bit to be less disparaging of other deployment models. We're introducing a new option, we're not recommending against the existing options. You could maybe drop a bunch of this paragraph and just say that we're introducing a new model to allow a separation between the control plane operator and the mesh admin, which allows the mesh control plane to run as a service, while still giving the mesh admin control over the configuration of the control plane.

I wouldn't even discuss service owners here since they shouldn't care right?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Central Cluster Deployment model
+
+After installing Istio using the [default installation profile](/docs/setup/install/istioctl/#install-istio-using-the-default-profile), you can deploy your services to the mesh, like the diagram below:
```
smawson, line 25: I would say something about how the default installation of Istio uses a single cluster for the control plane and the services, and then show the diagram. I wouldn't talk about deploying services since it isn't really relevant.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+{{< image width="100%"
+    link="single-cluster.svg"
+    alt="Istio mesh in a single cluster"
+    title="Single cluster Istio mesh"
+    caption="Istio mesh in a single cluster"
+    >}}
+
+From Istio 1.7, it is possible to run Istiod on a separate, dedicated cluster (the "central control plane cluster") as shown in the diagram below. With this new deployment model, mesh operators can work purely on the central control plane cluster when there is a need to upgrade or re-configure Istio. Platform/service owners work solely on the data plane cluster to deploy their applications and apply any Istio resources or configurations. They don't have direct access to the central control plane cluster to deploy anything there, neither can the central control plane cluster process any Istio resources in that cluster. If you are interested in exploring this, you can follow the [central istiod single cluster step by step guide](https://github.com/istio/istio/wiki/Central-Istiod-single-cluster-steps).
```
howardjohn, line 34: Do we want to say this is a "dedicated cluster"? It need not be a "cluster" - the key point is its not the main cluster. 

Part of my concern is the description is too narrow but I also don't want people thinking "Istio is so complicated just to run a simple mesh you need two clusters!"

smawson, line 34: Say something about how with this new deployment model, the control plane can run in a separate cluster from the services. The control plane cluster is owned by the mesh operator, and applications consume the control plane as a service.

The stuff about deploying to each others clusters is confusing, maybe drop that or try to clean it up to make it obvious why it is important.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+{{< image width="100%"
+    link="single-cluster-central-Istiod.svg"
+    alt="Istio mesh in a single cluster with Istiod outside"
+    title="Single cluster Istio mesh with Istiod outside"
+    caption="Single cluster Istio mesh with Istiod in a central control plane cluster"
+    >}}
+
+The diagram above shows a single cluster as the data plane for an Istio mesh. However, you can expand this deployment model to manage multiple Istio meshes from a central control plane cluster that runs multiple Istiod control planes:
```
smawson, line 43: I think you should try to be clear about your audience. Here you're talking to the mesh operator, and telling them that they can run multiple control plane instances in a single cluster.

It may be worthwhile to really target this blog post towards either mesh operators or mesh admins, or split into sections, one for each. Otherwise it is confusing who you're talking to and what you're trying to tell them.

For example, to the mesh operator, you want them to know:
- They can run the control plane in their own clusters, and provide it as a service to mesh admins.
- They can manage the lifecycle of the control plane (?)
- They can run multiple control planes in a single cluster.
- Mesh admins don't need access to their cluster.
- They can use their own istio mesh to manage their control plane clusters, using ingress to control access.

For mesh admins:
- They can use a control plane as a service provided by a mesh operator
- They can still configure mesh-wide settings using mesh config
- They can control when they upgrade to a new control plane version (?)

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+{{< image width="100%"
+    link="multiple-central-Istiods.svg"
+    alt="Istio meshes in single clusters with Istiod outside"
+    title="Multiple single clusters Istio meshes with Istiod outside"
+    caption="Multiple single clusters with multiple Istiod control planes in a central control plane cluster"
+    >}}
+
+Central control plane cluster can be used to host multiple Istiod control planes and each Istiod manages its own data plane clusters. In this model vendors could install another Istio mesh in the central control plane cluster and configure its `istio-ingress` gateway to route traffic from a data plane cluster to its corresponding Istiod control plane.
+
+You may further expand your data plane to multiple clusters, which are managed by the same Istiod running on the central control plane cluster. All of the data plane clusters will receive their configurations from the central control plane cluster per the diagram below.
```
smawson, line 54: You should describe the difference between primary and remote cluster here, where primary just means that is where the config for the mesh comes from? Otherwise I don't understand why one is labelled primary and the other remote since they look identical.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+{{< image width="100%"
+    link="multiple-clusters-central-Istiod.svg"
+    title="Multicluster Istio mesh with Istiod outside"
+    caption="Multicluster Istio mesh with Istiod in a central control plane cluster"
+    >}}
+
+## Conclusion
+
+This new deployment model enables the Istio control plane to be run and managed by others who have the operational expertise in Istio.  This reduces the complexity and provides the best experience to platform/service owners by allowing them to focus on the core business logic of their applications and the configuration of their services.
```
smawson, line 64:
```suggestion
This new deployment model enables the Istio control plane to be run and managed by others who have operational expertise in Istio.  This reduces complexity and provides the best experience for platform/service owners by allowing them to focus on the core business logic of their applications and the configuration of their services.
```

content/en/blog/2020/new-deployment-model/index.md
sdake: @linsun no need to run this through env WG. Our collective decision as I recall it was to let it be named what it is named with of course TOC oversight. My takeaway was the original  is already in place, and is such is sticky.

content/en/blog/2020/new-deployment-model/index.md
smawson, line 64: Here I would summarize benefits to the operator and the mesh admin (and again leave service owners out of it).

This really lets mesh operators be separate from mesh admins, allowing vendors to offer Istio control plane as a service, which was not really possible without this work. So for vendors this lets them provide a new offering, and for mesh admins it lets them use a managed service instead of operating the control plane themselves.

content/en/blog/2020/new-deployment-model/index.md
frankbu: I wouldn't use "central" because central brings to mind "single point of failure" which is not a good thing. It also brings up the question of how do you get control plane HA with this model if istiod is running in only one cluster?

As I said above, I also still think that this blog shouldn't be introducing new terminology, but instead introducing the example and referring to diagrams and terms that are added to the Deployment Models doc, not put them in the blog itself. The blog can be a very short intro and just point to the model it's implementing, instead of confusing readers by describing things that Istio docs don't mention.

Also, I don't think `primary cluster` can be used to mean `primary data plane cluster` in the managed/central istiod model, because primary -> control plane. None of the mesh clusters have a control plane so there are zero primary clusters in this model. All the clusters are `remote clusters`.

I think the fundamental difference between a `primary cluster` and a `control plane cluster` is that the a primary cluster is a "mesh cluster" with a control plane, while a control plane cluster is just a "plain K8s cluster" with a control plane.

I think this should all be described clearly in the deployment models doc, not in a blog.

content/en/blog/2020/new-deployment-model/index.md
frankbu: What is the purpose of describing 3 different roles in this blog? Other than the fact that there are mesh operators vs. others (2 things to separate), this deployment model doesn't support any further breakdown of roles, so describing the sub roles for configuring vs implementing services on the data-plane side has nothing to do with what this blog is providing. Or am I missing something?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,64 @@
...
+
```
frankbu, line 22: I think the last sentence is the only important one. Who else, other than managed Istio providers would configure Istio this way. IMO, a simpler way to describe this whole thing is a way to run a `managed control plane` in a Kubernetes cluster (i.e., `control plane cluster : a Kubernetes cluster running a managed control plane`). Talking about a `Mesh Operator` persona managing the installation and upgrade of Istio is just using new words to say that the control plane is managed separately (a `managed control plane`). Why make it more complicated than that?

content/en/blog/2020/new-deployment-model/index.md
linsun: You are right, the benefit introduced by this blog doesn't differ much for the 2 sub roles (under mesh admin/user).  It is described here for completeness of common mesh personas.

content/en/blog/2020/new-deployment-model/index.md
linsun: updated... of course, pending new naming under votes in environment wg now.

content/en/blog/2020/new-deployment-model/index.md
linsun: I'll remove dedicated.  It doesn't have to be dedicated.

content/en/blog/2020/new-deployment-model/index.md
linsun: Thanks, this is good feedback, I have been thinking to separate the roles in the diagram as well.  I'll rework on the split.

content/en/blog/2020/new-deployment-model/index.md
linsun: Some feedback on naming from https://istio.slack.com/archives/C6KA8TTSS/p1598301755064400 on environments slack channel is that managed control plane often indicated run by cloud provider... and it is too narrow.  It doesn't have to be cloud provider, it can be a separate team focusing on operating istio control plane.

content/en/blog/2020/new-deployment-model/index.md
linsun: @frankbu I like your comment about not introducing new terminology and make this simple to focus on user cases.   I agree with your difference about `primary cluster` vs `control plane cluster` as well.

Are you suggesting we update deployment model first before releasing a blog?

content/en/blog/2020/new-deployment-model/index.md
linsun: the key for service owner role is their experience with Istio mesh stay the same, whether they are using central istiod or not.

linsun:
Thanks to all for your feedback, I highly appreciate them!  I have done another pass to resolve all of them, except a few related to naming that I'll lean on @frankbu to create a PR to add to istio terminology.  Summary of changes:
- clarify targeted users and  associated benefits throughout the blog, including diagrams.
- removed central everywhere, and used control plane cluster
- removed word dedicated

PTAL and let me know any other feedback you have. @smawson @nmittler @howardjohn @craigbox

Help I need from @frankbu 's PR on namings:
- control plane cluster - any other term should i use?
- I've used remote cluster, and data plane cluster, primary data plane cluster in this blog and diagrams.  Would be good to have these consistent with the terminologies you are building.

content/en/blog/2020/new-deployment-model/index.md
linsun: Thanks!  clarified - PTAL

content/en/blog/2020/new-deployment-model/index.md
linsun: Thanks - Clarified!  PTAL.  I left your two question bullets out as that is out of scope for this blog.

content/en/blog/2020/new-deployment-model/index.md
linsun: Hopefully this is clarified now with the diagram updated for each personas and I dropped the portion about "deploying to each others cluster".  PTAL

content/en/blog/2020/new-deployment-model/index.md
linsun: Updated.  I talked about deploying services as the diagram had services.  But I agree, there is no change to that experience so not worthy highlighting.

content/en/blog/2020/new-deployment-model/index.md
linsun: Thanks, I updated this to be 2 paragraphs, 1 focuses on where we are today, and 2nd focuses on introducing the new model.  service owners are mentioned solely here only to let them know their experience is unchanged.

content/en/blog/2020/new-deployment-model/index.md
linsun: This is updated as well, PTAL

content/en/blog/2020/new-deployment-model/index.md
nmittler: To be clear, `primary cluster` does not imply anything about where user workloads run. It simply implies that the control plane runs in *that* cluster.

content/en/blog/2020/new-deployment-model/index.md
nmittler: > I think the fundamental difference between a primary cluster and a control plane cluster is that the a primary cluster is a "mesh cluster" with a control plane, while a control plane cluster is just a "plain K8s cluster" with a control plane.

I strongly disagree. The intent for `primary cluster` is simply to suggest that it is a cluster where the control plane runs. It has no distinction of whether user workloads run there or not. It's just a cluster that runs a control plane. You can certainly have a dedicated primary cluster that runs no user workloads ... that's fine. Adding `control plane cluster` will be a huge problem for terminology ... it will simply be confused with `primary cluster` in every conversation. In fact, when we decided on the term `primary cluster`, `control plane cluster` was one of the candidate terms. We're better off not introducing a new term for this and simply refer to it as a `primary cluster` that runs outside of the mesh.

frankbu: The problem is that implementation wise, Istiod deployed in a primary cluster is different than Istiod deployed as an external control plane. It's not just the same instructions, deploy Istio and then either do or don't run any services next to it. I'm not dwelling on the distinction in my proposed doc updates. PTAL at #8015

nmittler: I still don't think that's a valid distinction. The configuration difference IIUC is to enable the control plane to use another cluster as a source for config. That is orthogonal to whether or not the control plane is external. Theoretically we could use an in-mesh cluster configured to use another cluster as its source of config.

content/en/blog/2020/new-deployment-model/index.md
smawson: I think you can greatly simplify this background:

Istio currently requires the control plane to run in one of the clusters in the mesh, leading to a lack of separation between the mesh operator and the mesh admin.

Also the final paragraph is not background, its an overview of what the blog is about.

Maybe just rename this section to overview and delete all but the last paragraph, and introduce the concepts of mesh operator and mesh admin in that paragraph.

We're introducing a new deployment model for Istio which enables a separation of the role of a mesh operator (manages the mesh control plane installation and operation) and mesh admin (manages the mesh configuration and membership). This new deployment model allows istiod to be used as an external control plane (link to glossary entry).

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,67 @@
...
+We're introducing a new deployment model for Istiod which enables mesh operators to install and manage the mesh control plane on separate clusters. This new deployment model allows a clear separation between mesh operators and mesh admins. With this new model, Istio mesh operators can run Istio control plane for mesh admins while mesh admins can still control the configuration of the control plane without worrying about installing or managing the Istio control plane. This model is transparent to mesh users.
+
+## Control Plane Cluster Deployment model
+
+After installing Istio using the [default installation profile](/docs/setup/install/istioctl/#install-istio-using-the-default-profile), you have Istiod control plane installed in a single cluster like the diagram below:
```
smawson, line 26:
```suggestion
After installing Istio using the [default installation profile](/docs/setup/install/istioctl/#install-istio-using-the-default-profile), you will have an Istiod control plane installed in a single cluster like the diagram below:
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+{{< image width="100%"
+    link="single-cluster.svg"
+    alt="Istio mesh in a single cluster"
+    title="Single cluster Istio mesh"
+    caption="Istio mesh in a single cluster"
+    >}}
+
+With this new deployment model in Istio 1.7, it is possible to run Istiod on a separate cluster (the `control plane cluster`) from services, as shown in the diagram below. The `control plane cluster` is owned by mesh operator while the mesh admin owns the cluster that has services deployed in the mesh. Mesh admin has no access to the `control plane cluster`. If you are interested in exploring this, you can follow the [central istiod single cluster step by step guide](https://github.com/istio/istio/wiki/Central-Istiod-single-cluster-steps).
```
smawson, line 35: Reword to say that we can run istiod as an external control plane, for example running it in a cluster owned by the mesh operator. We can drop any mention of control plane cluster or primary cluster or whatever. Just say that the mesh operator owns the control plane, and the mesh admin has no access to it.

Then you can link to an example of how to run istiod as an external control plane for mesh operators to view.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+{{< image width="100%"
+    link="single-cluster-central-Istiod.svg"
+    alt="Istio mesh in a single cluster with Istiod outside"
+    title="Single cluster Istio mesh with Istiod outside"
+    caption="Single cluster Istio mesh with Istiod in a control plane cluster"
+    >}}
+
+Mesh admins can expand the service mesh to multiple clusters, which are managed by the same Istiod running in the `control plane cluster`. The `primary data plane cluster` also serves as the Istio configuration cluster, in addition to run services. The `control plane cluster` reads Istio configurations from the `primary data plane cluster` and pushes them to all of the data plane clusters per the diagram below.
```
smawson, line 44: Replace primary data plane cluster with config cluster, and say that the external control plane reads its configuration from the config cluster.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,67 @@
...
+    alt="Istio meshes in single clusters with Istiod outside"
+    title="Multiple single clusters Istio meshes with Istiod outside"
+    caption="Multiple single clusters with multiple Istiod control planes in a control plane cluster"
+    >}}
+
+The `control plane cluster` can be used to host multiple Istiod control planes and each Istiod manages its own `data plane cluster`. Mesh operators could install another Istio mesh in the `control plane cluster` and configure its `istio-ingress` gateway to route traffic from a data plane cluster to its corresponding Istiod control plane.
+
+## Conclusion
+
+This new deployment model enables the Istio control plane to be run and managed by mesh operators who have operational expertise in Istio. Mesh operators can run the control plane in their own control plane clusters, provide it as a service to mesh admins. Mesh operators can optionally run multiple Istiod control planes in the control plane cluster, deploy their own Istio mesh and use their `istio-ingress` gateway to control access to these Istiod control planes.
```
smawson, line 65: We should make it clear that mesh operators can run the control planes however they want (for example they could run them on VMs), they can run multiple per cluster, etc. That is their implementation choice, we provide some examples to get started but there is lots of flexibility.

content/en/blog/2020/new-deployment-model/index.md
linsun: Ok, I've removed most texts related to single cluster today and updated it as overview.

content/en/blog/2020/new-deployment-model/index.md
linsun: Thanks Sven!  Updated PTAL

content/en/blog/2020/new-deployment-model/index.md
linsun: Thanks Sven!  Updated PTAL

content/en/blog/2020/new-deployment-model/index.md
linsun: Thanks Sven!  Updated PTAL

linsun:
@smawson @frankbu Thanks for the new comments and the new terminologies!  This blog has been updated with all your comments along with the diagrams.  Please let me know if there is anything else needed before it can be merged.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
+---
+title: A New Istiod Deployment Model
```
dcberg, line 2: Consider ... "An Alternative Istiod Deployment Model".

The title could be interpreted as if the current deployment model is changing.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+From experience working with various service mesh users and vendors, we believe there are 3 key personas for a typical service mesh:
+
+* Mesh Operator, who manages the service mesh control plane installation and upgrade.
+
+* Mesh Admin, often referred as Platform Owner, who owns the service mesh platform and defines the overall strategy and implementation for service owners to adopt service mesh.
+
+* Mesh User, often referred as Service Owner, who owns one or more services in the mesh.
+
+Istio currently requires the control plane to run in one of the clusters in the mesh, leading to a lack of separation between the mesh operator and the mesh admin. We're introducing a new deployment model for Istiod which enables mesh operators to install and manage the mesh control plane on separate clusters. This new deployment model allows a clear separation between mesh operators and mesh admins. With this new model, Istio mesh operators can run Istio control plane for mesh admins while mesh admins can still control the configuration of the control plane without worrying about installing or managing the Istio control plane. This model is transparent to mesh users.
```
dcberg, line 20: consider rewording .... "introducing an alternative deployment model"

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+    title="Multiple single clusters Istio meshes with Istiod outside"
+    caption="Multiple single clusters with multiple Istiod control planes in an external control plane cluster"
+    >}}
+
+The `external control plane cluster` can be used to host multiple Istiod control planes and each Istiod manages its own `config cluster`. Mesh operators could install another Istio mesh in the `external control plane cluster` and configure its `istio-ingress` gateway to route traffic from a config cluster to its corresponding Istiod control plane.
+
+## Conclusion
+
+This new deployment model enables the Istio control plane to be run and managed by mesh operators who have operational expertise in Istio. Mesh operators can run the control plane in their own control plane clusters or other environments, provide it as a service to mesh admins. Mesh operators can optionally run multiple Istiod control planes in the control plane cluster, deploy their own Istio mesh and use their `istio-ingress` gateway to control access to these Istiod control planes. Through these examples provided here, mesh operators can explore different implementation choices and choose what works best for them.
```
dcberg, line 63: Consider .... "This new deployment model enables the Istio control plane to be run and managed by mesh operators who have operational expertise and providing a clean separation between service mesh control and data planes."

linsun: Updated, thanks Dan!

content/en/blog/2020/new-deployment-model/index.md
linsun: Updated, thanks Dan!

content/en/blog/2020/new-deployment-model/index.md
linsun: Good point, Updated, thanks Dan!

content/en/blog/2020/new-deployment-model/index.md
smawson: Updated text looks good to me.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+After installing Istio using the [default installation profile](/docs/setup/install/istioctl/#install-istio-using-the-default-profile), you will have an Istiod control plane installed in a single cluster like the diagram below:
+
+{{< image width="100%"
+    link="single-cluster.svg"
+    alt="Istio mesh in a single cluster"
+    title="Single cluster Istio mesh"
+    caption="Istio mesh in a single cluster"
+    >}}
+
+With this new deployment model in Istio 1.7, it is possible to run Istiod on a separate cluster (the `external control plane cluster`) from services, as shown in the diagram below. The `external control plane cluster` is owned by mesh operator while the mesh admin owns the cluster that has services deployed in the mesh. Mesh admin has no access to the `external control plane cluster`. For mesh operators, you can follow the [external istiod single cluster step by step guide](https://github.com/istio/istio/wiki/External-Istiod-single-cluster-steps) to explore more on this. This model was called central Istiod among Istio maintainers but external Istiod is the official name now.
```
smawson, line 33: Last sentence is a bit awkward, could reword:

Note: In internal discussions among istio maintainers this model was known as "central istiod."

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+{{< image width="100%"
+    link="single-cluster-external-Istiod.svg"
+    alt="Istio mesh in a single cluster with Istiod outside"
+    title="Single cluster Istio mesh with Istiod outside"
+    caption="Single cluster Istio mesh with Istiod in an external control plane cluster"
+    >}}
+
+Mesh admins can expand the service mesh to multiple clusters, which are managed by the same Istiod running in the `external control plane cluster`. The `config cluster` also serves as the Istio configuration cluster, in addition to run services. The `external control plane cluster` reads Istio configurations from the `primary data plane cluster` and pushes them to both the `config cluster` and `remote cluster` per the diagram below.
```
smawson, line 42:
```suggestion
Mesh admins can expand the service mesh to multiple clusters, which are managed by the same Istiod running in the `external control plane cluster`. The `config cluster` also serves as the Istio configuration cluster, in addition to running services. The `external control plane cluster` reads Istio configurations from the `config cluster` and Istiod pushes configuration to the data plane running in both the `config cluster` and `remote cluster` per the diagram below.
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+Mesh operators may further expand this deployment model to manage multiple Istio control planes from an external control plane cluster that runs multiple Istiod control planes:
+
+{{< image width="100%"
+    link="multiple-external-Istiods.svg"
+    alt="Istio meshes in single clusters with Istiod outside"
+    title="Multiple single clusters Istio meshes with Istiod outside"
+    caption="Multiple single clusters with multiple Istiod control planes in an external control plane cluster"
+    >}}
+
+The `external control plane cluster` can be used to host multiple Istiod control planes and each Istiod manages its own `config cluster`. Mesh operators could install another Istio mesh in the `external control plane cluster` and configure its `istio-ingress` gateway to route traffic from a config cluster to its corresponding Istiod control plane.
```
smawson, line 59: I wouldn't say Istiod manages a config cluster, rather it is pulling configuration from its own config cluster.

smawson, line 59: I think this was actually talking about remote clusters, not config clusters.

Each Istiod manages its own set of remote clusters. Mesh operators can route requests from a remote cluster to its corresponding Istio control plane, for example by using Istio Gateway to handle the routing.

You could also potentially drop the whole discussion of using Istio to route requests to control planes, it is really cool but confusing (cue inception meme).

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Conclusion
+
+This new deployment model enables the Istio control plane to be run and managed by mesh operators who have operational expertise in Istio, and provides a clean separation between service mesh control and data planes. Mesh operators can run the control plane in their own control plane clusters or other environments, provide it as a service to mesh admins. Mesh operators can optionally run multiple Istiod control planes in the control plane cluster, deploy their own Istio mesh and use their `istio-ingress` gateway to control access to these Istiod control planes. Through these examples provided here, mesh operators can explore different implementation choices and choose what works best for them.
```
smawson, line 63:
```suggestion
This new deployment model enables the Istio control plane to be run and managed by mesh operators who have operational expertise in Istio, and provides a clean separation between service mesh control and data planes. Mesh operators can run the control plane in their own clusters or other environments, providing the control plane as a service to mesh admins. Mesh operators can run multiple Istiod control planes in a single cluster, deploying their own Istio mesh and using `istio-ingress` gateways to control access to these Istiod control planes. Through the examples provided here, mesh operators can explore different implementation choices and choose what works best for them.
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+This model reduces complexity for mesh admins by allowing them to focus on mesh configurations without operating the control plane themselves. Mesh admin could continue to configure mesh-wide settings and Istio resources without any access to external control plane clusters. Mesh users can continue to interact with the service mesh without any change.
```
smawson, line 65:
```suggestion
This model reduces complexity for mesh admins by allowing them to focus on mesh configurations without operating the control plane themselves. Mesh admin can continue to configure mesh-wide settings and Istio resources without any access to external control plane clusters. Mesh users can continue to interact with the service mesh without any changes.
```

content/en/blog/2020/new-deployment-model/index.md
linsun: Thanks, reworded.

content/en/blog/2020/new-deployment-model/index.md
linsun: Updated. Resulted from simply changing remote cluster to config cluster. :(

linsun: Thanks, I struggle with this when writing the blog.  for single cluster scenario, the config cluster is also the remote cluster.  Yes it is indeed talking about remote clusters here.

I'll update the diagram that for single cluster, it is config + remote cluster.

I added a link to instructions about setting up Istio to route requests to istiod in case mesh operators want to learn more.  Hope this helps to clarify this a bit.

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
+---
+title: An Alternative Istiod Deployment Model
```
frankbu, line 2:
```suggestion
title: An External Control Plane Deployment Model
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+From experience working with various service mesh users and vendors, we believe there are 3 key personas for a typical service mesh:
+
+* Mesh Operator, who manages the service mesh control plane installation and upgrade.
+
+* Mesh Admin, often referred as Platform Owner, who owns the service mesh platform and defines the overall strategy and implementation for service owners to adopt service mesh.
+
+* Mesh User, often referred as Service Owner, who owns one or more services in the mesh.
+
+Istio currently requires the control plane to run in one of the clusters in the mesh, leading to a lack of separation between the mesh operator and the mesh admin. We're introducing an alternative deployment model for Istiod which enables mesh operators to install and manage the mesh control plane on separate clusters. This new deployment model allows a clear separation between mesh operators and mesh admins. With this new model, Istio mesh operators can run Istio control plane for mesh admins while mesh admins can still control the configuration of the control plane without worrying about installing or managing the Istio control plane. This model is transparent to mesh users.
```
frankbu, line 20:
```suggestion
Prior to version 1.7, Istio required the control plane to run in one of the clusters in the mesh, leading to a lack of separation between the mesh operator and the mesh admin. Istio 1.7 introduces a new {{< gloss >}}external control plane{{< /gloss >}} deployment model which enables mesh operators to install and manage mesh control planes on separate external clusters. This deployment model allows a clear separation between mesh operators and mesh admins. Istio mesh operators can now run Istio control planes for mesh admins while mesh admins can still control the configuration of the control plane without worrying about installing or managing the control plane. This model is transparent to mesh users.
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+After installing Istio using the [default installation profile](/docs/setup/install/istioctl/#install-istio-using-the-default-profile), you will have an Istiod control plane installed in a single cluster like the diagram below:
+
+{{< image width="100%"
+    link="single-cluster.svg"
+    alt="Istio mesh in a single cluster"
+    title="Single cluster Istio mesh"
+    caption="Istio mesh in a single cluster"
+    >}}
+
+With this new deployment model in Istio 1.7, it is possible to run Istiod on a separate cluster (the `external control plane cluster`) from services, as shown in the diagram below. The `external control plane cluster` is owned by mesh operator while the mesh admin owns the cluster that has services deployed in the mesh. Mesh admin has no access to the `external control plane cluster`. For mesh operators, you can follow the [external istiod single cluster step by step guide](https://github.com/istio/istio/wiki/External-Istiod-single-cluster-steps) to explore more on this. Note: In some internal discussions among Istio maintainers this model was known as "central istiod".
```
frankbu, line 33:
```suggestion
With the new deployment model in Istio 1.7, it's possible to run Istiod on an external cluster, separate from the mesh services as shown in the diagram below. The external control plane cluster is owned by the mesh operator while the mesh admin owns the cluster running services deployed in the mesh. The mesh admin has no access to the external control plane cluster. Mesh operators can follow the [external istiod single cluster step by step guide](https://github.com/istio/istio/wiki/External-Istiod-single-cluster-steps) to explore more on this. (Note: In some internal discussions among Istio maintainers, this model was previously referred to as "central istiod".)
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+{{< image width="100%"
+    link="single-cluster-external-Istiod.svg"
+    alt="Istio mesh in a single cluster with Istiod outside"
+    title="Single cluster Istio mesh with Istiod outside"
+    caption="Single cluster Istio mesh with Istiod in an external control plane cluster"
+    >}}
+
+Mesh admins can expand the service mesh to multiple clusters, which are managed by the same Istiod running in the `external control plane cluster`. The `config cluster` also serves as the Istio configuration cluster, in addition to running services. The `external control plane cluster` reads Istio configurations from the `config cluster` and Istiod pushes configuration to the data plane running in both the `config cluster` and `remote cluster` per the diagram below.
```
frankbu, line 42:
```suggestion
Mesh admins can expand the service mesh to multiple clusters, which are managed by the same Istiod running in the external cluster. None of the mesh clusters are {{< gloss "primary cluster" >}}primary clusters{{< /gloss >}}, in this case. They are all {{< gloss "remote cluster" >}}remote clusters{{< /gloss >}}. However, one of them also serves as the Istio configuration cluster, in addition to running services. The external control plane reads Istio configurations from the `config cluster` and Istiod pushes configuration to the data plane running in both the config cluster and other remote clusters as shown in the diagram below.
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+{{< image width="100%"
+    link="multiple-clusters-external-Istiod.svg"
+    title="Multicluster Istio mesh with Istiod outside"
+    caption="Multicluster Istio mesh with Istiod in an external control plane cluster"
+    >}}
+
+Mesh operators may further expand this deployment model to manage multiple Istio control planes from an `external control plane cluster` that runs multiple Istiod control planes:
```
frankbu, line 50:
```suggestion
Mesh operators can further expand this deployment model to manage multiple Istio control planes from an external cluster running multiple Istiod control planes:
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+{{< image width="100%"
+    link="multiple-external-Istiods.svg"
+    alt="Istio meshes in single clusters with Istiod outside"
+    title="Multiple single clusters Istio meshes with Istiod outside"
+    caption="Multiple single clusters with multiple Istiod control planes in an external control plane cluster"
+    >}}
+
+The `external control plane cluster` can be used to host multiple Istiod control planes and each Istiod manages its own `remote cluster`. Mesh operators could install their own Istio mesh in the `external control plane cluster` and configure its `istio-ingress` gateway to route traffic from a `remote cluster` to its corresponding Istiod control plane. To learn more about this, check out [these steps](https://github.com/istio/istio/wiki/External-Istiod-single-cluster-steps#deploy-istio-mesh-on-external-control-plane-cluster-to-manage-traffic-to-istiod-deployments).
```
frankbu, line 59:
```suggestion
The external cluster can also be used to host multiple Istiod control planes with each Istiod managing its own remote cluster(s). Mesh operators can even install their own Istio mesh in the external control plane cluster and configure its `istio-ingress` gateway to route traffic from remote clusters to their corresponding Istiod control planes. To learn more about this, check out [these steps](https://github.com/istio/istio/wiki/External-Istiod-single-cluster-steps#deploy-istio-mesh-on-external-control-plane-cluster-to-manage-traffic-to-istiod-deployments).
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Conclusion
+
+This new deployment model enables the Istio control plane to be run and managed by mesh operators who have operational expertise in Istio, and provides a clean separation between service mesh control and data planes. Mesh operators can run the control plane in their own clusters or other environments, providing the control plane as a service to mesh admins. Mesh operators can run multiple Istiod control planes in a single cluster, deploying their own Istio mesh and using `istio-ingress` gateways to control access to these Istiod control planes. Through the examples provided here, mesh operators can explore different implementation choices and choose what works best for them.
```
frankbu, line 63:
```suggestion
The external control plane deployment model enables the Istio control plane to be run and managed by mesh operators who have operational expertise in Istio, and provides a clean separation between service mesh control and data planes. Mesh operators can run the control plane in their own clusters or other environments, providing the control plane as a service to mesh admins. Mesh operators can run multiple Istiod control planes in a single cluster, deploying their own Istio mesh and using `istio-ingress` gateways to control access to these Istiod control planes. Through the examples provided here, mesh operators can explore different implementation choices and choose what works best for them.
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+This model reduces complexity for mesh admins by allowing them to focus on mesh configurations without operating the control plane themselves. Mesh admin can continue to configure mesh-wide settings and Istio resources without any access to external control plane clusters. Mesh users can continue to interact with the service mesh without any changes.
```
frankbu, line 65:
```suggestion
This new model reduces complexity for mesh admins by allowing them to focus on mesh configurations without operating the control plane themselves. Mesh admins can continue to configure mesh-wide settings and Istio resources without any access to external control plane clusters. Mesh users can continue to interact with the service mesh without any changes.
```

content/en/blog/2020/new-deployment-model/index.md
craigbox: "Deploying Istio control planes outside mesh clusters" ?

Much more clear and says what the post is about immediately

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
+---
+title: An Alternative Istiod Deployment Model
+subtitle: Clear Separation between Mesh Operator and Mesh Admin
```
craigbox, line 3: something like "Provide a clear separation of concerns between mesh admins and cluster admins"?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+description: A new deployment model for Istio.
+publishdate: 2020-08-26
+attribution: "Lin Sun (IBM), Iris Ding (IBM)"
+keywords: [istiod,deployment model,install,deploy,'1.7']
+---
+
+## Overview
+
+From experience working with various service mesh users and vendors, we believe there are 3 key personas for a typical service mesh:
```
craigbox, line 12: Comment disappeared on commit change and I don't seem to be able to add more text there.

I still think you would benefit from using "Service owner" so they don't all start with "Mesh" (b) I'm not sure I've seen evidence that Mesh Operator and Mesh Admin are two personas, _unless_ they're already consuming a managed service, or something like this.

content/en/blog/2020/new-deployment-model/index.md
craigbox: I still get lost with "mesh operators" and "mesh admins"; my head blurs them into the same persona. Maybe refer to them without both starting with "mesh" ?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+* Mesh Operator, who manages the service mesh control plane installation and upgrade.
+
+* Mesh Admin, often referred as Platform Owner, who owns the service mesh platform and defines the overall strategy and implementation for service owners to adopt service mesh.
+
+* Mesh User, often referred as Service Owner, who owns one or more services in the mesh.
+
+Istio currently requires the control plane to run in one of the clusters in the mesh, leading to a lack of separation between the mesh operator and the mesh admin. We're introducing an alternative deployment model for Istiod which enables mesh operators to install and manage the mesh control plane on separate clusters. This new deployment model allows a clear separation between mesh operators and mesh admins. With this new model, Istio mesh operators can run Istio control plane for mesh admins while mesh admins can still control the configuration of the control plane without worrying about installing or managing the Istio control plane. This model is transparent to mesh users.
+
+## External Control Plane Deployment model
```
craigbox, line 22:
```suggestion
## External control plane deployment model
```

maybe?

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+After installing Istio using the [default installation profile](/docs/setup/install/istioctl/#install-istio-using-the-default-profile), you will have an Istiod control plane installed in a single cluster like the diagram below:
```
craigbox, line 24: I maintain that it's not important that it's istiod, it's just that it's a control plane.  That's my opinion however, please feel free to ignore.

content/en/blog/2020/new-deployment-model/index.md
craigbox: That is another reason I don't think the term "primary cluster" is suitable.  You can have none of them! 🤨

frankbu: Maybe not ideal, but "primary cluster" seems OK to me. A standalone mesh needs at least one primary cluster in it. The alternative is to use an external control plane to control the mesh. We had a long discussion before agreeing on the name. One other reasonable suggestion, "control cluster", was rejected for some reason I don't remember. @nmittler @smawson Do you remember?

content/en/blog/2020/new-deployment-model/index.md
frankbu: "Deploying Istio Control Planes Outside the Mesh"

content/en/blog/2020/new-deployment-model/index.md
nmittler: I recall preferring `control cluster` specifically because `primary` implies `secondary`, which is commonly used more in failover scenarios. @smawson do u recall?

content/en/blog/2020/new-deployment-model/index.md
linsun: I think in this case, they are because one team owns control plane install/upgrade, one team managed the mesh data plane.

content/en/blog/2020/new-deployment-model/index.md
linsun: Yes,  you are right.  We had Istiod here because that is what the diagram had so easier for user to read + match what is in the diagram.

content/en/blog/2020/new-deployment-model/index.md
linsun: the reason I had mesh in the name is because cluster admin may not operate meshes...

linsun:
@frankbu thank you much for taking a pass on the docs.  I think the blog is *way* better with all of your help @frankbu @smawson @craigbox @rvennam @nmittler @dcberg, much appreciated!

There is some concern on `primary cluster` in the blog - which disappeared as I committed Frank's suggestion. I think we should move that discussion to a separate thread or doc channel because that involves changes in many places in istio.io if many think that name should be changed.

content/en/blog/2020/new-deployment-model/index.md
linsun: If we remove mesh, say operators or admins, I think it is too wide and can cause confusions as well :-(

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Deploying Istio Control Planes Outside the Mesh
+subtitle: Clear Separation between Mesh Operator and Mesh Admin
+description: A new deployment model for Istio.
+publishdate: 2020-08-26
```
frankbu, line 5:
```suggestion
publishdate: 2020-08-27
```

content/en/blog/2020/new-deployment-model/index.md
```diff
@@ -0,0 +1,65 @@
...
+Mesh operators can further expand this deployment model to manage multiple Istio control planes from an external cluster running multiple Istiod control planes:
+
+{{< image width="100%"
+    link="multiple-external-Istiods.svg"
+    alt="Istio meshes in single clusters with Istiod outside"
+    title="Multiple single clusters Istio meshes with Istiod outside"
+    caption="Multiple single clusters with multiple Istiod control planes in an external control plane cluster"
+    >}}
+
+The external cluster can also be used to host multiple Istiod control planes with each Istiod managing its own remote cluster(s). Mesh operators can even install their own Istio mesh in the external control plane cluster and configure its `istio-ingress` gateway to route traffic from remote clusters to their corresponding Istiod control planes. To learn more about this, check out [these steps](https://github.com/istio/istio/wiki/External-Istiod-single-cluster-steps#deploy-istio-mesh-on-external-control-plane-cluster-to-manage-traffic-to-istiod-deployments).
```
frankbu, line 59:
```suggestion
In this case, each Istiod manages its own remote cluster(s). Mesh operators can even install their own Istio mesh in the external control plane cluster and configure its `istio-ingress` gateway to route traffic from remote clusters to their corresponding Istiod control planes. To learn more about this, check out [these steps](https://github.com/istio/istio/wiki/External-Istiod-single-cluster-steps#deploy-istio-mesh-on-external-control-plane-cluster-to-manage-traffic-to-istiod-deployments).
```

craigbox:
My role in Steering is to check it's a well written post which reflects well on the project, which is now LGTM - would be my preference for @smawson to do to the signoff but more with his TOC hat on.

PR #7983, opened 2020-08-19
clarify request timeout default

linsun:
cc @GregHanson @rvennam 
Note: in the doc below, it also said ```In this task, you used Istio to set the request timeout for calls to the reviews microservice to half a second. By default the request timeout is disabled.```

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/request-timeouts/index.md
+++ b/content/en/docs/tasks/traffic-management/request-timeouts/index.md
@@ -28,7 +28,7 @@ This task shows you how to setup request timeouts in Envoy using Istio.
 ## Request timeouts
 
 A timeout for http requests can be specified using the *timeout* field of the [route rule](/docs/reference/config/networking/virtual-service/#HTTPRoute).
-By default, the timeout is 15 seconds, but in this task you override the `reviews` service
+By default, the request timeout is diabled, but in this task you override the `reviews` service
 timeout to 1 second.
 To see its effect, however, you also introduce an artificial 2 second delay in calls
 to the `ratings` service.


content/en/docs/tasks/traffic-management/request-timeouts/index.md
```diff
@@ -28,7 +28,7 @@ This task shows you how to setup request timeouts in Envoy using Istio.
 ## Request timeouts
 
 A timeout for http requests can be specified using the *timeout* field of the [route rule](/docs/reference/config/networking/virtual-service/#HTTPRoute).
-By default, the timeout is 15 seconds, but in this task you override the `reviews` service
+By default, the request timeout is diabled, but in this task you override the `reviews` service
```
frankbu, line 31: This was just changed the other way in May. https://github.com/istio/istio.io/pull/7255

frankbu:
It's unclear what the behavior is, so requesting review signoff from @istio/wg-networking-maintainers before merge.

content/en/docs/tasks/traffic-management/request-timeouts/index.md
linsun: I know which is why it is confusing.  We actually had a different default than envoy.

I learned this hard way from my envoy config in my pod:

```
          "route": {
           "cluster": "outbound|80||sleep.default.svc.cluster.local",
           **"timeout": "0s",**
           "retry_policy": {
            "retry_on": "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes",
            "num_retries": 2,
            "retry_host_predicate": [
             {
              "name": "envoy.retry_host_predicates.previous_hosts"
             }
            ],
            "host_selection_retry_max_attempts": "5",
            "retriable_status_codes": [
             503
            ]
           },
           "max_grpc_timeout": "0s"
          },
          "decorator": {
           "operation": "sleep.default.svc.cluster.local:80/*"
          },
          "name": "default"
         }
```

frankbu: In that case, it also needs to be fixed in the concepts doc: https://istio.io/latest/docs/concepts/traffic-management/#timeouts

Also, if the Istio API has a different default than Envoy, it should also say that in the [API reference](https://github.com/istio/api/blob/c78f387f78a22a3c4a1c49679dc44cded454530d/networking/v1beta1/virtual_service.proto#L604).

hzxuzhonghu: We do not respect the envoy default (15s).  In istio the default timeout value is `infinit` if `ISTIO_DEFAULT_REQUEST_TIMEOUT` env is 0.

hzxuzhonghu: If ISTIO_DEFAULT_REQUEST_TIMEOUT is set to `5s`, then the default timeout is 5s

PR #7986, opened 2020-08-20
Update index.md

johnzheng1975:
Change the helloworld service/ deploy to real content. It can run well in my testing.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/blog/2017/0.1-canary/index.md
+++ b/content/en/blog/2017/0.1-canary/index.md
@@ -43,13 +43,14 @@ We begin by defining the **helloworld** Service, just like any other Kubernetes 
 apiVersion: v1
 kind: Service
 metadata:
-name: helloworld
-labels:
-  app: helloworld
+  name: helloworld
 spec:
+  ports:
+  - port: 80
+    protocol: TCP
+    targetPort: 80
   selector:
     app: helloworld
-  ...
 {{< /text >}}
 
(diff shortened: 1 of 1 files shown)


frankbu:
The last line of this blog post points to the sample code used in the article: https://github.com/istio/istio/tree/release-1.7/samples/helloworld

Anyway, we generally do not want to update old blog posts. Blogs are considered single-point-in-time articles, and rarely updated. If we think this is a valuable enough doc, we could consider moving it (or adding something similar) to the [traffic managment tasks](https://istio.io/latest/docs/tasks/traffic-management/).

content/en/blog/2017/0.1-canary/index.md
```diff
@@ -59,33 +60,51 @@ apiVersion: apps/v1
...
   template:
     metadata:
       labels:
         app: helloworld
         version: v1
     spec:
       containers:
-      - image: helloworld-v1
-        ...
+      - image: nginx
```
frankbu, line 79: this is supposed to be the helloworld image at dockerhub.

frankbu:
Closing because we do not want to update the blog.

@johnzheng1975 Please open an issue if you think it would be worthwhile to add something similar in the mainline docs (instead of just in a blog).

johnzheng1975:
That's make sense - Do not change blog. Just leave this closed.
Thanks for your time.

PR #8003, opened 2020-08-23
automated test for zipkin dashboard and modify snippet generation to take boilerplate into account

su225:
https://preliminary.istio.io/latest/docs/tasks/observability/distributed-tracing/zipkin/

It is assumed that bookinfo app works as expected. So trace obtained should match expected one (like istio-ingressgateway should contact productpage, not the other way around). Full trace may not be available due to sampling. This resembles what is shown on the screen (both fetch from an API).

There is a retry loop because traces may not be available and we might have to wait some time. 

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


su225:
@howardjohn - Thanks. I see that it is already covered. Closing it... On a side note, shall I remove `test:yes/no` tag in `index.md` so that we don't have `X Page test` on top?

su225:
@frankbu - As per feedback on istio/istio.io#8006 i have removed verification part, but instead just test if it is accessible through `istioctl dashboard` port-forward.

content/en/docs/tasks/observability/distributed-tracing/zipkin/test.sh
```diff
@@ -0,0 +1,66 @@
...
+startup_bookinfo_sample
+
+# This shows up in many places. Should move to util? 
+function send_productpage_requests() {
+  _set_ingress_environment_variables
+  local GATEWAY_URL="$INGRESS_HOST:$INGRESS_PORT"
+  for _ in {1..250}; do
+    curl -s "http://$GATEWAY_URL/productpage" > /dev/null
+  done
+}
```
frankbu, line 44: There is a snip for this, that isn't currently being generated by scripts/snip.py, because it doesn't currently handle boiler-plate.
```
{{< boilerplate trace-generation >}}
```
If handled properly, the snip generator would output a snip function, something like this:
```
snip_trace_generation() {
for i in `seq 1 100`; do curl -s -o /dev/null http://$GATEWAY_URL/productpage; done
}
``` 
If you would like to fix the snip generator to output boilerplate snips, that would be great. Otherwise, for now I would just simulate it by adding a fake function that looks like the expected snip and use it in the test
```suggestion
# TODO fix snip generator to gen this function in snips.sh
snip_trace_generation() {
for i in `seq 1 100`; do curl -s -o /dev/null http://$GATEWAY_URL/productpage; done
}
```

content/en/docs/tasks/observability/distributed-tracing/zipkin/test.sh
```diff
@@ -0,0 +1,66 @@
...
+  curl -s -o /dev/null -w "%{http_code}" "$zipkin_url"
+}
+
+_verify_same access_zipkin_with_portforward "200"
+pgrep istioctl | xargs kill
+
+# @cleanup
+set +e
+cleanup_bookinfo_sample
+pgrep istioctl | xargs kill
```
frankbu, line 62: We really should use the documented cleanup snip in this and other similar tests. At least, add a TODO to figure out how to test with the "documented cleanup". Maybe we need to change the command being used in the doc for cleanup?

su225: Oh! Right. The same old problem in other two tests. I'll add it.

content/en/docs/tasks/observability/distributed-tracing/zipkin/test.sh
su225: Thanks. Let me check out that.

content/en/docs/tasks/observability/distributed-tracing/zipkin/test.sh
su225: Created an issue here - https://github.com/istio/istio.io/issues/8014

content/en/docs/tasks/observability/distributed-tracing/zipkin/test.sh
```diff
@@ -0,0 +1,66 @@
...
+function send_productpage_requests() {
+  _set_ingress_environment_variables
+  local GATEWAY_URL="$INGRESS_HOST:$INGRESS_PORT"
+  for _ in {1..250}; do
+    curl -s "http://$GATEWAY_URL/productpage" > /dev/null
+  done
+}
+
+snip_accessing_the_dashboard_1 &
+send_productpage_requests
```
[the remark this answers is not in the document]
su225: Fixed it

content/en/docs/tasks/observability/distributed-tracing/zipkin/test.sh
su225: Done. I generated snippets for boilerplate (of course, updated scripts/gen_snips.sh` and `scripts/snips.py`) and "sourced" it from `snip.sh` for each page. Reason - boilerplates are meant to be reusable. So it is better to have those snippets in one place and "source" them wherever needed. That's also why I used different prefix to avoid naming collision. Some minor changes are due to linter.

scripts/snip.py
```diff
@@ -158,6 +167,9 @@
 
 with open(os.path.join(snipdir, snipfile), 'w', encoding='utf-8') as f:
     f.write(HEADER % markdown.split("content/en/")[1] if "content/en/" in markdown else markdown)
+    for bp in boilerplates:
+        source_line = f"source 'content/en/boilerplates/snips/{bp}.md_snips.sh'\n"
```
frankbu, line 171: I think the file name can be a little simpler (remove the `.md_snips` part. Also, use double quotes for the file might be more consistent with the recommended pattern for adding sources manually.
```suggestion
        source_line = f"source \"content/en/boilerplates/snips/{bp}.sh\"\n"
```

su225: Done.

content/en/boilerplates/snips/index.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
frankbu, line 1: A really nice optimization would be to not generate a snip file for boilerplate.md files that don't have any `{{< text }}` snippets in them and then only source the .sh file if it exists.

We can leave that for a future PR, if you don't have time to do it now.

scripts/snip.py
```diff
@@ -158,6 +171,13 @@
 
 with open(os.path.join(snipdir, snipfile), 'w', encoding='utf-8') as f:
```
frankbu, line 172: How about only saving the file if it has snippets?
```
# Only generate snip file if it has content.
if len(boilerplates) and len(snippets):
    with open(os.path.join(snipdir, snipfile), 'w', encoding='utf-8') as f:
    ...
```

su225: That's what I did at first (but not committed, so not shown here). **However, when they are "sourced" from doc tests, shellcheck (rightly) complains that there is no such file.** To fix some way to check if boilerplate has some bash snippets associated with it. A simpler (but not so good as it introduces coupling with `scripts/gen_snips.sh`) design would be to define "boilerplate snippets" directory from where it can lookup if there is a snippet file (meaning it has some `{{<text>}}` section.

su225: As mentioned in comments in `snip.py` we already have some coupling with `scripts/gen_snips.sh`. Maybe I can write that optimization?

su225: Done. I applied both optimizations you suggested (and shellcheck is happy)

content/en/boilerplates/snips/index.sh
su225: Done already :) Number of files in the PR reduced from 31 to 23.

scripts/snip.py
su225: Fixed it. The bug was - consider the case where m boilerplates were found, but 0 snippets. If all m of them don't have snippets then we would have end up with an empty file. It is fixed by filtering boilerplates without snippets early on.

PR #8006, opened 2020-08-24
remove test:no/yes from index.md for zipkin, jaeger, lightstep, tracing overview and kiali

su225:
As per review in https://github.com/istio/istio.io/pull/8003#pullrequestreview-473052262 , tracing is covered by integration test in istio/istio or mesh visualization is more about testing Kiali's functionality than Istio. So `test:yes/no` can be removed as doc test is not necessary.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/observability/distributed-tracing/jaeger/index.md
+++ b/content/en/docs/tasks/observability/distributed-tracing/jaeger/index.md
@@ -6,7 +6,6 @@ keywords: [telemetry,tracing,jaeger,span,port-forwarding]
 aliases:
  - /docs/tasks/telemetry/distributed-tracing/jaeger/
 owner: istio/wg-policies-and-telemetry-maintainers
-test: no
 ---
 
 After completing this task, you understand how to have your application participate in tracing with [Jaeger](https://www.jaegertracing.io/),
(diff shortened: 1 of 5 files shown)


frankbu:
@su225 I agree with @howardjohn that much of your test in #8003 is feature testing, covered by the integration tests, but we do still need doc tests, to test that the "doc is correct". The doc says that you can use the istioctl dashboard command to access zipkin, it then says that if you run a snip that sends requests to bookinfo /productpage (the generator is currently ommitting this snip, because it's in a boilerplate - I think we should fix the generator for that), and then it shows what to expect in some screenshots. Since we can't really test screen shots, I think the verify should just do something minimal to check that there is something in zipkin. That would be a sufficient "doc page test" IMO. Your python verify code is overkill for a doc test, it's the part that is covered by the integration tests.

su225:
@frankbu - If I understand it correctly, we don't have to test if trace is correct as it is covered by integration test. However, we have to test if we can access dashboard by `istioctl dashboard` command - which is testing if it sets up port-forwarding correctly. So that would be sufficient for doc-test. From this perspective, we do need a test for https://preliminary.istio.io/latest/docs/tasks/observability/gateways/ as it is about exposing dashboards and making sure they are accessible as given in docs. Correct?

PR #8017, opened 2020-08-25
Describe external control planes

frankbu:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


linsun:
Thanks Frank!  LGTM

linsun:
@frankbu can you add what we call for remote cluster that also serves istio config?  I basically used primary remote cluster in my blog.

content/en/docs/reference/glossary/external-control-plane.md
```diff
@@ -0,0 +1,10 @@
+---
+title: External Control Plane
+test: n/a
+---
+
+An external control plane is a [control plane](/docs/reference/glossary/#control-plane)
+that externally manages mesh workloads running in their own [clusters](/docs/reference/glossary/#cluster)
+or other infrastructure. The control plane may, itself, be deployed in a cluster, although not
+in one of the [primary clusters](/docs/reference/glossary/#primary-cluster) of the mesh it's controlling.
```
nmittler, line 9: I don't think this sentence is correct. There is nothing about a `primary cluster` that implies it has to be part of the `mesh`. It's just a cluster that runs a control plane.

frankbu: I don't think the sentence is implying a primary cluster has to be part of the mesh. I intentionally avoided the issue by not calling the control plane's cluster anything. I'm just saying the control plane's cluster cannot be "one of the primary clusters in the mesh". Maybe slightly different wording would be better?
```suggestion
in one of the clusters that is part of the mesh it's controlling.
```

frankbu:
@linsun 

> can you add what we call for remote cluster that also serves istio config? I basically used primary remote cluster in my blog.

I think Nate suggested `config cluster` for that, which seems fine to me. I don't think it really belongs in this PR, though, since it's not really related to the `control plane model`. It's about which cluster(s) have the config in a multicluster deployment, regardless of whether it's using an `external control plane` or other control plane (e.g., multiple primary) deployment model.

Maybe we will want to document that concept at some point, and put it in the glossary, but for now I think it would be fine for you to just call it the config cluster in your blog, without the docs formally defining such a term.

content/en/docs/reference/glossary/external-control-plane.md
nmittler: Ah I think I misread it, my mistake. Your suggestion LG.

linsun:
@frankbu config cluster works for me.  Should I call the cluster config cluster + remote cluster as it is a remote cluster that runs users' services/workloads as well?

content/en/docs/reference/glossary/external-control-plane.md
```diff
@@ -0,0 +1,10 @@
+---
+title: External Control Plane
+test: n/a
+---
+
+An external control plane is a [control plane](/docs/reference/glossary/#control-plane)
+that externally manages mesh workloads running in their own [clusters](/docs/reference/glossary/#cluster)
+or other infrastructure. The control plane may, itself, be deployed in a cluster, although not
+in one of the clusters that is part of the mesh it's controlling.
+Its purpose is to cleanly separate the control plane from the data plane of a mesh.
```
sdake, line 10: @frankbu alot of the media has picked up on the word "central istiod". Would you mind adding a sentence that says something like "This was previously identified as `central Istiod`"?

frankbu: That belongs in Lin's blog, not here. We really should downplay that term IMO, because as we discussed on slack, it's not really in the center of anything.

linsun: Ok I'll add it to my blog.

linsun:
Hi @rcaballeromx I asked @frankbu to merge this because it has two TOC approval, and 1 env WG lead and 1 multicluster lead approvals.

I  had contacted you via email to get a source of the diagram but no reply.  Can you elaborate which piece is not complaint?  I followed https://istio.io/latest/about/contribute/diagrams/

frankbu:
@rcaballeromx As I told you on slack, if you'd like to send a followup PR to improve the diagram, that would be great.

PR #8030, opened 2020-08-26
Remove mention of "pilot agent"

howardjohn:
I have heard 5+ cases of folks thinking this means it going through the
control plane. https://github.com/istio/istio/issues/26737 for example.

We should not call it "pilot agent". We are callnig it istio-agent
internally, not sure if we want to make that in the docs or something
else -- anything but "pilot" agent.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


howardjohn:
This is sort of introducing a new terminology. I don't really care what the name is as long as its not "pilot", so feel free to add any suggestions

PR #8033, opened 2020-08-26
Fix lint_site.sh to make it (mostly) work on OSX.

ssuchter:
Change #1: OSX (bsd) sed requires an argument for -i
Change #2: OSX (bsd) xargs doesn't support -r, and it wasn't necessary to use anyway, so removed it

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# TODO: Use the below in every script that uses the -i option on sed
+find_inplace_sed() {
+    local tmpfile=`mktemp`
+    local POSSIBLE_INPLACE_SED_CMDS=("sed -i" "sed -i ''")
+    local sed_cmd
+    unset INPLACE_SED
+    for sed_cmd in "${POSSIBLE_INPLACE_SED_CMDS[@]}"; do
+        echo orig > ${tmpfile}
+        if eval ${sed_cmd} 's/orig/new/g' ${tmpfile} >/dev/null 2>&1 ; then
+            INPLACE_SED="${sed_cmd}"
+            break
+        fi
+    done
+    rm $tmpfile
+    if [ -z ${INPLACE_SED+x} ]; then
+        echo "Cannot find a valid inplace sed command from ${POSSIBLE_INPLACE_SED_CMDS}"
+        return 1
+    fi
+    return 0
+}
--- a/scripts/lint_site.sh
+++ b/scripts/lint_site.sh
@@ -24,6 +24,13 @@ else
     LANGS="en zh"
 fi
 
+SCRIPT_DIR="${BASH_SOURCE%/*}"
+if [[ ! -d "$SCRIPT_DIR" ]]; then SCRIPT_DIR="$PWD"; fi
+. "$SCRIPT_DIR/common.sh"
+if ! find_inplace_sed; then
+    exit 1
+fi
+
 # This performs spell checking and style checking over markdown files in a content
 # directory. It transforms the shortcode sequences we use to annotate code blocks
 # into classic markdown ``` code blocks, so that the linters aren't confused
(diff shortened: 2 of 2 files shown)


sdake:
@ssuchter nice job on the work. I would approve, although these files are copied from the repo https://github.com/istio/common-files. Can you submit your PR there instead? These files are then copied around to the various 15 repositories Istio maintains.

ssuchter:
An, thanks @sdake. I'll take a look, it will take a few hours for me to get to that. I'd definitely prefer to put things in the right home.

frankbu:
> these files are copied from the repo https://github.com/istio/common-files.

Only `common.sh` is copied over, `lint_site.sh` is a local file that you need to update here.

ssuchter:
I took a quick look. Given that it's in a 'common' directory already, I don't want to call it 'common.sh' anymore. I'd put it in setup_env.sh, but that seems to be focused on build-related things. I'd love it if setup_env.sh was called build_env_setup.sh, but I don't feel confident to know that that's a good idea.

What about this being called (in the istio:common/scripts directory) something like 'env_tools.sh' ? I'd like your thoughts on that naming.

frankbu:
> What about this being called (in the istio:common/scripts directory) something like 'env_tools.sh' ?

SGTM, but I think you need to have this discussion with the maintainers of the istio/common repo. Whatever you agree on with them, will then just be copied over here. Unless, you don't know of any other scripts that needs the new function, in which case why not just put it in lint_site.sh for now, and only make it reusable when another use for comes up.

ssuchter:
I'll make a pull against istio/common. Other scripts do need this, I just haven't modified them yet.

```
MBP:~/src/ssuchter-istio.io/scripts% grep -l 'sed -i' *.sh
archive_version.sh
build_archive_site.sh
build_site.sh
common.sh
create_version.sh
grab_reference_docs.sh
prepare_release.sh

```

sdake:
FWIW, we have had some complaints from our mac developer community that we use GNU shell features (which are mostly POSIX specified...) within MacOS. This mandates the need to install brew, to obtain some of the toolchain.

The build environment ideally should need make and docker, and nothing else. It was this way originally, but things change. As a result, any improvements in this area to remove gnu-isms and ensure `make build` for istio/istio or `make gen` (our generic generation target) work on macos would be *highly* valued.

I think it would be great to add some generalized helper functions (for things like sed).

In 2017 I wrote generic sed for a file headed for the bitbucket (that I still think is unretired. Might be useful for inspiration: https://github.com/istio/istio/pull/2303. There are multiple environmental conditions to deal with (homebrew macos, linux).

sdake:
@sshucker @frankbu also prototyping a common files change against a repo it is copied to is quite common, so please feel free to submit PRs here or istio/istio or istio/envoy that test your changes in the checks before ultimately copying them to common/common (where there are sadly, no checks).

Cheers,
-steve

ssuchter:
I wonder if we could consider some base version of python as what's
required (in addition to docker and make). A lot of bash-isms and use of
hard-to-understand complex command lines can be obviated with well-written
python. For example, this lint-site.sh would be infinitely easier to
understand and maintain if it were lint-site.py. We wouldn't need any
external grep, xargs, find, etc, commands. (We'd still probably fork
mdspell).

On my computers, I observe these are the "default" pythons installed by the
OS:
Centos 8 Linux: python3.6
OSX: python2.7 (!)

It's certainly possible to write valid python that will work on both
python2.7 and python3, until Apple gets their act together.

How would you feel about that dependence? What versions of python do you
happen to have already available?

Sean

ssuchter:
https://github.com/istio/istio/pull/26906

howardjohn:
> The build environment ideally should need make and docker, and nothing else. It was this way originally, but things change. 

Why is this not the case? don't we run everything inside of docker?

ssuchter:
Well, the motivating reason for me in this case is that I was making a doc
change, and I wanted to lint the doc change. Except for these unix tool
inconsistencies, it would have straight up worked.

ssuchter:
https://github.com/istio/common-files/pull/304/files

ericvn:
@ssuchter I'm all for making the changes (and I actually have used brew to get some commands in the past). I'm curious what failed for you in docs that necessitated this change? I assume you are trying to run something locally instead of via the makefile and its targets?

ssuchter:
I was trying to run ./scripts/lint_site.sh in the istio.io repo.

ericvn:
Simpliest way to do this without having to make MacOS changes to the script would be to simply run `make shell` which starts a build container and puts you at the prompt. From the prompt, you can simply run `./scripts/lint_site.sh`.

sdake:
there (should) be make targets for every type of operation required by a user. I believe this is `make lint`.

ericvn:
Agree that in most cases, simply running the `make target` is the simplest. But sometimes the target may not be available or do more that a person wants. In that case, falling back to the shell command line might be appropriate.

frankbu:
I'm starting to think this isn't really a very good idea because there are more incompatibilities that we really don't want to deal with. For example, I recently noticed that the extended regex syntax supported by the `-E` option is also different on OSX.

I think we should stick to using `make shell` or, in this case, `make lint`, so that we're guaranteed to get the same result as the build, when we locally run the scripts.

frankbu:
Closing for now.

@ssuchter Feel free to reopen, or open another PR, if you still want to make some of these changes.

PR #8042, opened 2020-08-27
Automated test - Remotely access telemetry addons (both secure and insecure ways)

su225:
https://istio.io/latest/docs/tasks/observability/gateways/

Automated docs test for accessing telemetry addons through gateways
For "secure access" test, self-signed certificate is used.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/tasks/observability/gateways/test_utils.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
```
frankbu, line 1: How about moving this to `tests/util/addons.sh`? In the future we can refactor and make it more generally reusable for other tests that need addons.

su225: Done. I made the function more flexible. (I'm not a bash expert. So if you know a better implementation than using `case`..`esac` then please suggest).

su225:
From test logs - `subject=CN = *..nip.io, O = example organization`
```bash
export INGRESS_HOST=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
```
It seems that the command to get the IP of the ingress gateway is failing as seen from error message. But there are many other tests where similar command is used and they're all fine. This does not seem to be profile dependent as the test with `profile=demo` has also failed.

su225:
Ok. Got it. I always thought that the test cluster had a load balancer. Instead of using the snippet I had to use `_set_ingress_environment_variables` which handles the case without load balancer.

content/en/docs/tasks/observability/gateways/secure_access_test.sh
```diff
@@ -0,0 +1,73 @@
...
+_deploy_and_wait_for_addons kiali prometheus grafana zipkin
+
+# Setup TLS certificates and ingress access
+_set_ingress_environment_variables
+INGRESS_DOMAIN="$INGRESS_HOST.nip.io"
+INGRESS_URL="$INGRESS_DOMAIN:$SECURE_INGRESS_PORT"
+
+snip_option_1_secure_access_https_1
+
+snip_option_1_secure_access_https_2
```
frankbu, line 39: This and a lot of other snips have corresponding _out variables. You should use them to confirm that the expected output in the doc is correct.

```suggestion
_verify_same snip_option_1_secure_access_https_2 "$snip_option_1_secure_access_https_2_out"
```

su225: Added it back...

PR #8044, opened 2020-08-27
Deployment models - clarifications and fixes

rshriram:
Cleaned up a bit of incorrect concepts in the doc, and align the explanations to match with the current state of Istio.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -29,15 +29,11 @@ or are there multiple control planes deployed to ensure high availability (HA)?
 Are all clusters going to be connected into a single {{< gloss >}}multicluster{{< /gloss >}}
 service mesh or will they be federated into a {{< gloss >}}multi-mesh{{< /gloss >}} deployment?
 
-All of these questions, among others, represent independent dimensions of configuration for an Istio deployment.
+Deploying a `single Istio mesh` involves three independent dimensions of configuration:
 
 1. single or multiple cluster
 1. single or multiple network
 1. single or multiple control plane
-1. single or multiple mesh
-
-All combinations are possible, although some are more common than others and
-some are clearly not very interesting (for example, multiple mesh in a single cluster).
 
 In a production environment involving multiple clusters, you can use a mix
 of deployment models. For example, having more than one control plane is recommended for HA,
@@ -123,66 +119,7 @@ You can configure inter-cluster communication based on the
 example, if two clusters reside on the same underlying network, you can enable
 cross-cluster communication by simply configuring firewall rules.
 
-### DNS with Multiple Clusters
-
-When a client application makes a request to some host, it must first perform a
-DNS lookup for the hostname to obtain an IP address before it can proceed with
-the request.
-
-In Kubernetes, the DNS server residing within the cluster typically handles
-this DNS lookup. Kubernetes `Service` configures the DNS server.
-
-Istio ignores the IP address returned from the DNS lookup, however. Instead, it
-uses the list of active endpoints for the host and load balances across them.
-Istio uses either Kubernetes `Service`/`Endpoint` or Istio `ServiceEntry` to
-configure its internal mapping of hostname to IP addresses.
-
-This two-tiered naming system becomes more complicated when you have multiple
-clusters. Istio is inherently multicluster-aware, but Kubernetes is not
-(today). Because of this, the client cluster must have a DNS entry for the
-service in order for the DNS lookup to succeed, and a request to be
-successfully sent. This is true even if there are no instances of that
-service's Pods running in the client cluster.
-
-To ensure that DNS lookup succeeds, you must deploy a Kubernetes `Service` to
-each cluster that consumes that service. This ensures that regardless of
-where the request originates, it will pass DNS lookup and be handed to Istio
-for proper routing.
-
-This can also be achieved with Istio `ServiceEntry`, rather than Kubernetes
-`Service`. `ServiceEntry` does not configure DNS, however. This means that
-DNS will have to be configured either manually or with automated tooling,
-such as the [Istio CoreDNS Plugin](https://github.com/istio-ecosystem/istio-coredns-plugin).
-
-{{< tip >}}
-There are a few efforts in progress that will help simplify the DNS story:
-
-- Istio will soon support DNS interception for all workloads with a sidecar
-  proxy. This will allow Istio to perform DNS lookup on behalf of the
-  application.
-
-- [Admiral](https://github.com/istio-ecosystem/admiral) is an Istio community
-  project that provides a number of multicluster capabilities, including
-  automatic creation of service DNS entries.
-
-- [Kubernetes Multi-Cluster Services](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api)
-  is a Kubernetes Enhancement Proposal (KEP) that defines an API for exporting
-  services to multiple clusters. This effectively pushes the responsibility of
-  service visibility and DNS resolution for the entire `clusterset` onto
-  Kubernetes. There is also work in progress to build layers of `MCS` support
-  into Istio, which would allow Istio to work with any cloud vendor `MCS`
-  controller or even act as the `MCS` controller for the entire mesh.
-
-While `Admiral` is available today, the Istio and Kubernetes communities are
-actively building more general solutions into their platforms. Stay tuned!
-{{< /tip >}}
-
 ## Network models
-
-Istio uses a simplified definition of {{< gloss >}}network{{< /gloss >}} to
-refer to {{< gloss >}}workload instance{{< /gloss >}}s that have direct
-reachability. For example, by default all workload instances in a single
-cluster are on the same network.
 
 Many production systems require multiple networks or subnets for isolation
 and high availability. Istio supports spanning a service mesh over a variety of
@@ -232,26 +169,12 @@ network of the consumers.
     caption="A service mesh with multiple networks"
     >}}
 
-This solution requires exposing all services (or a subset) through the gateway.
-Cloud vendors may provide options that will not require exposing services on
-the public internet. Such an option, if it exists and meets your requirements,
-will likely be the best choice.
-
-{{< tip >}}
-In order to ensure secure communications in a multi-network scenario, Istio
-only supports cross-network communication to workloads with an Istio proxy.
-This is due to the fact that we expose services at Ingress Gateway with TLS
-pass-through, which enables mTLS directly to the workload. A workload without
-an Istio proxy, however, will likely not be able to participate in mutual
-authentication with other workloads. For this reason, Istio filters
-out-of-network endpoints for proxyless services.
-{{< /tip >}}
-
 ## Control plane models
 
 An Istio mesh uses the {{< gloss >}}control plane{{< /gloss >}} to configure all
-communication between workload instances within the mesh. Workload instances
-connect to a control plane instance to get their configuration.
+communication between workload instances within the mesh. At one extreme, you can choose
+to deploy one control plane per cluster in the mesh, while at the other extreme,
+you can host one control plane for all clusters in the mesh.
 
 In the simplest case, you can run your mesh with a control plane on a single
 cluster.
@@ -263,36 +186,11 @@ cluster.
     caption="A single cluster with a control plane"
     >}}
 
-A cluster like this one, with its own local control plane, is referred to as
-a {{< gloss >}}primary cluster{{< /gloss >}}.
+A cluster like this one, with its own local control plane, is referred to as a {{< gloss >}}primary cluster{{< /gloss >}}.
 
-Multicluster deployments can have any number of primary clusters. Each primary
-cluster receives its configuration (i.e. `Service`/`ServiceEntry`,
-`DestinationRule`, etc.) from the Kubernetes API Server residing in the same
-cluster. Each primary cluster, therefore, has an independent source of
-configuration.
-
-This duplication of configuration across primary clusters does require
-additional steps when rolling out changes, however. Large production
-systems may automate this process with tooling, such as CI/CD systems, in
-order to manage configuration rollout.
-
-There may be use cases where having different configurations among primary
-clusters is desirable. For example, this affords the ability to canary
-configuration changes in a sub-section of the mesh controlled by a
-given primary cluster. Alternatively, visibility of a service could be
-restricted to part of the mesh, helping to establish service-level isolation.
-
-A {{< gloss >}}remote cluster{{< /gloss >}} refers to a cluster that contains
-user workloads but no control plane. Remote clusters rely on a control plane
-residing outside of the cluster (e.g. in a primary cluster). In this case, the
-control plane must be made accessible to the remote cluster.
-
-This can be achieved by exposing the control plane through an Istio Gateway.
-Cloud vendors may provide options, such as internal load balancers, for
-providing this sort of capability without exposing the control plane on the
-public internet. Such an option, if it exists and meets your requirements,
-will likely be the best choice.
+Multicluster deployments can also share control plane instances. In this case,
+the control plane instances can reside in one or more primary clusters.
+Clusters without their own control plane are referred to as {{< gloss "remote cluster" >}}remote clusters{{< /gloss >}}.
 
 {{< image width="75%"
     link="shared-control.svg"
@@ -301,10 +199,9 @@ will likely be the best choice.
     caption="A service mesh with a primary and a remote cluster"
     >}}
 
-Instead of running control planes in primary clusters inside the mesh, a
-service mesh composed entirely of remote clusters can be controlled by an
-{{< gloss >}}external control plane{{< /gloss >}}. This provides isolated
-management and complete separation of the control plane deployment from the
+Instead of running control planes in primary clusters inside the mesh, a service mesh composed entirely of
+remote clusters can be controlled by an {{< gloss >}}external control plane{{< /gloss >}}.
+This provides isolated management and complete separation of the control plane deployment from the
 data plane services that comprise the mesh.
 
 {{< image width="100%"
@@ -314,11 +211,10 @@ data plane services that comprise the mesh.
     caption="A single cluster with an external control plane"
     >}}
 
-A cloud vendor's {{< gloss >}}managed control plane{{< /gloss >}} is a
-typical example of an external control plane.
+A cloud vendor's {{< gloss >}}managed control plane{{< /gloss >}} is a typical example of an external control plane.
 
-For high availability, you should deploy a control plane across multiple
-clusters, zones, or regions.
+For high availability, you should deploy one control plane in each zone/region or set of clusters, and
+ensure that configuration is replicated across all control planes.
 
 {{< image width="75%"
     link="multi-control.svg"
@@ -330,74 +226,75 @@ clusters, zones, or regions.
 This model affords the following benefits:
 
 - Improved availability: If a control plane becomes unavailable, the scope of
-  the outage is limited to only that control plane.
+  the outage is limited to only workloads in clusters managed by that control plane.
 
-- Configuration isolation: You can make configuration changes in one cluster,
-  zone, or region without impacting others.
+- Improved configuration isolation: Configuration changes in one control plane affects only
+the workloads in clusters managed by this control plane.
 
-- Controlled Rollout: You have more fine-grained control over configuration
-  rollout (e.g. one cluster at a time). For large meshes with many clusters,
-  this process can be automated with CI/CD pipelines.
+### Control Plane Per Cluster
 
-- Selective Service Visibility: You can choose which `Service`/`ServiceEntry`
-  resources to make visible to each control plane. This can be used to
-  establish service-level isolation between clusters. For example, an
-  administrator may choose to deploy the `HelloWorld` service to Cluster A,
-  but not Cluster B. Any attempt to call `HelloWorld` from Cluster B will
-  fail the DNS lookup.
+For maximum availability and configuration isolation you can deploy a control plane in each cluster. Each cluster's control plane
+manages only the workloads in its own cluster. If the control plane is unavailable (due to upgrade or
+temporary outage), the workloads will continue to function with the existing sidecar configuration and
+will re-attempt to connect to the in-cluster control plane periodically.
 
-The following list ranks control plane deployment examples by availability:
+This model affords the following benefits:
 
-- One cluster per region (**lowest availability**)
-- Multiple clusters per region
-- One cluster per zone
-- Multiple clusters per zone
-- Each cluster (**highest availability**)
+- Highest level of availability: If a control plane becomes unavailable, the scope of
+  the outage is limited to only that cluster and the workloads within it.
 
-### Endpoint Discovery with Multiple Control Planes
+- Maximum configuration isolation: You can make configuration changes in one cluster without impacting
+  workloads deployed in any other cluster.
 
-An Istio control plane manages traffic within the mesh by providing each proxy
-with the list of service endpoints. In order to make this work in a
-multicluster scenario, each control plane must observe endpoints from the API
-Server in every cluster.
+### Availability vs Config Isolation
 
-To enable endpoint discovery for a cluster, an administrator generates a
-`remote secret` and deploys it to each primary cluster in the mesh. The
-`remote secret` contains credentials, granting access to the API server in the
-cluster.
+The following list ranks control plane deployment examples by availability and configuration isolation:
 
-The control planes will then connect and discover the service endpoints for
-the cluster, enabling cross-cluster load balancing for these services.
+- One control plane per region or a single global control plane (**lowest availability**, **lowest isolation**)
+- Multiple control planes per region
+- One control plane per cluster (**highest availability**, **highest isolation**)
 
-{{< image width="75%"
-    link="endpoint-discovery.svg"
-    caption="Primary clusters with endpoint discovery"
+## Mesh models
+
+Istio supports having all of your services in a
+{{< gloss "service mesh" >}}single mesh{{< /gloss >}}, or federating multiple meshes
+together, which is also known as {{< gloss >}}multi-mesh{{< /gloss >}}.
+
+### Single mesh
+
+The simplest Istio deployment is a single mesh. A single mesh can span
+[one or more clusters](#cluster-models) and [one or more networks](#network-models).
+All workloads in a mesh share a common certificate authority. You can choose to use a single shared
+  [trust domain](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#21-trust-domain) or a separate trust domain for each cluster.
+If you use a single shared trust domain, a service account in a given namespace is globally unique - i.e. if
+two clusters `cluster-1` and `cluster-2` have the same service account `foo` in namespace `ns`, they will be
+treated the same for purposes of security. With a trust domain for each cluster, there are no
+restrictions on the uniqueness of service accounts, and namespaces. You can apply Istio Authorization
+policies that allow or deny all traffic from one cluster (irrespective of the namespace/service) to another
+based on the source principals of incoming traffic.
+
+### Multiple meshes
+
+Multiple mesh deployments result from {{< gloss >}}mesh federation{{< /gloss >}}, with the following capabilities:
+
+- Organizational boundaries: lines of business.
+- Potential for cross vendor meshes to communicate with each
+  other if the trust bundles are setup properly.
+
+You can enable inter-mesh communication with {{< gloss >}}mesh federation{{</gloss >}}.
+When federating, each mesh can expose a set of services and identities, which all participating meshes can recognize.
+
+{{< image width="50%"
+    link="multi-mesh.svg"
+    alt="Multiple service meshes"
+    title="Multi-mesh"
+    caption="Multiple service meshes"
     >}}
 
-By default, Istio will load balance requests evenly between endpoints in
-each cluster. In large systems that span geographic regions, it may be
-desirable to use [locality load balancing](/docs/ops/configuration/traffic-management/locality-load-balancing)
-to prefer that traffic stay in the same zone or region.
-
-In some advanced scenarios, load balancing across clusters may not be desired.
-For example, in a blue/green deployment, you may deploy different versions of
-the system to different clusters. In this case, each cluster is effectively
-operating as an independent mesh. This behavior can be achieved in a couple of
-ways:
-
-- Do not exchange remote secrets between the clusters. This offers the
-  strongest isolation between the clusters.
-
-- Use `VirtualService` and `DestinationRule` to disallow routing between two
-  versions of the services.
-
-In either case, cross-cluster load balancing is prevented. External traffic
-can be routed to one cluster or the other using an external load balancer.
-
-{{< image width="75%"
-    link="blue-green.svg"
-    caption="Blue-green deployment without cross-cluster load balancing"
-    >}}
+When federating two meshes that do not share the same certificate authority, you must
+{{< gloss "mesh federation">}}federate{{< /gloss >}}
+{{< gloss >}}identity{{< /gloss >}} and exchange **trust bundles** between them. See the
+section on [Trust between meshes](#trust-between-meshes) for an overview.
 
 ## Identity and trust models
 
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -29,15 +29,18 @@ or are there multiple control planes deployed to ensure high availability (HA)?
 Are all clusters going to be connected into a single {{< gloss >}}multicluster{{< /gloss >}}
 service mesh or will they be federated into a {{< gloss >}}multi-mesh{{< /gloss >}} deployment?
 
-All of these questions, among others, represent independent dimensions of configuration for an Istio deployment.
+## Definition of a Mesh in Istio
+
+Before answering the questions above, it is imperative to understand the definition of a `mesh` in Istio.
+
+_A mesh is one single administrative domain, consisting of one single shared root CA trusted by
```
smawson, line 36: A mesh can actually have multiple roots, so long as all clusters trust the same set of roots it is a single mesh.

I'm not even sure how to define a mesh anymore, since it reality Istio is really flexible and even service discovery doesn't have to be the same in every cluster.

But I think administrative domain / shared set of trusted roots / single service discovery / service naming domain is kind of the right direction?

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -176,9 +179,16 @@ network of the consumers.
...
 An Istio mesh uses the {{< gloss >}}control plane{{< /gloss >}} to configure all
-communication between workload instances within the mesh. You can replicate the
-control plane, and workload instances connect to any control plane instance to
-get their configuration.
+communication between workload instances within the mesh. At one extreme, you can choose
+to deploy one control plane per cluster in the mesh, while at the other extreme,
+you can host one single shared control plane for all clusters in the mesh.
+
+_Note: Each control plane will only honor configurations (such as Gateways, VirtualServices, etc.)
+that are present in its local Kubernetes cluster. If you wish to apply a configuration to all workloads in the mesh
```
smawson, line 187: This is no longer true with the external control plane stuff, there is a concept of a config cluster where an external control plane can pull its config. So we should maybe add a parenthetical that mentions that external control planes pull their config from a single config cluster.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -176,9 +179,16 @@ network of the consumers.
...
+across control planes, it is your responsibility to replicate the configuration to the
+Kubernetes clusters associated with each control plane._
+
+### Shared control plane models
```
smawson, line 191: I think this should move below the single cluster simple case and go right before "Multicluster deployments can also share control plane instances."

A single cluster mesh isn't really a shared control plane.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,15 +240,20 @@ clusters, zones, or regions.
 This model affords the following benefits:
 
 - Improved availability: If a control plane becomes unavailable, the scope of
-  the outage is limited to only that control plane.
+  the outage is limited to only that control plane, and the workloads in clusters managed by that control plane.
```
smawson, line 243:
```suggestion
  the outage is limited to only workloads in clusters managed by that control plane.
```

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
     caption="A service mesh after a control plane instance fails"
     >}}
 
-The following list ranks control plane deployment examples by availability:
+### Independent Control Planes
+
+You can also deploy a control plane per cluster, with a shared root CA. Each cluster's control plane
```
smawson, line 267: I'd like to reword this a tiny bit to make it more part of a continuum, so rather than "also"something like:

For maximum availability and configuration isolation you can deploy a control plane in each cluster.

We don't need to mention shared root CA since that is the same for shared control planes and independent control planes.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+
+This model affords the following benefits:
+
+- Highest level of availability: If a control plane becomes unavailable, the scope of
+  the outage is limited to only that cluster and the workloads within it.
+
+- Maximum configuration isolation: You can make configuration changes in one cluster without impacting
+  workloads deployed in any other cluster.
+
+The following list ranks control plane deployment examples by availability, configuration isolation:
```
smawson, line 280: Maybe a new section since this isn't part of independent control planes.

smawson, line 280:
```suggestion
The following list ranks control plane deployment examples by availability and configuration isolation:
```

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+
+- One control plane per region (**lowest availability**, **lowest isolation**)
```
smawson, line 282: One before this is a single global control plane.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+
+### Single mesh
+
+The simplest Istio deployment is a single mesh. A single mesh can span
+[one or more clusters](#cluster-models) and [one or more networks](#network-models).
+All workloads in a mesh share a common certificate authority or the same trust
+domain depending on the following situations:
+
+- If you are using a vendor hosted CA that is shared across different customers,
+  then it is imperative to use a single trust domain for the entire mesh, as the
```
smawson, line 302: This is not necessarily true (for example google supports a single mesh with multiple trust domains using our managed CA).

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+  root CA across different customers could be the same (i.e. the vendor's CA itself).
+
+- If you have your own root CA for all your clusters, you can choose to use a single shared
```
smawson, line 305: We're more general than this, we can support multiple roots and multiple trust domains for each root. Or at least that was the design...

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+  trust domain or a separate trust domain for each cluster.
+  
+If you use a single shared trust domain, a service in a given namespace is globally unique - i.e. if
```
smawson, line 308: This should talk specifically about service accounts or service identities, not just services.

Also we support aliasing so even if two clusters have the same service account foo in namespace ns but are in different trust domains, they _may_ be treated the same if the trust domains are aliased.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -299,95 +378,30 @@ those identities.
...
-## Tenancy models
-
-In Istio, a **tenant** is a group of users that share
-common access and privileges to a set of deployed workloads. Generally, you
-isolate the workload instances from multiple tenants from each other through
-network configuration and policies.
-
-You can configure tenancy models to satisfy the following organizational
-requirements for isolation:
+Istio does not have a formally defined notion of tenancy. However, you can achieve a great
```
smawson, line 383: Is this really true? Our notion of tenancy was defined in this doc, which is that namespaces are used for soft tenancy. Also we have lots of code that helps make namespace isolation work. So I'm not sure we don't have a formal notion of tenancy. Maybe we can just drop that first sentence.

You can achieve a great level of isolation across different teams ...

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -396,49 +410,9 @@ to restrict access to only the appropriate callers.
...
-    caption="Two isolated service meshes with two clusters and two namespaces"
-    >}}
-
-Since a different team or organization operates each mesh, service naming
-is rarely distinct. For example, the `mysvc` in the `foo` namespace of
-`cluster-1` and the `mysvc` service in the `foo` namespace of
-`cluster-2` do not refer to the same service. The most common example is the
-scenario in Kubernetes where many teams deploy their workloads to the `default`
-namespace.
+### Cluster isolation
```
smawson, line 413: This used to be more about using meshes as tenancy (it was probably mistitled). 

I think we want to do more work to make cluster tenancy actually work well, and also think we should talk about using multiple meshes as tenancy.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -299,95 +378,30 @@ those identities.
     caption="Multiple service meshes with certificate authorities"
     >}}
 
-## Mesh models
+## Isolating teams within the mesh
```
smawson, line 381: I think we should talk about multi-mesh here too, so re-title as Isolating teams with Istio or something.

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: Is this the right place for this definition? We already have [What is a Service Mesh](https://istio.io/latest/docs/concepts/what-is-istio/#what-is-a-service-mesh) ... maybe this re-work should go there instead and we could link to it?

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -176,9 +179,16 @@ network of the consumers.
 ## Control plane models
 
 An Istio mesh uses the {{< gloss >}}control plane{{< /gloss >}} to configure all
-communication between workload instances within the mesh. You can replicate the
-control plane, and workload instances connect to any control plane instance to
-get their configuration.
+communication between workload instances within the mesh. At one extreme, you can choose
+to deploy one control plane per cluster in the mesh, while at the other extreme,
+you can host one single shared control plane for all clusters in the mesh.
```
nmittler, line 184: nit: suggest rephrasing to `you can host one control plane that is shared by all clusters in the mesh`

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,15 +240,20 @@ clusters, zones, or regions.
...
+- Improved configuration isolation: Configuration changes in one control plane affects only
+the workloads in clusters managed by this control plane.
 
-You can improve control plane availability through fail over. When a control
-plane instance becomes unavailable, workload instances can connect to
-another available control plane instance. Fail over can happen across clusters,
-zones, or regions.
+#### Failover in Shared Control Planes
+
+You can improve shared control plane availability through fail over. When a shared control
```
nmittler, line 250: Does failover across cluster boundaries actually work today?

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -217,8 +227,8 @@ data plane services that comprise the mesh.
 
 A cloud vendor's {{< gloss >}}managed control plane{{< /gloss >}} is a typical example of an external control plane.
 
-For high availability, you should deploy a control plane across multiple
-clusters, zones, or regions.
+For high availability, you should deploy one shared control plane in each zone/region or sets of clusters, and
```
nmittler, line 230: remove `shared` ... that's not a term we want to use anymore.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -203,7 +213,7 @@ Clusters without their own control plane are referred to as {{< gloss "remote cl
     caption="A service mesh with a primary and a remote cluster"
     >}}
 
-Instead of running control planes in primary clusters inside the mesh, a service mesh composed entirely of
+Instead of running shared control planes in primary clusters inside the mesh, a service mesh composed entirely of
```
nmittler, line 216: remove `shared`

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
     caption="A service mesh after a control plane instance fails"
     >}}
 
-The following list ranks control plane deployment examples by availability:
+### Independent Control Planes
```
nmittler, line 265: This term has not been defined or agreed upon. We probably shouldn't use it. In the old terminology, this would have effectively been "replicated". We had previously considered using the term `independent control planes` to represent the case where both are primaries and there is no cross-cluster load balancing between the clusters

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,15 +240,20 @@ clusters, zones, or regions.
...
-- Configuration isolation: You can make configuration changes in one cluster,
-  zone, or region without impacting others.
+- Improved configuration isolation: Configuration changes in one control plane affects only
+the workloads in clusters managed by this control plane.
 
-You can improve control plane availability through fail over. When a control
-plane instance becomes unavailable, workload instances can connect to
-another available control plane instance. Fail over can happen across clusters,
-zones, or regions.
+#### Failover in Shared Control Planes
```
nmittler, line 248: Rephrase to just `Failover in Control Planes`

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
     caption="A service mesh after a control plane instance fails"
     >}}
 
-The following list ranks control plane deployment examples by availability:
+### Independent Control Planes
+
+You can also deploy a control plane per cluster, with a shared root CA. Each cluster's control plane
+manages only the workloads in its own cluster. If the control plane is unavailable (due to upgrade or
+temporary outage), the workloads will continue to function with the existing sidecar configuration and
+will re-attempt to connect to the local control plane periodically.
```
nmittler, line 270: replace `local` with something like `in-cluster`

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+
+The following list ranks control plane deployment examples by availability, configuration isolation:
+
+- One control plane per region (**lowest availability**, **lowest isolation**)
+- Multiple control planes per region
+- One control plane per zone
+- Multiple control planes per zone
+- One control plane per cluster (**highest availability**, **highest isolation**)
+
+## Mesh models
```
nmittler, line 288: Can we move Mesh models back to where it was? It's built on top of `trust models`, so I'd like to see that come first.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+{{< gloss >}}identity{{< /gloss >}} and **trust bundles** between them. See the
+section on [Multiple Trust Domains](#trust-between-meshes) for an overview.
 
 ## Identity and trust models
 
 When a workload instance is created within a service mesh, Istio assigns the
 workload an {{< gloss >}}identity{{< /gloss >}}.
 
-The Certificate Authority (CA) creates and signs the certificates used to verify
+The Certificate Authority (CA) in the control plane creates and signs the certificates used to verify
```
nmittler, line 343: This wording seems to imply `citadel` (i.e. part of istiod). Was there some motivation for adding this?

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: Isn't this just a rename of `Tenancy Models`? I'm not sure this is an improvement ... or am I missing something?

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -299,95 +378,30 @@ those identities.
...
-Istio supports two types of tenancy models:
+Istio supports two types of soft-tenancy models:
 
-- [Namespace tenancy](#namespace-tenancy)
-- [Cluster tenancy](#cluster-tenancy)
+- [Namespace isolation](#namespace-isolation)
+- [Cluster isolation](#cluster-isolation)
 
-### Namespace tenancy
+### Namespace isolation
```
nmittler, line 398: Why is `isolation` preferred to just `tenancy`?

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: does each mesh have its own trust bundle ?

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: Yes but the implementation is not there. there is quite a bit of handwaving to say 'make sure to replicate trust bundles everywhere' using some SPIFFE design doc but nothing to say how to keep these bundles updated as you add more clusters/CAs. That is the reason i chose to keep this concept doc concrete as I had one end user ask me today on how to accomplish the trust bundle replication described in the doc.

Listing in concepts but having no implementation feels a bit like rhetorical trick :P.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: Right, thats different trust domains. the sentence is about single trust domain

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: what does it even mean? and why does it matter to a single Istio mesh? I mean the persona doing this is viewing the doc from the perspective of a single mesh, based on what is possible today in Istio. We dont have examples/guides for multi-mesh or guides explaining how to do trust bundle updates, replication, etc.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: I have absolutely no clue if it works. But it was listed in the doc (you can actually see it live on our site)!. So I am simply repeating that.. Heck it even has diagrams!

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -29,15 +29,18 @@ or are there multiple control planes deployed to ensure high availability (HA)?
...
 
-All of these questions, among others, represent independent dimensions of configuration for an Istio deployment.
+## Definition of a Mesh in Istio
+
+Before answering the questions above, it is imperative to understand the definition of a `mesh` in Istio.
+
+_A mesh is one single administrative domain, consisting of one single shared root CA trusted by 
+all workloads in the mesh, irrespective of the clusters or control planes in the system._
+
+All of these questions, among others, represent independent dimensions of configuration for a `single Istio mesh` deployment.
```
frankbu, line 39:
```suggestion
Deploying a `single Istio mesh` involves three independent dimensions of configuration:
```

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: as I said, I dont know if the failover works or is supported. It just seemed hand wavy. I can use replicated or shared. But its what it is.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: yes but trust models is also talking about multi-mesh.. So its chicken and egg.

content/en/docs/ops/deployment/deployment-models/index.md
frankbu: Even if you move this header after the single cluster simple case, it still isn't appropriate because the external control plane example, later on, is also not a shared control plane model. Do we really need this heading at all?

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: Because we dont have a formal tenancy concept. No one knows what a tenant is. Docs have been flying around about this "tenancy" thing. Isolation != tenancy. We shouldn't oversell this bad :)

content/en/docs/ops/deployment/deployment-models/index.md
frankbu: Agree with @nmittler. You could have a mesh with multiple primary clusters, so there's no sharing in that case.

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: +1 for removing the heading

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: I'd be in favor of removing it from the doc.

@smawson WDYT?

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: Then can we move it back for diff purposes at least? We can reorganize in a follow-up if we think it's necessary.

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: Still I think I'd rather keep the tenancy docs as-is until we finalize the new conceptual model for tenancy. I'm not sure this rename is an improvement.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: Should we add external control plane here as well?

content/en/docs/ops/deployment/deployment-models/index.md
linsun: +1, local k8s cluster is really our new 'config cluster' now, as it doesn't have to be local.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -176,9 +179,16 @@ network of the consumers.
...
-control plane, and workload instances connect to any control plane instance to
-get their configuration.
+communication between workload instances within the mesh. At one extreme, you can choose
+to deploy one control plane per cluster in the mesh, while at the other extreme,
+you can host one single shared control plane for all clusters in the mesh.
+
+_Note: Each control plane will only honor configurations (such as Gateways, VirtualServices, etc.)
+that are present in its local Kubernetes cluster. If you wish to apply a configuration to all workloads in the mesh
+across control planes, it is your responsibility to replicate the configuration to the
+Kubernetes clusters associated with each control plane._
```
linsun, line 189: Think this `it is your responsibility to replicate the configuration` - do you mean istio config or k8s objects for services?  I assume it is later for shared control plane and perhaps both for replicated control plane... this can generate confusion to users.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: is there workload outage when istio config/resources aren't changed?  Seems odd we would highlight workloads outage here.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: would be good to remove this before we have steps and validation.  Last time we discussed, i think mainly @costinm knows how to do this.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: I do like independent control planes way better than replicated control planes :)

content/en/docs/ops/deployment/deployment-models/index.md
linsun: +1, the key is config isolation.  Perhaps max availability as well.  (especially if we remove shared control plane with failover).

content/en/docs/ops/deployment/deployment-models/index.md
linsun: Note: in this model, the control plane could be external as well.

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: Yeah I think that makes sense. We can probably just add that in a follow-on PR.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+  workloads deployed in any other cluster.
+
+The following list ranks control plane deployment examples by availability, configuration isolation:
+
+- One control plane per region (**lowest availability**, **lowest isolation**)
+- Multiple control planes per region
+- One control plane per zone
+- Multiple control planes per zone
+- One control plane per cluster (**highest availability**, **highest isolation**)
+
```
linsun, line 287: What is confusing here is cluster can be multi-zone as well.  So users could be deploy control plane per zone in a multi-zone cluster.   In fact, we recommend users to use large clusters with zone HA - https://istio.io/latest/docs/ops/best-practices/deployment/#deploy-across-multiple-availability-zones

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: I'm not sure that's true. It sounds like the description here is specifically regarding in-cluster control planes.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: Also, I could have combination of these... i can have multiple control planes per region, and within each region, i have have one cp per zone.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: do we really want to put vendor statement in istio.io docs here?

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -247,20 +262,85 @@ zones, or regions.
...
+[one or more clusters](#cluster-models) and [one or more networks](#network-models).
+All workloads in a mesh share a common certificate authority or the same trust
+domain depending on the following situations:
+
+- If you are using a vendor hosted CA that is shared across different customers,
+  then it is imperative to use a single trust domain for the entire mesh, as the
+  root CA across different customers could be the same (i.e. the vendor's CA itself).
+
+- If you have your own root CA for all your clusters, you can choose to use a single shared
+  trust domain or a separate trust domain for each cluster.
```
linsun, line 306: Do we support separate trust domain for a single mesh?  Any examples?  Looking at our replicated control plane doc, we are using same root CA, and configuring auto_passthrough on the cluster gateways, so it is still single trust domain imho.

linsun: Or you consider our replicated control plane is different trust domains with same root CA?

ericvn:
If we need these changes in `master` we need the cherrypick label set.

content/en/docs/ops/deployment/deployment-models/index.md
frankbu: The `What is a Service Mesh` section of the concepts is very high level, so I don't think we want this there, but I do think we should clean up the [service mesh glossary entry](https://preliminary.istio.io/latest/docs/reference/glossary/#service-mesh) should say this, and here we can refer to and highlight the single admin domain part

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: this is addressing the google CA thing.. as this restriction in google CA is now making the security people to suggest that each cluster should have its own trust domain (and yet share a root CA)

rshriram: and still be different meshes!!!!!

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: we certainly support separate trust domains per cluster but with same CA. Heck, we are running this in production at several customer sites. its there in our docs - some obscure trust domain flag/field that people can set at control plane level.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: fixed

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,30 +230,76 @@ clusters, zones, or regions.
...
+
+### Single mesh
 
-- Configuration isolation: You can make configuration changes in one cluster,
-  zone, or region without impacting others.
+The simplest Istio deployment is a single mesh. A single mesh can span
+[one or more clusters](#cluster-models) and [one or more networks](#network-models).
+All workloads in a mesh share a common certificate authority. You can choose to use a single shared
+  trust domain or a separate trust domain for each cluster.
+
```
frankbu, line 273:
```suggestion

```

content/en/docs/ops/deployment/deployment-models/index.md
smawson: The target audience is a platform admin. They need to know how to deal with tenancy within their platform. One option is cluster tenancy, which doesn't quite work right by being overly tied to control plane deployments, but is something we should work on improving. Another option is mesh tenancy. We should discuss these models here, if we're going to discuss one model that doesn't quite work right we should discuss the other too :)

content/en/docs/ops/deployment/deployment-models/index.md
nmittler: The problem is that it's not defined and we haven't reached agreement on the term. There doesn't really seem to be a need for a new term. We could just as easily call this section `Control plane per cluster`.

nmittler (commented):
I have added content to this doc that now overlaps a bit with this PR. In particular, the "independent control plane" section duplicates some of the content in the control planes intro that discusses each primary having a separate source or config.

Should probably give the. new page a read and see if any re-arranging or merging of content is needed.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: Okay changed to cluster tenancy and also added mesh tenancy

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,30 +226,75 @@ clusters, zones, or regions.
...
+The simplest Istio deployment is a single mesh. A single mesh can span
+[one or more clusters](#cluster-models) and [one or more networks](#network-models).
+All workloads in a mesh share a common certificate authority. You can choose to use a single shared
+  trust domain or a separate trust domain for each cluster.
+If you use a single shared trust domain, a service account in a given namespace is globally unique - i.e. if
+two clusters `cluster-1` and `cluster-2` have the same service account `foo` in namespace `ns`, they will be
+treated the same for purposes of security. With a trust domain for each cluster, there are no
+restrictions on the uniqueness of service accounts, and namespaces. You can apply enforcement
+policies that allow or deny all traffic from one cluster (irrespective of the namespace/service) to another.
```
linsun, line 273: @rshriram @frankbu what are the examples of enforcement policies here?  Istio RBAC or network policies?  I assume Istio RBAC doesn't work across clusters when there are multiple trust domains.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: Thanks found it in meshconfig trust domain.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,30 +226,75 @@ clusters, zones, or regions.
...
+Istio supports having all of your services in a
+{{< gloss "service mesh" >}}single mesh{{< /gloss >}}, or federating multiple meshes
+together, which is also known as {{< gloss >}}multi-mesh{{< /gloss >}}.
+
+### Single mesh
+
+The simplest Istio deployment is a single mesh. A single mesh can span
+[one or more clusters](#cluster-models) and [one or more networks](#network-models).
+All workloads in a mesh share a common certificate authority. You can choose to use a single shared
+  trust domain or a separate trust domain for each cluster.
```
linsun, line 268:
```suggestion
  [trust domain](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#21-trust-domain) or a separate trust domain for each cluster.
```

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,30 +226,75 @@ clusters, zones, or regions.
...
+- One control plane per cluster (**highest availability**, **highest isolation**)
+
+## Mesh models
+
+Istio supports having all of your services in a
+{{< gloss "service mesh" >}}single mesh{{< /gloss >}}, or federating multiple meshes
+together, which is also known as {{< gloss >}}multi-mesh{{< /gloss >}}.
+
+### Single mesh
+
```
linsun, line 264: For single mesh, if clusters have different trust domain for each cluster, why aren't those multi-meshes?  Seems it is intuitive to use trust domain to scope the boundary of mesh.

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,30 +226,75 @@ clusters, zones, or regions.
...
-zones, or regions.
+You can enable inter-mesh communication with {{< gloss >}}mesh federation{{</gloss >}}.
+When federating, each mesh can expose a set of services and identities, which all participating meshes can recognize.
 
 {{< image width="50%"
-    link="failover.svg"
-    alt="A service mesh after a control plane instance fails"
-    title="Control plane fail over"
-    caption="A service mesh after a control plane instance fails"
+    link="multi-mesh.svg"
```
linsun, line 287: Confused about this diagram, how is this different than single mesh multiple trust domains mentioned earlier where it also says `You can apply enforcement
policies that allow or deny all traffic from one cluster (irrespective of the namespace/service) to another`

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -230,30 +226,75 @@ clusters, zones, or regions.
...
-- One cluster per region (**lowest availability**)
-- Multiple clusters per region
-- One cluster per zone
-- Multiple clusters per zone
-- Each cluster (**highest availability**)
+When federating two meshes that do not share the same
+{{< gloss >}}trust domain{{< /gloss >}}, you must
+{{< gloss "mesh federation">}}federate{{< /gloss >}}
+{{< gloss >}}identity{{< /gloss >}} and **trust bundles** between them. See the
+section on [Multiple Trust Domains](#trust-between-meshes) for an overview.
```
linsun, line 297: Here we are referring to multiple trust domains to a link that is trust between meshes.  but earlier the doc says you could have multiple trust domains for a single mesh (each cluster has its own trust domain).

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: as i mentioned earlier, it does and we use it in production.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: Why does trust domain (an opaque string) have to be tied to the CA or mesh? Intuitive is very subjective. To me, it is unintuitive to run single Istio for many clusters or allow pods in one cluster to access internal k8s services of another cluster :). Yet, we have many formulations of this (single/independent/central control planes). The trust domain falls under the same umbrella. You can use a single global trust domain for entire mesh (like central istiod) or many trust domains (one per cluster). Its just essentially build your own adventure, like the rest of this doc.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: Why does trust domain (an opaque string) have to be tied to the CA or mesh? Intuitive is very subjective. To me, it is unintuitive to run single Istio for many clusters or allow pods in one cluster to access internal k8s services of another cluster :). Yet, we have many formulations of this (single/independent/central control planes). The trust domain falls under the same umbrella. You can use a single global trust domain for entire mesh (like central istiod) or many trust domains (one per cluster). Its just essentially build your own adventure, like the rest of this doc.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: good catch.. will remove

content/en/docs/ops/deployment/deployment-models/index.md
```diff
@@ -396,49 +372,16 @@ to restrict access to only the appropriate callers.
...
+them to manage their own configurations. Alternatively, if you can ensure that namespaces
+are globally unique across clusters, you can use a shared control plane.
 
-Since a different team or organization operates each mesh, service naming
-is rarely distinct. For example, the `mysvc` in the `foo` namespace of
-`cluster-1` and the `mysvc` service in the `foo` namespace of
-`cluster-2` do not refer to the same service. The most common example is the
-scenario in Kubernetes where many teams deploy their workloads to the `default`
-namespace.
+### Mesh Tenancy
```
nmittler, line 382: So it appears that you removed the `Mesh models` section in favor of using mesh as a unit of tenancy. I'm not entirely against this as it does seem to simplify this a bit.

@smawson thoughts?

content/en/docs/ops/deployment/deployment-models/index.md
linsun: Can we clarify it is Istio authZ policy (vs a general enforcement policies) if that is what you are referring to here?

Note: we don't have examples for this at istio.io or in our github.  So this could be perceived as unsupported.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: Maybe it doesn't... What I'm trying to understand is how can I differentiate if this is single mesh or multi-meshes, based on your latest doc. From our prior doc, my understanding is that it is based on trust domains and whether admin needs to federate trust bundles.  Now, I don't know, and that is the confusion I had.

content/en/docs/ops/deployment/deployment-models/index.md
linsun: I realize you may not have changed diagrams... you reshaped what is single mesh that could include multiple trust domains, which made me think this diagram isn't only applied to multi-meshes based on the latest doc thus the confusion.

linsun:
I think the update looks good in general. I'm very much in favor of removing contents that we don't officially support or at least provide some clear indication when something is not supported.

I do think we need clarity on single mesh (when there is multiple trust domains) vs multi-meshes.

content/en/docs/ops/deployment/deployment-models/index.md
rshriram: done

nmittler (commented):
@rshriram I think you need to apply this to the `master` branch first and then cherry-pick into `release-1.7`. This PR doesn't have my recent changes.

frankbu:
Outdated PR. Superseded by #8487 and #8503

PR #8086, opened 2020-09-05
Doc test for accessing Jaeger and Grafana dashboards by port-forwarding

su225:
Very similar to Zipkin - https://preliminary.istio.io/latest/docs/tasks/observability/distributed-tracing/jaeger/
Accessing Grafana dashboards (mesh and service) - https://preliminary.istio.io/latest/docs/tasks/observability/metrics/using-istio-dashboard/

**NOTE**: We cannot verify screenshots (but still possible to verify API calls and results. But not sure, if that needs to be done, or is covered in some integration test in istio.io/istio repo)

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/observability/distributed-tracing/jaeger/test.sh
+++ b/content/en/docs/tasks/observability/distributed-tracing/jaeger/test.sh
@@ -39,7 +39,7 @@ snip_accessing_the_dashboard_1 &
 # Although test says, take a look at traces, we don't have to do that in this task
 # as it is covered by an integration test in istio/istio.
 function access_jaeger_by_port_forward() {
-  curl -s -o /dev/null -w '%{http_code}' http://localhost:16686/jaeger
+  curl -s -o /dev/null -w '%{http_code}' "http://localhost:16686/jaeger/api/traces?service=productpage.default&limit=20"
 }
 
 _verify_same access_jaeger_by_port_forward "200"
--- a/content/en/docs/tasks/observability/distributed-tracing/jaeger/test.sh
+++ b/content/en/docs/tasks/observability/distributed-tracing/jaeger/test.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC1090,SC2154,SC2155,SC2034
+
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -u
+set -o pipefail
+
+source "tests/util/helpers.sh"
+source "tests/util/samples.sh"
+source "tests/util/addons.sh"
+
+# @setup profile=default
+
+# NOTE: This test is very similar to the one for zipkin.
+_deploy_and_wait_for_addons jaeger
+
+kubectl label namespace default istio-injection=enabled --overwrite
+startup_bookinfo_sample
+_set_ingress_environment_variables
+GATEWAY_URL="$INGRESS_HOST:$INGRESS_PORT"
+bpsnip_trace_generation__1
+
+snip_accessing_the_dashboard_1 &
+
+# Although test says, take a look at traces, we don't have to do that in this task
+# as it is covered by an integration test in istio/istio.
+function access_jaeger_by_port_forward() {
+  curl -s -o /dev/null -w '%{http_code}' http://localhost:16686/jaeger
+}
+
+_verify_same access_jaeger_by_port_forward "200"
+pgrep istioctl | xargs kill
+
+# @cleanup
+set +e
+
+# TODO: Fix issue of killing twice (https://github.com/istio/istio.io/issues/8014)
+pgrep istioctl | xargs kill
+cleanup_bookinfo_sample
+_undeploy_addons jaeger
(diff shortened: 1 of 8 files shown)


content/en/docs/tasks/observability/distributed-tracing/jaeger/test.sh
```diff
@@ -0,0 +1,54 @@
...
+_set_ingress_environment_variables
+GATEWAY_URL="$INGRESS_HOST:$INGRESS_PORT"
+bpsnip_trace_generation__1
+
+snip_accessing_the_dashboard_1 &
+
+# Although test says, take a look at traces, we don't have to do that in this task
+# as it is covered by an integration test in istio/istio.
+function access_jaeger_by_port_forward() {
+  curl -s -o /dev/null -w '%{http_code}' http://localhost:16686/jaeger
```
frankbu, line 42: In the zipkin test your curl was to `local zipkin_url='http://localhost:9411/zipkin/api/v2/traces?serviceName=productpage.default'` but here it's just `http://localhost:16686/jaeger`.

I agree you don't need to look at the traces in any detail, but I think (like you did in the zipkin test) you should at least confirm that there is an expected `productpage` trace.

content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
```diff
@@ -0,0 +1,62 @@
...
+kubectl label namespace default istio-injection=enabled --overwrite
+startup_bookinfo_sample
+_set_ingress_environment_variables
+GATEWAY_URL="$INGRESS_HOST:$INGRESS_PORT"
+bpsnip_trace_generation__1
+
+snip_viewing_the_istio_dashboard_3 &
+
+# For verification, we only check if the dashboards are accessible, but not its actual contents
+# TODO: Is it worth checking API calls and output for Grafana case?
```
frankbu, line 43: Yes, similar to my comment for jaeger, above, I think you should generate some traffic (as the doc describes), and then do some minimal check to confirm that something was generated. We don't need to confirm it's exactly correct, leave that to the istio/istio integration tests, but just do a basic sanity check to confirm that logging did happen as a result of the generated traffic.

content/en/docs/tasks/observability/distributed-tracing/jaeger/test.sh
su225: Fixed it.

content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
su225: I added some checks. I had to dig into the queries used by Grafana to contact Prometheus to populate the dashboard. It is long and ugly. I have added it anyways.

content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
```diff
@@ -0,0 +1,102 @@
...
+GATEWAY_URL="$INGRESS_HOST:$INGRESS_PORT"
+
+bpsnip_trace_generation__1
+
+snip_viewing_the_istio_dashboard_3 &
+
+# For verification, we only check if the dashboards are accessible, but not its actual contents
+# TODO: Is it worth checking API calls and output for Grafana case? 
+function access_grafana_istio_mesh_dashboard() {
+  curl -L -s -o /dev/null -w '%{http_code}' "http://localhost:3000/dashboard/db/istio-mesh-dashboard"
```
frankbu, line 46: Instead of this curl followed by the long encoded curl, can you just check for the expected text by curling the dashboard itself?
```suggestion
  curl -L -s "http://localhost:3000/dashboard/db/istio-mesh-dashboard"
```
Then change the the call, see below.

content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
```diff
@@ -0,0 +1,102 @@
...
+  | jq -r '.data.result[].metric.destination_workload_var' | sort
+}
+
+function query_request_from_productpage_workload() {
+  curl -L -s 'http://localhost:3000/api/datasources/proxy/1/api/v1/query?query=%20sum(istio_requests_total%7Breporter%3D%22source%22%2C%20source_workload%3D~%22productpage-v1%22%2C%20source_workload_namespace%3D~%22default%22%7D)%20by%20(destination_service)%20or%20sum(istio_tcp_sent_bytes_total%7Breporter%3D%22source%22%2C%20source_workload%3D~%22productpage-v1%22%2C%20source_workload_namespace%3D~%22default%22%7D)%20by%20(destination_service)' \
+  | jq -r '.data.result[].metric.destination_service' | sort
+}
+
+_verify_same access_grafana_istio_mesh_dashboard "200"
+_verify_lines query_request_for_all_workloads "
```
frankbu, line 76:
```suggestion
_verify_lines access_grafana_istio_mesh_dashboard "
```

content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
su225: image

There is a problem through. It seems like this is an Angular app. So all it does is load angular.js, app.js and other vendor scripts, then fetches data by API calls and then renders it. **So rendering is done by client-side JavaScript after fetching data from the API**. So we should either live with long and ugly URLs or remove and just check if we can access dashboard.

content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
su225: Not sure if we should remove those checks. As mentioned before, dashboards are rendered by client-side JavaScript. So grepping for DOM elements does not work.

content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
frankbu: That's too bad. Thanks for trying. It would have been a lot cleaner if we could. So, let's leave it as you have it for now.

frankbu (approved):

PR #8088, opened 2020-09-05
[WIP] Enhance prow/** to set up for multicluster tests

su225:
Enhance `prow/**` to set up multiple clusters (distinguished by `TOPOLOGY` parameter)

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/prow/lib-multicluster.sh
+++ b/prow/lib-multicluster.sh
@@ -1,0 +1,199 @@
+#!/bin/bash
+
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# shellcheck source=prow/lib.sh
+source "${ROOT}/prow/lib.sh"
+
+# Cluster names for multicluster configurations.
+export CLUSTER1_NAME=${CLUSTER1_NAME:-"cluster1"}
+export CLUSTER2_NAME=${CLUSTER2_NAME:-"cluster2"}
+export CLUSTER3_NAME=${CLUSTER3_NAME:-"cluster3"}
+
+export CLUSTER_NAMES=("${CLUSTER1_NAME}" "${CLUSTER2_NAME}" "${CLUSTER3_NAME}")
+export CLUSTER_POD_SUBNETS=(10.10.0.0/16 10.20.0.0/16 10.30.0.0/16)
+export CLUSTER_SVC_SUBNETS=(10.255.10.0/24 10.255.20.0/24 10.255.30.0/24)
+
+export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
+
+# Cleans up the clusters created by setup_kind_clusters
+function cleanup_kind_clusters() {
+  for c in "${CLUSTER_NAMES[@]}"; do
+     cleanup_kind_cluster "${c}"
+  done
+}
+
+function setup_kind_clusters() {
+  TOPOLOGY="${1}"
+  IMAGE="${2:-kindest/node:v1.18.2}"
+
+  KUBECONFIG_DIR="$(mktemp -d)"
+
+  # The kind tool will error when trying to create clusters in parallel unless we create the network first
+  # TODO remove this when kind support creating multiple clusters in parallel - this will break ipv6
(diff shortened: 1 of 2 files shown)


su225:
@nmittler - Please review. Much of multi-cluster setup is copied from the other the main istio repo.

su225:
Output of dummy test
```
+source tests/util/helpers.sh
+__cluster_snapshot
+__cluster_snapshot(): __cluster_state
+__cluster_state(): kubectl get ns -o name
+__cluster_state(): sed /istio-system/d
+__cluster_state(): kubectl get all --ignore-not-found -n default
+__cluster_state(): kubectl get istiooperators --ignore-not-found -n default
+__cluster_state(): kubectl get destinationrules --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get envoyfilters --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get gateways --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get serviceentries --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get sidecars --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get virtualservices --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get workloadentries --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get authorizationpolicies --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get peerauthentications --ignore-not-found -n default -n istio-system
+__cluster_state(): kubectl get requestauthentications --ignore-not-found -n default -n istio-system
+set -e
+set -u
+set -o pipefail
+echo 'start multicluster test with KUBECONFIG=/home/a269f991,/home/a07ba57c,/home/63aaf8e0'
+IFS=,
+read -r -a CONFIGS
+for kcfg in "${CONFIGS[@]}"
+kubectl --kubeconfig=/home/a269f991 get pods -A
+for kcfg in "${CONFIGS[@]}"
+kubectl --kubeconfig=/home/a07ba57c get pods -A
+for kcfg in "${CONFIGS[@]}"
+kubectl --kubeconfig=/home/63aaf8e0 get pods -A
```

pkg/test/istioio/framework.go
```diff
@@ -184,14 +193,18 @@ func NeedSetup(config string) bool {
 // TestDocs traverses through all test cases and runs those that need the
 // setup config specified by the input. The (*testing.T) variable comes
 // from the TestDocs function in each tests/setup/*/doc_test.go
-func TestDocs(t *testing.T, config string) {
+func TestDocs(t *testing.T, config string, multicluster bool) {
```
nmittler, line 196: I don't think we want to fish `multicluster` through the test framework.  If we assume that we have a separate prow job for the multicluster tests, a better approach might be to just have that job select the specific tests to run via `TEST` (see https://github.com/istio/istio.io/tree/master/tests).

prow/lib-multicluster.sh
```diff
@@ -0,0 +1,199 @@
+#!/bin/bash
```
nmittler, line 1: @su225 Can we reduce this PR to just the files under the `prow` directory. I'd like to focus the review on just setting up the clusters.

@stevenctl mind taking a look to compare to your work on istio/istio?

su225: Ok. I'll move other changes to another PR.

pkg/test/istioio/framework.go
su225: Anyways this will be moved to another PR.  But I have a few questions
* Are we planning to put multicluster tests into a single directory? (Unlike what we do for single cluster tests)
* If not, then wouldn't we end up updating prow job config for new tests? (like `TEST=dir1` and then `TEST=dir1,dir2`)

su225: Alternatively, make file can be edited to avoid the issue mentioned in second point. But still a hassle.

nmittler: > Are we planning to put multicluster tests into a single directory? (Unlike what we do for single cluster tests)

Yes, the multicluster tests will follow the same directory structure as the site content, so the prow target would use something like `TEST=setup/install/multicluster`

stevenctl:
I wonder if it makes sense to move some of this to common-files. I have a feeling KinD upgrades and other changes should apply to both. At least the `lib*.sh` files and then the main script share the logic from that. Things should be broken down well enough in istio/istio's `lib.sh`, and there are optimizations like the docker registry in kind.

su225:
@stevenctl - Same feeling here. This is a good opportunity to get rid of duplication (both single and multicluster).

su225:
Created https://github.com/istio/istio.io/issues/8097 to get rid of duplication

su225:
This PR is broken down and is done in another PR.
https://github.com/istio/istio.io/pull/8127

su225:
Closing it as #8127 is merged.

PR #8091, opened 2020-09-08
Add note regarding current operator merge behaviour

### LINKED ISSUE #8090: Document Istio operator's merge behaviour
The Istio operator has some undocumented merge behaviour, as per the discussion here: https://github.com/istio/istio/issues/27032

This should be documented, as it's a pretty big issue. PR incoming shortly.

Niksko:

Closes #8090.

Questions for reviewers:

* Does this need an example? The example that lead me to this was adding different affinity settings (required, then preferred) and both of them ending up in the gateway pod spec. However perhaps a simpler example would be better.
* Does this need info on when it will be fixed? It will be fixed when Istio no long supports K8s <1.16 (because then we can use server-side apply)
* Should this provide an alternative e.g. 'If this behaviour is incompatible with your desired Istio installation, consider using X installation method instead'?

[X] Docs
[X] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/operator/index.md
+++ b/content/en/docs/setup/install/operator/index.md
@@ -123,6 +123,12 @@ Now, with the controller running, you can change the Istio configuration by edit
 the `IstioOperator` resource. The controller will detect the change and respond by updating
 the Istio installation correspondingly.
 
+{{< warning >}}
+The current behavior of the Istio operator is to _merge_ configuration of some components rather than to _override_
+configuration. This applies to components that have immutable fields such as the Kubernetes `Service` object's
+`clusterIP` field. This can cause Istio components to display a mix of current configuration and previous configuration.
+{{< /warning >}}
+
 For example, you can switch the installation to the `default`
 profile with the following command:
 


content/en/docs/setup/install/operator/index.md
```diff
@@ -123,6 +123,12 @@ Now, with the controller running, you can change the Istio configuration by edit
 the `IstioOperator` resource. The controller will detect the change and respond by updating
 the Istio installation correspondingly.
 
+{{< warning >}}
+The current behavior of the Istio operator is to _merge_ configuration of some components rather than to _override_
+configuration. This applies to components that have immutable fields such as the Kubernetes `Service` object's
+`clusterIP` field. This can cause Istio components to display a mix of current configuration and previous configuration.
```
frankbu, line 129: Yes, I think this could use a more specific example. What kind of "merge" happens when you try to update a field of a Service object? Is it only the immutable fields that have this issue? What exactly happens if you change the clusterIP of a service field, how does it merge 2 clusterIP valuse?

Probably should also mention a workaround and/or ETA for a fix??
cc @ostromart

ostromart: it's a bit unclear to me what the difference between merge and override is, since merge is an override. an example would be good explaining that distinction.

Niksko: @frankbu @ostromart not the best example, I will remove it. This was the example given to me in the original issue I raised, but it isn't helpful here.

For reference, here is the original issue, the comments add context: https://github.com/istio/istio/issues/27032

Is the example I provided there (modifying pod anti-affinity settings, and ending up with a merged result) a more appropriate example?

As far as a workaround, there really isn't any, which is why I wanted to add the warning. As you can see in the linked issue above, the suggested workaround is to delete the `IstioOperator` resource and create a new one, which clearly isn't feasible if you want your Istio installation to stay up.

Happy to include an ETA for the fix, it's basically whenever Istio drops support for K8s 1.16 (k8s 1.17 and higher have server side apply, which solves this issue), which if I'm not mistaken will be Istio 1.9?

Niksko: Actually, it's possible that this has been resolved. Waiting on info in istio/istio#27032

frankbu:
@Niksko What is the status of this PR? Has the issue been resolved, so we can close it? Otherwise, do you want to update/clarify the example, so we can merge?

frankbu:
Closing because the issue seems to have been addressed.

@Niksko please reopen if I am mistaken.

PR #8103, opened 2020-09-09
Update test reference - master

ericvn:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
Will need some additional source changes to match stuff done in istio/istio.

ericvn:
Bookinfo test is failing because of an earlier my-nginx not being cleaned:

```
 my-nginx      ClusterIP   10.105.13.53     <none>        443/TCP    5m39s
```

ericvn:
Also, per John, https://github.com/istio/istio/pull/27247 may fix the DNS issue, so will try updating to commit with that when available and remove the DNS flag added earlier.

ericvn:
Updated to a commit including the fix noted above and removed flag to work around original issue.

ericvn:
Two failures in egress tests where tests expect `HTTP/1.1 301 Moved Permanently` however they get a `HTTP/1.1 503 Service Unavailable`. Mentioned in #networking, a John says he's has found the breakage (https://github.com/istio/istio/pull/26409). Waiting on a PR and will move to that when ready.

Continuing on with fixing other failures.

ericvn:
One of the failures is getting Sleep pods and using it for a curl. The problems that the deployment is ready when on pod is ready, but the first one is still termination. So the command to get the pod, gets two pods and then the curl fails:
```
"sleep-8f795f47d-gnb4s sleep-8f795f47d-h56qm" not found"
```
Need to think about some logic to not get the terminated pod(s).

ericvn:
The egress test continue to fail. I Slacked with @howardjohn last night and he was able to recreate (on GKE, the tests pass). The current thought it is ipv6 related (although CI is thought to be ipv4 ).

ericvn:
There appears to be an issue in istioctl experimental wait. The fix might take a few days to get through review, etc, so shortening the wait time from 5m so the tests should complete in an hr. Will observe if things are happening too early with a 30s wait.

content/en/docs/tasks/security/authorization/authz-td-migration/index.md
```diff
@@ -86,7 +86,7 @@ Notice that it may take tens of seconds for the authorization policy to be propa
 1. Install Istio with a new trust domain.
 
     {{< text bash >}}
-    $ istioctl install --set profile=demo --set values.global.trustDomain=new-td
+    $ istioctl install --skip-confirmation --set profile=demo --set values.global.trustDomain=new-td
```
shamsher31, line 89: Can we use short notation wherever possible and keep it at the end?

```
$ istioctl install --set profile=demo --set values.global.trustDomain=new-td -y
```

content/en/docs/setup/getting-started/index.md
```diff
@@ -75,7 +75,7 @@ Follow these steps to get started with Istio:
     profiles for production or performance testing.
 
     {{< text bash >}}
-    $ istioctl install --set profile=demo
+    $ istioctl install --skip-confirmation --set profile=demo
```
shamsher31, line 78: Same as https://github.com/istio/istio.io/pull/8103/files#r489327294
```
$ istioctl install --set profile=demo -y
```

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
```diff
@@ -51,7 +51,7 @@ echo '*** istioctl-analyze step 7 ***'
 istioctl analyze --help
 
 echo '*** istioctl-analyze step 8 ***'
-snip_install_with_custom_config_analysis
+echo y | snip_install_with_custom_config_analysis
```
frankbu, line 54: Why are you not using the `yes` command for this?
```suggestion
yes | snip_install_with_custom_config_analysis
```

content/en/docs/tasks/security/authorization/authz-td-migration/index.md
ericvn: Will be picked up in a later cleanup commit, one tests pass. And since they are now passing, it's that time.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
ericvn: Lack of knowing the command was there. Will pick up change(s) in cleanup commit now that tests are working.

tests/util/verify.sh
```diff
@@ -237,7 +237,7 @@ __verify_with_retry() {
         local funcret="$?"
 
         # shellcheck disable=SC2001
-        out=$(sed 's/[[:space:]]*$//g' <<< "$out")
+        out=$(sed -e 's/[[:space:]]*$//g' -e '/./,$!d' <<< "$out")
```
frankbu, line 240: Deleting leading empty lines seems like it will cause confusion. If the output is expected to have empty lines at the start, we should include it in the expected output, or use `_verify_contains` if that's more convenient.

Deleting trailing spaces, on the other hand, is needed. It is solving the problem where sometimes the output has trailing space, but because of the lint checker does not allow trailing spaces in .md files, we can't include it in the expected output.

ericvn: @frankbu I'm fine moving to _verify_contains, but I found no way to actually add the empty lines at the start. If I add them in the index.md, it's carried to the snips, but the _verify_like(I think that's what that was using) still failed.

frankbu: Hmm, that sound like there may be a bug in the verify function. Handling white space properly has always been hard to get right because bash read does some stripping itself. If you can reproduce an example, open an issue for it.

content/en/docs/tasks/security/authentication/mtls-migration/index.md
```diff
@@ -78,7 +78,8 @@ the policies to enforce STRICT mutual TLS between the workloads.
 
     {{< text bash >}}
     $ kubectl get destinationrule --all-namespaces
-    No resources found.
+        NAMESPACE      NAME                      HOST                                    AGE
+        istio-system   meshexpansion-dr-istiod   istiod.istio-system.svc.cluster.local   5m4s
```
frankbu, line 82: Changing this in the doc is not acceptable, since the DR has nothing to do with this task, and is inconsistent with the text above that says make sure there are no existing DRs. I think this is only there if the cluster is configured for multicluster or VM integration. I don't think our test cluster for these tasks should be, but if they will be, we should just change the test.sh to handle this, not the .md (unless we want to update the doc to explain better).

ericvn: Well, the cluster isn't configured for multicluster or VM, unless the test case is doing it.

Will need to see if it's always there.

ericvn: I wonder if the test framework changing to allow multicluster is what's causing this. Maybe need to move from default to none and do the istioctl install would be appropriate.

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
ericvn: Tried the change to `yes |` and the tests fail:
```
2020-09-16T15:23:24.676474Z	info	tf	Running command script tasks_security_authorization_authz-td-migration_test
    script.go:111: script tasks_security_authorization_authz-td-migration_test returned an error: exit status 141. Output:
        This will install the Istio profile into the cluster. Proceed? (y/N) 
        - Processing resources for Istio core.
        ✔ Istio core installed
        - Processing resources for Istiod.
        - Processing resources for Istiod. Waiting for Deployment/istio-system/istiod
        ✔ Istiod installed
        - Processing resources for Egress gateways, Ingress gateways.
        - Processing resources for Egress gateways, Ingress gateways. Waiting for Deployment/istio-system...
        ✔ Egress gateways installed
        ✔ Ingress gateways installed
        - Pruning removed resources
        ✔ Installation complete
```
Trying the command locally yields a SIGPIPE(13) failure.

frankbu: Ah, so it's probably failing because we run the tests with `set -o pipefail`. I don't know of that's really needed.

content/en/docs/tasks/security/authorization/authz-td-migration/index.md
ericvn: Fixed

content/en/docs/setup/getting-started/index.md
ericvn: Fixed

tests/util/verify.sh
ericvn: Removed and will fix the failing test case. Will add TODO: and comment here in case someone wants to see about a bug in the verify.

content/en/docs/tasks/security/authentication/mtls-migration/index.md
ericvn: Removed, removed verify. TODO: added to verify ways around having the extra DR in the test (when the docs say one won't be there).

content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh
ericvn: I'm not sure about changing the pipefail. Can we do it file wide? Or is this something we would change around each invocation? For now, `echo y` is being used and working.

frankbu: Not worth changing it around every invocation, better to use `echo y` in that case, but maybe we could change it file wide. It's just a convention we've been using in the test.sh files, but I'm not sure if it's needed or not. Anyway, another thing to investigate when we have some free time.

tests/util/verify.sh
ericvn: Interestingly, the error with the empty line either fixed itself as it's not reappearing after the change.

nmittler (approved):

frankbu (approved):

PR #8110, opened 2020-09-09
update docs to point to 1.6 release

jacob-delgado:
Announce Istio 1.6.9

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/news/releases/1.6.x/announcing-1.6.9/index.md
+++ b/content/en/news/releases/1.6.x/announcing-1.6.9/index.md
@@ -1,0 +1,41 @@
+---
+title: Announcing Istio 1.6.9
+linktitle: 1.6.9
+subtitle: Patch Release
+description: Istio 1.6.9 patch release.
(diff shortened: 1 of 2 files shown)


ericvn:
I'm not sure this is necessarily needed. If we merge this change and then rebuild the 1.6 archive, I believe the page would show up here: https://istio.io/v1.6/news/releases/1.6.x/. (note that this is the page if you select to look at the 1.6 version of the documentation). I could see where that might be nice, hasn't been done in the past, and would have to verify if the new archive building tool would create the appropriate changes in the release-1.7 branch for this file.

The data/args.yml file wouldn't be used as it's not copied into the archives directory.

jacob-delgado:
Closing. We don't need to point to the latest release as long as it is on the main istio site. We should be fine.

PR #8104, opened 2020-09-09
Update test reference for 1.7.1

ericvn:
Please provide a description for what this PR is for.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #8118, opened 2020-09-10
Bump istio dependency

howardjohn:
Not sure if I need to do more for this or just go.mod?


ericvn:
@howardjohn Looks like you are trying to so similar to my #8104 which is to move to the latest Istio/istio. I haven't looked at the required changes as the test framework has undergone some changes.

howardjohn:
closing in favor of https://github.com/istio/istio.io/pull/8104. Make sure you match the `replace` statements here

ericvn:
Thanks John. Picked those changes up in a commit.

PR #8131, opened 2020-09-11
Adding test targets for each profile

nmittler:
Also bumping up timeout slightly as a temporary fix for flakes.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/tests/tests.mk
+++ b/tests/tests.mk
@@ -1,8 +1,9 @@
-export TIMEOUT ?= 40m
+DEFAULT_TIMEOUT=60m
+export TIMEOUT ?= ${DEFAULT_TIMEOUT}
 
 # gocache disabled by -count=1
 # tests in different packages forced to be sequential by -p=1
-%.test: init | $(JUNIT_REPORT)
+doc.test.%: init | $(JUNIT_REPORT)
 	@${GO} test ${REPO_ROOT}/tests/setup/$*/... \
 		-v -timeout=${TIMEOUT} -count=1 -p=1 \
 		-istio.test.hub=$(HUB) \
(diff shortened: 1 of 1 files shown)


tests/tests.mk
```diff
@@ -1,4 +1,13 @@
-export TIMEOUT ?= 30m
+export TIMEOUT ?= 40m
```
ericvn, line 1: Lets make this 60 minutes. And use the value in the line specifying default below

tests/tests.mk
```diff
@@ -1,4 +1,13 @@
...
+
+# gocache disabled by -count=1
+# tests in different packages forced to be sequential by -p=1
+%.test: init | $(JUNIT_REPORT)
```
ericvn, line 5: I'd also be inclined to prefix with `doc.` like the others. Otherwise we could remove doc. from the front of help since doc.test will go away.

ericvn (approved):

adammil2000 (approved):

frankbu:
@nmittler @ericvn Why is this necessary? Doesn't having N separate test jobs just complicate the build?

ericvn:
@frankbu I don't necessarily want to spread things out further, but we were seeing bucket 1 hit a 30 min timeout, then bucket 2 hit a timeout, and finally bucket 3 would finish, so maybe 1:20 for length of the test cases. Bumping the timeout to 60 mins, meant that if we still timed out, tests would increase to 2:20. To prevent a really long running tests, there are now 3, each with will take a max of 60 mins. So, including timeouts, the longest we should wait for test completion would be 1:00.

Once we get things in a better state, we could potentially move them back, but note that the time to run the tests will be approximately the total time of the three individual tests.

PR #8162, opened 2020-09-17
Expand deployment models to include multicluster considerations

nmittler:
This is a prerequisite for the new multicluster installation guide.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,179 @@
...
+
+By default, Istio will load balance requests evenly between endpoints in each cluster. In large systems that span
+geographic regions, it may be desirable to use locality load balancing to prefer that traffic stay in the same zone
+or region.
+
+In some advanced scenarios, load balancing across clusters may not be desired. For example, in a blue/green
+deployment, you may deploy different versions of the system to different clusters. In this case, each cluster is
+effectively operating as an independent mesh. This behavior can be achieved in a couple of ways:
+
+Do not exchange remote secrets between the clusters. This offers the strongest isolation between the clusters.
```
smawson, line 72: Need to make these a numbered list or bullet list or something.

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,179 @@
...
+uses the list of active endpoints for the host and load balances across them. Istio uses either Kubernetes
+`Service`/`Endpoint` or Istio `ServiceEntry` to configure its internal mapping of hostname to IP addresses.
+
+This two-tiered naming system becomes more complicated when you have multiple clusters. Istio is inherently
+multicluster-aware, but Kubernetes is not (today). Because of this, for a client in one cluster to make a request to
+a service in another cluster, there must be a Kubernetes service for the host defined in the client cluster in order
+for the DNS lookup to succeed. This is true even if there are no instances of that service's Pods running in the
+client cluster.
+
+To generalize this, you must deploy a Kubernetes service to each cluster that consumes that service. This ensures that
```
smawson, line 117: Should we mention that there are istio ecosystem projects that help with this (admiral) and also k8s is working on addressing this (MCS)? Otherwise it feels like we're just leaving this as an exercise to the reader instead of helping them.

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,179 @@
...
+{{< /tip >}}
+
+{{< image width="80%"
+    link="east-west.svg"
+    caption="Multiple networks with services exposed through gateway"
+    >}}
+
+### Understanding `MeshNetworks`
+
+The `MeshNetworks` configuration within the `IstioOperator` CRD, provides information about all of the networks
```
smawson, line 146:
```suggestion
The `MeshNetworks` configuration within the `IstioOperator` CRD provides information about all of the networks
```

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,179 @@
...
+within the mesh. It is effectively a map of where the network IDs are the keys.
```
smawson, line 147:
```suggestion
within the mesh. It is effectively a map where the network IDs are the keys.
```

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,179 @@
...
+A workload without an Istio proxy, however, will likely not be able to participate in mutual authentication with
+other workloads. For this reason, Istio filters out-of-network endpoints for proxyless services.
+{{< /tip >}}
+
+{{< image width="80%"
+    link="east-west.svg"
+    caption="Multiple networks with services exposed through gateway"
+    >}}
+
+### Understanding `MeshNetworks`
```
smawson, line 144: This seems different than the rest of this page. I suggest moving it to the multicluster installation guide, since it is more practical than conceptual, and isn't really a consideration. Also it is cruft from us not making this as easy as we should and will go away :)

nmittler:
@smawson PTAL

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+other using an external load balancer.
+
+{{< image width="80%"
+    link="blue-green.svg"
+    caption="Blue-green deployment without cross-cluster load balancing"
+    >}}
+
+### Remote Clusters
+
+A {{< gloss >}}remote cluster{{< /gloss >}} refers to a cluster that contains user workloads but no control plane.
```
stevenctl, line 88: Is this doc only valid once we get istiod-less remotes? Or we need to clarify that "control plane" in this context only refers to serving xDS, but not Certs/Injection.

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Configuration
+
+An Istio control plane in a {{< gloss >}}primary cluster{{< /gloss >}} receives its configuration
+(i.e. `Service`/`ServiceEntry`, `DestinationRule`, etc.) from the Kubernetes API Server residing in the same
+{{< gloss >}}cluster{{< /gloss >}}.
+
+When a mesh contains multiple primary clusters, each control plane has an independent source of configuration.
+
+{{< image width="80%"
```
stevenctl, line 29: Should this image clarify that we're reading _istio_ configuration from the local api server? Could be confusing later when we introduce reading k8s objects from remote api servers.

stevenctl: Actually, it's plenty clear in that diagram.

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+Remote clusters rely on a control plane residing either in a primary cluster or
+{{< gloss "external control plane" >}}external to the mesh{{< /gloss >}}. In either case, the control plane must be
+made accessible to the remote cluster.
+
+Cloud vendors may provide options, such as internal load balancers, for providing this sort of capability
+without exposing the control plane on the public internet. Such an option, if it exists and meets your requirements,
+will likely be the best choice.
+
+{{< image width="80%"
+    link="remote-clusters.svg"
```
stevenctl, line 98: nit: can we show where discovery comes from for the remote proxy?

stevenctl (approved):
Just the one question about the remotes – but overall this will be really helpful to understand the different deployment models!

smawson (approved):

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+In Kubernetes, the DNS server residing within the cluster typically handles this DNS lookup. Kubernetes
+services configure this server.
+
+Istio ignores the IP address returned from the DNS lookup, however. Instead, it
+uses the list of active endpoints for the host and load balances across them. Istio uses either Kubernetes
+`Service`/`Endpoint` or Istio `ServiceEntry` to configure its internal mapping of hostname to IP addresses.
+
+This two-tiered naming system becomes more complicated when you have multiple clusters. Istio is inherently
+multicluster-aware, but Kubernetes is not (today). Because of this, for a client in one cluster to make a request to
+a service in another cluster, there must be a Kubernetes service for the host defined in the client cluster in order
```
craigbox, line 116:
```suggestion
a service in another cluster, there must be a Kubernetes `Service` for the host defined in the client cluster in order
```

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+for the DNS lookup to succeed. This is true even if there are no instances of that service's Pods running in the
+client cluster.
+
+To generalize this, you must deploy a Kubernetes service to each cluster that consumes that service. This ensures that
```
craigbox, line 120:
```suggestion
To generalize this, you must deploy a Kubernetes `Service` to each cluster that consumes that service. This ensures that
```

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+weight: 20
+keywords:
+  - multiple-clusters
+  - control-plane
+  - networks
+owner: istio/wg-environments-maintainers
+test: n/a
+---
+
+Before you install Istio on {{< gloss "multicluster" >}}multiple clusters{{< /gloss >}}, there are some things to
```
linsun, line 13: Before you expand Istio to multiple clusters?

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+This model has the following advantages:
+
+* Reduces the blast radius of bad configuration and helps to maintain high availability for the control plane.
+* Allows administrators more fine-grained control over configuration rollout (e.g. one cluster at a time). For large
+meshes with many clusters, this process can be automated with CI/CD pipelines.
+* Allows selective exposure of `Service`/`ServiceEntry` across cluster boundaries. For example, an administrator may
+choose to deploy the HelloWorld service to Cluster A, but not Cluster B. This is an advanced use case, but may be
+desirable for stronger isolation between clusters.
+
+This duplication of configuration across primary clusters can lead to issues, however. In particular, Kubernetes
```
linsun, line 43: "can lead to issues" - is this too strong?  seems a red flag but it is not.   I think it is more like:
This duplication of configuration across primary clusters needs additional steup, either by deploying the service to the consuming cluster or create service entry on the consuming cluster.

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+* Use `VirtualService` and `DestinationRule` to disallow routing between two versions of the services.
+
+In either case, cross-cluster load balancing is prevented. External traffic can be routed to one cluster or the
+other using an external load balancer.
+
+{{< image width="80%"
+    link="blue-green.svg"
+    caption="Blue-green deployment without cross-cluster load balancing"
+    >}}
+
```
linsun, line 85: Regarding blue green deployment, is this still 1 single mesh? I could argue this is 2 meshes.

My preference is to remove endpoint discovery section.  It is great detail and I love it, I just think it is not a consideration for multicluster istio.  endpoint discovery is required else it is not 1 mesh.

content/en/docs/ops/deployment/multicluster-considerations/index.md
nmittler: That was my thought. I'm trying to be generic here, so that we don't fumble around to much WRT "types" of remote clusters. So yeah, I'm considering the source of XDS as the control plane in this case.

content/en/docs/ops/deployment/multicluster-considerations/index.md
nmittler: That would be assuming that they're starting with a single cluster mesh. I think the current wording is slightly more generic in that it doesn't assume any particular starting state.

content/en/docs/ops/deployment/multicluster-considerations/index.md
nmittler: Funny thing ... I *think* that I added this section based on your feedback in the gdoc :). You're right, of course... it is effectively a separate mesh and I say so above. I do believe this still belongs here, however, because blue-green deployment is often how multicluster upgrades are performed.

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,155 @@
...
+In a multi-network scenario, each cluster will have to expose any services needed by workloads in other networks.
+Cloud vendors may provide options that will not require exposing services on the public internet. Such an option,
+if it exists and meets your requirements, will likely be the best choice.
+
+{{< tip >}}
+Istio currently only supports cross-network communication to workloads with an Istio proxy. This is due to the
+fact that we expose services at Ingress Gateway with TLS pass-through, which enables mTLS directly to the workload.
+A workload without an Istio proxy, however, will likely not be able to participate in mutual authentication with
+other workloads. For this reason, Istio filters out-of-network endpoints for proxyless services.
```
linsun, line 149: This is good content!  wonder if we position as this is by design to ensure the traffic is secured among clusters.   When service A call service B across clusters, service A and B must both have istio proxies to ensure the mtls communication among the two.

linsun:
This is great contents, Nathan!  thanks for putting this together!  LGTM in general.  My only comment is in a few places about the DNS lookup + must create the service in the consuming cluster.   Also if we want to expose endpoint discover to user as a choice.

content/en/docs/ops/deployment/multicluster-considerations/index.md
nmittler: Good point. Reworded the first sentence a bit ... PTAL

content/en/docs/ops/deployment/multicluster-considerations/index.md
linsun: My bad if I was asking different things. :-(  Maybe I think i was asking how to fold the replicated control plane piece in while we want to maintain the boundary of the clusters and allow admins to selectively expose and consume services acorss clusters.  

Looking at your preview url, I didn't get you are saying this is separate mesh (it is marked as 1 mesh in the diagram).  If there is a case this is useful for multicluster upgrade, i think we should highlight it here.  It is not obvious to me atm when reading it.  I don't understand how this relates to multicluster from users' perspective as to when to choose not to enable endpoint discovery.  Are you saying you recommend undo "create-remote-secrete' on the to-be-upgraded cluster?

nmittler: ah good point WRT the diagram. Thing is, we *still* don't have a really solid definition of when it's a single mesh vs multi-mesh. In the case of blue-green, they very well could be using the same root of trust ... in which case, *maybe* they're still conceptually a single mesh?

@smawson WDYT?

content/en/docs/ops/deployment/multicluster-considerations/index.md
nmittler: I've added some clarification to the DNS section regarding `ServiceEntry`.

ericvn:
For the lint error, I opened a `make shell`, did a `make lint` and then looked at the generate file. The line in question is:
```
        <div class="content">There are a few efforts in progress that will help simplify the DNS story in the future:</p>
```

ericvn:
Things looked OK in a `make serve` or the netlify output.

I tried a few things, and maybe @frankbu knows the rules, but indenting in the tip seemed to pass the lint locally and still look correct::
```
{{< tip >}}
    There are a few efforts in progress that will help simplify the DNS story in the future:

    * Istio will soon support DNS interception for workloads with a sidecar proxy. This will allow Istio to perform
    DNS lookup on behalf of the application.
    * [Admiral](https://github.com/istio-ecosystem/admiral) is an Istio community project that provides a number of
    multicluster capabilities, including automatic creation of service DNS entries.
    * [Kubernetes Multi-Cluster Services](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api)
    is a Kubernetes Enhancement Proposal (KEP) that defines an API for exporting services to multiple clusters.
{{< /tip >}}
```

ericvn:
Shucks on the lint error. Let me see if I can look at it some more.

nmittler:
@ericvn FYI I've pinged @davidhauck offline ... he's going to dig into it.

ericvn:
I wasn't able to come up with a solution that looked good to me. Something might be able to be fixed that allows lists inside a _tip_ @davidhauck.

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,161 @@
...
+    * [Admiral](https://github.com/istio-ecosystem/admiral) is an Istio community project that provides a number of
+    multicluster capabilities, including automatic creation of service DNS entries.
+    * [Kubernetes Multi-Cluster Services](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api)
+    is a Kubernetes Enhancement Proposal (KEP) that defines an API for exporting services to multiple clusters.
+{{< /tip >}}
+
+## Networks
+
+Istio uses a simplified definition of [network](/docs/ops/deployment/deployment-models#network-models) to refer to
+pods that have direct reachability. For example, by default all the pods in a single cluster are on a single network.
```
frankbu, line 141: This sounds strange. Where's the `simplified definition of network`? The wording makes it expected at the link target. Maybe there should be a glossary entry with this simplified definition? Or maybe just say it here:
```suggestion
All pods in an Istio [network](/docs/ops/deployment/deployment-models#network-models) can reach each other directly. For example, by default all the pods in a single cluster are on a single network.
```

nmittler: @frankbu I'm not too sure about the revised wording. It makes it sound that Istio implements a network, rather than just defining a term/concept.

frankbu:
This is good information, but I'm not sure about how it's organized, esp. relative to the Deployment Models doc. It seems to have some duplicate information, and some stuff in the two docs seem like they should really be co-located. It sort of reads like just some random selection of notes, that belong in other places, as opposed to a self-contained standalone doc. 

 I'm OK with merging this for now, in preliminary, but I think it needs to be reorganized along with the Deployment Models doc, in a followup PR, before we publish on istio.io. For example, maybe much of this (along with much of what's in Deployment Models currently, should be moved into a Control Planes doc, as part of a more general restructure. I also wonder if all this deployment model stuff should move from Operations to Concepts?

content/en/docs/ops/deployment/multicluster-considerations/index.md
frankbu: I suppose that's true. Maybe we really need to add a glossary entry for network with the "Istio definition". Then, we could use your wording, but using the glossary reference, instead of a link to the other doc

content/en/docs/ops/deployment/multicluster-considerations/index.md
```diff
@@ -0,0 +1,161 @@
...
+{{< tip >}}
+    There are a few efforts in progress that will help simplify the DNS story in the future:
+
+    * Istio will soon support DNS interception for workloads with a sidecar proxy. This will allow Istio to perform
+    DNS lookup on behalf of the application.
+    * [Admiral](https://github.com/istio-ecosystem/admiral) is an Istio community project that provides a number of
+    multicluster capabilities, including automatic creation of service DNS entries.
+    * [Kubernetes Multi-Cluster Services](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api)
+    is a Kubernetes Enhancement Proposal (KEP) that defines an API for exporting services to multiple clusters.
+{{< /tip >}}
```
frankbu, line 136: I'm not sure exactly what the generator doesn't like, but I think that using `-` (instead of `*`) is preferred. Also I think it might need space between the list entries (just when nested in a tip???). Anyway, I think the following will work:
```suggestion
{{< tip >}}
There are a few efforts in progress that will help simplify the DNS story in the future:

- Istio will soon support DNS interception for workloads with a sidecar proxy. This will allow Istio to perform
  DNS lookup on behalf of the application.

- [Admiral](https://github.com/istio-ecosystem/admiral) is an Istio community project that provides a number of
  multicluster capabilities, including automatic creation of service DNS entries.

- [Kubernetes Multi-Cluster Services](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api)
  is a Kubernetes Enhancement Proposal (KEP) that defines an API for exporting services to multiple clusters.
{{< /tip >}}
```
Note that you'll also need to change your other lists to `-`, because mixed list types are not allowed.

nmittler:
@frankbu @smawson @stevenctl @linsun PTAL

I've merged this content into `deployment models`, since there was a lot of overlap. We should probably eventually find a new home for deployment models ... possibly several docs under `concepts`?

content/en/docs/ops/deployment/multicluster-considerations/index.md
nmittler: I've added a glossary entry in this PR.

craigbox:
> I've merged this content into `deployment models`, since there was a lot of overlap. We should probably eventually find a new home for deployment models ... possibly several docs under `concepts`?

We are working towards an information architecture review/change later this year, so my suggestion is, get the content wherever you can, and it will be audited and moved to the right place in time.

content/en/docs/reference/glossary/network.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Network
+test: n/a
+---
+
+Istio uses a simplified definition of Network based on general connectivity.
+[Workload instances](/docs/reference/glossary/#workload-instance) are on the same
+Network if they able to communicate directly, without a gateway.
```
frankbu, line 8:
```suggestion
Istio uses a simplified definition of network based on general connectivity.
[Workload instances](/docs/reference/glossary/#workload-instance) are on the same
network if they are able to communicate directly, without a gateway.
```

frankbu (approved):

PR #8163, opened 2020-09-17
Try different timeout for tasks/security/authorization/authz-ingress

ericvn:
I'm seeing this test fail frequently. I suspect it's do to a smaller timeout. I had 1/2 failures locally with 30s timeout. Increasing to 60s, I had 0/5 failures.

I am not seeing messages that we have hit the timeout value (which was a typical indication earlier), so this might not be the issue. It may just be an additional small amount of time for connections or other configuration to take effect.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/tests/util/helpers.sh
+++ b/tests/util/helpers.sh
@@ -50,19 +50,15 @@ _wait_for_deployment() {
 }
 
 # Wait for Istio config to propagate
-# usage: _wait_for_istio <kind> <namespace> <name> <optional timeout> <optional if set, don't fail>
+# usage: _wait_for_istio <kind> <namespace> <name>
 _wait_for_istio() {
     local kind="$1"
     local namespace="$2"
     local name="$3"
-    local timeout="${4:-5m}"
-    start=$(date +%s)
-    if ! istioctl experimental wait --for=distribution --timeout="$timeout" "$kind" "$name.$namespace"; then
-        echo "Failed distribution of $kind $name in namespace $namespace after $(($(date +%s)-start)) second(s) duration"
-        if [ -z "$5" ]; then
-          istioctl ps
-          exit 1
-        fi
+    local start=$(date +%s)
+    if ! istioctl experimental wait --for=distribution --timeout=10s "$kind" "$name.$namespace"; then
+        echo "Failed distribution of $kind $name in namespace $namespace"
+        istioctl ps
         echo "TEST: wait for failed, but continuing."
     fi
     echo "Duration: $(($(date +%s)-start)) seconds"
--- a/tests/util/helpers.sh
+++ b/tests/util/helpers.sh
@@ -55,9 +55,10 @@ _wait_for_istio() {
     local kind="$1"
     local namespace="$2"
     local name="$3"
-    if ! istioctl experimental wait --for=distribution --timeout=1m "$kind" "$name.$namespace"; then
+    if ! istioctl experimental wait --for=distribution --timeout=5m "$kind" "$name.$namespace"; then
         echo "Failed distribution of $kind $name in namespace $namespace"
         istioctl ps
+        exit 1
     fi
 }
 
--- a/tests/util/helpers.sh
+++ b/tests/util/helpers.sh
@@ -55,7 +55,7 @@ _wait_for_istio() {
     local kind="$1"
     local namespace="$2"
     local name="$3"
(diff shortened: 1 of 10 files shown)


johnma14 (approved):

johnma14 (commented):
Just saw the WIP label on the commit message

tests/util/helpers.sh
```diff
@@ -55,7 +55,7 @@ _wait_for_istio() {
     local kind="$1"
     local namespace="$2"
     local name="$3"
-    if ! istioctl experimental wait --for=distribution --timeout=30s "$kind" "$name.$namespace"; then
+    if ! istioctl experimental wait --for=distribution --timeout=1m "$kind" "$name.$namespace"; then
```
frankbu, line 58: Why not just make it 5m, same as _wait_for_deployment(). It's just a max wait, so if it should never happen.
```suggestion
    if ! istioctl experimental wait --for=distribution --timeout=5m "$kind" "$name.$namespace"; then
```

ericvn: @frankbu Because there is a problem where occasionally the test waits the full timeout even if the requirements are satisfied. So we had multiple 5 min waits happening across the tests, and then we would blow the 60 min timeout as well. Maybe that's not an issue anymore. Need to look at the tests again.

ericvn: I looked at the test results for the three tests here, and don't see any timeout failures. Maybe we can go back to the 5m as you suggest and turn the failure back on if we timeout and see what happens.  Commit upcoming.

ericvn:
Looking at the output with a `5m` timeout: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/8163/doc.test.profile_demo_istio.io/90 you can see that once we get to the egress tests, they start taking more than 5 mins to complete, but don't fail. This was the failure we were seeing earlier and why we made the timeout less. More investigation is needed, but this is likely related to https://github.com/istio/istio/issues/27330.

ericvn:
Running demo bucket locally (with additional output to show duration of the waits). Tabulation of waits > 1 sec (note the bucket times out at 60 m):
DR - 6 - 150sec(3), 180sec(1), 210 sec(3)
GW - 3 - 120sec(2) 150sec(1)
VS - 4 - 2sec(1), 120sec(2), 150sec(1).

ericvn:
I updated the helper _wait_for_istio() by adding two parameters. The first is a timeout that can override the default for the framework (changed back to 5m earlier in this PR). The second will cause the function to return a 0 (passing) no matter if the timeout fails or not. This will allow a successive `verify` to possibly iterate and continue the wait.

ericvn:
Changing those `_wait_for_istio`'s that were taking a long time to use `2s true` showed that any verifies did not appreciably extend the time back to the longer time (I didn't measure for any increase in the verify's) as test times came down to a minute or two. Note that as I changed some of the _wait_for_istio`'s, the longest timeouts moved to others. After getting the times down, a majority of the _wait_for_istio`'s were changed. It would be simpler to just change the method instead of chasing the long delays and updating calls piecemeal. That's what was done in the last PR, so we end up back where we started.

Once we are back where we started (shortened timeout and always pass), the question remains as to which timeout to use. If the 30 sec timeout is causing issues with one test, and moving to something shorter actually causes the test to pass (and doesn't negatively impact the others), it might be best to move to that in the near term.

Once we get istioctl wait to not have the extended delays, moving back to a longer timeout and error makes sense.

ericvn:
And the authz test seems to be behaving better, so maybe no change is needed in the short term.

ericvn:
Simply not delaying in the wait (since it has some problems), and expecting a verify repeat and running the tests, yields a failure in `TestDocs/tasks/traffic-management/ingress/secure-ingress/test.sh`. 

The code does:
```
_wait_for_istio gateway default mygateway

_verify_failure snip_configure_a_mutual_tls_ingress_gateway_3
```
The wait isn't happening and the `_verify_failure` doesn't do a retry yields the failure.

ericvn:
Looking at the failure, I suspect it's due to a missing wait on sleep to deploy before trying to get the pod name. New commit coming with that change.

ericvn:
@frankb I think this is ready for review. Should satisfy our discussions.

frankbu (approved):

PR #8180, opened 2020-09-22
Rewrite multicluster installation docs

nmittler:
This completely replaces the existing docs for multi-primary, primary-remote, and single/multi-network.

Changed blogs dependent on the `.global` hack to reference the 1.6 docs.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


su225:
Minor issue - https://deploy-preview-8180--preliminary-istio.netlify.app/latest/docs/setup/install/multicluster/
Arrow indicates that there are more items where as we only have one multicluster related install documentation.
image

su225:
@nmittler - https://github.com/istio/test-infra/pull/2907 is merged. We now have a multicluster prow job.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -1,6 +1,6 @@
 ---
```
frankbu, line 1: Rename this file: `_index.md` -> `index.md`

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -9,13 +9,715 @@ aliases:
 keywords: [kubernetes,multicluster]
 test: n/a
```
frankbu, line 10:
```suggestion
test: yes
owner: istio/wg-environments-maintainers
```

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -9,13 +9,715 @@ aliases:
...
 ---
+When installing Istio across multiple clusters, there are more things to
+consider than with a [single cluster installation](/docs/setup/getting-started).
+
+There is no default installation procedure for a multicluster mesh, as it will
+be different based on your cluster and network architecture. We therefore
+provide options for:
```
frankbu, line 17:
```suggestion
Although there is no default installation procedure for a multicluster mesh, as it will
be different based on your cluster and network architecture, this guide covers some
of the most common setups:
provide options for:
```

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -9,13 +9,715 @@ aliases:
...
+
+- [Network topologies](/docs/ops/deployment/deployment-models#network-models):
+  one or two networks
+
+- [Control plane topologies](/docs/ops/deployment/deployment-models#control-plane-models):
+  multi-primary, primary-remote
+
+Before you begin, review the [deployment models](/docs/ops/deployment/deployment-models)
+which provides foundational concepts used throughout this guide.
```
frankbu, line 26:
```suggestion
Before you begin, review the [deployment models guide](/docs/ops/deployment/deployment-models)
which describes the foundational concepts used throughout this guide.
```

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -9,13 +9,715 @@ aliases:
...
+accessible from the other cluster, ideally using L4 network load balancers
+(NLBs). Not all cloud providers support NLBs and some require special
+annotations to use them, so please consult your cloud provider’s documentation
+for enabling NLBs for service object type load balancers. When deploying on
+platforms without NLB support, it may be necessary to modify the health checks
+for the load balancer to register the ingress gateway.
+
+## Environment Variables
+
+This guide will refer to two clusters named `FRED` and `BARNEY`. The following
```
frankbu, line 44: From the last doc WG call, I had the TODO to find out if the Flintstones reference in this document is [too culturally specific](https://developers.google.com/style/translation#be-inclusive).

I think the answer is that's fine. From IBM doc team: `unless the knowledge of the reference is required to understand what's happening in the example, then it doesn't really matter ... you dont need to know anything about the flintstones to understand the interactions between the clusters or anything, right? if not i think it's cool to keep it as-is. doesnt really matter if it is lost on people bc the reference isnt required to understand what's happening`

craigbox: That linked doc does go on to say:

`Don't use the names of fictional characters from material that's in copyright. (This includes virtually all movies and TV shows.)`

While I think we would be OK (and indeed while I suggested we do something like this), I agree that changing it would be sensible.

Bike-shedding on names now open on Slack :)

frankbu: Actually, the IBMer I talked to did also say `the only other thing to consider is copyright infringement but doesnt look like there's any of that here, just general names`. Basically, she thinks it's OK because we're not actually using the name `Flintstones`, just `Fred` and `Barney`. Anyway, if there's any doubt I agree we should change it.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+      multiCluster:
+        clusterName: FRED
+      network: EAST
+      meshNetworks:
+        EAST:
+          endpoints:
+          - fromRegistry: FRED
+          - fromRegistry: BARNEY
+          gateways:
+          - registryServiceName: istio-ingressgateway.istio-system.svc.cluster.local
```
stevenctl, line 128:
```suggestion
          - registryServiceName: istio-eastwestgateway.istio-system.svc.cluster.local
```

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+      multiCluster:
+        clusterName: BARNEY
+      network: EAST
+      meshNetworks:
+        EAST:
+          endpoints:
+          - fromRegistry: BARNEY
+          - fromRegistry: FRED
+          gateways:
+          - registryServiceName: istio-ingressgateway.istio-system.svc.cluster.local
```
stevenctl, line 154:
```suggestion
          - registryServiceName: istio-eastwestgateway.istio-system.svc.cluster.local
```

stevenctl (commented):
I don't know if its a big concern, but users don't technically have to specify the network name for single-network, could be confusing to a user reading this guide to add another cluster to their existing single-network mesh. 

It could be good to mention somewhere that specifying the network name is a good practice for future-proofing the mesh to add networks later.

content/en/docs/setup/install/multicluster/_index.md
nmittler: @craigbox any thoughts? Are you fine going with the names as-is or do you have a suggestion for a change?

ericvn:
profile none test failure:
serviceaccount "sleep" deleted
        service "sleep" deleted
        deployment.apps "sleep" deleted
        serviceaccount/sleep created
        service/sleep created
        deployment.apps/sleep created
        error: timed out waiting for the condition on pods/sleep-854565cb79-tgtm2
        NAME                     READY   STATUS        RESTARTS   AGE
        sleep-6cf6844d46-cwz6q   2/2     Running       0          61s
        sleep-854565cb79-tgtm2   2/2     Terminating   0          3m51s
        SOURCE_POD: sleep-6cf6844d46-cwz6q sleep-854565cb79-tgtm2
        Error from server (NotFound): pods "sleep-6cf6844d46-cwz6q sleep-854565cb79-tgtm2" not found

The code is basically deploying sleep (and there must have been one earlier that is being terminated) and then the `_wait_for_deployment default sleep` is timing out waiting on the terminating pod.

nmittler:
@ericvn there is no test in this PR. There shouldn't be any way that this is causing that failure.

ericvn:
@nmittler My comment was nothing really to do with what's in your PR, just noting that there seems to be a test flake for posterity.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
+---
+title: Multicluster Installation
+description: Install an Istio mesh across multiple Kubernetes clusters.
+weight: 30
+aliases:
+    - /docs/setup/kubernetes/multicluster-install/
+    - /docs/setup/kubernetes/multicluster/
+    - /docs/setup/kubernetes/install/multicluster/
+keywords: [kubernetes,multicluster]
+test: yes
```
ericvn, line 10: Although you do have test: set to yes which generates the snips.sh, and there are not test scripts created. The yes does show a checkmark on the page that this page is tested, when in fact there are no tests.

That said, I noticed that the scrips.sh did get lint tested to highlight problems in the snippets on the page.

nmittler: @frankbu added `yes` ... I assumed that was the goal, but not sure. I'm fine setting it back to `no`. I'd like to do the tests with a follow-on PR since it will be huge and complex.

ericvn: OK. I'll let @frankbu respond if it should go back to no now that the scripts are tested and being there are no tests at the moment. As noted above, I can see a reason to have the scripts.sh created and linted when updating pages with snippets. But it also leaves a false impression that a page is tested.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+owner: istio/wg-environments-maintainers
+---
+When installing Istio across multiple clusters, there are more things to
+consider than with a [single cluster installation](/docs/setup/getting-started).
+
+Although there is no default installation procedure for a multicluster mesh, as it will
```
smawson, line 16: I'd rephrase:

This guide covers some of the most common concerns when creating a multi-cluster mesh using Istio:
...

I don't think we need the first part about things being different.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+
+- [Control plane topologies](/docs/ops/deployment/deployment-models#control-plane-models):
+  multi-primary, primary-remote
+
+Before you begin, review the [deployment models guide](/docs/ops/deployment/deployment-models)
+which describes the foundational concepts used throughout this guide.
+
+## Requirements
+
+This guide requires that you have two Kubernetes clusters with any of the
```
smawson, line 31: Maybe be specific that we mean kubernetes versions here? Unclear if this means you need two clusters with Istio installed.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+[supported versions](/docs/setup/platform-setup).
+
+If the clusters are on different networks (i.e. no direct connectivity between
+pods on different clusters), they must have ingress gateway services which are
```
smawson, line 35: Make it clear that those can be set up using Istio after Istio is installed.

Otherwise this reads like they need to set up ingress gateways before installing Istio.

Alternatively just drop this from here and talk about it when configuring the second cluster on a different network.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+between all clusters in the mesh. Depending on the requirements for your
+system, there may be multiple options available for establishing trust.
+See [certificate management](/docs/tasks/security/cert-management/) for
+detailed descriptions and instructions for all available options.
+Depending on which option you choose, the installation instructions for
+Istio may change slightly.
+
+This guide will assume that you use a common root to generate intermediate
+certificates for each cluster. Follow the [instructions](/docs/tasks/security/cert-management/plugin-ca-cert/)
+to generate and push a secret to both the `FRED` and `BARNEY` clusters.
```
smawson, line 72:
```suggestion
to generate and push ca certificate secrets to both the `FRED` and `BARNEY` clusters.
```

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+
+{{< tip >}}
+If you currently have a single cluster with a self-signed CA (as described
+in [Getting Started](/docs/setup/getting-started/)), you’ll need to
+[change the CA](/docs/tasks/security/cert-management/) to support multiple
```
smawson, line 77: That link isn't clear, what do they actually need to do if they already have Istio installed with a self-signed cert? Can we point them at explicit instructions on how to migrate from self-signed to intermediate?

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+clusters and then reinstall Istio using the standard installation
+instructions below.
+{{< /tip >}}
+
+## Install Istio
+
+The steps for installing Istio on multiple clusters depend on your
+requirements for network and control plane topology. This section
+illustrates the options for two clusters. Large systems may employ more
```
smawson, line 86: Meshes that span many clusters may employ more than one of these options.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+    link="multi-primary.svg"
+    caption="Multiple primary clusters on the same network"
+    >}}
+
+### Configure `FRED` as a primary
+
+Install Istio on `FRED`:
+
+{{< text bash >}}
+$ istioctl --context=${CTX_FRED} -f - <<EOF
```
smawson, line 112: Super minor nit: I'd rather show us putting that operator config into a file (echo to a file for example) and then passing that file to istioctl install, rather than inlining it. Sets them up better for editing that file as they desire.

Also you missed the "install" part of the command.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+
+{{< /tab >}}
+
+{{< tab name="Multi-Primary, Multi-Network" category-value="multi-primary-multi-network" >}}
+
+The following steps will install the Istio control plane on both `FRED` and
+`BARNEY`, making each a {{< gloss >}}primary cluster{{< /gloss >}}. `FRED` is
+on the `WEST` network, while `BARNEY` is on the `EAST` network.
+
+Both `FRED` and `BARNEY` observe the API Servers in each cluster for endpoints.
```
smawson, line 187: I think this image is wrong, it has istiod directly observing the api server in the other cluster, but if it is on a different network it can't do that right? Is istiod using the E-W gateway to get to the api server in the other cluster? Or is the api server exposed some other way? Maybe we should mention how that part works and any requirements on istiod having access to the api server for other clusters?

smawson: I think many api servers from cloud providers are publicly accessible through an NLB but we should be explicit about that requirement. Some don't meet that and API servers are private, which means this may not work. And for on-prem Kubernetes installs all bets are off...

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+Install a gateway in `FRED` that is dedicated to
+[east-west](https://en.wikipedia.org/wiki/East-west_traffic) traffic. By
+default, this gateway will be public on the Internet. Production systems may
+require additional access restrictions (e.g. via firewall rules) to prevent
+external attacks. Check with your cloud vendor to see what options are
+available.
+
+{{< text bash >}}
+$ CLUSTER=FRED NETWORK=WEST \
+    samples/multicluster/gen-eastwest-gateway.sh | \
```
smawson, line 237: Same here, I'd prefer putting in a file and then having them kubectl apply it so they know they can save that file into their source control system.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+
+The following steps will install the Istio control plane on `FRED` (the
+{{< gloss >}}primary cluster{{< /gloss >}}) and configure `BARNEY` (the
+{{< gloss >}}remote cluster{{< /gloss >}}) to use the control plane in `FRED`. Both
+clusters reside on the `EAST` network.
+
+`FRED` will be configured to observe the API Servers in both clusters for
+endpoints. In this way, the control plane will be able to provide service
+discovery for workloads in both clusters. Service workloads across cluster
+boundaries communicate indirectly, via dedicated
```
smawson, line 329: This is wrong, they're going direct, it is only the istiod connection that goes through a gateway.

We should talk here about how control plane traffic goes through a gateway but data plane traffic goes direct. The control plane is indirect because of bootstrapping, we need a stable IP address for istiod and we use an ingress gateway to get that.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+
+Install a gateway in the `FRED` that is dedicated to
+[east-west](https://en.wikipedia.org/wiki/East-west_traffic) traffic. By
+default, this gateway will be public on the Internet. Production systems may
+require additional access restrictions (e.g. via firewall rules) to prevent
+external attacks. Check with your cloud vendor to see what options are
+available.
+
+{{< text bash >}}
+$ CLUSTER=FRED NETWORK=WEST \
```
smawson, line 371: We don't need the E-W gateway (there is no WEST anyway).

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+The following steps will install the Istio control plane on `FRED` (the
+{{< gloss >}}primary cluster{{< /gloss >}}) and configure `BARNEY` (the
+{{< gloss >}}remote cluster{{< /gloss >}}) to use the control plane in `FRED`.
+`FRED` is on the `WEST` network, while `BARNEY` is on the `EAST` network.
+
+`FRED` will be configured to observe the API Servers in both clusters for
+endpoints. In this way, the control plane will be able to provide service
+discovery for workloads in both clusters.
+
+Services in the `BARNEY` will reach the control plane in the `FRED` via a dedicated
```
smawson, line 437:
```suggestion
Services in the `BARNEY` cluster will reach the control plane in the `FRED` cluster via a dedicated
```

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+Enable automatic sidecar injection for the `sample` namespace:
+
+{{< text bash >}}
+$ kubectl label --context=${CTX_FRED} namespace sample \
+    istio-injection=enabled
+$ kubectl label --context=${CTX_BARNEY} namespace sample \
+    istio-injection=enabled
+{{< /text >}}
+
+Deploy the `HelloWorld` service to both clusters:
```
smawson, line 600:
```suggestion
Create the `HelloWorld` service in both clusters:
```

content/en/docs/setup/install/multicluster/index.md
frankbu: As long as the test is coming in a followup PR, before the 1.8 release (which is our policy for new docs), I think it's OK if it's set to yes in preliminary.istio.io, so that (as @ericvn pointed out) we can at least be sure the commands have no lint errors.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,724 @@
...
+
+Each control plane observes the API Servers in both clusters for endpoints.
+Service workloads communicate directly (pod-to-pod) across cluster boundaries.
+
+{{< image width="75%"
+    link="multi-primary.svg"
+    caption="Multiple primary clusters on the same network"
+    >}}
+
+### Configure `FRED` as a primary
```
frankbu, line 107: All of the section headings that are inside a `{{< tab ... >}}` don't work properly. Notice that if you click on any of these headings in the outline, it doesn't go to them. Not sure, what the best solution is, but one option is just to turn these headings into bold text.
```suggestion
**Configure `FRED` as a primary**
```

content/en/docs/setup/install/multicluster/index.md
nmittler: Yeah that was the assumption .. that API Servers are publicly available. I'll add that to the requirements at the top.

content/en/docs/setup/install/multicluster/index.md
nmittler: We're using the term E-W here to describe the traffic direction, it's not referring to the network names.

It is a bit confusing since we named our networks EAST and WEST. I guess we could rename the gateway to something like `in-mesh` or `inter-cluster`, but `east-west` is a fairly common term for this. Perhaps we rename the networks in these examples?

content/en/docs/setup/install/multicluster/index.md
nmittler: I've removed this and added some basic text to the multicluster sections.

content/en/docs/setup/install/multicluster/index.md
nmittler: @frankbu if we make this change they simply won't appear in the outline, right? I'm not sure if that's much better. From a readability standpoint, having them as headings (rather than just bold) seems more clear.

content/en/docs/setup/install/multicluster/_index.md
craigbox: Hmm.

Your networks were named WILMA and BETTY and I renamed them EAST and WEST because people who know the Flintstones would have thought they were the same class of object.

Now that's causing trouble with east-west traffic I see..

I checked in with a TW, and her guidance was "use CLUSTER-1, CLUSTER-2, NETWORK-1, NETWORK-2."  That's in line with Google docs guidance, any opinion from IBM?

I'd say "go with what we have" and look at new guidelines and updates later in a separate commit.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,744 @@
...
+Each control plane observes the API Servers in both clusters for endpoints.
+
+Service workloads communicate directly (pod-to-pod) across cluster boundaries.
+
+{{< image width="75%"
+    link="multi-primary.svg"
+    caption="Multiple primary clusters on the same network"
+    >}}
+
+**Configure `CLUSTER1` as a primary**
```
frankbu, line 109: Maybe to get around the lint error you can make these dummy headings list items for now:
```suggestion
- **Configure `CLUSTER1` as a primary**
```
We can revisit the layout and improve in a followup PR.

frankbu: Maybe something like this is another option? https://stackoverflow.com/questions/52062672/rmarkdown-add-header-under-tabset

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,744 @@
...
+clusters.
+
+{{< tip >}}
+If you currently have a single cluster with a self-signed CA (as described
+in [Getting Started](/docs/setup/getting-started/)), you need to
+change the CA using one of the methods described in
+[certificate management](/docs/tasks/security/cert-management/). Changing the
+CA typically requires reinstalling Istio. The installation instructions
+below may have to be altered based on your choice of CA.
+{{< /tip >}}
```
linsun, line 80: Could we have some clarification around `Changing the
CA typically requires reinstalling Istio`?  Do we mean just reinstall the control plane?  Is restarting data plane needed or can Istio auto distribute the new key/certs to each sidecar?

linsun: cc @myidpt @howardjohn

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,744 @@
...
+          endpoints:
+          - fromRegistry: CLUSTER2
+          - fromRegistry: CLUSTER1
+          gateways:
+          - registryServiceName: istio-eastwestgateway.istio-system.svc.cluster.local
+            port: 15443
+EOF
+$ istioctl install --context=${CTX_CLUSTER2} -f cluster2.yaml
+{{< /text >}}
+
```
linsun, line 162: For multi primary, the diagram didn't show gateways while the plan showed gateways.  This is a bit confusing when I reviewed this tab.

nmittler: Are you referring to the presence of `istio-eastwestgateway` in the `IstioOperator` yaml? Thing is that the gateway isn't needed for this case, so we don't actually deploy it. Not sure if it's valid to leave out `gateways` altogether, but if that works maybe that's a better solution here.

@stevenctl WDYT?

linsun: I assume the gateways aren't used for workloads to communicate if so, it is confusing we also call it eastwestgateway.

stevenctl: I do agree that we don't need to show the gateways in the multi-primary guide, they're not really necessary. 

@linsun 
The eastwest gateways are used for:
- multinetwork workload traffic
- discovery from outside the clsuter

eastwest sort of implies it's mostly used for workload traffic, so I can see how that name is confusing.

nmittler: @linsun east-west is a pretty standard term, but I'm open to other names if they're more clear.

linsun: If removing the gateways are okay for this example, I'd vote for removing it because it is quite confusing to have it :) @stevenctl @nmittler

linsun: @nmittler the name east-west is good.  It is just confusing here when it is not used for workload traffic.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,744 @@
...
+$ CLUSTER=CLUSTER1 NETWORK=NETWORK1 \
+    samples/multicluster/gen-eastwest-gateway.sh | \
+    kubectl apply --context=${CTX_CLUSTER1} -f -
+{{< /text >}}
+
+Since the clusters are on separate networks, we need to expose all services
+(*.local) on the east-west gateway in both clusters. While this gateway is
+public on the Internet, services behind it can only be accessed by services
+with a trusted mTLS certificate and workload ID, just as if they were on the
+same network.
```
linsun, line 252: Ok this is where I think there is some gap when comparing with our previous replicated control plane model.   The doc here instructs users to expose all services while the replicated control plane instructs users to selectively expose services.   I don't think we have that function ready yet as we are waiting on k8s for the multicluster service API.

This is also where I'm not sure if we are ready to remove the replicated control plane doc yet without an equivalent replacement.  

@nmittler @stevenctl @rshriram @frankbu

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,744 @@
...
+      multiCluster:
+        clusterName: CLUSTER1
+      network: NETWORK1
+      meshNetworks:
+        ${NETWORK1}:
+          endpoints:
+          - fromRegistry: CLUSTER1
+          - fromRegistry: CLUSTER2
+          gateways:
+          - registryServiceName: istio-eastwestgateway.istio-system.svc.cluster.local
```
linsun, line 371: Similar comment as earlier - Is gateway required here?

content/en/docs/setup/install/multicluster/index.md
nmittler: I'm not sure there is a gap, really. This should be the default behavior for a service mesh that spans multiple clusters. The updated [Deployment Models](https://preliminary.istio.io/latest/docs/ops/deployment/deployment-models/#dns-with-multiple-clusters) doc discusses selective service exposure via `Service`/`ServiceEntry`, but that is beyond the scope of this guide.

content/en/docs/setup/install/multicluster/index.md
linsun: Ok i think it is required here for workloads on remote cluster to reach out to istiod... but not for workloads from remote cluster to reach out to cluster 1.

nmittler: yup, exactly.

content/en/docs/setup/install/multicluster/index.md
linsun: Are you saying it is exposed to the other cluster by default but what services to import are controlled by creating service entry or the corresponding service in the consuming cluster?

nmittler: You can further clamp it down by being more selective when we expose services on the gateway. I can add a note in the text if that helps?

linsun:
I think this LGTM overall.  My only concern is removal of replicated control plane - wanted a few people who wrote that doc to sign off @rshriram @frankbu.   

@nmittler one thing i think it is confusing is that in your diagrams, you had service A (cluster 1) and service B (cluster 2) where service A calls service B.  What is not explained in the diagram is that service B also need to exist in cluster 1 (or at least a service entry for service B needs to exist).

This is explained briefly later in the hello examples but could be easily missed by users.

nmittler:
@linsun 

> @nmittler one thing i think it is confusing is that in your diagrams, you had service A (cluster 1) and service B (cluster 2) where service A calls service B. What is not explained in the diagram is that service B also need to exist in cluster 1 (or at least a service entry for service B needs to exist).

Yeah understood. That is part of user application deployment, however, not control plane setup. So I think the verification step (where we actually deploy services) is the right home for it. In addition, there is an extensive discussion in the updated deployment models and I don't think we want to duplicate all of that here. We're upfront in telling the user to review the deployment models before beginning the installation.

linsun:
Just had a discussion with @nmittler @stevenctl - regarding my concern earlier with replacing replicated control plane with the multi-primary & multi-network model model, I think I'm good with using gateway resource to selectively choose what services to expose to the remote cluster(s).  Basically, if admin wants to maintain cluster boundary and selectively exposes service A in c1 , they can control it via gateway resources (within hosts config to choose what hosts to expose).  And to consume the service A(from c1) in the consuming c2, just create the dummy service A in the c2 to establish the dns..  No need to use service entry or core dns in this model.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -0,0 +1,744 @@
...
+Each control plane observes the API Servers in both clusters for endpoints.
+
+Service workloads communicate directly (pod-to-pod) across cluster boundaries.
+
+{{< image width="75%"
+    link="multi-primary.svg"
+    caption="Multiple primary clusters on the same network"
+    >}}
+
+- **Configure `CLUSTER1` as a primary**
```
frankbu, line 109: This doesn't look so good in the preview. Have you tried using html tags?
```suggestion
<h3>Configure `CLUSTER1` as a primary</h3>
```
If that doesn't work either, then maybe we should merge it as is and investigate other ideas in a followup PR.

nmittler:
@frankbu take a look ... if you're ok with it, let's just get it in.

frankbu (approved):

PR #8184, opened 2020-09-22
Logo changes

craigbox:
[X] Docs

Add four logos, remove one at user request, and change the weighting of the About section.
(Requires cherrypicking for 1.6 please.)


frankbu:
On the customers page it says: `Here are just a few of the organizations we know are using Istio`. Does `just a few` sound a bit arrogant, given the list isn't small. Maybe `Here are some of the ...` would be better?

frankbu:
> (Requires cherrypicking for 1.6 please.)

 Why is it needed in the archive docs? Did you mean 1.7 (i.e., on istio.io)?

craigbox:
> > (Requires cherrypicking for 1.6 please.)
> 
> Why is it needed in the archive docs? Did you mean 1.7 (i.e., on istio.io)?

Erm, sorry. Yes, there's been a release since now and then. 😄

craigbox:
> On the customers page it says: `Here are just a few of the organizations we know are using Istio`. Does `just a few` sound a bit arrogant, given the list isn't small. Maybe `Here are some of the ...` would be better?

I think I changed it away from that wording in the last big upgrade, because it didn't sound as natural to me. This will be addressed in the larger update to happen so I'd recommend we leave it until then.

frankbu:
> so I'd recommend we leave it until then.

OK, but I would fix it soon. Unless there are another hundred or more customers that we know are using Istio, the `just a few of the organizations we know` part is an outright lie.

frankbu (approved):

PR #8193, opened 2020-09-23
[release-1.7] Cherry-pick #8184 for logo changes

craigbox:
[X] Docs

Manually cherry-pick #8184, fix failed automated attempt in #8185.


craigbox:
https://deploy-preview-8193--istio-staging.netlify.app/latest/about/community/customers/ spot-check LGTM

ericvn:
I admit I was expecting 3 files to be changed here, and not a lot of the .svg files. Do you know if the changes in files are those caused by the IMGBOT and it's shrinking of files?

craigbox:
Yes, after diffing the directories I saw that the files had been optimized.  Not 100% sure why, but it seemed a win to move them. I can skip this step if it's disruptive.

frankbu (approved):
I guess it's fine to replace the exiting images with the optimized ones, but I would have thought that the imagebot would/should have done that automatically. Maybe we are supposed to add the cherrypick label to imagebot PRs, but we haven't been doing that?

ericvn:
imagebot only runs agains master as far as I know.  I was always unsure of cherrypicking because as things diverge, the changes may or may not be correct for both versions.

ericvn:
And if the cherrypick fails, the imagebot isn't going to understand an issue being opened and assigned to it to do it manually.

PR #8205, opened 2020-09-24
Add korean localization

markruler:
[x] **Docs**
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Hi! :raised_hand_with_fingers_splayed:

I started translating this site into Korean.
Please review it positively.

### Add
i18n/ko.toml
content/ko/...

### Update
config.toml
src/ts/lang.ts
layouts/partial/header.html
layouts/index.redir
.gitignore


markruler:
@googlebot I signed it!

frankbu:
@markruler Thanks for the PR. We have been reevaluating the process for supporting translations of the Istio documentation at istio.io. The problem with publishing an early partial translation that contains mostly just copies of English docs, is that many of those English docs will soon be outdated, given how quickly Istio docs change. The result is that we end up having multiple copies of the same English doc, corresponding to a particular Istio version, that Google searches will return. Other than by looking at the actual URL, there is no way users can know if they are looking at an actual up to date English doc or not. Therefore, we need to be sure that all the docs are translated, before we publish them. That's why we were forced to remove all of the files from the pt-br version that hadn't been translated yet.

Do you think it would be possible to translate the entire set of docs in this PR, or at least the titles, so that it is clear it is a partially translated file, instead of the English doc?

Please note that the Istio docs are still changing frequently, so I would recommend a process to keep track of day-to-day changes, so that your translated version can be kept up to date. I know that even the zh version, which is a complete translation, also struggles to stay up to date.

cc @istio/technical-oversight-committee @istio/steering-committee

PR #8224, opened 2020-09-29
More retries for flakes

ericvn:
Still seeing 4 of last 20 failing for the authorization/authz-ingress:
```
    script.go:111: script tasks_security_authorization_authz-ingress_test returned an error: exit status 1. Output:
        namespace/foo created
        serviceaccount/httpbin created
        service/httpbin created
        deployment.apps/httpbin created
        gateway.networking.istio.io/httpbin-gateway created
        virtualservice.networking.istio.io/httpbin created
        Waiting for deployment "httpbin" rollout to finish: 0 of 1 updated replicas are available...
        deployment "httpbin" successfully rolled out
        service/istio-ingressgateway patched
        authorizationpolicy.security.istio.io/ingress-policy created
        Duration: 0 seconds
        authorizationpolicy.security.istio.io/ingress-policy configured
        Duration: 0 seconds
        VERIFY FAILED snip_ipbased_allow_list_and_deny_list_4: received: "403", expected: "200"
```

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/security/authorization/authz-ingress/test.sh
+++ b/content/en/docs/tasks/security/authorization/authz-ingress/test.sh
@@ -22,7 +22,7 @@ set -o pipefail
 # @setup profile=default
 
 # Set retries to a higher value because config update is slow.
-export VERIFY_RETRIES=10
+export VERIFY_RETRIES=15
 
 export CLIENT_IP
 


content/en/docs/tasks/security/authorization/authz-ingress/test.sh
```diff
@@ -22,7 +22,7 @@ set -o pipefail
 # @setup profile=default
 
 # Set retries to a higher value because config update is slow.
-export VERIFY_RETRIES=10
+export VERIFY_RETRIES=15
```
frankbu, line 25: This will take way too long to time out. With the exponential backoff, 15 retries will take about 8 hours before failing. 10 retries is about the longest I would ever go (about 17 mins).

ericvn: Well, the test bucket will fail at 60 mins, but maybe we can increase the retries by 1 or 2. Not sure if there is something else we can do that may get the config written so that the tests do pass.

frankbu: You could bump it to 11 (about 35 mins), but I doubt that waiting more than 17 mins will help. It should never need half an hour to propagate config. My guess is there must be some kind of timing bug where the config change gets missed in which case waiting won't help.

ericvn: Making this DNM for now. Maybe can figure out if there is something we can do to get the config written and may need doc updates, etc.

PR #8231, opened 2020-09-29
Update test ref to 1.7.3

ericvn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
Problem is that with the private build for the security fix, the normal images we get from grr.io/istio-testing for the commit aren't there. I tried to override the HUB and TAG with the actual release (seems to work for 1.6), but the installs done directly in the scripts don't pick up the env vars as the test created control planes do. Not sure if we want to update your tests to use them (maybe it's setting HUB and TAG when we build istioctl so things are built in) or maybe its a change in the private build process to also copy the images back to the public test repo.

frankbu:
> I tried to override the HUB and TAG with the actual release (seems to work for 1.6)

It's only working in 1.6 because we didn't have any tests that do install yet.

If we can't easily make the security images available in the usual place, I think we should just not update the test build for security releases, since those builds don't change anything that is likely to break docs.

frankbu (approved):

PR #8240, opened 2020-09-30
Create 404.html

adityakumar48:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/404.html
+++ b/404.html
@@ -1,0 +1,13 @@
+---
+layout: base
+title: "404 - File Not Found"
+permalink: /404.html
+---
+<div class="row">
+  <div class="small-12 columns">
+     <h1>We could not Found that page</h1>
+     <article>
+     Please Check Your Link .
+     </article>
+  </div>
+</div>


frankbu:
@adityakumar48 Can you explain why this is needed? Where would it be used?

adityakumar48:
> @adityakumar48 Can you explain why this is needed? Where would it be used?

Dear Sir, because If any one mistake to go your wrong site  to  go and show to error 404 this page is not found then they to know its wrong then they are write our correct address

frankbu:
@adityakumar48 But isn't this already handled? For example, if I access a missing page in my browser: https://preliminary.istio.io/latest/docs/setup/getting-startd/, it displays a nice html page with message `We're sorry, the page you requested cannot be found`

adityakumar48:
> @adityakumar48 But isn't this already handled? For example, if I access a missing page in my browser: https://preliminary.istio.io/latest/docs/setup/getting-startd/, it displays a nice html page with message `We're sorry, the page you requested cannot be found`

Ok Sir

ericvn:
Marking as spam from Hacktoberfest and closing.

Issue #8245, opened 2020-10-01
Add health check using TCP socket


shamsher31:
Ref: https://preliminary.istio.io/latest/docs/ops/configuration/mesh/app-health-check/

Other than `command` and `HTTP` probes, Kubernetes also provide health checks using a TCP socket.

PR #8246, opened 2020-10-01
Add health check using TCP socket

shamsher31:
To fix https://github.com/istio/istio.io/issues/8245

```
$ kubectl get pods
NAME                             READY   STATUS    RESTARTS   AGE
liveness-http-86c7cc8f45-s4h7m   1/1     Running   0          2m
liveness-tcp-59f6d56f7-69f7k     1/1     Running   0          2m3s
```
Using the HTTP probe
```
$ kubectl describe pod liveness-http-86c7cc8f45-s4h7m
...
Liveness:       http-get http://:8001/foo delay=5s timeout=1s period=5s #success=1 #failure=3
```

Using TCP probe
```
$ kubectl describe pod liveness-tcp-59f6d56f7-69f7k
...
Liveness:       tcp-socket :8001 delay=5s timeout=1s period=5s #success=1 #failure=3
```

### DIFF
--- a/content/en/docs/ops/configuration/mesh/app-health-check/index.md
+++ b/content/en/docs/ops/configuration/mesh/app-health-check/index.md
@@ -17,8 +17,9 @@ test: yes
 [Kubernetes liveness and readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/)
 describes several ways to configure liveness and readiness probes including:
 
-1. Command
-1. HTTP request
+1. [Command](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command)
+1. [HTTP request](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request)
+1. [TCP Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe)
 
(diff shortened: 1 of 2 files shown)


frankbu:
From the previous cleanup PR for this doc, it seemed that the plan was to remove options from this doc, now that Istio handles health checking by default.

Does this change make sense if that is the case?

cc @istio/wg-user-experience-maintainers 

I also think that the corresponding test.sh for this doc needs to be testing all of the documented options.

shamsher31:
> I also think that the corresponding test.sh for this doc needs to be testing all of the documented options.

I'm thinking to move `http` and `tcp` example into `samples/health-check`. I can update the doc and test if that is ok?

frankbu:
> I'm thinking to move http and tcp example into samples/health-check. I can update the doc and test if that is ok?

@shamsher31 I'd like the [owner of this document](https://github.com/istio/istio.io/blob/master/DOC_OWNERS.md#istiowg-user-experience-maintainers-32-docs) to decide what the future direction should be, since it's unclear to me how it should be aligned with the current release of Istio. That's why I added the wg label and reviewer.

shamsher31:
@istio/wg-user-experience-maintainers PTAL.

shamsher31:
cc: @therealmitchconnors @esnible

esnible (approved):

frankbu (approved):

howardjohn:
I am not sure this is right. AFAIK TCP socket health check is broken by design in Istio, since its not possible to proxy a "can I connect" check.

@shamsher31 @frankbu

PR #8248, opened 2020-10-01
Fixes for multicluster install doc

nmittler:
[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/multicluster/index.md
+++ b/content/en/docs/setup/install/multicluster/index.md
@@ -130,7 +130,7 @@ spec:
           - registryServiceName: istio-eastwestgateway.istio-system.svc.cluster.local
             port: 15443
 EOF
-$ istioctl install --context=${CTX_CLUSTER1} -f cluster1.yaml
+$ istioctl install --context=${CTX_CLUSTER1} --skip-confirmation -f cluster1.yaml
 {{< /text >}}
 
 <h3>Configure CLUSTER2 as a primary</h3>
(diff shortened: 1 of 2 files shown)


ericvn (commented):
LGTM. Need to fix the lint error. I'll also let @frankbu weigh in on wether we want the `--skip-confirmation`. I know there was discussion earlier about having it only on select pages, like Getting Started. From a test standpoint for the other pages, we just do an echo y into the snippet.

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -130,7 +130,7 @@ spec:
           - registryServiceName: istio-eastwestgateway.istio-system.svc.cluster.local
             port: 15443
 EOF
-$ istioctl install --context=${CTX_CLUSTER1} -f cluster1.yaml
+$ istioctl install --context=${CTX_CLUSTER1} --skip-confirmation -f cluster1.yaml
```
frankbu, line 133: We should only add `--skip-confirmation` if we really think confirming in this task is unnecessary. If it's only to avoid it in the doc test, we should instead just pipe the input in the test, for example:
```
echo y | snip_install_istio_1
```
Also, in places where we do want to skip confirmation for simplicity, we have been trying not to highlight the fact by just adding the short version of the option at the end of the command:
```
$ istioctl install --context=${CTX_CLUSTER1} -f cluster1.yaml -y
```

ericvn: Thanks Frank. All that makes sense.

nmittler:
@frankbu PTAL

frankbu (approved):

PR #8250, opened 2020-10-01
README update on how to update build versions tests use

ericvn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,3 +222,27 @@ We'd appreciate it if you'd like to help create one! Our goal is to eventually h
 for every testable document published on the Istio site.
 
 See the [tests README](tests/README.md) for more information.
+
+### Updating the test reference for a given release stream
+
+The release streams starting with `release-1.6` contain tests for the test content. Each release tests against
+a particular istio version/commit. When the release team has a build, `1.x.y`, ready for their long running tests, they
+should come to the docs team to have the testing for that release run start running against the build.
+
+For public builds:
+1. Get the istio/istio commit that was used for the build from https://gcsweb.istio.io/gcs/istio-release/releases/1.x.y/manifest.yaml file.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -222,3 +222,27 @@ We'd appreciate it if you'd like to help create one! Our goal is to eventually h
...
+
+The release streams starting with `release-1.6` contain tests for the test content. Each release tests against
+a particular istio version/commit. When the release team has a build, `1.x.y`, ready for their long running tests, they
+should come to the docs team to have the testing for that release run start running against the build.
+
+For public builds:
+1. Get the istio/istio commit that was used for the build from https://gcsweb.istio.io/gcs/istio-release/releases/1.x.y/manifest.yaml file.
+1. In the release branch:  Run `go get istio.io/istio@commit && go mod tidy`.
+
+For private builds (this is done after the build is released):
```
frankbu, line 236: I don't understand what a "private build" is. I understand that there are two kinds of builds:

1. Release builds published at docker.io
2. Dev builds published at gcr.io

They are both "public", aren't they?

ericvn: Release and dev builds are public. Private builds are those where we can't reveal much beforehand. Typically it's an advance notice that a release will happen in two weeks for CVEs. Since we can't reveal anything before the actual release, the source and built images are in private repos. As the source and images are private, we can't actually move to them until they are publicly released, and thus there is no early testing of the release in istio.io. The difference here is that the images we test against in the gcr.io repository were never created for the release, so in that case we use the actual images.

Does that make sense?

I'll try and add something in the doc.

frankbu (approved):

PR #8253, opened 2020-10-01
[Test Framework] Fixes to support multicluster

nmittler:
[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/observability/distributed-tracing/zipkin/multicluster_test.sh
+++ b/content/en/docs/tasks/observability/distributed-tracing/zipkin/multicluster_test.sh
@@ -21,8 +21,6 @@ set -o pipefail
 
 # @setup multicluster
 
-source "tests/util/helpers.sh"
-
 # Initialize KUBECONFIG_FILES
 _set_kube_vars
 
--- a/content/en/docs/tasks/observability/distributed-tracing/zipkin/multicluster_test.sh
+++ b/content/en/docs/tasks/observability/distributed-tracing/zipkin/multicluster_test.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# shellcheck disable=SC1090,SC2154,SC2155,SC2034
+# shellcheck disable=SC1090,SC2154,SC2155,SC2031,SC2034
 
 # Copyright Istio Authors
 #
@@ -21,14 +21,18 @@ set -o pipefail
 
 # @setup multicluster
 
+source "tests/util/helpers.sh"
+
+# Initialize KUBECONFIG_FILES
+_set_kube_vars
+
 # TODO: Remove this once we have a real multicluster test
 
 echo "start multicluster test with KUBECONFIG=${KUBECONFIG}"
 echo "DOCTEST_KUBECONFIG=${DOCTEST_KUBECONFIG}"
 echo "DOCTEST_NETWORK_TOPOLOGY=${DOCTEST_NETWORK_TOPOLOGY}"
 
-IFS=',' read -r -a CONFIGS <<< "${KUBECONFIG}"
-for kcfg in "${CONFIGS[@]}"; do
+for kcfg in "${KUBECONFIG_FILES[@]}"; do
   kubectl --kubeconfig="$kcfg" get pods -A
 done
 
--- a/tests/util/helpers.sh
+++ b/tests/util/helpers.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# shellcheck disable=SC2155
+# shellcheck disable=SC2155,SC2030,SC2031
 
 # Copyright Istio Authors
 #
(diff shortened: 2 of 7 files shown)


ericvn:
Overall, this looks good. My main concern, and I'll have to do some testing, is related to the KUBECONFIG and multiple files. I worked with @sdake early on with multiple files when using the build container. I also typically have multiple contexts within a given file (mostly 2 ,but occasionally 4), and that may not work here.

nmittler:
> Overall, this looks good. My main concern, and I'll have to do some testing, is related to the KUBECONFIG and multiple files. I worked with @sdake early on with multiple files when using the build container. I also typically have multiple contexts within a given file (mostly 2 ,but occasionally 4), and that may not work here.

@ericvn I've updated the PR to use the current context from each kubeconfig file. PTAL

tests/util/helpers.sh
```diff
@@ -15,6 +15,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# _set_kube_vars initializes the following variables based on the value of KUBECONFIG:
+#
+# KUBECONFIG_FILES: an array containing the individual files in the order specified.
+# KUBE_CONTEXTS: the names of the kube contexts, in the order of the KUBECONFIG files specified.
+function _set_kube_vars()
+{
+  # Split out the kube config files and then get the first context in
```
frankbu, line 24:
```suggestion
  # Split out the kube config files and then get the current context in
```

tests/util/helpers.sh
```diff
@@ -15,6 +15,38 @@
...
+  # each. We do this because the contexts are stored in a map, which
+  # means that order of the context returned by
+  # `kubectl config get-contexts` is not guaranteed. By pulling out
+  # the context on a per-file basis, we maintain the order of the
+  # files in the KUBECONFIG variable. It should be noted, however,
+  # that this logic assumes one context per file.
```
frankbu, line 30: Which logic assumes this? Is it not possible to run the tests using one KUBECONFIG containing multiple contexts/clusters?

tests/util/verify.sh
```diff
@@ -280,30 +281,59 @@ __cluster_state() {
...
-    kubectl get serviceentries --ignore-not-found -n default -n istio-system 2>&1 | grep -v "error: the server doesn't have a resource type"
-    kubectl get sidecars --ignore-not-found -n default -n istio-system 2>&1 | grep -v "error: the server doesn't have a resource type"
-    kubectl get virtualservices --ignore-not-found -n default -n istio-system 2>&1 | grep -v "error: the server doesn't have a resource type"
-    kubectl get workloadentries --ignore-not-found -n default -n istio-system 2>&1 | grep -v "error: the server doesn't have a resource type"
-    kubectl get authorizationpolicies --ignore-not-found -n default -n istio-system 2>&1 | grep -v "error: the server doesn't have a resource type"
-    kubectl get peerauthentications --ignore-not-found -n default -n istio-system 2>&1 | grep -v "error: the server doesn't have a resource type"
-    kubectl get requestauthentications --ignore-not-found -n default -n istio-system 2>&1 | grep -v "error: the server doesn't have a resource type"
+    kubectl get istiooperators --ignore-not-found -n default 2>&1
```
frankbu, line 284: I'm wondering why the grep is no longer needed. It was a workaround (hack actually) because some of the `profile_none` tests were failing because the Istio CRDs didn't exist when the snapshot was taken (which output these missing type messages), but they were there at the end of the test.

content/en/docs/tasks/observability/distributed-tracing/zipkin/multicluster_test.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
```
frankbu, line 1: This test is just a temporary dummy to test the multicluster profile, right? It has nothing to do with the zipkin task, so we should remove it as soon as we have a real multicluster test.

tests/util/verify.sh
nmittler: it was causing failures when I was running. And CI seems fine as well.

content/en/docs/tasks/observability/distributed-tracing/zipkin/multicluster_test.sh
```diff
@@ -21,14 +21,18 @@ set -o pipefail
 
 # @setup multicluster
 
+source "tests/util/helpers.sh"
```
frankbu, line 24: This isn't needed because helpers.sh is one of the auto included files: https://github.com/istio/istio.io/tree/master/tests#include-files

tests/util/verify.sh
```diff
@@ -280,30 +281,59 @@ __cluster_state() {
...
+    kubectl get virtualservices --ignore-not-found -n default -n istio-system 2>&1
+    kubectl get workloadentries --ignore-not-found -n default -n istio-system 2>&1
+    kubectl get authorizationpolicies --ignore-not-found -n default -n istio-system 2>&1
+    kubectl get peerauthentications --ignore-not-found -n default -n istio-system 2>&1
+    kubectl get requestauthentications --ignore-not-found -n default -n istio-system 2>&1
 }
 
-__cluster_snapshot() {
-    __cluster_state > __cluster_snapshot.txt 2>&1
+__create_cluster_snapshots() {
```
frankbu, line 297: Can you also change the comment on line 268: __cluster_snapshot -> __create_cluster_snapshots

content/en/docs/tasks/observability/distributed-tracing/zipkin/multicluster_test.sh
nmittler: yup .. that's the plan.

frankbu (approved):

frankbu (approved):

PR #8254, opened 2020-10-01
Revert to -o pipefail and fix lint-yaml, revert istio commit

ericvn:
First thing to do is get the failing test to show they fail and the linter to pass.

At this point I don't expect a number of the tests to pass (something is broken with the reference commit and our framework), so seeing a failure is a good thing.

Next up is what to do to get the tests to pass. I believe that is part of @nmittler's PR. 

Trying these changes with the kubeconfig changes did not fix the test failures :(.

Several issues opened for the failure, https://github.com/istio/istio/issues/27692 and https://github.com/istio/istio/issues/27693.

Reverting to the prior istio commit so tests pass.


Makefile.core.mk
```diff
@@ -99,13 +99,18 @@ opt:
 clean:
 	@rm -fr resources .htmlproofer tmp generated public out samples install go tests/integration/ manifests
 
-lint: clean_public build_nominify lint-copyright-banner lint-python lint-yaml lint-dockerfiles lint-scripts lint-sass lint-typescript lint-go
+# Use docs-lint-yaml as a target around the common-files lint-yaml, but set +o pipefail for that call
+# otherwise it fails.
+docs-lint-yaml:
+	@set +o pipefail; $(MAKE) -f common/Makefile.common.mk lint-yaml
```
nmittler, line 105: This is a little surprising ... how does setting this for the lint target affect tests? The tests aren't dependent on this target, are they?

ericvn: You will noticed that I removed ~line 200 which was setting this for all the targets, and then just added it back here for the single target that needs it from the istio.io repo's perspective.

frankbu (approved):

PR #8257, opened 2020-10-02
Revert to common-files lint-yaml after fixing failure in common-files

ericvn:
This change removes the istio.io wrapper around the common-file's lint-yaml and thus the pipefail is being set. To prevent the `grep -L -e "{{"` from returning a `1` and causing the lint-yaml to fail, an extra yaml file is added causing the `grep` to return a `0`.

Not sure if having the wrapper or an extra file is the more appropriate workaround. Otherwise, there may be a way to fix the lint-yaml target in common-files to not fail. Looking at that some more.

### DIFF
--- a/empty.yaml
+++ b/empty.yaml
@@ -1,4 +1,0 @@
-# This is an empty yaml file so make lint-yaml passes.
-# lint-yaml removes files with {{ in them using grep -L.
-# If it fails to remove any yaml files, it returns an error.
-{{ }}
--- a/empty.yaml
+++ b/empty.yaml
@@ -1,0 +1,4 @@
+# This is an empty yaml file so make lint-yaml passes.
+# lint-yaml removes files with {{ in them using grep -L.
+# If it fails to remove any yaml files, it returns an error.
+{{ }}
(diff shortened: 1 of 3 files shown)


empty.yaml
```diff
@@ -0,0 +1,4 @@
+# This is an empty yaml file so make lint-yaml passes.
+# lint-yaml removes files with {{ in them using grep -L.
```
frankbu, line 2: I still don't understand why we can't fix this properly. Change the script to something like this:
```
find . -name "*.yml" -not -exec grep "{{" {} \; -print
```

ericvn: I've tried a version of Frank's suggestion in the common-files as a test. We wouldn't actually check that in here, only this repo's Makefile change once the common-files change is propagated.

ericvn:
Initial lint test failure was expected as I removed the upcoming common-file PR change. Once that change is merged, the lint test can run again and should pass.

ericvn:
This one should be ready to go once the tests pass.

frankbu (approved):

PR #8262, opened 2020-10-02
[Test Framework] Add multicluster test for multi-primary

nmittler:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/setup/install/multicluster/common.sh
```diff
@@ -0,0 +1,112 @@
...
+{
+  kubectl delete ns istio-system sample --context="${CTX_CLUSTER1}" --ignore-not-found
+}
+
+# _delete_namespaces_cluster2 removes the istio-system and sample namespaces on both
+# CLUSTER2.
+function _delete_namespaces_cluster2()
+{
+  kubectl delete ns istio-system sample --context="${CTX_CLUSTER2}" --ignore-not-found
+}
```
frankbu, line 57: I wonder if this bunch is even more common than for just this doc's tests, and could be moved up to tests/util. WDYT?

nmittler: I don't think so. It relies specifically on the environment variables `CLUSTER1` and `CLUSTER2`

content/en/docs/setup/install/multicluster/common.sh
```diff
@@ -0,0 +1,112 @@
...
+
+    # Wait before trying again
+    sleep 2
+  done
+
+  if [[ -z "$LOAD_BALANCED" ]]; then
+    echo "Traffic was not load balanced between the clusters"
+    exit 1
+  fi
+}
```
frankbu, line 112: This doesn't properly handle failures and retries. Better to use the built-in framework functions to get all that for free. Also, don't use `_` prefix for verify functions that aren't part of the framework standard library.

Maybe something like this:
```suggestion
  # Verify that traffic is load balanced from both clusters.
  _verify_lines repeated_snip_verifying_crosscluster_traffic_1 "
  + Hello version: v2,
  + Hello version: v1,
  "
  _verify_lines repeated_snip_verifying_crosscluster_traffic_3 "
  + Hello version: v2,
  + Hello version: v1,
  "
}

repeated_snip_verifying_crosscluster_traffic_1() {
  for i in {1..10}; snip_verifying_crosscluster_traffic_1; done
}
 
repeated_snip_verifying_crosscluster_traffic_3() {
  for i in {1..10}; snip_verifying_crosscluster_traffic_3; done
}
```

content/en/docs/setup/install/multicluster/common.sh
```diff
@@ -77,36 +77,15 @@ function verify_load_balancing()
...
+  local EXPECTED_RESPONSE_FROM_CLUSTER1="Hello version: v1, instance:"
+  local EXPECTED_RESPONSE_FROM_CLUSTER2="Hello version: v2, instance:"
 
-    # Wait before trying again
-    sleep 2
-  done
+  # Verify we hit both clusters from CLUSTER1
+  VERIFY_RETRIES=9
+  _verify_contains snip_verifying_crosscluster_traffic_1 "$EXPECTED_RESPONSE_FROM_CLUSTER1"
+  _verify_contains snip_verifying_crosscluster_traffic_1 "$EXPECTED_RESPONSE_FROM_CLUSTER2"
```
frankbu, line 86: This approach should work, but it could be slow because retries back off exponentially. The first retry delay is 1 sec, the next is 2 sec, 4 sec, 8 sec, 16 sec ... so if it takes all 9 retries before you get what you're looking for, the total delay will be more than 5 minutes I think. That's why I suggested just doing 10 quick curls and make sure you get at least 1 of each version in the output.

nmittler: It seems fine in my runs so far, and it's simpler. I'd say let's keep it as-is until we see its'a an issue.

frankbu (approved):

PR #8271, opened 2020-10-06
Suggestion from Jonh Wendell

sdake:
This comment came on slack for how to setup Istio on OpenShift. Without
this flag, OpenShift does not run.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


sdake:
nm, I see this value is set in preliminary but not 1.7.

I'll submit this there.

PR #8276, opened 2020-10-07
Update protocol selection page

howardjohn:
Add info about server first protocols

Make auto section the first since it is the most important

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
+++ b/content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
@@ -26,9 +26,9 @@ Istio can automatically detect HTTP and HTTP/2 traffic. If the protocol cannot a
 
 Some protocols are "Server First" protocols, which means the server will send the first bytes. In these cases, automatic protocol detection
 will not function correctly, as Istio will determine the protocol based on the initial bytes. If you are using a server first protocol, you should
-explicitly declare the protocol following the instructions below.
+follow the [explicit protocol selection instructions](#explicit-protocol-selection) to declare the protocol.
 
-The following protocols are known to be incompatible with automatic protocol selection, and will be treated as TCP if not explicitly declared:
+The following protocols are known to be incompatible with automatic protocol selection, and must also be explicitly declared or they will be treated as TCP:
 
 |Protocol|Port|
 |--------|----|
@@ -41,7 +41,7 @@ The following protocols are known to be incompatible with automatic protocol sel
 Since TLS communication is not server first, any TLS based traffic will support protocol detection, even if it is wrapping a protocol that does not support detection, such as MySQL.
 {{< /tip >}}
 
-## Manual protocol selection
+## Explicit protocol selection
 
 Protocols can be specified manually in the Service definition.
 
--- a/content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
+++ b/content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
@@ -12,9 +12,34 @@ owner: istio/wg-networking-maintainers
 test: no
 ---
 
-Istio supports proxying all TCP traffic by default, but in order to provide additional capabilities,
-such as routing and rich metrics, the protocol must be determined.
-This can be done automatically or explicitly specified.
+Istio supports proxying any TCP traffic. This includes HTTP, HTTPS, gRPC, as well as raw TCP protocols.
+In order to provide additional capabilities, such as routing and rich metrics, the protocol must be determined. This can be done automatically or explicitly specified.
+
+Non-TCP based protocols, such as UDP, are not proxied. These protocols will continue to function as normal, without
+any interception by the Istio proxy. They are not supported at gateways.
+
+## Automatic protocol selection
+
+Istio can automatically detect HTTP and HTTP/2 traffic. If the protocol cannot automatically be determined, traffic will be treated as plain TCP traffic.
+
+### Server first protocols
+
+Some protocols are "Server First" protocols, which means the server will send the first bytes. In these cases, automatic protocol detection
+will not function correctly, as Istio will determine the protocol based on the initial bytes. If you are using a server first protocol, you should
+explicitly declare the protocol following the instructions below.
+
+The following protocols are known to be incompatible with automatic protocol selection, and will be treated as TCP if not explicitly declared:
+
+|Protocol|Port|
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -12,9 +12,34 @@ owner: istio/wg-networking-maintainers
...
 ---
 
-Istio supports proxying all TCP traffic by default, but in order to provide additional capabilities,
-such as routing and rich metrics, the protocol must be determined.
-This can be done automatically or explicitly specified.
+Istio supports proxying any TCP traffic. This includes HTTP, HTTPS, gRPC, as well as raw TCP protocols.
+In order to provide additional capabilities, such as routing and rich metrics, the protocol must be determined. This can be done automatically or explicitly specified.
+
+Non-TCP based protocols, such as UDP, are not proxied. These protocols will continue to function as normal, without
+any interception by the Istio proxy. They are not supported at gateways.
```
frankbu, line 19: "They are not supported at gameways." is a little vague. Can you clarify? Would something like this be correct?
```suggestion
any interception by the Istio proxy but cannot be used in proxy-only components such as ingress or egress gateways.
```

howardjohn: I think that makes sense. Basically the idea is UDP is not "supported" its just not touched at all. This means you cannot use it at gateway, but in a pod it can send UDP out untouched.

content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -12,9 +12,34 @@ owner: istio/wg-networking-maintainers
...
+
+## Automatic protocol selection
+
+Istio can automatically detect HTTP and HTTP/2 traffic. If the protocol cannot automatically be determined, traffic will be treated as plain TCP traffic.
+
+### Server first protocols
+
+Some protocols are "Server First" protocols, which means the server will send the first bytes. In these cases, automatic protocol detection
+will not function correctly, as Istio will determine the protocol based on the initial bytes. If you are using a server first protocol, you should
+explicitly declare the protocol following the instructions below.
```
frankbu, line 29:
```suggestion
follow the [explicit protocol selection instructions](#explicit-protocol-selection) to declare the protocol.
```

howardjohn: Should it be `#manual-protocol-selection` or do I not understand how this works?

content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -12,9 +12,34 @@ owner: istio/wg-networking-maintainers
...
+| SMTP   |25  |
+| DNS    |53  |
+| MySQL  |3306|
+| MongoDB|27017|
+
+{{< tip >}}
+Since TLS communication is not server first, any TLS based traffic will support protocol detection, even if it is wrapping a protocol that does not support detection, such as MySQL.
+{{< /tip >}}
 
 ## Manual protocol selection
```
frankbu, line 44:
```suggestion
## Explicit protocol selection
```

content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -12,9 +12,34 @@ owner: istio/wg-networking-maintainers
...
+
+Istio can automatically detect HTTP and HTTP/2 traffic. If the protocol cannot automatically be determined, traffic will be treated as plain TCP traffic.
+
+### Server first protocols
+
+Some protocols are "Server First" protocols, which means the server will send the first bytes. In these cases, automatic protocol detection
+will not function correctly, as Istio will determine the protocol based on the initial bytes. If you are using a server first protocol, you should
+explicitly declare the protocol following the instructions below.
+
+The following protocols are known to be incompatible with automatic protocol selection, and will be treated as TCP if not explicitly declared:
```
frankbu, line 31:
```suggestion
The following protocols are known to be incompatible with automatic protocol selection, and must also be explicitly declared or they will be treated as TCP:
```

content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
frankbu: I thought renaming the section (manual -> explicit) would be better, since the text always uses the word explicit instead of manual. But that's just my preference.

howardjohn: I agree

content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -12,11 +12,36 @@ owner: istio/wg-networking-maintainers
...
+|Protocol|Port|
+|--------|----|
+| SMTP   |25  |
+| DNS    |53  |
+| MySQL  |3306|
+| MongoDB|27017|
+
+{{< tip >}}
+Since TLS communication is not server first, any TLS based traffic will support protocol detection, even if it is wrapping a protocol that does not support detection, such as MySQL.
+{{< /tip >}}
```
frankbu, line 42: One more nit. The incompatible protocols part looks like it's in the `Server first protocols` subsection, so I suggest either removing the section heading or reordering the text like this:
```suggestion
The following protocols are known to be incompatible with automatic protocol selection, and must be
[explicitly declared](#explicit-protocol-selection) or they will be treated as TCP:

|Protocol|Port|
|--------|----|
| SMTP   |25  |
| DNS    |53  |
| MySQL  |3306|
| MongoDB|27017|

{{< tip >}}
Since TLS communication is not server first, any TLS based traffic will support protocol detection, even if it is wrapping a protocol that does not support detection, such as MySQL.
{{< /tip >}}

### Server first protocols

Some protocols are "Server First" protocols, which means the server will send the first bytes. In these cases, automatic protocol detection
will not function correctly, as Istio will determine the protocol based on the initial bytes. If you are using a server first protocol, you should
follow the explicit protocol selection instructions, below, to declare the protocol.
```

frankbu (approved):

PR #8309, opened 2020-10-15
Fixes for multicluster tests

nmittler:
[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


stevenctl (approved):

content/en/docs/setup/install/multicluster/index.md
```diff
@@ -620,13 +620,13 @@ Deploy the `helloworld-v1` application to `cluster1`:
...
     -f @samples/helloworld/helloworld.yaml@ \
-    -l app=helloworld -l version=v1 -n sample
+    -l version=v1 -n sample
 {{< /text >}}
 
 Confirm the `helloworld-v1` pod status:
 
 {{< text bash >}}
-$ kubectl get pod --context="${CTX_CLUSTER1}" -n sample
+$ kubectl get pod --context="${CTX_CLUSTER1}" -n sample -l app=helloworld
```
frankbu, line 629: The `-l app=helloworld` isn't really needed since we just created the sample namespace, so we know that the only pod that should be running is the helloworld one. I think it's just a distraction that may confuse some people.

frankbu: Oh, I see, you're not doing things in the test in the exact same order as the doc instructions, so the app label is needed. On the other hand, it is generally best to execute the test instructions in the same order that the doc tells users to run them if you want to really "test the instructions" as documented.

nmittler: In this case, it's a performance optimization. I want to kick off all the deploys first and then check later.

nmittler:
@ericvn @frankbu this is ready to go now ... PTAL

frankbu (approved):

nmittler:
@ericvn @frankbu where are things at with the release? Does this need to be cherry-picked?

frankbu:
@nmittler no need to cherrypick anything. Master branch is the 1.8 release docs. We cut the release branch when the release goes out.

PR #8310, opened 2020-10-16
Update reference script to not strip leading istio.io on /vx.y

ericvn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
https://github.com/istio/istio.io/pull/8238 has been held out for a few weeks because when importing links like <a href="https://istio.io/v1.5/docs/reference/config/installation-options/"> we typically strip the leading `https://istio.io` leaving `/v1.5...`. The linter fails on that link since it doesn't exist. This is trying to not strip the `https://istio.io` when the link starts with a `v` or `V`.

ericvn:
Not quite correct yet. Seeing links like 
/docs/reference/config/istio.mesh.v1alpha1/ocs/reference/config/networking/destination-rule/ in the generated output

ericvn:
This PR is ready to go. The code changes in commits #1 and #2 will prevent the stripping of the `https://istio.io` in the case where the first character after the succeeding / is a v or a V. This leaves the full URL for the one failing case.

Commit #3 is the resulting changes running `make update_ref_docs` with the change.

frankbu (approved):

PR #8323, opened 2020-10-20
Extra notes on TLS for protocol sniffing docs

howardjohn:

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
+++ b/content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
@@ -38,6 +38,13 @@ Some protocols are "Server First" protocols, which means the server will send th
 will not function correctly, as Istio will determine the protocol based on the initial bytes. If you are using a server first protocol, you should
 follow the explicit protocol selection instructions, below, to declare the protocol.
 
+Additionally, plaintext requests to servers configured in `mTLS` [`PERMISSIVE` mode](/docs/reference/config/security/peer_authentication/#PeerAuthentication-MutualTLS-Mode) will fail. 
+This can be addressed with any of the following steps:
+
+* Configure `mTLS` mode `STRICT` for the server. This will enforce all traffic is TLS encrypted.
+* Configure `mTLS` mode `DISABLE` for the server. This will disable the TLS sniffing, allowing server first protocols to be used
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -38,6 +38,13 @@ Some protocols are "Server First" protocols, which means the server will send th
 will not function correctly, as Istio will determine the protocol based on the initial bytes. If you are using a server first protocol, you should
 follow the explicit protocol selection instructions, below, to declare the protocol.
 
+Additionally, plaintext requests to servers configured in `mTLS` [`PERMISSIVE` mode](/docs/reference/config/security/peer_authentication/#PeerAuthentication-MutualTLS-Mode) will fail.
```
frankbu, line 41: I'm not really following this, maybe because I don't know much about server first protocols? IIUC it's talking about a server that implements a server first protocol, so the previous paragraph says it needs to be manually declared in the service port name. The following tip says that TLS is not server first, so it always works. This would lead me to believe that the only thing that is special about PERMISSIVE mode is that the request may or may not be mTLS, so the protocol should be manually declared, so that it will work either way. If that's all this is about, then I would just say that. Otherwise, I think it needs more explanation about why it will fail. Or both?

howardjohn: Basically, there are many parts.

First is outbound. If app sends TLS directly, its never server first. Otherwise, if its server first, it will fail if we do not declare the protocol, since we try to detect the protocol.

Next is inbound. Inbound is more complicated. If the inbound request is TLS (mTLs or direct from app, either works), we have no problems. If its permissive mode, we attempt to detect if its TLS - if it was not TLS, it would hang, and fail. DISABLE is ok since we do not detect, and STRICT is ok since its always TLS. Then, we also attempt to detect the protocol unless it is explicitly declared. 

So: `outboundSuccess == protocolNamed || requestIsTls`

`inboundSuccess == requestIsTls || (protocolNamed && (STRICT || DISABLE))`

frankbu:
This is done, moved to https://preliminary.istio.io/latest/docs/ops/deployment/requirements#server-first-protocols

PR #8340, opened 2020-10-22
Add regex matching semantics documentation

xulingqing:
Add regex matching semantics documentation in Security Value Matching section. 
- code change https://github.com/istio/istio/pull/27984

istio.io website change:
- istio/istio.io    https://github.com/istio/istio.io/pull/8340
- istio/api          https://github.com/istio/api/pull/1707

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/concepts/security/index.md
+++ b/content/en/docs/concepts/security/index.md
@@ -656,6 +656,9 @@ schemas:
    that a field must be present, use the `fieldname: ["*"]`format. This is
    different from leaving a field unspecified, which means match anything,
    including empty.
+- Regex match: `"regex:"` is used as special prefix. Anything after "regex:" is 
+   interpreted as regular expression. For example, `'regex:abc.com\/[0-9]?'`
+   matches `"abc.com/"`, `"abc.com/0"`, etc.
 
 There are a few exceptions. For example, the following fields only support exact
 match:
@@ -665,7 +668,7 @@ match:
 - The `ports` field under the `to` section
 
 The following example policy allows access at paths with the `/test/*` prefix
-or the `*/info` suffix.
+or the `*/info` suffix, or paths that match regex expression `abc.com\/[0-9]?`.
 
 {{< text yaml >}}
 apiVersion: security.istio.io/v1beta1
@@ -681,7 +684,7 @@ spec:
   rules:
   - to:
     - operation:
-        paths: ["/test/*", "*/info"]
+        paths: ["/test/*", "*/info", 'regex:abc.com\/[0-9]?']
 {{< /text >}}
 
 #### Exclusion matching


content/en/docs/concepts/security/index.md
```diff
@@ -656,6 +656,9 @@ schemas:
    that a field must be present, use the `fieldname: ["*"]`format. This is
```
frankbu, line 656:
```suggestion
   that a field must be present, use the `fieldname: ["*"]` format. This is
```

content/en/docs/concepts/security/index.md
```diff
@@ -656,6 +656,9 @@ schemas:
...
    different from leaving a field unspecified, which means match anything,
    including empty.
+- Regex match: `"regex:"` is used as special prefix. Anything after "regex:" is
```
frankbu, line 659:
```suggestion
- Regex match: If a string begins with `regex:`, the following text is 
```

content/en/docs/concepts/security/index.md
```diff
@@ -656,6 +656,9 @@ schemas:
    that a field must be present, use the `fieldname: ["*"]`format. This is
    different from leaving a field unspecified, which means match anything,
    including empty.
+- Regex match: `"regex:"` is used as special prefix. Anything after "regex:" is 
+   interpreted as regular expression. For example, `'regex:abc.com\/[0-9]?'`
```
frankbu, line 660:
```suggestion
   interpreted as a regular expression. For example, `'regex:abc.com\/[0-9]?'`
```

yangminzhu (changes requested):
This feature is disabled by default and requires to set a flag to opt-in, so it shouldn't be documented in the concept page as if it's the default behavior.

Instead I suggest to move the changes to be a new section in https://istio.io/latest/docs/tasks/security/authorization/authz-http/#configure-access-control-for-workloads-using-http-traffic, and add instructions about how to set the flag to test it.

Also putting this on hold as it is not in 1.8, should not be merged before 1.8 is released.

xulingqing:
> This feature is disabled by default and requires to set a flag to opt-in, so it shouldn't be documented in the concept page as if it's the default behavior.
> 
> Instead I suggest to move the changes to be a new section in https://istio.io/latest/docs/tasks/security/authorization/authz-http/#configure-access-control-for-workloads-using-http-traffic, and add instructions about how to set the flag to test it.
> 
> Also putting this on hold as it is not in 1.8, should not be merged before 1.8 is released.

Sure, willdo. Thanks!

frankbu:
Closing for now. @xulingqing please reopen or send a new PR when its ready to be merged.

PR #8346, opened 2020-10-23
Add wait for gateway to multicluster tests

nmittler:
This required some other changes WRT verification:

- Change __cmp_like to allow for not accepting <pending> for an IP address.

- Change __verify_with_retry to use a timeout rathan than number of retries. This is a more intuitive interface and aligns with the way we do retries in istio/istio. I also got rid of exponential backoff and allow both the timeout and delay between retries to be configured.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/multicluster/_index.md
+++ b/content/en/docs/setup/install/multicluster/_index.md
@@ -27,3 +27,11 @@ This guide covers some of the most common concerns when creating a
 - [Control plane topologies](/docs/ops/deployment/deployment-models#control-plane-models):
   multiple {{< gloss "primary cluster" >}}primary clusters{{< /gloss >}},
   a primary and {{< gloss >}}remote cluster{{< /gloss >}}
+
+{{< tip >}}
+For meshes that span more than two clusters, you can extend the steps in this
+guide to configure more complex topologies.
+
+See [deployment models](/docs/ops/deployment/deployment-models) for more
+information.
+{{< /tip >}}
--- a/tests/README.md
+++ b/tests/README.md
@@ -182,36 +182,42 @@ expected output. The framework includes the following built-in verify functions:
 1. **`_verify_same`** `func` `expected`
 
    Runs `func` and compares the output with `expected`. If they are not the same,
-   exponentially back off and try again, 7 times by default. The number of retries
-   can be changed by setting the `VERIFY_RETRIES` environment variable.
+   wait a second and try again, up to two minutes by default. The retry behavior
+   can be changed by setting the `VERIFY_TIMEOUT` and `VERIFY_DELAY` environment
+   variables.
 
 1. **`_verify_contains`** `func` `expected`
 
    Runs `func` and compares the output with `expected`. If the output does not
-   contain the substring `expected`, exponentially back off and try again, 7 times
-   by default. The number of retries can be changed by setting the `VERIFY_RETRIES`
-   environment variable.
+   contain the substring `expected`,
+   wait a second and try again, up to two minutes by default. The retry behavior
+   can be changed by setting the `VERIFY_TIMEOUT` and `VERIFY_DELAY` environment
+   variables.
 
 1. **`_verify_not_contains`** `func` `expected`
 
    Runs `func` and compares the output with `expected`. If the command execution fails
    or the output contains the substring `expected`,
-   exponentially back off and try again, 7 times by default. The number of retries
-   can be changed by setting the `VERIFY_RETRIES` environment variable.
+   wait a second and try again, up to two minutes by default. The retry behavior
+   can be changed by setting the `VERIFY_TIMEOUT` and `VERIFY_DELAY` environment
+   variables.
 
 1. **`_verify_elided`** `func` `expected`
 
    Runs `func` and compares the output with `expected`. If the output does not
    contain the lines in `expected` where "..." on a line matches one or more lines
-   containing any text, exponentially back off and try again, 7 times by default.
-   The number of retries can be changed by setting the `VERIFY_RETRIES` environment
-   variable.
+   containing any text,
+   wait a second and try again, up to two minutes by default. The retry behavior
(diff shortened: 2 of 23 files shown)


tests/README.md
```diff
@@ -182,44 +182,53 @@ expected output. The framework includes the following built-in verify functions:
...
    Like implies:
 
    - Same number of lines
    - Same number of whitespace-seperated tokens per line
    - Tokens can only differ in the following ways:
 
      1. different elapsed time values (e.g., `30s` is like `5m`)
-     1. different ip values (e.g., `172.21.0.1` is like `10.0.0.31`)
+     1. different ip values (e.g., `172.21.0.1` is like `10.0.0.31`). Allows
+         `<none>` and `<pending>` by default. This can be customized by setting
```
frankbu, line 229: Allowing `<pending>` by default seems wrong, since that normally means that it's not ready yet, and that we should really keep retrying until we get a real IP. Usually it's only acceptable if it's "perpetually" pending.

Given the special cases where this is used, do we need two env variables, or would it be good enough to just have one flag `ALLOW_NO_IP` which defaults to false, but when true would allow either `<none` or `<pending>`?

tests/README.md
```diff
@@ -182,44 +182,53 @@ expected output. The framework includes the following built-in verify functions:
 1. **`_verify_same`** `func` `expected`
 
    Runs `func` and compares the output with `expected`. If they are not the same,
-   exponentially back off and try again, 7 times by default. The number of retries
-   can be changed by setting the `VERIFY_RETRIES` environment variable.
+   wait a second and try again, up to two minutes by default. The retry behavior
+   can be changed by setting the `VERIFY_TIMEOUT` and `VERIFY_DELAY` environment
```
frankbu, line 186: Why do you want to change to a fixed delay, instead of exponential back off? Previously it would only call the function 7 times before giving up in 2 mins, but now it will call the func about 120 times before timing out when there really is a problem.

nmittler: The number of calls isn't really a concern from the perspective of a test author or someone debugging a test ... it's duration (i.e. timeout).  Using timeout also provides more expected behavior in that the test terminates when you expect it to. If using number of attempts, you actually have to factor in how long each attempt takes, which could be several seconds. This means you can have tests that actually take way longer than 2 minutes.

Exponential backoff is more than we need here in general and also leads to confusing behavior WRT the overall duration. In addition, I can't think of anything we would actually want to exponentially backoff ... there's no real advantage. Using a delay and timeout is a much simpler UI from a test authoring perspective.

tests/README.md
nmittler: I'm actually just maintaining current behavior here. I had to make this change because we weren't retying on `<pending>`. I could change the default, but I'd be concerned other tests might fail .. but maybe it's not a big deal.

nmittler: > Given the special cases where this is used, do we need two env variables, or would it be good enough to just have one flag ALLOW_NO_IP which defaults to false, but when true would allow either <none or <pending>?

The meaning of `none` and `pending` is significantly different. Not sure you can capture that in one flag ... we could try?

tests/README.md
frankbu: The expected bahavior argument goes both ways. If we don't know how long the function itself will take, we don't know how many retries we'll get before the timeout, so to handle things like curl with network glitches, we can't be sure how many retries we're going to get.

Anyway, I agree that timeout/delay is simpler for people to understand, so OK with the change. I still wonder if ~120 retries by default is too much. Maybe the default delay should be 2 secs?

nmittler: > Maybe the default delay should be 2 secs?

Possibly, but I'm not sure if there is any real concern. If it becomes a problem we can always bump it up.

tests/README.md
frankbu: So the history of this was for setup docs that say `the external IP may be <none> or perpetually <pending>` in which case a nodeport is used for the ingress IP. The framework was trying to handle the case that it's being used on a platform that has no external LB. Matching the `<pending>` then, is only reasonable if there is some kind of `_wait_for` call before the verify call, so retries should never really happen in that case. If, on the other hand, we're using the verify retries as the way to wait for the IP to be ready (as you are doing), the `<pending>` check is wrong, because it is not the final expected state.

So, given all this, I'm thinking maybe should just remove the `<pending>` and `<none>` checks from `_verify_like`. I don't think we do want to test for the nodeport case, and even if we decide to in the future, we will probably do the verify differently anyway. WDYT?

nmittler: ok, let me try that out and see what happens.

content/en/docs/setup/install/multicluster/_index.md
```diff
@@ -22,7 +22,7 @@ This guide covers some of the most common concerns when creating a
 {{< gloss >}}multicluster{{< /gloss >}} mesh:
 
 - [Network topologies](/docs/ops/deployment/deployment-models#network-models):
-  one or two networks
+  one or more networks.
```
frankbu, line 25: If not being specific about the cases covered by the doc, then maybe you should also change line 29 for consistency:
```
primary and {{< gloss "remote cluster">}}remote clusters{{< /gloss >}}
```

nmittler: I've reverted this and added a tip below.

nmittler:
@frankbu PTAL

I've disabled matching `<pending>` and `<none>` by default, but had to leave that option in for now due to https://github.com/istio/istio.io/issues/8353.  Once that's resolved, we should be able to remove that logic altogether.

tests/README.md
nmittler: Looks like there is one test that requires `<pending>`. I've created an issue and captured the details below.

frankbu (approved):

PR #8369, opened 2020-10-26
fix vms gateway

stevenctl:
cc @howardjohn @nmittler 

Not sure when this moved to `multicluster/`. Is that the appropriate home since this is also used for VMs? 

This would also need to be updated with instructions to deploy eastwest gateway.

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -57,7 +57,7 @@ Install Istio and expose the control plane so that your virtual machine can acce
 1. Expose the control plane using the provided sample configuration.
 
     {{< text bash >}}
-    $ kubectl apply -f @samples/istiod-gateway/istiod-gateway.yaml@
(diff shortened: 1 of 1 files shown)


ericvn (approved):
Matches rename from https://github.com/istio/istio/pull/27222 and is in istio/istio release-1.8 branch.

content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -57,7 +57,7 @@ Install Istio and expose the control plane so that your virtual machine can acce
 1. Expose the control plane using the provided sample configuration.
 
     {{< text bash >}}
-    $ kubectl apply -f @samples/istiod-gateway/istiod-gateway.yaml@
+    $ kubectl apply -f @samples/multicluster/expose-istiod.yaml@
```
frankbu, line 60: This seems like a bad directory for a yaml file that is used for accessing istiod in many places, not just multcluster (VM integration, external control plane, primary cluster control plane).

PR #8373, opened 2020-10-26
Remove production usage as Helm support in pre-alpha

nrjpoddar:

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/helm/index.md
+++ b/content/en/docs/setup/install/helm/index.md
@@ -14,7 +14,7 @@ This feature is actively in development and is considered `pre-alpha`.
 {{< /warning >}}
 
 Follow this guide to install and configure an Istio mesh using
-[Helm](https://helm.sh/docs/) for in-depth evaluation or production use.
+[Helm](https://helm.sh/docs/) for in-depth evaluation.
 
 The Helm charts used in this guide are the same underlying charts used when
 installing Istio via [Istioctl](/docs/setup/install/istioctl/) or the
--- a/content/en/docs/setup/install/helm/index.md
+++ b/content/en/docs/setup/install/helm/index.md
@@ -1,7 +1,7 @@
 ---
 title: Install with Helm [Experimental]
 linktitle: Install with Helm
-description: Install and configure Istio for in-depth evaluation or production use.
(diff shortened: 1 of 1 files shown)


nrjpoddar:
cc @ericvn

sdake (commented):
I have reviewed the full document as a whole.

1. I would change line 13 "pre-alpha" to "experimental." Istio has an [experimental feature status](https://github.com/istio/enhancements/issues/new?template=experimental_feature.md), but not a pre-alpha stage.
1. I am not keen on line 115 - the migration from non-Helm environments. I think the stand-alone operator for Istio needs to be stopped first (deletion of the deployment as a minimum - the whole manifest ideally) before proceeding down this path.
1. I think it is important to clarify the uninstall warning n 122 - the CRs need not be deleted if the CRDs don't need to be deleted. Do the CRDs need to be deleted?
1. Are the optional steps really optional? I would think the ingress chart is needed in most cases unless using @ostromart 's operational guide for gateways.
1. You may consider linking to @ostromart 's operational gate for gateways here as well, to show more advanced operational procedures.
1. the workflow on 198 is not quite right. The gist is you go from 1.7.0->canary (1.8.0) - test the canary then ->istio 1.8.0 (not canary). This is a two-step process. I think this would be a stretch goal for the documentation.
1. Please delete lines 246-254. There is no reason that I am aware of to delete CRDs other than leaking the CRDs has some minor cost with Kubernetes etcde. While they leak from the cluster, 

If you prefer I file these issues separately, I can do that, or you can expand this PR to address these problems. Your choice. Would like to see the outstanding issues resolved in the documentation for 1.7 - as the problems pointed out are likely to cause problems.

@costinm @howardjohn @ostromart thoughts on my review welcome.

nrjpoddar:
Thanks @sdake for your detailed feedback. I want to keep this PR for a small fix so let's file an issue so we can discuss your recommendations and make subsequent PRs. While I agree with a lot of your suggestions but not with all of them so let's start an issue to hash it out.

frankbu (approved):

PR #8385, opened 2020-10-28
Update istioctl installation docs

ostromart:


ericvn:
If I reviewed this correctly, you basically removed the `Configure component settings` section and moved the `Configure gateways` to under the `Customize Kubernetes settings`.

content/en/docs/setup/install/istioctl/index.md
```diff
@@ -364,40 +364,89 @@ The `IstioOperator` API defines components as shown in the table below:
 `cni` |
 `istiodRemote` |
 
-### Configure component settings
+These components share a common API for setting Kubernetes resources.
```
frankbu, line 367: Removing the detailed `configure compnent settings` section seems reasonable, but replacing it with just this one sentence seems to have gone a little too far. What's the point of showing the table? I think it's to tell people where in API to set values for a given component. So, at a minimum, I think it should still say something like:
```suggestion
The configurable settings for each of these components are available in the API under `components.<component name>`. 
For example, to use the API to change (to false) the `enabled` setting for the `telemetry` component, use `--set components.telemetry.enabled=false`, or set it in an `IstioOperator` resource like this:

{{< text yaml >}}
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  components:
    telemetry:
      enabled: false
{{< /text >}}

All of the components also share a common API for setting Kubernetes resources, under `components.<component name>.k8s`, as described in the following section.
```

content/en/docs/setup/install/istioctl/index.md
```diff
@@ -364,40 +364,89 @@ The `IstioOperator` API defines components as shown in the table below:
...
+          operator: Exists
 {{< /text >}}
 
-1. Use the `telemetry_off.yaml` overlay file with the `istioctl install` command:
+Use `istioctl install` to apply the modified settings to the cluster:
 
-{{< text bash >}}
-$ istioctl install -f telemetry_off.yaml
+{{< text syntax="bash" repo="operator" >}}
+$ istioctl install -f samples/operator/pilot-k8s.yaml
```
frankbu, line 425: The merge conflict is because of a recent fix in the sample yaml files paths. If you fix this, you're version is correct.
```suggestion
$ istioctl install -f operator/samples/pilot-k8s.yaml
```

content/en/docs/setup/install/istioctl/index.md
ostromart: done, using pilot (telemetry is deprecated)

content/en/docs/setup/install/istioctl/index.md
ostromart: done

frankbu (approved):

PR #8390, opened 2020-10-28
update authz docs for remote.ip

kylebevans:
remote.ip has been added as an Authorization Condition and the Ingress
Gateway Authorization task has been updated to include it.

[ x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/security/authorization/authz-ingress/index.md
+++ b/content/en/docs/tasks/security/authorization/authz-ingress/index.md
@@ -29,7 +29,7 @@ through the Istio ingress gateway with this command:
 * Turn on RBAC debugging in Envoy for the ingress gateway:
 
     {{< text bash >}}
-    $ kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do istioctl proxy-config log "$pod" -n istio-system --level rbac:debug; done
+    $ kubectl get pods -n istio-system -o name -l istio=ingressgateway | sed 's|pod/||' | while read -r pod; do istioctl proxy-config log "$pod" -n istio-system --level rbac:debug; done
     {{< /text >}}
 
 *  Follow the instructions in
@@ -270,7 +270,7 @@ EOF
 Find your original client IP address if you don't know it and assign it to a variable:
 
 {{< text bash >}}
-$ CLIENT_IP=$(kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do kubectl logs "$pod" -n istio-system | grep remoteIP; done | head -1 | awk -F, '{print $3}' | awk -F: '{print $2}' | sed 's/ //') && echo "$CLIENT_IP"
+$ CLIENT_IP=$(kubectl get pods -n istio-system -o name -l istio=ingressgateway | sed 's|pod/||' | while read -r pod; do kubectl logs "$pod" -n istio-system | grep remoteIP; done | head -1 | awk -F, '{print $3}' | awk -F: '{print $2}' | sed 's/ //') && echo "$CLIENT_IP"
 192.168.10.15
 {{< /text >}}
 
@@ -300,7 +300,7 @@ EOF
 Find your original client IP address if you don't know it and assign it to a variable:
 
 {{< text bash >}}
-$ CLIENT_IP=$(kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do kubectl logs "$pod" -n istio-system | grep remoteIP; done | head -1 | awk -F, '{print $4}' | awk -F: '{print $2}' | sed 's/ //') && echo "$CLIENT_IP"
+$ CLIENT_IP=$(kubectl get pods -n istio-system -o name -l istio=ingressgateway | sed 's|pod/||' | while read -r pod; do kubectl logs "$pod" -n istio-system | grep remoteIP; done | head -1 | awk -F, '{print $4}' | awk -F: '{print $2}' | sed 's/ //') && echo "$CLIENT_IP"
 192.168.10.15
 {{< /text >}}
 
@@ -403,7 +403,7 @@ different client IP to verify the request is allowed.
 * If you are not getting the responses you expect, view the ingress gateway logs which should show RBAC debugging information:
 
     {{< text bash >}}
-    $ kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do kubectl logs "$pod" -n istio-system; done
+    $ kubectl get pods -n istio-system -o name -l istio=ingressgateway | sed 's|pod/||' | while read -r pod; do kubectl logs "$pod" -n istio-system; done
     {{< /text >}}
 
 ## Clean up
--- a/content/en/docs/tasks/security/authorization/authz-ingress/index.md
+++ b/content/en/docs/tasks/security/authorization/authz-ingress/index.md
@@ -63,7 +63,7 @@ Caching and propagation overhead can cause a delay.
 
 * If a packet goes through an external load balancer and/or kube-proxy, then the original source IP address of the client is lost.  There are a few ways to handle this if you need the original client IP for logging or security purposes:
 
-      1. If you are using an HTTP/HTTPS external load balancer, it can put the original client IP address in the X-Forwarded-For header.  Istio can extract the client IP address from this header with some configuration.  See [Configuring Gateway Network Topology](/latest/docs/ops/configuration/traffic-management/network-topologies/). Quick example if using a single load balancer in front of Kubernetes:
+      1. If you are using an HTTP/HTTPS external load balancer, it can put the original client IP address in the X-Forwarded-For header.  Istio can extract the client IP address from this header with some configuration.  See [Configuring Gateway Network Topology](/docs/ops/configuration/traffic-management/network-topologies/). Quick example if using a single load balancer in front of Kubernetes:
 
         {{< text yaml >}}
         apiVersion: install.istio.io/v1alpha1
--- a/content/en/docs/tasks/security/authorization/authz-ingress/index.md
+++ b/content/en/docs/tasks/security/authorization/authz-ingress/index.md
@@ -7,12 +7,7 @@ owner: istio/wg-security-maintainers
 test: yes
 ---
 
-This task shows you how to enforce access control on an Istio ingress gateway
-using an authorization policy.
-
-An Istio authorization policy supports IP-based allow lists or deny lists as well as
-the attribute-based allow lists or deny lists previously provided by Mixer policy.
-The Mixer policy is deprecated in *5 and not recommended for production use.
+This task shows you how to enforce IP-based access control on an Istio ingress gateway using an authorization policy.
 
 ## Before you begin
 
@@ -61,37 +56,7 @@ Caching and propagation overhead can cause a delay.
 
 ## Source IP address of the original client
 
-If a packet goes through an external proxy load balancer and/or kube-proxy, then the original source IP address of the client is lost.  Below are some strategies for preserving the original client IP for logging or security purposes.  For reference, here are the types of load balancers created by Istio with a `LoadBalancer` service on popular managed Kubernetes environments:
-
-|Cloud Provider | Load Balancer Name            | Load Balancer Type
-----------------|-------------------------------|-------------------
-|AWS EKS        | Classic Elastic Load Balancer | TCP Proxy
-|GCP GKE        | TCP/UDP Network Load Balancer | Network
-|Azure AKS      | Azure Load Balancer           | Network
-|DO DOKS        | Load Balancer                 | Network
-
-{{< tip >}}
-You can instruct AWS EKS to create a Network Load Balancer when you install Istio by using a `serviceAnnotation` like below:
-
-{{< text yaml >}}
-apiVersion: install.istio.io/v1alpha1
-kind: IstioOperator
-spec:
-  meshConfig:
-    accessLogEncoding: JSON
-    accessLogFile: /dev/stdout
-  components:
-    ingressGateways:
-    - enabled: true
-      k8s:
-        hpaSpec:
-          maxReplicas: 10
-          minReplicas: 5
-        serviceAnnotations:
-          service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
-{{< /text >}}
-
-{{< /tip >}}
+If a packet goes through an external proxy load balancer and/or kube-proxy, then the original source IP address of the client is lost.  Below are some strategies for preserving the original client IP for logging or security purposes.
 
 {{< tabset category-name="lb" >}}
 
@@ -185,6 +150,38 @@ spec:
 {{< /tab >}}
 
 {{< /tabset >}}
+
+For reference, here are the types of load balancers created by Istio with a `LoadBalancer` service on popular managed Kubernetes environments:
+
+|Cloud Provider | Load Balancer Name            | Load Balancer Type
+----------------|-------------------------------|-------------------
+|AWS EKS        | Classic Elastic Load Balancer | TCP Proxy
+|GCP GKE        | TCP/UDP Network Load Balancer | Network
+|Azure AKS      | Azure Load Balancer           | Network
+|DO DOKS        | Load Balancer                 | Network
+
+{{< tip >}}
+You can instruct AWS EKS to create a Network Load Balancer when you install Istio by using a `serviceAnnotation` like below:
+
+{{< text yaml >}}
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  meshConfig:
+    accessLogEncoding: JSON
+    accessLogFile: /dev/stdout
+  components:
+    ingressGateways:
+    - enabled: true
+      k8s:
+        hpaSpec:
+          maxReplicas: 10
+          minReplicas: 5
+        serviceAnnotations:
+          service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
+{{< /text >}}
+
+{{< /tip >}}
 
 ## IP-based allow list and deny list
 
--- a/content/en/docs/tasks/security/authorization/authz-ingress/index.md
+++ b/content/en/docs/tasks/security/authorization/authz-ingress/index.md
@@ -31,17 +31,11 @@ through the Istio ingress gateway with this command:
     $ kubectl apply -f <(istioctl kube-inject -f @samples/httpbin/httpbin-gateway.yaml@) -n foo
     {{< /text >}}
 
-* See [Source IP for Services with `Type=NodePort`](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport)
-for more information. Update the ingress gateway to set `externalTrafficPolicy: local` to preserve the
-original client source IP on the ingress gateway using the following command:
+* Turn on RBAC debugging in Envoy for the ingress gateway:
 
     {{< text bash >}}
-    $ kubectl patch svc istio-ingressgateway -n istio-system -p '{"spec":{"externalTrafficPolicy":"Local"}}'
+    $ kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do istioctl proxy-config log "$pod" -n istio-system --level rbac:debug; done
     {{< /text >}}
-
-*  Follow the instructions in
-    [Determining the ingress IP and ports](/docs/tasks/traffic-management/ingress/ingress-control/#determining-the-ingress-ip-and-ports)
-    to define the `INGRESS_HOST` and `INGRESS_PORT` environment variables.
 
 * Verify that the `httpbin` workload and ingress gateway are working as expected using this command:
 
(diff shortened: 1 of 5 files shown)


kylebevans:
The two doc.test failures are for different tests unrelated to this PR.

The linting failure is because /latest/docs/ops/configuration/traffic-management/network-topologies/ gives a 404.  I'm at a loss as to why that's happening since that doc is there. Does anyone else see the issue?

content/en/docs/tasks/security/authorization/authz-ingress/test.sh
```diff
@@ -37,24 +37,35 @@ _set_ingress_environment_variables
...
+
+snip_ipbased_allow_list_and_deny_list_9
+_wait_for_istio authorizationpolicy istio-system ingress-policy
+
+snip_ipbased_allow_list_and_deny_list_10
+_wait_for_istio authorizationpolicy istio-system ingress-policy
+
+snip_ipbased_allow_list_and_deny_list_11
+
+snip_ipbased_allow_list_and_deny_list_12
```
frankbu, line 68: Hmm, is there nothing to verify for any of these snips?

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -86,7 +169,28 @@ Create the authorization policy:
...
+      selector:
+        matchLabels:
+          app: istio-ingressgateway
+      action: ALLOW
+      rules:
+      - from:
+        - source:
+            remoteIpBlocks: ["1.2.3.4", "5.6.7.0/24"]
     EOF
     {{< /text >}}
```
frankbu, line 195: Instead of making these remoteIpBlocks examples another step, you should put the two examples into a {{< tabset >}} in the previous step.
https://preliminary.istio.io/latest/about/contribute/shortcodes/#use-tabs

content/en/docs/tasks/security/authorization/authz-ingress/test.sh
kylebevans: Great question, the original docs use examples with ipBlocks and assume that the cluster is configured with "externalTrafficPolicy: local" so that the actual original client IP address is available as the packet source address in k8s.  The $CLIENT_IP env var was also originally set with this in mind.  However, with the remoteIpBlocks examples, the way remote.ip gets set is highly dependent on the k8s environment: which load balancer, proxy protocol or X-Forwarded-For header, and meshConfig.

I don't know of a clean way to get the remoteIP for the CLIENT_IP setting, so I left the remoteIP examples without a verify.  The original ipBlocks examples still have the verify though.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
kylebevans: Good point, I'll change it to a tab.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,40 +31,123 @@ through the Istio ingress gateway with this command:
...
+
+* What if the Pod that is handling traffic from the `NodePort` or `LoadBalancer` isn't running on the worker node that received the traffic?  Kubernetes has its own internal proxy called kube-proxy that receives the packets and forwards them to the correct node.
+
+## Source IP address of the original client
+
+* If a packet goes through an external load balancer and/or kube-proxy, then the original source IP address of the client is lost.  There are a few ways to handle this if you need the original client IP for logging or security purposes:
+
+      1. If you are using an HTTP/HTTPS external load balancer, it can put the original client IP address in the X-Forwarded-For header.  Istio can extract the client IP address from this header with some configuration.  See [Configuring Gateway Network Topology](/docs/ops/configuration/traffic-management/network-topologies/). Quick example if using a single load balancer in front of Kubernetes:
```
yangminzhu, line 66: would it be better if we put the 3 options in a tab?

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -114,7 +231,28 @@ Create the authorization policy:
       rules:
       - from:
         - source:
-           ipBlocks: ["1.2.3.4", "5.6.7.0/24", "$CLIENT_IP"]
+            ipBlocks: ["1.2.3.4", "5.6.7.0/24", "$CLIENT_IP"]
+    EOF
+    {{< /text >}}
+
+1. With `remoteIpBlocks`:
```
yangminzhu, line 238: also probably use tab here, I think we want the user to choose one of {ipBlocks, remoteIpBlocks} but not both.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -144,7 +282,28 @@ not allowed to access the ingress gateway:
       rules:
       - from:
         - source:
-           ipBlocks: ["$CLIENT_IP"]
+            ipBlocks: ["$CLIENT_IP"]
+    EOF
+    {{< /text >}}
+
+1. With `remoteIpBlocks`:
```
yangminzhu, line 289: same here for tab.

content/en/docs/tasks/security/authorization/authz-ingress/test.sh
yangminzhu: I think as long as the original ipBlocks field is covered, we're mostly good here as it's just about using two fields in the API, the doc structure, steps and commands are mostly the same (except the environment setup that is kind out of the scope and hard to reproduce in test).

And we also have e2e tests covering both fields in the istio/istio repo.

yangminzhu (commented):
Thanks for the PR, the doc update looks great! just some small comments.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
frankbu: This one is better left as a numbered list, IMO.

kylebevans: I made it a tab before I saw Frank's comment.  Let me know whether you like it this way or the other way.  I can change it back if need be.

content/en/docs/tasks/security/authorization/authz-ingress/test.sh
kylebevans: Breaking things up into tabs made it easier to add verifications, so I added some more.  I got the client ip from the envoy debug logs, which feels a little bit like gaming it, but it does work.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -12,7 +12,7 @@ using an authorization policy.
 
 An Istio authorization policy supports IP-based allow lists or deny lists as well as
 the attribute-based allow lists or deny lists previously provided by Mixer policy.
-The Mixer policy is deprecated in 1.5 and not recommended for production use.
+The Mixer policy is deprecated in *5 and not recommended for production use.
```
frankbu, line 15: What is `*5`? Maybe:
```suggestion
Mixer policy has been deprecated since Istio 1.5 and is not recommended for production use.
```
Or is not even available in 1.8? If so, Maybe remove the sentence completely?

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +31,388 @@ through the Istio ingress gateway with this command:
...
+
+* A `LoadBalancer` is just like a `NodePort`, except it also creates an environment specific external load balancer to handle distributing traffic to the worker nodes.  For example, in AWS EKS, the `LoadBalancer` service will create a Classic ELB with your worker nodes as targets. If your Kubernetes environment does not have a `LoadBalancer` implementation, then it will just behave like a `NodePort`.  An Istio ingress gateway creates a `LoadBalancer` service.
+
+* What if the Pod that is handling traffic from the `NodePort` or `LoadBalancer` isn't running on the worker node that received the traffic?  Kubernetes has its own internal proxy called kube-proxy that receives the packets and forwards them to the correct node.
+
+## Source IP address of the original client
+
+If a packet goes through an external proxy load balancer and/or kube-proxy, then the original source IP address of the client is lost.  Below are some strategies for preserving the original client IP for logging or security purposes.  For reference, here are the types of load balancers created by Istio with a `LoadBalancer` service on popular managed Kubernetes environments:
```
frankbu, line 64: I would move the `For reference, here ...` to after the tabset.
```
If a packet goes through an external proxy load balancer and/or kube-proxy, then the original source IP address of the client is lost. Here are some strategies for preserving the original client IP for logging or security purposes:

{{< tabset category-name="lb" >}}
...
{{< /tabset >}}

For reference, here are the types of load balancers created by Istio with a `LoadBalancer` service on popular managed Kubernetes environments:

|Cloud Provider | Load Balancer Name            | Load Balancer Type
----------------|-------------------------------|-------------------
|AWS EKS        | Classic Elastic Load Balancer | TCP Proxy
|GCP GKE        | TCP/UDP Network Load Balancer | Network
|Azure AKS      | Azure Load Balancer           | Network
|DO DOKS        | Load Balancer                 | Network

{{< tip >}}
You can instruct AWS EKS to create a Network Load Balancer when you install Istio by using a `serviceAnnotation` like below:

{{< text yaml >}}
kind: IstioOperator
spec:
  meshConfig:
    accessLogEncoding: JSON
    accessLogFile: /dev/stdout
  components:
    ingressGateways:
    - enabled: true
      k8s:
        hpaSpec:
          maxReplicas: 10
          minReplicas: 5
        serviceAnnotations:
          service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
{{< /text >}}

{{< /tip >}}
```

content/en/docs/tasks/security/authorization/authz-ingress/index.md
kylebevans: That was a search/replace error.  I'll try to find out if Mixer is gone or not.

kylebevans: mixer is gone, so i'll remove it.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
frankbu: Yeah, tabset looks fine to me, as long as you move the "for reference" part after it.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
     {{< /text >}}
 
-* See [Source IP for Services with `Type=NodePort`](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport)
-for more information. Update the ingress gateway to set `externalTrafficPolicy: local` to preserve the
-original client source IP on the ingress gateway using the following command:
+* Turn on RBAC debugging in Envoy for the ingress gateway:
 
     {{< text bash >}}
-    $ kubectl patch svc istio-ingressgateway -n istio-system -p '{"spec":{"externalTrafficPolicy":"Local"}}'
+    $ kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do istioctl proxy-config log "$pod" -n istio-system --level rbac:debug; done
```
frankbu, line 32: This seems like a pretty crazy way to access the ingress gateway. Why not something like this?
```suggestion
    $ istioctl proxy-config log $(kubectl get pod -l istio=ingressgateway -n istio-system) -n istio-system --level rbac:debug
```

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
     {{< /text >}}
 
-*  Follow the instructions in
-    [Determining the ingress IP and ports](/docs/tasks/traffic-management/ingress/ingress-control/#determining-the-ingress-ip-and-ports)
-    to define the `INGRESS_HOST` and `INGRESS_PORT` environment variables.
-
```
frankbu, line 45: You can't delete this, because you're using the env vars in the next step.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
 * Verify that the `httpbin` workload and ingress gateway are working as expected using this command:
 
     {{< text bash >}}
     $ curl "$INGRESS_HOST":"$INGRESS_PORT"/headers -s -o /dev/null -w "%{http_code}\n"
```
frankbu, line 38: A best practice in bash is to put quotes around the whole arg, not each variable:
```suggestion
    $ curl "$INGRESS_HOST:$INGRESS_PORT/headers" -s -o /dev/null -w "%{http_code}\n"
```

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
-    {{< /text >}}
-
 {{< warning >}}
 If you don’t see the expected output, retry after a few seconds.
 Caching and propagation overhead can cause a delay.
 {{< /warning >}}
 
+## Getting traffic into Kubernetes and Istio
+
+* All methods of getting traffic into Kubernetes involve opening a port on all worker nodes.  The main features that accomplish this are the `NodePort` service and the `LoadBalancer` service.  Even the Kubernetes `Ingress` resource must be backed by an Ingress controller that will create either a `NodePort` or a `LoadBalancer` service.
```
frankbu, line 49: This doesn't sound like it should be a bullet
```suggestion
All methods of getting traffic into Kubernetes involve opening a port on all worker nodes.  The main features that accomplish this are the `NodePort` service and the `LoadBalancer` service.  Even the Kubernetes `Ingress` resource must be backed by an Ingress controller that will create either a `NodePort` or a `LoadBalancer` service.
```

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
+
+* A `NodePort` just opens up a port in the range 30000-32767 on each worker node and uses a label selector to identify which Pods to send the traffic to.  You have to manually create some kind of load balancer in front of your worker nodes or use Round-Robin DNS.
+
+* A `LoadBalancer` is just like a `NodePort`, except it also creates an environment specific external load balancer to handle distributing traffic to the worker nodes.  For example, in AWS EKS, the `LoadBalancer` service will create a Classic ELB with your worker nodes as targets. If your Kubernetes environment does not have a `LoadBalancer` implementation, then it will just behave like a `NodePort`.  An Istio ingress gateway creates a `LoadBalancer` service.
+
+* What if the Pod that is handling traffic from the `NodePort` or `LoadBalancer` isn't running on the worker node that received the traffic?  Kubernetes has its own internal proxy called kube-proxy that receives the packets and forwards them to the correct node.
```
frankbu, line 55:
```suggestion
What if the Pod that is handling traffic from the `NodePort` or `LoadBalancer` isn't running on the worker node that received the traffic?  Kubernetes has its own internal proxy called kube-proxy that receives the packets and forwards them to the correct node.
```

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
+        serviceAnnotations:
+          service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
+{{< /text >}}
+
+{{< /tip >}}
+
 ## IP-based allow list and deny list
 
-1. The following command creates the authorization policy, `ingress-policy`, for
+**`ipBlocks` vs `remoteIpBlocks`:** If you are using the X-Forwarded-For HTTP header or the Proxy Protocol to determine the original client IP address, then you should use `remoteIpBlocks` in your `AuthorizationPolicy`. If you are using `externalTrafficPolicy: Local`, then you should use `ipBlocks` in your `AuthorizationPolicy`.
```
frankbu, line 188: In the `TCP/UDP Proxy Load Balancer` tab in the previous section, neither `X-Forwarded-For` nor `externalTrafficPolicy` is mentioned. Which one (ipBlocks or remoteIpBlocks) to use in that case?

This paragraph should also reference back to the previous section to make it clear that this decision is related to that information.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< /tabset >}}
+
+* Verify that a request to the ingress gateway is denied:
 
     {{< text bash >}}
     $ curl "$INGRESS_HOST":"$INGRESS_PORT"/headers -s -o /dev/null -w "%{http_code}\n"
```
frankbu, line 390:
```suggestion
    $ curl "$INGRESS_HOST:$INGRESS_PORT/headers" -s -o /dev/null -w "%{http_code}\n"
```

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< /tabset >}}
+
+* Verify that a request to the ingress gateway is allowed:
 
     {{< text bash >}}
     $ curl "$INGRESS_HOST":"$INGRESS_PORT"/headers -s -o /dev/null -w "%{http_code}\n"
```
frankbu, line 329:
```suggestion
    $ curl "$INGRESS_HOST:$INGRESS_PORT/headers" -s -o /dev/null -w "%{http_code}\n"
```

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
+{{< /tip >}}
+
 ## IP-based allow list and deny list
 
-1. The following command creates the authorization policy, `ingress-policy`, for
+**`ipBlocks` vs `remoteIpBlocks`:** If you are using the X-Forwarded-For HTTP header or the Proxy Protocol to determine the original client IP address, then you should use `remoteIpBlocks` in your `AuthorizationPolicy`. If you are using `externalTrafficPolicy: Local`, then you should use `ipBlocks` in your `AuthorizationPolicy`.
+
+*  Follow the instructions in
+    [Determining the ingress IP and ports](/docs/tasks/traffic-management/ingress/ingress-control/#determining-the-ingress-ip-and-ports)
+    to define the `INGRESS_HOST` and `INGRESS_PORT` environment variables.
```
frankbu, line 192: This needs to have been done already in "Before you begin".
```suggestion
```

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
+* Update the `ingress-policy` to include your client IP address:
+
+{{< tabset category-name="source" >}}
+
+{{< tab name="ipBlocks" category-value="ipBlocks" >}}
+
+Find your original client IP address if you don't know it and assign it to a variable:
+
+{{< text bash >}}
+$ CLIENT_IP=$(kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do kubectl logs "$pod" -n istio-system | grep remoteIP; done | head -1 | awk -F, '{print $3}' | awk -F: '{print $2}' | sed 's/ //') && echo "$CLIENT_IP"
```
frankbu, line 267: Same question as https://github.com/istio/istio.io/pull/8390/files#r516295115

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="remoteIpBlocks" category-value="remoteIpBlocks" >}}
+
+Find your original client IP address if you don't know it and assign it to a variable:
+
+{{< text bash >}}
+$ CLIENT_IP=$(kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do kubectl logs "$pod" -n istio-system | grep remoteIP; done | head -1 | awk -F, '{print $4}' | awk -F: '{print $2}' | sed 's/ //') && echo "$CLIENT_IP"
```
frankbu, line 297: https://github.com/istio/istio.io/pull/8390/files#r516295115

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -31,143 +26,390 @@ through the Istio ingress gateway with this command:
...
     {{< /text >}}
 
-1. You could use an online proxy service to access the ingress gateway using a
+* You could use an online proxy service to access the ingress gateway using a
 different client IP to verify the request is allowed.
 
+* If you are not getting the responses you expect, view the ingress gateway logs which should show RBAC debugging information:
+
+    {{< text bash >}}
+    $ kubectl get pods -n istio-system | grep ingress | awk '{print $1}' | while read -r pod; do kubectl logs "$pod" -n istio-system; done
```
frankbu, line 400: https://github.com/istio/istio.io/pull/8390/files#r516295115

content/en/docs/tasks/security/authorization/authz-ingress/index.md
kylebevans: You have to run istioctl proxy-config for each ingress gateway pod, so you can't escape doing a loop if there is more than one ingress gateway.  Also, you need to grab just the pod name output from get pods with cut or awk. I used awk.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
kylebevans: Ah, I'll move that back up.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
kylebevans: Same reasoning here; we can't be sure which ingress gateway got the request in the case of multiple ingress gateway pods.  We need to loop through them all and print the logs to find the request.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
kylebevans: I added a table here to make that more clear.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
kylebevans: I made the other changes, and the remaining comments are related to this kind of loop.  Let me know if my answer is acceptable here, or if you still want me to rework it.

frankbu: I see. The grep and awk still seem a little ugly. Would something like this work?
```
kubectl get pods -n istio-system -l istio=ingressgateway -o name | while ...
```

kylebevans: That is close, but it prints the pods in this format `pod/istio-ingressgateway-99b56479c-46zt2` and istioctl doesn't accept the `pod/` part.  We could get rid of the grep that way, but the awk (or sed) will end up being a little more complex since a substitution will be necessary to remove `pod/`.  What do you think?

frankbu: I don't know. Does this look any better?
```
kubectl get pods -n istio-system -l istio=ingressgateway -o name | sed 's|pod/||' | while ...
```
I'll leave it up to you.

kylebevans: Sure, that seems reasonable.

frankbu (approved):

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -46,128 +39,383 @@ original client source IP on the ingress gateway using the following command:
...
+  workloadSelector:
+    labels:
+      istio: ingressgateway
+  configPatches:
+  - applyTo: LISTENER
+    patch:
+      operation: MERGE
+      value:
+        listener_filters:
+        - name: envoy.listener.proxy_protocol
```
Kyslik, line 87:
```
        - name: envoy.listener.tls_inspector
```

Is not needed?

kylebevans: It's not needed just to enable the proxy protocol.  If you already have it in the config, the EnvoyFilter above won't remove it.  I'm not sure whether they are compatible or not though.

Kyslik: Without `envoy.listener.tls_inspector` `https://` does not work; maybe add it as note? Also its mentioned in here https://github.com/istio/istio.io/blob/master/content/en/docs/ops/configuration/traffic-management/network-topologies/index.md#proxy-protocol and in here https://github.com/istio/istio/pull/12748#issuecomment-640274792.

kylebevans: I didn't notice they had also added Proxy protocol documentation to the Network Topologies doc, but you're right the two EnvoyFilters there and here should probably match.

kylebevans: FYI, PR here: https://github.com/istio/istio.io/pull/8472

Kyslik: Perfect; thank you. Much appreciated.

PR #8392, opened 2020-10-29
Fix invalid operator sample

nak3:
https://istio.io/latest/docs/setup/install/istioctl explains that the `pilot-k8s.yaml` exists in `samples/operator/`,
but it does not. It should be [operator/samples/pilot-k8s.yaml](https://github.com/istio/istio/blob/master/operator/samples/pilot-k8s.yaml).

[X] Docs


shamsher31 (approved):
Thanks.

frankbu (approved):

ericvn (approved):

PR #8301, opened 2020-10-14
Global trustDomain setting is deprecated in favour of meshConfig.trustDomain

shamsher31:
Part of https://github.com/istio/istio/issues/27734

Related: https://github.com/istio/istio/pull/27806

PR #8399, opened 2020-10-29
Update test reference to 1.8.0-alpha.2

### LINKED ISSUE #8404: docs/tasks/observability/logs/access-log has changed output with 1.8.0-alpha.2
[2020-10-29T21:51:57.945Z] "GET /status/418 HTTP/1.1" 418 - "-" 0 135 2 2 "-" "curl/7.69.1" "f65c68d3-619a-44c7-a9c1-44384b3055ba" "httpbin:8000" "127.0.0.1:80" inbound|8000|| 127.0.0.1:52328 10.244.0.14:80 10.244.0.13:43430 outbound_.8000_._.httpbin.default.svc.cluster.local default", expected: "inbound|8000|http|httpbin.default.svc.cluster.local"

From Slack, John states that:
`inbound|<port>|<name>|<hostname> -> inbound|<port>|| is the expectation`

ericvn:
Update the test reference which lead to several breakages.

Fixes #8404


ericvn:
Noted that `tasks/observability/logs/access-log` is failing again, so looking at that test locally.

ericvn:
tasks/security/authorization/authz-td-migration fails and I opened https://github.com/istio/istio.io/issues/8405. I also asked about the 503 vs 403 in Slack: https://istio.slack.com/archives/C3TEGNZ7W/p1604061921132200

content/en/docs/tasks/observability/logs/access-log/test.sh
```diff
@@ -46,7 +46,8 @@ _verify_elided snip_test_the_access_log_1 "$snip_test_the_access_log_1_out"
 
 # Check the logs
 _verify_contains snip_test_the_access_log_2 "outbound|8000||httpbin.default.svc.cluster.local"
-_verify_contains snip_test_the_access_log_3 "inbound|8000|http|httpbin.default.svc.cluster.local"
+_verify_contains snip_test_the_access_log_3 "inbound|8000||"
+_verify_contains snip_test_the_access_log_3 ".httpbin.default.svc.cluster.local"
```
frankbu, line 50: Can the check be more specific?
```suggestion
_verify_contains snip_test_the_access_log_3 "outbound_.8000_._.httpbin.default.svc.cluster.local"
```

content/en/docs/tasks/security/authorization/authz-td-migration/index.md
```diff
@@ -22,7 +22,7 @@ In Istio 1.4, we introduce an alpha feature to support {{< gloss >}}trust domain
 1. Install Istio with a custom trust domain and mutual TLS enabled.
 
     {{< text bash >}}
-    $ istioctl install --set profile=demo --set values.global.trustDomain=old-td
+    $ istioctl install --set profile=demo --set meshConfig.trustDomain=old-td
```
frankbu, line 25: There's another PR (#8301) that fixes this plus an IstioOperator example as well. It shouldn't be needed here, though, because the old helm API should also work for backward compatibility, right?

content/en/docs/tasks/observability/logs/access-log/test.sh
ericvn: What I did was separate the prior check into two checks for the pieces that surround the new data, so we are checking for the same things as before.  If we had a regex check we could have the old check and say to ignore all the characters in the middle of the string.

content/en/docs/tasks/security/authorization/authz-td-migration/index.md
ericvn: Agree that his change isn't needed here. Leaving it the old ways, generates a warning message in the test output:
```
! values.global.trustDomain is deprecated; use meshConfig.trustDomain instead
```

Making this change just removes the warning. And since the value is deprecated, we are ready for the future.

content/en/docs/tasks/observability/logs/access-log/test.sh
frankbu: If you want to check two things, you should use _verify_lines instead of two _verify_contains calls (you don't want to call the snip twice).
```
_verify_lines snip_test_the_access_log_3 "
+ inbound|8000||
+ .httpbin.default.svc.cluster.local"
```
That said, I think my suggested check (above) is a better and simpler check that the request was logged.

ericvn: I updated to use the single check you mentioned above.

frankbu (approved):

PR #8401, opened 2020-10-29
chore: add meshery in benchmarking tool

kushthedude:
[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

- This PR lists Meshery, the service mesh performance, and management plane, into the available benchmarking tools.


ericvn:
@craigbox OK to add this?

craigbox:
I don't think this is a list of possible benchmarking tools one could use, but instead a list of what Istio currently uses to generate the numbers and graphs it publishes.  With that in mind, I think we should close the PR without merging.

@kushthedude: you might like to write a blog post on how you can use Meshery to do performance tests on Istio?

ericvn:
Pre @craigbox, I'll close this PR and wait for a possible blog entry.

PR #8432, opened 2020-11-02
Update operator upgrade docs

ostromart:


content/en/docs/setup/install/operator/index.md
```diff
@@ -172,9 +172,32 @@ $ kubectl logs -f -n istio-operator $(kubectl get pods -n istio-operator -lname=
...
+## In-place Upgrade
+
+Download and extract the `istioctl` corresponding to the version of Istio you wish to upgrade to. Reinstall the operator
+at the target Istio version:
+
+{{< text bash >}}
+$ <extracted-dir>/bin/istioctl operator init
+{{< /text >}}
+
+You should see that the `istiod` pod has restarted and its version has changed to the target version:
```
richardwxn, line 184: `istio-operator` pod not `istiod`

content/en/docs/setup/install/operator/index.md
```diff
@@ -172,9 +172,32 @@ $ kubectl logs -f -n istio-operator $(kubectl get pods -n istio-operator -lname=
 Refer to the [`IstioOperator` API](/docs/reference/config/istio.operator.v1alpha1/#IstioOperatorSpec)
 for the complete set of configuration settings.
 
+## In-place Upgrade
+
+Download and extract the `istioctl` corresponding to the version of Istio you wish to upgrade to. Reinstall the operator
+at the target Istio version:
+
+{{< text bash >}}
+$ <extracted-dir>/bin/istioctl operator init
```
richardwxn, line 181: maybe worth mentioning a bit about prerequisite? Such as the existence of the IstioOperator CR, and revision need to match between the new deployment and the existing CR(either both no revision or revision should be the same).

btw I am still worried about the installed-state CR, which should be used for recording purpose only while it is possible to be reconciled by any of the operator pods.

content/en/docs/setup/install/operator/index.md
ostromart: done.

content/en/docs/setup/install/operator/index.md
ostromart: i think it's a given that there's a CR since this is an upgrade?
the revision matching is important but not for in-place. i think we can probably assume that if they are using in-place, they are not interested in revisions, so adding it here may lead to confusion.

frankbu (approved):

Issue #8396, opened 2020-10-29
Need to document the new vendor profiles


ericvn:
https://preliminary.istio.io/latest/docs/setup/additional-setup/config-profiles/ does not contain any information about the new vendor profiles like `openshift`

PR #8424, opened 2020-11-02
add empty and preview profile to the table

shamsher31:
Before:

before

After:

after

PR #8337, opened 2020-10-22
Update the docs to add OpenShift install profile

hickeyma:
OpenShift install profile was added in https://github.com/istio/istio/pull/28076. This PR is to update the docs to inform users about the profile.

[ ] Configuration Infrastructure
[x] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #8434, opened 2020-11-03
Add tip for vendor profiles

shamsher31:
To fix https://github.com/istio/istio.io/issues/8396

Based on https://github.com/istio/istio.io/pull/8424#issuecomment-720494703

Ref: https://preliminary.istio.io/latest/docs/setup/additional-setup/config-profiles/

### DIFF
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -54,3 +54,8 @@ The components marked as **X** are installed within each profile:
 To further [customize Istio and install addons](/docs/setup/install/istioctl/#customizing-the-configuration), you can add one or more `--set <key>=<value>` options in the
 `istioctl install` command that you use when installing Istio.
 Refer to [integration](/docs/ops/integrations) for more details.
+
+{{< tip >}}
+Some additional vendor-specific configuration profiles are also available.
+For a complete list, refer to [Vendor Optimized Profiles]({{< github_file >}}/manifests/profiles/PROFILES.md).
+{{< /tip >}}
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -42,6 +42,11 @@ for your specific needs. The following built-in configuration profiles are curre
 1. **preview**: the preview profile contains features that are experimental. This is intended to explore new features
                 coming to Istio. Stability, security, and performance are not guaranteed - use at your own risk.
 
+{{< tip >}}
+Some additional vendor-specific configuration profiles are also available.
+For a complete list, refer to [Vendor Optimized Profiles]({{< github_file >}}/manifests/profiles/PROFILES.md).
+{{< /tip >}}
+
 The components marked as **X** are installed within each profile:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -42,14 +42,18 @@ for your specific needs. The following built-in configuration profiles are curre
 1. **preview**: the preview profile contains features that are experimental. This is intended to explore new features
                 coming to Istio. Stability, security, and performance are not guaranteed - use at your own risk.
 
+1. **openshift**: used to install Istio on OpenShift platform. Note that the OpenShift is incompatible with the Istio
```
ericvn, line 45: I believe we need to add a section called `Vendor Optimized Profiles` to include profiles like this.

See https://github.com/istio/istio/blob/master/manifests/profiles/PROFILES.md. The reason is that this profiles are supported by the vendor and not necessarily the Istio community so could see one of them going away at some point and it might not be handled in the normal community deprecation?

ericvn:
Also, see https://github.com/istio/istio.io/pull/8337

content/en/docs/setup/additional-setup/config-profiles/index.md
shamsher31: Do you mean to add a separate table for Vendor profile?
Is it ok to have `opnshift` description here or it also needs to be moved below the table?

frankbu: I don't think we want anything vendor specific in these docs. Maybe add a tip here, instead of the profile:
```
{{< tip >}}
Some additional vendor-specific configuration profiles are also available.
For a complete list, refer to [Vendor Optimized Profiles]({{< github_file >}}/manifests/profiles/PROFILES.md).
{{< /tip >}}
```

shamsher31: So that means, nothing needs to be added other than just a `<tip>`?

shamsher31: Done.

content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -54,3 +54,8 @@ The components marked as **X** are installed within each profile:
 To further [customize Istio and install addons](/docs/setup/install/istioctl/#customizing-the-configuration), you can add one or more `--set <key>=<value>` options in the
 `istioctl install` command that you use when installing Istio.
 Refer to [integration](/docs/ops/integrations) for more details.
+
+{{< tip >}}
+Some additional vendor-specific configuration profiles are also available.
+For a complete list, refer to [Vendor Optimized Profiles]({{< github_file >}}/manifests/profiles/PROFILES.md).
+{{< /tip >}}
```
frankbu, line 61: Maybe this tip could be a little more prominently displayed?  How about right after the list of standard profiles (at line 44)?

shamsher31: Sounds good, moved.

frankbu (approved):

ericvn (approved):

PR #8456, opened 2020-11-06
Fix profile warning in getting started doc

frankbu:
[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #8464, opened 2020-11-09
update warn for custom profile

shamsher31:
Based on https://github.com/istio/istio.io/pull/8337#discussion_r518228960


frankbu:
Fixed in #8456

PR #8466, opened 2020-11-09
Document vendor profiles

brian-avery:
This moves the vendor profile documentation to istio.io rather than /manifests/profiles/PROFILES.md

### DIFF
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -44,7 +44,7 @@ for your specific needs. The following built-in configuration profiles are curre
 
 {{< tip >}}
 Some additional vendor-specific configuration profiles are also available.
-For a complete list, refer to [Vendor Optimized Profiles]({{< github_blob >}}/manifests/profiles/PROFILES.md).
+For more information, refer to the [setup instructions](/docs/setup/platform-setup) for your platform.
 {{< /tip >}}
 
 The components marked as **X** are installed within each profile:


brian-avery:
Related to https://github.com/istio/istio/pull/28589

content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -56,6 +52,13 @@ The components marked as **X** are installed within each profile:
 | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`istio-ingressgateway` | X | X | | | | X |
 | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`istiod` | X | X | X | | | X |
 
+
+### Vendor Profiles
+
+Some additional vendor-specific configuration profiles are also available.
+
+1. **openshift**: used for configuring Istio in an OpenShift environment.
```
brian-avery, line 60: @sdake @jwendell The yaml of the openshift profile is quite a bit different than any of the other profiles. Should we do a table of features enabled for it?

jwendell: IMO, not. Those are details that profiles intend to hide.

jwendell, line 60: s/OpenShift/Red Hat OpenShift/ ?

frankbu:
Quite some time ago, we decided to minimize platform specific information on istio.io. We prefer, instead, to add tips, etc., that point off to documentation on the vendor site. In this case, however, I think it could be documented on the platform setup page at istio.io, instead of some vendor site itself, but not in the mainline profiles doc.

brian-avery:
@frankbu Ok. Thanks. Are you suggesting a second profile page specifically to hold any vendor profiles?

content/en/docs/setup/additional-setup/config-profiles/index.md
brian-avery: So we hide all of the details of what features a user gets with a profile? Seems like it would be better to use one of the profiles that does document what it's doing instead. I don't think they can stack? Ideally I think we'd have like "install on OpenShift" or "Install on Anthos" with x features.

frankbu:
> @frankbu Ok. Thanks. Are you suggesting a second profile page specifically to hold any vendor profiles?

I think it depends on whether we really need a central list of these profiles or not. I was thinking that the `openshift` profile would be described on the [OpenShift setup page](https://preliminary.istio.io/latest/docs/setup/platform-setup/openshift/), same with any other ones, but I don't think there currently are any others. If we decide that listing of all the vendor profiles is needed, I would put it on the [top-level setup page](https://preliminary.istio.io/latest/docs/setup/platform-setup/openshift/), but I'm not sure that's needed.

frankbu:
Whether or not we have a full listing of vendor profiles anywhere, maybe we should change line 47 from this:
```
For a complete list, refer to [Vendor Optimized Profiles]({{< github_blob >}}/manifests/profiles/PROFILES.md).
```
to something like this:
```
For more information, refer to the [setup instructions](/docs/setup/platform-setup/) for your platform.
```

brian-avery:
Thanks. Currently, OpenShift is the only other documented platform that we have. Putting that content directly in the OpenShift platform page and documenting it sounds great to me as well as the text change that you suggested above.

frankbu (approved):

PR #8472, opened 2020-11-09
Make proxy protocol EnvoyFilter consistent with Network Topology doc

kylebevans:
Make the proxy protocol EnvoyFilter identical to the one in
docs/ops/configuration/traffic-management/network-topologies/

[ x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/bin/init.sh
+++ b/bin/init.sh
@@ -51,7 +51,7 @@ git checkout "$ISTIO_SHA"
 ISTIOCTL_ARTIFACT="${ISTIO_OUT}/release/"
 case "$GOOS_LOCAL" in
   linux)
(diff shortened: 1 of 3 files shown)


ericvn:
Just wondering how you decided which way to move the change? A quick look seems to show a lot of other pages having `workloadSelector` toward to top of the yaml file. That said, I do see a few lower in the file.

frankbu (approved):

kylebevans:
I have made edits to the authz-ingress task before, but not the network topology doc so I just deferred to the style there.  I personally like the `workloadSelector` near the top.  I can change them both if you want.

PR #8482, opened 2020-11-10
Add in removed tests (had regression failure) to verify fix

ericvn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
Verified the regression still exists:
```
VERIFY FAILED snip_migrate_trust_domain_without_trust_domain_aliases_4 (timeout after 120s): received: "503", expected: "403"
```

ericvn:
I worked with @howardjohn on the problem, and it was determined that `istiod` is not being redeployed on an `istioctl install` as it was in the past.

I reworked the doc/test to add a redeployment of istiod. Interestingly, we had a statement in the original test case to wait on the `istiod` redeployment to finish after the istioctl install.

The question remains if this is a breaking change from before. Older versions automatically restarted `istiod` so the additional step wasn't needed. This is also a test of an un-preferred path.

howardjohn (approved):

frankbu (approved):

PR #8491, opened 2020-11-12
Few fixes for ext control plane setup

frankbu:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/additional-setup/external-controlplane/index.md
+++ b/content/en/docs/setup/additional-setup/external-controlplane/index.md
@@ -56,7 +56,7 @@ Variable | Description
 -------- | -----------
 `CTX_EXTERNAL_CLUSTER` | The context name in the default [Kubernetes configuration file](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/) used for accessing the external control plane cluster.
 `CTX_REMOTE_CLUSTER` | The context name in the default [Kubernetes configuration file](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/) used for accessing the remote cluster.
-`EXTERNAL_ISTIOD_ADDR` | The hostname for the ingress gateway on the external control plane cluster. This is used by the `remote_cluster` to access the external control plane.
+`EXTERNAL_ISTIOD_ADDR` | The hostname for the ingress gateway on the external control plane cluster. This is used by the remote cluster to access the external control plane.
 `SSL_SECRET_NAME` | The name of the secret that holds the TLS certs for the ingress gateway on the external control plane cluster.
 
 Set the `CTX_EXTERNAL_CLUSTER` and `CTX_REMOTE_CLUSTER` now. You will set the others later.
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -233,7 +233,7 @@ spec:
       istioNamespace: external-istiod
       meshID: mesh1
       multiCluster:
-        clusterName: remote_cluster
+        clusterName: $CTX_REMOTE_CLUSTER
```
linsun, line 236: Think this won't work because it is the context, not the cluster name.  If you want cluster name as var, user would need to set them as env var first.

linsun:
Other changes LGTM

content/en/docs/setup/additional-setup/external-controlplane/index.md
frankbu: Good point. Added another env var. PTAL

linsun (approved):
LG!  Thanks @frankbu

content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -56,14 +56,16 @@ Variable | Description
...
+`EXTERNAL_ISTIOD_ADDR` | The hostname for the ingress gateway on the external control plane cluster. This is used by the remote cluster to access the external control plane.
 `SSL_SECRET_NAME` | The name of the secret that holds the TLS certs for the ingress gateway on the external control plane cluster.
 
-Set the `CTX_EXTERNAL_CLUSTER` and `CTX_REMOTE_CLUSTER` now. You will set the others later.
+Set the `CTX_EXTERNAL_CLUSTER`, `CTX_REMOTE_CLUSTER`, and `REMOTE_CLUSTER_NAME` now. You will set the others later.
 
 {{< text bash >}}
 $ export CTX_EXTERNAL_CLUSTER=external_cluster
 $ export CTX_REMOTE_CLUSTER=remote_cluster
+$ export REMOTE_CLUSTER_NAME=remote_cluster
```
ericvn, line 68: minor nit: Since you have `remote_cluster` for both items, should you make them `external_/remote_cluster_ctx` and remote_cluster_name?

frankbu: The `CTX_` names are a consistent naming convention used in the multicluster docs, so we should leave them that way. Doing the opposite, `NAME_REMOTE_CLUSTER`, sounds awkward. So, I'd leave them as they are.

ericvn: My comment was more of the values on the right side of the `=` sign. You have remote_cluster twice for different variables. You probably want to highlight to the reader that these values can be different.

frankbu: Oh, I see. I think it's fine. Usually the cluster name and context name are the same, but not always. That's why we need two variables. Users should be able to figure it out, given the table, above, with clear descriptions of all the vars.

ericvn (approved):

PR #8509, opened 2020-11-16
Small update to clarify the mirror flow

sb1975:
Small clarification - This route rule sends 100% of the traffic from `v1` to `v2`

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/mirroring/index.md
+++ b/content/en/docs/tasks/traffic-management/mirroring/index.md
@@ -235,7 +235,7 @@ log entries for `v1` and none for `v2`:
     EOF
     {{< /text >}}
 
-    This route rule sends 100% of the traffic to `v1`. The last stanza specifies
+    This route rule sends 100% of the traffic from `v1` to `v2`. The last stanza specifies
     that you want to mirror to the `httpbin:v2` service. When traffic gets mirrored,
     the requests are sent to the mirrored service with their Host/Authority headers
     appended with `-shadow`. For example, `cluster-1` becomes `cluster-1-shadow`.
--- a/content/en/docs/tasks/traffic-management/mirroring/index.md
+++ b/content/en/docs/tasks/traffic-management/mirroring/index.md
@@ -235,8 +235,8 @@ log entries for `v1` and none for `v2`:
     EOF
     {{< /text >}}
 
-    This route rule sends 100% of the traffic from `v1` to `v2`. The last stanza specifies
-    that you want to mirror to the `httpbin:v2` service. When traffic gets mirrored,
+    This route rule sends 100% of the traffic to `v1`. The last stanza specifies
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/traffic-management/mirroring/index.md
```diff
@@ -235,7 +235,7 @@ log entries for `v1` and none for `v2`:
     EOF
     {{< /text >}}
 
-    This route rule sends 100% of the traffic to `v1`. The last stanza specifies
+    This route rule sends 100% of the traffic from `v1` to `v2`. The last stanza specifies
```
frankbu, line 238: This change is not correct. As documented, the route rule is sending 100% of the traffic to the `v1` version of httpbin. See the fields `weight: 100` and `subset: v1`. In addition it's also mirroring 100% of the traffic to `v2`. That's not the same as sending the v1 traffic to v2.

sb1975 (commented):
updated to clarify

content/en/docs/tasks/traffic-management/mirroring/index.md
```diff
@@ -236,7 +236,7 @@ log entries for `v1` and none for `v2`:
     {{< /text >}}
 
     This route rule sends 100% of the traffic to `v1`. The last stanza specifies
-    that you want to mirror to the `httpbin:v2` service. When traffic gets mirrored,
+    that you want to mirror 100% of traffic from `httpbin:v1` to the `httpbin:v2` service. When traffic gets mirrored,
```
frankbu, line 239: This is better, but still not exactly correct. It's not mirroring traffic "from" `httpbin:v1`. It's mirroring the traffic that is being sent "to" `httpbin:v1`. More precisely, it's mirroring "all the incoming traffic", regardless of where's it's actually being sent. You could, for example, set up a rule to send 50% of traffic to v1, 50% to v2, and mirror 100% to v3. The mirror means what percentage of incoming traffic should go to a mirror destination. How about clarifying like this:
```suggestion
    that you want to mirror (i.e., also send) 100% of the same traffic to the
    `httpbin:v2` service. When traffic gets mirrored,
```

frankbu:
@sb1975 would you like to update this PR as suggested in https://github.com/istio/istio.io/pull/8509#discussion_r526160596, and then sign the CLA so we can merge it? Thanks.

frankbu:
@sb1975 You still need to sign the CLA before we can merge your PR. Please see https://github.com/istio/istio.io/pull/8509#issuecomment-731922208.

frankbu (approved):

sb1975:
@googlebot I signed it!

sb1975:
wondering why its failing still..

PR #8517, opened 2020-11-17
add verify traffic step for VM doc

stevenctl:
Adds a step to verify VM to Kubernetes traffic using in-agent DNS rather than just checking that the logs look ok. 

Other fixups include:
* fixing `--set` flag for auto-registration
* removing misplaced backtick from a command
* including RPM removal instruction for CentOS

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -271,7 +271,7 @@ Run the following commands on the virtual machine you want to add to the Istio m
     $ 2020-08-21T01:32:20.695478Z info cache GenerateSecret default
     $ 2020-08-21T01:32:20.695595Z info sds resource:default pushed key/cert pair to proxy
     {{< /text >}}
- 
+
 1. Create a Namespace to deploy a Pod-based Service:
 
     {{< text bash >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -271,6 +271,26 @@ Run the following commands on the virtual machine you want to add to the Istio m
...
+1. Deploy the `HelloWorld` Service:
+
+    {{< text bash >}}
+    $ kubectl apply -f @samples/helloworld/helloworld.yaml@
+    {{< /text >}}
+    
+1. Send requests from your Virtual Machine to the Service:
+
+    {{< text bash >}}
+    $ curl helloworld.sample.svc:5000/hello
```
stevenctl, line 291: Since this requires in-agent DNS, and https://github.com/istio/istio/pull/28966 is unlikely going to make it for 1.8.0, I think I need to add a step to enable it by hand.

ericvn (approved):

content/en/docs/setup/install/virtual-machine/index.md
frankbu: @stevenctl Now that this PR has merged, Is this doc broken in 1.8.0? If DNS won't be working, can you fix this up somehow so the instructions will work?

PR #8522, opened 2020-11-18
have user edit generated file to enable DNS

stevenctl:
Follow up to #8517 – until 1.8.1 users will need to do this step to enable DNS capturing on their VM.

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -145,12 +145,6 @@ Install Istio and expose the control plane so that your virtual machine can acce
        * `root-cert.pem`: The root certificate used to authenticate.
        * `hosts`: An addendum to `/etc/hosts` that the proxy will use to reach istiod for xDS.*
 
-    {{< idea >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -176,6 +170,20 @@ Install Istio and expose the control plane so that your virtual machine can acce
...
 
+1. (Optional) Enable Capturing DNS
+
+    {{< idea >}}
+    A sophisticated option involves configuring DNS within the virtual
+    machine to reference an external DNS server. This option is beyond
+    the scope of this guide.
+    {{< /idea >}}
+
+    To allow your VM to resolve Kubernetes Services by hostname, you can append a line to the generated `cluster.env`.
```
ericvn, line 181:
```suggestion
    To allow your VM to resolve Kubernetes Services by hostname, add an additional configuration option to the generated `cluster.env`.
```

stevenctl:
https://github.com/istio/istio/pull/28987 will actually be in 1.8.0 – this change isn't needed.

PR #8402, opened 2020-10-29
Rework locality LB docs into tasks

nmittler:
Removing the old locality docs entirely and replacing with tested tasks.

[ ] Configuration Infrastructure
[x] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #8524, opened 2020-11-18
Several test framework improvements

nmittler:
Overview of the changes:

- Adding ability to verify that expected output occurs a number of times consecutively. This is needed for https://github.com/istio/istio.io/pull/8402.

- Moving snapshot checking logic to Go code so that it can be separated out into separate test steps, which are timed and contain their own output directories/files. This makes the code cleaner and also makes the snapshot logic more transparent.

- Updating debug.sh to use newer bash syntax that allows it to dynamically select a free file descriptor. Without this, I was seeing all commands echoed to my console in goland.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/mtls_test.sh
+++ b/content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/mtls_test.sh
@@ -75,6 +75,7 @@ _verify_contains snip_configure_mutual_tls_origination_for_egress_traffic_5 "GET
 
 # @cleanup
 set +e # ignore cleanup errors
+kubectl label namespace default istio-injection- || true
 snip_cleanup_the_mutual_tls_origination_example_1
 snip_cleanup_the_mutual_tls_origination_example_2
 snip_cleanup_the_mutual_tls_origination_example_3
--- a/pkg/test/istioio/framework.go
+++ b/pkg/test/istioio/framework.go
@@ -48,21 +48,15 @@ var (
 	contentFolder = fmt.Sprintf("%v/%v/", os.Getenv("REPO_ROOT"), defaultPath)
 
 	// scripts that are sourced for all tests
-	helperTemplate = `
-		cd ${REPO_ROOT}
-		source "%v/%v" # snips.sh
-		source "tests/util/verify.sh"
-		source "tests/util/debug.sh"
-		source "tests/util/helpers.sh"
-	`
-
-	createClusterSnapshots = `
-		__create_cluster_snapshots
-	`
-
-	clusterCleanupCheck = `
-		__cluster_cleanup_check
-	`
+	scriptPrefixTemplate = `
+### BEGIN INJECTED SCRIPT ###
+cd ${REPO_ROOT}
+source "%v/%v" # snips.sh
+source "tests/util/verify.sh"
+source "tests/util/debug.sh"
+source "tests/util/helpers.sh"
+### END INJECTED SCRIPT ###
+`
 
 	snipsFileSuffix = "snips.sh"
 	testFileSuffix  = "test.sh"
@@ -139,7 +133,6 @@ func checkFile(path string) (*TestCase, error) {
 		)
 		return testCase, err
 	}
-	helperScript := getHelperScript(shortPath)
 	testScript := splitScript[0]
 	cleanupScript := splitScript[1]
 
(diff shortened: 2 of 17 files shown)


content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/mtls_test.sh
```diff
@@ -75,6 +75,7 @@ _verify_contains snip_configure_mutual_tls_origination_for_egress_traffic_5 "GET
 
 # @cleanup
 set +e # ignore cleanup errors
+kubectl label namespace default istio-injection- || true
```
frankbu, line 78: Is `|| true` needed even though it's running after `set +e`?

Actually, I wonder if the `set +e` should be automatically added by the frameworks, rather than first line of every test? WDYT?

pkg/test/istioio/framework.go
```diff
@@ -162,19 +155,48 @@ func checkFile(path string) (*TestCase, error) {
...
+	needToAdd := true
+
+	// Add the prefix before the first uncommented line.
+	for i, line := range strings.Split(script, "\n") {
+		// First check if the script doesn't start with the shell line.
+		// In that case, just start the output with a default shell line and
+		// the prefix.
+		if i == 0 && !strings.HasPrefix(line, "#!") {
+			//
+			out = fmt.Sprintf("#!/usr/bin/env bash\n%s\n%s", prefix, out)
```
frankbu, line 183: Why do we want to hide this special shebang injection here? Shouldn't it be required on line 1 of any script we call this function with, or otherwise not require it so we can ignore the issue?

pkg/test/istioio/snapshot.go
```diff
@@ -0,0 +1,286 @@
...
+					scopes.Framework.Debugf("failed listing pods in namespace %s: %v", namespace, err)
+				} else {
+					for _, pod := range pods.Items {
+						nsSnapshot.Pods = append(nsSnapshot.Pods, pod.Name)
+					}
+					sort.Strings(nsSnapshot.Pods)
+				}
+			}
+
+			if includeIstioResources {
```
frankbu, line 168: One of the problems we had before is that when starting with a clean cluster, the Istio CRDs are not defined, so listing these resources would fail. Does that problem go away with the Go functions that your calling now?

pkg/test/istioio/snapshot.go
```diff
@@ -0,0 +1,286 @@
...
+)
+
+var (
+	istioOperatorGVK = schema.GroupVersionResource{
+		Group:    "install.istio.io",
+		Version:  "v1alpha1",
+		Resource: "istiooperators",
+	}
+	namespaceBlacklist = map[string]bool{
+		"istio-system": true,
```
frankbu, line 43: General problem with the old snapshot implementation, and IIUC still with this impl, is that we really should snapshot the controlplane state. As is, we aren't detecting changes to the controlplane that might be left around.

nmittler: That's fine, but beyond the scope of this PR

pkg/test/istioio/snapshot.go
nmittler: No that wouldn't change with this implementation. Just maintaining existing functionality.

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/mtls_test.sh
nmittler: > Is || true needed even though it's running after set +e?

It's not, but it doesn't hurt

> Actually, I wonder if the set +e should be automatically added by the frameworks, rather than first line of every test? WDYT?

I'm probably fine with that, but would rather not change it in this PR.

pkg/test/istioio/framework.go
nmittler: done.

nmittler:
@frankbu PTAL

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/mtls_test.sh
frankbu: OK, I'll do that in a followup PR.

frankbu (approved):

PR #8527, opened 2020-11-19
fix older releases link on archive page

davidhauck:
Archives are now located at istio.io/archive instead of archive.istio.io.

[X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -23,6 +23,7 @@
 1.6.x.
 1.7.x
 1.7.x.
+1.8.x
 1.x
 10ms
 10s
@@ -111,6 +112,7 @@ backend
 backends
 base64
 Bazel
+Bian
 BigQuery
 bitpipe
 BlackHole
@@ -493,6 +495,7 @@ passthrough
 Passthrough
 PassthroughCluster
 peek
+Pengyuan
 pem
 performant
 pipelined
--- a/content/en/news/releases/1.8.x/_index.md
+++ b/content/en/news/releases/1.8.x/_index.md
@@ -1,0 +1,8 @@
+---
+title: 1.8.x Releases
+description: Announcements for the 1.8 release and its associated patch releases.
+weight: 21
+list_by_publishdate: true
+layout: release-grid
+decoration: dot
+---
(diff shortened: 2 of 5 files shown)


archive/archive/index.html
```diff
@@ -4,7 +4,7 @@
...
     <a tabindex=-1 role=menuitem href="https://github.com/istio/istio.io/issues/new?title=Issue%20with%20_index.html">Report Site Bugs</a><div role=separator></div><h6>Other versions of this site</h6><a tabindex=-1 role=menuitem onclick="navigateToUrlOrRoot('https://istio.io/\/');return false;">Current Release</a>
     <a tabindex=-1 role=menuitem onclick="navigateToUrlOrRoot('https://preliminary.istio.io/\/');return false;">Next Release</a>
-    <a tabindex=-1 role=menuitem href=https://archive.istio.io>Older Releases</a></div></div><button id=search-show title="Search this site" aria-label=Search><svg class="icon"><use xlink:href="./img/icons.svg#magnifier"/></svg></button></div><form id=search-form name=cse role=search><input type=hidden name=cx value=002184991200833970123:iwwf17ikgf4>
+    <a tabindex=-1 role=menuitem href=https://istio.io/archive>Older Releases</a></div></div><button id=search-show title="Search this site" aria-label=Search><svg class="icon"><use xlink:href="./img/icons.svg#magnifier"/></svg></button></div><form id=search-form name=cse role=search><input type=hidden name=cx value=002184991200833970123:iwwf17ikgf4>
```
frankbu, line 7: I think this is also broken on all the`archive/vX.Y/index.html` pages too.

davidhauck: Looks like >= V1.4 work correctly while <= V1.3 fail. It's going to take a while to go back and rebuild all of those site, so it may be easier to add a redirect from archive.istio.io to istio.io/archive

frankbu: Hmm, would it be fixed if you did a grep for `archive.istio.io` to find all the index.html files that are broken and make this same change to all of them? Or is there more to it than that?

PR #8608, opened 2020-12-01
Added Gateway creation to kubernetes-ingress page

s-bauer:
I think it's required to have a gateway listening in order for istio to serve requests to kubernetes ingress resources. Correct me if I'm wrong and/or missing something!

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/kubernetes-ingress/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/kubernetes-ingress/index.md
@@ -23,6 +23,28 @@ A [Kubernetes Ingress Resources](https://kubernetes.io/docs/concepts/services-ne
 
 Let's see how you can configure a `Ingress` on port 80 for HTTP traffic.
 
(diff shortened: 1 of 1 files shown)


s-bauer:
@googlebot I signed it!

howardjohn (commented):
This shouldn't be needed since Istio 1.6+

howardjohn:
Istiod will internally auto-generate such a gateway when you use Ingress. You can see the autogenerated Gateway by `curl localhost:8080/debug/configz` in istiod

s-bauer:
Running 1.7.3, let me verify it again!

s-bauer:
Yep verified it's not needed. Why is nothing working on my first attempt today. I really need to get some sleep :)
Anyway, thanks for your help, really appreciate it. I might still open another PR for the cert-manager integration as there could be more detail.

PR #8611, opened 2020-12-02
Use minimal profile instead of default

rytswd:
The default profile may be a good place to start with, but when the control plane should be handled separately from the gateway, minimal profile seems to better suit the gateway definition being in a separate setup.

As a side note, I find the use of `profile: minimal` is quite limited in current documentation. The minimal and empty profiles allow multiple IstioOperator CRs for managing the Istio installation (e.g. gateways and control planes to be separate in separate files).

(I'm even inclined to suggest default profile to be without ingress gateway, and some other profile to include both control plane and gateway - I understand this should be a separate discussion elsewhere.)

--

[x] Docs
[x] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -32,7 +32,8 @@ for your specific needs. The following built-in configuration profiles are curre
     This profile enables high levels of tracing and access logging so it is not suitable for performance tests.
     {{< /warning >}}
 
-1. **minimal**: the minimal set of components necessary to use Istio's [traffic management](/docs/tasks/traffic-management/) features.
+1. **minimal**: same as the default profile, but only the control plane components are installed.
+    This allows you to configure the control plane and data plane components (e.g., gateways) using [separate profiles](/docs/setup/upgrade/gateways/#installation-with-istioctl).
 
 1. **remote**: used for configuring {{< gloss "remote cluster" >}}remote clusters{{< /gloss >}} of a
     [multicluster mesh](/docs/ops/deployment/deployment-models/#multiple-clusters).
--- a/content/en/docs/setup/upgrade/gateways/index.md
+++ b/content/en/docs/setup/upgrade/gateways/index.md
@@ -43,12 +43,7 @@ the control plane.
     metadata:
       name: control-plane # REQUIRED
     spec:
-      profile: default
-      components:
-        ingressGateways:
-          - name: istio-ingressgateway
-            # MUST BE DISABLED
-            enabled: false
+      profile: minimal
     {{< /text >}}
 
 1.  Create a separate `IstioOperator` CR for the gateway(s), ensuring that it has a name and has the `empty` profile:


frankbu:
One question that this brings up is what exactly is the `minimal` profile these days? Is it the same as the `default` profile, but only the control plane part? If so, I think this PR make sense and we should also update the documentation [for the minimal profile](https://istio.io/latest/docs/setup/additional-setup/config-profiles/) to say something like: `minimal: same as the default profile, but only the control plane components are installed. This can be used to separate management of data plane components (e.g., gateways).`

If the `minimal` profile is not quite the same as `default w/o gateway`, maybe it would make sense to add another profile, e.g., `controlplane` which is == `default`, but control plane components only.

rytswd:
As to the diff, it looks as though `minimal` is really `default w/o ingress gateway`.

```bash
$ {
    cat << EOF > /tmp/istio-minimal.yaml
kind: IstioOperator
metadata:
  name: control-plane
spec:
  profile: minimal
EOF

    cat << EOF > /tmp/istio-default.yaml
kind: IstioOperator
metadata:
  name: control-plane
spec:
  profile: default
  components:
    ingressGateways:
        - name: istio-ingressgateway
          enabled: false
EOF

    istioctl manifest generate -f /tmp/istio-default.yaml > /tmp/istio-cp-default.yaml
    istioctl manifest generate -f /tmp/istio-minimal.yaml > /tmp/istio-cp-minimal.yaml

    diff /tmp/istio-cp-default.yaml /tmp/istio-cp-minimal.yaml && echo "no diff found"
}

no diff found

$
```

I think the word "minimal" is pretty vague. I was actually about to propose a change in the profile doc to clarify this installs control plane components such as istiod. I just wasn't certain how prevalent the word "control plane", though.

ostromart:
I think the minimal profile use case is too vague atm to start putting it into docs. @costinm or @linsun may have more context and know the current status quo.

linsun:
Should we rename min profile to control plane profile given we are now recommending users to install cp separately from data plane?  min profile isn't that much useful when istio cp components are merged to istiod.

frankbu:
So it sounds like the `minimal` profile, which initially was intended to be a light weight configuration (no Mixer, Citadel, etc) that people could use to experiment with traffic routing features only, is not very interesting anymore, now that everything is in the single istiod component. There is no use case for it anymore, other than the "control plane" only scenario (with gateways managed separately).

Therefore, I would suggest that we make the change in this PR along with my previous suggestion of changing the [minimal profile doc](https://preliminary.istio.io/latest/docs/setup/additional-setup/config-profiles/) to align with what it actually is these days:
```
1. **minimal**: same as the default profile, but only the control plane components are installed.
    This can be used to enable separate management of data plane components (e.g., gateways).
```
If we think that `minimal` is not a very good name for this profile anymore, we could deprecate it and add a new better profile name for it, but that would first require an `istio/istio` change, before we change it in the docs, so that would be done in a future PR.

@ostromart, @linsun WDYT?

@rytswd Can you please change the `minimal` configuration profiles doc, as suggested, as well.

rytswd:
Thanks for the input, @frankbu, I will be making the documentation update on the profile as well.

As to the actual wording for that, if there is a control-plane only installation, I'd think data plane management should be done in a separate operator input with empty profile - just like how the experimental [Managing Gateways with Multiple Revisions](https://istio.io/latest/docs/setup/upgrade/gateways/) suggests. What do you think about the following with some minor edit to your suggestion?

```
1. **minimal**: same as the default profile, but only the control plane components are installed.
    This allows custom configuration of data plane components (e.g., gateways), and even having a separate profile to manage data plane components with the empty profile.
```

At that point, it might be better to move up the "empty" profile bullet so that the "minimal" explanation flows without introducing new jargon.

--

Also, I can certainly look to work on introducing a new profile over in the main repo. The naming is something we'd need to discuss - I'd be more than happy to initiate that discussion in a separate issue.

frankbu:
Thanks @rytswd. I like your suggestion about making the doc for the `minimal` profile a little more clear about how it can be used to separate the data plane components, but I think we should avoid being too specific about using it in conjunction with the empty profile, at least until the pattern (or patterns) is better understood. Maybe just referencing the example would be best:
```
1. **minimal**: same as the default profile, but only the control plane components are installed.
    This allows you to configure the control plane and data plane components (e.g., gateways) using [separate profiles](/docs/setup/upgrade/gateways/#installation-with-istioctl).
```

frankbu:
Thanks @rytswd. I like your suggestion about making the doc for the `minimal` profile a little more clear about how it can be used to separate the data plane components, but I think we should avoid being too specific about using it in conjunction with the empty profile, at least until the pattern is better understood. Maybe just referencing the example would be best:
```
1. **minimal**: same as the default profile, but only the control plane components are installed.
    This allows you to configure the control plane and data plane components (e.g., gateways) using [separate profiles](/docs/setup/upgrade/gateways/#installation-with-istioctl).
```

linsun:
@frankbu I'm good to clarify min profile as cp only profile.  better than what we have right now :)

sdake:
@frankbu SGTM.

Cheers,
-steve

rytswd:
Thanks folks, I have taken @frankbu's suggestion and updated the profile doc accordingly 👍

frankbu (approved):

PR #8631, opened 2020-12-05
Remove common problem port name conflict

zufardhiyaulhaq:
We found that port name conflicting in multiple gateway is already resolved. https://istio.io/docs/ops/common-problems/network-issues/#port-conflict-when-configuring-multiple-tls-hosts-in-a-gateway

I also try to spike https://github.com/istio/istio/issues/21158#issuecomment-738122949 and working fine. If this is true, we can directly merge this PR.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


frankbu:
I don't think we can remove this entry without first better understanding it. According to @howardjohn, the port name is used to key into a map, so if two gateways use the same port name, there will be a conflict. The exact symptom, when things break may be different now, so I think we need to figure out under what conditions something will break, how it will break, and update this doc entry accordingly.

zufardhiyaulhaq:
sure @frankbu, where is the discussion with @howardjohn, can you give some link related to the thread?

frankbu:
Hi @zufardhiyaulhaq

> where is the discussion with @howardjohn, can you give some link related to the thread?

There's been no discussion yet. @howardjohn just mentioned the portname key -> map problem to me offline.

@howardjohn Do you have any idea how to proceed with this issue, maybe pointers to code that we can look at to better understand the current behaviour? Maybe if two ports have the same name but different number or protocol is the current gotcha? Or does the map key cover more than just the port properties?

howardjohn:
https://github.com/istio/istio/issues/21394

frankbu:
Closing for now.

@zufardhiyaulhaq Please feel free to reopen if you determine that this entry still needs to be updated.

PR #8648, opened 2020-12-10
added port 15053 to the docs

rinormaloku:
Port 15053 is exposed by the pod but it wasn't documented.

The short description is based on the following comment: https://github.com/istio/istio/pull/22369#issuecomment-604513008 from @costinm  (which is back in March). 

Would be great if costin can take a look if the short description is correct.

### DIFF
--- a/content/en/docs/ops/deployment/requirements/index.md
+++ b/content/en/docs/ops/deployment/requirements/index.md
@@ -98,6 +98,7 @@ The following ports and protocols are used by the Istio sidecar proxy (Envoy).
 | 15008 | TCP | Envoy tunnel port (inbound) | No |
 | 15020 | HTTP | Merged Prometheus telemetry from Istio agent, Envoy, and application | No |
 | 15021 | HTTP | Health checks | No |
+| 15053 | DNS | DNS resolution | Yes |
 | 15090 | HTTP | Envoy Prometheus telemetry | No |
 
 The following ports and protocols are used by the Istio control plane (istiod).


content/en/docs/ops/deployment/requirements/index.md
```diff
@@ -109,6 +109,7 @@ The following ports and protocols are used by the Istio control plane (istiod).
 | 8080 | HTTP | Debug interface (deprecated) | No |
 | 443 | HTTPS | Webhooks | No |
 | 15014 | HTTP | Control plane monitoring | No |
+| 15053 | DNS | DNS resolution for Service Entries | No |
```
howardjohn, line 112: This is a legacy of an experiment that wasn't fully cleaned up, its only in the yaml manifest not the code. we should fully clean it up there ,not document it.

howardjohn: Or document that its in the agent (table just above this)

costinm:
Agree with John, it is not something that has reached 'beta' so may be better to not document it for now.

content/en/docs/ops/deployment/requirements/index.md
rinormaloku: Ah, so this was initially the design which is replaced/postceded by the local DNS Agent.

I'll move it to the table above as suggested.

rinormaloku:
@costinm I moved the doc to the section "ports and protocols are used by the Istio sidecar proxy". If you think it's superfluous, feel free to reject it. For me, it was confusing while I was investigating how the DNS Agent works.

costinm (commented):
My concern is that this is not a 'beta' or supported feature - we did use it internally as one of the DNS options for VMs, but now a better option exists and as such it's unlikely we'll move this to beta.

IMO the docs should reflect supported features and port.

rinormaloku:
@costinm makes sense. Closing this.

PR #8649, opened 2020-12-10
Update supported versions without known CVEs

jacob-delgado:
1.8.1 was released. https://istio.io/latest/news/security/istio-security-2020-011/ shows 1.8.0 contains a CVE.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


frankbu (approved):
We should change our process so that we update this file in the same PR as the release notes. This way we avoid having this file outdated for a couple of days after a new release is published.

PR #8654, opened 2020-12-11
keep meshID untouched for external cp test

linsun:
Think this should work :)  else, I need to understand why we have meshID in our initial doc.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/additional-setup/external-controlplane/test.sh
+++ b/content/en/docs/setup/additional-setup/external-controlplane/test.sh
@@ -40,7 +40,6 @@ snip_set_up_the_remote_cluster_1_modified() {
     # Update config file: delete CA certificates and meshID
     sed -i \
         -e '/proxyMetadata:/,+2d' \
-        -e '/meshID: mesh1/,+2d' \
         remote-config-cluster.yaml
 }
 
@@ -50,7 +49,6 @@ snip_set_up_the_control_plane_in_the_external_cluster_2_modified() {
     # Update config file: delete CA certificates and meshID, and update pilot vars
     sed -i \
         -e '/proxyMetadata:/,+2d' \
-        -e '/meshID: mesh1/,+2d' \
         -e '/INJECTION_WEBHOOK_CONFIG_NAME: ""/d' \
         -e "s/VALIDATION_WEBHOOK_CONFIG_NAME: \"\"/ISTIOD_CUSTOM_HOST: ${EXTERNAL_ISTIOD_ADDR}/" \
         external-istiod.yaml


frankbu:
Hmm, this brings up another question. Why are these needed in the doc at all, given that it's not a multiCluster setup?
```
      meshID: mesh1
      multiCluster:
        clusterName: $REMOTE_CLUSTER_NAME
```

linsun:
@nmittler requested to add these.  My understanding is these values are helpful when you want to grow from 1 cluster to multiclusters.

frankbu:
>  My understanding is these values are helpful when you want to grow from 1 cluster to multiclusters.

That makes sense, but without explaining it, this is just making things more confusing by conflating two different features (multicluster and external controlplane) into one example. I think it would be better to add a tip or some text that says that even though this examples is a single cluster mesh, it configures the multiCluster properties so that additional clusters can be easily added to the same mesh in the future.

@nmittler WDYT?

nmittler:
@frankbu @linsun we want our instructions for single cluster to be able to easily expand to multicluster. If our single cluster instructions exclude cluster name, then they would effectively have to re-install the first cluster before adding clusters.

linsun:
Yep agreed.  I think we probably should update our single cluster install instruction as well.  We are just slightly ahead with external cp doc as it is relatively new.

frankbu:
> we want our instructions for single cluster to be able to easily expand to multicluster.

Seems rather than complicating the single cluster instructions with this extra config, can't we just set them by default, e.g., meshID defaults to "mesh1", or something like that, and we extract the clusterName from the Kubeconfig where Istio is being installed?

frankbu (approved):

nmittler:
> Seems rather than complicating the single cluster instructions with this extra config, can't we just set them by default, e.g., meshID defaults to "mesh1", or something like that, and we extract the clusterName from the Kubeconfig where Istio is being installed?

This has been discussed in the past but there was never any real traction. IIRC part of the concern was wanting human-readable names, but also wanting the user to actively think about them to avoid conflicts. So if they create 2 meshes, they wouldn't automatically both be `mesh1`, for example. Cluster ID is a little different ... by default the tool uses a hash that is extracted from the `kube-system` namespace, but it's not human-readable. As part of the k8s MCS work, kubernetes is considering incorporating the concept of real cluster IDs .... if that happens, we'll just use that by default.

But for now, making the user think about it and choose names up front is the best way to prevent them from shooting themselves in the foot when they want to expand the mesh.

frankbu:
> But for now, making the user think about it and choose names up front is the best way to prevent them from shooting themselves in the foot when they want to expand the mesh.

Makes sense, but our current instructions aren't doing that. All of the single cluster examples, default installation profiles, etc., are currently pointing the gun straight at the users foot :) Doc isn't really the answer either, since most people don't read them. Maybe `istioctl install` should generate an error (or at least a warning) if the meshID and clusterName aren't set?

linsun:
Thanks @frankbu @nmittler for the discussion!  I think having istioctl install generate a warning makes sense given where we are, before we can automatically generate it for the user in a meaningful way.

@frankbu do you want to open an issue for UX WG?

PR #8656, opened 2020-12-13
clarifying VM installations for multi-cluster meshes

rinormaloku:
This took me quite some time to figure out, as the actual issue is so simple that I kept overlooking it (added it to the docs to spare other people's debugging time).

Though with some extra effort the --clusterID flag could be found out at execution time.

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -176,6 +176,11 @@ Install Istio and expose the control plane so that your virtual machine can acce
 
     {{< /tabset >}}
 
+    {{< warning >}}
+    The above command, for Multi-Cluster meshes requires the flag `--clusterID`
+    to indicate to which cluster the Workload is authenticating to.
+    {{< /warning >}}
+
 ## Configure the virtual machine
 
 Run the following commands on the virtual machine you want to add to the Istio mesh:
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -177,8 +177,8 @@ Install Istio and expose the control plane so that your virtual machine can acce
     {{< /tabset >}}
 
     {{< warning >}}
-    The above command, for Multi-Cluster meshes requires the flag `--clusterID`
-    to indicate to which cluster the Workload is authenticating to.
+    When connecting a VM to a multicluster Istio mesh, you will need to add the `--clusterID` argument to
+    the above command. Set the value to the name of the cluster corresponding to the `istioctl` context.
     {{< /warning >}}
 
 ## Configure the virtual machine


content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -176,6 +176,11 @@ Install Istio and expose the control plane so that your virtual machine can acce
 
     {{< /tabset >}}
 
+    {{< warning >}}
+    The above command, for Multi-Cluster meshes requires the flag `--clusterID`
```
frankbu, line 180: What value do you need to set --clusterID to? Is it the cluster name? Which of the multiple clusters do you set it to? Any of them? The design for VM integration with a multicluster mesh seems like it needs more explanation, assuming it is all thought through and works.

@ostromart PTAL

ostromart: @nmittler is a better person to comment on this.

rinormaloku: @frankbu you set it to the clusterName as set on Istio installation, e.g. if you follow [Install Multi-Primary](https://istio.io/latest/docs/setup/install/multicluster/multi-primary/#configure-cluster1-as-a-primary) it would be `cluster1` or `cluster2`
```
$ cat <<EOF > cluster1.yaml
kind: IstioOperator
spec:
  values:
    global:
      meshID: mesh1
      multiCluster:
        clusterName: cluster1
      network: network1
EOF
```

Clarifying that is a good point as well. Wouldn't know how to fit it here without going down a rabbit hole.

stevenctl: The clusterID should match what `istioctl` was pointing to when we ran `istioctl x workload entry configure`. That command generates a ServiceAccount token and we need to tell istiod which k8s to validate the token against.

rinormaloku: @stevenctl are you 100% sure? 
I think most of the time that would be the case (if you have the context and the clusterName of the same value), but what stops you from naming the cluster anything else?

stevenctl: Sorry, I should be more clear. The clusterID given to `x workload entry` should be the name of the cluster (set during installation) of the cluster that `istioctl` is pointing to.

stevenctl: I plan on updating the command itself to infer the clusterID, per @frankbu's excellent suggestion.

stevenctl: Actually, for now it's best to have users specify the cluster ID. Inferring it is not so simple.. the reliable places to find it in a live cluster is are istiod deployment, and the injector configmap. We need to be able to fetch it from a more standard location.

rinormaloku: @stevenctl do you think we should make the proposed change to the docs? If yes is it clear enough?

content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -176,6 +176,11 @@ Install Istio and expose the control plane so that your virtual machine can acce
...
+    to indicate to which cluster the Workload is authenticating to.
+    {{< /warning >}}
```
frankbu, line 182:
```suggestion
    {{< warning >}}
    When connecting a VM to a multicluster Istio mesh, you will need to add the `--clusterID` argument to
    the above command. Set the value to the name of the cluster corresponding to the `istioctl` context.
    {{< /warning >}}
```

content/en/docs/setup/install/virtual-machine/index.md
stevenctl: I'd make it clear that the "cluster the Workload is authenticating to" is the cluster istioctl is connecting to (current kubeconfig/context)

frankbu: I don't think you need to need to explain here the reason (workload is authenticating to it), just make it clear what to set it to. See https://github.com/istio/istio.io/pull/8656#discussion_r543426311

stevenctl (approved):

stevenctl:
Thanks @rinormaloku!

frankbu (approved):

PR #8657, opened 2020-12-14
No need to create namespace manually for --watchNamespaces

shamsher31:
After this https://github.com/istio/istio/pull/29557, now it's no longer required to create namespaces manually.

The following config will automatically create a namespace for `--watchNamespaces`.

```
$ istioctl operator init --watchedNamespaces=ns1
```
```
# cat iop.yaml
kind: IstioOperator
metadata:
  namespace: ns1
spec:
  values:
    global:
      istioNamespace: ns1
    meshConfig:
      rootNamespace: ns1
```
```
$ istioctl install -f iop.yaml
```

### DIFF
--- a/content/en/docs/setup/install/operator/index.md
+++ b/content/en/docs/setup/install/operator/index.md
@@ -108,6 +108,8 @@ spec:
   values:
     global:
       istioNamespace: istio-namespace1
+    meshConfig:
+      rootNamespace: istio-namespace1
 {{< /text >}}
 
 {{< tip >}}


content/en/docs/setup/install/operator/index.md
```diff
@@ -108,6 +108,8 @@ spec:
   values:
     global:
       istioNamespace: istio-namespace1
+    meshConfig:
+      rootNamespace: istio-namespace1
```
frankbu, line 112: What does this have to do with what this example is showing (how to install istiod in a namespace other than istio-system)? Configuring rootNamespace is an unrelated function, so showing it here is just extra noise.

frankbu: btw, what is the default value of `rootNamespace`? It should default to the value of `rootNamespace` IMO.

frankbu: Also, this change doesn't seem to have any relationship to the title of this PR (No need to create namespace manually ...). If that's what this change is about, I expect to see a create ns call removed from the doc somewhere, not some added config.

shamsher31: > What does this have to do with what this example is showing (how to install istiod in a namespace other than istio-system)? Configuring rootNamespace is an unrelated function, so showing it here is just extra noise.


Yes, it looks unrelated but it is required to create EnvoyFilter in custom namespaces. see https://github.com/istio/istio/pull/29557#issuecomment-743236276

shamsher31: > what is the default value of rootNamespace

The default is `istio-system`and its used [here](https://github.com/istio/istio/blob/410453d1a22a3a867e09a669751db58810488832/manifests/charts/istio-control/istio-discovery/templates/telemetryv2_1.9.yaml#L8).

frankbu: Are you saying that you cannot change istioNamespace, without also setting rootNamespace as well? If so, that's definitely a bug, otherwise we don't need two different variables.

shamsher31: Are you saying that you cannot change istioNamespace, without also setting rootNamespace as well? 

Yes, you can try this step https://github.com/istio/istio/pull/29557#issuecomment-743109707 and see the output. That is why we need to set one additional parameter https://github.com/istio/istio.io/pull/8657#discussion_r542474497.

> If so, that's definitely a bug, otherwise we don't need two different variables.

I feel the same https://github.com/istio/istio/pull/29557#issuecomment-743244075

frankbu: Looks to me like it would also work if you create the `istio-system` namespace before applying the IOP.

Rather than adding a mysterious bug-workaround in the above example, I suggest opening an issue for the bug, and depending on the resolution of the issue, if this still can be easily misconfigured, add an entry to https://preliminary.istio.io/latest/docs/ops/common-problems/validation/, and probably `istioctl validate` or `istioctl analyze` should add checks to detect this misconfiguration.

shamsher31: > Looks to me like it would also work if you create the istio-system namespace before applying the IOP.

Yes, it works. 

Closing this. Thanks.

PR #8711, opened 2020-12-28
correct secret naming spec using in Gateway configuration

SerenaFeng:

In practice, secret name used in Gateway.tls.credentialName is not restricted to ending with "-cacert". Just that, when suffixed with it, "-cacert" suffix can be ignored for short.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination-sds/index.md
+++ b/content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination-sds/index.md
@@ -215,8 +215,7 @@ to hold the configuration of the NGINX server:
     $ kubectl create secret generic client-credential-cacert --from-file=ca.crt=example.com.crt -n istio-system
     {{< /text >}}
 
-    Note that the secret name for an Istio CA-only certificate must end with `-cacert` and the secret **must** be
-    created in the same namespace as Istio is deployed in, `istio-system` in this case.
+    Note that the secret **must** be created in the same namespace as Istio is deployed in, `istio-system` in this case.
 
     {{< warning >}}
     The secret name **should not** begin with `istio` or `prometheus`, and the secret **should not** contain a `token` field.
@@ -322,7 +321,7 @@ to hold the configuration of the NGINX server:
             number: 443
           tls:
             mode: SIMPLE
-            credentialName: client-credential # this must match the secret created earlier without the "-cacert" suffix
+            credentialName: client-credential # this must match the secret created earlier, if the secret name ends with "-cacert", "-cacert" suffix can be ignored
             sni: my-nginx.mesh-external.svc.cluster.local
     EOF
     {{< /text >}}


SerenaFeng:
> /ok-to-test

@ericvn , I saw gencheck_istio.io test failed, but can't see why from the logs, would you please help me with this issue?

content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination-sds/index.md
```diff
@@ -322,7 +321,7 @@ to hold the configuration of the NGINX server:
             number: 443
           tls:
             mode: SIMPLE
-            credentialName: client-credential # this must match the secret created earlier without the "-cacert" suffix
+            credentialName: client-credential # this must match the secret created earlier, if the secret name ends with "-cacert", "-cacert" suffix can be ignored
```
frankbu, line 324: The "secret created earlier" DOES have a `-cacert` suffix. There is no "if" in this case.

SerenaFeng: Here, I want to express that if the secret name is not suffixed with "-cacert", the credentialName must be exactly matched with the secret name, if so, then "-cacert" suffix can be ignored. do you suggest I change the description to " because the secret name ends with "-cacert", "-cacert" can be ignored here" ?

frankbu: I would not try to explain that in the DestinationRule comment.

Seems to me that there are some strange rules for secret names ending in `-cacert` that should be documented, probably in here: https://preliminary.istio.io/latest/docs/reference/config/networking/destination-rule/#ClientTLSSettings.

The earlier step in this task said that `the secret name for an Istio CA-only certificate must end with -cacert`, but you are also changing that. If that's no longer the case, then why not change the secret name in this task to not have the `-cacert` suffix. That way the task doesn't even need to address the issue at all.

cc @istio/wg-security-maintainers

howardjohn: https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-mutual-tls-ingress-gateway (bottom of that section, I think we are missing a header...) explains this.

Basically, there are various formats possible:
```
A TLS Secret with keys tls.key and tls.crt, as described above. For mutual TLS, a ca.crt key can be used.
A generic Secret with keys key and cert. For mutual TLS, a cacert key can be used.
A generic Secret with keys key and cert. For mutual TLS, a separate generic Secret named <secret>-cacert, with a cacert key. For example, httpbin-credential has key and cert, and httpbin-credential-cacert has cacert.
```

craigbox:
@SerenaFeng @howardjohn Is this ready for merging?

howardjohn (commented):
we should just link to https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-mutual-tls-ingress-gateway (the heading is missing, fixed in https://github.com/istio/istio.io/pull/9276), and not try to explain the logic in N different places

ericvn:
PR seems to be abandoned. Please reopen/rebase if you want this considered again.

PR #8738, opened 2021-01-06
multicluster: use cluster name from exported context

shamsher31:
Ref: https://preliminary.istio.io/latest/docs/setup/install/multicluster/multi-primary

This makes it easy to copy a snippet and run it directly even when the cluster name is different.

### DIFF
--- a/content/en/docs/setup/install/multicluster/multi-primary/index.md
+++ b/content/en/docs/setup/install/multicluster/multi-primary/index.md
@@ -38,7 +38,7 @@ spec:
     global:
       meshID: mesh1
       multiCluster:
-        clusterName: cluster1
+        clusterName: $CTX_CLUSTER1
       network: network1
 EOF
 {{< /text >}}
@@ -62,7 +62,7 @@ spec:
     global:
       meshID: mesh1
       multiCluster:
-        clusterName: cluster2
+        clusterName: $CTX_CLUSTER2
       network: network1
 EOF
 {{< /text >}}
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/install/multicluster/multi-primary/index.md
```diff
@@ -38,7 +38,7 @@ spec:
     global:
       meshID: mesh1
       multiCluster:
-        clusterName: cluster1
+        clusterName: $CTX_CLUSTER1
```
frankbu, line 41: This is not correct. The env variable contains the "context name", which may be different from the "cluster name".

nmittler: Frank is right ... we don't have a variable representing the name of the cluster. It's simply hard-coded to `cluster1`. We could potentially add variables for the cluster names, but then we'd also want to add variables for the mesh and network names for consistency. I'm not sure it's worth it.

shamsher31: So the cluster name can be anything and should not be matching the context name? 
If that is the case then I think we don't need variables and hardcodoing works fine.

nmittler: Right, the cluster name is something human-readable, like `cluster1`. The name of the kubernetes context, however, isn't always up to the user ... it could be anything and is often generated by the cloud provider.

shamsher31: Thanks @nmittler. I will close this PR.

shamsher31:
Closing bases on https://github.com/istio/istio.io/pull/8738#discussion_r553489334

PR #8745, opened 2021-01-07
note the --revision flag for eastwest gateway

stevenctl:
This is something that can be easily misconfigured, and difficult for users to discover. 

I've been able to replicate this issue: https://github.com/istio/istio/issues/29234 when not specifying the revision flag on the eastwest script, but installing the control plane with a revision, and having an old control plane without a revision.


content/en/docs/setup/install/multicluster/multi-primary_multi-network/index.md
```diff
@@ -76,6 +76,10 @@ $ @samples/multicluster/gen-eastwest-gateway.sh@ \
     istioctl --context="${CTX_CLUSTER1}" install -y -f -
 {{< /text >}}
 
+{{< tip >}}
```
frankbu, line 79: Would a warning be more appropriate?

```suggestion
{{< warning >}}
```

nmittler (approved):

PR #8754, opened 2021-01-08
Add info about using Metallb for multi-cluster setup on kind

shamsher31:
To test primary-remote multi-cluster setup on `kind`, The `cluster2` requires `remotePilotAddress`  which can be `eastwestgateway` load balancer ip. Considering KinD does not come with load balancer, `Metallb` helps provide external IP as a load balancer.

### DIFF
--- a/content/en/docs/setup/install/multicluster/primary-remote/index.md
+++ b/content/en/docs/setup/install/multicluster/primary-remote/index.md
@@ -17,7 +17,7 @@ Before proceeding, be sure to complete the steps under
 [before you begin](/docs/setup/install/multicluster/before-you-begin).
 
 {{< tip >}}
-If you are testing multicluster setup on `kind` you can use [MetalLB](https://metallb.universe.tf/installation/) to make use of `EXTERNAL-IP` for `LoadBalancer` services.
+{{< boilerplate multi-cluster-with-metallb >}}
 {{< /tip >}}
 
 In this configuration, cluster `cluster1` will observe the API Servers in
--- a/content/en/docs/setup/install/multicluster/primary-remote/index.md
+++ b/content/en/docs/setup/install/multicluster/primary-remote/index.md
@@ -16,9 +16,7 @@ connectivity between the pods in both clusters.
 Before proceeding, be sure to complete the steps under
 [before you begin](/docs/setup/install/multicluster/before-you-begin).
 
-{{< tip >}}
 {{< boilerplate multi-cluster-with-metallb >}}
-{{< /tip >}}
 
 In this configuration, cluster `cluster1` will observe the API Servers in
 both clusters for endpoints. In this way, the control plane will be able to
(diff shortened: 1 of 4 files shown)


content/en/docs/setup/install/multicluster/primary-remote/index.md
```diff
@@ -128,6 +128,10 @@ $ istioctl x create-remote-secret \
 
 Save the address of `cluster1`’s east-west gateway.
 
+{{< tip >}}
+If you are testing primary-remote multi-cluster setup on `Kind` you can use [Metallb](https://metallb.universe.tf/installation/) to make use of EXTERNAL-IP for Load Balancer.
```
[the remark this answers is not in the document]
shamsher31: Thanks.

content/en/docs/setup/install/multicluster/primary-remote/index.md
```diff
@@ -128,6 +128,10 @@ $ istioctl x create-remote-secret \
 
 Save the address of `cluster1`’s east-west gateway.
 
+{{< tip >}}
+If you are testing primary-remote multicluster setup on `Kind` you can use [MetalLB](https://metallb.universe.tf/installation/) to make use of `EXTERNAL-IP` for `LoadBalancer` services.
```
ericvn, line 132: I believe they do not uppercase the `k` in kind.

shamsher31: Done.

content/en/docs/setup/install/multicluster/primary-remote/index.md
```diff
@@ -128,6 +128,10 @@ $ istioctl x create-remote-secret \
 
 Save the address of `cluster1`’s east-west gateway.
 
+{{< tip >}}
```
frankbu, line 131: Why would we have a tip for just one specific platform type, and only for one of the 4 setup variations? If we need such a tip, it should be generic, for any platform without an external LB, and it should be either moved up to the before you begin section, or made into a boilerplate and included everywhere it applies.

cc @nmittler

shamsher31: I think this is needed for a single network. for multinetwork it works fine without an external LB.  Different cloud vendors already provide external LB and this is just for local testing. I found a ref for using it with `kind` but not sure about `minikube`.

I will move this instruction to the top.

ericvn: @shamsher31 In terms of using a common message, Frank had mentioned turning it into a boilerplate. See https://preliminary.istio.io/latest/about/contribute/shortcodes/#use-boilerplate-text for that.

shamsher31: @ericvn Thanks. I have created a boilerplate for the same.

ericvn (approved):
Confirmed with @shamsher31 this is only needed on the two pages here. Also, the multicluster tests are expected to fail right now, for example: https://github.com/istio/istio/issues/29389

frankbu (changes requested):
The tip and /tip parts should also be in the boilerplate

frankbu (approved):

PR #8746, opened 2021-01-07
update virtual machine docs and fix multi-network doc

stevenctl:
- [X] Use heredoc/file for workload group instead of `istioctl x workload group create` cc @howardjohn 
- [X] Multi-network doc uses new installation steps 
- [ ] examples/virtual-machines endpoint validation should use hello-world instead of httpbin to match setup/virtual-machine

fixes https://github.com/istio/istio/issues/29499
fixes https://github.com/istio/istio.io/issues/8602

Not sure how we track translated docs becoming out of sync...

PR #8755, opened 2021-01-08
fix rendering issues in vm doc

stevenctl:
https://github.com/istio/istio.io/pull/8746 merged prematurely

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -208,11 +208,11 @@ metadata:
   namespace: "${VM_NAMESPACE}"
 spec:
   metadata:
-        labels:
-          app: "${VM_APP}"
-      template:
-        serviceAccount: "${SERVICE_ACCOUNT}"
-        network: "${VM_NETWORK}"
+    labels:
+      app: "${VM_APP}"
+  template:
+    serviceAccount: "${SERVICE_ACCOUNT}"
+    network: "${VM_NETWORK}"
 EOF
 {{< /text >}}
 
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -44,16 +44,16 @@ and `SERVICE_ACCOUNT`
     $ CLUSTER="Kubernetes"
     {{< /text >}}
 
-    {{< tab name="Multi-Network" category-value="multiple" >}}
-
     {{< /tab >}}
+
+    {{< tab name="Multi-Network" category-value="multiple" >}}
 
     {{< text bash >}}
     $ VM_APP="<the name of the application this VM will run>"
     $ VM_NAMESPACE="<the name of your service namespace>"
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -194,64 +194,64 @@ Install Istio and expose the control plane so that your virtual machine can acce
 
 1. Create a template `WorkloadGroup` for the VM(s)
 
-    {{< tabset category-name="registration-mode" >}}
+{{< tabset category-name="registration-mode" >}}
```
frankbu, line 197: If you don't indent this 4 spaces, the numbered list won't work (the next list item will reset back to `1.`, instead of increment to `2.`).

stevenctl:
@frankbu do you know why the yaml in this heading isn't indented properly:

https://deploy-preview-8755--preliminary-istio.netlify.app/latest/docs/setup/install/virtual-machine/#create-files-to-transfer-to-the-virtual-machine

content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -235,15 +237,15 @@ Install Istio and expose the control plane so that your virtual machine can acce
...
+            labels:
+              app: "${VM_APP}"
+          template:
+            serviceAccount: "${SERVICE_ACCOUNT}"
+            network: "${VM_NETWORK}"
     EOF
     {{< /text >}}
 
-    1. Push the `WorkloadGroup` to the cluster:
+1. Push the `WorkloadGroup` to the cluster:
```
frankbu, line 248: This one is still not rendering correctly as item `2.`. Try indenting the list items 4 spaces and the text sections 8 spaces. If that doesn't work, maybe nested list items don't work in which case I would suggest just making them straight text (remove the numbering), since there are only 2 steps here anyway.

stevenctl: I've just removed the numbering. Looks good now, thanks for the tip!

content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -192,108 +193,107 @@ Install Istio and expose the control plane so that your virtual machine can acce
...
-    {{< /text >}}
+{{< text bash >}}
+$ kubectl --namespace ${VM_NAMESPACE} apply -f workloadgroup.yaml
+{{< /text >}}
 
-    {{< /tab >}}
+{{< /tab >}}
 
-    {{< /tabset >}}
+{{< /tabset >}}
```
frankbu, line 253: Shouldn't these be moved above line 245 (i.e., the `Then, to allow ...` part should be after the tabset)?

stevenctl: That tab should contain both steps, they're both auto-registration related.

stevenctl: Instead, I've moved "First, create a template..." into each tab.

frankbu (approved):

PR #8773, opened 2021-01-13
Silence curl command

ericvn:
The curl commands usually have some extraneous output (at least on my Mac):
```
kubectl exec "$(kubectl get pod -l app=ratings -o jsonpath='{.items[0].metadata.name}')" -c ratings -- curl productpage:9080/productpage | grep -o "<title>.*</title>"


  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  5183  100  5183    0     0   8072      0 --:--:-- --:--:-- --:--:--  8073
<title>Simple Bookstore App</title>
```

Adding the `-s` removes the progress meter. I know if would also possibly remove the error coming back if things were setup incorrectly, but the grep would have removed that as well.

### DIFF
--- a/content/en/docs/examples/bookinfo/index.md
+++ b/content/en/docs/examples/bookinfo/index.md
@@ -138,7 +138,7 @@ If you use GKE, please ensure your cluster has at least 4 standard GKE nodes. If
     example from `ratings`:
 
     {{< text bash >}}
-    $ kubectl exec "$(kubectl get pod -l app=ratings -o jsonpath='{.items[0].metadata.name}')" -c ratings -- curl -s productpage:9080/productpage | grep -o "<title>.*</title>"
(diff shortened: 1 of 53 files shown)


frankbu:
Same problem here: https://preliminary.istio.io/latest/docs/setup/install/multicluster/verify/#verifying-cross-cluster-traffic

Maybe should add -s there (probably also in the verify external control plane instructions). Maybe other places as well.

ericvn:
@frankbu Maybe too aggressive, but I did look at all the files where we were doing `curl` and basically added `s` (silent) and `S` (show errors even if silent) to the curl commands. 

Three of the tests failed, so I removed the `S`. With the `S` would now see something like:
```
curl: (56) Recv failure: Connection reset by peer
```
instead of the prior (no `S`)
```
command terminated with exit code 56
```

I probably could have updated the output in the index.md file in lieu of removing the `S`.

nmittler (approved):

PR #8802, opened 2021-01-20
fix mirror task doc

richardwxn:
the mirror_percent is deprecated, use mirrorPercent instead


ericvn:
Looks like the you need to do a `make gen` to create the new snips.sh.

frankbu (approved):

PR #8803, opened 2021-01-20
dashboard controlz: append ns to pod for controlz output

shamsher31:
To fix https://github.com/istio/istio/issues/30208

Controlz does not accept the `-n` flag, instead, the namespace is inferred to the pod name.

Install Istio.
```
[istio-1.9.0-beta.0]$ ./bin/istioctl install --set values.global.jwtPolicy=first-party-jwt
```

List pod.
```
[istio-1.9.0-beta.0]$ k get po -n istio-system
NAME                                   READY   STATUS    RESTARTS   AGE
istio-ingressgateway-758896c45-q6vpk   1/1     Running   0          13m
istiod-8645cf86d4-9gh29                1/1     Running   0          13m
```

Controlz dashboard
```
[istio-1.9.0-beta.0]$ istioctl dashboard controlz istiod-8645cf86d4-9gh29.istio-system
http://localhost:9876
```

### DIFF
--- a/content/en/docs/ops/diagnostic-tools/controlz/index.md
+++ b/content/en/docs/ops/diagnostic-tools/controlz/index.md
@@ -28,7 +28,7 @@ To access the ControlZ page of istiod, you can port-forward its ControlZ endpoin
 locally and connect through your local browser:
 
 {{< text bash >}}
-$ istioctl dashboard controlz <istiod pod name>.istio-system
+$ istioctl dashboard controlz deployment/istiod.istio-system
 {{< /text >}}
 
 This will redirect the component's ControlZ page to `http://localhost:9876` for remote access.
--- a/content/en/docs/ops/diagnostic-tools/controlz/index.md
+++ b/content/en/docs/ops/diagnostic-tools/controlz/index.md
@@ -28,7 +28,7 @@ To access the ControlZ page of istiod, you can port-forward its ControlZ endpoin
 locally and connect through your local browser:
 
 {{< text bash >}}
-$ istioctl dashboard controlz <istiod pod name> -n istio-system
+$ istioctl dashboard controlz <istiod pod name>.istio-system
 {{< /text >}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/diagnostic-tools/controlz/index.md
```diff
@@ -28,7 +28,7 @@ To access the ControlZ page of istiod, you can port-forward its ControlZ endpoin
 locally and connect through your local browser:
 
 {{< text bash >}}
-$ istioctl dashboard controlz <istiod pod name> -n istio-system
+$ istioctl dashboard controlz <istiod pod name>.istio-system
```
frankbu, line 31: A good question is why doesn't it support `-n`?
cc @esnible

esnible: I verified this works in _istioctl_ 1.6.12 but is broken in _istioctl_ 1.7.2.  We should fix this there.

However, an improvement for this line, if we fixed the 1.7.x regression, would be `istioctl dashboard controlz deployment/istiod -n istio-system`.  That way the user could just cut-and-paste.

esnible (changes requested):

content/en/docs/ops/diagnostic-tools/controlz/index.md
shamsher31: That looks reasonable to me.
 
However currently, this does not work with 1.9 beta.

```
[istio-1.9.0-beta.0]$ ./bin/istioctl dashboard controlz deployment/istiod -n istio-system
Error: deployments.apps "istiod" not found
```

So I'm keeping the deployment name until we fix the 1.7 regression.

```
[stio-1.9.0-beta.0]$ ./bin/istioctl dashboard controlz deployment/istiod.istio-system
http://localhost:9876
```

esnible (approved):
I love it.  Shorter, easier, gets around the bug discovered with `-n`.

frankbu (approved):

Issue #8797, opened 2021-01-19
Automate "Wait for Resource Status to Apply Configuration"


esnible:
We should automate [Wait for Resource Status to Apply Configuration](https://istio.io/latest/docs/ops/configuration/mesh/config-resource-ready/).

PR #8823, opened 2021-01-21
Test Automation for ops/configuration/mesh/config-resource-ready

esnible:
Resolves https://github.com/istio/istio.io/issues/8797

### DIFF
--- a/content/en/docs/ops/configuration/mesh/config-resource-ready/test.sh
+++ b/content/en/docs/ops/configuration/mesh/config-resource-ready/test.sh
@@ -31,8 +31,8 @@ _verify_contains snip_apply_and_wait_for_httpbin_vs "$snip_apply_and_wait_for_ht
 
 
 # @cleanup
-kubectl apply -f samples/httpbin/httpbin.yaml
-kubectl apply -f samples/httpbin/httpbin-gateway.yaml
+kubectl delete -f samples/httpbin/httpbin.yaml
+kubectl delete -f samples/httpbin/httpbin-gateway.yaml
 # Delete the Istio this test installed
 kubectl delete ValidatingWebhookConfiguration istiod-istio-system
 kubectl get mutatingwebhookconfigurations -o custom-columns=NAME:.metadata.name --no-headers | xargs kubectl delete mutatingwebhookconfigurations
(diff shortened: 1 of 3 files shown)


content/en/docs/ops/configuration/mesh/config-resource-ready/test.sh
```diff
@@ -0,0 +1,39 @@
...
+
+echo '*** config-resource-ready step 1 ***'
+echo y | snip_install_with_enable_status
+
+echo '*** istioctl-analyze step 2 ***'
+_verify_contains snip_apply_and_wait_for_httpbin_vs "$snip_apply_and_wait_for_httpbin_vs_out"
+
+
+# @cleanup
+kubectl delete -f samples/httpbin/httpbin.yaml || true
```
frankbu, line 34: Hmm, why do you have `|| true` here? It shouldn't be needed since the framework automatically adds a `set +e` command at the beginning of the cleanup script: https://github.com/istio/istio.io/blob/47c8caa3090d5c81d331d2396776d259cd786ed2/pkg/test/istioio/framework.go#L63

therealmitchconnors (approved):

esnible:
/retest
`profile_none: TestDocs/tasks/observability/metrics/customize-metrics/test.sh/test.sh` failed, which is not my test.  Either a flake or my test has a cleanup problem?

esnible:
@frankbu Any hints on what is going wrong with the `profile: none` test?

frankbu:
@esnible Off hand, I don't see anything wrong with your test. The way you cleanup the webhook seems a little strange, but I don't see anything in the failing test that indicates any related problem. Maybe it's worth trying to use istioctl uninstall instead? Otherwise, my approach in the past was to just try adding/removing things from my test to see what happens. Maybe the failure is just a flake?

esnible:
I don't think it's a flake, because `/retest` didn't fix it.
I copied the cleanup script from _content/en/docs/ops/diagnostic-tools/istioctl-analyze/test.sh_ which was written before there was `istioctl uninstall`.  I will try switching to `uninstall`.

frankbu (approved):

PR #8848, opened 2021-01-25
Add missing step to multicluster tests

nmittler:
This shouldn't actually be needed by the tests, since it only affects pods without a sidecar. However, adding for completeness.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


nmittler:
Putting a hold on this, since it appears we've found a real issue with the injector: https://github.com/istio/istio/issues/30376

nmittler:
@ericvn this is no-longer required as the real fix was https://github.com/istio/istio/pull/30377. 

We should still commit this, however, since it makes the test match the docs.

ericvn (approved):

PR #8854, opened 2021-01-26
[test] Update istio test ref to post beta.1 to see if which tests would now pass

ericvn:
Discussion if some of the multicluster tests may now pass with a PR that went in post beta release.


nmittler:
@stevenctl looks like recent test framework changes might have broken docs. Can you investigate?

ericvn:
Will need to get a later istio/istio that has the cherry-pick of https://github.com/istio/istio/pull/30417 in it as it fixes the problem of a private field in a struct (made private in an earlier framework PR).

ericvn:
It looks like `multi-primary_multi-network` might be working now. I haven't looked at the other tests to see if their failures have changed.

ericvn:
The two Primary Remotes tests are now failing when trying to verify cross cluster traffic.

ericvn:
Looks like the `Getting Started` test is back to passing.

ericvn:
As the beta-1 PR was never merged and that is now for rc.0, this PR is no longer needed.

PR #8855, opened 2021-01-27
add sleep

kyessenov:
https://github.com/istio/istio/issues/29400

### DIFF
--- a/content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
+++ b/content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
@@ -52,6 +52,8 @@ kubectl get istiooperator installed-state -n istio-system -o yaml
 _wait_for_istio envoyfilter istio-system stats-filter-1.8
 _wait_for_istio envoyfilter istio-system stats-filter-1.9
 
+sleep 60s
+
 ## Verify if patching works correctly
 send_productpage_requests
 _verify_lines snip_verify_the_results_2 "
--- a/content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
+++ b/content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
@@ -52,6 +52,8 @@ kubectl get istiooperator installed-state -n istio-system -o yaml
 _wait_for_istio envoyfilter istio-system stats-filter-1.8
 _wait_for_istio envoyfilter istio-system stats-filter-1.9
 
+# TODO: remove this delay once we can reliably detect the stats extension configuration update
+# has been applied.
 sleep 60s
 
 ## Verify if patching works correctly


content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
```diff
@@ -52,6 +52,8 @@ kubectl get istiooperator installed-state -n istio-system -o yaml
 _wait_for_istio envoyfilter istio-system stats-filter-1.8
 _wait_for_istio envoyfilter istio-system stats-filter-1.9
 
+sleep 60s
```
mandarjog, line 55: Is there any wait_for we can have here instead of 1 minute?

kyessenov: There's no metric exposed indicating that envoy received EnvoyFilter, and v8 has booted, and the config has been applied, so it's not clear what to wait for. Available metrics are here https://www.envoyproxy.io/docs/envoy/latest/configuration/other_features/wasm. There was an issue with repeated re-application of config in the past which prevented a special metric in stats extension, but that might have been solved.

mandarjog: @therealmitchconnors does status field work with EnvoyFilter CR? You could wait for that. 
how widely is the EnvoyFilter configuration being applied?

therealmitchconnors: Yes, use https://github.com/istio/istio.io/blob/master/tests/util/helpers.sh#L86 _wait_for_istio() to wait for the envoyfilter to distribute to all proxies in the mesh.  This does not guarantee that V8 has started, but it's probably a better heuristic than sleeping alone, and provides valuable output when pilot fails to reconcile all changes to the proxies.

kyessenov: I think wait_for_istio is already there, yet the test still flakes.

ericvn: Simply adding a sleep is not enough. The documentation markdown is obviously still wrong as it says the cert will expire after 5 secs but it doesn't.

ericvn (changes requested):

content/en/docs/tasks/observability/metrics/customize-metrics/test.sh
frankbu: I think _wait_for_istio should work, but if you look at helpers.sh, the code is commented out, so it's actually just a NO-OP at the moment. https://github.com/istio/istio.io/blob/master/tests/util/helpers.sh#L86

This test's use of _wait_for_istio is different than most other tests.  Other tests aren't flaky because _wait_for_istio is really optional for them, because they call an _verify_xxx function, which retires the operation if necessary until the result is found (or it times out). This use is different because the operation being verified is to check the log, the operation to populate the log only happens once.

If `istioctl wait` is fixed now, we should uncomment the _wait_for_istio code and hopefully that will fix the flake here.

ericvn: Agree we need to figure out if istioctl wait is actually fixed. I think there was an issue for the original problem.

kyessenov: In absence of `istioctl wait`, can we use sleep for now? Flakey tests are a productivity drain for the entire team.

ericvn: One more question that I'm curious about is what changed in this code to cause the filters to not become active as quick as before? The test flakiness is not observed in 1.8 where we should running this same test.

Also, I would add a TODO and a comment on why the sleep is there so someone can potentially come back and implement something better than the sleep.

kyessenov: I'm not aware of any internal changes in the stats extension. I haven't tracked environmental changes (switch to v8, sizing) that might have an effect.

ericvn: The TODO looks good to me. @frankbu, do you think this is sufficient? Do we want to add something to the index.md that says the user may need to repeat the curl after waiting for the config to become active?

frankbu: Seems like the sleep is all that's needed for now, until we get _wait_for_istio fixed and that it waits for the filter to be not only present, but also active as @kyessenov pointsed out. Nobody has complained about the issue when running the instructions manually, so I assume that means that the human delay between running commands is long enough.

frankbu (approved):

PR #8868, opened 2021-01-28
Add troubleshooting guide for VMs

howardjohn:
This was requested as part of the VM graduation support ability review

### DIFF
--- a/content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
+++ b/content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
@@ -1,0 +1,121 @@
+---
+title: Debugging Virtual Machines
+description: Describes tools and techniques to diagnose issues with Virtual Machines.
+weight: 20
+keywords: [debug,virtual-machines,envoy]
+owner: istio/wg-environments-maintainers
+test: n/a
+---
+
+This page describes how to troubleshoot issues with Istio deployed to Virtual Machines.
+Before reading this, you should the steps in [Virtual Machine Installation](/docs/setup/install/virtual-machine/).
+
+Troubleshooting an Istio Virtual Machine installation is similar to troubleshooting issues with proxies running inside Kubernetes, but there are some key differences to be aware of.
+
+While much of the same information is available on both platforms, accessing this information differs.
+
+## Monitoring health
+
+The Istio sidecar is typically run as a `systemd` unit. To ensure its running properly, you can check that status:
+
+{{< text bash >}}
+$ systemctl status istio
+{{< /text  >}}
+
+Additionally, the sidecar health can be programmatically check at its health endpoint:
+
+{{< text bash >}}
+$ curl localhost:15021/healthz/ready -I
+{{< /text  >}}
+
+## Logs
+
+Logs for the Istio proxy can be found in a few places.
+
+To access the `systemd` logs, which has details about the initialization of the proxy:
+
+{{< text bash >}}
+$ journalctl -f -u istio -n 1000
+{{< /text  >}}
+
+The proxy will redirect `stderr` and `stdout` to `/var/log/istio/istio.err.log` and  `/var/log/istio/istio.log`, respectively.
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
```diff
@@ -0,0 +1,121 @@
...
+$ istioctl proxy-status
+NAME           CDS        LDS        EDS        RDS      ISTIOD                    VERSION
+vm-1.default   SYNCED     SYNCED     SYNCED     SYNCED   istiod-789ffff8-f2fkt     {{< istio_full_version >}}
+{{< /text  >}}
+
+However, `istioctl proxy-config` relies on functionality in Kubernetes to connect to a proxy, which will not work for virtual machines.
+Instead, a file containing the configuration dump from Envoy can be passed. For example:
+
+{{< text bash >}}
+$ curl -s localhost:15000/config_dump | istioctl proxy-config clusters --file -
```
jasonwzm, line 80: There should be a step before this to move istioctl to the VM, correct?

content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
```diff
@@ -0,0 +1,121 @@
...
+the virtual machine to become a part of a `Service`, similar to an `Endpoint` in Kubernetes.
+
+To check these are created correctly:
+
+{{< text bash >}}
+$ kubectl get workloadentries
+NAME             AGE   ADDRESS
+vm-10.128.0.50   14m   10.128.0.50
+{{< /text  >}}
+
```
jasonwzm, line 100: I think we should add some details on troubleshooting workload entries not created automatically. grepping the istiod logs may be good enough for now.

content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
howardjohn: I don't think users should actually put istioctl on the VM. The should extract a config dump through various means specific to their environment and then run istioctl somewhere else. We cannot really easily document how to run commands on a VM - all deployments are bespoke. Instead, we just explain how they can be used

content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
howardjohn: SGTM, I will add a note to look at logs. I also filed a bug to add metrics which can be helpful

content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
stevenctl: What about `ssh user@vm 'curl localhost:15000/config_dump' | istioctl ...` to make it more clear?

stevenctl (approved):

content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
jasonwzm: Yea, just some way to let the user know the config dump needs to be pulled from the VM.

content/en/docs/ops/diagnostic-tools/virtual-machines/index.md
```diff
@@ -0,0 +1,121 @@
...
+title: Debugging Virtual Machines
+description: Describes tools and techniques to diagnose issues with Virtual Machines.
+weight: 20
+keywords: [debug,virtual-machines,envoy]
+owner: istio/wg-environments-maintainers
+test: n/a
+---
+
+This page describes how to troubleshoot issues with Istio deployed to Virtual Machines.
+Before reading this, you should the steps in [Virtual Machine Installation](/docs/setup/install/virtual-machine/).
```
rootsongjc, line 11:
```suggestion
Before reading this, you should take the steps in [Virtual Machine Installation](/docs/setup/install/virtual-machine/).
```

nmittler (approved):

PR #8895, opened 2021-02-03
Change baseEjectionTime to be less than maxEjectionTime (default 300s)

hzxuzhonghu:
@ericvn @ramaraochavali @nmittler 

Hope this can fix

### DIFF
--- a/content/en/docs/tasks/traffic-management/locality-load-balancing/distribute/index.md
+++ b/content/en/docs/tasks/traffic-management/locality-load-balancing/distribute/index.md
@@ -56,7 +56,7 @@ spec:
     outlierDetection:
       consecutive5xxErrors: 100
       interval: 1s
-      baseEjectionTime: 10m
+      baseEjectionTime: 1m
 EOF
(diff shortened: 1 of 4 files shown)


ericvn:
@hzxuzhonghu You shouldn't be updating go.* files at all, especially changing it to a different branch. That requires additional changes in other files.

Are you thinking the just updating the DR will help, or are you thinking there is another fix needed?

hzxuzhonghu:
Only dr change

ericvn:
A quick sanity check running the DR change shows the tests should now pass. 

https://github.com/istio/istio/pull/30602 will fix the issue where 10m could still have been specified and the maxEjectionTime would have been increased to be greater than or equal the baseEjectionTime.

Thanks @hzxuzhonghu.

ericvn:
Only test failure at this point was the customize metrics flake.

ericvn:
The customize metrics test is a flake. No reason to keep testing.

ericvn (approved):

PR #8914, opened 2021-02-05
[release-1.8] Update to use the release-1.8 common-files.

ericvn:
It appears the release-1.8 branch was never updated to use the release-1.8 branch from common-files when the branch was cut. This PR moves the release-1.8 istio.io branch to use the 1.8 common-files.

Need to verify if the branching script does this for the new branch cut.


ericvn:
Need to add something to the script around here: https://github.com/istio/istio.io/blob/master/scripts/create_version.sh#L140

doing a `UPDATE_BRANCH=release-${PREV_MINOR} make update-common`

frankbu (approved):

PR #8923, opened 2021-02-07
fix yaml in gateway reference doc

zackzhangkai:
Signed-off-by: zackzhangkai <zhangkaiamm@gmail.com>

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/archive/v1.1/docs/reference/config/networking/v1alpha3/gateway/index.html
+++ b/archive/v1.1/docs/reference/config/networking/v1alpha3/gateway/index.html
@@ -128,8 +128,7 @@ spec:
   hosts:
   - mongosvr.prod.svc.cluster.local # name of internal Mongo service
   gateways:
-  - some-config-namespace/my-gateway # can omit the namespace if gateway is in same
-                                       namespace as virtual service.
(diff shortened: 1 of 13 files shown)


rvennam:
I believe these files are generated from istio/api and not intended to be modified directly. @ericvn

ericvn:
Correct @rvennam.  The top of the files has:
```
source_repo: https://github.com/istio/api
title: Gateway
description: Configuration affecting edge load balancer.
```

Which identifies the repo that this file comes from. One might have to do a little searching there to figure out which pb.html file matches this one. Once the file is updated there, the nightly runs will autoupdate the file in this repo.

We typically don't update the archives, as those files actually come from the release branches when they are pulled into the repo (happens when the branch is cut, and when an update in an older release branch warrants it (there is a script to do this)).

Also, we don't typically mix translated updates with English updates. That requires additional approvals which can delay the PR merging. Normally, we update the English version, and then the translators will pick the change into the translated stream. There are some changes in the works where translated pages which are still in English (never translated) will pick up the English pages.

ericvn:
I added the file location in https://github.com/istio/istio/issues/30728

PR #8932, opened 2021-02-09
Automate microservices tutorial

### LINKED ISSUE #8429: Add support to test framework to use the same test environment across to test multiple docs
The current test framework environment supports setting up an Istio environment and deleting the resources created when testing the functionality on a page. In cases where we have to use the same environment across multiple docs, the test framework does not support that scenario. An example where this support is needed is the "Learn Microservices using Kubernetes and Istio" example ( https://istio.io/latest/docs/examples/microservices-istio/) where we need to build on the test environment across multiple pages.

johnma14:
This PR tries to automate the microservices tutorial in the example section of the Istio docs.
It consists of 2 parts:
- update to the istio.io test framework to use the same test framework across multiple tests (parent-child relationship)
- added automated tests for the microservices tutorial which has the parent-child relationship

Updates to the test framework to support this scenario includes 3 new optional tags to the TestCase object -
- @parent: this is the parent id that will be used to identify the parent test case
- @child: this indicates that the test case if the child of a parent id. The parent id set using the @parent tag will
          be used here.
- @order: this tag will be used to specify the order in which the child tests need to be run

The child tests will be run as a subtest of the parent test.

Fixes #8429

[X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


craigbox:
Mariam, please have a look at the test/lint output so we can move this forward.  Thank you!

craigbox (commented):
If this is still relevant I'd love to get it merged.  @johnma14 ?

linsun:
@johnma14 no longer works on Istio any more. @ericvn thoughts?  Would you and someone willing to step up to get this to finish line? My thoughts are we are probably okay without the tutorial.

PR #8945, opened 2021-02-10
add Wasm extension feature stage.

bianpengyuan:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/about/feature-stages/index.md
+++ b/content/en/about/feature-stages/index.md
@@ -63,7 +63,7 @@ Below is our list of existing features and their current phases. This informatio
 
 | Feature           | Phase
 |-------------------|-------------------
-| [WebAssembly Extension](/docs/ops/configuration/extensibility/wasm-module-distribution) | Alpha
+| WebAssembly Extension | Alpha
 
 ### Security and policy enforcement
 
--- a/content/en/about/feature-stages/index.md
+++ b/content/en/about/feature-stages/index.md
@@ -59,6 +59,12 @@ Below is our list of existing features and their current phases. This informatio
 | [Trace Sampling](/docs/tasks/observability/distributed-tracing/configurability/#trace-sampling) | Beta
 | [Request Classification](/docs/tasks/observability/metrics/classify-metrics/) | Beta
 
+### Extensibility
+
+| Feature           | Phase
+|-------------------|-------------------
+| [WebAssembly Extension](/docs/ops/configuration/extensibility/wasm-module-distribution) | Alpha
+
 ### Security and policy enforcement
 
 | Feature           | Phase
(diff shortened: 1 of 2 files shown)


duderino (approved):
Thanks for doing this Peter.

douglas-reid (approved):

duderino (changes requested):
Actually, @bianpengyuan I think this should be experimental because it still relies on EnvoyFilter and that's not really an API.

@mandarjog @douglas-reid @PiotrSikora was something else agree on?

bianpengyuan:
@duderino Yeah make sense, we should block Wasm extension distribution alpha on API. I have updated this PR to mark this feature as experimental.  I thought alpha is the `initial` stage of any feature, but actually alpha also needs promotion. but I keep overall Wasm feature status as `alpha`, which is what we claimed at Istio 1.5.

content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
```diff
@@ -10,6 +10,8 @@ owner: istio/wg-policies-and-telemetry-maintainers
 test: no
 ---
 
+{{< boilerplate experimental-feature-warning >}}
```
ericvn, line 13: Normally [Experimental] is added to the end of the title as well.

content/en/about/feature-stages/index.md
```diff
@@ -59,6 +59,12 @@ Below is our list of existing features and their current phases. This informatio
 | [Trace Sampling](/docs/tasks/observability/distributed-tracing/configurability/#trace-sampling) | Beta
 | [Request Classification](/docs/tasks/observability/metrics/classify-metrics/) | Beta
 
+### Extensibility
+
+| Feature           | Phase
+|-------------------|-------------------
+| [WebAssembly Extension](/docs/ops/configuration/extensibility/wasm-module-distribution) | Alpha
```
ericvn, line 66: I does seem strange to have an Alpha feature have a link that leads to an Experimental page.

content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
bianpengyuan: Done.

content/en/about/feature-stages/index.md
bianpengyuan: I removed the doc link. We don't really have any other doc about Wasm except this linked one, or a blog post created back in 1.5 where we claimed alpha support for Wasm. Looks like there are other features in this page which does not have link attached, so I simply remove the link.

duderino (approved):

duderino:
TOC previously approved Wasm at Alpha, so I think that's valid.   Introducing Wasm remote loading as a separate feature at experimental also lgtm.  Splitting the pre-existing support/maturity from the new feature nicely addresses the issue.

ericvn (approved):

PR #8966, opened 2021-02-12
Fix Documentation on MetricConfig field: tagsToRemove

crhuber:
Fixes Documentation on MetricConfig field: tagsToRemove.
Currently if you configure your manifest file with tagsToRemove, it has no effect on the tags that are removed by Istio-proxy ie:

has no effect on tags:
```
    telemetry:
      enabled: true
      v1:
        enabled: false
      v2:
        enabled: true
        prometheus:
          configOverride:
            inboundSidecar:
              debug: false
              stat_prefix: istio
              metrics:
                - name: requests_total
                  tagsToRemove: ["destination_principal", "source_principal"]
```

However, this works:

```
    telemetry:
      enabled: true
      v1:
        enabled: false
      v2:
        enabled: true
        prometheus:
          configOverride:
            inboundSidecar:
              debug: false
              stat_prefix: istio
              metrics:
                - name: requests_total
                  tags_to_remove: ["destination_principal", "source_principal"]
```

This PR updates the documentation to reflect that `tags_to_remove` is the correct field name

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


crhuber:
@googlebot I signed it!

content/en/docs/reference/config/proxy_extensions/stats/index.html
```diff
@@ -54,7 +54,7 @@ <h2 id="MetricConfig">MetricConfig</h2>
 </td>
 </tr>
 <tr id="MetricConfig-tags_to_remove">
-<td><code>tagsToRemove</code></td>
+<td><code>tags_to_remove</code></td>
```
ericvn, line 57: This change comes from a file in another repo. Making the change here will be overwritten the next day.

I think the update might need to be here: https://github.com/istio/proxy/blob/master/extensions/stats/config.proto#L44

ericvn (changes requested):

content/en/docs/reference/config/proxy_extensions/stats/index.html
ericvn: I believe that comes from here: https://github.com/istio/proxy/blob/master/extensions/stats/config.proto#L44

I don'r know enough about the generation from the .proto to the .pb.html.

bianpengyuan:
https://github.com/istio/proxy/pull/3191 should supersede this.

ericvn:
https://github.com/istio/proxy/pull/3191 does have a change for the master branch which will update preliminary.istio.io. I did ask about cherrypicking it back to 1.9 which will fix istio.io (at least latest).

ericvn:
Closing this PR as istio/proxy#3191 should fix the problem and it is incorrect to fix in this repo.

PR #8973, opened 2021-02-13
Rewrite docs regarding load balancing

chris-schra:
Make clear that the zones and regions have
to match the actual K8s labels.
Add convenience scripts.

[X] Docs
[ ] Installation
[X] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md
+++ b/content/en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md
@@ -36,10 +36,12 @@ The following environment variables will be used for the various contexts:
 Variable | Description
 -------- | -----------
 `CTX_PRIMARY` | The context used for applying configuration to the primary cluster.
-`CTX_R1_Z1` | The context used to interact with pods in `region1.zone1`.
-`CTX_R1_Z2` | The context used to interact with pods in `region1.zone2`.
-`CTX_R2_Z3` | The context used to interact with pods in `region2.zone3`.
-`CTX_R3_Z4` | The context used to interact with pods in `region3.zone4`.
+`CTX_WORKERS` | The contexts used to interact with pods in the worker clusters.
+
+## Set the worker clusters
+
+{{< text bash >}}
+$ export CTX_WORKERS=('cluster1' 'cluster2'); // adjust to your contexts!
 
 ## Create the `sample` namespace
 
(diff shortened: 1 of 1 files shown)


chris-makaio:
@googlebot I signed it!

content/en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md
```diff
@@ -57,49 +59,23 @@ metadata:
...
   do \
     kubectl --context="$CTX" apply -f sample.yaml; \
+    for LOC in $(kubectl --context="$CTX" get nodes -o \
+      custom-columns=REGION:'{.metadata.labels.topology\.kubernetes\.io/region}',ZONE:'{.metadata.labels.topology\.kubernetes\.io/zone}' --no-headers  \
+      | uniq | sed -E 's/<none>/default/g' | sed -E 's/[ \t]+/\./g')
+    do \
+      ./@samples/helloworld/gen-helloworld.sh \
+        --version "$LOC" > "helloworld-${LOC}.yaml"; \
+        kubectl --context="$CTX" apply -f "helloworld-${LOC}.yaml"; \
+    done
```
rvennam, line 77: This is terse and clever, but I'm afraid this make it less readable and more difficult to modify. Open to other opinions

chris-makaio: I take that as a compliment :)
Of course you‘re right, but my point here is that I think it should be clear that we have to use the actual region & zone labels from the clusters in Istio to make load balancing work. I can split that commands, no problem...

related (different repo):
https://github.com/istio/istio/issues/30832

content/en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md
```diff
@@ -36,10 +36,12 @@ The following environment variables will be used for the various contexts:
 Variable | Description
 -------- | -----------
 `CTX_PRIMARY` | The context used for applying configuration to the primary cluster.
-`CTX_R1_Z1` | The context used to interact with pods in `region1.zone1`.
-`CTX_R1_Z2` | The context used to interact with pods in `region1.zone2`.
-`CTX_R2_Z3` | The context used to interact with pods in `region2.zone3`.
-`CTX_R3_Z4` | The context used to interact with pods in `region3.zone4`.
+`CTX_WORKERS` | The contexts used to interact with pods in the worker clusters.
```
nmittler, line 39: Using an array and relying on index positions seems generally less clear IMO. I think we should just revert this

content/en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md
nmittler: +1 ... we need to prioritize readability here.

chris-makaio:
guys, I understand and accept that you don't like the PR 😆 

The point is that it's not obvious how all the values relate to each other. Again, we could just simply set the values outside an array, but the important thing is -  I think - that one can't just ANY names for regions and zones. If I read "region1" and "zone1" it looks to me like I just can use any value as long as it is the same when passed to gen-helloworld

nmittler:
@chris-makaio The [intro page](https://istio.io/latest/docs/tasks/traffic-management/locality-load-balancing/) for the task does describe where these values from from (node labels). Does that help?

craigbox:
Chris; could you modify the PR to just make clear in comments the things you found confusing?

Xunzhuo (commented):
please fix lint erros below thanks!

./en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md:62: MD009 Trailing spaces
./en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md:63: MD009 Trailing spaces
./en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md:51: MD031 Fenced code blocks should be surrounded by blank lines
./en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md:60: MD031 Fenced code blocks should be surrounded by blank lines
./en/docs/tasks/traffic-management/locality-load-balancing/before-you-begin/index.md:66: MD031 Fenced code blocks should be surrounded by blank lines

craigbox (commented):
Thanks for taking a stab at this, @chris-schra. Given there wasn't any follow up from the comments the first time around, I'm going to close this; feel free to re-open if you wish.

PR #8442, opened 2020-11-04
Remove rbac instruction for GKE

upodroid:
I removed the RBAC step from GKE because this is no longer an issue for k8s version 1.11+.

https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control

[ ] Configuration Infrastructure
[ ] Docs
[X] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #8990, opened 2021-02-16
Revert "Remove rbac instruction for GKE"

howardjohn:
Reverts istio/istio.io#8442
Fixes https://github.com/istio/istio/issues/29483

### DIFF
--- a/content/en/docs/setup/platform-setup/gke/index.md
+++ b/content/en/docs/setup/platform-setup/gke/index.md
@@ -69,3 +69,13 @@ Follow these instructions to prepare a GKE cluster for Istio.
         --zone $ZONE \
         --project $PROJECT_ID
     {{< /text >}}
+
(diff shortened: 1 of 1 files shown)


smawson:
So why did we think it was safe to remove these before? Mostly curious since we removed these in November... Do we need to add any caveats here about when this step is needed?

howardjohn:
The original PR was sent by a new contributor and approved by someone who doesn't use GKE (I assume). I don't see anything suggesting it was ever safe to remove - although I am not an RBAC expert. I suspect the default permissions gives the owner of the cluster admin access, but if you are in a cluster not in your own project its needed?

smawson (approved):
I asked because you were one of the approvers on the original PR :-D

howardjohn:
> approved by someone who doesn't use GKE (I assume)

Hah, I did not look close enough, didn't see my name there...

PR #8999, opened 2021-02-18
[release-1.9] Update of reference docs

ericvn:
We usually wait and update reference docs whenever a patch release is made.

This is just an early run. I haven't verified if all these changes are already in 1.9.0 or if some are things that will be out with 1.9.1. Will add another commit when 1.9.1 is built and will update this.


ericvn:
The addition of `PILOT_ENABLE_DESTINATION_RULE_INHERITANCE` was in https://github.com/istio/istio/pull/30066 which was after the 1.9.0 build. Will want to wait until 1.9.1 is released to update the docs.

craigbox (approved):

PR #9016, opened 2021-02-19
[release-1.9] Bump basic auth Wasm extension version.

bianpengyuan:
Bump basic auth wasm extension version for 1.9 branch to include a fix: https://github.com/istio/istio/issues/4693#issuecomment-776296897

### DIFF
--- a/content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
+++ b/content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
@@ -86,7 +86,7 @@ spec:
                code:
                  remote:
                    http_uri:
-                     uri: https://github.com/istio-ecosystem/wasm-extensions/releases/download/{{< istio_version >}}.0/basic-auth.wasm
+                     uri: https://github.com/istio-ecosystem/wasm-extensions/releases/download/{{< istio_version >}}.1/basic-auth.wasm
                    # Optional: specifying sha256 checksum will let istio agent verify the checksum of downloaded artifacts.
                    # It is **highly** recommended to provide the checksum, since missing checksum will cause the Wasm module to be downloaded repeatedly.
                    # To compute the sha256 checksum of a Wasm module, download the module and run `sha256sum` command with it.


content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
```diff
@@ -86,7 +86,7 @@ spec:
                code:
                  remote:
                    http_uri:
-                     uri: https://github.com/istio-ecosystem/wasm-extensions/releases/download/{{< istio_version >}}.0/basic-auth.wasm
+                     uri: https://github.com/istio-ecosystem/wasm-extensions/releases/download/{{< istio_version >}}.1/basic-auth.wasm
```
ericvn, line 89: I'm not sure if there is a better mechanism to get the latest version updated automatically.

bianpengyuan: I think that extension will likely remain static for a very long time in the future, so the need to bump its version is very low. Automating this probably won't be very valuable.

ericvn: Also not sure of a good way to handle this across release boundaries. For example, preliminary.istio.io has: `https://github.com/istio-ecosystem/wasm-extensions/releases/download/1.10.0/basic-auth.wasm` which isn't a valid release at the moment. Does it make sense to remove istio_version?

bianpengyuan: The basic auth extension is tested with the corresponding Istio release, so I think it would be safe to keep the doc followig Istio release.

I am going to work on prow automation to release Wasm extension from the `istio-ecosystem/wasm-extensions` repo, so that when Istio has a release branch cut, a new release will be made from that repo, which makes sure that the istio.io doc will always be valid at the release time.

ericvn: > The basic auth extension is tested with the corresponding Istio release, so I think it would be safe to keep the doc followig Istio release.
> 
> I am going to work on prow automation to release Wasm extension from the `istio-ecosystem/wasm-extensions` repo, so that when Istio has a release branch cut, a new release will be made from that repo, which makes sure that the istio.io doc will always be valid at the release time.

So preliminary.istio.io will be valid for the time from branch cut to release (we cut the new release-1.10 branch you create the new extensions), but from release to the next branch cut preliminary.istio.io will not have a valid link (as it will expect a 1.11.0 (similar to expecting a 1.10.0 now).

ericvn (approved):

content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
bianpengyuan: > So preliminary.istio.io will be valid for the time from branch cut to release (we cut the new release-1.10 branch you create the new extensions), but from release to the next branch cut preliminary.istio.io will not have a valid link (as it will expect a 1.11.0 (similar to expecting a 1.10.0 now).

Yeah.. would this be problemtic?

PR #9017, opened 2021-02-19
Add out-of-mesh server metadata info into telemetry v2 faq.

bianpengyuan:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/faq/metrics-and-logs/telemetry-v1-vs-v2.md
+++ b/content/en/faq/metrics-and-logs/telemetry-v1-vs-v2.md
@@ -22,6 +22,9 @@ v2 which are listed below:
   missing peer attributes like workload name, namespace and labels.
   However, if both peers have proxies injected all the labels mentioned
   [here](/docs/reference/config/metrics/) are available in the generated metrics.
+  At Istio 1.8 release, a solution was added to distribute server workload metadata
+  to client sidecar, which makes client side metric have server workload metadata labels
+  filled even when the server workload is out of the mesh.
 
 * **TCP metadata exchange requires mTLS**
   TCP metadata exchange relies on the [Istio ALPN protocol](/docs/tasks/observability/metrics/tcp-metrics/#understanding-tcp-telemetry-collection)
--- a/content/en/faq/metrics-and-logs/telemetry-v1-vs-v2.md
+++ b/content/en/faq/metrics-and-logs/telemetry-v1-vs-v2.md
@@ -22,9 +22,9 @@ v2 which are listed below:
   missing peer attributes like workload name, namespace and labels.
   However, if both peers have proxies injected all the labels mentioned
   [here](/docs/reference/config/metrics/) are available in the generated metrics.
-Starting with the Istio 1.8 release, a solution was added to distribute server workload metadata
-  to client sidecar, which makes client side metric have server workload metadata labels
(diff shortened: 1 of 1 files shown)


content/en/faq/metrics-and-logs/telemetry-v1-vs-v2.md
```diff
@@ -22,6 +22,9 @@ v2 which are listed below:
   missing peer attributes like workload name, namespace and labels.
   However, if both peers have proxies injected all the labels mentioned
   [here](/docs/reference/config/metrics/) are available in the generated metrics.
+  At Istio 1.8 release, a solution was added to distribute server workload metadata
+  to client sidecar, which makes client side metric have server workload metadata labels
+  filled even when the server workload is out of the mesh.
```
rvennam, line 27: The way the section reads now is a bit confusing. If starting from 1.8, it now has workload labels for server workloads that are not part of the mesh, then why not just remove this whole section? It seems strange to say there is a problem and then say the solution is there since an older Istio release.

bianpengyuan: Stats reporting both at client proxy and server proxy. The 1.8 solution only resolves missing server metadata at client side. Stats reporting at server side still lacks metadata is the client is outside the mesh.

content/en/faq/metrics-and-logs/telemetry-v1-vs-v2.md
```diff
@@ -22,6 +22,9 @@ v2 which are listed below:
   missing peer attributes like workload name, namespace and labels.
   However, if both peers have proxies injected all the labels mentioned
   [here](/docs/reference/config/metrics/) are available in the generated metrics.
+  At Istio 1.8 release, a solution was added to distribute server workload metadata
```
ericvn, line 25:
```suggestion
Starting with the Istio 1.8 release, a solution was added to distribute server workload metadata
```

content/en/faq/metrics-and-logs/telemetry-v1-vs-v2.md
```diff
@@ -22,6 +22,9 @@ v2 which are listed below:
...
+  to client sidecar, which makes client side metric have server workload metadata labels
```
ericvn, line 26:
```suggestion
  to the client sidecar causing client side metrics to have server workload metadata labels
```

ericvn (commented):
Trying to fix the grammar a little. Maybe something like this.

content/en/faq/metrics-and-logs/telemetry-v1-vs-v2.md
rvennam: I suggest removing "Starting with the Istio 1.8 release, a solution was added" as I don't believe the history is important...especially in future releases. 

When the server workload is out of the mesh, server workload metadata is distributed to client sidecar, causing client side metrics to have server workload metadata labels filled.

bianpengyuan: Updated to remove 1.8 release. thanks!

rvennam (approved):

PR #9023, opened 2021-02-22
download updates

xichengliudui:
Fixed: #9024


craigbox:
Hi xichengliudui,

This script already downloads the latest release - I just tested it now and it pulls 1.9.0.
This is correct, as when we update to 1.10 for example, we don't want to have to open the docs.

I'm closing this PR and will reply on the bug to figure out what is happening.

PR #9025, opened 2021-02-22
Add blog for zero configuration Istio

howardjohn:
The intent here is to show off what Istio provides out of the box, to
attempt to counteract some of the reputation Istio has gotten for being
over complicated/requiring too many CRDs.

### DIFF
--- a/content/en/blog/2021/life-without-crds/index.md
+++ b/content/en/blog/2021/life-without-crds/index.md
@@ -34,9 +34,9 @@ Istio automatically generates detailed telemetry for all service communications 
 This telemetry provides observability of service behavior, empowering operators to troubleshoot, maintain, and optimize their applications – without imposing any additional burdens on service developers.
 Through Istio, operators gain a thorough understanding of how monitored services are interacting, both with other services and with the Istio components themselves.
 
-All of this functionality is added by Istio without any configuration. [Integrations](/latest/docs/ops/integrations/) with tools such as Prometheus, Grafana, Jaeger, Zipkin, and Kiali are also available.
+All of this functionality is added by Istio without any configuration. [Integrations](/docs/ops/integrations/) with tools such as Prometheus, Grafana, Jaeger, Zipkin, and Kiali are also available.
 
-For more information about the observability Istio provides, check out the [observability overview](/latest/docs/concepts/observability/).
+For more information about the observability Istio provides, check out the [observability overview](/docs/concepts/observability/).
 
 ## Traffic Management
 
--- a/content/en/blog/2021/life-without-crds/index.md
+++ b/content/en/blog/2021/life-without-crds/index.md
@@ -1,0 +1,81 @@
+---
+title: "Life Without CRDs"
+description: Understanding the benefits Istio brings, even when no configuration is used.
+publishdate: 2021-02-25
+attribution: "John Howard (Google)"
+---
+
+Often times when a new user encounters Istio for the first time, they are overwhelmed by the vast feature
+set exposed by Istio. Unfortunately, this occasionally gives the impression that Istio is needlessly complex,
+and not fit for small teams or clusters.
+
+One great part about Istio, however, is that it aims to bring as much value to users out of the box, without any configuration at all.
+This enables users to get most of the benefits of Istio with minimal efforts. For some users with simple requirements, custom configurations
+may never be required at all, while others will be able to incrementally add Istio configurations as they need them, such as to add
+ingress routing, fine-tune networking settings, or lock down security policies.
+
+## Getting started
+
+To get started, check out our [getting started](/docs/setup/getting-started/) documentation, where you will learn how to install Istio.
+If you are already familiar, you can simply run `istioctl install`.
+
+Next, we will explore all the benefits Istio provides us, without configuration.
+
+## Security
+
+Istio automatically enables [mutual TLS](/docs/concepts/security/#mutual-tls-authentication) for traffic between pods in the mesh.
+This enables applications to forgo complex TLS configuration and certificate management, and offload all transport layer security to the sidecar.
+
+Once comfortable with automatic TLS, you may choose to [allow only mTLS traffic](/docs/tasks/security/authentication/mtls-migration/), or configure custom [authorization policies](/docs/tasks/security/authorization/) for your needs.
+
+## Observability
+
+Istio automatically generates detailed telemetry for all service communications within a mesh.
+This telemetry provides observability of service behavior, empowering operators to troubleshoot, maintain, and optimize their applications – without imposing any additional burdens on service developers.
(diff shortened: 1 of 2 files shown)


howardjohn:
Not sure why I get 404s, those links look valid to me

ericvn:
@howardjohn Remove the `latest/` in the failing links.

content/en/blog/2021/life-without-crds/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "Life Without CRDs"
+description: Understanding the benefits Istio brings, even when no configuration is used.
+publishdate: 2021-02-25
+attribution: "John Howard (Google)"
+---
+
+Often times when a new user encounters Istio for the first time, they are overwhelmed by the vast feature
+set exposed by Istio. Unfortunately, this occasionally gives the impression that Istio is needlessly complex,
```
ericvn, line 9:
```suggestion
set exposed by Istio. Unfortunately, this occasionally gives the impression that Istio is needlessly complex
```

content/en/blog/2021/life-without-crds/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "Life Without CRDs"
```
smawson, line 2: I would call this something like "Zero Config Istio"

The CRDs are all still there :)

content/en/blog/2021/life-without-crds/index.md
```diff
@@ -0,0 +1,81 @@
...
+description: Understanding the benefits Istio brings, even when no configuration is used.
+publishdate: 2021-02-25
+attribution: "John Howard (Google)"
+---
+
+Often times when a new user encounters Istio for the first time, they are overwhelmed by the vast feature
+set exposed by Istio. Unfortunately, this occasionally gives the impression that Istio is needlessly complex,
+and not fit for small teams or clusters.
+
+One great part about Istio, however, is that it aims to bring as much value to users out of the box, without any configuration at all.
```
ericvn, line 12:
```suggestion
One great part about Istio, however, is that it aims to bring as much value to users out of the box without any configuration at all.
```

smawson (commented):
This is a great idea, love it.

content/en/blog/2021/life-without-crds/index.md
```diff
@@ -0,0 +1,81 @@
...
+This enables users to get most of the benefits of Istio with minimal efforts. For some users with simple requirements, custom configurations
+may never be required at all, while others will be able to incrementally add Istio configurations as they need them, such as to add
```
ericvn, line 14:
```suggestion
may never be required at all. While others will be able to incrementally add Istio configurations as they need them, such as to add
```

content/en/blog/2021/life-without-crds/index.md
```diff
@@ -0,0 +1,81 @@
...
+This telemetry provides observability of service behavior, empowering operators to troubleshoot, maintain, and optimize their applications – without imposing any additional burdens on service developers.
+Through Istio, operators gain a thorough understanding of how monitored services are interacting, both with other services and with the Istio components themselves.
+
+All of this functionality is added by Istio without any configuration. [Integrations](/docs/ops/integrations/) with tools such as Prometheus, Grafana, Jaeger, Zipkin, and Kiali are also available.
+
+For more information about the observability Istio provides, check out the [observability overview](/docs/concepts/observability/).
+
+## Traffic Management
+
+While Kubernetes provides a lot of networking functionality, such as service discovery and DNS, because this is done at the L4 connection level we lose a lot of potential functionality.
```
ericvn, line 43:
```suggestion
While Kubernetes provides a lot of networking functionality, such as service discovery and DNS, this is done at the L4 connection level and we lose a lot of potential functionality.
```

content/en/blog/2021/life-without-crds/index.md
howardjohn: I was actually planning to not install the CRDs, but unfortunately telemetry would break so its a bit awkward. I think we could actually not setup the CRDs, and exclude envoyfilter/wait for Telemetry API to land, or change the title. 

I think changing the title is probably for the best, unless people actually think CRD == complex, rather than using CRDs as a proxy for complexity?

smawson: Yeah I think people use CRDs as a proxy for complexity, so changing the title is good. If we don't install the CRDs then there is friction to go from zero-config to first-config, and we don't want that IMO. It is really about what they are forced to learn about in order to get their job done.

linsun: +1 changing makes sense.  :) The key is no change to their code and apply Istio resources as needed incrementally.

howardjohn:
Addressed all the comments, thanks everyone

content/en/blog/2021/zero-config-istio/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+One great part about Istio, however, is that it aims to bring as much value to users out of the box without any configuration at all.
+This enables users to get most of the benefits of Istio with minimal efforts. For some users with simple requirements, custom configurations
+may never be required at all. Others will be able to incrementally add Istio configurations once they are more comfortable and as they need them, such as to add
+ingress routing, fine-tune networking settings, or lock down security policies.
+
+## Getting started
+
+To get started, check out our [getting started](/docs/setup/getting-started/) documentation, where you will learn how to install Istio.
+If you are already familiar, you can simply run `istioctl install`.
```
rvennam, line 20:
```suggestion
Essentially, you can simply run `istioctl install`, label your namespace and deploy your application pods.
```

content/en/blog/2021/zero-config-istio/index.md
```diff
@@ -0,0 +1,81 @@
...
+Hostname=echo-cb96f8d94-rpfqz
+Hostname=echo-cb96f8d94-cgmxr
+Hostname=echo-cb96f8d94-wf4xk
+Hostname=echo-cb96f8d94-rpfqz
+Hostname=echo-cb96f8d94-cgmxr
+{{< /text >}}
+
+Here we can see our requests are [round-robin](/docs/concepts/traffic-management/#load-balancing-options) load balanced between all backends.
+
+In addition to these better defaults, Istio offers customization of a [variety of traffic management settings](/docs/concepts/traffic-management/), including timeouts, retries, and much more.
```
rvennam, line 81: Are there any out-of-the-box default retry logic worth mentioning here?

rvennam (approved):

craigbox:
Did we hold off cherrypicking this for a reason?  Needs to go to 1.9 to be published.

howardjohn:
/cherrypick release-1.9

Nope, not intentionally on my part at least

PR #9046, opened 2021-02-23
Fix istio.io tests when moving to later kubectl

ericvn:
The CI pipeline moved to a later build image which contains an updated kubectl which seems to be causing some issue with the doc tests.  I am working on fixing them up with this PR.


howardjohn (approved):
possibly safe is to add `kubectl wait --for condition=Established CRD/something`. Syntax is not exactly right :slightly_smiling_face:

PR #9049, opened 2021-02-24
Ignore some errors

ericvn:
Moving the CI pipeline to a later version of `kubectl` causes some failures which didn't occur before. We need to ignore the now `expected` failures to let the tests continue.

### DIFF
--- a/content/en/docs/setup/additional-setup/external-controlplane/test.sh
+++ b/content/en/docs/setup/additional-setup/external-controlplane/test.sh
@@ -78,7 +78,10 @@ export EXTERNAL_ISTIOD_ADDR=$(kubectl \
     -n istio-system get svc istio-ingressgateway \
     -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 snip_set_up_the_remote_cluster_1_modified
+
+set +e #ignore failures here
 echo y | snip_set_up_the_remote_cluster_2
+set -e
 
 # Install istiod on the external cluster.
 


rvennam (approved):

content/en/docs/setup/additional-setup/external-controlplane/test.sh
```diff
@@ -78,7 +78,10 @@ export EXTERNAL_ISTIOD_ADDR=$(kubectl \
     -n istio-system get svc istio-ingressgateway \
     -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 snip_set_up_the_remote_cluster_1_modified
+
+set +e #ignore failures here
 echo y | snip_set_up_the_remote_cluster_2
```
frankbu, line 83: What is the `expected` error here? This snip is installing the ingress gateway on the cluster, so I can't see how the following validation part of this test will succeed, if the install fails here.

frankbu: Oh, I see. The install worked, just spits out some errors about EnvoyFilter.
```
service/istio-ingressgateway created
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
unable to recognize "STDIN": no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
```
Any idea what's causing the messages?

ericvn: @frankbu and I haven't confirmed if the messages existed with the older Kubectl (just need to look back at older passing tests), and if maybe Kubectl was just passing back an error code now...

I am still concerned that there is a failure in what's deployed and the test isn't actually complete enough to test the `remote` cluster.

In Slack, John asked about changing to `istioctl install` (see another PR where I tried that) and that also has error messages and the gateway seems to fail. Overall the tests pass if I skip failing on the error code there.

I'm hoping to see some feedback on if the test is actually accurate in that PR.

PR #9050, opened 2021-02-24
Rename dir name metadata_exchange.html. to metadata_exchange

hzxuzhonghu:
Windows does not support `metadata_exchange.html.`

```
git clone https://github.com/istio/istio.io.git

Cloning into 'istio.io'...


remote: Enumerating objects: 126, done.
remote: Counting objects: 100% (126/126), done.
remote: Compressing objects: 100% (88/88), done.
remote: Total 113481 (delta 39), reused 72 (delta 21), pack-reused 113355
Receiving objects: 100% (113481/113481), 119.00 MiB | 3.39 MiB/s, done.
Resolving deltas: 100% (66853/66853), done.
error: invalid path 'archive/v1.6/docs/reference/config/proxy_extensions/metadata_exchange.html./index.html'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```


ericvn:
The precursor to this problem is actually this line having a period on the end:

https://github.com/istio/proxy/blob/release-1.6/extensions/metadata_exchange/config.pb.html#L4

This change could be overwritten, if we ever ran an update_ref_docs on the release-1.6 branch and recreated the archive for 1.6 changes. I think that is pretty unlikely, and there is a bunch of work needed (including a release-1.6 branch commit in istio/proxy) to fix this correctly, so this change will most likely endure.

hzxuzhonghu:
How could we fix it permanently?

ericvn:
Permanent fix would be to update the istio/proxy file in the release-1.6 branch. Then in the release-1.6 branch here do a update-ref-docs to pull that change into this repo (which is basically what you have done, and would be overwritten if we would ever do a update-ref-docs on 1.6 (not likely)), and then run the command such that we update the archive directory in 1.9 with the new changes from 1.6. That still leaves the subdirectory with the trailing `.` in the archives of the other branches, so should it be pulled there as well?

Would something ever be severe enough to cause us to do all this? I don't think so. So making a change, without changing the upstream is most likely fine.

It probably makes sense to carry this change to there release-1.9 branch since we still actively build there.

ericvn:
A few notes:  I did make the same change in the release-1.6 branch and ran `make lint` to see if there was any issues and it passed. I also did a `make serve` and was able to get to the page.

Using the netlify, test I tried to see if this change was OK, at it appears that things will maybe break. The `https://deploy-preview-9050--preliminary-istio.netlify.app/v1.6/docs/reference/config/proxy_extensions/` page still has the trailing `.` on the link so clicking the link will 404.

ericvn:
@davidhauck See my prior remark about the netlify task still having an old pointer after the change in this PR. Is that expected? What else needs to be done to make things work?

davidhauck:
Remember that the archive is pure html, so changing the name of a file won't update any links to it. This means that we will have to update all links to it manually, e.g. https://github.com/istio/istio.io/blob/master/archive/v1.6/docs/reference/config/proxy_extensions/index.html

Like you said, the 'correct' way to do this would be to update istio/proxy on release-1.6, then build an updated version of release-1.6 here. I don't think that should be too difficult?

hzxuzhonghu:
I am super confused, could you help

ericvn:
@hzxuzhonghu 

As David suggests, we should follow a process similar to what I laid out above.

1. Please create a PR fixing https://github.com/istio/proxy/blob/release-1.6/extensions/metadata_exchange/config.pb.html#L4 (by removing the trailing period). This PR will need to be approved by one of the release managers. 
2. Once that it done, someone from the Docs team (probably David or myself) will need to do a few steps to get the correction moved from the proxy repo to the release-1.6 branch here, and then that re-archived into the master and release-1.9 branches.

ericvn:
Fixed in #9179 via instructions above.

hzxuzhonghu:
Thanks @ericvn

PR #9067, opened 2021-02-25
[release-1.9] Bump basic auth 1.9 Wasm extension module to 1.9.2

bianpengyuan:
Dump basic auth Wasm module version again to 1.9.2 to include hosts option: https://github.com/istio-ecosystem/wasm-extensions/issues/50


rvennam:
Do we have to update this every patch release? @ericvn

bianpengyuan:
@rvennam Yeah, I think so, but this extension should have pretty much no change in the future, and I would expect that it does not need patch releases once stable. It churned a bit recently just because it is new and some basic functions were just added according to user's feedback.

ericvn (approved):

PR #9070, opened 2021-02-26
Polidea update

wbartosz:
This pull request includes About Ecosystem page.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/about/ecosystem/index.md
+++ b/content/en/about/ecosystem/index.md
@@ -1,0 +1,46 @@
+---
+title: Ecosystem
+description: Ecosystem.
+subtitle: "Istio is a service mesh that provides cross-cutting functions that all micro services environments need. You can get Istio from 27 distributors"
+weight: 34
+skip_toc: true
+skip_byline: true
+skip_pagenav: true
+doc_type: about
+---
+
+{{< tabset category-name="ecosystem-type" class="tabset--ecosystem" >}}
+
+    {{< tab
+        name="distributors"
+        category-value="default"
+        description="Locality-weighted load balancing allows administrators to control the distribution of traffic to endpoints based on the localities of where the traffic originates and where it will terminate."
+    >}}
+
+    {{< distributors >}}
+
+    {{< /tab >}}
+
+    {{< tab
+        name="pro services" 
+        category-value="pro_services"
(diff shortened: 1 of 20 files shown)


content/en/about/ecosystem/index.md
```diff
@@ -0,0 +1,46 @@
+---
```
davidhauck, line 1: Can we add a todo on this page as well

layouts/shortcodes/interactive_panels.html
```diff
@@ -0,0 +1,10 @@
+{{ $firstItems := .Get "first" }}
+{{ $items := cond (eq (.Get "items") "integrations") .Site.Data.companies.integrations .Site.Data.companies.pro_services }}
```
davidhauck, line 2: Is there a reason we're passing in what's essentially a bool and then loading them? It seems like it would be better to just pass in the list itself

davidhauck (dismissed):

content/en/about/ecosystem/index.md
wbartosz: Done.
Added an extra one in the `data/companies.yml` because it also contains a list of placeholder distributors, integrations and pro services.

layouts/shortcodes/interactive_panels.html
wbartosz: I have replaced the bool with an actual list item in the `.Site.Data.companies` file.

davidhauck (commented):
LGTM, @craigbox whenever you want feel free to approve this if it looks good to you. Small comment

layouts/shortcodes/interactive_panels.html
```diff
@@ -0,0 +1,10 @@
+{{ $firstItems := .Get "first" }}
```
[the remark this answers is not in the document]
wbartosz: It's not. The idea was to use it in similar way to distributors where we could define how many maximum elements we want to list. That component isn't really used anywhere else apart from Ecosystem.

I will get rid of that variable.

craigbox (approved):

PR #9072, opened 2021-02-26
fix workaround for revision validation bug

Monkeyanator:
Previous documented workaround was flawed for users on in-cluster operator (per [this comment](https://github.com/istio/istio/pull/30286#issuecomment-786595621)). This removes the incorrect workaround and gives a code snippet for users to help with the other workaround

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/upgrade/canary/index.md
+++ b/content/en/docs/setup/upgrade/canary/index.md
@@ -57,7 +57,7 @@ istio-sidecar-injector-canary   1          3m18s
 {{< /text >}}
 
 {{< warning >}}
-Due to [a bug](https://github.com/istio/istio/issues/28880) in the creation of the `ValidatingWebhookConfiguration` during install, initial installations of Istio __must not__ specify a revision. As a temporary workaround, for Istio resource validation to continue working after removing the non-revisioned Istio installation, the `ValidatingWebhookConfiguration` must be manually pointed to the revision that should handle validation.
+Due to [a bug](https://github.com/istio/istio/issues/28880) in the creation of the `ValidatingWebhookConfiguration` during install, initial installations of Istio __must not__ specify a revision. As a temporary workaround, for Istio resource validation to continue working after removing the non-revisioned Istio installation, the `istiod` service must be manually pointed to the revision that should handle validation.
 
 One way to accomplish this is to manually create a service called `istiod` pointing to the target revision using [this service]({{< github_blob >}}/manifests/charts/istio-control/istio-discovery/templates/service.yaml) as a template. Another option is to run the command below, where `<REVISION>` is the name of the revision that should handle validation. This command creates an `istiod` service pointed to the target revision.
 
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/upgrade/canary/index.md
```diff
@@ -57,11 +57,14 @@ istio-sidecar-injector-canary   1          3m18s
 {{< /text >}}
 
 {{< warning >}}
-Due to [a bug](https://github.com/istio/istio/issues/28880) in the creation of the `ValidatingWebhookConfiguration` during install, initial installations of Istio __must not__ specify a revision. As a temporary workaround, for Istio resource validation to continue working after removing the non-revisioned Istio installation, the `ValidatingWebhookConfiguration` must be manually pointed to the revision that should handle validation.
+Due to [a bug](https://github.com/istio/istio/issues/28880) in the creation of the `ValidatingWebhookConfiguration` during install, initial installations of Istio __must not__ specify a revision. As a temporary workaround, for Istio resource validation to continue working after removing all revisioned Istio installations, the `istiod` service must be manually pointed to the revision that should handle validation.
```
smawson, line 60: I think the text is wrong here, the workaround is for after you remove the non-revisioned install, not when you remove revisioned installs.

Monkeyanator: Good catch thanks, the phrasing was indeed backwards. Fixed now

content/en/docs/setup/upgrade/canary/index.md
```diff
@@ -59,9 +59,12 @@ istio-sidecar-injector-canary   1          3m18s
 {{< warning >}}
 Due to [a bug](https://github.com/istio/istio/issues/28880) in the creation of the `ValidatingWebhookConfiguration` during install, initial installations of Istio __must not__ specify a revision. As a temporary workaround, for Istio resource validation to continue working after removing the non-revisioned Istio installation, the `ValidatingWebhookConfiguration` must be manually pointed to the revision that should handle validation.
```
smawson, line 60: I liked the other part of the change here though, telling the user they need to create an istiod service, I thought that was the point of the bug? It isn't enough just to change ValidatingWebhookConfiguration, you need to create a revision-less istiod service and point it the right place?

smawson: Suggested text:

As a temporary workaround, for Istio resource validation to continue working after removing the non-revisioned Istio installation, the `istiod` service must be manually pointed to the revision that should handle validation.

Monkeyanator: Yup, makes sense, changed. Wasn't sure if users would see the connection between the `ValidatingWebhook` and the `istiod` service but it's probably better we don't mention "modifying where the webhook points" at all.

smawson (approved):

PR #9126, opened 2021-03-06
Link Wasm blog post from the Wasm remote fetch task.

bianpengyuan:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
+++ b/content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
@@ -107,6 +107,8 @@ spec:
 {{< /text >}}
 
 The Istio agent will only intercept and download remote Wasm modules configured via ECDS resources.
+You can find more technical details about this feature in this [blog post](/blog/2021/wasm-progress/#wasm-module-distribution-via-the-istio-agent).
+
 This feature is enabled by default.
 To disable ECDS interception and Wasm downloading in the Istio agent, set the `ISTIO_AGENT_ENABLE_WASM_REMOTE_LOAD_CONVERSION` environment variable to `false`.
 For example, to set it globally:


content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
```diff
@@ -107,6 +107,8 @@ spec:
 {{< /text >}}
 
 The Istio agent will only intercept and download remote Wasm modules configured via ECDS resources.
+You can find more technical details about this feature in this [blog post](/blog/2021/wasm-progress/#wasm-module-distribution-via-the-istio-agent).
```
ericvn, line 110: I'm not sure if we typically have pointers to blogs in the regular docs, at least I typically don't see them.  I'll leave the pointer to blog up to @frankbu or @craigbox.

craigbox: I would think that if it was worth documenting, it should be documented.  Frank?

ericvn: I think what you are saying is the if it's mentioned in the docs, we should place it in the docs, and that content is longer lived. So, in this case, I would think we want to convert/merge the blog into the docs. Is that correct? I believe that what Frank used to say, and what I expected him to come back with for a comment. But I wanted to make sure.

bianpengyuan: Is the concern that blog post is likely to be out of dated?

ericvn: A blog post is something that goes out of date, or becomes stale and shouldn't be in the docs (at least that they way I've been thought). Docs are something that remain consistent over a longer time.

My concern here that the we are adding something in the longer term docs that is pointing to a shorter term blog. This implies we would need to do a update in the short term to the docs to remove the blog link when it becomes non-relevent. That's what I am trying to avoid, and why I asked if we put links in long term docs to short term blogs.

If the blog is really a long term item, we should have it in the docs instead of as a blog (so suggestion would be to move it there). In that case, a link makes sense since its a link in a long term document to a long term document. 

Does that answer your question?

bianpengyuan: Thanks for the detailed explanation! I think it makes sense to not mix short term snapshot of the state (blog) with an evolving doc. I will think about merging part of the blog post into the existing doc. Thanks!

PR #9151, opened 2021-03-08
Fix typo: inherite -> inherit

Sovietaced:
[ X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/zh/docs/reference/config/security/peer_authentication/index.html
+++ b/content/zh/docs/reference/config/security/peer_authentication/index.html
@@ -74,7 +74,7 @@ spec:
       mode: DISABLE
 </code></pre>
 
-<p>Policy to inherite mTLS mode from namespace (or mesh) settings, and overwrite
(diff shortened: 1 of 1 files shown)


ericvn:
Note that this change is in a file that is generated elsewhere and pulled into istio.io. 

A `make update-ref-docs` should probably be run against the Chinese version of the docs (not sure if that's possible with the scripts), or all the reference files should be updated from their English counterparts.

ericvn:
It does look like one could change `scripts/grab_reference_docs.sh` replacing /en/ with another language (maybe set via an env var).

Sovietaced:
> Note that this change is in a file that is generated elsewhere and pulled into istio.io.

Oops sorry! Looks like the typo was actually fixed ~18 days ago.

PR #9157, opened 2021-03-09
Update index.md

zackzhangkai:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/platform-setup/kubesphere/index.md
+++ b/content/en/docs/setup/platform-setup/kubesphere/index.md
@@ -8,7 +8,7 @@ owner: istio/wg-environments-maintainers
 test: no
 ---
 
-This page was last updated February 17, 2018.
+This page was last updated March 9, 2021.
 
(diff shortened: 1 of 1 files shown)


zackzhangkai:
I'm a member of istio , in the fact. My github name is already in the community doc. https://github.com/istio/community/blob/master/org/members.yaml#L72
But why does nobody invite me to the istio org?

craigbox (approved):

craigbox:
> I'm a member of istio , in the fact. My github name is already in the community doc. https://github.com/istio/community/blob/master/org/members.yaml#L72
> But why does nobody invite me to the istio org?

I think you might have to be in [developers.yaml](https://github.com/istio/community/blob/master/org/developers.yaml) instead?

I think we are looking to merge these two soon. I'll follow up.

Issue #9154, opened 2021-03-09
Can't checkout the code in Windows platform.


elkan1788:
Hi ,

There I want to commit my Chinese translate content, but when I fork this repository and git clone to my computer which OS is windows, Then I meet the error that see this below:

```sh
"git" clone -v --recurse-submodules --progress "git@github.com:elkan1788/istio.io.git" "D:/GitRepos/istio.io"
Cloning into 'D:/GitRepos/istio.io'...
remote: Enumerating objects: 11, done.
remote: Counting objects:  36% (4/11)eceiving objects:   0% (1/123878)remote: Counting objects: 100% (11/11), done.
remote: Compressing objects: 100% (10/10), done.
remote: Total 123878 (delta 0), reused 5 (delta 0), pack-reused 123867R
Receiving objects: 100% (123878/123878), 129.39 MiB | 993.00 KiB/s, done.
Resolving deltas: 100% (72543/72543), done.
error: invalid path 'archive/v1.6/docs/reference/config/proxy_extensions/metadata_exchange.html./index.html'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'

Done

Press Enter or Esc to close console...
```

I try to checkout from HEAD that the error was happened again.  So is there anyone see this error same with me, please help me what is wrong with me and fix it ,  Thanks.

ericvn:
See #9050 which was the original PR to fix the issue. The problem is that the directory name is specified upstream, so we are working through the process to get that changed (done), the 1.6 branch to be update (PR waiting to be merged) and then rebuild the archive from 1.6 in the main branch.

elkan1788:
@ericvn  I waiting for this PR merge to do my translate work, hope there can get good news form this issue #9050 , Thank you.

ericvn:
@elkan1788 I had asked for someone to approve my PR yesterday and again today  so that I can continue with the next PR which once merged should fix the issue. At the mercy of maintainers to get things in at the moment.

elkan1788:
@ericvn Thanks for your work. I will keep focus on it .

ericvn:
Still waiting on approval for #9179. David said he would approve, but haven't seen that yet. Hopefully soon.

elkan1788:
@ericvn  I see the code had change, Thanks a lot.

PR #9179, opened 2021-03-10
Build a new 1.6 archive to remove trailing . in directory name

ericvn:
Fixes: https://github.com/istio/istio.io/issues/9154

Generated via `make build-old-archive-1.6.0`

@davidhauck Do all these changes make sense for doing that?

The key change `...nsions/metadata_exchange.html./index.html → ...y_extensions/metadata_exchange/index.html ` removing the trailing `.` is included.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
This should fix the cloning problem for the main branch. I'm not sure what happens if thy check out one of the other branches, like release-19, which will still contain the directory in the 1.6 archive.

davidhauck:
So the process to generate this was to first update istio/api to fix the link, and then we re-ran the generate command here which detected that change?

ericvn:
@davidhauck 

First PR was https://github.com/istio/proxy/pull/3214. This removed the trailing period on the location line of the pb.html so the file should be copied to a directory without the `.`

Second PR, https://github.com/istio/istio.io/pull/9166, was to run the istio.io command to actually pull the changes from the other repos. I was hoping there would only be the one file change, but there were a couple command updates as well.

Third PR is this one. Ran the command to build the new 1.6 archive here. I do notice the directory name changed in several places, so we certainly needed to update more than the the directory name as was attempted in a few other PRs.

ericvn:
I checked a dozen files or so, and for those, the change was something like:
```
href=/v1.6/docs/reference/config/proxy_extensions/metadata_exchange.html./>Metadata Exchange Config</a></li>
href=/v1.6/docs/reference/config/proxy_extensions/metadata_exchange/>Metadata Exchange Config</a></li>
```
within the search button?

davidhauck:
Got it, LGTM

ericvn:
One thing I noticed in the second PR above is that the html also got removed from the directory name. Not sure which part of the code did that. But it matches the change I see in a few of the files..
```
...nsions/metadata_exchange.html./index.html → ...y_extensions/metadata_exchange/index.html
```

ericvn:
I did move around the new 1.6 archive, looking specifically at the page with the new URL and well as a bunch of others, and things look ok: https://deploy-preview-9179--preliminary-istio.netlify.app/v1.6/docs/reference/config/proxy_extensions/metadata_exchange/ 

I don't think we need to cherry pick this change back to release-1.9, so there will be no change to the .6 archive there.

ericvn:
/test doc.test.profile_none_istio.io

as there was VM test failure (I think that one is a much less frequent flake).

ericvn:
The two failing tests are known failures until https://github.com/istio/istio/issues/31317 is fixed.

davidhauck (approved):

PR #9189, opened 2021-03-11
Polidea update - Website docs

agatastpa:
This PR contains website docs - instructions how to contribute to Istio page.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ see the Istio [contribution guidelines](https://github.com/istio/community/blob/
 - [Testing document content](#testing-document-content)
 - [Multi-language support](#multi-language-support)
 - [Regular maintenance](#regular-maintenance)
+- [Website documentation](./WEBSITE.md)
 
 ## Editing and building
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ see the Istio [contribution guidelines](https://github.com/istio/community/blob/
 - [Testing document content](#testing-document-content)
 - [Multi-language support](#multi-language-support)
 - [Regular maintenance](#regular-maintenance)
-- [Website documentation](./WEBSITE.md)
+- [Custom Pages](./CUSTOM_PAGES.md)
 
 ## Editing and building
 
--- a/CUSTOM_PAGES.md
+++ b/CUSTOM_PAGES.md
@@ -1,5 +1,5 @@
-# Contributing to the Istio website
-Here are some instructions how you can contribute to the [Istio website](https://www.istio.io).
+# Custom Pages
+Here are some instructions how you can customize pages of the [Istio website](https://www.istio.io).
 
 ## Pages
 1. [Homepage](##homepage)
--- a/WEBSITE.md
+++ b/WEBSITE.md
@@ -1,0 +1,141 @@
+# Contributing to the Istio website
+Here are some instructions how you can contribute to the [Istio website](https://www.istio.io).
+
+## Pages
+1. [Homepage](##homepage)
+2. [About](##about)
+3. [FAQ](##faq)
+4. [Blog](##blog)
+5. [Get involved](##get-involved)
+6. [Documentation](##documentation)
+7. [Info banner](##info-banner)
+
+## Homepage
+Homepage landing panels are created using the ```{{< content_panel >}}``` shortcode providing the `type` of the panel, `title`, `text`, `button` and `url` (example below).
+```
+{{< content_panel type="dark" title="service_mesh" text="You can find what you need to make Istio do exactly what you need it to do." button="learn_more" url="/about/service-mesh" >}}
+```
+
+### Concepts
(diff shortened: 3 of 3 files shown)


craigbox:
LGTM, defer to David

aijamalnk:
Thank you @agatastpa !

Could we please mention and link to this file in the README so it's easily disoverable?

README.md
```diff
@@ -23,6 +23,7 @@ see the Istio [contribution guidelines](https://github.com/istio/community/blob/
 - [Testing document content](#testing-document-content)
 - [Multi-language support](#multi-language-support)
 - [Regular maintenance](#regular-maintenance)
+- [Website documentation](./WEBSITE.md)
```
davidhauck, line 26: This is all documentation for the "website," so I'm not sure that's the best name. Would a better place for this be here? https://istio.io/latest/about/contribute/

ericvn: What happening to all the documentation at https://istio.io/latest/about/contribute/? How should contributors find that going forward?

craigbox: I would propose moving that content to GitHub.  Contributor docs for code are on GitHub; contributor docs for the web should be there too. If the team wanted to keep this on the site, it would need to move under Docs.

aijamalnk: Perhaps we could call the link on README.md 'Changing and deploying the website'? I also like having it in GitHub. Many projects have instructions to change and set up a local environment in their readmes/similar files as the last section of the README.

craigbox:
We'll need to think about where the other assorted content from https://istio.io/latest/about/ goes - I think that https://istio.io/latest/about/contribute/ could go to GitHub as MD files, and this file would go with that.

I think that Feature Status, Reporting Bugs, Security Vulnerabilities etc should probably become headlines in Docs, not About. @ericvn, would you have any objections?  Should I take this to Docs WG next week?

ericvn:
I do notice that there is no way to get at the pages like https://istio.io/latest/docs/concepts/ at all in the new layout.  In the old, you could use the link to get back to it. For example, to get to the above, go to https://istio.io/latest/docs/concepts/what-is-istio/ and click on `concepts.`

ericvn:
@craigbox I'm not to worried about where the about stuff lands, but several I think are pretty popular (I don't have the number of hits per page, but I hit them a lot). Feature Status and Supported Releases are two. 

Doc contribution could be moved to GH or a wiki. We use a wiki for code development, so using one for doc development would be fine. Again, we just need an easy way for people to find it.

As for issue creation, I would really just like to see a good way to direct people to the correct place. Maybe we only take them in istio/istio (disable elsewhere) and work on the templates. I think it funny that we have istio operation issues being written in istio/istio.io and istio/community for example. We should be able to better direct users on issue creation.

aijamalnk:
I see we have docs contribution guide in the website itself. If we were to put this content in the website, since the whole PR is mainly an overview of the architecture of each page of the website, we could add a new section called "Architecture of the Istio site" under the Contributing Documentation section(https://istio.io/latest/about/contribute/). WDYT?

craigbox:
The plan of record is to move the "how to edit the website" content out of the "advocacy" half of the website.  It can go to GitHub (my preference) or it can go to /docs.  "how to edit the code" content isn't on istio.io, it's on GitHub.

We'll need to ensure all pages are redirected, especially "Feature Status" and "Supported Releases", but they are not particularly discoverable in today's design so I'm not worried about that getting worse in the new world.

aijamalnk:
@craigbox - as @davidhauck mentioned, currently there are plenty of documentation on changing and updating the website. And I agree with David that these docs should be added to the current docs. My suggestions is that we add the whole PR as a new section called "Architecture of the Istio site" under the [Contributing Documentation section](https://istio.io/latest/about/contribute/) as it is mainly an overview of the architecture of each page of the website. 

If you want the "contribute" docs out of About section, this should be discussed with Docs WG and agreed with the rest of the group on where to put and access them in the future. 

These discussions however shouldn't block this PR. This is just additional info on the site and nothing else.

craigbox:
I agree the content should all live together.  I just want to point out
that if we do nothing, content under "about" in the old design and not
under "about" in the new design, may no longer be accessible.

>

davidhauck:
The easier way forward might be to simply move all of the current contributing documents to markdown files that can be viewed in github. That should really only take a few minutes to move them. The current contribute docs get <50 page views per week. If we wanted to ensure people don't get lost, we could add a small page at /about/contribute that redirects users to the github pages.

aijamalnk:
Okay - @agatastpa would you be able to move existing Website contribution docs under About to `.md` files on Github as per @davidhauck 's comment?

davidhauck (approved):

PR #9190, opened 2021-03-11
Polidea update - News

Jakub-Sadowski:
This PR contains: news list page and single news page

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure


ericvn:
I notice the Release Announcements is now a list of all the announcements and not by release. Makes it hard to find if you are looking for a particular release.

brian-avery:
Buttons don't render here:

image

Also, I know it's part of the design, but the previous security listing allowed users to look at the table and see if a CVE affected their release and what it's score is. This is now missing from the table. Can we add it back? Without it, users have to dive into each security issue.

craigbox:
We didn't actually do a design for News explicitly; we just made it look
like the blog.  Maybe it might suit the docs layout better?

>

brian-avery:
Maybe. I think the stream flow works for this. I like the idea of presenting everything in a stream that was presented here. All three are related and being able to scroll through all of them in cronological order makes sense to me. But I do think we need to have a way to present the CVE details on that first page. Could do this as a second line, could do this as labels. Could do a table as we did previously. Not sure what allows a user to most easily identify the important details:
*  How severe is this CVE
* Does it affect me?

aijamalnk:
- The selection works in a strange manner. Try the following:
    1. Select release announcements
    2. Select security bulletins
    3. Click on security bulletins again
    - Outcome: All buttons become white. Every category is selected.
    - Correct outcome: Security bulletins is unselected. Only release announcements is selected.

**Perhaps we should only allow users to select one item at a time**. Selecting item X unselects other items. When first entering the page, everything shows up. To have everything again, click on your current selection.

As for security bulletins - **could we show the [Disclosure Details](https://user-images.githubusercontent.com/14306210/111111844-51082980-851c-11eb-8e93-833086219dd1.png
) in the [post snippet](https://user-images.githubusercontent.com/14306210/111111969-8c0a5d00-851c-11eb-88bc-be52a6fca3cd.png)?**

@Jakub-Sadowski

Jakub-Sadowski:
> * The selection works in a strange manner. Try the following:
>   
>   1. Select release announcements
>   2. Select security bulletins
>   3. Click on security bulletins again
>   
>   
>   * Outcome: All buttons become white. Every category is selected.
>   * Correct outcome: Security bulletins is unselected. Only release announcements is selected.
> 
> **Perhaps we should only allow users to select one item at a time**. Selecting item X unselects other items. When first entering the page, everything shows up. To have everything again, click on your current selection.
> 
> As for security bulletins - **could we show the [Disclosure Details](https://user-images.githubusercontent.com/14306210/111111844-51082980-851c-11eb-8e93-833086219dd1.png) in the [post snippet](https://user-images.githubusercontent.com/14306210/111111969-8c0a5d00-851c-11eb-88bc-be52a6fca3cd.png)?**
> 
> @Jakub-Sadowski

@aijamalnk I just fixed the problem with multi-select, about Disclosure details I can add any field from photo below. Do you want to add just keyword in this case 'CVE' or some different information which text will be dependent on this value?
image

craigbox:
Can we have it say, on three separate lines, _if_ each value is populated?

CVE(s): $CVES
CVSS: $CVSS
Affected versions: $RELEASES

ericvn:
Selection is still broke.

aijamalnk:
Hi @ericvn, thanks for the feedback!

I think the selection is working as designed right now. When users land to the page, the content is "unflitered", therefore it shows everything, and then a user can select a particular type of news and then it shows only those. The same behavior is applied to Blog and has been approved and merged. 

Since we need to wrap up this project, can we move forward and wait for user feedback? We can create a bug to track this, and prioritize the fix if user feedback requests it.

ericvn:
@aijamalnk I think the design is broke, but that's my opinion and I will wait and see what the user feedback is.

layouts/news/list.html
```diff
@@ -0,0 +1,86 @@
...
+                            <div class="post-summary">
+                                <p>{{ $item.Params.description }}</p>
+                                {{ if $item.Params.cves }}
+                                    <strong>CVE(s)</strong>: {{ range $i, $cve := $item.Params.cves }} {{ . }} {{ if ne $i (sub (len $item.Params.cves ) 1) }} , {{ end }} {{ end }}</br>
+                                {{ end }}
+                                {{ if $item.Params.cvss }}
+                                    <strong>CVSS</strong>: {{ with $item.Params.cvss }} {{ . }} {{ end }}</br>
+                                {{ end }}
+                                {{ if  $item.Params.releases }}
+                                    <strong>Affected versions</strong> : {{ range $i, $cve := $item.Params.releases }} {{ . }} {{ if ne $i (sub (len $item.Params.releases ) 1) }} , {{ end }} {{ end }}
```
craigbox, line 48:
```suggestion
                                    <strong>Affected versions</strong>: {{ range $i, $cve := $item.Params.releases }} {{ . }} {{ if ne $i (sub (len $item.Params.releases ) 1) }}, {{ end }} {{ end }}
```

davidhauck:
Looks good to me if we are planning to wait for user feedback.

layouts/news/list.html
```diff
@@ -0,0 +1,86 @@
+{{ define "main" }}
+<main class="primary blog-content">
+    <h1>News</h1>
+    <p class="blog-description">Istio addresses the challenges developers and operators face as monolithic applications transition towards a distributed microservice architecture. </p>
```
davidhauck, line 4: Let's throw in a todo here

Jakub-Sadowski: okay, it's done

davidhauck (approved):

PR #9192, opened 2021-03-11
Update external control plane installation docs

frankbu:
1. Added explanatory text, and verification steps.
2. Restructured to separate instructions by role (mesh operator / mesh admin).
3. Uses template injection to start gateways (a mesh admin role now).
4. Include multicluster external control plane instructions.


content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -70,272 +76,443 @@ $ export REMOTE_CLUSTER_NAME=<your remote cluster name>
...
+        ingressGateways:
+        - namespace: external-istiod
+          name: istio-ingressgateway
+          enabled: true
+      values:
+        gateways:
+          istio-ingressgateway:
+            injectionTemplate: gateway
+    EOF
+    $ istioctl install -f istio-ingressgateway.yaml --context="${CTX_REMOTE_CLUSTER}"
```
frankbu, line 448: This command will currently fail. Latest master build fails to inject the gateway template because the deployment has `istio.io/rev: default` (instead of no `istio.io/rev` label), which the sidecar-injector webhook doesn't currently allow.

@Monkeyanator I understand (from @howardjohn) that you have an upcoming fix for this. Is that correct?

Monkeyanator: Yes, that's correct. I was going to hold off until the default revision work merged, but since that might be delayed a bit, I'll get an implementation up ASAP

frankbu: Great, thanks!

content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -70,272 +76,443 @@ $ export REMOTE_CLUSTER_NAME=<your remote cluster name>
...
+    {{< text bash >}}
+    $ kubectl create sa istiod-service-account -n external-istiod --context="${CTX_EXTERNAL_CLUSTER}"
+    $ istioctl x create-remote-secret \
+      --context="${CTX_REMOTE_CLUSTER}" \
+      --type=config \
+      --namespace=external-istiod | \
+      kubectl apply -f - --context="${CTX_EXTERNAL_CLUSTER}"
+    {{< /text >}}
+
+1. Before you deploy the external istiod, you need to create the root namespace on the config (remote) cluster.
```
frankbu, line 166: Would it be possible for `istioctl install` to do this automatically, just like it does for local istio-system ns, so we can eliminate this step?

content/en/docs/setup/additional-setup/external-controlplane/index.md
howardjohn: it should work with https://github.com/istio/istio/pull/31344 anyways?

frankbu: Oh, yes @howardjohn, looks like it is already working with #31344. Thanks.

ericvn (approved):

PR #9198, opened 2021-03-12
revise "Getting Started" page to be friendly for new istio users to use bookinfo sample

monsteredc:
Originally in the page `setup/Getting Started`, when deploying the sample application using 

`kubectl apply -f samples/bookinfo/platform/kube/bookinfo.yaml`

the destination rules are not applied at all, which will cause routing problems when doing other tasks of the sample application. 
In the page `examples/bookinfo application`, the destination rules are mentioned to be required for the routing rules. However, there is a tip showing that `If you installed Istio using the Getting Started instructions, you already have Bookinfo installed and you can skip these steps.`. New users may skip applying the destination rules, and get stuck for a moment.

This repo is to copy & paste the destination rule part in `examples/bookinfo application` to `setup/Getting Started` for better new user experience.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/getting-started/index.md
+++ b/content/en/docs/setup/getting-started/index.md
@@ -348,6 +348,30 @@ by viewing the Bookinfo product page using a browser.
 
 1.  Paste the output from the previous command into your web browser and confirm that the Bookinfo product page is displayed.
 
+### Apply default destination rules
+
+Before you can use Istio to control the Bookinfo version routing, you need to define the available
+versions, called *subsets*, in [destination rules](/docs/concepts/traffic-management/#destination-rules).
+
+Run the following command to create default destination rules for the Bookinfo services:
+
+{{< text bash >}}
+$ kubectl apply -f @samples/bookinfo/networking/destination-rule-all.yaml@
+{{< /text >}}
+
+{{< tip >}}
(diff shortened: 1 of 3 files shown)


monsteredc:
@googlebot I signed it!

craigbox:
Thanks for catching this, Xiaopeng!

I wonder though, seeing as it's not actually using the destination rules on this page, if we should just change the installation instructions in "Deploy the sample application" to install the destination rules too?

ericvn:
When this page was originally designed and implemented, there was no need for the DestinationRules, and they didn't get added. 

I do see the problem in that this page points to pages in `Next Steps` where the base DRs are now missing, and things may/may not work the same as coming from the BookInfo Application page.

I would tend to want to keep the BookInfo page and this page somewhat close so having the separate section probably matches better.

monsteredc:
@craigbox  Thanks for the advice. I think the suggestion you provided looks great as well. However, as @ericvn said, providing a seperate section will make the two pages look close, and I think it will make more sense to new users.

monsteredc:
@googlebot I fixed it.

monsteredc:
@googlebot I fixed it.

monsteredc:
@googlebot I fixed it.

monsteredc:
@googlebot I fixed it.

PR #9230, opened 2021-03-15
Adding destination rules part to the tip section in "bookinfo" page

monsteredc:
Originally in the page `setup/Getting Started`, when deploying the sample application using

`kubectl apply -f samples/bookinfo/platform/kube/bookinfo.yaml`

the destination rules are not applied at all, which will cause routing problems when doing other tasks of the sample application.
In the page `examples/bookinfo` application, the destination rules are mentioned to be required for the routing rules. However, there is a tip showing that `If you installed Istio using the Getting Started instructions, you already have Bookinfo installed and you can skip these steps.`. New users may skip applying the destination rules, and get stuck for a moment.

This PR is to add the destination rule tip to the tip section in "bookinfo" page which reminds new users to apply destination rules in order to try the tasks.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/examples/bookinfo/index.md
+++ b/content/en/docs/examples/bookinfo/index.md
@@ -16,8 +16,7 @@ to demonstrate various Istio features.
 {{< tip >}}
 If you installed Istio using the [Getting Started](/docs/setup/getting-started/)
 instructions, you already have Bookinfo installed and you can skip most of these steps
-and go direct to [Apply Default Destination Rules](/docs/examples/bookinfo/#apply-default-destination-rules)
-part.
+and go directly to [Apply Default Destination Rules](/docs/examples/bookinfo/#apply-default-destination-rules).
 {{< /tip >}}
 
 The application displays information about a
--- a/content/en/docs/examples/bookinfo/index.md
+++ b/content/en/docs/examples/bookinfo/index.md
@@ -15,7 +15,9 @@ to demonstrate various Istio features.
 
 {{< tip >}}
 If you installed Istio using the [Getting Started](/docs/setup/getting-started/)
-instructions, you already have Bookinfo installed and you can skip these steps.
+instructions, you already have Bookinfo installed and you can skip most of these steps
+and go direct to [Apply Default Destination Rules](/docs/examples/bookinfo/#apply-default-destination-rules)
+part.
 {{< /tip >}}
 
 The application displays information about a
--- a/content/en/docs/setup/getting-started/index.md
+++ b/content/en/docs/setup/getting-started/index.md
@@ -348,30 +348,6 @@ by viewing the Bookinfo product page using a browser.
 
 1.  Paste the output from the previous command into your web browser and confirm that the Bookinfo product page is displayed.
 
-### Apply default destination rules
-
-Before you can use Istio to control the Bookinfo version routing, you need to define the available
-versions, called *subsets*, in [destination rules](/docs/concepts/traffic-management/#destination-rules).
-
-Run the following command to create default destination rules for the Bookinfo services:
-
-{{< text bash >}}
-$ kubectl apply -f @samples/bookinfo/networking/destination-rule-all.yaml@
-{{< /text >}}
-
-{{< tip >}}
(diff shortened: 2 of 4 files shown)


monsteredc:
@ericvn Hello Eric, could you please help me review the PR? Thanks! It seems like the doc test framwork has some bugs, and two of them cannot be passed.

ericvn:
@monsteredc The egress and timeout tests are known failing tests (see https://github.com/istio/istio/issues/31317).

content/en/docs/setup/getting-started/index.md
```diff
@@ -348,6 +348,30 @@ by viewing the Bookinfo product page using a browser.
...
+{{< /tip >}}
+
+Wait a few seconds for the destination rules to propagate.
+
+You can display the destination rules with the following command:
+
+{{< text bash >}}
+$ kubectl get destinationrules -o yaml
+{{< /text >}}
+
```
ericvn, line 374: While I'm not necessarily keen on adding this section, it does make sense if we have the `Next steps` section at the bottom where the user may follow a link which and their Istio environment may not be correct (those pages are written to follow the `BookInfo` page from which this section is borrowed). I'm not sure how many of the follow-on pages would need updating (similar to what was added to request-routing in this commit), but at quick glance, maybe 3.

content/en/docs/setup/getting-started/index.md
```diff
@@ -348,6 +348,30 @@ by viewing the Bookinfo product page using a browser.
 
 1.  Paste the output from the previous command into your web browser and confirm that the Bookinfo product page is displayed.
 
+### Apply default destination rules
```
ericvn, line 351: One question I have, and I know @craigbox talked about moving this section up (I prefer to keep it the example back and here), is if we want to outdent this section like the BookInfo page and use ##?

content/en/docs/tasks/traffic-management/request-routing/index.md
```diff
@@ -15,9 +15,10 @@ microservice.
 ## Before you begin
 
 * Setup Istio by following the instructions in the
-[Installation guide](/docs/setup/).
+  [Installation guide](/docs/setup/).
 
-* Deploy the [Bookinfo](/docs/examples/bookinfo/) sample application.
+* Deploy the [Bookinfo](/docs/examples/bookinfo/) sample application including the
+  [default destination rules](/docs/examples/bookinfo/#apply-default-destination-rules).
```
frankbu, line 21: This seems unnecesary since there's already a warning on line 43 that the destinations rules are needed.

content/en/docs/setup/getting-started/index.md
frankbu: Rather than copying the whole section here, why not just fix the problem by correcting the tip in the bookinfo example: `If you installed Istio using the Getting Started instructions, you already have Bookinfo installed and you can skip most of these steps and proceed straight to Apply default destination rules(link).`

monsteredc: Thanks for the suggestions. I think revising the tip section makes sense.

content/en/docs/setup/getting-started/index.md
monsteredc: Thanks.

content/en/docs/tasks/traffic-management/request-routing/index.md
monsteredc: Thanks.

content/en/docs/examples/bookinfo/index.md
```diff
@@ -15,7 +15,9 @@ to demonstrate various Istio features.
 
 {{< tip >}}
 If you installed Istio using the [Getting Started](/docs/setup/getting-started/)
-instructions, you already have Bookinfo installed and you can skip these steps.
+instructions, you already have Bookinfo installed and you can skip most of these steps
+and go direct to [Apply Default Destination Rules](/docs/examples/bookinfo/#apply-default-destination-rules)
+part.
```
ericvn, line 20:
```suggestion
and go directly to [Apply Default Destination Rules](/docs/examples/bookinfo/#apply-default-destination-rules).
```

ericvn (approved):

PR #9276, opened 2021-03-16
Add info about SNI routing

howardjohn:
Fixes https://github.com/istio/istio/issues/31421

This is a very very common issue - I have probably seen it come up 10 times this month. So certainly worth an explanation in the docs I think.

Note: Previously the stuff about key formats was in the previous section, so I also split that out into its own section

### DIFF
--- a/content/en/docs/ops/common-problems/network-issues/index.md
+++ b/content/en/docs/ops/common-problems/network-issues/index.md
@@ -496,3 +496,24 @@ Then, simply bind both `VirtualServices` to it like this:
 - `Gateway` configuration `gw` with host `*.test.com`, selector `istio: ingressgateway`, and TLS using gateway's mounted (wildcard) certificate
 - `VirtualService` configuration `vs1` with host `service1.test.com` and gateway `gw`
 - `VirtualService` configuration `vs2` with host `service2.test.com` and gateway `gw`
+
+### Configuring SNI routing when not sending SNI
+
+An HTTPS `Gateway` configuring the `hosts` field will perform an [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) match on incoming requests.
+For example, the following configuration would only allow requests that match `*.example.com` in the SNI:
+
+{{< text yaml >}}
+servers:
+- port:
+    number: 443
+    name: https
+    protocol: HTTPS
+  hosts:
+  - "*.example.com"
+{{< /text >}}
+
+This may cause certain requests to fail.
+
+For example, if you do not have DNS set up and are instead directly setting the host header, such as `curl 1.2.3.4 -H "Host: app.example.com"`, no SNI will be set, causing the request to fail. Instead, you can set up DNS or use the `--resolve` flag of `curl`.
+
+Another common issue is load balancers in front of Istio. Most cloud load balancers will not forward the SNI, so if you are terminating TLS in your cloud load balancer you may need to either configure it to instead passthrough the TLS connection, or changes `hosts` to match `*`, which will disable the SNI matching.
--- a/content/en/docs/ops/common-problems/network-issues/index.md
+++ b/content/en/docs/ops/common-problems/network-issues/index.md
@@ -520,7 +520,7 @@ Instead, you can set up DNS or use the `--resolve` flag of `curl`. See the [Secu
 Another common issue is load balancers in front of Istio.
 Most cloud load balancers will not forward the SNI, so if you are terminating TLS in your cloud load balancer you may need to do one of the following:
 
-* Configure the cloud load balancer to instead passthrough the TLS connection
-* Disable SNI matching in the `Gateway` by setting the hosts field to `*`
+- Configure the cloud load balancer to instead passthrough the TLS connection
+- Disable SNI matching in the `Gateway` by setting the hosts field to `*`
 
 A common symptom of this is for the load balancer health checks to succeed while real traffic fails.
--- a/content/en/docs/ops/common-problems/network-issues/index.md
+++ b/content/en/docs/ops/common-problems/network-issues/index.md
@@ -499,7 +499,7 @@ Then, simply bind both `VirtualServices` to it like this:
 
 ### Configuring SNI routing when not sending SNI
 
-An HTTPS `Gateway` configuring the `hosts` field will perform an [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) match on incoming requests.
+An HTTPS `Gateway` that specifies the `hosts` field will perform an [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) match on incoming requests.
 For example, the following configuration would only allow requests that match `*.example.com` in the SNI:
 
 {{< text yaml >}}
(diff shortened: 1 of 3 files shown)


content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
```diff
@@ -423,12 +423,37 @@ $ kubectl create -n istio-system secret generic httpbin-credential --from-file=t
...
+
 Istio supports reading a few different Secret formats, to support integration with various tools such as [cert-manager](/docs/ops/integrations/certmanager/):
 
 * A TLS Secret with keys `tls.key` and `tls.crt`, as described above. For mutual TLS, a `ca.crt` key can be used.
 * A generic Secret with keys `key` and `cert`. For mutual TLS, a `cacert` key can be used.
 * A generic Secret with keys `key` and `cert`. For mutual TLS, a separate generic Secret named `<secret>-cacert`, with a `cacert` key. For example, `httpbin-credential` has `key` and `cert`, and `httpbin-credential-cacert` has `cacert`.
 
+## SNI Routing
+
+A HTTPS `Gateway` configuring the `hosts` field will perform an [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) match on incoming requests.
```
frankbu, line 438:
```suggestion
An HTTPS `Gateway` configuring the `hosts` field will perform an [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) match on incoming requests.
```

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
```diff
@@ -423,12 +423,37 @@ $ kubectl create -n istio-system secret generic httpbin-credential --from-file=t
...
+    number: 443
+    name: https
+    protocol: HTTPS
+  hosts:
+  - "*.example.com"
+{{< /text >}}
+
+This may cause certain requests to fail.
+
+For example, if you do not have DNS set up an are instead directly setting the host header, such as `curl 1.2.3.4 -H "Host: app.example.com"`, no SNI will be set, causing the request to fail. Instead, you can set up DNS or use the `--resolve` flag of `curl`.
```
frankbu, line 453:
```suggestion
For example, if you do not have DNS set up and are instead directly setting the host header, such as `curl 1.2.3.4 -H "Host: app.example.com"`, no SNI will be set, causing the request to fail. Instead, you can set up DNS or use the `--resolve` flag of `curl`.
```

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
```diff
@@ -423,12 +423,37 @@ $ kubectl create -n istio-system secret generic httpbin-credential --from-file=t
...
+
+Another common issue is load balancers in front of Istio. Most cloud load balancers will not forward the SNI, so if you are terminating TLS in your cloud load balancer you may need to either configure it to instead passthrough the TLS connection, or changes `hosts` to match `*`, which will disable the SNI matching.
```
frankbu, line 455: This should probably also (or instead) be in the common problems doc: https://preliminary.istio.io/latest/docs/ops/common-problems/network-issues/

howardjohn: I think it should be "also". Its important to be here as well I think

content/en/docs/ops/common-problems/network-issues/index.md
```diff
@@ -496,3 +496,24 @@ Then, simply bind both `VirtualServices` to it like this:
 - `Gateway` configuration `gw` with host `*.test.com`, selector `istio: ingressgateway`, and TLS using gateway's mounted (wildcard) certificate
 - `VirtualService` configuration `vs1` with host `service1.test.com` and gateway `gw`
 - `VirtualService` configuration `vs2` with host `service2.test.com` and gateway `gw`
+
+### Configuring SNI routing when not sending SNI
+
+An HTTPS `Gateway` configuring the `hosts` field will perform an [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) match on incoming requests.
```
nmittler, line 502: Maybe `configuring` -> `that specifies`

content/en/docs/ops/common-problems/network-issues/index.md
```diff
@@ -496,3 +496,24 @@ Then, simply bind both `VirtualServices` to it like this:
...
+    number: 443
+    name: https
+    protocol: HTTPS
+  hosts:
+  - "*.example.com"
+{{< /text >}}
+
+This may cause certain requests to fail.
+
+For example, if you do not have DNS set up and are instead directly setting the host header, such as `curl 1.2.3.4 -H "Host: app.example.com"`, no SNI will be set, causing the request to fail. Instead, you can set up DNS or use the `--resolve` flag of `curl`.
```
nmittler, line 517: By "do not have DNS set up" I assume you mean that your domain name can be looked up publicly? Maybe clarify the text a bit and add a link.

Also, it would be good to explain what the `--resolve` flag does and how it fixes the issue.

content/en/docs/ops/common-problems/network-issues/index.md
```diff
@@ -496,3 +496,24 @@ Then, simply bind both `VirtualServices` to it like this:
...
+
+Another common issue is load balancers in front of Istio. Most cloud load balancers will not forward the SNI, so if you are terminating TLS in your cloud load balancer you may need to either configure it to instead passthrough the TLS connection, or changes `hosts` to match `*`, which will disable the SNI matching.
```
nmittler, line 519: `changes hosts` -> `change hosts`

Also, you should be clear that you're referring to the `hosts` field in the `Gateway`. Maybe it could be restated `... or disable SNI matching in the Gateway by setting the hosts field to *`. Or maybe better yet, use a bullet list:

```
Another common issue is load balancers in front of Istio. Most cloud load
balancers will not forward the SNI, so if you are terminating TLS in your
cloud load balancer you may need to do one of the following:

- Configure it to instead passthrough the TLS connection
- Disable SNI matching in the `Gateway` by setting the hosts field to `*`
```

content/en/docs/ops/common-problems/network-issues/index.md
howardjohn: It might still be a good idea, but this page uses `--resolve` many times already. But yes that is what I meant about DNS, I wasn't sure a good way to phrase it.

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
frankbu: Mentioning it in this doc makes sense, but duplicating the same information in both places seems kind of ugly. How about instead of adding this section here, just add a warning around line 95, something like this:
```
{{< warning >}}
An HTTPS `Gateway` with a `hosts` field value other than `*` will perform
[SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) matching before forwarding a request,
which may cause some requests to fail. See [configuring SNI routing](link to common problem) for details.
{{< /warning >}}
```

content/en/docs/ops/common-problems/network-issues/index.md
mandarjog: @howardjohn how about specifically mentioning how health checks work general sni routing. I think we should use that example.

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
mandarjog: we should also mention the symptom. Health check passes but real traffic gets a 404.

howardjohn: Why does the health check succeed? is it just doing connection establishment?

The symptom would be connection closed btw not 404 - we never complete tls handshake. I will add it

howardjohn: or I guess the health check is port 15021, nevermind

howardjohn:
@frankbu can you take another look? I think I have addressed all comments

frankbu (approved):

PR #9280, opened 2021-03-16
Multicluster with external control plane instructions

frankbu:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -538,6 +539,163 @@ Unlike the first remote cluster, the second and subsequent clusters added to the
...
+    is set to `remote` this time, instead of `config`. Also note that the secret is applied in the remote (config)
+    cluster, instead of the external cluster, because that is where the external istiod is watching for additions.
+
+1. Create the remote Istio install configuration, which installs the injection webhook that uses the
+    external control plane's injector, instead of a locally deployed one:
+
+    {{< text syntax=bash snip_id=get_second_config_cluster_iop >}}
+    $ cat <<EOF > second-config-cluster.yaml
+    apiVersion: install.istio.io/v1alpha1
+    kind: IstioOperator
```
frankbu, line 573: As discussed in the env WG meeting, this is now only installing the injection webhook in the remote cluster, nothing else. It's also passing the network and cluster as env variables through the URL.

@costinm @howardjohn  Does this look right, i.e., like what you were suggesting?

I wonder if we should have a new `remote` profile that configures everything before the last 3 lines of this IOP?

Note that passing env variables through the URL path this way doesn't work for values containing / chars.

content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -538,6 +539,163 @@ Unlike the first remote cluster, the second and subsequent clusters added to the
...
+      values:
+        pilot:
+          configMap: false
+        telemetry:
+          enabled: false
+        global:
+          omitSidecarInjectorConfigMap: true
+          istioNamespace: external-istiod
+        istiodRemote:
+          injectionURL: https://${EXTERNAL_ISTIOD_ADDR}:15017/inject/cluster/${SECOND_CLUSTER_NAME}/net/network2
```
frankbu, line 590: Note that passing env variables through the URL path this way doesn't work for values containing / chars.

content/en/docs/setup/additional-setup/external-controlplane/index.md
howardjohn: I am not sure need a profile for a single YAML object. Maybe we should just have users directly `kubectl` apply it?

howardjohn: `istioctl x tag generate default --revision=default` is pretty close to doing this was well... cc @Monkeyanator

Monkeyanator: `istioctl tag` could be used here... only problem is it doesn't create selectors for `istio-inejction=enabled` or `sidecar.istio.io/inject`. Maybe we can add a `--default` flag that adds them

content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -522,11 +482,15 @@ including gateways, if needed.
 
     {{< text bash >}}
     $ curl -s "http://${GATEWAY_URL}/hello"
-    Hello version: v1, instance: helloworld-v1-5b75657f75-ncpc5
+    Hello version: v1, instance: helloworld-v1-776f57d5f6-s7zfc
     {{< /text >}}
 
 ## Adding clusters to the mesh (optional) {#adding-clusters}
 
+{{< warning >}}
```
ericvn, line 490: Does it make sense to have this match the `experimental` boilerplate, and maybe update that with the `experimental` link?

ericvn (approved):

Issue #9338, opened 2021-03-21
Kubernetes Service APIs point to nowhere


AssafKatz3:
Hi,
[Kubernetes Service APIs page](https://istio.io/latest/docs/tasks/traffic-management/ingress/service-apis/) contains link to github.com/kubernetes-sigs/service-apis/config/crd?ref=v0.1.0 but this page isn't existed.
Thanks

PR #9346, opened 2021-03-22
Fix configuring k8 gateway api

shamsher31:
To fix https://github.com/istio/istio.io/issues/9338

Ref: https://github.com/kubernetes-sigs/gateway-api
Service APIs is renamed to Gateway API

After update.
```
$ curl -s -I -HHost:httpbin.example.com "http://$INGRESS_HOST:$INGRESS_PORT/get"
HTTP/1.1 200 OK
server: istio-envoy
date: Mon, 22 Mar 2021 10:03:45 GMT
content-type: application/json
content-length: 1730
access-control-allow-origin: *
access-control-allow-credentials: true
x-envoy-upstream-service-time: 2
```


craigbox:
Is this superceded by #9347?

shamsher31:
> Is this superceded by #9347?

Yes, I can close this and it will be handle in  #9347

Issue #9348, opened 2021-03-22
Wildcard configuration for arbitrary domains with non root user


wangmeng:
For doc: https://istio.io/latest/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#wildcard-configuration-for-arbitrary-domains

The "Wildcard configuration for arbitrary domains" provides great capabilities to support arbitrary domains with egress gateway.
While in the current set up, both "istio-egressgateway-with-sni-proxy" and the nginx as sni-proxy are running as root defined via below 2 sections in IstioOperator definition. 

```
              securityContext:
                runAsNonRoot: false
                runAsUser: 0
```

```
  values:
    gateways:
      istio-egressgateway:
        runAsRoot: true
```

The above setting could be improved to run with non root user for both containers:
1. As for istio egressgateway, PR https://github.com/istio/istio/pull/23174 introduced the capability of `manifests: run gateways as non-root user`
2. nginx could run as non root by specifying the pid file location

wangmeng:
PR merged, so close this issue. Thanks.

PR #9325, opened 2021-03-19
Update istio test ref - fix timeout failures?

ericvn:
[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #9350, opened 2021-03-22
Wildcard configuration for arbitrary domains with non root user

wangmeng:
This PR is to track the change for issue: https://github.com/istio/istio.io/issues/9348

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

It contains the necessary change to run the solution with non root user. 

**The change includes:**
- The created egress gateway (istio-proxy container) will use port 8443 to handle the request for 443 request from sidecar, so egress gateway can run as non-root user. 
- The sni-proxy in the same  egressgateway pod will change to listen on port 18443 instead, since 8443 is used by istio-proxy in the same pod. 
- sni-proxy nginx pid file setup with custom path that do not require root access. 
- The port section in egressgateway-with-sni-proxy.yaml is optional, since it is actually the current default behavior for gateway. Still leave it here to demonstrate this Wildcard configuration solution actually could be used not only for the 443 port, but can be customized to support any port with tls protocol. 
```
              ports:
              - port: 443
                targetPort: 8443
                name: https
```

I made the change for files:
`content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md`
`content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/test.sh`

For file: `content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/snips.sh`
As it mentioned as below, I assume it would be automatically updated. 
```
THIS IS AN AUTO-GENERATED FILE, DO NOT EDIT. PLEASE MODIFY THE ORIGINAL MARKDOWN FILE:
#          docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md
```

Thanks.

### DIFF
--- a/content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md
+++ b/content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md
@@ -297,7 +297,8 @@ The SNI proxy will forward the traffic to port `443`.
 
     {{< text bash >}}
     $ cat <<EOF > ./sni-proxy.conf
-    user www-data;
+    # setup custom path that do not require root access
+    pid /tmp/nginx.pid;
 
     events {
     }
@@ -312,7 +313,7 @@ The SNI proxy will forward the traffic to port `443`.
       # tcp forward proxy by SNI
       server {
         resolver 8.8.8.8 ipv6=off;
-        listen       127.0.0.1:8443;
+        listen       127.0.0.1:18443;
         proxy_pass   \$ssl_preread_server_name:443;
         ssl_preread  on;
       }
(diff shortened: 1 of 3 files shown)


craigbox:
/ok-to-test

>

ericvn:
@wangmeng You need to run `make gen` to update the other files.

ericvn:
@istio/wg-networking-maintainers Looking for approval. Thanks.

wangmeng:
Thanks for looking into the PR. 

After using `make gen`, as suggested by @ericvn, to regenerate the `snips.sh` for wildcard-egress-hosts, the wildcard related test can pass. 

The only failure case now is the timeout setting doesn't work as expected for egress-control, which is not related to this PR's change. And tracking via: https://github.com/istio/istio.io/pull/9325

howardjohn (approved):
This is reasonable to me, thanks

ericvn (approved):

PR #9356, opened 2021-03-22
Bump to latest istio

howardjohn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/additional-setup/external-controlplane/index.md
+++ b/content/en/docs/setup/additional-setup/external-controlplane/index.md
@@ -185,15 +185,28 @@ and installing needed webhooks, configmaps, and secrets on the remote cluster so
         pilot:
           enabled: true
           k8s:
-            volumes:
-            - name: config-volume
-              volumeSource:
-                configMap:
-                  localObjectReference:
+            overlays:
+            - kind: Deployment
+              name: istiod
+              patches:
+              - path: spec.template.spec.volumes[100]
+                value: |-
+                  name: config-volume
+                  configMap:
                     name: istio
-            volumeMounts:
-            - name: config-volume
-              mountPath: /etc/istio/config
+              - path: spec.template.spec.volumes[100]
+                value: |-
+                  name: inject-volume
+                  configMap:
+                    name: istio-sidecar-injector
+              - path: spec.template.spec.containers[0].volumeMounts[100]
+                value: |-
+                  name: config-volume
+                  mountPath: /etc/istio/config
+              - path: spec.template.spec.containers[0].volumeMounts[100]
+                value: |-
+                  name: inject-volume
+                  mountPath: /var/lib/istio/inject
             env:
             - name: INJECTION_WEBHOOK_CONFIG_NAME
               value: ""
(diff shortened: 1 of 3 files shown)


ericvn:
@howardjohn Not sure if it was desired or not, but I don't see that you are actually using the latest istio.

ericvn:
The 408 change matches what I had done (assuming that is what should be returned now).

content/en/docs/setup/additional-setup/external-controlplane/snips.sh
```diff
@@ -93,15 +93,28 @@ spec:
...
-        volumes:
-        - name: config-volume
-          volumeSource:
-            configMap:
-              localObjectReference:
+        overlays:
+        - kind: Deployment
+          name: istiod
+          patches:
+          - path: spec.template.spec.volumes[100]
```
frankbu, line 100: Don't we use [-1] to mean append to end of list? Like in step 3 here: https://istio.io/latest/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#setup-egress-gateway-with-sni-proxy

howardjohn: I guess either must work? I copy+pasted from somewhere else.. there is a PR to add support for `volumes` directly so maybe we won't need it for long

frankbu (approved):

content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -398,6 +398,7 @@ and installing needed webhooks, configmaps, and secrets on the remote cluster so
     istio-leader                           0      2m9s
     istio-namespace-controller-election    0      2m11s
     istio-sidecar-injector                 2      2m1s
+    istio-validation-controller-election   0      2m6s
```
frankbu, line 401: @howardjohn I noticed that when I run locally the validatingwebhookconfiguration is NOT there, but it is when we run in CI. You also mentioned in the env WG call that only the mutatingwebhook is needed on the remote cluster. I'm wondering:
1. Why is it not there when I run locally?
2. Why isn't it needed? How does the validation happen if it's not there?

Do you understand what's going on?

howardjohn: Its probably because of all the fake sed stuff we do. With a real root cert, we don't have it since we don't need to patch the validatingwebhookconfiguration, which is why we don't have the configmap.

Why we don't have the actual validatingwebhookconfiguration itself, I am not entirely sure. But its almost certainly due to the fact the doc and doc test are not the same

howardjohn: Also its not needed in the remote-only cluster, since there is no config. In the config cluster you want it. If you install the CRDs you probably ought to install the webhook as well though.

frankbu: Oh, actually, I see this is the configmap, not the webhook, but I suppose my questions still apply? Any idea why the configmap is there in CI, but not when running locally?

Also, the second question still applies if I understood what you said in the env WG, that only the mutatingwebhook is needed in the remote cluster.

howardjohn: Yes, the test is doing fake stuff the doc doesn't do. Because we have no real certificate, we need to patch the webhook with our self-signed cert, and that results in th configmap.

If you are a remote ONLY cluster (no config), you don't *need* CRDs or webhook, since they do nothing. If you are a remote-config cluster, you still want them

frankbu: That all makes sense. Thanks!

PR #9367, opened 2021-03-23
Fix lint URL

craigbox:
[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/scripts/lint_site.sh
+++ b/scripts/lint_site.sh
@@ -63,7 +63,7 @@ check_content() {
     pushd "${TMP}" >/dev/null
 
     if ! find . -type f -name '*.md' -print0 | xargs -0 -r mdspell "${LANG}" --ignore-acronyms --ignore-numbers --no-suggestions --report; then
-        echo "To learn how to address spelling errors, please see https://istio.io/about/contribute/creating-and-editing-pages/#linting"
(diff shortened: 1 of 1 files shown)


scripts/lint_site.sh
```diff
@@ -35,7 +35,7 @@ check_content() {
 
     # check for use of ```
     if grep -nr -e "\`\`\`" --include "*.md" "${DIR}"; then
-        echo "Ensure markdown content uses {{< text >}} for code blocks rather than \`\`\`. Please see https://istio.io/about/contribute/creating-and-editing-pages/#embedding-preformatted-blocks"
+        echo "Ensure markdown content uses {{< text >}} for code blocks rather than \`\`\`. Please see https://istio.io/latest/about/contribute/code-blocks/"
```
ericvn, line 38: I believe we don't want `latest` in the links.

craigbox:
I personally don't want /latest anywhere, but 🤷‍♀️

What's the reasoning? Should we just link /about, which is a redirect to
/latest/about?

The URL will need to redirect when these move to /docs in 1.10

ericvn:
My thought was we just have the links point to `about` removing `latest`.

We likely will have a number of changes after the pages are moved.

ericvn (approved):

PR #9372, opened 2021-03-23
Separate Functionality and API for Gateway status

brian-avery:
This separates API and Gateway into two separate fields since they have different maturity. Follow up from https://github.com/istio/enhancements/pull/16

I had a failure on my local system with make lint that I don't think is related.

### DIFF
--- a/content/en/about/feature-stages/index.md
+++ b/content/en/about/feature-stages/index.md
@@ -37,7 +37,7 @@ Below is our list of existing features and their current phases. This informatio
 | Traffic Control: label/content based routing, traffic shifting | Stable
 | Resilience features: timeouts, retries, connection pools, outlier detection | Stable
 | Gateway Functionality (Ingress, Egress for all protocols) | Stable
-| Gateway APIs (Ingress, Egress for all protocols) | Stable
+| Gateway APIs (Ingress, Egress for all protocols) | Beta
 | TLS termination and SNI Support in Gateways | Stable
 | SNI (multiple certs) at ingress | Stable
 | [Locality load balancing](/docs/tasks/traffic-management/locality-load-balancing/) | Beta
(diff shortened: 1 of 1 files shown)


brian-avery:
@costinm @lizan @PiotrSikora can you please provide your thoughts around this?

content/en/about/feature-stages/index.md
```diff
@@ -36,7 +36,8 @@ Below is our list of existing features and their current phases. This informatio
 | Protocols: Websockets / MongoDB  | Stable
 | Traffic Control: label/content based routing, traffic shifting | Stable
 | Resilience features: timeouts, retries, connection pools, outlier detection | Stable
-| Gateway: Ingress, Egress for all protocols | Stable
+| Gateway Functionality (Ingress, Egress for all protocols) | Stable
+| Gateway APIs (Ingress, Egress for all protocols) | Stable
```
howardjohn, line 40: They have the same maturity listed here? So I am not sure the goal. But fine with me either way

content/en/about/feature-stages/index.md
```diff
@@ -36,7 +36,8 @@ Below is our list of existing features and their current phases. This informatio
 | Protocols: Websockets / MongoDB  | Stable
 | Traffic Control: label/content based routing, traffic shifting | Stable
 | Resilience features: timeouts, retries, connection pools, outlier detection | Stable
-| Gateway: Ingress, Egress for all protocols | Stable
+| Gateway Functionality (Ingress, Egress for all protocols) | Stable
+| Gateway APIs (Ingress, Egress for all protocols) | Beta
```
howardjohn, line 40: Maybe i am alone in thinking this, as its not written down anywhere, but my impression was `apiVersion` != feature status phase. The networking APIs are stable whether we call it that in the apiVersion (which is due to a technical constraint of the CRD API) or not.

For example, thinking we could deprecate the Gateway API in 3 months is completely impractical.

howardjohn (changes requested):
I think the API should/already is considered stable

content/en/about/feature-stages/index.md
brian-avery: Thanks. So we currently have: 

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
```

But the API itself is stable.

Issue #9409, opened 2021-03-26
Our alias to handle moved page no longer works?


linsun:
When I search Kubernetes service API in istio.io, this came first:  https://istio.io/latest/docs/tasks/traffic-management/ingress/service-apis/

but this page is getting 404.

Checked in our code, we do have the [alias](https://github.com/istio/istio.io/blob/593e85fb663511db852ce839507c1da764aceef6/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md) listed properly.  Is our alias no longer working?

@davidhauck can you look at this?

linsun:
cc @craigbox

craigbox:
The alias is /docs/foo but the page is /latest/doc/foo.

Adding that alias will fix it.

David and I are going to look at other solutions for /latest with Netlify's
new edge handlers

linsun:
Ah I missed that minor difference :) I'll submit a small PR

PR #9410, opened 2021-03-26
fix k8s service doc broken link

linsun:
fixes: https://github.com/istio/istio.io/issues/9409

Thanks @craigbox 

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
@@ -3,7 +3,7 @@ title: Kubernetes Gateway API
 description: Describes how to configure the Kubernetes Gateway API with Istio.
 weight: 50
 aliases:
-    - /docs/tasks/traffic-management/ingress/service-apis/
+    - /latest/docs/tasks/traffic-management/ingress/service-apis/
 keywords: [traffic-management,ingress]
 owner: istio/wg-networking-maintainers
 test: yes


ericvn (approved):

content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
```diff
@@ -3,7 +3,7 @@ title: Kubernetes Gateway API
 description: Describes how to configure the Kubernetes Gateway API with Istio.
 weight: 50
 aliases:
-    - /docs/tasks/traffic-management/ingress/service-apis/
+    - /latest/docs/tasks/traffic-management/ingress/service-apis/
```
frankbu, line 6: This doesn't look right. Won't this result in preliminary.istio.io redirecting to istio.io?

Also, this looks like a much more general problem ... aren't ALL aliases on the site broken this same way?

frankbu (changes requested):

content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
frankbu: cc: @davidhauck @craigbox

craigbox:
> This doesn't look right. Won't this result in preliminary.istio.io redirecting
to istio.io?

No

> Also, this looks like a much more general problem ... aren't ALL aliases
on the site broken this same way?

Yes

content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
linsun: While this is a general issue, I haven't noticed any other link broke other than this one because we recently renamed it.

ericvn: Agree that this is more of a recent problem as someone needed to save a link with `latest` and then have the page renamed. Earlier (pre using `latest`) URLs which were saved without `latest` in the link continue to work as the redirect from the non-`latest` links do continue to the new page.

ericvn: That said, people do need to realize that saving links with `latest` in them may not always point to the same info as a page may get updated release to release. While it may have pointed to the release-1.8 docs a few months ago, it might  point the the release-1.10 docs in another month.

craigbox:
I have to assume the number of people who "save links" is dwarfed by the search engine implications.

This is a problem which I'll bring up in a different issue:

image

frankbu:
Fixed this one, along with the external control plane alias, in #9740. It's a temporary hack until we fix the alias problem properly.

PR #9425, opened 2021-03-30
Fix the OpenGraph preview image

craigbox:
The <og:image> tag in the site metadata is used to display an image when you paste an Istio URL into Google Chat, for example.  At the moment it doesn't take the right image:

image

The URL in there currently is relative (`/latest/img/istio-whitelogo-bluebackground-framed.svg`) but it should be absolute according to https://ogp.me/.

I don't currently have access to our Netlify setup and I'm not sure if it passes the hosted URL through to Hugo in order that this would works, so let's test it.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -34,10 +34,10 @@
         {{ if .Permalink }}
             <meta property="og:url" content="{{ .Permalink }}">
         {{ end }}
-        <meta property="og:image" content="{{ .Site.BaseURL }}/img/istio-whitelogo-bluebackground-framed.svg">
+        <meta property="og:image" content="{{ .Site.BaseURL | absURL }}/img/istio-whitelogo-bluebackground-framed.svg">
         <meta property="og:image:alt" content="Istio Logo">
-        <meta property="og:image:width" content="112">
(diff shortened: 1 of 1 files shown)


craigbox:
From the deployment preview:

`<meta property="og:url" content="/latest/"><meta property="og:image" content="/latest/img/istio-whitelogo-bluebackground-framed.svg">`

That's a fail then.

It looks like to fix this dynamically, we would have to either (a) push `URL` or `DEPLOY_URL` from Netlify into the build process, all for two URLs (b) hard-code the URL in the page and then update the linter not to complain.

@davidhauck, any objection to b?

craigbox:
Works on https://www.opengraph.xyz/ (though the shape isn't great).
Doesn't work in Google Chat.

Following up with that team now.

ericvn:
Adding a DNM since the PR isn't ready yet. Remove the label when this is ready to merge.

ericvn (approved):

PR #9438, opened 2021-04-01
fix the ext-authz test flaky

yangminzhu:
There are possibly 2 timing issues: 

1. the extension provider is defined but the external authz service is not ready, causing istiod to fallback to a deny-all config.
2. the request for 200 is sent first where the authz config might not be distributed to proxy, causing the ext-authz flow not triggered.

This PR fix this by first deploying and waiting for the external authz service ready, and then define the extension provider, and also send the request for 403 first (this makes sure the ext-authz is already working in the proxy)

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn (approved):
LGTM.  Thanks for the quick update.

PR #9510, opened 2021-04-12
Polidea update - About breadcrumb

wbartosz:
This pull request includes a change redirecting user to / from the About breacrumb.

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/about/_index.md
+++ b/content/en/about/_index.md
@@ -5,4 +5,5 @@ description: Get a bit more in-depth info about the Istio project.
 sidebar_none: true
 weight: 15
 doc_type: about
+go_root_breadcrumb: true
 ---
--- a/layouts/partials/breadcrumbs.html
+++ b/layouts/partials/breadcrumbs.html
@@ -10,7 +10,14 @@
 {{ range $index := split $url "/" }}
     {{ if $page }}
         {{ if not $done }}
-            {{ $.Scratch.Add "urls" (slice $page.Permalink) }}
+            {{ $pageUrl := "" }}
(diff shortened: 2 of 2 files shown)


craigbox:
/latest/about/ still exists and can be reached. I think we will want to remove the page and have it just be a redirect? (It's in the Google search index.)

craigbox:
breadcrumbs do what we want, thanks!
going to https://deploy-preview-9510--preliminary-istio.netlify.app/latest/about/ does not.  (this URL is presumably in the Google search index.)

Should we just make that page use a blank layout or something to get rid of the "flash of unstyled content"

davidhauck:
I'd rather have a redirect to `/` than a blank layout

craigbox:
is that something we should do in `layouts/index.redir`?

davidhauck:
Yeah, I think we could just do a hardcoded redirect from /about -> /

wbartosz:
I have tried adding `/about /` as well as `/latest/about /latest` but those didn't work. Don't you have to edit the netlify.toml?

craigbox:
I think that `/latest/about / 301!` would work, according to the tester at https://play.netlify.com/redirects.
the `!` is needed to tell it to ignore the fact `/latest/about` exists and force it anyway

craigbox:
LGTM, leaving for David to approve.

davidhauck (approved):

kebe7jun (approved):

craigbox (approved):

PR #9523, opened 2021-04-13
Bump istio version to head of 1.10

howardjohn:
If this DOESN'T fail we have a problem. https://github.com/istio/istio/commit/27816e7b44c74a4e41d08658333b42a9de88d186 should make it fail, to match 1.9 behavior

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ replace github.com/evanphx/json-patch => github.com/evanphx/json-patch v0.0.0-20
 require (
 	github.com/golang/sync v0.0.0-20180314180146-1d60e4601c6f
 	github.com/pmezard/go-difflib v1.0.0
-	istio.io/istio v0.0.0-20210412210842-f171f4daffb4
+	istio.io/istio v0.0.0-20210413235758-a57ed8a5b016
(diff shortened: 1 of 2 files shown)


ericvn:
I don't think we ran into the timeout failure we would expect:
Egress test:
```
VERIFY FAILED snip_manage_traffic_to_external_services_3 (timeout after 120s):
received:
"200
real	0m 5.09s
user	0m 0.00s
sys	0m 0.00s"
expected:
"408
real    0m3.149s
user    0m0.004s
sys     0m0.004s"
```

request timeout:
```
expected:
"Sorry, product reviews are currently unavailable for this book."
```

howardjohn:
I think we need an envoy update

ericvn:
I did try locally using the latest `master` branch commit and running the `egress/egress-control` and it continued to pass with a `408` being returned (I did need to make a small change as an additional header, "Accept-Encoding": "gzip", was being returned in one of the steps.

ericvn:
We have verified that this didn't fail on the 1.10-alpha.1 build which has the Envoy and Istio changes and A further comment was made in https://github.com/istio/istio/issues/31616 and it looks like we will need another Envoy change. Not sure if you want to leave this PR open.

PR #9555, opened 2021-04-19
update standard metrics documentation

douglas-reid:
As part of 1.10 release, ensure that the standard metrics documentation matches behavior. This also removes mentions of Mixer from the page, as that is no longer relevant.

[ X ] Policies and Telemetry

### DIFF
--- a/content/en/docs/reference/config/metrics/index.md
+++ b/content/en/docs/reference/config/metrics/index.md
@@ -112,8 +112,8 @@ For TCP traffic, Istio generates the following metrics:
     destination_canonical_revision
     {{< /text >}}
 
-*   **Destination Cluster**: Name of the cluster for the destination workload.
+*   **Destination Cluster**: This identifies the cluster of the destination workload.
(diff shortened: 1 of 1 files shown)


bianpengyuan (approved):

content/en/docs/reference/config/metrics/index.md
```diff
@@ -113,3 +111,9 @@ For TCP traffic, Istio generates the following metrics:
     destination_canonical_service
     destination_canonical_revision
     {{< /text >}}
+
+*   **Destination Cluster**: Name of the cluster for the destination workload.
```
smawson, line 115: Should we match the wording of all the other items?

"This identifies the cluster of the destination workload."

Same for source.

douglas-reid: Sure. This was just copying over the dropped text from before, but we should update it to match.

smawson (approved):

PR #9563, opened 2021-04-19
Filter mixer docs from related content sections

douglas-reid:
This PR is meant to prevent Mixer blog posts from 2017 (and similar) from showing up in the `See Also` section of pages in our docs. This is to prevent user confusion and stop talking about Mixer (which has been deprecated and removed from Istio).

It also attempts to add more appropriate keywords in a few places to link docs together hopefully.

[ X ] Policies and Telemetry

### DIFF
--- a/layouts/partials/primary_bottom.html
+++ b/layouts/partials/primary_bottom.html
@@ -3,8 +3,11 @@
 {{ $skipPageNav := .Scratch.Get "skipPageNav" }}
 
             {{ if .Scratch.Get "seeAlso" }}
-                {{ $related := .Site.RegularPages.Related . | first 6 }}
-                {{ with $related }}
+                {{ $related := .Site.RegularPages.Related . | first 8 }}
+                {{ $mixerSPOF := where $related "RelPermalink" "/blog/2017/mixer-spof-myth/"}}
+                {{ $mixerAdapter := where $related "RelPermalink" "/blog/2017/adapter-model/"}}
+                {{ $seeAlso := $related | complement $mixerSPOF $mixerAdapter | first 6}}
+                {{ with $seeAlso }}
                     <nav id="see-also">
                         <h2>{{ i18n "see_also" }}</h2>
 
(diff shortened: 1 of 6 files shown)


layouts/partials/primary_bottom.html
```diff
@@ -3,8 +3,11 @@
 {{ $skipPageNav := .Scratch.Get "skipPageNav" }}
 
             {{ if .Scratch.Get "seeAlso" }}
-                {{ $related := .Site.RegularPages.Related . | first 6 }}
-                {{ with $related }}
+                {{ $related := .Site.RegularPages.Related . | first 8 }}
+                {{ $mixerSPOF := where $related "RelPermalink" "/blog/2017/mixer-spof-myth/"}}
+                {{ $mixerAdapter := where $related "RelPermalink" "/blog/2017/adapter-model/"}}
+                {{ $seeAlso := $related | complement $mixerSPOF $mixerAdapter | first 6}}
+                {{ with $seeAlso }}
```
ericvn, line 10: @davidhauck Does this sound OK?

davidhauck: I'd rather have these pages be in a constant somewhere. That could be a list of directories in data/*.yaml or using [front-matter](https://gohugo.io/content-management/front-matter/#user-defined) on pages that we don't want appearing here, and then filtering on that instead. The latter is probably preferred.

douglas-reid: OK. switched to the front-matter approach. ptal.

douglas-reid:
ping on review. it would be nice to have this in before 1.10.

ericvn:
@davidhauck Review please. Thanks.

layouts/partials/primary_bottom.html
davidhauck: Sorry missed this, looks good!

davidhauck (approved):

ericvn (approved):

Issue #9529, opened 2021-04-14
Automate https://preliminary.istio.io/latest/docs/tasks/observability/kiali/


esnible:
Notes:
- The instructions should ask for a Prometheus install.  I followed https://istio.io/latest/docs/ops/integrations/prometheus/
- I missed the directive "This task assumes the Bookinfo application is installed in the bookinfo namespace."
- The instructions say "This task assumes the Bookinfo application is installed...", but it the pictures were generated using more than the vanilla _samples/bookinfo/platform/kube/bookinfo.yaml_.  For example they expect three versions of **ratings**.
- I suspect the instruction at https://preliminary.istio.io/latest/docs/tasks/observability/kiali/#viewing-and-editing-istio-configuration-yaml to do `kubectl apply -f samples/bookinfo/networking/destination-rule-all.yaml` should have a `-n bookinfo`.

I don't want to pre-req some Ajax web testing tool in the Istio docs repo.  I am trying to decide how serious to make these tests.

- Just test that the commands in the documentation can execute
- Do the commands in the documentation, but instead of testing the Kiali UI use the Kiali API, e.g. `curl "localhost:20001/kiali/api/namespaces/graph?namespaces=bookinfo&graphType=app"`
- Really test the Kiali UI

PR #9570, opened 2021-04-20
Automation for 'visualize your mesh' with Kiali

esnible:
Resolves https://github.com/istio/istio.io/issues/9529


douglas-reid:
@esnible what is the status of this work?  This is one of the two "automate" tasks assigned to E&T this release cycle. It'd be amazing if we could get this in.

PR #9595, opened 2021-04-23
Add IOP to glossary

brian-avery:
I realized we use the term IOP all over the place without actually defining what it is. This adds it to the glossary.
[x] Docs
[x] Installation
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/reference/glossary/iop.md
+++ b/content/en/docs/reference/glossary/iop.md
@@ -1,0 +1,7 @@
+---
+title: IOP
+test: n/a
+---
+
+IOP, short for Istio Operator, is the custom resource used to configure
+an Istio installation when installing using the `istioctl install` command.
--- a/content/en/docs/reference/glossary/iop.md
+++ b/content/en/docs/reference/glossary/iop.md
@@ -4,4 +4,5 @@ test: n/a
 ---
 
 IOP, short for Istio Operator, is the custom resource used to configure
-an Istio installation when installing using the `istioctl install` command.
+an Istio installation when installing using the `istioctl install` command or
+the in-cluster operator.
--- a/content/en/docs/reference/glossary/istio-operator-custom-resource.md
+++ b/content/en/docs/reference/glossary/istio-operator-custom-resource.md
@@ -3,6 +3,7 @@ title: Istio Operator Custom Resource
 test: n/a
 ---
 
-The Istio Operator Custom Resource is the custom resource used to configure
+The Istio Operator Custom Resource, often referred to using the short form `IOP` or `IO`,
+is the custom resource used to configure
 an Istio installation when installing using the `istioctl install` command or
 the in-cluster operator.
--- a/content/en/docs/reference/glossary/istio-operator-custom-resource.md
+++ b/content/en/docs/reference/glossary/istio-operator-custom-resource.md
@@ -1,0 +1,8 @@
+---
+title: Istio Operator Custom Resource
+test: n/a
+---
+
+The Istio Operator Custom Resource is the custom resource used to configure
+an Istio installation when installing using the `istioctl install` command or
+the in-cluster operator.
--- a/content/en/docs/reference/glossary/istio-resource.md
+++ b/content/en/docs/reference/glossary/istio-resource.md
@@ -1,5 +1,5 @@
 ---
-title: IOP
+title: Istio Resource
 test: n/a
 ---
 
--- a/content/en/docs/reference/glossary/istio-resource.md
+++ b/content/en/docs/reference/glossary/istio-resource.md
@@ -1,6 +1,0 @@
----
-title: Istio Resource
-test: n/a
----
-
-See [Istio Operator Custom Resource](#istio-operator-custom-resource)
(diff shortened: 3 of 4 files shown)


craigbox:
_Do_ we use it all over the place?  I don't see many mentions in a quick grep of the docs.

Maybe an entry that says "ISTIOOPERATOR, IO, IOP" or something?  All are valid objects afaict.

brian-avery:
Do we support aliases in the glossary? That might be more useful...

brian-avery:
The problem with that using `ISTIOOPERATOR, IO, IOP` is that any attempts to use it in the glossary will break as the glossary is looking for the specific term. It looks like we also refer to this by the term `Istio CRDs`. 

I saw 7 references to `iop` when I looked and it's also in the release notes for 1.9.4. I realized while reviewing those that I only had a vague idea of what it actually was and that if I didn't know, many users probably didn't know either. Looking at those references though, in the case of `iop`, many of them are snippet IDs and duplicates between English and Chinese. 

Regardless, this is something that feels like it's assumed to be a common term but is specific to Istio and doesn't have a definition recorded anywhere.

Closes that I found to a definition was:

`Added io as short name for Istio Operator resources in addition to iop. (Issue #27159)` from the 1.8 release notes.

brian-avery:
What about something like this:

**IOP**
IOP is the custom resource used to configure
an Istio installation when installing using the `istioctl install` command.

**Istio Operator Custom Resource**
See IOP

**Istio Resource** 
See IOP

**IO**
See IOP

brian-avery:
As @ericvn pointed out, we can define aliases directly in the gloss field:

https://preliminary.istio.io/latest/docs/releases/contribute/shortcodes/#glossary-terms

content/en/docs/reference/glossary/iop.md
```diff
@@ -0,0 +1,7 @@
+---
+title: IOP
+test: n/a
+---
+
+IOP, short for Istio Operator, is the custom resource used to configure
+an Istio installation when installing using the `istioctl install` command.
```
morvencao, line 7: Not only for standalone operator(`istioctl install`) but also for in-cluster operator, they share the same IOP resource.

morvencao (approved):

ericvn (approved):

brian-avery:
@frankbu do you have any thoughts on how to handle multiple words for the same thing? I can't imagine this is the only place this occurs in Istio.

frankbu:
@brian-avery Glossary references can use different words than the glossary entry. For example: `{{< gloss "IOP" <}}Istio operator custom resource{{< /gloss >}}`

brian-avery:
Thanks @frankbu. So would you list the term as I have it in this PR: 

```
Or would you do something like:
**IOP** 
IOP is the custom resource used to configure
an Istio installation when installing using the istioctl install command.
```
and then just provide links with the various synonyms specified? 

```
Or would you do something like:
**ISTIOOPERATOR, IO, IOP** 
IOP is the custom resource used to configure
an Istio installation when installing using the istioctl install command.
```

in the glossary and then specify each term individually when linking to it? 

It seems like the glossary currently only lists one form of a term rather than multiple, but I could see someone navigating directly to the glossary (I did that in fact) and trying to look up a term there. 

The other option I thought of was: 

```
IOP
IOP is the custom resource used to configure
an Istio installation when installing using the istioctl install command.

Istio Operator Custom Resource
See IOP

Istio Resource
See IOP

IO
See IOP
```

ericvn:
I'm partial to including each in the Glossary (except I'd make `Istio Operator Custom Resource` the one with the full description) for those that just peruse the Glossary, and then within the docs, you can point to the Glossary item with the description, will still using the other name. I hope that makes sense.

content/en/docs/reference/glossary/istio-resource.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Istio Resource
```
craigbox, line 2: Is this too generic?  I would have thought "Istio resource" could apply to any API object created against an Istio CRD.

(Otherwise LGTM)

brian-avery: Let me remove that one. I could have sworn I saw somewhere where the docs were simply referring to it as "your Istio resource" but now I'm not seeing it. I agree that it's too generic.

brian-avery:
@ericvn is this good to go? If so, I'll remove the dnm.

content/en/docs/reference/glossary/istio-operator-custom-resource.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Istio Operator Custom Resource
+test: n/a
+---
+
+The Istio Operator Custom Resource is the custom resource used to configure
```
frankbu, line 6:
```suggestion
The Istio Operator Custom Resource, often referred to using the short form `IOP` or `IO`,
is the custom resource used to configure
```

frankbu:
This LGTM now, but I think you should you add references to it, in all (or at least many of) the docs that use the terms?

Issue #9605, opened 2021-04-26
the wrong link in doc


ha5ky:
hi there, there is a wrong link at https://istio.io/latest/docs/tasks/observability/metrics/customize-metrics/, 
![image](https://user-images.githubusercontent.com/23353351/116045781-f62e2a00-a6a4-11eb-9f5e-8907a12e9d23.png)
when i click this link, the page notice me 404. 
it is confusing me which yaml file shoud i choose？

PR #9606, opened 2021-04-26
Customizing Istio Metrics: make the telemetry link dynamic

shamsher31:
To fix https://github.com/istio/istio.io/issues/9605

### DIFF
--- a/content/en/docs/tasks/observability/metrics/customize-metrics/index.md
+++ b/content/en/docs/tasks/observability/metrics/customize-metrics/index.md
@@ -23,7 +23,7 @@ By default, Istio defines and generates a set of standard metrics (e.g.
 
 Istio uses the Envoy proxy to generate metrics and provides its configuration in
 the `EnvoyFilter` at
-[`manifests/charts/istio-control/istio-discovery/templates`]({{<github_blob>}}/manifests/charts/istio-control/istio-discovery/templates/telemetryv2_{{< istio_version >}}.yaml).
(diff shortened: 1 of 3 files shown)


content/en/docs/tasks/observability/metrics/customize-metrics/index.md
```diff
@@ -23,7 +23,7 @@ By default, Istio defines and generates a set of standard metrics (e.g.
 
 Istio uses the Envoy proxy to generate metrics and provides its configuration in
 the `EnvoyFilter` at
-[`manifests/charts/istio-control/istio-discovery/templates/telemetryv2_1.7.yaml`]({{<github_blob>}}/manifests/charts/istio-control/istio-discovery/templates/telemetryv2_1.7.yaml).
+[`manifests/charts/istio-control/istio-discovery/templates`]({{<github_blob>}}/manifests/charts/istio-control/istio-discovery/templates/telemetryv2_{{< istio_version >}}.yaml).
```
ericvn, line 26: My only concern is we have removed the version from the link and we are now pointing to the folder. I suspect that's because we single quotes. Might try removing the quotes.

shamsher31: Removing quotes will render it as plain text and it won't be more recognizable. We are saying in the sentence that `EnvoyFilter is at location` so it should be fine to point them to the directory. Anyways we are redirecting users to the exact file they are looking for the configuration.

ericvn: It would be plain text, but it is still highlighted in blue as a link which should be OK as it is a link.

shamsher31: Yup, it looks good without a qoute. Thanks.

ericvn (approved):

PR #9608, opened 2021-04-26
Highlight and recommend canary upgrade

shamsher31:
Part of https://github.com/istio/istio/issues/32384

### DIFF
--- a/content/en/docs/setup/upgrade/in-place/index.md
+++ b/content/en/docs/setup/upgrade/in-place/index.md
@@ -13,7 +13,7 @@ criteria. Also, it alerts the user if it detects any changes in the profile
 default values between Istio versions.
 
 {{< tip >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/upgrade/in-place/index.md
```diff
@@ -12,6 +12,10 @@ the upgrade, it checks that the Istio installation meets the upgrade eligibility
 criteria. Also, it alerts the user if it detects any changes in the profile
 default values between Istio versions.
 
+{{< tip >}}
+[Canary Upgrade](/docs/setup/upgrade/canary/) is much safer than doing an in-place upgrade and is the highly recommended upgrade method.
```
ericvn, line 16: My concern here is the `much` and `highly`. Is it better to say `Canary Upgrade is safer than doing an in-place upgrade and is the recommended upgrade method.`?

shamsher31: Sounds better

ericvn (approved):

PR #9616, opened 2021-04-27
Use new external profile

frankbu:
Depends on https://github.com/istio/istio/pull/32469

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/install/multicluster/primary-remote_multi-network/index.md
+++ b/content/en/docs/setup/install/multicluster/primary-remote_multi-network/index.md
@@ -174,7 +174,6 @@ $ cat <<EOF > cluster2.yaml
 apiVersion: install.istio.io/v1alpha1
 kind: IstioOperator
 spec:
-  profile: remote
   values:
     global:
       meshID: mesh1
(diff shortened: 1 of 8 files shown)


content/en/docs/setup/install/multicluster/primary-remote_multi-network/index.md
```diff
@@ -174,7 +174,6 @@ $ cat <<EOF > cluster2.yaml
 apiVersion: install.istio.io/v1alpha1
 kind: IstioOperator
 spec:
-  profile: remote
```
smawson, line 177: Should these use external profile as base or no profile?

frankbu: They should be using the `external` profile, but that will require several other changes to the example, which currently relies on a local istiod for CA and injection. So, for now, I just removed the `profile=remote`, because it's deprecated. The current example continues to work with no profile (i.e., default settings). I opened #9618 to fix them to properly use `profile=external` in a future PR.

ericvn (approved):

PR #9623, opened 2021-04-28
Minor grammatical changes to traffic management

jacob-delgado:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/concepts/traffic-management/index.md
+++ b/content/en/docs/concepts/traffic-management/index.md
@@ -313,7 +313,7 @@ value, a prefix, or a regex.
 You can add multiple match conditions to the same `match` block to AND your
 conditions, or add multiple match blocks to the same rule to OR your conditions.
 You can also have multiple routing rules for any given virtual service. This
-lets you make your routing conditions as complex or simple as you like within a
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/traffic-management/index.md
```diff
@@ -313,7 +313,7 @@ value, a prefix, or a regex.
 You can add multiple match conditions to the same `match` block to AND your
 conditions, or add multiple match blocks to the same rule to OR your conditions.
 You can also have multiple routing rules for any given virtual service. This
-lets you make your routing conditions as complex or simple as you like within a
+lets you make your routing conditions be as complex or simple as you like within a
```
craigbox, line 316: Both read more correctly to me in the older version than the newer one.  For example, you're saying "make something be"  which is an odd construct.  "This lets your routing conditions be" sounds more correct, but is no longer active voice.

(I am not a copy editor)

smawson:
IMO both changes are unnecessary, it read better beforehand. Also not a copyeditor but I play one on tv :-D

PR #9624, opened 2021-04-28
Remove citadel reference

jacob-delgado:
Citadel was deprecated in Istio 1.5 and was removed in Istio 1.6

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/setup/upgrade/in-place/index.md
+++ b/content/en/docs/setup/upgrade/in-place/index.md
@@ -35,7 +35,7 @@ Before you begin the upgrade process, check the following prerequisites:
 
 {{< warning >}}
 Traffic disruption may occur during the upgrade process. To minimize the disruption, ensure
-that at least two replicas of each component (except Citadel) are running. Also, ensure that
+that at least two replicas of each component are running. Also, ensure that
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/upgrade/in-place/index.md
```diff
@@ -35,7 +35,7 @@ Before you begin the upgrade process, check the following prerequisites:
 
 {{< warning >}}
 Traffic disruption may occur during the upgrade process. To minimize the disruption, ensure
-that at least two replicas of each component (except Citadel) are running. Also, ensure that
+that at least two replicas of each component are running. Also, ensure that
```
craigbox, line 38: What are the components?  Do we just mean istiod, or do we also mean gateways?

jacob-delgado: I think this is in reference to the deployments. Only istiod is shipped.

craigbox: could we then say "two replicas of Istiod"?

craigbox (approved):

PR #9639, opened 2021-04-29
Fix broken href in docs about upgrade istio

laststem:
I found broken link, please see [link](https://istio.io/v1.6/docs/setup/upgrade)

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
While I wouldn't expect a re-archive of the release-1.6 branch, if it was done, it would over-ride this change back to the olde version. It would probably be best to make this update in the `release-1.6` branch, and then re-archive it to pull in the change (and cherry-pick that to the istio.io branch (currently 1.9)). That would fix the problem at the source so it wouldn't have a possibility of getting changed, as well as getting it updated in the current preliminary.istio.io and istio.io repos.

PR #9649, opened 2021-04-30
Add publish dates

brian-avery:
This adds publish dates for the initial release notes for all releases. Preliminary.istio.io sorts by the date when displaying notes and because it's not present, they all get sorted to the bottom.

Changes in this PR: 
* synchronized minor release front matter with patch release
* added publish date for all upgrade notes and minor release notes
* updated listing to only list pages with publish dates set in "news" rather than sorting those without publish dates to the bottom. **Support announcements, security releases, and release notes should all have publish dates set.**
* updated listing to show all pages rather than just regular pages -- the minor releases have a `_index.md` instead of `index.md`. 
* fixed filters for blog & news
* fixed a broken link that was being covered up by 1.8 not being scanned

`_index.md` tells Hugo to use that page as a directory listing. If we use `index.md`, then Hugo will only generate content for the one page in the directory, skipping upgrade notes and release notes. Instead, we now request all pages (including both regular pages and listing pages)


ericvn:
One thing I notice is that all the `Update Notes` and `Change Notes` pages don't have a release in the title. Scrolling through the list you will find a number of these pages, and you need to read the lower text to figure out which one they are for.

image

ericvn (commented):
LGTM

craigbox:
This is great, Brian.  Thanks, approved.

To the future reader: as discussed on Slack, the plan of record is not to use the new "river" layout for news, so I'm noting that with this design, [/releases](https://deploy-preview-9649--preliminary-istio.netlify.app/latest/news/releases/) and [/releases/1.9](https://deploy-preview-9649--preliminary-istio.netlify.app/latest/news/releases/1.9.x/) are both empty pages, but we don't care to fix that right now.

craigbox (approved):

ericvn (approved):

PR #9657, opened 2021-04-30
[WIP] Add security icon

brian-avery:
Something we discussed in chat the other day was potentially adding a security icon to indicate releases that fixed vulnerabilities. This explores the idea.

### DIFF
--- a/content/en/news/releases/1.1.x/announcing-1.1/_index.md
+++ b/content/en/news/releases/1.1.x/announcing-1.1/_index.md
@@ -1,8 +1,8 @@
 ---
-title: Announcing Istio 1.1
-linktitle: 1.1
+title: Announcing Istio 1.1.0
+linktitle: 1.1.0
 subtitle: Major Update
-description: Istio 1.1 release announcement.
+description: Istio 1.1.0 release announcement.
 publishdate: 2019-03-19
 release: 1.1.0
 skip_list: true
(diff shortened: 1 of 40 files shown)


craigbox:
Are you concerned that this might disincentivize people?  Is there a case where we'll silently fix a released bug that we don't want to call out?

brian-avery:
The feedback that we received in the first survey was that there were a very large number of people running vulnerable versions of Istio that didn't know that they were vulnerable. We now post these announcements to Twitter, which likely helps, but I don't know if it's worth calling out what releases are important to consider from a security perspective vs what you might want to upgrade to if you have a bug that you're experiencing in your environment. I.e. more people are likely interested in upgrading for the former. 

From a security point of view, I think any issues that we deem are important enough to warrant a security release are big enough to announce. I don't think we'd make exceptions for things we do security releases for. We do have lower priority issues that we fix without security issues, but those are different. 

And in terms of disincentivizing people, there are vulnerabilities in all sorts of software every day that aren't caught and lead to security breaches. As a user, I think I'd much rather be aware that a security release is out there and has been fixed than be vulnerable and not know it. That being said, my sample size there is limited to myself, so I'd be interested to hear what the greater opinion is.

brian-avery:
This was just an exploration of what we could do and how we could tie the disjoint security, releases, and support sections that we currently have together. I think users are likely more concerned about a release with a security fix than just a security notice itself.

brian-avery:
@ericvn suggestion here was to only show the lock next to the latest security fix for each version, which makes sense to me. Everything below that at that point is insecure anyway.

Issue #9569, opened 2021-04-20
Large header takes up too much space in doc windows


frankbu:
There seems to be too much wasted space when reading docs in the new istio.io layout. The header wastes about 5 or 6 lines of content, and once you've scrolled past the outline, there's a big block of empty space on the left. The actual document of interest ends up too small and squished in the middle.

![image](https://user-images.githubusercontent.com/2752495/115449467-d9977b00-a1e8-11eb-8a72-3e0f090892ed.png)

craigbox:
See also #9515 from @howardjohn 

> The new header on the website is problematic I think. It is large in general, and not responsive. Most sites with such a large header either are not sticky, shrink when the screen is smaller, or shrink when scrolling. The new header does none of this, leading to a poor UX

Flagging this for options.

We can:

(a) not keep the header on the page for docs
(b) shrink the header on docs pages
(c) resize on scroll.

I would prefer (c) if it's easy to do. We already trigger a CSS class when the page is scrolled.  We could reduce the padding with some kind of easing (and shrinking of the logo?)

frankbu:
I agree with @howardjohn that the header is problematic, in fact, I think it is probably the single most annoying problem with the new layout. Option (c) SGTM, most websites seem to do that kind of thing - i.e., scroll away to a much smaller header when the user scrolls down in the page.

craigbox:
Devil's advocate: Is there anything in the header so valuable we'd want to leave even a smaller version of it visible at all times _for someone who has cared enough to scroll down on a docs page?_

Things I can think: the search box, an easy link to the home page

rytswd:
I also found it really annoying to have so much space wasted with the header - so I tried a quick fix for smaller header with the existing scroll spy. I'm not proud of this rather ugly change, but hopefully that works to demonstrate how I wished for header to shrink.

craigbox:
@rytswd That is lovely, and with another edit to resize the "Try Istio button, is probably what we're going to go with. 💟

rytswd:
Got the Try Istio button a bit smaller on the scroll as well, but it looks like the text jumps around for a split second. I'm not too happy with it TBH, but would need much more time digging into the code structure if I were to properly get it fixed...

craigbox:
Polidea: please implement this based on Ryota's draft PR and fixing the alignment issues on the Istio logo and the "Try Istio" button.  We are undecided on if we want this to apply to all pages or only docs pages; Aizhamal and I will make that call when we see the output.

[...]

PR #9659, opened 2021-04-30
Add max-height and transition to resize header

rytswd:

Related to #9569

This is a bit hacky as the existing scroll spy was only meant to update single element. With this update, the header size gets updated along with some other elements, and that makes the SCSS really scattered...

I just wanted to propose how this change improves the UX significantly with small CSS changes in a few places. A proper fix may need a bit of restructuring in SCSS.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/src/sass/misc/_header.scss
+++ b/src/sass/misc/_header.scss
@@ -9,13 +9,13 @@
     z-index: $header-z;
     padding: 1.25rem 0;
     transition: background-color .2s, box-shadow .2s,
-                max-height .2s, 
-                padding-top .2s, padding-bottom .2s;
+    max-height .2s,
+    padding-top .2s, padding-bottom .2s;
 
     &.active {
         box-shadow: 0 6px 9px 0 rgba(0, 0, 0, .16);
         max-height: 3.4375rem;
-        padding: 0.6125rem 0;
+        padding: .6125rem 0;
 
         .logo svg {
             height: 2.125rem;
@@ -46,8 +46,8 @@
         .logo {
             fill: $secondaryColor;
 
-            svg { 
-                width: 106px; 
+            svg {
+                width: 106px;
                 transition: height .2s;
             }
         }
--- a/src/sass/misc/_header.scss
+++ b/src/sass/misc/_header.scss
@@ -1,16 +1,30 @@
 .main-navigation {
     position: fixed;
     height: $headerHeight;
+    max-height: $headerHeight;
     background-color: #ffffff;
     top: 0;
     right: 0;
     left: 0;
     z-index: $header-z;
     padding: 1.25rem 0;
-    transition: background-color .2s, box-shadow .2s;
+    transition: background-color .2s, box-shadow .2s,
+                max-height .2s, 
+                padding-top .2s, padding-bottom .2s;
 
     &.active {
         box-shadow: 0 6px 9px 0 rgba(0, 0, 0, .16);
+        max-height: 3.4375rem;
+        padding: 0.6125rem 0;
+
+        .logo svg {
+            height: 2.125rem;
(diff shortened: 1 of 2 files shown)


howardjohn:
Looks great! one small issue onthe main page only, when you scroll to top it flashes the fullsize white header for a split second before becoming blue: https://deploy-preview-9659--preliminary-istio.netlify.app/latest/

rytswd:
Ah thanks for the catch, it looks like there is some other transition setup overriding my update, taking a look at that

rytswd:
OK so I think I got the main page transition fixed, and the Try Istio button a bit smaller when scrolled. If I had more time I'd look more into making Istio logo several pixels lower, and fix up the Try Istio button animation (it looks like the text jumps around a bit) - but I give up on getting dot perfect... At least this looks better than before IMO 😅

craigbox:
Lovely - we'll turn this over to our dev team to fix up!

wbartosz:
I don't think changing the max-height of the buttons (Try Istio) is a good idea. They are concise throughout the app and then squash on the scroll. We could make the header a bit shorter but don't you think it is too tiny now?

wbartosz:
I don't think changing the max-height of the buttons (Try Istio) is a good idea. They are concise throughout the app and then that single CTA one in the header squeezes on scroll.

We could make the header a bit shorter but don't you think it is too tiny now?

rytswd:
@wbartosz Thanks for checking, some points are being discussed in the referenced issue (https://github.com/istio/istio.io/issues/9569), but let me answer all of them.

> I don't think changing the max-height of the buttons (Try Istio) is a good idea.

I originally kept the size of the button, but it was rather large and stood out a bit too much. I personally don't see why "Try Istio" button needs to stand out for every single page you go into, so having it smaller was one approach.

> We could make the header a bit shorter but don't you think it is too tiny now?

As I referenced in the related issue, I find the current size just about right - there are some other sites I use for reference:

- https://helm.sh/docs/ - a bit taller header, but the background texture helps make it feel the space is not wasted. Also, I find Helm is not as complicated and feature rich as Istio, so this rather larger header is OK, but with Istio with so much words throughout the site, I'd want less space wasted by the header
- https://vector.dev/ - very slim header, just enough for folks who come to check out the documentation
- https://argo-cd.readthedocs.io/en/latest/ - about the same height, slightly shorter
- https://fluxcd.io/docs/ - slightly taller
- https://tekton.dev/docs/ - slightly taller, about the same size with Flux CD
- https://grafana.com/docs/grafana/latest/dashboards/ - about the same size when scrolled down to the main content

I think this is about the right size, could be made slightly taller or shorter, but never as big as what we have in the preliminary doc right now. It is about making header easily accessible, but do not disturb users when they are scrolling down and going through the content.

> There's also a scrollbar within the header in mobile view once you start scrolling.

Ah I didn't notice this one, we can put overflow control for that.

craigbox:
@wbartosz has implemented much of the new design.  I asked him to take a go at this based on your work - he has submitted #9687 - [PTAL at the render](https://deploy-preview-9687--preliminary-istio.netlify.app/latest/docs/).

It's taller than yours, but smaller than most others, which gives the "Try Istio" button room to breathe and lets it still be a viable tap target on mobile.  I think it's a good compromise and it's getting good play in #docs on Slack too.

Let me know your thoughts?

rytswd:
Thanks @craigbox , the new PR looks perfectly reasonable, and much better in terms of code / style management. I'm closing this PR in favour of #9687.

PR #9670, opened 2021-05-03
Only 7 columns, not 8 in istio 1.10.0-rc.0 istioctl

jacob-delgado:
Printing out $8 results in an empty line.

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
Is there an issue with changing the output of commands? What's the likelihood of someone coding up a prior lookup of column 8, and now its column 7?

ericvn:
In my case column 8 is `1.10.0-rc.0` on rc.0 and latest build.

ericvn:
However, Going back to release-1.9, it appears that the output is the same, so this isn't a change in behavior from 1.9.

ericvn (approved):

Issue #9668, opened 2021-05-03
values.global.tracer.zipkin.address is deprecated


esnible:
Following instructions at https://preliminary.istio.io/latest/docs/ops/integrations/jaeger/

``` bash
istioctl-1.10.0-rc.0 install \
  --set values.global.tracer.zipkin.address=jaeger-collector.istio-system.svc.cluster.local:9411 \
   && say installed || say failed
```

Istio installs, but produces the message `! values.global.tracer.zipkin.address is deprecated; use meshConfig.defaultConfig.tracing.zipkin.address instead`

esnible:
The docs at https://preliminary.istio.io/latest/docs/ops/integrations/jaeger/ say "This can be configured with `--set values.global.tracer.zipkin.address=<jaeger-collector-address>:9411`" but @bianpengyuan [commented](https://github.com/istio/istio/issues/32596#issuecomment-831429151) that the port should be 14268.

I don't know if 9411 or 14268 is the correct port, but the fix to use the non-deprecated syntax should include the correct port for Jaeger.

douglas-reid:
@bianpengyuan why did the port change? https://github.com/istio/istio/blob/master/samples/addons/jaeger.yaml#L84 has been our sample deployment for serving the zipkin API since forever, and it officially documented as the zipkin support path: https://www.jaegertracing.io/docs/1.22/getting-started/

> 9411 | HTTP | collector | Zipkin compatible endpoint (optional)

bianpengyuan:
I think they just use jaeger collector service with 9411: https://github.com/istio/istio/blob/00082f164c21d8229c80479f7410e4b55d6f65ba/samples/addons/jaeger.yaml#L93 because that seems to be what the doc suggests.

esnible:
I misunderstood the docs.  I thought the instruction to use `--set values.global.tracer.zipkin.address=<jaeger-collector-address>:9411` was for everyone, not just people doing option 2.

Someone (Istio docs, or Jaeger folks) should clean up the doc.  I plan to use my own Jaeger builds in the future, so I need to understand how to do this.

I thought the worst I could do by specifying the tracer address if not needed was a no-op, but it didn't work.  All of the problems I had were caused by doing combinations of `--set values.global.tracer.zipkin.address=` and `--set meshConfig.defaultConfig.tracing.zipkin.address=` with `jaeger-collector.istio-system.svc.cluster.local:<port>`.  When I just did the simple thing,

```bash
[...]

PR #9674, opened 2021-05-03
Use zipkin collector address instead of jaeger to avoid confusion.

bianpengyuan:
There is no need to mention port number. Operator should know which port they should be using for zipkin and jaeger trace report. It has caused some confusions: https://github.com/istio/istio.io/issues/9668 and https://github.com/istio/istio/issues/32026#issuecomment-831561949

### DIFF
--- a/content/en/docs/ops/integrations/jaeger/index.md
+++ b/content/en/docs/ops/integrations/jaeger/index.md
@@ -25,7 +25,7 @@ This will deploy Jaeger into your cluster. This is intended for demonstration on
 
 Consult the [Jaeger documentation](https://www.jaegertracing.io/) to get started. No special changes are needed for Jaeger to work with Istio.
 
-Once Jaeger is installed, you will need to point Istio proxies to send traces to the deployment. This can be configured with `--set values.global.tracer.zipkin.address=<jaeger-collector-address>:9411` at installation time. See the [`ProxyConfig.Tracing`](/docs/reference/config/istio.mesh.v1alpha1/#Tracing) for advanced configuration such as TLS settings.
+Once Jaeger is installed, you will need to point Istio proxies to send traces to the deployment. This can be configured with `--set values.global.tracer.zipkin.address=<jaeger-collector-address>` at installation time. See the [`ProxyConfig.Tracing`](/docs/reference/config/istio.mesh.v1alpha1/#Tracing) for advanced configuration such as TLS settings.
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


douglas-reid:
I'm not sure we want this PR. We are serving the zipkin API, which on most deployments of both services (and in our sample deployments) is on port `9411`. If traffic to port `9411` is not working, then there must be a separate issue (or a backwards incompatibility that was introduced).

bianpengyuan:
> I'm not sure we want this PR. We are serving the zipkin API, which on most deployments of both services (and in our sample deployments) is on port `9411`. If traffic to port `9411` is not working, then there must be a separate issue (or a backwards incompatibility that was introduced).

Jaeger collector service does not use port 9411, which makes the doc confusing. I can remove the jaeger one only, if we want to keep zipkin 9411 documented.

douglas-reid:
> > I'm not sure we want this PR. We are serving the zipkin API, which on most deployments of both services (and in our sample deployments) is on port `9411`. If traffic to port `9411` is not working, then there must be a separate issue (or a backwards incompatibility that was introduced).
> 
> Jaeger collector service does not use port 9411, which makes the doc confusing. I can remove the jaeger one only, if we want to keep zipkin 9411 documented.

It should. The all-in-one should be enabling the optional Zipkin port. That's been our jaeger integration story from 0.1. Did we remove it?

douglas-reid:
> > > I'm not sure we want this PR. We are serving the zipkin API, which on most deployments of both services (and in our sample deployments) is on port `9411`. If traffic to port `9411` is not working, then there must be a separate issue (or a backwards incompatibility that was introduced).
> > 
> > 
> > Jaeger collector service does not use port 9411, which makes the doc confusing. I can remove the jaeger one only, if we want to keep zipkin 9411 documented.
> 
> It should. The all-in-one should be enabling the optional Zipkin port. That's been our jaeger integration story from 0.1. Did we remove it?

Maybe we should update with a warning about enabling the optional port, per our sample config: https://github.com/istio/istio/blob/master/samples/addons/jaeger.yaml#L33-L34 ?

bianpengyuan:
I think the problem is 9411 is not exposed at jaeger-collector service: https://github.com/istio/istio.io/issues/9668#issuecomment-831579695

ericvn:
Any updates on this PR?

ericvn:
Waiting on approval from @douglas-reid

PR #9587, opened 2021-04-23
Add SAP as a consumer

Shegox:
Hello team, 

this PR adds the logo/name of SAP as user of Istio. With the ongoing Polidea revamp of the webpage I wasn't sure where I should add the logo. I opted now for just the [new case studies page](https://preliminary.istio.io/latest/about/case-studies), but if I should add it to the [`community/customers`](https://istio.io/latest/about/community/customers/) page as well, please just let me know.

Our colleagues from the trademark team asks us to include following notice about the usage of the logo, I hope that is fine:

> Purpose: Use of SAP Logo as a user on https://istio.io/latest/about/case-studies
> Logo Identification: SAP Logo
> 
> Actual use of the SAP logo described above commits the Requesting Company to the following:
> - Use of the SAP logo is limited solely to the Purpose
> - Use of the SAP logo must conform to [SAP’s trademark use guidelines](https://www.sap.com/corporate/en/legal/trademark.html#proper-use)
> - The context in which the SAP logo is displayed shall be positive to SAP and its products and services
> 
> SAP hereby consents to the use of its logo for the limited Purpose stated above and Company’s use of the logo will serve as its acknowledgment of the accuracy of the representations upon which this consent relies. SAP may revoke this consent at any time and without notice if Company violates any of the representations or conditions contained in this email. SAP may revoke this notice under any other circumstances with thirty (30) days notice to you.
> SAP reserves all intellectual property rights in and to its trademarks. Company’s use of the SAP trademarks in violation of this consent letter may be subject to legal action under applicable law. SAP’s consent and the logos and/or trademarks to which the consent applies are provided to Company as-is with no warranties of any type.

[ ] Configuration Infrastructure
[x] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

PR #9677, opened 2021-05-04
add Kyma as integration and SAP BTP, Kyma Runtime as provider

Shegox:
This PR adds Kyma as an integration leveraging Istio and SAP BTP, Kyma Runtime as a provider where you get a managed Kyma including Istio.

As both sections are still under constructions this is just a first proposal and I'm happy to adjust it depending on how it should look like. 

The changes are separated out of #9587 in order to allow easier iterations.
//cc @craigbox 

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


data/companies.yml
```diff
@@ -1,9 +1,9 @@
 # TODO: Replace placeholders
 
 providers:
-  - name: "Istio"
-    logo: "/logos/istio.svg"
-    url: "https://istio.io/"
+  - name: "SAP Business Technology Platform, Kyma Runtime"
+    logo: "/logos/sap-btp-kyma-runtime.svg"
```
craigbox, line 5: This should be the SAP logo - the goal of the logo wall is to identify the providers that people will be familiar with.

Otherwise, great stuff - thank you!!

craigbox (changes requested):

data/companies.yml
Shegox: good point, updated in 7250ac2.

craigbox (approved):

PR #9679, opened 2021-05-04
Remove workaround for #32244

frankbu:
Depends on https://github.com/istio/istio/pull/32616

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


content/en/docs/setup/additional-setup/external-controlplane/index.md
```diff
@@ -211,6 +211,8 @@ and installing the sidecar injector webhook configuration on the remote cluster
               value: ""
             - name: VALIDATION_WEBHOOK_CONFIG_NAME
               value: ""
+            - name: EXTERNAL_ISTIOD
```
stevenctl, line 214: https://github.com/istio/istio/blob/688973e58828ffbcff2ccd9eeab41a12527c217a/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml#L151-L152

Would it be better to set `Values.global.externalIstiod`? That will set this env as well as make sure the permissions are set up.

frankbu: Hmm, IIUC `Values.global.externalIstiod` is being overloaded. When we install Istio on the external (control plane) cluster, we don't install the Roles/ClusterRoles, we just install istiod, so `Values.global.externalIstiod` would only affect the istiod `EXTERNAL_ISTIOD` env variable. On the remote clusters we don't install Istiod, so the only thing we get from `Values.global.externalIstiod` are the extra ClusterRole permissions.

So it seems to mean two different things, depending on which cluster's install it's used on:

1. on a control plane cluster, it means the installed istiod supports remote clusters.
2. on a remote cluster it means the cluster is using an external control plane.

I think it might be best to just use it for the second case, and change the istiod deployment template to not use it to set EXTERNAL_ISTIOD:

```
          - name: EXTERNAL_ISTIOD
            value: "false"
```

@stevenctl WDYT?

stevenctl: Ah, yeah that's a bit awkward. SGTM – although setting `env` directly in the docs doesn't seem ideal. Honestly, I'd rather get rid of the feature flag and just turn on those features when we detect a cluster is an istiodless-remote.

stevenctl (approved):

content/en/docs/setup/additional-setup/external-controlplane/index.md
frankbu: I agree, would be great to get rid of the feature flag and just detect in the future.

For now, just removed `Values.global.externalIstiod` for setting of EXTERNAL_ISTIOD in https://github.com/istio/istio/pull/32616

nmittler (approved):

PR #9690, opened 2021-05-06
Add doc test for IstioOperator install

Monkeyanator:
Test the install and update sections of the `istioctl operator` install docs. Doesn't test the upgrade portions since that requires older istioctls

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
@Monkeyanator I think you could safely use curl to download the istioctl version to use for the older install. I'm concerned if we marked the page as tested, when we only test 1/3 of the page (figuring install and then both upgrades).

Monkeyanator:
@ericvn agreed. Maybe we can merge without marking as tested so we have the coverage without the bad signalling?

ericvn (approved):

PR #9706, opened 2021-05-10
Add documentation for gateway installation and management

howardjohn:
This is not 100% ready, but wanted some feedback before proceeding more. Preview: https://deploy-preview-9706--preliminary-istio.netlify.app/latest/docs/setup/additional-setup/gateway/

Open questions:
* Do we want to document non-injection gateway? It would basically be "remove the `injectionTemplate` section and understand there are now these limitations: ...".
* Do we want upgrade here or on its own page?

TODO:
* Fix yaml issues with multipart docs
* Add gateway upgrade sections
* Remove other sections describing gateway install and point here


content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+
+This guide requires the Istio control plane [to be installed](/docs/setup/install/) before proceeding.
+
+## Installing the gateway with injection
+
+To support users that already utilize existing deployment tools, we provide a few different ways to deploy a gateway.
+Each method will result in the same outcome; we recommend choosing the method you are most familiar with.
+If you are not sure which to choose, we suggest the Simple YAML method.
+
+As a security best practice, it is recommended to deploy the gateway in different namespace from the control plane.
```
smawson, line 25:
```suggestion
As a security best practice, it is recommended to deploy the gateway in a different namespace from the control plane.
```

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Prerequisites
+
+This guide requires the Istio control plane [to be installed](/docs/setup/install/) before proceeding.
+
+## Installing the gateway with injection
+
+To support users that already utilize existing deployment tools, we provide a few different ways to deploy a gateway.
+Each method will result in the same outcome; we recommend choosing the method you are most familiar with.
+If you are not sure which to choose, we suggest the Simple YAML method.
```
smawson, line 23: Linkify the "Simple YAML" part? Assuming we can link to that?

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+  selector:
+    matchLabels:
+      istio: ingressgateway
+  template:
+    metadata:
+      annotations:
+        # Enable gateway injection
+        inject.istio.io/templates: gateway
+      labels:
+        # Set a unique label for the gateway. This is required to ensure Gateway's
```
smawson, line 69:
```suggestion
        # Set a unique label for the gateway. This is required to ensure Gateways
```

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+        # can select this workload
+        istio: ingressgateway
+        # Enable gateway injection. If connecting to a revisioned control plane, replace with
```
smawson, line 72: Both the annotation above and label here have a comment saying "enable injection". Maybe clarify why both are needed?

On the annotation we could indicate that it can be used to select different templates if desired.

content/en/docs/setup/install/gateway/index.md
smawson, line 23: Why are we recommending that one? It actually looks the most complicated in terms of what the user needs to manage.

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+$ helm install istio-ingress manifests/charts/gateways/istio-ingress -n istio-ingress -f values.yaml
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< /tabset >}}
+
+## Managing gateways
+
+Below describes how to manage gateways after installation. For more information on their usage, follow
```
smawson, line 182:
```suggestion
The following describes how to manage gateways after installation. For more information on their usage, follow
```

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+the [Ingress](/docs/tasks/traffic-management/ingress/) and [Egress](/docs/tasks/traffic-management/egress/) tasks.
+
+### Gateway selectors
+
+The labels chosen on the gateway pods are important as they are used by Gateway selectors, so it is important your
```
smawson, line 187:
```suggestion
The labels chosen on the gateway pods are used by Gateway selectors, so it is important that your
```

content/en/docs/setup/install/gateway/index.md
howardjohn: Its subjective which is "best", but I prefer it because:
* Its using plain Kubernetes APIs, which are understood by all users, rather than our own (undocumented) Helm values API
* It is the most flexible
* It is the only way we have right now to safely canary upgrade a gateway. The (current unwritten) upgrade section will likely depend on this for some upgrade options
* No extra tools required

I would not be surprised at all if many disagree. But if I was not familiar with helm, I would probably not want to go learn a new tool; If I am familiar with helm the line above recommends choosing what the user is familiar with.

I am ok with changing the recommended path if others have a different preference

smawson: Well if they're using Istio they're already likely using either istioctl or helm right? It seems unlikely they don't have either of those...And both of those are MUCH shorter than the full yaml.

To avoid this we could just drop the sentence and not make a recommendation at all, then the previous sentence just tells them to use what they're familiar with and we're good.

nrjpoddar:
@howardjohn the placement of this guide will be a bit confusing to users. If you follow any of the other installation guides you will end up with a gateway. Should we augment those guides to exclude gateway installations and point users to this new workflow? If we end up marking this feature as experimental, we should highlight its feature status on top of the page.

The gateway topology section might belong to architecture/best practices, Operations/deployment might be a better place to house this section.

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+
+To support users that already utilize existing deployment tools, we provide a few different ways to deploy a gateway.
+Each method will result in the same outcome; we recommend choosing the method you are most familiar with.
+If you are not sure which to choose, we suggest the Simple YAML method.
+
+As a security best practice, it is recommended to deploy the gateway in different namespace from the control plane.
+
+Using the same mechanisms as [Istio sidecar injection](/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection), Istio allows the required
+options for gateways to run to be automatically injected.
+This enables developers to have full control over the gateway deployment, while simplifying operations - when a new upgrade is available, or a configuration has changed, gateway pods can simply be restarted to be updated.
```
nrjpoddar, line 29: What are the customizations that can be done currently when using auto gateway injection and how can users do that?

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+  namespace: istio-ingress
+spec:
+  selector:
+    matchLabels:
+      istio: ingressgateway
+  template:
+    metadata:
+      annotations:
+        # Enable gateway injection
+        inject.istio.io/templates: gateway
```
nrjpoddar, line 67: Is this annotation documented in the reference/annotations section? What are the supported values?

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+      labels:
+        # Set a unique label for the gateway. This is required to ensure Gateway's
+        # can select this workload
+        istio: ingressgateway
+        # Enable gateway injection. If connecting to a revisioned control plane, replace with
+        # `istio.io/rev: <revision-name>`.
+        sidecar.istio.io/inject: "true"
```
nrjpoddar, line 74: We need this label too?

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+    spec:
+      containers:
+      - name: istio-proxy
+        image: auto # The image will automatically update each time the pod starts.
```
nrjpoddar, line 78: 'auto' looks like a magic word here. Where will we document this API?

howardjohn: This is the documentation :-). Its also here: https://preliminary.istio.io/latest/docs/setup/additional-setup/sidecar-injection/#customizing-injection

content/en/docs/setup/install/gateway/index.md
howardjohn: You can either label the entire namespace or label the pod itself. This is how injection works in general, and is not specific to gateway injection. I will make it more clear

content/en/docs/setup/install/gateway/index.md
howardjohn: I did add it in https://github.com/istio/api/pull/1956 but I don't see it on istio.io so I don't know whats going on there.. maybe the reference docs are not updated?

content/en/docs/setup/install/gateway/index.md
howardjohn: Its just a standard Kubernetes pod/deployment. You should be able to set whatever options you want following Kubernetes docs (`kubectl explain pod`)

howardjohn:
> @howardjohn the placement of this guide will be a bit confusing to users. If you follow any of the other installation guides you will end up with a gateway. Should we augment those guides to exclude gateway installations and point users to this new workflow? If we end up marking this feature as experimental, we should highlight its feature status on top of the page.
> 
> The gateway topology section might belong to architecture/best practices, Operations/deployment might be a better place to house this section.

I see what you mean but this seems similar to multicluster, which the comments apply to equally, so it may be good to make them consistent?

content/en/docs/setup/install/gateway/index.md
```diff
@@ -0,0 +1,228 @@
...
+Along with creating a service mesh, Istio allows managing [gateways](/docs/concepts/traffic-management/#gateways),
+which run at the edge of the mesh and allow fine-grained control over traffic entering and leaving the mesh.
+
+Follow this guide to deploy one or more gateways.
+
+## Prerequisites
+
+This guide requires the Istio control plane [to be installed](/docs/setup/install/) before proceeding.
+
+## Installing the gateway with injection
```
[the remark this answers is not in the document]
frankbu: Ah, I see this is a TODO in your description above. My suggestion:

Add a section, before this one, that talks about "Gateway installation options". You could include the material from the doc I referenced above, and then also discuss the shortcomings and that template injection is a better approach. That would be followed by this section, which is an implied recommended approach, even if we don't say that explicitly yet, since the feature is still alpha.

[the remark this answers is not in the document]
frankbu: > I see what you mean but this seems similar to multicluster, which the comments apply to equally, so it may be good to make them consistent? Seems like either both should be under install/, or they should both be split?

I tend to agree that this is not the right place for this doc. The difference is that every other doc actually installs Istio (i.e., starts 1 or more control planes). This is really part of configuring the mesh. The external control plane doc really makes this clear by describing the gateway install steps under "mesh admin steps", while the the rest of the Istio setup is the "mesh operator steps". (As an aside, for the same reason, I think that maybe the External Control Plane installation doc should be moved here.)

That said, I tend to think that maybe the best place for this is under "More Guides". It really is important, so I also think that every other install doc should have one or more clear references to this guide, so users don't miss it.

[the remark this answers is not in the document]
frankbu: Also, if we move this to the "More Guides" section, we could also add a reference on the _index.md page, i.e., after `Choose the guide that best suits your needs and platform.`, add another sentence something like `To install gateways in a mesh, refer to [Installing Gateways](link) for details`.

content/en/docs/setup/install/gateway/index.md
frankbu: Maybe it needs a cherrypick to release-1.10?

content/en/docs/setup/install/gateway/index.md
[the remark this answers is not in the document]
frankbu: https://github.com/istio/istio.io/pull/9727

content/en/docs/setup/install/gateway/index.md
howardjohn: That was it, thanks Frank! I sent a cherrypick

content/en/docs/setup/install/gateway/index.md
howardjohn: I added a note

content/en/docs/setup/install/gateway/index.md
nrjpoddar: I see the new documentation https://preliminary.istio.io/latest/docs/setup/additional-setup/sidecar-injection/#custom-templates-experimental and the annotation getting added but I still don't see what are the templates available in Istio. Is it sidecar and gateway currently?

nrjpoddar (commented):
This naming `inject.istio.io/templates` and `sidecar.istio.io/inject` is a bit confusing.

I wouldn't have expected gateways to require `sidecar.istio.io/inject` annotation as there's no sidecar involved here. It feels like wrong UX to me.

content/en/docs/setup/install/gateway/index.md
nrjpoddar: Yeah, using `sidecar.istio.io/inject` annotation for gateways feels wrong to me..

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,296 @@
...
+      istio-ingressgateway:
+        # Enable gateway injection
+        injectionTemplate: gateway
+{{< /text >}}
+
+Then install using standard `istioctl` commands:
+
+{{< text bash >}}
+$ kubectl create namespace istio-ingress
+$ istioctl install -f ingress.yaml
```
linsun, line 154: I assume the operator and helm approach doesn't require label the namespace Istio-ingress with auto sidecar injection?

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,296 @@
...
+#### Dedicated application gateway
+
+In this model, an application namespace has its own dedicated gateway installation.
+This allows giving full control and ownership to a single namespace.
+This level of isolation can be helpful for critical applications that have strict performance or security requirements.
+
+{{< image width="50%" link="user-gateway.svg" caption="Dedicated application gateway" >}}
+
+Unless there is another load balancer in front of Istio, this typically means that each application will have its own IP address,
+which may complicate DNS configurations.
```
linsun, line 235: @howardjohn for this model, is there any difference in the 3 choices of gateway installation methods?  Are all 3 support this?  I feel the operator and helm had some limitation where something are not customizable?

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,296 @@
...
+### In place upgrade
+
+Because gateways utilize pod injection, new gateway pods that are created will automatically be injected with the latest configuration, which includes the version.
+
+To pick up changes to the gateway configuration, the pods can simply be restarted, using commands such as `kubectl rollout restart deployment`.
+
+If you would like to change the [control plane revision](/docs/setup/upgrade/canary/) in use by the gateway, you can set the `istio.io/rev` label on the gateway Deployment, which will also trigger a rolling restart.
+
+{{< image width="50%" link="inplace-upgrade.svg" caption="In place upgrade in progress" >}}
+
```
linsun, line 248: We should highlight if users are changing gateway to be deployed from Istio-system to Istio-ingress, canary upgrade must be used.

content/en/docs/setup/install/gateway/index.md
frankbu: I think the problem is that the `inject` annotation name is just too specific. It would be better named `proxy.istio.io/inject`. Could we add a new annotation (and maybe deprecate `sidecar.istio.io/inject`)?

howardjohn: Aside from normal API migration pains, it would be extremely difficult to make this change. The mutating webhook already has 8 distinct label selection clauses in order to handle the complex (internal) logic we want to expose to users (in a hopefully simple to understand scheme) while keeping backwards compatibility. It would probably be exponentially worse to try to add another one.

It is a label btw, not an annotation

content/en/docs/setup/install/gateway/index.md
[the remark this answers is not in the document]
howardjohn: I moved it under more guides

content/en/docs/setup/additional-setup/gateway/index.md
howardjohn: None of the methods do. All of them explicitly turn injection on in the deployment. This is following standard injection logic - https://preliminary.istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy. Nothing here is special for gateways.

If they don't want to label the deployment, they can just label the namespace.

content/en/docs/setup/additional-setup/gateway/index.md
howardjohn: That is related to upgrades only. This is more about the API model which doesn't care about install method

content/en/docs/setup/additional-setup/gateway/index.md
howardjohn: A service cannot select pods in another namespace so you cannot really do a cross namespace canary upgrade. If you want to do this you need a higher level control like

content/en/docs/setup/install/gateway/index.md
frankbu: I see your point. The name's not ideal, but it is what it is.

content/en/docs/setup/install/_index.md
```diff
@@ -12,6 +12,8 @@ test: n/a
 If you need help choosing, refer to our
 [which Istio installation method should I use?](/about/faq/#install-method-selection) FAQ page.
 
+To install gateways in a mesh, refer to [Installing Gateways](/docs/setup/additional-setup/gateway) for details
```
frankbu, line 15:
```suggestion
To install gateways in a mesh, refer to [Installing Gateways](/docs/setup/additional-setup/gateway) for details.
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,296 @@
...
+title: Installing Gateways
+description: Install and customize Istio Gateways.
+weight: 40
+keywords: [install,gateway,kubernetes]
+owner: istio/wg-environments-maintainers
+test: no
+---
+
+Along with creating a service mesh, Istio allows managing [gateways](/docs/concepts/traffic-management/#gateways),
+which run at the edge of the mesh and allow fine-grained control over traffic entering and leaving the mesh.
```
frankbu, line 11:
```suggestion
Along with creating a service mesh, Istio allow you to manage [gateways](/docs/concepts/traffic-management/#gateways),
which run at the edge of the mesh, providing fine-grained control over traffic entering and leaving the mesh.
```

howardjohn:
@frankbu @nrjpoddar @smawson I have (attempted to) addressed all the comments and made the changes discussed in the TOC, so this is ready to go from my perspective. Let me know what you think

content/en/docs/setup/additional-setup/gateway/high-level-canary.svg
```diff
@@ -0,0 +1,3 @@
...
...m="translate(-0.5 -0.5)"><switch><foreignObject style="overflow: visible; text-align: left;" pointer-events="none" width="100%" height="100%" requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility"><div xmlns="http://www.w3.org/1999/xhtml" style="display: flex; align-items: unsafe center; justify-content: unsafe center; width: 258px; height: 1px; padding-top: 30px; margin-left: 91px;"><div style="box-sizing: border-box; font-size: 0; text-align: center; "><div style="display: inline-block; font-size: 12px; font-family: Helvetica; color: #000000; line-height: 1.2; pointer-events: all; white-space: normal; word-wrap: normal; ">External Load Balancer<br />OR<br />DNS configuration</div></div></div></foreignObject><text x="220" y="34" fill="#000000" font-family="Helvetica" font-size="12px" text-anchor="middle">External Load Balancer...</text></switch></g></g><switch><g requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility"/><a transform="translate(0,-5)" xlink:href="https://www.diagrams.net/doc/faq/svg-export-text-problems" target="_blank"><text text-anchor="middle" font-size="10px" x="50%" y="100%">Viewer does not support full SVG 1.1</text></a></switch></svg>
```
frankbu, line 3: Lint and netlify are failing. Something about this image is causing the image shortcode to get messed up here: https://github.com/istio/istio.io/blob/692f67753a7a4de42c37724c575ae836edadba8d/layouts/shortcodes/image.html#L91

Maybe it's a bug with negative viewBox values? `viewBox="-0.5 -0.5 421 431"`

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+only the [Kubernetes YAML](/docs/setup/additional-setup/gateway/#tabset-docs-setup-additional-setup-gateway-1-2-tab) method is supported for this upgrade method.
+{{< /warning >}}
+
+### Canary upgrade with external traffic shifting (advanced)
+
+A variant of this the [Canary upgrade](#canary-upgrade) approach is to shift the traffic between the versions using a high level construct outside Istio, such as an external load balancer or DNS.
+
+{{< image width="50%" link="high-level-canary.svg" caption="Canary upgrade in progress with external traffic shifting" >}}
+
+This offers fine-grained control, but may be unsuitable or complicated to setup in some environments.
```
frankbu, line 307:
```suggestion
This offers fine-grained control, but may be unsuitable or complicated to set up in some environments.
```

content/en/docs/setup/additional-setup/gateway/high-level-canary.svg
howardjohn: yeah I have to manually change to `0 0 ...`. Not sure why. I thought I did, let me fix it

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+If you use another load balancer in front of Istio, you may also use that to control the traffic distribution.
+
+{{< warning >}}
+Because other installation methods bundle the gateway Service, which controls its external IP address, with the gateway Deployment,
+only the [Kubernetes YAML](/docs/setup/additional-setup/gateway/#tabset-docs-setup-additional-setup-gateway-1-2-tab) method is supported for this upgrade method.
+{{< /warning >}}
+
+### Canary upgrade with external traffic shifting (advanced)
+
+A variant of this the [Canary upgrade](#canary-upgrade) approach is to shift the traffic between the versions using a high level construct outside Istio, such as an external load balancer or DNS.
```
frankbu, line 303:
```suggestion
A variant of the [Canary upgrade](#canary-upgrade) approach is to shift the traffic between the versions using a high level construct outside Istio, such as an external load balancer or DNS.
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+To support users that already utilize existing deployment tools, we provide a few different ways to deploy a gateway.
+Each method will result in the same outcome; we recommend choosing the method you are most familiar with.
+
+As a security best practice, it is recommended to deploy the gateway in a different namespace from the control plane.
+
+Using the same mechanisms as [Istio sidecar injection](/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection), Istio allows the required
+options for gateways to run to be automatically injected.
+This enables developers to have full control over the gateway deployment, while simplifying operations - when a new upgrade is available, or a configuration has changed, gateway pods can simply be restarted to be updated.
+This makes the experience of operating a gateway deployment the same as operating sidecars.
```
frankbu, line 29:
```suggestion
Using the same mechanisms as [Istio sidecar injection](/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection), the Envoy proxy configuration for gateways can similarly be auto-injected.
This gives developers full control over the gateway deployment, while also simplifying operations. When a new upgrade is available, or a configuration has changed, gateway pods can be updated by simply restarting them.
This makes the experience of operating a gateway deployment the same as operating sidecars.

To support users with existing deployment tools, Istio provides a few different ways to deploy a gateway.
Each method will produce the same result.
Choose the method you are most familiar with:

{{< tip >}}
As a security best practice, it is recommended to deploy the gateway in a different namespace from the control plane.
{{< /tip >}}
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+  name: istio-ingressgateway-sds
+subjects:
+- kind: ServiceAccount
+  name: default
+{{< /text >}}
+
+{{< tip >}}
+This example shows a bare minimum example to get a gateway running. For production usage, additional
+configuration such as Horizontal Pod Autoscaler, Pod Disruption Budget, and resource requests/limits are recommended.
+{{< /tip >}}
```
frankbu, line 171:
```suggestion
This example shows the bare minimum needed to get a gateway running. For production usage, additional
configuration such as Horizontal Pod Autoscaler, Pod Disruption Budget, and resource requests/limits are recommended.

{{< tip >}}
The `sidecar.istio.io/inject` label on the pod is used in this example to enable injection. Just like application sidecar injection, this can instead be controlled at the namespace level.
See [Controlling the injection policy](/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy) for more information.
{{< /tip >}}
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+$ kubectl create namespace istio-ingress
+$ kubectl apply -f ingress.yaml
+{{< /text >}}
+
+{{< tip >}}
+The example above enables injection on the pod. This can also be controlled at the namespace level.
+See [Controlling the injection policy](/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy) for more information.
+{{< /tip >}}
+
+{{< /tab >}}
```
frankbu, line 185:
```suggestion
{{< /tab >}}
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+The following describes how to manage gateways after installation. For more information on their usage, follow
+the [Ingress](/docs/tasks/traffic-management/ingress/) and [Egress](/docs/tasks/traffic-management/egress/) tasks.
+
+### Gateway selectors
+
+The labels chosen on the gateway pods are used by Gateway selectors, so it is important that your
+Gateway objects are configured to match these labels.
+
+For example, in the deployments above we set the `istio=ingressgateway` label on the pods.
+To apply a Gateway to these deployments, we would select the same label:
```
frankbu, line 200:
```suggestion
The labels on a gateway deployment's pods are used by `Gateway` configuration resources, so it's important that
your `Gateway` selector matches these labels.

For example, in the above deployments, the `istio=ingressgateway` label is set on the gateway pods.
To apply a `Gateway` to these deployments, you need to select the same label:
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+  selector:
+    istio: ingressgateway
+...
+{{< /text >}}
+
+### Gateway deployment topologies
+
+Depending on your mesh configuration and use cases, you may wish to deploy gateways in different ways.
+Below shows a few different patterns for gateway deployments.
+Note that multiple options can be utilized within the same cluster.
```
frankbu, line 217:
```suggestion
A few different gateway deployment patterns are shown below.
Note that more than one of these patterns can be used within the same cluster.
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+#### Shared gateway
+
+In this model, a single centralized gateway is used by many applications, possibly across many namespaces.
+Gateway(s) in the `ingress` namespace delegate ownership of routes to application namespaces, but retain control over TLS configuration.
+
+{{< image width="50%" link="shared-gateway.svg" caption="Shared gateway" >}}
+
+This model works well when you have many applications you want to expose externally, as they are able to use shared infrastructure.
+It also works well in models that have the same domain or TLS certificates shared by many applications.
```
frankbu, line 227:
```suggestion
It also works well in use cases that have the same domain or TLS certificates shared by many applications.
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+To pick up changes to the gateway configuration, the pods can simply be restarted, using commands such as `kubectl rollout restart deployment`.
+
+If you would like to change the [control plane revision](/docs/setup/upgrade/canary/) in use by the gateway, you can set the `istio.io/rev` label on the gateway Deployment, which will also trigger a rolling restart.
+
+{{< image width="50%" link="inplace-upgrade.svg" caption="In place upgrade in progress" >}}
+
+### Canary upgrade (advanced)
+
+This upgrade method depends on using [control plane revisions](/docs/setup/upgrade/canary/).
```
frankbu, line 254:
```suggestion
{{< warning >}}
This upgrade method depends on control plane revisions, and therefore can only be used in conjunction with
[control plane canary upgrade](/docs/setup/upgrade/canary/).
{{< /warning >}}
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+If you would like to more slowly control a rollout of a new revision, we can run multiple versions of the gateway deployment.
+For example, if I want to do a controlled rollout of the `canary` revision, create a copy of your current deployment with the `istio.io/rev=canary` label set:
```
frankbu, line 257:
```suggestion
If you would like to more slowly control the rollout of a new control plane revision, you can run multiple versions of a gateway deployment.
For example, if you want to roll out a new revision, `canary`, create a copy of your gateway deployment with the `istio.io/rev=canary` label set:
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+      labels:
+        istio: ingressgateway
+        istio.io/rev: canary # Set to the control plane revision you want to deploy
+    spec:
+      containers:
+      - name: istio-proxy
+        image: auto
+{{< /text >}}
+
+When this is created, we will now have two versions of the gateway, both selected by the same Service:
```
frankbu, line 282:
```suggestion
When this is deployment is created, you will then have two versions of the gateway, both selected by the same Service:
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+{{< text bash >}}
+$ kubectl get endpoints -o "custom-columns=NAME:.metadata.name,PODS:.subsets[*].addresses[*].targetRef.name"
+NAME                   PODS
+istio-ingressgateway   istio-ingressgateway-788854c955-8gv96,istio-ingressgateway-canary-b78944cbd-mq2qf
+{{< /text >}}
+
+{{< image width="50%" link="canary-upgrade.svg" caption="Canary upgrade in progress" >}}
+
+Unlike when deploying your own applications, you cannot utilize [Istio traffic shifting](/docs/tasks/traffic-management/traffic-shifting/) to control, as the traffic to the gateway is coming from external clients that we do not control.
```
frankbu, line 292:
```suggestion
Unlike application services deployed inside the mesh, you cannot use [Istio traffic shifting](/docs/tasks/traffic-management/traffic-shifting/) to distribute the traffic between the gateway versions because their traffic is coming directly from external clients that Istio does not control.
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+only the [Kubernetes YAML](/docs/setup/additional-setup/gateway/#tabset-docs-setup-additional-setup-gateway-1-2-tab) method is supported for this upgrade method.
+{{< /warning >}}
+
+### Canary upgrade with external traffic shifting (advanced)
+
+A variant of the [Canary upgrade](#canary-upgrade) approach is to shift the traffic between the versions using a high level construct outside Istio, such as an external load balancer or DNS.
+
+{{< image width="50%" link="high-level-canary.svg" caption="Canary upgrade in progress with external traffic shifting" >}}
+
+This offers fine-grained control, but may be unsuitable or complicated to set up in some environments.
```
frankbu, line 307:
```suggestion
This offers fine-grained control, but may be unsuitable or overly complicated to set up in some environments.
```

content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -0,0 +1,307 @@
...
+{{< /text >}}
+
+{{< image width="50%" link="canary-upgrade.svg" caption="Canary upgrade in progress" >}}
+
+Unlike when deploying your own applications, you cannot utilize [Istio traffic shifting](/docs/tasks/traffic-management/traffic-shifting/) to control, as the traffic to the gateway is coming from external clients that we do not control.
+Instead, you can control the distribution of traffic by the number of replicas of each deployment.
+If you use another load balancer in front of Istio, you may also use that to control the traffic distribution.
+
+{{< warning >}}
+Because other installation methods bundle the gateway Service, which controls its external IP address, with the gateway Deployment,
```
frankbu, line 297:
```suggestion
Because other installation methods bundle the gateway `Service`, which controls its external IP address, with the gateway `Deployment`,
```

howardjohn:
thanks for the comments @frankbu , this is updated

frankbu (approved):

frankbu:
@googlebot I consent.

PR #9710, opened 2021-05-11
Add security grid, release grid and general list templates to new News layout

craigbox:
[X] Docs


ericvn:
- First column, Disclosure, is not wide enough on the Security page.
- The releases page does not have a way to get to the `Announcing 1.9.0` page.

ericvn (approved):

PR #9775, opened 2021-05-18
move docs menu to the 2nd item

shamsher31:
I think documentation is the most widely used menu item on the list and it should be moved as early as possible to quickly access it. Not sure if there is a standard way to keep it at the end of the menu, but I saw most of the websites keep it at the start of the menu list.
 
Old menu items:

old-menu

New menu items

new-menu

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -165,31 +165,31 @@ disableAliases = true
       weight = 6
 
   [[menu.main]]
+    identifier = "docs"
+    name = "Documentation"
+    title = "Documentation"
+    url = "/docs/"
+    weight = 2
+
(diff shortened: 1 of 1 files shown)


craigbox:
Thanks for the PR, Shamsher.   Docs is on the right hand end for exactly that reason of discoverability. 🙂.  Our UX team suggested it's more obvious where it is than being hidden in the middle of the menu.  It's also tied to "search" which is right next to it.

We can revisit this later if you wish - we are looking to what we can do with the UX of the docs later this year, pending staffing - but right now, hours before the launch, it would make more sense to go live with what we have planned.

PR #9779, opened 2021-05-18
Update master after 1.10 release

ericvn:
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


davidhauck (approved):

ericvn:
Need to update the istio test reference as well since the istioctl analyzer test uses the version output from istioctl, which when using the 1.10 test ref will return 1.10, and it's now expecting 1.11.

ericvn:
Curious what other test issues there will be. I do expect some additional istioctl analyzer output may need to be updated in the docs as well.

ericvn:
https://github.com/istio/istio/issues/32926 opened for a set of failing tests.

PR #9784, opened 2021-05-18
Update banner to animate on scroll

rytswd:
Fix banner to move up when scroll spy updates the header height.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/src/sass/misc/_banners.scss
+++ b/src/sass/misc/_banners.scss
@@ -1,6 +1,12 @@
 .banner-container {
     display: none;
     top: $headerHeight;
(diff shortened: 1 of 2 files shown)


craigbox (approved):
LGTM. Who knows how this might work if we ever had more than one banner, but we didn't plan for that in the new design 😁

brian-avery (approved):
Much better. Thanks :)

PR #9797, opened 2021-05-19
Remove faq/applications

howardjohn:
This documentation was all about working around networking which are all
resolved automatically by
https://istio.io/latest/blog/2021/upcoming-networking-changes/


linsun (approved):
Thanks! I was going to remove it :)

linsun:
+1 I like that @craigbox

content/en/about/faq/traffic-management/statefulsets.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Can Istio be used with StatefulSets and headless Services?
```
ericvn, line 2: I think the capitalization of headless should match services. I'm in favor of uppercasing the `h`.

howardjohn: From k8s docs: "In this case, you can create what are termed "headless" Services". "headless" is not a proper noun while Services and StatefulSets are referring to the object names?

ericvn: Interesting as I found k8s docs like, https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/, that have `StatefulSets currently require a Headless Service to be responsible...`

I'll take your comment as a guide.

content/en/about/faq/traffic-management/statefulsets.md
```diff
@@ -0,0 +1,6 @@
...
+weight: 30
```
ericvn, line 3: Can we update this to 25 or 35, just so we aren't overloading 30?

howardjohn: updated

ericvn (approved):

PR #9829, opened 2021-05-21
Add BlaBlaCar company logo

etifontaine:
Add BlaBlaCar, we have been using Istio in production for more than 2 years.
It has help us to easily migrate from on-prem to the Cloud, and support more than 90 millions users worldwide.

We have made contributions to Istio and we would like to be part of this list.

https://github.com/blablacar

Thank you :) 

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


craigbox (approved):

ericvn:
Note that the test will continue to fail. I will have a PR shortly which should fix the failure, and allow the test to pass.

PR #9843, opened 2021-05-25
Add update_test_reference target and cleanup old target

ericvn:
update_examples target removed in https://github.com/istio/istio.io/pull/6105

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


Makefile.core.mk
```diff
@@ -140,7 +140,10 @@ netlify: netlify_install
 update_ref_docs:
 	@scripts/grab_reference_docs.sh $(SOURCE_BRANCH_NAME)
 
-update_all: update_ref_docs update_examples
```
davidhauck, line 143: Do we not need update_examples any more?

ericvn: There is no target, it was removed in an earlier PR (noted above). And I removed the script called in that target.

davidhauck (approved):

PR #9890, opened 2021-06-07
remove namespce for cluster scope resources.

morvencao:
[ ] Docs
[ X ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
@frankbu Thoughts on this change as I know you were working on the tests a short time ago? The updated line is run in the test.sh so I believe this change is OK.

frankbu (approved):
I agree this is better. Since it's a cluster-scoped resource, it's in every namespace, so explicitly using -n external-istiod is somewhat confusing.

ericvn (approved):

ericvn:
circuit breaker test is still flakey.

Issue #9766, opened 2021-05-17
Remove [Experimental] from page titles


craigbox:
Experimental content is flagged inline on the page.  We should remove the bulky [Experimental] tag from `title` and `linktitle` metadata and look to flag it with something like a typographical dagger (†).

ericvn:
Where would one add the footnote/description for the dagger? Is this something we need to put on each page?

johnma14:
I believe we would have to add such a note on each page where the dagger shows. It could be a text that appears when we hover over the dagger.

craigbox:
For the sidebar: It would be worth having the dagger have hover text.
For the pages themselves: I don't think it needs to be displayed in the title, as each page will already have a note on it saying it's experimental?

johnma14:
That sounds good to me. We should be able to do this on the sidebar with a dagger.

PR #9895, opened 2021-06-07
Flag experimental pages with dagger

ChristinaMak:
Addresses https://github.com/istio/istio.io/issues/9766

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
Looking at the deploy/netlify output (click on `Details`), the &dagger is not substituted if that was expected.

craigbox:
I think it would be better to have a flag in the front patter that says something like

```
status: experimental
```

and rendered a dagger - only in the sidebar - if so.

layouts/partials/navigation_level.html
```diff
@@ -28,32 +28,40 @@
         {{ $pageLocation := (path.Dir (path.Dir .File.Dir)) }}
         {{ if and (eq $parentDir $pageLocation) (not .Params.draft) }}
             {{ $desc := .Description }}
+            {{ $linktitle := .LinkTitle }}
+            {{ if isset .Params "status" }}
+                {{ $status := .Params.status }}
+                {{ if in $status "experimental" }}
```
johnma14, line 34: So for experimental feature we use a dagger, what about for status: beta/experimental. I am not even sure what that status means. @craigbox any thoughts on that one?

content/en/docs/ops/configuration/extensibility/wasm-module-distribution/index.md
```diff
@@ -8,6 +8,7 @@ aliases:
 keywords: [extensibility,Wasm,WebAssembly]
 owner: istio/wg-policies-and-telemetry-maintainers
 test: no
+status: experimental
```
johnma14, line 11: 'E' needs to be upper case

content/en/docs/tasks/observability/distributed-tracing/configurability/index.md
```diff
@@ -1,10 +1,11 @@
 ---
-title: Configurability [Beta/Experimental]
-description: How to configure tracing options (beta/experimental).
+title: Configurability
+description: How to configure tracing options.
 weight: 60
 keywords: [telemetry,tracing]
 owner: istio/wg-policies-and-telemetry-maintainers
 test: no
+status: beta/experimental
```
johnma14, line 8: 'B' for beta and 'E' for experimental needs to be upper case.

layouts/partials/navigation_level.html
craigbox: I think some parts of the feature are Beta and some are experimental.

I think you use the same symbol for anything and people then mouse-over to see the status.
That might lead to a discussion on if we should do that for Beta APIs too; I suspect the answer is no.

craigbox:
LGTM, nicely done.  Bikeshedding time, perhaps an asterisk without a space, or a space before the dagger?
Either one encourages you to think "What's that", and it's great that the mouseover now adds (Experimental).

layouts/partials/navigation_level.html
```diff
@@ -28,32 +28,40 @@
         {{ $pageLocation := (path.Dir (path.Dir .File.Dir)) }}
         {{ if and (eq $parentDir $pageLocation) (not .Params.draft) }}
             {{ $desc := .Description }}
+            {{ $linktitle := .LinkTitle }}
+            {{ if isset .Params "status" }}
+                {{ $status := .Params.status }}
+                {{ if in $status "Experimental" }}
+                    {{ $linktitle = printf "%s%s" ($linktitle) "*" }}
```
johnma14, line 35: Can we also add a space between the text and the asterisk? Thanks

layouts/partials/navigation_level.html
```diff
@@ -28,32 +28,40 @@
         {{ $pageLocation := (path.Dir (path.Dir .File.Dir)) }}
         {{ if and (eq $parentDir $pageLocation) (not .Params.draft) }}
             {{ $desc := .Description }}
+            {{ $linktitle := .LinkTitle }}
+            {{ if isset .Params "status" }}
+                {{ $status := .Params.status }}
+                {{ if in $status "Experimental" }}
```
johnma14, line 34: @ChristinaMak I agree with @craigbox comments that we can use the the same symbol for any status - Experimental/Alpha/Beta (especially since we are including what it means in the description in the hover text). In that case, we can remove this check.

layouts/partials/navigation_level.html
```diff
@@ -31,9 +31,7 @@
             {{ $linktitle := .LinkTitle }}
             {{ if isset .Params "status" }}
                 {{ $status := .Params.status }}
-                {{ if in $status "Experimental" }}
-                    {{ $linktitle = printf "%s%s" ($linktitle) "*" }}
-                {{ end }}
+                {{ $linktitle = printf "%s%s" ($linktitle) " *" }}
```
johnma14, line 34: this should be: {{ $linktitle = printf "%s %s" ($linktitle) "*" }}

johnma14:
@ChristinaMak I verified the changes and they look good to me. Thanks for addressing the comments. I just found one more thing we need to verify. Under `Operations-> Configuration-> Traffic Management -> Configuring Gateway Network Topology (Alpha)`, I think we need to update this one also to move the word `Alpha` out of the title and put it under status field in the front matter. Thoughts @craigbox ?

craigbox:
> 2. Under `Operations-> Configuration-> Traffic Management -> Configuring Gateway Network Topology (Alpha)`, I think we need to update this one also to move the word `Alpha` out of the title and put it under status field in the front matter. Thoughts @craigbox ?

Seems reasonable!

johnma14 (approved):
LGTM. Thank you @ChristinaMak for the changes.

craigbox (approved):

PR #9909, opened 2021-06-09
fix for nginx deployment yaml

razeel-mohammed:
Egress Gateways with TLS Origination

As the path /etc/nginx already exists in the container the volume mount to the same location fails and the pod goes to crashloop. To get rid of this issue, mount the nginx.config to a separate location and copy the it to the nginx folder during the pod start followed by a nginx reload.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination-sds/index.md
+++ b/content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination-sds/index.md
@@ -182,11 +182,19 @@ to hold the configuration of the NGINX server:
           containers:
           - name: my-nginx
             image: nginx
+            lifecycle:
+              postStart:
+                exec:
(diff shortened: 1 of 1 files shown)


craigbox:
Hi @razeel-mohammed,

Thanks for identifying this.  It looks like it was a valuable contribution.  I see you closed the branch; would you mind if I asked it it was a result of the hoops that you were asked to jump through to submit it, or something else?

Thanks,
Craig

razeelmohammed:
@craigbox I am sorry I totally missed your comment. I used my wrong github handle to raise this PR that's  the reason why I closed it. I am glad that you found the PR useful.

PR #9916, opened 2021-06-10
Fix config_source for ECDS

dgn:
Fixes https://github.com/istio/istio/issues/32415

### DIFF
--- a/content/en/docs/reference/config/networking/envoy-filter/index.html
+++ b/content/en/docs/reference/config/networking/envoy-filter/index.html
@@ -340,12 +340,8 @@ spec:
         name: my-wasm-extension # This must match the name above
         config_discovery:
           config_source:
-            api_config_source:
-              api_type: GRPC
-              transport_api_version: V3
-              grpc_services:
-              - envoy_grpc:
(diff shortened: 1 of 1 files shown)


ericvn (changes requested):
Note that this file is imported from another repo and is not changed here. The change should start here: https://github.com/istio/api/blob/release-1.10/networking/v1alpha3/envoy_filter.proto#L365

Starting with the main branch change, it will be updated in preliminary.istio.io after a daily reference doc update. Then the cherry-pick to the 1.10 branch can be made. It would then be pulled into the istio.io repo and istio.io when the next patch release is made.

dgn:
> Note that this file is imported from another repo and is not changed here. The change should start here: https://github.com/istio/api/blob/master/networking/v1alpha3/envoy_filter.proto
> 
> Starting with the main branch change, it will be updated in preliminary.istio.io after a daily reference doc update. Then the cherry-pick to the 1.10 branch can be made. It would then be pulled into the istio.io repo and istio.io when the next patch release is made.

Thanks, should have looked more closely :) Will submit an equivalent PR to api

PR #9957, opened 2021-06-21
update zh enableAutoMtls default true

yufan022:
https://github.com/istio/istio.io/blob/296c1ba498af93e9fe41f0b3fa8fe1e73e176846/content/en/docs/reference/config/istio.mesh.v1alpha1/index.html#L277
in en docs, enableAutoMtls default is true.

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/zh/docs/reference/config/istio.mesh.v1alpha1/index.html
+++ b/content/zh/docs/reference/config/istio.mesh.v1alpha1/index.html
@@ -667,7 +667,7 @@ No
 <td><code><a href="https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#boolvalue">BoolValue</a></code></td>
 <td>
 <p>This flag is used to enable mutual TLS automatically for service to service communication
-within the mesh, default false.
+within the mesh, default true.
 If set to true, and a given service does not have a corresponding DestinationRule configured,
 or its DestinationRule does not have TLSSettings specified, Istio configures client side
 TLS configuration appropriately. More specifically,


ericvn:
This file on the English side is copied from the istio/api repository, and nothing is typically updated in the istio.io repository. Looking at the English doc, https://github.com/istio/istio.io/blob/master/content/en/docs/reference/config/istio.mesh.v1alpha1/index.html#L277, it seems this is correct there. I'm not sure how translation works in this case, but is there a reason that this document is not just a copy of the English document?

yufan022:
> This file on the English side is copied from the istio/api repository, and nothing is typically updated in the istio.io repository. Looking at the English doc, https://github.com/istio/istio.io/blob/master/content/en/docs/reference/config/istio.mesh.v1alpha1/index.html#L277, it seems this is correct there. I'm not sure how translation works in this case, but is there a reason that this document is not just a copy of the English document?

I'm not sure how it's translated, @rootsongjc Do you know how it works?

yufan022:
@googlebot I signed it!

SataQiu (commented):
/ok-to-test

kebe7jun:
> This file on the English side is copied from the istio/api repository, and nothing is typically updated in the istio.io repository. Looking at the English doc, https://github.com/istio/istio.io/blob/master/content/en/docs/reference/config/istio.mesh.v1alpha1/index.html#L277, it seems this is correct there. I'm not sure how translation works in this case, but is there a reason that this document is not just a copy of the English document?

Maybe we should keep the same sync as the English document's, I'll follow up to implement it.

kebe7jun (approved):

PR #9967, opened 2021-06-23
Remove outdated usage

hzxuzhonghu:
Currently we donnot need ServicEntry for multicluster mesh anymore

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


craigbox:
Is it right to say you cannot use this any more, or you just don't need to in the most common case?

hzxuzhonghu:
https://istio.io/latest/docs/setup/install/multicluster/  

Se is not necessary for the multicluster models listed here. Previously it is needed for multi-controlplane mesh collaborated with istio-coredns, which is also deprecated

craigbox (approved):

PR #10010, opened 2021-06-28
Add table of contents test status

brian-avery:
This adds a table of contents test type to indicate documentation that contains only a table of contents and no other documentation (i.e. top level document listings for a section). It also extends `make test_status` in order to account for these documents.

This is useful in figuring out which tests we can skip in testing.


ericvn (approved):

ericvn:
@kebe7jun Can you please take a look?

kebe7jun (approved):

PR #10015, opened 2021-06-29
add explicit call out about the subtle difference between empty vs omitted

rsalmond:
[X ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

At first glance, a quick reading of this example the highlighted description text below appears to contradict it.

image

This change adds an explicit call out describing the subtle but crucial difference between "omitted" and "empty".

### DIFF
--- a/content/en/docs/reference/config/security/authorization-policy/index.html
+++ b/content/en/docs/reference/config/security/authorization-policy/index.html
@@ -234,6 +234,9 @@ one rule matches the request.</p>
 <p>If not set, the match will never occur. This is equivalent to setting a
 default of deny for the target workloads.</p>
 
+<p>If set but left empty, the match will always occur. This is equivalent to
+setting a default of allow for the target workloads.</p>
+
 </td>
 <td>
 No


craigbox:
/ok-to-test

I expect this has to be made somewhere different, as it's generated into here?  @ericvn

rsalmond:
Oh! I should have looked harder, I see it [over here](https://github.com/istio/api/blob/2af152115f81571984b4146e93274f82ecc73f3e/security/v1beta1/authorization_policy.proto#L244-L245). Shall I close this and cut it over there instead?

ericvn (changes requested):
This PR is trying to change a file which is automatically pulled from another repo. Changes in this file will be reverted on the next automated pull.

The change should be made after the same section here: https://github.com/istio/api/blob/master/security/v1beta1/authorization_policy.proto#L244-L245

The change in that file will be generated into the file pulled into the `master branch here which is reflected in preliminary.istio.io. The istio/api PR can be cherry-picked to the release-1.10 branch and the change will be reflected in istio.io when the next patch release happens.

rsalmond:
Thanks for the heads up @ericvn, closing in favour of istio/api#2038

PR #10032, opened 2021-07-03
Add the information that you can concatenate CA certs

ldemailly:
Add the information that you can concatenate CA certs if you want to accept MTLS from client providing certificate signed by different CAs

[x ] Docs
[ x] Installation
[ ] Networking
[ ] Policies and Telemetry
[ x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
@@ -432,7 +432,7 @@ Istio supports reading a few different Secret formats, to support integration wi
 * A TLS Secret with keys `tls.key` and `tls.crt`, as described above. For mutual TLS, a `ca.crt` key can be used.
 * A generic Secret with keys `key` and `cert`. For mutual TLS, a `cacert` key can be used.
 * A generic Secret with keys `key` and `cert`. For mutual TLS, a separate generic Secret named `<secret>-cacert`, with a `cacert` key. For example, `httpbin-credential` has `key` and `cert`, and `httpbin-credential-cacert` has `cacert`.
-* The `cacert` entry can be a CA bundle: concatenated individual CA certificates.
+* The `cacert` can be a CA bundle consisting of concatenated individual CA certificates.
 
 ### SNI Routing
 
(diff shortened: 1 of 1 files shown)


ldemailly:
Preview: https://deploy-preview-10032--istio.netlify.app/latest/docs/tasks/traffic-management/ingress/secure-ingress/#key-formats

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
```diff
@@ -432,6 +432,7 @@ Istio supports reading a few different Secret formats, to support integration wi
 * A TLS Secret with keys `tls.key` and `tls.crt`, as described above. For mutual TLS, a `ca.crt` key can be used.
 * A generic Secret with keys `key` and `cert`. For mutual TLS, a `cacert` key can be used.
 * A generic Secret with keys `key` and `cert`. For mutual TLS, a separate generic Secret named `<secret>-cacert`, with a `cacert` key. For example, `httpbin-credential` has `key` and `cert`, and `httpbin-credential-cacert` has `cacert`.
+* The `cacert` entry can be a CA bundle: concatenated individual CA certificates.
```
ericvn, line 435: Maybe, "The `cacert` entry can be a CA bundle consisting of concatenated individual CA certificates."

ericvn:
@istio/wg-security-maintainers Thoughts?

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
ldemailly: Thanks, changed. added "value" for clarity too but maybe it should be

`cacert` key's value can be...

I put so far

`cacert` key value can be...

problem with english as a second language, is I'm not sure, lmk

ldemailly:
ps: I tested this using 2 CAs and btw it's nice because you can add/remove without errors on existing traffic

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
shankgan: nit: can we clarify that these are "individual Root CA certificates"

ldemailly: It should work with any CA, why wouldn't it for intermediate vs root? as long as envoy can verify the signer

shankgan: agreed, but it would it be a good idea to add a test to "tests/integration/security/sds_ingress/" before we claim support for this in a release-branch? Not aware of any other scenarios where we pass an intermediate cert to envoy as a TLS trust anchor; please enlighten me if you know of any

ldemailly: here is an unintegrated test (the test I used to check that indeed it's working and CAs can be changed without errors):

https://github.com/istio/istio/pull/33880/files

ldemailly:
can we get the doc in even without an integrated test; as it's very useful information to have and 1 line of clarification

JimmyCYJ (approved):

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
ldemailly: I tested with intermediate CAs and that works as well but you need to put the whole chain in the cacert (ie both the intermediate and the root) - see the MR above for the test

shankgan: LGTM to make change

craigbox (approved):

PR #10038, opened 2021-07-07
fix ratelimt envoyfilter crd

zirain:
accroding [ratelimit config](https://github.com/envoyproxy/envoy/blob/d3a90127f8c408651e8ab44e961408ca1fd972b4/source/extensions/filters/http/ratelimit/config.cc#L27) timeout should use envoy::extensions::filters::http::ratelimit::v3::RateLimit.timeout

otherwise there will be some grpcstatus 14 error during high qps

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


ericvn:
@istio/wg-policies-and-telemetry-maintainers Would you provide feedback if this change is correct. Thanks.

bianpengyuan (commented):
Can you update istio/istio example and test yaml file with this? Specifically https://github.com/istio/istio/blob/master/samples/bookinfo/policy/productpage_envoy_ratelimit.yaml and https://github.com/istio/istio/blob/master/tests/integration/telemetry/policy/testdata/enable_envoy_ratelimit.yaml. This will make sure the configuration is tested.

Issue #10052, opened 2021-07-09
noticed that the current sni-proxy snippet for the egress gateways doesn't work in kubernetes clusters


AndreaM12345:
in testing the sni-proxy snippet mentioned in this doc https://istio.io/latest/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#setup-egress-gateway-with-sni-proxy in my kubernetes cluster noticed that some things were not working

In sending the traffic noticed that the proxy_pass would add a `\` which would not work so removed that and also noticed that the kube dns works better than the google dns for the resolver.

AndreaM12345:
would like to take this issue

PR #10053, opened 2021-07-09
modify sni-proxy.conf snippet for egress to work in kube

AndreaM12345:
This is to fix https://github.com/istio/istio.io/issues/10052

[ x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ x] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/snips.sh
+++ b/content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/snips.sh
@@ -192,7 +192,7 @@ stream {
   server {
     resolver 8.8.8.8 ipv6=off; #if using in kubernetes it may help to use the kube dns resolver (kube-dns.kube-system.svc.cluster.local)
     listen       127.0.0.1:18443;
-    proxy_pass   $ssl_preread_server_name:443;
+    proxy_pass   \$ssl_preread_server_name:443; #should remove the `\` when running as its an artifact of bash scripting
     ssl_preread  on;
   }
 }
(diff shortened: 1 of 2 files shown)


ericvn:
Will probably need the same change in the .md file.

content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/snips.sh
```diff
@@ -190,9 +190,9 @@ stream {
 
   # tcp forward proxy by SNI
   server {
-    resolver 8.8.8.8 ipv6=off;
+    resolver 8.8.8.8 ipv6=off; #if using in kubernetes it may help to use the kube dns resolver (kube-dns.kube-system.svc.cluster.local)
     listen       127.0.0.1:18443;
-    proxy_pass   \$ssl_preread_server_name:443;
```
frankbu, line 195: I think the `\` is needed to escape the `$` char because of the way the `cat` command is used to put the contents into the file `sni-proxy.conf`. That bash command will remove the `\`.

Btw, you should never modify `snips.sh` files directly. Instead, go to the corresponding .md file (`wildcard-egress-hosts/index.md` in this case), make the change there, and then run `make gen` to regen the snips file. The way that snips are generated from md files is described here: https://github.com/istio/istio.io/blob/master/tests/README.md

AndreaM12345:
@googlebot I signed it!

craigbox (commented):
@AndreaM12345 want to come back to this PR with the benefit of experience? :)

frankbu:
This documentation has been removed long ago, so we can close this.

PR #10130, opened 2021-07-28
fix apiVersion for gw install

linsun:
getting an error for current doc if I use this yaml with server side istio-operator.   

```
error: unable to recognize "iop.yaml": no matches for kind "IstioOperator" in version "operator.istio.io/v1alpha1"
```

use `apiVersion: install.istio.io/v1alpha1` would work for both `istioctl install -f` or `kubectl apply -f`

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/additional-setup/gateway/index.md
+++ b/content/en/docs/setup/additional-setup/gateway/index.md
@@ -52,7 +52,7 @@ As a security best practice, it is recommended to deploy the gateway in a differ
 First, setup an `IstioOperator` configuration file, called `ingress.yaml` here:
 
 {{< text yaml >}}
-apiVersion: operator.istio.io/v1alpha1
+apiVersion: install.istio.io/v1alpha1
 kind: IstioOperator
 metadata:
   name: ingress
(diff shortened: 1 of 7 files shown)


craigbox:
@istio/wg-environments-maintainers: https://github.com/istio/istio/pull/21562 suggests it changed name around 1.5; are we OK to `s/operator.istio.io/install.istio.io/` throughout the docs?

(Lin: if so, there are a few other locations that we should do too)

rvennam:
I recently noticed this in https://istio.io/latest/docs/setup/install/external-controlplane/#enable-gateways

morvencao (approved):

howardjohn (approved):

kebe7jun (approved):

linsun:
@craigbox I brought this up in env WG mtg today, yes, it should be `install.istio.io`.  The istioctl cmd worked because it didn't check for the apiVersion correctness. cc @istio/wg-environments-maintainers 

I've updated all other places I could find.  PTAL.

craigbox (approved):
Great work Lin, thank you!

PR #10139, opened 2021-08-01
Documentation: change Secret name in Sidecar Injection Problems section

weibeld:
Change Secret name from `istiod-service-account-token` to `istio-ca-secret`.

I guess the latter is the correct Secret to use, as it's the only Secret containing a `root-cert.pem` field. And the former seems to be a ServiceAccount token Secret as created by the [Token Controller](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#token-controller) for every ServiceAccount.

However, I didn't look into the code, so correct me if I'm wrong.

Link to the section in the docs: https://istio.io/latest/docs/ops/common-problems/injection/#x509-certificate-related-errors

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/common-problems/injection/index.md
+++ b/content/en/docs/ops/common-problems/injection/index.md
@@ -148,7 +148,7 @@ Verify the `caBundle` in the `mutatingwebhookconfiguration` matches the
 {{< text bash >}}
 $ kubectl get mutatingwebhookconfiguration istio-sidecar-injector -o yaml -o jsonpath='{.webhooks[0].clientConfig.caBundle}' | md5sum
 4b95d2ba22ce8971c7c92084da31faf0  -
-$ kubectl -n istio-system get secret istiod-service-account-token -o jsonpath='{.data.root-cert\.pem}' | md5sum
+$ kubectl -n istio-system get secret istio-ca-secret -o jsonpath='{.data.root-cert\.pem}' | md5sum
 4b95d2ba22ce8971c7c92084da31faf0  -
 {{< /text >}}
 
--- a/content/en/docs/ops/common-problems/injection/index.md
+++ b/content/en/docs/ops/common-problems/injection/index.md
@@ -146,9 +146,9 @@ Verify the `caBundle` in the `mutatingwebhookconfiguration` matches the
    root certificate mounted in the `istiod` pod.
 
 {{< text bash >}}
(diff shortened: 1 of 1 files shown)


lei-tang:
- When following the guide content/en/docs/ops/common-problems/injection/index.md, what is the error?
- After changing Secret name from istiod-service-account-token to istio-ca-secret, is the error resolved?

content/en/docs/ops/common-problems/injection/index.md
```diff
@@ -148,7 +148,7 @@ Verify the `caBundle` in the `mutatingwebhookconfiguration` matches the
 {{< text bash >}}
 $ kubectl get mutatingwebhookconfiguration istio-sidecar-injector -o yaml -o jsonpath='{.webhooks[0].clientConfig.caBundle}' | md5sum
 4b95d2ba22ce8971c7c92084da31faf0  -
-$ kubectl -n istio-system get secret istiod-service-account-token -o jsonpath='{.data.root-cert\.pem}' | md5sum
+$ kubectl -n istio-system get secret istio-ca-secret -o jsonpath='{.data.root-cert\.pem}' | md5sum
```
howardjohn, line 151: This is empty in my out-of-box `istioctl install`

howardjohn (changes requested):
I agree the current doc is clearly wrong, but not sure this is right either.

content/en/docs/ops/common-problems/injection/index.md
weibeld: Yes, `root-cert.pem` is empty. The correct field seems to be `ca-cert.pem`. I fixed it in a new commit.

weibeld:
> * When following the guide content/en/docs/ops/common-problems/injection/index.md, what is the error?

The second command (`kubectl -n istio-system get secret istiod-service-account-token ...`) yields an empty output because the field within the Secret doesn't exist.

> * After changing Secret name from istiod-service-account-token to istio-ca-secret, is the error resolved?

Yes, including changing `jsonpath='{.data.root-cert\.pem}'` to `jsonpath='{.data.ca-cert\.pem}'`. The MD5 sums of the two commands are now identical.

Tested for Istio 1.10.3.

howardjohn (approved):

ericvn (approved):

PR #10143, opened 2021-08-02
Remove "file mount" egress gw doc

howardjohn:
This is actively leading users down a bad practice. We previously did
the same for Ingress - the results were we got a lot less bugs about
file mount being very hard to use.

As is, users are directed here as the default - only if they happen to
know what "SDS" is (an implementation detail) will they realize the
other doc is better.

This removes the "file mount" vs "SDS" distinction - instead there is just a single doc (which uses SDS, but never mentions it)

Testing: this doc is fully doc-tested, so we should be covered here.


zioproto:
@howardjohn how can we get some traction on this PR ? thanks

Issue #10151, opened 2021-08-05
Known Kubernetes versions won't work


linsun:
Given we move to v1 API for CRD such as mutating webhook configuration, I think this means k8s 1.15 won't work with Istio 1.11.  If this is true, I prefer us add a statement around "Istio 1.11 won't work with k8s 1.15 or older" in release notes and platform support pages.

@howardjohn @ericvn @brian-avery thoughts?

howardjohn:
We broke 1.15 support in Istio 1.10. I think it's already in the notes and platform setup

ericvn:
I haven't looked at `preliminary.istio.io`, but at least `istio.io` doesn't list 1.15 as a supported version starting with 1.7, and is no longer listed as untested starting with 1.10: https://istio.io/latest/docs/releases/supported-releases/, so I'm not thinking we need to add one.

I recall in TOC mention of adding a banner that 1.9.x won't work with k8s 1.22 so maybe there is some combined statement?

howardjohn:
The reason it's important to announce the 1.22 one is (1) it's the first
time a new Kubernetes version has broken an old Istio version and (2) users
Kubernetes clusters may automatically update to 1.22 and break Istio.

Old k8s versions not working I'd much less risky and better understood by
users

brian-avery:
I agree with John  and Eric. Iirc, we had breaking changes with 1.16 as well, so there's already issues with older versions of Kubernetes and we state that we don't support them.

linsun:
There is a difference between unsupported versions vs versions it won't work.  

If you look at our platform setup page(https://istio.io/latest/docs/setup/platform-setup/), we don't tell users k8s 1.15 won't work for Istio.  We did say k8s 1.17 or older is not supported.

If we know some versions won't work for sure, we should tell it.

brian-avery:
Are you thinking we'll always keep the list of definitely broken, possibly broken, and supported updated based on user issues or our own testing? 

We can call out 1.15 and 1.16, but I'm not sure what that implies for the support of 1.17 or how deep we'd want to go on specific features that are broken. 

My feeling is it should be assumed that anything below the supported versions is "your mileage may vary".

linsun:
[...]

PR #10153, opened 2021-08-05
Test banner for warning of using istio < 1.10 with k8s 1.22

ericvn:
A PR to get capture discussion on publicizing the fact that Istio 1.9 and earlier do not work with Kubernetes 1.22 which was just released started in https://github.com/istio/istio.io/issues/10151.

Fixes https://github.com/istio/istio.io/issues/10154


ericvn:
I will take any comments/changes to the message. Toning it done some:
```
Istio versions prior to 1.10 do not work with Kubernetes 1.22 or later. Verify your Istio version is supported when upgrading your Kubernetes version.
```

content/en/events/banners/kubernetes-1.22-with-older-releases.md
```diff
@@ -0,0 +1,11 @@
...
+title: Kubernetes 1.22 Release statement
+period_start: 2021-08-04
+period_duration: 180
+max_impressions: 20
+timeout: 120
+link: /docs/releases/supported-releases#support-status-of-istio-releases
+
+---
+
+Warning! Kubernetes 1.22 is not supported by Istio release 1.9.x. Do not upgrade your Kubernetes version if you are using Istio versions prior to 1.10.
```
brian-avery, line 11: Is this better? 

```suggestion
Istio release 1.9.x is not compatible with Kubernetes 1.22. If you upgrade your Kubernetes cluster, please consider upgrading to Istio 1.10 or newer. 
```

ericvn: I'm not sure `consider` is accurate, at least if they are upgrading to 1.22.

ericvn:
I suspect, depending on when this merges, it makes sense to CP to `istio.io`

craigbox:
How about something.nice and positive like "Kubernetes 1.22 is supported by
Istio 1.10 and above. Removal of deprecated APIs (link) will  break older
versions of Istio. Check here for the supported version table"

brian-avery:
I like it @craigbox

brian-avery:
Is it worth it to clarify that by older versions, we mean 1.9 and below? Call them out explicitly?

craigbox:
Something went wrong here...

image

ericvn:
I wonder if it's a length limitation. I've shortened it and it seems to work.
Screen Shot 2021-08-06 at 8 23 13 AM

Now to see if the wording should be tweaked some more.

craigbox:
Still failing for me, though my screen is _super wide_: but funnily enough, it fixes after a few refreshes.  Looking to see if I can find out why...

image

ericvn:
I'm seeing the same when using the `Deploy` here, and seemed to correct itself on the 3rd refresh. Worked fine locally with `make serve`. Thanks for looking into it.

content/en/events/banners/kubernetes-1.22-with-older-releases.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Kubernetes 1.22 Release statement
+period_start: 2021-08-04
+period_duration: 180
+max_impressions: 8
+timeout: 120
+link: /docs/releases/supported-releases#support-status-of-istio-releases
+---
+[Removal of deprecated APIs](https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/) in
+Kubernetes 1.22 will break Istio 1.9 and below. Click here for the supported version table
```
craigbox, line 10:
```suggestion
Kubernetes 1.22 will only work with Istio 1.10 and above. Check here for the supported version table.
```

ericvn:
At this point, the `deploy` seems to be OK (although may have to use a private window to get the banner to show). The link still takes you to the istio.io `support releases` page, but there is a PR to update that and it should be active after the release next week.

craigbox (approved):

PR #10163, opened 2021-08-09
Update direct upgrade warning

therealmitchconnors:
As of May, skipping a single minor version is supported.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/upgrade/_index.md
+++ b/content/en/docs/setup/upgrade/_index.md
@@ -7,5 +7,5 @@ test: n/a
 ---
 
 {{< warning >}}
-Upgrading across more than one minor version (e.g., `1.6.x` to `1.8.x`) in one step is not officially tested or recommended.
+Upgrading across more than two minor versions (e.g., `1.6.x` to `1.9.x`) in one step is not officially tested or recommended.
 {{< /warning >}}


craigbox (commented):
Is this only true for canary upgrades?  Do we want to call that out if so?

therealmitchconnors:
I will leave it to @Monkeyanator to clarify what variety of upgrades we test.

therealmitchconnors:
I just got clarification from @jacob-delgado that Helm upgrades are also tested.  It's not clear whether non-revision based upgrades are included.

Xunzhuo (commented):
LGTM

craigbox:
> It's not clear whether non-revision based upgrades are included.

Assuming they are still supported, who can clear that up?  @howardjohn may know

howardjohn:
You can do single version in place upgrade, 2 version upgrade with canary

craigbox:
ok, on looking again, I think this is fine to go in because there is _a_ method to do two versions, and we probably do recommend the canary approach :)

craigbox (approved):

PR #10164, opened 2021-08-09
refactor istio.io security page Authz part

xulingqing:

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/concepts/security/index.md
```diff
@@ -569,366 +556,25 @@ An authorization policy includes a selector, an action, and a list of rules:
...
-         methods: ["GET", "HEAD"]
-{{< /text >}}
+Value matching in authorization policies supports Exact match, Prefix match with an ending `"*"`,
+Suffix match with a starting `"*"`, Presence match with `fieldname: ["*"]` format to specify
+that a field must be present.
+At the same time, user can use Exclusion matching to match negative conditions like `notValues` in the `when` field, `notIpBlocks`
+in the `source` field, `notPorts` in the `to` field. Check [Security Tasks]() for further details.
 
-#### Value matching
+In the [Security Policy Examples](docs/ops/configuration/security/security-policy-examples/), we introduce more policy use cases like:
```
yangminzhu, line 572: I think you should make sure the new content is ready when deleting it from the concept page. You could do it in a single PR that adds the new content in example/task/reference page while deleting it from concept page.

Otherwise there is chance that some contents are removed in this PR but forget to be added in other places, and more importantly, separating the change will make it harder to review and compare the newly added content to the old one, and it puts the preminary.istio.io in an incomplete status until all changes are merged.

yangminzhu (commented):
I

content/en/docs/concepts/security/index.md
xulingqing: Thanks for your suggestion. Yes, the changes to the authz part will be in the same PR. But let's first have consensus of how will the concept page look like. The concept page changes will be finished in the 1st commit, and I will commit the tasks/reference changes in this PR as a different commit.

xulingqing: Once we reach agreement on the concept page, @justinpettit will work with me on finding a section that fits for the delete content. In this way, we can greatly reduce moving the content around. Thanks~

craigbox (commented):
@xulingqing and @justinpettit, any interest in finishing this off?

PR #10181, opened 2021-08-13
Ensure extra padding when banner is present

rytswd:
Ensure CSS padding setup is correctly handled when banner is present for pages such as Docs, News and Get Involved.
Some other pages such as FAQ, Docs, etc. seem to have no issue as there is no conflicting padding setup.

Before this change:

Blog
image

News
image

Get Involved
image

--

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/src/sass/misc/_banners.scss
+++ b/src/sass/misc/_banners.scss
@@ -59,6 +59,6 @@
 }
 @media screen AND (min-width: $bp-md) {
     .banner-padding {
-        padding-top: 5rem;
+        padding-top: 5rem !important;
     }
 }


rytswd:
This seems to work for Blog and News
image

image

But for Get Involved, because the original margin / padding setup is slightly different from other pages, having double banners would still cause strange rendering. It may be worth tackling, but I feel like this style consolidation should be done for all the pages, and thus could perhaps be tackled separately.

image

craigbox (approved):

Issue #9858, opened 2021-05-31
Display issue - overlapping tables (see screenshot)


su225:
![image](https://user-images.githubusercontent.com/41095737/120193836-7e6e9480-c23a-11eb-9ef3-86c1400bbbd7.png)

Page - https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec
Browser - Google Chrome (Version 91.0.4472.77 (Official Build) (64-bit))

The overlapping is weird and should be fixed.

Arhell:
done https://github.com/istio/istio.io/pull/10238

PR #10238, opened 2021-08-25
fix overlapping table

Arhell:
fix issue https://github.com/istio/istio.io/issues/9858

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/src/sass/base/_base.scss
+++ b/src/sass/base/_base.scss
@@ -96,6 +96,15 @@ table {
         display: block;
         overflow-x: auto;
     }
+
+    p {
+        word-break: break-word;
+
+        a,
+        code {
+            word-break: break-word;
+        }
+    }
 }
 
 table p:first-of-type {
--- a/src/sass/misc/_article.scss
+++ b/src/sass/misc/_article.scss
@@ -29,3 +29,9 @@
         }
     }
 }
+
+article {
+    h2 {
+        word-break: break-word;
+    }
+}


shamsher31 (commented):
Looks good. But need to test if it affects any other part of the website.
https://deploy-preview-10238--preliminary-istio.netlify.app/latest/docs/reference/config/istio.operator.v1alpha1/#KubernetesResourcesSpec

ericvn:
https://deploy-preview-10238--preliminary-istio.netlify.app/latest/docs/reference/config/istio.mesh.v1alpha1/ still has an overlap.

rvennam:
@Arhell Can you show us an example of where words are not always transferred correctly after your PR? GlobalMeshOptions and IstioOperator referece pages seem OK with your PR.

Arhell:
> @Arhell Can you show us an example of where words are not always transferred correctly after your PR? GlobalMeshOptions and IstioOperator referece pages seem OK with your PR.

https://deploy-preview-10238--preliminary-istio.netlify.app/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ExtensionProvider-EnvoyExternalAuthorizationGrpcProvider

craigbox:
If the problem is only that some words are broken (which are too long for
the line anyway), I think that's a win.  I'm not sure there's a fix we
could apply to break words with no obvious break - it would have to include
hyphenation, which would introduce technical incorrectness?

>

Arhell:
> If the problem is only that some words are broken (which are too long for the line anyway), I think that's a win. I'm not sure there's a fix we could apply to break words with no obvious break - it would have to include hyphenation, which would introduce technical incorrectness?
> […](#)

Technically, unlikely, I isolated the styles to the maximum so that it would not affect other content (word wrap). Just readability can be a little difficult

rvennam:
Agree that the new change is better than before. I say we merge this and then look into making those subtitiles smaller - h3 instead of h2 to give it more room

rvennam (approved):

Arhell:
cc @craigbox

craigbox (approved):

PR #10332, opened 2021-09-14
add task for JWT claim based routing

yangminzhu:
The doc is ready for review, the test will fail until the PR https://github.com/istio/istio/pull/34974 gets merged.

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [x] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
@@ -224,6 +224,126 @@ Let's see how you can configure a `Gateway` on port 80 for HTTP traffic.
     ...
     {{< /text >}}
 
+## Configuring ingress routing based on JWT claims
+
+The Istio ingress gateway also supports routing based on authenticated JWT, which is useful for routing based on end user
+identity and more secure compared using the unauthenticated HTTP attributes (e.g. path or header).
+
+This section describes how to configure the ingress gateway to route based on JWT claims.
+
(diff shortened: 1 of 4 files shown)


yangminzhu:
> Useful content. Putting it between "Configuring ingress using an Istio gateway" and "Accessing ingress services using a browser" sections breaks the flow of the page.
> 
> I think this section should be moved to a different page: https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/

Thanks for the review, I'm fine to move this to wherever the proper place but I think the https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/ is a little bit strange as this task does not use any of the authorization policies.

My preference is still to put this under networking tasks as this is more about routing but security is fine to me, we can add a new one if that is better. @howardjohn @costinm @liminw please also let me know what you think, thanks.

howardjohn:
I agree it should be under security section. Or at the very least, not on this page - if we really want it on traffic management we can put it on its own page (preferably not under ingress/). This doc is like the #3 viewed istio.io page, we should keep it as small as possible.

rvennam:
@yangminzhu Could you move this to security or a new page under traffic mgmt?

yangminzhu:
> @yangminzhu Could you move this to security or a new page under traffic mgmt?

yeah, will move to security once the implementation PR is merged

yangminzhu:
@myidpt @liminw moved the task to security page, ptal, thanks.

Issue #10321, opened 2021-09-10
The doc for "Istio Operator Install" seems to contain incorrect info


syedrakib:
This issue is related to https://github.com/istio/istio.io/issues/9892#issuecomment-917119319

## Steps to reproduce the error

I deleted all istio namespaces from my GKE cluster

```bash
$ kubectl delete namespace istio-operator istio-system
```

Installed **istioctl version 1.10.2** in MacOS following https://istio.io/latest/docs/ops/diagnostic-tools/istioctl/?_ga=2.53425067.2032442266.1631216965-1634768249.1631055742#install-hahahugoshortcode-s2-hbhb

Then installed istio-operator

```bash
$ istioctl operator dump > operator_dump.yaml
$ kubectl apply -f operator_dump.yaml
```

Confirmed that the istio-operator is up & running

```bash
$ kubectl rollout status deployment/istio-operator -n=istio-operator
deployment "istio-operator" successfully rolled out
```

## The Error

The next step involves applying the IstioOperator CRD following https://istio.io/latest/docs/setup/install/operator/?_ga=2.63248428.2032442266.1631216965-1634768249.1631055742#install

```bash
kubectl apply -f - <<EOF
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
  name: example-istiocontrolplane
spec:
  profile: demo
EOF
```

At this point i expect it to just work. But i got the following error:

```
Error from server (NotFound): error when creating "STDIN": namespaces "istio-system" not found
```

## But documentation says otherwise

The docs at https://istio.io/latest/docs/setup/install/operator/?_ga=2.233666461.2032442266.1631216965-1634768249.1631055742#install says 

> Prior to Istio 1.10.0, the namespace istio-system needs to be created before installing the operator. As of Istio 1.10.0, this is no longer required.

---

Looks like the istio operator needs a fix or the docs needs to be revised.

craigbox:
@jacob-delgado @shamsher31 could you take a look?  I think you updated the comment/feature and it may not be correct.

shamsher31:
Based on this change https://github.com/istio/istio/pull/31237, The namespace `istio-system` gets automatically created if you use the following command.
```
$ istioctl operator init
```
and then apply IstioOperator CRD

```yaml
kubectl apply -f - <<EOF
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
[...]

PR #10339, opened 2021-09-15
Clarify warn for istio-system namespace creation

shamsher31:
To fix https://github.com/istio/istio.io/issues/10321


craigbox:
PTAL at a more thorough rewrite in #10343.

PR #10349, opened 2021-09-16
Fix grafana dashboard URLS after grafana update

ericvn:

https://github.com/istio/istio/pull/35196 updated the Grafana version installed by the samples. This caused the `using-istio-dashboard` test to fail as it's URL has changed. I updated the URLS in the index.md to match what I am seeing as well as in the test.sh.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/observability/metrics/using-istio-dashboard/index.md
+++ b/content/en/docs/tasks/observability/metrics/using-istio-dashboard/index.md
@@ -54,7 +54,7 @@ the example application throughout this task.
     $ istioctl dashboard grafana
     {{< /text >}}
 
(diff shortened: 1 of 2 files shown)


howardjohn (approved):
We should open an issue to see if we can get the nice URLs back so we don't forget.

thanks!

rvennam (approved):

ericvn:
> We should open an issue to see if we can get the nice URLs back so we don't forget.
> 
> thanks!

https://github.com/istio/istio/issues/35226 created

PR #10380, opened 2021-09-23
Include sidecar.env for VM configuration

xulingqing:

BUG: https://github.com/istio/istio/issues/34058
PR: https://github.com/istio/istio/pull/35048 
If the revision is set, we try to use the same discovery address that we setup in /etc/hosts. It involves using *sidecar.env* to pass a revisioned istiod address.

- [X] Docs
- [X] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -285,6 +285,7 @@ If third party tokens are not enabled, you should add the option `--set values.g
 
 Next, use the `istioctl x workload entry` command to generate:
 
+* `sidecar.env`: Contains metadata specific to each VM like istiod address to fetch CA.
 * `cluster.env`: Contains metadata that identifies what namespace, service account, network CIDR and (optionally) what inbound ports to capture.
 * `istio-token`: A Kubernetes token used to get certs from the CA.
 * `mesh.yaml`: Provides `ProxyConfig` to configure `discoveryAddress`, health-checking probes, and some authentication options.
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/install/virtual-machine/index.md
```diff
@@ -285,6 +285,7 @@ If third party tokens are not enabled, you should add the option `--set values.g
 
 Next, use the `istioctl x workload entry` command to generate:
 
+* `sidecar.env`: Contains metadata specific to each VM like istiod address to fetch CA.
```
howardjohn, line 288: This is confusing since its not actually how its used. I think reality is more like "cluster.env: contains metadata developer A added. sidecar.env: contains metadata developer B added" :slightly_smiling_face: .

In the current VM installation method, it doesn't seem like there is a whole lot of motivation to split these files. It may be better suited to just consolidate all of them to cluster.env

xulingqing: @stevenctl what's your idea? do we need to add the sidecar.env change to cluster.env?

stevenctl: Agree with John. We can still support sidecar.env, but no real reason to generate both from our tool. 

https://github.com/istio/istio/pull/35340

rvennam:
If sidecar.env is being merged into cluster.env, can this PR be closed?

craigbox (commented):
This seems to be out of date given #35340.  If you still have the problem, please feel free to revive the PR.

PR #10404, opened 2021-10-05
add cookie consent banner

### LINKED ISSUE #10005: Review website cookies/analytics/opt-out?
A few people were having a conversation about the "cookie banner" on knative.dev, and we noticed that https://istio.io/ didn't have such a banner. Digging further, it looks like I'm served Google Analytics javascript and reports are collected even in an incognito window with `DNT: 1` cookie.

I'm in the US, so maybe this is okay here, but I suspect that our European friends would be surprised by this.

davidhauck:
Closes #10005 

- [ X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


Arhell (commented):
lgtm

layouts/_default/baseof.html
```diff
@@ -192,5 +193,55 @@
                 </a>
             </div>
         {{ end }}
+
+        <!-- Analytics Consent -->
+        <!-- Uses https://github.com/osano/cookieconsent (MIT License) -->
```
rvennam, line 198: @craigbox do we need blessing from steering?

PR #10420, opened 2021-10-12
Add usage warning for in-cluster operator

Monkeyanator:
Per Environments and TOC discussion that we should discourage use but not deprecate or introduce a new maintenance mode. Moves in-cluster operator to the bottom of suggested install methods, adds a warning on the `install method selection` page, and adds a warning to the top of the in-cluster operator install docs.


marshallford:
> Per Environments and TOC discussion

Anything public that can be linked to regarding this decision? The operator install method appears to be the sane way to install Istio via GitOps.

PR #10453, opened 2021-11-01
Fix subzone link not found

hanxiaop:
In link https://istio.io/latest/docs/tasks/traffic-management/locality-load-balancing/, the `topology.istio.io/subzone` link does not exist.

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/locality-load-balancing/_index.md
+++ b/content/en/docs/tasks/traffic-management/locality-load-balancing/_index.md
@@ -32,7 +32,7 @@ triplet defines a locality:
 - **Sub-zone**: Allows administrators to further subdivide zones for more
   fine-grained control, such as "same rack". The sub-zone concept doesn't exist
   in Kubernetes. As a result, Istio introduced the custom node label
-  [`topology.istio.io/subzone`](https://github.com/istio/api/blob/master/label/label.go#L42)
+  [`topology.istio.io/subzone`](https://github.com/istio/api/blob/master/label/labels.yaml#L76)
   to define a sub-zone.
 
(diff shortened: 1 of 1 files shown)


Arhell (commented):
lgtm
cc @craigbox

craigbox (commented):
@ericvn thoughts on if this should be tied to `master`?  Either it goes stale if the file changes, or we use a `release-1.12` link and it goes stale as versions increase.

monsteredc:
@craigbox I had the same doubt at the beginning. Before using the `master` version, I was tried to use the permalink. I agree that both of the implementations have their own flaws, and we may discuss which one is more preferred.

ericvn:
Can we just use something like `github_blob` which takes into account the currently built branch? Since this is in `api`, we'd have create something similar.  It doesn't help if the line number changes in the file across releases.

craigbox:
@hanxiaop would love to see this moved forward, so have marked it to be open.

hanxiaop:
@ericvn How about other links using `https://github.com/istio/api/blob`? After I added the lint check for `github_api_blob`, I realized there are lots of places using the `api` link right now, and some of the links will be invalid as well if I replace them with the `github_api_blob` form. Should they keep the original permalink form or should we do something?

ericvn:
Replaced by https://github.com/istio/istio.io/pull/13242.

hanxiaop:
> Replaced by #13242.

@ericvn nice work, thanks!

Issue #10492, opened 2021-11-10
Add more detailed instructions for exposing an external istiod to remote clusters


frankbu:
Step 3 in the [external control plane setup instructions](https://istio.io/latest/docs/setup/install/external-controlplane/#set-up-a-gateway-in-the-external-cluster) should provide more details on how to expose the ingress gateway, for example, provide cut-and-paste commands for doing it using `kind` clusters.

PR #10510, opened 2021-11-17
Fixed the issue #10492

litong01:
This PR will fix the issue described in #10492

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/external-controlplane/index.md
+++ b/content/en/docs/setup/install/external-controlplane/index.md
@@ -57,6 +57,41 @@ need to modify the installation procedure to enable access. For example, the
 [multicluster configuration](#adding-clusters) could also be used to enable access
 to the API server.
 
+### Istio CA certificates
+
+Since Istio remote cluster will communicate with Istio external control plane cluster
+via TLS connections invoked by Istio mutating web hooks running in remote cluster,
+it is essential that the Istio mutating webhook in remote cluster trusts the
+Istio control plane injection service whose connection shall be encrypted by either
+the ingress gateway or the external control plane itself. This is especially important
+when your Istio certificates are self signed. Thus the external control plane
+root certificate (not to be confused with its private key) shall be available when set
+up your Istio component in remote cluster such as the mutating web hook.
+
+There are number of ways to obtain such certificate. If you are setting up a production environment,
+most likely your certificates come from Trusted Certificate Authority such as `VeriSign` or
+`SecureTrust`, you should already have the certificate in your possession, and you do not have to configure
+your Istio web hooks to trust additional certificates.
+
+{{< tip >}}
+When you use self signed certificates for your environment, the certificate used by external control plane
+must be trusted by the Istio web hooks running in your remote cluster, extra configuration step is needed.
+Please see the later section on using self signed certificates.
+{{< /tip >}}
+
+Please refer to [Certificate Management](https://kubernetes.io/docs/tasks/security/cert-management/) for
+more information about Istio certificates management and [Plug in CA Certificates](https://kubernetes.io/docs/tasks/security/cert-management/plugin-ca-cert/) how to generate self signed certificates and apply the generated
+certificates for your Istio components. Regardless where the certificates may come from, the root certificate
+used by external control plane needs to be available during the setup process, this document assumes that
+you have access to this certificate and its base 64 encoded content is saved in an environment variable named CACERT
+
+For example: If your root cert is named `myrootcert.crt` in your current directory, you may run the following
+command to get the certificate base 64 encoded into the CACERT environment variable.
+
+{{< text syntax=bash snip_id=none >}}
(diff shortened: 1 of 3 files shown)


Arhell (commented):
cc @craigbox

content/en/docs/setup/install/external-controlplane/index.md
```diff
@@ -229,9 +304,6 @@ and installing the sidecar injector webhook configuration on the remote cluster
         rootNamespace: external-istiod
         defaultConfig:
           discoveryAddress: $EXTERNAL_ISTIOD_ADDR:15012
-          proxyMetadata:
```
frankbu, line 232: Removing these breaks the "production" instructions.

content/en/docs/setup/install/external-controlplane/index.md
```diff
@@ -311,19 +383,17 @@ and installing the sidecar injector webhook configuration on the remote cluster
             protocol: https
             name: https-XDS
           tls:
-            mode: SIMPLE
-            credentialName: $SSL_SECRET_NAME
+            mode: PASSTHROUGH
```
frankbu, line 386: We don't want to change the config to use passthrough in the official doc. Just mention (in a tip) that you can change it. Btw, that's exactly what the test program does with a couple of sed commands. I'm actually surprised that the test isn't broken with your changed snip.

content/en/docs/setup/install/external-controlplane/index.md
```diff
@@ -57,6 +57,38 @@ need to modify the installation procedure to enable access. For example, the
...
+Istio control plane injection service whose connection shall be encrypted by either
+the ingress gateway or the external control plane itself. This is especially important
+when your Istio certificates are self signed. Thus the external control plane
+root certificate (not to be confused with its private key) shall be available when set
+up your Istio component in remote cluster such as the mutating web hook.
+
+There are number of ways to obtain such certificate. If you are setting up a production environment,
+most likely your certificates come from Trusted Certificate Authority such as `VeriSign` or
+`SecureTrust`, you should already have the certificate in your possession, and you do not have to configure
+your Istio web hooks to trust additional certificates. However if you use self signed certificates either
```
frankbu, line 74: I don't think we want the self-signed passthrough approach to be a first class part of these instructions. I think we should just mention in a `{{< tip >}}` that you can do it and also in other tips (at the appropriate places in the instructions) show what to change in the config.

content/en/docs/setup/install/external-controlplane/index.md
```diff
@@ -173,6 +210,44 @@ and installing the sidecar injector webhook configuration on the remote cluster
...
+              name: istio-sidecar-injector-external-istiod
+              patches:
+              - path: webhooks[0].clientConfig.caBundle
+                value: "${CACERT}"
+              - path: webhooks[1].clientConfig.caBundle
+                value: "${CACERT}"
+              - path: webhooks[2].clientConfig.caBundle
+                value: "${CACERT}"
+              - path: webhooks[3].clientConfig.caBundle
+                value: "${CACERT}"
```
frankbu, line 247: Instead of showing a copy of the whole config here, add a tip that just shows just the overlay part and says add this to the config.

content/en/docs/setup/install/external-controlplane/index.md
litong01: @frankbu this should not break production setup since cacertificates file is the default. I can add it back if you insist.

content/en/docs/setup/install/external-controlplane/index.md
litong01: ok. will mention it in a tip.

content/en/docs/setup/install/external-controlplane/index.md
litong01: ok. tip as well

content/en/docs/setup/install/external-controlplane/index.md
litong01: I was thinking about that but sometimes I found it is a bit difficult to describe it instead of a good example. will try to describe it though.

content/en/docs/setup/install/external-controlplane/index.md
frankbu: Oh, in that case it sounds like unnecessary noise in the instructions with or without your other changes. I don't know who added it in the first place, but if it works without it we should simplify the config and remove it.

litong01: right. this should not be needed.

content/en/docs/setup/install/external-controlplane/index.md
litong01: PASSTHROUGH should work in more cases than the TLS simple mode.

litong01:
take a different direction at present.

PR #10521, opened 2021-11-19
Add 1.10 pre-EOL announcement

Monkeyanator:
EOL warning for 1.10

### DIFF
--- a/content/en/news/support/announcing-1.10-eol/index.md
+++ b/content/en/news/support/announcing-1.10-eol/index.md
@@ -1,0 +1,14 @@
+---
+title: Support for Istio 1.10 ends on December 30th, 2021
+subtitle: Support Announcement
+description: Upcoming Istio 1.10 end of life announcement.
+publishdate: 2021-11-19
+aliases:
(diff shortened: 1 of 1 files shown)


ericvn:
While I believe that the n+2 + 6 weeks is correct for the 1.10 release, it was mentioned to present to the TOC about carrying this forward for future releases. Since we mention a formula here and there was a discussion of a blog for support going forward, I think we can wait until the TOC meeting on Nov 29 before merging this. That date still leaves a 30 day notice for EOL notice.

Monkeyanator:
> While I believe that the n+2 + 6 weeks is correct for the 1.10 release, it was mentioned to present to the TOC about carrying this forward for future releases. Since we mention a formula here and there was a discussion of a blog for support going forward, I think we can wait until the TOC meeting on Nov 29 before merging this. That date still leaves a 30 day notice for EOL notice.

Sounds good.

Monkeyanator:
From 11/29 TOC we will be carrying the policy forward.

frankbu (approved):
The publish date is a little old here, but I better late than never :)

PR #10545, opened 2021-11-29
fix weak algorithm

### LINKED ISSUE #10544: CA signature digest algorithm too weak
https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/
```bash
curl -HHost:httpbin.example.com --resolve "httpbin.example.com:443:172.18.0.6" --cacert example.com.crt "https://httpbin.example.com"
curl: (60) SSL certificate problem: CA signature digest algorithm too weak
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

liuxu623:

Fix #10544 

- [x] Docs

### DIFF
--- a/content/en/docs/tasks/observability/gateways/index.md
+++ b/content/en/docs/tasks/observability/gateways/index.md
@@ -59,7 +59,7 @@ This example uses self-signed certificates, which may not be appropriate for pro
     $ mkdir -p ${CERT_DIR}
     $ openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -subj "/O=example Inc./CN=*.${INGRESS_DOMAIN}" -keyout ${CERT_DIR}/ca.key -out ${CERT_DIR}/ca.crt
     $ openssl req -out ${CERT_DIR}/cert.csr -newkey rsa:2048 -nodes -keyout ${CERT_DIR}/tls.key -subj "/CN=*.${INGRESS_DOMAIN}/O=example organization"
-    $ openssl x509 -req -days 365 -CA ${CERT_DIR}/ca.crt -CAkey ${CERT_DIR}/ca.key -set_serial 0 -in ${CERT_DIR}/cert.csr -out ${CERT_DIR}/tls.crt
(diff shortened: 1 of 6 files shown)


Arhell (commented):
/ok-to-test

lei-tang (approved):
Thanks for the fix! 
Nit: can you add a comment on the default digest algorithm used by openssl?

howardjohn (changes requested):
Do we really need to make it more complex? This is clearly just an example of a minimal portable command to generate a cert. No one should be using this in production - it even has `example.com` in the command.

IMO a demo command should set the minimal flags possible. As is, a user will now need to try to grok why 'sha256' is used. Is it because the author likes it? Because its recommended? Because its required? It adds a lot of cognitive overload.

liuxu623:
> Do we really need to make it more complex? This is clearly just an example of a minimal portable command to generate a cert. No one should be using this in production - it even has `example.com` in the command.
> 
> IMO a demo command should set the minimal flags possible. As is, a user will now need to try to grok why 'sha256' is used. Is it because the author likes it? Because its recommended? Because its required? It adds a lot of cognitive overload.

I try to fix #10544, and I figured it out. Because I use macOS's openssl generate certificate, and use Ubuntu's curl test.
```
macOS openssl version
LibreSSL 2.8.3
```
```
Ubuntu openssl version
OpenSSL 1.1.1f  31 Mar 2020
```
The macOS's openssl default digest algorithm is sha1, and if I use Ubuntu's openssl  generate certificate digest algorithm is sha256.

howardjohn (approved):
lgtm, didn't realize the curl issue.

Have to love osx...

liuxu623:
> Thanks for the fix! Nit: can you add a comment on the default digest algorithm used by openssl?

```
macOS openssl version
LibreSSL 2.8.3
```
```
Ubuntu openssl version
OpenSSL 1.1.1f  31 Mar 2020
```
The macOS's openssl default digest algorithm is sha1, and Ubuntu's openssl is sha256.

craigbox (approved):
/cherrypick release-1.12

PR #10572, opened 2021-12-06
Garyan refactoring

aryan16:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [X] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/concepts/security/index.md
+++ b/content/en/docs/concepts/security/index.md
@@ -602,28 +602,6 @@ spec:
 
 This [example](http://istio.io/latest/docs/reference/config/security/authorization-policy/#deny-policy) shows an authorization policy that denies requests if the source is not the `foo` namespace:
 
-<!-- {{< text yaml >}}
-apiVersion: security.istio.io/v1beta1
-kind: AuthorizationPolicy
-metadata:
- name: httpbin-deny
- namespace: foo
-spec:
- selector:
-   matchLabels:
-     app: httpbin
-     version: v1
- action: DENY
- rules:
- - from:
-   - source:
-       notNamespaces: ["foo"]
-{{< /text >}}
-
-The deny policy takes precedence over the allow policy. Requests matching allow
-policies can be denied if they match a deny policy. Istio evaluates deny
-policies first to ensure that an allow policy can't bypass a deny policy. -->
-
 #### Policy Target
 
 You can specify a policy's scope or target with the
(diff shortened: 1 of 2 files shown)


aryan16:
@xulingqing could you please have a look in this pr.

Arhell (commented):
/ok-to-test

craigbox (changes requested):
`content/en/docs/reference/config/security/authorization-policy/index.html` is generated from https://github.com/istio/api/tree/master/security/v1beta1.  Changes to it here will be overwritten.  You'll have to make this change in the protos in the reference docs.

justinpettit:
Can you please not delete content in one commit and then add it back in a different commit?  It makes it hard to track the changes.  If content is moved, it will be easier to follow within a commit.

aryan16:
@justinpettit sorry I couldn't understand your comment properly. We want to have one single commit for the change right ? Right now we are following that only, the other commit which you see had few issues. I forgot to squash both the commits.

ericvn:
> 2. Added a new section named **Examples** in references/configuration/security/Authorization Policy.

As noted above by Craig, this change will need to be done in the istio/api repo (https://github.com/istio/api/blob/master/security/v1beta1/authorization_policy.pb.html) and that will be pulled in (there is a daily job that runs that will pull the change). Note that the file is generated from https://github.com/istio/api/blob/master/security/v1beta1/authorization_policy.proto.

aryan16:
> 
noted.

aryan16:
I have moved some of the authorization examples to ops/configurations/security/security-policy-examples. I will add the tests once I get approval for this change.
cc : @myidpt

justinpettit:
> @justinpettit sorry I couldn't understand your comment properly. We want to have one single commit for the change right ? Right now we are following that only, the other commit which you see had few issues. I forgot to squash both the commits.

I'm not saying that it should only be a single commit.  I'm requesting that if you move text, please remove it from one file and add it to the other file in the same commit.  If you are making changes that are logically separate (e.g., authn changes and authz changes), those should be in separate commits.

aryan16:
> > @justinpettit sorry I couldn't understand your comment properly. We want to have one single commit for the change right ? Right now we are following that only, the other commit which you see had few issues. I forgot to squash both the commits.
> 
> I'm not saying that it should only be a single commit. I'm requesting that if you move text, please remove it from one file and add it to the other file in the same commit. If you are making changes that are logically separate (e.g., authn changes and authz changes), those should be in separate commits.
>
> Understood!! In the new commit there are changes in 2 files. I removed content from one file and added the same content in the other file.

aryan16:
cc : @justinpettit

craigbox (approved):

craigbox (commented):
Hey @aryan16 ; sorry this sat for so long.  I think it might be good to go after a rebase?

PR #10717, opened 2022-01-04
docs: update the networking folder for the zh-docs

saltbo:


zirain:
@saltbo are you still WIP? IMHO, we should not copy en doc to zh folder.

Arhell (commented):
/ok-to-test

saltbo:
> @saltbo are you still WIP? IMHO, we should not copy en doc to zh folder.

I see that the en docs are automatically generated, and the previous zh docs are not translated. 

What is the correct posture? @zirain

zirain:
> > @saltbo are you still WIP? IMHO, we should not copy en doc to zh folder.
> 
> I see that the en docs are automatically generated, and the previous zh docs are not translated.
> 
> What is the correct posture? @zirain

AFAIK, reference docs are generated from istio/api repo, @ericvn can you share more details about it?

ericvn:
@zirain You are correct in that there is a nightly job which updates various reference documents based on thing in other repos. The job runs `make update_ref_docs` which calls the `scripts/grab_reference_docs.sh` script. If you look at the script, it only updates the `en` directory, leaving the other language directories alone. I suspect the reason is that an updated reference docs could point to new istio.io documentation (see Note below), and if the translation hasn't been updated on that istio.io page, the linter could fail with a bad link.

Ignoring that issue, another question would be if these reference files should/could contain translated text? If they could include translations, then we would not want to replace them every time. If they should remain the same across translations, then I think we could upgrade the script to update them all in each language. I'm curious what @craigbox and the translation teams think. Should these files have translations, or should the be consistent across languages?

Note: If we update the script, we probably need to tweak the links that the reference docs point to (see the lint errors):
```
Ensure translated content doesn't include references to English content
```

ericvn:
I am doing a quick test in https://github.com/istio/istio.io/pull/10719 to see what might happen if we tried to updating the /zh/ version of the reference files using the tooling. It looks like we would have to do some more updates for some issues I noted there.

ericvn:
@craigbox Any thoughts on if we should update the reference docs in the `zh` tree? Do we just do it manually as I did in #10719 or do we let the translation team move them as they see fit?

ericvn:
No response on whether we should update the reference docs in the /zh/ tree so closing my test PR for now. If we want to update all the reference docs, we can recreate the PR, as there have been updates since that PR was done.

ericvn:
https://github.com/istio/istio.io/pull/11252 updates the reference docs in the `zh` folder.

PR #10763, opened 2022-01-10
Delete kubernetes-1.22-with-older-releases.md

craigbox:
As Istio 1.10 is out of support, and Kubernetes 1.23 is now current, I suggest it's safe to remove this banner (currently shown to all site visitors).

- [X] Docs


ericvn (approved):
LGTM - We are over 5 months into the 6 month period of the banner and it would be going away soon.

PR #10779, opened 2022-01-12
Vm blog part 2

nmnellis:

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
+---
+title: "Part 2 - Implementing an easier way to add virtual machines to Istio service mesh"
+description: How to statically configure envoy to join a mesh.
```
rvennam, line 3:
```suggestion
description: How to statically configure envoy in a VM to join a mesh.
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+publishdate: 2022-01-17
+attribution: "Nick Nellis (Solo.io)"
+keywords: [virtualmachine,Istio,networking,envoy]
+---
+![Virtual Machine](./istio-setup.png)
+
+Part 1 - [An easier way to add virtual machines to Istio service mesh](/blog/2021/simple-vms/)
+
+A lot of Istio users would like to connect their virtual machines to their service mesh but the current process is very cumbersome and hard to automate. This series explains a simpler way to connect outside applications securely to your applications running in a service mesh. If you would like to know more about why this might work for you, check out part 1 here.
```
rvennam, line 12:
```suggestion
A lot of Istio users would like to connect their virtual machines to their service mesh but the current process is very cumbersome and hard to automate. This series explains a simpler way to connect outside applications securely to your applications running in a service mesh. If you would like to know more about why this might work for you, check out [part 1 here](/blog/2021/simple-vms/).
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Preface
+
+As explained in [part 1](/blog/2021/simple-vms/), to reduce complexity of the virtual machine initialization we will not be registering it with Istio. Instead we will be issuing a certificate directly to the virtual machine and manually configuring the envoy instance to communicate inside the mesh via the east/west gateway. First we will need to organize our certificates so that the service mesh applications will trust these external workloads.
```
rvennam, line 16:
```suggestion
As explained in [part 1](/blog/2021/simple-vms/), to reduce the complexity of virtual machine initialization, we will not be registering it with Istio using the currently [documented approach](/docs/setup/install/virtual-machine/#configure-the-virtual-machine). Instead, we will be issuing a certificate directly to the virtual machine and manually configuring the envoy instance running in the virtual machine to communicate to the mesh via the east/west gateway. First, we will need to organize our certificates so that the service mesh applications will trust these external workloads.
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Certificate Management
+
+Since Istio will not be issuing the certificates for the virtual machines, it will need to be created and managed by some other means. In order for Istio to trust the traffic coming from the virtual machines, the certificates issued must be rooted in the same chain of trust as the service mesh applications.
+
+![Certificate Management](./cert-architecture.png)
+
+The simplest way to organize your certificates is to create a shared root CA. Then two intermediate CA certificates can be issued and used for creating Istio workload certificates and one for virtual machine certificates. The virtual machine CA could easily be managed by a third party certificate management system like Hashicorp's Vault.
```
rvennam, line 24:
```suggestion
The simplest way to organize your certificates is to create a shared root CA. Then, two intermediate CA certificates can be issued and used for creating the Istio workload and virtual machine certificates. The virtual machine CA could easily be managed by a third party certificate management system like Hashicorp's Vault.
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Istio Setup and Deployment
+
+![Istio Setup Deployment](./istio-setup.png)
+
+The setup for Istio to accept outside secured connections is very trivial. Using the IstioOperator configuration an east/west gateway needs to be created and configured to pass through connections.
+
+### Creating the east/west gateway
```
rvennam, line 32:
```suggestion
### Creating the east-west gateway
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+
+![Certificate Management](./cert-architecture.png)
+
+The simplest way to organize your certificates is to create a shared root CA. Then two intermediate CA certificates can be issued and used for creating Istio workload certificates and one for virtual machine certificates. The virtual machine CA could easily be managed by a third party certificate management system like Hashicorp's Vault.
+
+## Istio Setup and Deployment
+
+![Istio Setup Deployment](./istio-setup.png)
+
+The setup for Istio to accept outside secured connections is very trivial. Using the IstioOperator configuration an east/west gateway needs to be created and configured to pass through connections.
```
rvennam, line 30:
```suggestion
The setup for Istio to accept outside secured connections is very trivial. Using the IstioOperator configuration, an east-west gateway needs to be created and configured to passthrough connections.
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+
+### Creating the east/west gateway
+
+Below is an example of an Istio installation of just istiod and the east/west gateway shown above. `PILOT_SKIP_VALIDATE_TRUST_DOMAIN` is enabled so that the virtual machines can have their own SPIFFE ID format.
```
rvennam, line 34:
```suggestion
Below is an example of an Istio installation of just istiod and the east-west gateway shown above. `PILOT_SKIP_VALIDATE_TRUST_DOMAIN` is enabled so that the virtual machines can have their own SPIFFE ID format.
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+            - name: status-port
+              port: 15021
+              targetPort: 15021
+            # Port for mTLS passthrough
+            - port: 15443
+              targetPort: 15443
+              name: tls
+{{< /text >}}
+
+We also need to tell the east/west gateway to forward secure requests to the intended destination. To do this we create a simple Istio Gateway configuration.
```
rvennam, line 84:
```suggestion
We also need to tell the east-west gateway to forward secure requests to the intended destination inside the mesh. To do this we create a simple Istio Gateway configuration.
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+      number: 15443
+      name: tls
+      protocol: TLS
+    tls:
+      mode: AUTO_PASSTHROUGH
+    hosts:
+    - "*"
+{{< /text >}}
+
+The `AUTO_PASSTRHOUGH` config automatically registers the mesh applications with the east/west gateway. You can view the available services by running the following `istioctl` command.
```
rvennam, line 106:
```suggestion
The `AUTO_PASSTRHOUGH` config automatically registers the mesh applications with the east-west gateway. You can view the available services by running the following `istioctl` command.
```

content/en/blog/2021/simple-vms-part-2/index.md
```diff
@@ -0,0 +1,290 @@
...
+0.0.0.0 15443 SNI: outbound_.15014_._.istiod.istio-system.svc.cluster.local;
+0.0.0.0 15443 SNI: outbound_.15012_._.istiod.istio-system.svc.cluster.local;
+0.0.0.0 15443 SNI: outbound_.15010_._.istiod.istio-system.svc.cluster.local;
+{{< /text >}}
+
+## Virtual Machine Configuration
+
+![Virtual Machine Configuration](./vm-configuration.png)
+
+Because we will not be connecting envoy to an external xDS server, we will need to provide the configuration manually. We will need to create an inbound listener for the virtual machine application to communicate through.
```
rvennam, line 137:
```suggestion
Because we will not be connecting the envoy in the virtual machine to an external xDS server, we will need to provide its configuration manually. We will need to create an inbound listener for the virtual machine application to communicate through.
```

content/en/blog/2021/simple-vms-part-2/index.md
rvennam, line 12: Recommend adding another sentence that makes it clear that is only for users who need one-way

howardjohn, line 12: It seems a bit concerning that we are calling Istio's VM operation cumbersome and recommending manually configuring Envoy, requiring deep knowledge of Istio internals that are subject to change in the future. It's not clear to me how this is simpler in any way. I don't feel great about recommending people to not verify trust domains either, at least not without sufficient warning.

nmnellis: They way VMs need to be provisioned today with Istio is very difficult for a number of reasons. All of our customers who are interested in adding their VMs to Istio service meshes are asking for an alternative to the current process. We instead do have customers running the above solution in production already.

howardjohn: What is so hard about the provisioning process that this is easier? I am not sure I understand this - seems like perhaps there is a massive gap. One of the main values of Istio is doing a bunch of things for you so you don't have to - like provisioning certs, configuring envoy, etc. Using alternative cert provisioning mechanisms is perfectly reasonable, but I don't see why users are wanting to manually configure envoy config directly - even the Envoy team states that XDS is *not* intended for humans.

I don't know what our guidelines are for blogs, but I do not *personally* feel good about publishing information that I think is harmful to users and gives Istio a bad reputation. I would much prefer we work towards making the current VM state easier to use.

nmnellis: Here are some of the top frustrations with the current process
* Requirement from CI to communicate with the k8s API before a VM is started
* Issues on-boarding VMS in AWS
  * https://github.com/istio/istio/issues/33268
  * https://github.com/istio/istio/issues/31624
* Iptables manipulation
* No docker support
* Windows support

howardjohn: All really great points - thanks! Just want to clarify as well that I was not trying to say Istio VM story is perfect right now, just that our (long term) strategy should be to improve it rather than sidestep it. Totally understand if users want to do that for now though.

> Requirement from CI to communicate with the k8s API before a VM is started

Totally agree. I think this is actually largely a documentation gap - using an external cert provisioning tool, as discussed here, is perfectly valid 

> AWS

I think this all comes down to the LoadBalancer service having hostname instead of IP? If so, https://github.com/istio/istio/pull/36422/ might resolve some or all of this. If not, worth looking into more.

> iptables

Istio does support running without iptables and using HTTP_PROXY or other mechanisms. However, its under documented (and tested!).

> Docker support

Again, I think this is largely a documentation issue. I don't think there are any technical blockers (after all, we run in docker on k8s).

> Windows

How does it work in this blog. I saw we use `image: gcr.io/istio-release/proxyv2:1.12.0`. In windows case are you using a custom istio/proxy built for windows or just using upstream Envoy?


Overall I think we can improve a lot of the VM UX through improvements to istio.io. I don't think many changes are really needed in Istio code, just to expose them to users.

nmnellis: Thanks for the great responses to the issues. I agree that over time it would be great to improve Istio so this solution isn't needed but a this time a lot of people are simply choosing not to do any mesh VMs. The solution i wrote about actually came from a customer looking to lower the barrier of entry. They aren't the only customer choosing this solution over the existing Istio one. In my opinion, its more beneficial to share how people are actually using Istio than some ideal "right way". The intent of this blog is to not only display how Istio internals work, but add more creative thought on how it can utilized to provide benefit.

nmnellis:
After some discussion, we will be moving this blog over to Solo.io.

craigbox:
How does that decision impact https://istio.io/latest/blog/2021/simple-vms/ ?  Would you like to update it to reflect the new blog being on Solo, or move it to your corporate blog to match?

nmnellis:
> How does that decision impact https://istio.io/latest/blog/2021/simple-vms/ ? Would you like to update it to reflect the new blog being on Solo, or move it to your corporate blog to match?

The use cases and evaluation of how VMs work seem relevant to the larger istio community. I can remove the mention of part 2 if that is better, or i can also link it to Solos site once we get it up on our side.

craigbox:
I'm OK with either; please send a PR when you've published.

PR #10902, opened 2022-02-13
Experiment with setting Netlify URL in the RSS

craigbox:
Fixes #8357, eventually.

When you run `make serve` locally, it sets a base URL of `http://localhost:1313/latest/`, but `make netlify` just uses `/latest`. 

This PR fetches the Netlify URL from `DEPLOY_PRIME_URL` and tacks it on the front of the static entries in the RSS feed/s.

PR #10922, opened 2022-02-15
Change the env var used by the RSS feed

craigbox:
#10902 doesn't use the right URL.  Changing the variable from `DEPLOY_PRIME_URL` to `URL` may fix this.


craigbox:
https://deploy-preview-10922--preliminary-istio.netlify.app/latest/feed.xml now shows `preliminary.istio.io` as the URL.  This, while not technically correct for a branch build, is probably the right outcome and probably means that main RSS feed — the only one we care about — will work.

ericvn (approved):

craigbox:
https://preliminary.istio.io/latest/blog/feed.xml LGTM, so cherrypicking.

Issue #10924, opened 2022-02-16
Accessibility update aria links


Arhell:
![Screenshot_2](https://user-images.githubusercontent.com/26163841/154170224-b922261e-173a-4e7e-8247-970b02f8fd57.png)

PR #10929, opened 2022-02-16
Accessibility update aria links

Arhell:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

resolve https://github.com/istio/istio.io/issues/10924

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -37,7 +37,7 @@
 
                 {{ range $supportedLanguages}}
                     {{ $active := eq $currentLanguage .code }}
-                    <a tabindex="-1" role="menuitem" lang="{{.code}}" id="switch-lang-{{.code}}" class="footer-languages-item {{ if $active }}active{{ end }}">
+                    <a tabindex="-1" lang="{{.code}}" id="switch-lang-{{.code}}" class="footer-languages-item {{ if $active }}active{{ end }}">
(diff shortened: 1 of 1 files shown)


Arhell:
cc @raigbox
no errors after checking

Arhell:
cc @craigbox
no errors after checking

craigbox:
LGTM, but I'm not familiar with the tabindex settings.  Can you please give me a sentence explaining the change?

Arhell:
> LGTM, but I'm not familiar with the tabindex settings. Can you please give me a sentence explaining the change?

just clarified, left it, it does not affect the check

Arhell:
here is what is written in Accessibility Insights for Web

> An ARIA role attribute can be added to an element to instruct assistive technologies to treat the element as something other than its native HTML element type. For example, an <li> element with role="option" is to be treated as a selectable option in a listbox control, not as a static list item.
> 
> Some ARIA "child" roles identify managed controls that are always part of a larger composite control, identified by a "parent" role. For example, role="option" identifies a child control that is managed by a parent control identified by role="listbox". People who use assistive technologies might find it difficult or impossible to use a child control if its managing control lacks the required parent role.

for our page I think this attribute is redundant

craigbox (approved):

PR #10943, opened 2022-02-18
Add note about gateways to protocol selection

howardjohn:


content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -63,3 +63,15 @@ spec:
   - number: 80
     name: http-web
 {{< /text >}}
+
+## Gateways
```
howardjohn, line 67: This is a bit subtle, here is what I wanted to get across:

* Downsteam protocol and upstream protocol are unrelated by default for Gateway, but not for mesh
* This means you ought to configure the upstream protocol if you want HTTP/2, since otherwise we use HTTP/1
* If you really know what you are doing, you can set useClientProtocol, but should be advised that many clients WILL use http/2, even if you didn't want them to, due to ALPN negotiation.

frankbu, line 67:
```suggestion
## HTTP gateway protocol selection
```

content/en/docs/ops/configuration/traffic-management/protocol-selection/index.md
```diff
@@ -63,3 +63,15 @@ spec:
...
+
+For HTTP gateways, if the protocol is not explicitly declared to be HTTP/1.1 (`http`) or HTTP/2 (`http2` or  `grpc`), the Gateway cannot automatically detect which protocol
+to use when forwarding to the backend Service.
+Instead, it will always use HTTP/1.1.
+If you want to automatically use the protocol the client used, the [`useClientProtocol`](https://istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-HTTPSettings) option can be enabled for that Service.
+
+{{< warning >}}
+HTTPS Gateways will always [advertise](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation) support for HTTP/1.1 and HTTP/2.
+This means that modern clients will often use HTTP/2, which may cause issues when using `useClientProtocol` if your application does not support HTTP/2.
+{{< /warning >}}
```
frankbu, line 77:
```suggestion
Unlike sidecars, gateways are by default unable to automatically detect the specific HTTP protocol to use when forwarding requests to backend services. Therefore, unless explicit protocol selection is used to specify HTTP/1.1 (`http`) or HTTP/2 (`http2` or  `grpc`), gateways will forward all incoming HTTP requests using HTTP/1.1.

Instead of using explicit protocol selection, you can instead instruct gateways to forward requests using the same protocol as the incoming request by setting the [`useClientProtocol`](/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-HTTPSettings) option for a Service. Note, however, that using this option with services that do not support HTTP/2 can be risky, because HTTPS gateways always [advertise](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation) support for HTTP/1.1 and HTTP/2, so modern clients will often choose HTTP/2.
```

frankbu (approved):

PR #11000, opened 2022-03-02
Removed the helm istio-operator install

litong01:
Since the istio-operator helm charts have been
removed, this is no longer possible to do. Remove
the instructions. Please also see this issue

https://github.com/istio/istio/issues/36157

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/operator/index.md
+++ b/content/en/docs/setup/install/operator/index.md
@@ -68,27 +68,6 @@ $ istioctl operator init --watchedNamespaces=istio-namespace1,istio-namespace2
 
 See the [`istioctl operator init` command reference](/docs/reference/commands/istioctl/#istioctl-operator-init) for details.
 
-{{< tip >}}
-You can alternatively deploy the operator using Helm:
-
-1. Create a namespace `istio-operator`.
-
-    {{< text bash >}}
-    $ kubectl create namespace istio-operator
-    {{< /text >}}
-
-1. Install operator using Helm.
-
-    {{< text bash >}}
-    $ helm install istio-operator manifests/charts/istio-operator \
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/install/operator/index.md
```diff
@@ -68,27 +68,6 @@ $ istioctl operator init --watchedNamespaces=istio-namespace1,istio-namespace2
...
-    $ helm install istio-operator manifests/charts/istio-operator \
-        --set watchedNamespaces="istio-namespace1\,istio-namespace2" \
-        -n istio-operator
-    {{< /text >}}
-
-Note that you need to [download the Istio release](/docs/setup/getting-started/#download)
-to run the above command.
-{{< /tip >}}
-
 {{< warning >}}
```
ericvn, line 71: Also remove the warning? Talks about creating the namespace which is being removed in the tip.

litong01: will remove

ericvn:
The removed commands work fine for me. Trying to determine what is really removed.

content/en/docs/setup/install/operator/index.md
ericvn: Actually, re-reading this, the `init` is what creates the namespace, and you are not updating that, so you could leave the tip. It talks about differences from 1.10 which is no longer supported, so removing it is fine as well.

litong01:
The public helm repo has removed the operator charts, however, a user can still use local charts to install operator, thus leave the instruction alone, but one who is familiar with helm and not wanting to get charts first, will run into issues.

PR #11022, opened 2022-03-08
doc:remove the * character in istio operator install doc

tanjunchen:

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

image


craigbox:
The * is there to indicate that the Istio operator is experimental.  It's working as intended.  What caused you to want to change this?

(You're also submitting this PR to the archive for 1.12; what we would do if we agreed we needed to make a change was commit the change and regenerate the archive HTML.)

tanjunchen:
> The * is there to indicate that the Istio operator is experimental. It's working as intended. What caused you to want to change this?
> 
> (You're also submitting this PR to the archive for 1.12; what we would do if we agreed we needed to make a change was commit the change and regenerate the archive HTML.)

ok. I get it. Because I feel that it looks a bit ugly. If it is an experimental feature, I suggest putting it at the front of this article and giving it a hint.

ericvn (changes requested):
As noted, this is as designed. See https://github.com/istio/istio.io/issues/9766.

tanjunchen:
> As noted, this is as designed. See #9766.

get it.thanks.
/close

PR #11060, opened 2022-03-16
fix page display

zirain:

the root case is table without `table-layout: fixed;`

fix table display

before:
image

after:

image

@ericvn PTAL.

### DIFF
--- a/content/en/docs/tasks/observability/logs/access-log/index.md
+++ b/content/en/docs/tasks/observability/logs/access-log/index.md
@@ -85,6 +85,7 @@ Istio will use the following default access log format if `accessLogFormat` is n
 
 The following table shows an example using the default access log format for a request sent from `sleep` to `httpbin`:
 
+{{<fixed-table "fixed-table">}}
 | Log operator | access log in sleep | access log in httpbin |
 |--------------|---------------------|-----------------------|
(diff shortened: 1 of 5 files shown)


craigbox:
What are the implications for this on mobile?
If it's OK, shouldn't this CSS be applied to all tables?

zirain:
> What are the implications for this on mobile? 

I have check via chrome mobile monitor, it's ok.

> If it's OK, shouldn't this CSS be applied to all tables?

maybe someone who is familiar with infra can answer.

ericvn:
@davidhauck Thoughts?

ericvn:
This PR will fail to cherry-pick until the otel provider PR merges. Maybe wait and run the cherry-pick at that point.

davidhauck:
I think we should be fine to apply this to all tables on the site, I can't think of anything off the top of my head that it would break.

zirain:
> This PR will fail to cherry-pick until the otel provider PR merges. Maybe wait and run the cherry-pick at that point.

I'll try to apply to all tables on this site, after that it would be fine to cherry-pick.

craigbox:
PTAL at #11065

zirain:
duplicate with https://github.com/istio/istio.io/pull/11065

PR #11061, opened 2022-03-16
[release-1.13] Update istio test reference

zirain:

Update istio test reference
this will fix https://github.com/istio/istio.io/pull/11058

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.5
 require (
 	github.com/golang/sync v0.0.0-20180314180146-1d60e4601c6f
 	github.com/pmezard/go-difflib v1.0.0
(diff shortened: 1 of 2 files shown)


ericvn (changes requested):
So this seems like we will need a version of istio post 1.13.2 (which should be the old test reference). Once a new version of 1.13 is released that supports this, we can merge the original doc and test.

zirain:
> So this seems like we will need a version of Istio post 1.13.2 (which should be the old test reference). Once a new version of 1.13 is released that supports this, we can merge the original doc and test.

need this commit https://github.com/istio/istio/commit/b70734d0558b13e15879ebb9847e7e9529bcb83a

ericvn:
This PR can be closed as we have updated the tests to use the 1.13.3 test reference. That PR fixes https://github.com/istio/istio.io/pull/11058 and that PR can be merged once 1.13.3 is live.

Issue #11070, opened 2022-03-17
Hugo warnings


ericvn:
I see a Hugo warning in the logs:
```
WARN 2022/03/17 09:23:41 markup.defaultMarkdownHandler=blackfriday is deprecated and will be removed in a future release. See https://gohugo.io//content-management/formats/#list-of-content-formats

Re-run Hugo with the flag --panicOnWarning to get a better error message.
Start building sites …
hugo v0.93.3-44E3C002 linux/amd64 BuildDate=2022-03-08T10:15:28Z VendorInfo=gohugoio
WARN 2022/03/17 09:24:42 .Path when the page is backed by a file is deprecated and will be removed in a future release. We plan to use Path for a canonical source path and you probably want to check the source is a file. To get the current behaviour, you can use a construct similar to the one below:

  {{ $path := "" }}
  {{ with .File }}
	{{ $path = .Path }}
  {{ else }}
	{{ $path = .Path }}
  {{ end }}


Re-run Hugo with the flag --panicOnWarning to get a better error message.
```

ericvn:
Blackfriday deprecation also noted here: https://github.com/istio/istio.io/issues/10341

ericvn:
We also use an older Hugo version for the build:
https://github.com/istio/istio.io/blob/master/netlify.toml#L5

PR #11090, opened 2022-03-22
Attempt to build on netlify with 0.95.0 hugo

jacob-delgado:
Attempt to upgrade hugo on netlify

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,7 @@
     publish = "public"
 
 [build.environment]
-    HUGO_VERSION = "0.61.0"
+    HUGO_VERSION = "0.95.0"
     NODE_VERSION = "12.8.0"
     BUILD_WITH_CONTAINER = "0"
 
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,7 +3,7 @@
 
 [build.environment]
     HUGO_VERSION = "0.95.0"
-    NODE_VERSION = "12.8.0"
+    NODE_VERSION = "12.22.11"
     BUILD_WITH_CONTAINER = "0"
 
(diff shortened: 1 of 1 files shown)


jacob-delgado:
See https://github.com/istio/istio.io/issues/11070

ericvn:
@davidhauck Not sure how to best verify that the updated Hugo, etc don't have an adverse effect. Is it a wait and see if anyone finds something wrong with preliminary.istio.io? How about logs from the build? Can we compare to what we see in our current pipeline? Similar to https://github.com/istio/istio.io/issues/11070?

craigbox:
When we change the markdown renderer, we should diff the `public` directory before and after.  I'm not sure there's a useful equivalent for here?  Download a few pages before/after?

jacob-delgado:
The pages now have better human readable html in the newer Hugo versions. I'm not sure a diff would show anything.

davidhauck:
I'm not sure there's a great way to check what the differences are with the newer hugo version. Probably just wait and see if anyone finds anything wrong with preliminary.istio.io. 

I do remember trying to update hugo in the past and had some problems. Maybe the newest hugo versions don't have those problems any more. Either way, we should look around in the preview site to make sure everything is still working

ericvn (approved):
I'm fine adding this for now. We'll want to fix some of the other items, change from blackfriday, etc. but can do in another PR. We can revert this if we find issues with preliminary.istio.io.

PR #11103, opened 2022-03-24
Fix workload aggregating rules for TCP byte metrics

### LINKED ISSUE #10311: Observability Best Practices: example recording rules do not record TCP traffic
In Observability Best Practices (../docs/ops/best-practices/observability/#workload-level-aggregation-via-recording-rules) in the example prometheus configuration, the last six recording rules refer to istio_tcp_sent_bytes_total_count, istio_tcp_sent_bytes_total_sum, istio_tcp_sent_bytes_total_bucket, istio_tcp_received_bytes_total_count, istio_tcp_received_bytes_total_sum, istio_tcp_received_bytes_total_bucket.

Following the Integrations: Prometheus Option 2: Customized scraping configurations these metrics do not exist (I cant say if they do exist if following opt 1 as I have not tried that), so the TCP sent/received bytes data does not get aggregated or scraped by the 'production' prometheus. 

However, the metrics istio_tcp_sent_bytes_total and istio_tcp_received_bytes_total do exist, so replacing the last six recording rules with something like the following is one way to address the problem:
```yaml
        - record: "workload:istio_tcp_sent_bytes_total"
          expr: |
            sum without(instance, namespace, pod_name) (istio_tcp_sent_bytes_total)

        - record: "workload:istio_tcp_received_bytes_total"
          expr: |
            sum without(instance, namespace, pod_name) (istio_tcp_received_bytes_total)
```

nickgooding:
Fixes some broken suggested recording rules that treated
`istio_tcp_sent_bytes_total` and `istio_tcp_sent_received_total` as if
they were distribution type metrics rather than counters.

I have also reordered the metrics to more closely align with the order
on the "Istio Standard Metrics" page.

Fixes #10311

- [X] Docs
- [ ] Installation
- [ ] Networking
- [x] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/best-practices/observability/index.md
+++ b/content/en/docs/ops/best-practices/observability/index.md
@@ -81,6 +81,14 @@ groups:
     expr: |
       sum without(instance, kubernetes_namespace, kubernetes_pod_name) (istio_response_bytes_bucket)
 
+  - record: "workload:istio_tcp_sent_bytes_total"
+    expr: |
+      sum without(instance, kubernetes_namespace, kubernetes_pod_name) (istio_tcp_sent_bytes_total)
+
+  - record: "workload:istio_tcp_received_bytes_total"
+    expr: |
+      sum without(instance, kubernetes_namespace, kubernetes_pod_name) (istio_tcp_received_bytes_total)
+
   - record: "workload:istio_tcp_connections_opened_total"
     expr: |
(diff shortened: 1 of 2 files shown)


ericvn:
@istio/wg-policies-and-telemetry-maintainers Would you please review.

ericvn (approved):

PR #11134, opened 2022-04-01
Add 'istio' and 'kubernetes' to .spelling

my-git9:
Add 'istio' and 'kubernetes' to .spelling

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -457,6 +457,7 @@ IPs
 iptables
 IPv4
 IPv6
+istio
 Istio
 istio-agent
 istio-control
@@ -536,6 +537,7 @@ kubeconfig
 kubectl
 Kubelet
 kubelet
+kubernetes
 Kubernetes
 kubernetes.default
 kube-proxy


ericvn (changes requested):
Thanks for the PR, but istio and kubernetes are incorrect spellings and need capitalization, thus the reason they are not in the .spelling file.

ericvn:
I do see in one of the PR's the request that this change should be added. 

It is related to specifying a link for a header and the normal link for a header is all lowercase. Example from another file:
```
## Istio 的好处{#benefits-of-Istio}
```
The spellchecker checks each of the entries in this link, and if it was `istio` as it is by default, that fails. So in the translated files, where we actually specify the links, we have some of the links using uppercase characters to satisfy the spelling. So far that seems to be the norm. In the English docs, we don't override the links, so they aren't spellchecked. 

I'm reluctant to add the lowercase versions of the specific names as it would allow someone later to possibly not capitalize the name. I'm wondering if we can not spellcheck the specified links.

@craigbox What are your thoughts.

craigbox:
I like the idea of not spell checking link anchors, but I don't know how much effort that is.

ericvn:
https://github.com/istio/istio.io/pull/11159 removes the need to add entries for heading anchors.

Issue #11140, opened 2022-04-04
Helm install test is failing in 1.13


ericvn:
The same test was failing in master and in looking at that test, the install uses the latest version on the manifests which changed and needed the istio reference updated. 

I'm wondering if the test here is also using, incorrectly, the master `latest` manifests.

ericvn:
Confirmed the ingress pod is using:
```
  Normal   Pulling    28s                kubelet            Pulling image "gcr.io/istio-testing/proxyv2:latest"
```

ericvn:
Fixed by https://github.com/istio/istio.io/pull/11142

PR #11142, opened 2022-04-04
[release-1.13] Use the currently referenced images in the helm test

ericvn:

Fixes: https://github.com/istio/istio.io/issues/11140

Will cherry-pick to main branch later as that is currently passing.


content/en/docs/setup/install/helm/test.sh
```diff
@@ -27,7 +27,7 @@ function rewrite-repo() {
   cmd="$(echo "${cmd}" | sed 's|istio/base|manifests/charts/base|')"
   cmd="$(echo "${cmd}" | sed 's|istio/istiod|manifests/charts/istio-control/istio-discovery|')"
   cmd="$(echo "${cmd}" | sed 's|istio/gateway|manifests/charts/gateway|')"
-  eval "${cmd}"
+  eval "${cmd} --set global.tag=${ISTIO_IMAGE_VERSION=SHOULD_BE_SET}.${ISTIO_LONG_SHA=latest}"
```
ericvn, line 30: I need to set default values for these, or override the `-u`. These values are set in the Makefile. If the values aren't set, the defaults will lead to the test failing which is OK as there aren't really a good set of defaults.

frankbu (approved):

PR #11144, opened 2022-04-04
Pin basic auth extension image at 1.12.0.

bianpengyuan:
There is no breaking change at Wasm ABI since Istio 1.8, so no need to release wasm extensions continuously with Istio release. This change pin the version of basic auth module at 1.12.0.


craigbox:
Version mismatch check: I assume you haven't built it for 1.13 and don't intend to do so?

bianpengyuan:
> Version mismatch check: I assume you haven't built it for 1.13 and don't intend to do so?

Yes, that is the intention here.

craigbox (approved):

PR #11151, opened 2022-04-06
Fix 1.12.6 release date

dgn:
The release was held up for a week, so now the date is wrong.


craigbox (approved):
should have caught this.  thanks!

PR #11155, opened 2022-04-07
Change multicluster docs to use istiodless remotes

frankbu:

Depends on https://github.com/istio/istio/pull/38283

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/multicluster/before-you-begin/index.md
+++ b/content/en/docs/setup/install/multicluster/before-you-begin/index.md
@@ -56,8 +56,17 @@ detailed descriptions and instructions for all available options.
 Depending on which option you choose, the installation instructions for
 Istio may change slightly.
 
+{{< tip >}}
+If you are planning to deploy only one primary cluster (i.e., one of the
+Primary-Remote installations, below), you will only have a single CA
+(i.e., `istiod` on `cluster1`) issuing certificates for both clusters.
+In that case, you can skip the following CA certificate generation step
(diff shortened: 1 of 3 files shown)


craigbox (approved):

stevenctl:
cc @nmittler

stevenctl:
DNM just so we don't accidentally merge before getting the istio/istio PR in.

stevenctl (commented):
Just to confirm I understand all the pieces here:

* We install istio in the primary cluster
  * EXTERNAL_ISTIOD makes it so the primary will try to write things on remote clusters
* We install istio on the remote cluster
  * Create and annotate the namespace to allow the primary to write things
  * Install the external profile which sets webhooks to point at the primary cluster/stops istiod from getting installed
  * Push a remote secret so the primary can start doing discovery. 
  
LGTM; but I think the one missing thing is clarity on what a primary cluster is writing into the remote clusters (certs on webhooks, certs on the namespace controller). Not sure exactly where that fits in, but its a new thing that users will care a  lot a bout when they start running into issues.

stevenctl:
Also, for production clusters I think we talked about recommending using properly signed DNS certs for webhook access across clusters. I understand this takes extra work to test in our integration tests but it still seems important.

stevenctl:
cc @costinm

Issue #11156, opened 2022-04-08
Feature request - Ignore spelling check for heading anchors


ericvn:
In the English docs, heading anchors are not typically specified and we us the default. An example might be `## Benefits of istio` would be referenced using #benefits-of-Istio. 

In the translated docs, we have an issue in that we need to specify the heading anchor (ex: ## Istio 的好处{#benefits-of-istio}). If we specify #benefits-of-istio as the anchor, the spellchecker does not like istio, and we have been uppercasing them to pass (ex: ## Istio 的好处{#benefits-of-Istio}).

We could add istio (and kubernetes and maybe others) to the dictionary, but the we lose the ability to catch cases where they are not capitalized in other docs.

Is it possible to ignore spellcheck of the specified heading anchors?

ericvn:
See https://github.com/istio/istio.io/pull/11134

PR #11159, opened 2022-04-09
Remove heading anchors before spell-checking

ericvn:

Fixes https://github.com/istio/istio.io/issues/11156

Update the lint_site script to also remove heading anchors from the temporary files before spell checking.


craigbox (approved):
tested a few examples on https://regex101.com/, LGTM

```### This is a string {#this-is-a-string}
### This is a string
This is not a string {though it looks like one}
# This is a string {#this-is-a-string}
### This is a string {#this-is-a-string}
This is not a string {#though it looks like one}
[URL](https://en.wikipedia.org/wiki/Example#URL)```

Issue #10341, opened 2021-09-15
Change markdown generator for istio.io


craigbox:
We currently (#6105) use the `blackfriday` Markdown generator in Hugo.  It is deprecated (Hugo throws a warning as of 0.87) and will be removed in an upcoming release:

``` hugo serve --baseURL "http://localhost:1313/latest/" --bind 0.0.0.0 --disableFastRender
WARN 2021/09/15 12:39:55 markup.defaultMarkdownHandler=blackfriday is deprecated and will be removed in a future release. See https://gohugo.io//content-management/formats/#list-of-content-formats
Start building sites … 
hugo v0.87.0+extended darwin/amd64 BuildDate=unknown
```

We should look to change (back?) to `goldmark`.

howardjohn:
https://github.com/istio/istio.io/pull/11161

PR #11161, opened 2022-04-10
remove blackfriday

Arhell:
resolve https://github.com/istio/istio.io/issues/11070 and https://github.com/istio/istio.io/issues/10341

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -17,23 +17,6 @@ staticDir = ["static", "generated"]
 disableAliases = true
 
 [markup]
-  defaultMarkdownHandler = "blackfriday"
-  [markup.blackfriday]
-    angledQuotes = false
-    footnoteAnchorPrefix = ""
-    footnoteReturnLinkContents = ""
-    fractions = true
-    hrefTargetBlank = false
-    latexDashes = true
-    nofollowLinks = false
-    noreferrerLinks = false
-    plainIDAnchors = true
-    skipHTML = false
-    smartDashes = true
-    smartypants = true
-    smartypantsQuotesNBSP = false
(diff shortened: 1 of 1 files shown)


ericvn:
I did take a quick look at https://deploy-preview-11161--preliminary-istio.netlify.app/ and didn't see anything obviously wrong. I did try links, buttons, and some other items. Will let others check, and if noting noted, we can merge and hopefully get more eyes during the testing week.

ericvn:
This does not fix #11070 as there are several remaining. From there lint_istio details:
```
Start building sites … 
hugo v0.95.0-9F2E76AF linux/amd64 BuildDate=2022-03-16T14:20:19Z VendorInfo=gohugoio
WARN 2022/04/10 17:19:42 .Path when the page is backed by a file is deprecated and will be removed in a future release. We plan to use Path for a canonical source path and you probably want to check the source is a file. To get the current behaviour, you can use a construct similar to the one below:

  {{ $path := "" }}
  {{ with .File }}
	{{ $path = .Path }}
  {{ else }}
	{{ $path = .Path }}
  {{ end }}


Re-run Hugo with the flag --panicOnWarning to get a better error message.
```

ericvn (approved):
No other comments. Let's merge this and see if there is any negative feedback during the upcoming testing.

craigbox:
Sorry, I was late to this.  I'd be pleasantly surprised if there's no obvious changes; I've seen all sorts of things changing between these engines before. We should definitely make sure it's on the agenda for testing!

(I did a quick look for screenshot comparison tools but haven't installed anything yet.)

PR #11174, opened 2022-04-14
fix Hugo warnings

Arhell:
resolve https://github.com/istio/istio.io/issues/11070


ericvn (changes requested):
I'm seeing this now:
```
WARN 2022/04/14 11:55:28 .File.Path on zero object. Wrap it in if or with: {{ with .File }}{{ .Path }}{{ end }}
```

ericvn (approved):
Latest changes look good in `make serve` and I took a cursory look at the netlify output.

PR #10719, opened 2022-01-04
WIP - Test updating reference docs in /zh/

ericvn:
Please provide a description for what this PR is for.

This is a test PR to see what happens if we would try and update the reference doc files in the /zh/ tree from the upstream repos similar to how the /en/ files are.

PR #11182, opened 2022-04-17
zh-translation: docs/reference/config/networking/destination-rule/ind…

my-git9:

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/zh/docs/reference/config/networking/destination-rule/index.html
+++ b/content/zh/docs/reference/config/networking/destination-rule/index.html
@@ -3,7 +3,7 @@ WARNING: THIS IS AN AUTO-GENERATED FILE, DO NOT EDIT. PLEASE MODIFY THE ORIGINAL
 source_repo: https://github.com/istio/api
 title: Destination Rule
 description: Configuration affecting load balancing, outlier detection, etc.
-location: https://istio.io/zh/docs/reference/config/networking/destination-rule.html
+location: https://istio.io/docs/reference/config/networking/destination-rule.html
 layout: protoc-gen-docs
 generator: protoc-gen-docs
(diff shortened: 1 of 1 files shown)


zirain:
reference pages ares auto-generated by bots.
@ericvn do we update the infra?

ericvn:
@zirain I had looked into the possibility of updating the translated reference docs to match the English versions in a prior PR, but did not get feedback if we should make it happen and I closed it. For reference, it was https://github.com/istio/istio.io/pull/10719.

@craigbox Thoughts on re-doing my earlier PR to update the `zh` version of the reference files as part of our normal automated reference doc updates?

craigbox:
not enough context to have an opinion sorry.

kebe7jun:
> @zirain I had looked into the possibility of updating the translated reference docs to match the English versions in a prior PR, but did not get feedback if we should make it happen and I closed it. For reference, it was #10719 and that referred to #10717.
> 
> @craigbox Thoughts on re-doing my earlier PR to update the `zh` version of the reference files as part of our normal automated reference doc updates?

I think it is necessary to update the reference of `zh` automatically, this part of the content is not quite suitable for manual translation into Chinese, it may be more obscure and difficult to understand, build at this stage, if the update can be automatic, keep the accuracy of the content I think more good.

ericvn:
Talking with @craigbox I think it makes sense to update the translated docs at the same time as the English docs. I'll create a PR for that soon.

ericvn:
https://github.com/istio/istio.io/pull/11252 was created to update the `zh` side of the reference docs at the same time as the `en` directory. The daily automated job would now update both.

PR #11188, opened 2022-04-18
1.13 notes cp

stevenctl:
cherry-pick 1.13 notes


ericvn (approved):
Looks good. Marking DNM so you can unlabel and merge when ready.

ericvn:
Looks like we need to do two things in the `istio.io` related repo going forward. Besides updating data/args.yml, we also need to do a `make gen` as a page is using the latest version.

Issue #11190, opened 2022-04-18
update netlify hugo 


Arhell:
update netlify hugo last version

PR #11193, opened 2022-04-19
update netlify hugo

Arhell:
resolve https://github.com/istio/istio.io/issues/11190

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,7 @@
     publish = "public"
 
 [build.environment]
-    HUGO_VERSION = "0.95.0"
+    HUGO_VERSION = "0.97.3"
     NODE_VERSION = "12.22.11"
     BUILD_WITH_CONTAINER = "0"
 


craigbox:
Anything in particular this version fixes or that we should watch out for in testing?
Or is this just for being up to date?

Arhell:
Yes, just up to date.

craigbox:
Looking at https://github.com/gohugoio/hugo/releases, there's been a bunch of 0.97 point releases in the last few days.  I think it might be best to leave this sit for a week or so and see if they find any other major bugs.  Would that be OK, @Arhell ?

Arhell:
> Looking at https://github.com/gohugoio/hugo/releases, there's been a bunch of 0.97 point releases in the last few days. I think it might be best to leave this sit for a week or so and see if they find any other major bugs. Would that be OK, @Arhell ?

There were already bugs in 0.97.2. There were problems with the build.
I checked, everything works + I did not find any bugs for our site.

craigbox (approved):
ok, we can always upgrade/downgrade I guess.

ericvn:
Not sure why these tests are failing. I'll try and look later today.

PR #11196, opened 2022-04-19
zh-translation:sync virtual-machines/test.sh

my-git9:

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/zh/docs/examples/virtual-machines/test.sh
+++ b/content/zh/docs/examples/virtual-machines/test.sh
@@ -64,7 +64,7 @@ run_in_vm "
 
 # We do not have systemd, need to start mysql manually
 docker exec --privileged -d vm mysqld --skip-grant-tables
-# Wait for mysql to be ready
+# Wait for mysql to become ready
 run_in_vm "while ! sudo mysql 2> /dev/null; do echo retrying mysql...; sleep 5; done"
 
 run_in_vm snip_running_mysql_on_the_vm_3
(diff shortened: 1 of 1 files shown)


zirain:
@ericvn do you think we need do test in other language
IMHO, tests running on English is enough.

ericvn:
@zirain I'm not convinced that the translated document tests would continue to pass. In some instances, we need to make changes to the English pages, say in the command output or the commands as well as the tests scripts as Istio evolves. Since we don't update the translated packages/tests at the same time as we do the English pages, I expect that the non-English tests may become broken over time.

The problem is support for those translated pages/tests. In some cases, I may have spent a week to get the English pages back to passing (after switching the back to testing against the main branch following a release). I'm not sure who we can get to do the same things on the translated pages. Making the tests non-required would allow PRs to continue to merge, but I don't want to continually have non-required tests failing.

This does bring up whether we should be showing any of the translated pages with a check mark meaning we are running a test against them.

zirain:
@ericvn refer to [here](https://github.com/istio/istio.io/blob/5bc9c873e362f64b96eeb1c378475be1b6ae5c7f/tests/tests.mk#L48), I think we do not need to update tests in other language.

ericvn:
@zirain Correct. We only run tests against the English docs today. As I noted above, if we wanted to run against the translated docs, it may or may not work. My prior remarks were against enabling tests on the translated pages, unless support would come from some group. Translated tests don't need to be updated unless those tests would be enabled at some point.

ericvn (approved):
Their is no translated text in this PR. only English updates.

PR #11220, opened 2022-04-24
Update Istio test ref from latest on 1.14 branch

ericvn:
Update the Istio test reference to the latest on the 1.14 branch.

Since https://github.com/istio/istio.io/pull/11212 hasn't merged yet, the pipeline is referencing commits on the Istio main branch (now for 1.15), instead of the 1.14 branch.

This is pulling what is currently the latest commit in the 1.14 branch.


ericvn:
However, looking at the repository, I'm not sure the images are actually being built in the 1.14 pipeline...

PR #11252, opened 2022-05-02
Update reference docs in zh path

ericvn:

The first pass of this PR is to update the `zh` path reference docs by modifying the grab_reference_docs script (new script also checked in). This resulted in files being deleted, new files being created, as well as many modified files. Given the updated reference docs, see if the various linting checks still pass and there aren't outstanding errors within the docs pointing the the newly updated reference docs or pointing to removed docs.

If this works, then we can try updating the grab_reference_docs script to simply update both sets of documents when it is run. This would be in the next commit.


content/zh/docs/reference/config/security/peer_authentication/index.html
```diff
@@ -7,7 +7,7 @@
 layout: protoc-gen-docs
 generator: protoc-gen-docs
 schema: istio.security.v1beta1.PeerAuthentication
-aliases: [/docs/reference/config/security/v1beta1/peer_authentication]
+aliases: [/zh/docs/reference/config/security/v1beta1/peer_authentication]
```
ericvn, line 10: Does the /zh make sense here?

ericvn:
Lint passed on commit `#1`, so updating grab_reference_docs to do both `en` and `zh` branches. Running the script showed no changes in the reference files from prior commit. Commit `#2` has the updated script.

kebe7jun (approved):

content/zh/docs/reference/config/security/peer_authentication/index.html
kebe7jun: yes I think.

craigbox (approved):

PR #11254, opened 2022-05-02
hugo update

Arhell:


ericvn:
I did test locally last week with Hugo 0.98.0 and things didn't seem worse. It was part of my testing to see if the blackfriday change was causing an issue on the Ecosystem page, or if was the Hugo update. That issue seems related to changing from blackfriday and not the Hugo release.

Arhell:
> I did test locally last week with Hugo 0.98.0 and things didn't seem worse. It was part of my testing to see if the blackfriday change was causing an issue on the Ecosystem page, or if was the Hugo update. That issue seems related to changing from blackfriday and not the Hugo release.

This up to date update

ericvn (changes requested):
Let this sit for a while unless we really need to move up. The 1.14 build-tooling has Hugo 0.97.3 and the main istio.io branch will become the 1.14 branch at 1.14 release so it make sense to leave this at 0.97.3. We can update the main branch build tooling and Hugo here after the release.

PR #11281, opened 2022-05-09
Add webhooks to test framework snapshot

frankbu:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/examples/virtual-machines/test.sh
+++ b/content/en/docs/examples/virtual-machines/test.sh
@@ -104,6 +104,6 @@ _verify_elided run_curl "${snip_reaching_kubernetes_services_from_the_virtual_ma
 # @cleanup
 docker stop vm
 kubectl delete -f samples/multicluster/expose-istiod.yaml -n istio-system --ignore-not-found=true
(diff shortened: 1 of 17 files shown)


frankbu:
The cleanup instructions in the various docs are a mix. Some do complete uninstall,  some do partial, and some have no cleanup instructions at all. The ones that have partial or nothing have uninstall instructions in the test.sh, but many of those are only partially correct. Bottom line is that we need figure out what should be the "correct" way to cleanup (might require fixes in istioctl commands also), and then we can do a pass through the docs/tests to do a better job, but that is out of scope for this PR which is just adding more to the before/after snapshot checking and patching up the broken tests (consistently with how they are currently written). There are some comments in the code mentioning some specific TODOs for cleaning up tests.

ericvn (approved):
It looks like my comment on cleanup and tests using what is in the docs is only happening for 3 of the changed tests, two cert-management and another.  All the other tests are doing random cleanup in the script outside of the doc.

So while my concern is correct, it seems that we are already doing something else for a bunch of tests (which we should clean up later as mentioned by Frank).

PR #11291, opened 2022-05-11
Wildcard egress: remove arbitrary domain section

howardjohn:
This doc has been a nuisance for many years. It recommends an extremely
complex and dangerous pattern, relying on deploying nginx, extremely
complex EnvoyFilters enabling unsupported, custom, alpha Envoy c++
filters, and a number of other scary practices. IMO this does not belong
in Istio docs at all, and certainly not in our top level taks.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


frankbu:
I think we should either remove this whole document, or if not, update the earlier material in the doc based on the second section being removed. For example, earlier in the doc it says: `The first, and simplest, way to access a set of hosts within a common domain ...`, so that needs to be reworded given that there is only the "first way" in the doc. It should probably, at a minimum, explain the limitation of the config used in the task, and allude to what would need to be done in the general case, even if it doesn't include an example config in the task.

content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md
```diff
@@ -118,520 +118,6 @@ the default) is used in the service entry below.
...
-the set of wildcard domains are served by a single common host.
-This is the case for _*.wikipedia.org_. All of the language-specific sites are served by every
-one of the _wikipedia.org_ servers. You can route the traffic to an IP of any _*.wikipedia.org_ site,
-including _www.wikipedia.org_, and it will [manage to serve](https://en.wikipedia.org/wiki/Virtual_hosting)
-any specific site.
-
-In the general case, where all the domain names of a wildcard are not served by a single hosting server,
-a more complex configuration is required.
-
-### Wildcard configuration for a single hosting server
```
frankbu, line 133: Does this section need to be deleted too? Seems that it's really only the next section that uses the nginx proxy and is the overly complex config that we don't want to recommend.

howardjohn: Good call, added it back

content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md
```diff
@@ -254,383 +242,6 @@ $ kubectl delete virtualservice direct-wikipedia-through-egress-gateway
...
-    $ kubectl delete destinationrule disable-mtls-for-sni-proxy
-    $ kubectl delete configmap egress-sni-proxy-configmap -n istio-system
-    $ kubectl delete -f ./egressgateway-with-sni-proxy.yaml
-    {{< /text >}}
-
-1.  Remove the configuration files you created:
-
-    {{< text bash >}}
-    $ rm ./sni-proxy.conf ./egressgateway-with-sni-proxy.yaml
-    {{< /text >}}
```
frankbu, line 245: This blank line is causing lint error.

```suggestion
```

craigbox (approved):
/cherrypick release-1.14

joshperry:
I have no qualms with the "shouldn't be in our main tasks" change, but it seems kind of wrong to just delete this information from existence... I'm at this PR because I remember reading this method when we first deployed istio about a year back and had to look at the docs repo commit history to figure out why it disappeared.

Does it make sense to bring this back somewhere, or to maybe get some heads together to modernize the methodology? Not being able to handle wildcard TLS origination makes using these gateways for eastern egress really quite painful.

Often a complex solution that works is a good starting point to improve; we're already building a custom proxy with two custom C++ filters to shift TLS origination for NATS and postgres into the sidecar, so this seems fairly straightforward from my perspective.

psandhu79:
So there is no alternative apart from putting in all host entries

PR #11313, opened 2022-05-17
update proxy config yaml

xichengliudui:
The new field is `ImageType`. `type` is no longer used

- [x] Configuration Infrastructure

### DIFF
--- a/content/en/docs/reference/config/networking/proxy-config/index.html
+++ b/content/en/docs/reference/config/networking/proxy-config/index.html
@@ -33,7 +33,7 @@ metadata:
 spec:
   concurrency: 0
   image:
-    type: distroless
+    imageType: distroless
 </code></pre>
 
 <p>For namespace level configuration, put the resource in the desired namespace without a workload selector:</p>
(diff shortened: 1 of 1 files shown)


zirain:
this is not the right way to fix a reference page.
IIRC, you'd better send a PR in api repo, update the comments, then the automator will sync those changes here.

zirain:
hold for https://github.com/istio/istio/issues/38968

ericvn (changes requested):
This file is copied from the istio/api repo and needs to be changed there.

content/en/docs/reference/config/networking/proxy-config/index.html
```diff
@@ -60,7 +60,7 @@
       app: ratings
   concurrency: 0
   image:
-    type: debug
+    imageType: debug
```
ericvn, line 63: Example: This comes from  https://github.com/istio/api/blob/release-1.14/networking/v1beta1/proxy_config.pb.html#L60-L61

xichengliudui: ok

xichengliudui: waiting for confirmation to update API or update documentation

PR #11314, opened 2022-05-17
optimize html style

xichengliudui:
`#` belongs to markdown syntax, markdown syntax does not take effect in HTML

- [x] Docs
image

### DIFF
--- a/content/en/docs/reference/commands/install-cni/index.html
+++ b/content/en/docs/reference/commands/install-cni/index.html
@@ -250,9 +250,9 @@ If it is not installed already, you can install it via your OS&#39;s package man
 <p>To load completions in your current shell session:</p>
 <p>	source &lt;(install-cni completion bash)</p>
 <p>To load completions for every new session, execute once:</p>
-<p>#### Linux:</p>
+<p>Linux:</p>
 <p>	install-cni completion bash &gt; /etc/bash_completion.d/install-cni</p>
-<p>#### macOS:</p>
+<p>macOS:</p>
 <p>	install-cni completion bash &gt; /usr/local/etc/bash_completion.d/install-cni</p>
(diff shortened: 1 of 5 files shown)


ericvn (changes requested):
Note that these files are created by running the commands. If you look at the header, you will see a message to not update theses files directly.  I'm not sure if this change should be in the scripts that post process the file or within the command output itself.

xichengliudui:
> Note that these files are created by running the commands. If you look at the header, you will see a message to not update theses files directly. I'm not sure if this change should be in the scripts that post process the file or within the command output itself.
> 
> See https://github.com/istio/istio.io/blob/master/scripts/grab_reference_docs.sh#L150-L160



> Note that these files are created by running the commands. If you look at the header, you will see a message to not update theses files directly. I'm not sure if this change should be in the scripts that post process the file or within the command output itself.
> 
> See https://github.com/istio/istio.io/blob/master/scripts/grab_reference_docs.sh#L150-L160

The correct fix would be to convert '####' to `<h4></h4>` label

https://github.com/istio/pkg/blob/9f66753e4ffc7894b0cc1f432e6977de085a246f/collateral/control.go#L432-L527
we may need to update this func, `#### Linux` and `#### macOS` is matched as text
```go
	if cmd.Long != "" {
		g.emitText(cmd.Long)
	} else if cmd.Short != "" {
		g.emitText(cmd.Short)
	}
```

xichengliudui:
I currently do not know how to match, do you have any clue?

ericvn:
> I currently do not know how to match, do you have any clue?

I'm not sure if the ### Linux is part of the string, or is it added as part of the EscapeSting(). Maybe there is some snippet oc code to simply replace `####` in the emitText message.

craigbox:
Any update, @xichengliudui?

PR #11322, opened 2022-05-19
fix external-controlplane webhook error

zirain:
fix fail tests cause by https://github.com/istio/istio/pull/38387

PR #11324, opened 2022-05-19
fix external-controlplane webhook error

zirain:

fix https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/11304/doc.test.multicluster_istio.io_release-1.13/1525856549031907328

https://github.com/istio/istio.io/pull/11322 test release-1.13 contains PR https://github.com/istio/istio/pull/38387

cc @frankbu can you take a look?

### DIFF
--- a/content/en/docs/setup/install/external-controlplane/index.md
+++ b/content/en/docs/setup/install/external-controlplane/index.md
@@ -285,6 +285,12 @@ and installing the sidecar injector webhook configuration on the remote cluster 
     $ istioctl manifest generate -f external-istiod.yaml | kubectl apply --context="${CTX_EXTERNAL_CLUSTER}" -f -
     {{< /text >}}
 
+    At last, clean `istiod-default-validator` webhook:
+
+    {{< text bash >}}
+    $ kubectl delete validatingwebhookconfiguration istiod-default-validator --context="${CTX_EXTERNAL_CLUSTER}"
+    {{< /text >}}
+
 1. Confirm that the external istiod has been successfully deployed:
 
     {{< text bash >}}
--- a/content/en/docs/setup/install/external-controlplane/test.sh
+++ b/content/en/docs/setup/install/external-controlplane/test.sh
@@ -104,10 +104,14 @@ snip_set_up_the_control_plane_in_the_external_cluster_2
 snip_get_external_istiod_iop_modified
 echo y | istioctl install -f external-istiod.yaml --context="${CTX_EXTERNAL_CLUSTER}" --set values.pilot.env.ISTIOD_CUSTOM_HOST="${EXTERNAL_ISTIOD_ADDR}"
 
-_verify_like snip_set_up_the_control_plane_in_the_external_cluster_5 "$snip_set_up_the_control_plane_in_the_external_cluster_5_out"
+_verify_like snip_set_up_the_control_plane_in_the_external_cluster_6 "$snip_set_up_the_control_plane_in_the_external_cluster_6_out"
+
+snip_set_up_the_control_plane_in_the_external_cluster_6
+
+snip_set_up_the_control_plane_in_the_external_cluster_5
 
 snip_get_external_istiod_gateway_config_modified
-snip_set_up_the_control_plane_in_the_external_cluster_7
+snip_set_up_the_control_plane_in_the_external_cluster_8
 
 # Validate the installation.
 
@@ -194,6 +198,11 @@ kubectl delete validatingwebhookconfigurations istiod-default-validator --contex
 kubectl delete validatingwebhookconfigurations istiod-default-validator --context="${CTX_EXTERNAL_CLUSTER}"
 kubectl delete mutatingwebhookconfigurations istio-revision-tag-default-external-istiod --context="${CTX_REMOTE_CLUSTER}"
 kubectl delete mutatingwebhookconfigurations istio-revision-tag-default-external-istiod --context="${CTX_EXTERNAL_CLUSTER}"
+
+# clean webhook
+istioctl x uninstall --purge -y --context="${CTX_EXTERNAL_CLUSTER}"
+istioctl x uninstall --purge -y --context="${CTX_REMOTE_CLUSTER}"
+istioctl x uninstall --purge -y --context="${CTX_SECOND_CLUSTER}"
 
 kubectl delete ns istio-system external-istiod --context="${CTX_EXTERNAL_CLUSTER}"
 kubectl delete ns external-istiod --context="${CTX_REMOTE_CLUSTER}"
--- a/content/en/docs/setup/install/external-controlplane/snips.sh
+++ b/content/en/docs/setup/install/external-controlplane/snips.sh
@@ -177,10 +177,14 @@ istioctl manifest generate -f external-istiod.yaml | kubectl apply --context="${
 }
 
 snip_set_up_the_control_plane_in_the_external_cluster_5() {
+kubectl delete validatingwebhookconfiguration istiod-default-validator --context="${CTX_EXTERNAL_CLUSTER}"
+}
+
+snip_set_up_the_control_plane_in_the_external_cluster_6() {
 kubectl get po -n external-istiod --context="${CTX_EXTERNAL_CLUSTER}"
 }
 
-! read -r -d '' snip_set_up_the_control_plane_in_the_external_cluster_5_out <<\ENDSNIP
+! read -r -d '' snip_set_up_the_control_plane_in_the_external_cluster_6_out <<\ENDSNIP
 NAME                      READY   STATUS    RESTARTS   AGE
 istiod-779bd6fdcf-bd6rg   1/1     Running   0          70s
 ENDSNIP
@@ -264,7 +268,7 @@ spec:
 EOF
 }
 
-snip_set_up_the_control_plane_in_the_external_cluster_7() {
+snip_set_up_the_control_plane_in_the_external_cluster_8() {
 kubectl apply -f external-istiod-gw.yaml --context="${CTX_EXTERNAL_CLUSTER}"
 }
 


go.mod
```diff
@@ -11,17 +11,17 @@ replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.5
 require (
 	github.com/golang/sync v0.0.0-20180314180146-1d60e4601c6f
 	github.com/pmezard/go-difflib v1.0.0
-	istio.io/istio v0.0.0-20220518135041-b4c427158e47
-	istio.io/pkg v0.0.0-20220511193026-88c004ec99a0
-	k8s.io/apimachinery v0.23.5
-	k8s.io/client-go v0.23.5
+	istio.io/istio v0.0.0-20220513020623-8013126ce68b
```
zirain, line 14: this will be revert after test verified

ericvn (changes requested):
I'm not sure where this is actually testing the upstream changes. The original breaking change went into istio/istio master and was cherry-picked to 1.13 (but never 1.14). When the 1.13 branch of istio.io was tested (which points to istio/istio 1.13, the regression was noted and the istio/istio 1.13 PR was reverted.

Note that a problem was never uncovered in the istio/istio.io master branch as it is currently testing using istio/istio 1.14 which never had the regression. You should not see a problem here until after 1.14.0 is released at which point a new istio.io 1.14 branch will be created and istio.io master will point back to the istio master branch.

I did a quick test pointing the istio.io master branch to the istio master branch and the regression was noted in the istio master branch as well.

content/en/docs/setup/install/external-controlplane/index.md
```diff
@@ -285,6 +285,12 @@ and installing the sidecar injector webhook configuration on the remote cluster
     $ istioctl manifest generate -f external-istiod.yaml | kubectl apply --context="${CTX_EXTERNAL_CLUSTER}" -f -
     {{< /text >}}
 
+    At last, clean `istiod-default-validator` webhook:
```
ericvn, line 288: Can you better explain what this is doing. In this part of the doc we are setting up the control plane. Then it mentions we are `cleaning` a web hook by deleting it. I suspect there should be some words related to why the web hook should not be there and we should delete it followed by the kubectl delete.

zirain: this doc guide user install two istiod(one in istio-system namesapce, another in external-istiod namespace) in external cluster via `istioctl install`
without https://github.com/istio/istio/pull/38387 `istiod-default-validator` webhook always point to `istiod.istio-system.svc:443` even you change `global.istioNamespace`.

frankbu:
This PR seems to be a messy band-aid solution to the problems that we already have temporary fixes for.

I don't think this is needed at all after #11041 and #11281 have been merged.

frankbu:
This PR seems to be a messy band-aid solution to the problems that we already have temporary fixes for.

I don't think this is needed at all after #11041 and #11281 have been merged.

zirain:
> This PR seems to be a messy band-aid solution to the problems that we already have temporary fixes for.
> 
> I don't think this is needed at all after #11041 and #11281 have been merged.

yes, let's cherry-pick them to release-1.13 and retest https://github.com/istio/istio/pull/38387?

content/en/docs/setup/install/external-controlplane/index.md
ericvn: Explain in the document text, what is happening.

content/en/docs/setup/install/external-controlplane/test.sh
```diff
@@ -104,10 +104,14 @@ snip_set_up_the_control_plane_in_the_external_cluster_2
 snip_get_external_istiod_iop_modified
 echo y | istioctl install -f external-istiod.yaml --context="${CTX_EXTERNAL_CLUSTER}" --set values.pilot.env.ISTIOD_CUSTOM_HOST="${EXTERNAL_ISTIOD_ADDR}"
 
-_verify_like snip_set_up_the_control_plane_in_the_external_cluster_5 "$snip_set_up_the_control_plane_in_the_external_cluster_5_out"
+_verify_like snip_set_up_the_control_plane_in_the_external_cluster_6 "$snip_set_up_the_control_plane_in_the_external_cluster_6_out"
+
+snip_set_up_the_control_plane_in_the_external_cluster_6
```
ericvn, line 109: snip_set_up_the_control_plane_in_the_external_cluster_6 is getting run twice.

content/en/docs/setup/install/external-controlplane/test.sh
```diff
@@ -104,10 +104,14 @@ snip_set_up_the_control_plane_in_the_external_cluster_2
...
+
+snip_set_up_the_control_plane_in_the_external_cluster_5
```
ericvn, line 111: 5 is in the doc before 6, should this line be moved up so things are in order. Here 5 is run after 6.

frankbu:
Closing this for now, since it should not be needed.

zirain:
@frankbu can you take a look https://github.com/istio/istio.io/pull/11324#discussion_r877148667 , we still need this PR or something else.

https://github.com/istio/istio.io/blob/db2b88790e61828f17a806d6c2485b56d6e56b49/content/en/docs/setup/install/external-controlplane/snips.sh#L50

frankbu:
> @frankbu can you take a look https://github.com/istio/istio.io/pull/11324#discussion_r877148667 , we still need this PR or something else.

Sorry, forgot to mention one more fix https://github.com/istio/istio/pull/38600 for that too.

All together there are 3 PRs that fix (or workaround) the problems with validating webhooks:

1. #11041
2. #11281
3. https://github.com/istio/istio/pull/38600

frankbu:
@zirain unless I'm missing something, there is no test failure in the master branch anymore, with or without this PR. Is that not true? If it is, we don't need this PR.

zirain:
> @zirain unless I'm missing something, there is no test failure in the master branch anymore, with or without this PR. Is that not true? If it is, we don't need this PR.

right now the istio.io run test on release-1.14 not the master branch.

I sent https://github.com/istio/istio/pull/38387 to master, and cherry-pick to release-1.13 which cause https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/11304/doc.test.multicluster_istio.io_release-1.13/1525856549031907328

I also open a new PR https://github.com/istio/istio.io/pull/11329, test release-1.13 with https://github.com/istio/istio/pull/38387.

there're two options:
1. can you confirm https://github.com/istio/istio/pull/38387 is a good fix? if not, let's revert it from master branch
2. let's find out the root cause, my current fix may not perfect.

ericvn:
@frankbu If we were to test istio.io against the istio master branch, there is a failure (at least when I tried it a few days ago). There is a PR in istio master that causes the failure. 

When that PR was cherry-picked to the istio 1.13 branch, the 1.13 istio.io tests failed once it was moved to use a build with that commit. The cherry-pick was reverted istio 1.13 so the istio.io 1.13 tests would pass.

Had that same PR been cherry-picked to release 1.14 (no idea why that didn't happen), then we would see the same failure here (since we are currently testing against 1.14 builds).

Once 1.14.0 releases, istio.io will move back to testing against the istio master branch, and we will see the test fail.

Question is if the change made in istio master is correct? If it is, should it be cherry-picked back to 1.4 and 1.13?  If it is correct, does the documentation and its test need to be updated? Will this cause a failure for the current users who may have based behavior on incorrect behavior? Once the 1.14 PR is merged, we will need to update the doc test for it, and also cherrypick that back to 1.13. The test won't need cherry-picking to 1.14 as we will get that at branch cut time.

frankbu:
@zirain 

> I also open a new PR https://github.com/istio/istio.io/pull/11329, test release-1.13 with https://github.com/istio/istio/pull/38387.I also open a new PR https://github.com/istio/istio.io/pull/11329, test release-1.13 with https://github.com/istio/istio/pull/38387.

If you look at the [output from the failed test](https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/11329/doc.test.multicluster_istio.io_release-1.13/1527451912977256448/artifacts/tests-setup-multicluster-9b450f/TestDocs/setup/install/external-controlplane/test.sh/test.sh/_test_context/test.sh_output.txt), it is using `istioctl install` (not `istioctl manfiest generate`) to install the external control plane:

```
This will install the Istio 1.13.0 empty profile into the cluster. Proceed? (y/N) 
- Processing resources for Istiod.
- Processing resources for Istiod. Waiting for Deployment/external-istiod/istiod
✔ Istiod installed
- Pruning removed resources
✔ Installation completeMaking this installation the default for injection and validation.
```

That's what is causing it to create the validation webhook that (with your change) breaks the other istiod. Before your fix, it created the webhook incorrectly pointing to the other istiod, so even though we don't want it to generate any webhook, it previously didn't have any effect because the incorrectly generated webhook didn't change anything.

The proper workaround for this is to use `istioctl manifest generate` for the install, which doesn't generate the webhooks. The change is in #11041, which seems like it has been cherry-picked, but from the output message, above, I can see that it is not in the test that you ran.

zirain:
> @zirain
> 
> > I also open a new PR #11329, test release-1.13 with [istio/istio#38387](https://github.com/istio/istio/pull/38387) also open a new PR #11329, test release-1.13 with [istio/istio#38387](https://github.com/istio/istio/pull/38387).
> 
> If you look at the [output from the failed test](https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/11329/doc.test.multicluster_istio.io_release-1.13/1527451912977256448/artifacts/tests-setup-multicluster-9b450f/TestDocs/setup/install/external-controlplane/test.sh/test.sh/_test_context/test.sh_output.txt), it is using `istioctl install` (not `istioctl manfiest generate`) to install the external control plane:
> 
> ```
> This will install the Istio 1.13.0 empty profile into the cluster. Proceed? (y/N) 
> - Processing resources for Istiod.
> - Processing resources for Istiod. Waiting for Deployment/external-istiod/istiod
> ✔ Istiod installed
> - Pruning removed resources
> ✔ Installation completeMaking this installation the default for injection and validation.
> ```
> 
> That's what is causing it to create the validation webhook that (with your change) breaks the other istiod's validation. Before your fix, it created the webhook incorrectly pointing to the other istiod, so even though we don't want it to generate any webhook, it previously didn't have any effect because the incorrectly generated webhook didn't change anything.
> 
> The proper workaround for this is to use `istioctl manifest generate` for the install, which doesn't generate the webhooks. The change is in #11041, which seems like it has been cherry-picked, but from the output message, above, I can see that it is not in the test that you ran.

there're logs like following:
```console
Thank you for installing Istio 1.13.  Please take a few minutes to tell us about your install/upgrade experience!  https://forms.gle/pzWZpAvMVBecaQ9h9
Error from server (InternalError): error when creating "external-istiod-gw.yaml": Internal error occurred: failed calling webhook "validation.istio.io": Post "https://istiod.external-istiod.svc:443/validate?timeout=10s": x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "cluster.local")
Error from server (InternalError): error when creating "external-istiod-gw.yaml": Internal error occurred: failed calling webhook "validation.istio.io": Post "https://istiod.external-istiod.svc:443/validate?timeout=10s": x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "cluster.local")
```

[this line](https://github.com/istio/istio.io/blob/db2b88790e61828f17a806d6c2485b56d6e56b49/content/en/docs/setup/install/external-controlplane/snips.sh#L50) still create `istio-default-validator`, where there're two istiod pods in same cluster and same revision(here both are `default`), there will be a race condition.

in this case, the endpoint is `istiod.external-istiod.svc:443`, but the caBundle from istiod in istio-system.

zirain:
https://github.com/istio/istio.io/blob/2b7afa36d78ccbcda610d5adf739ca38e210a55e/content/en/docs/setup/install/external-controlplane/test.sh#L105

maybe this's the root cause, the `test.sh` still use `istioctl install`

frankbu:
> @frankbu maybe this's the root cause, the test.sh still use istioctl install

Yes, good catch. That would explain it.

zirain:
replace by https://github.com/istio/istio.io/pull/11349

PR #11369, opened 2022-06-02
Remove doc on "Istio DNS Certificate Management"

howardjohn:
This document gives harmful advice. This feature was intended to be used
for signing control plane certificates, and actually doesn't work for
other cases (cross namespace or any modern Kubernetes version are
completely broken).


craigbox:
Does this make https://istio.io/latest/blog/2019/dns-cert/ surplus to requirements?

howardjohn:
> Does this make https://istio.io/latest/blog/2019/dns-cert/ surplus to requirements?

Do we remove old blogs?

frankbu:
> Do we remove old blogs?

No, we've never removed blogs. Updating broken links to the archive version, as you've done, is the correct fix.

PR #11383, opened 2022-06-07
enforcing policies on egress traffic

nauticalmike:
Please provide a description for what this PR is for: enforcing policies on egress traffic

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


frankbu:
Why is there both a blog post and task with the same material? Should be one or the other. If a task, it probably should be in the security section, not policy enforcement section.

nauticalmike:
@frankbu the material is also related to policy enforcement which seems more appropriate in my opinion than just security, but where do we want to put it? is there a process to decide? Based on the description I found [here](https://istio.io/latest/docs/releases/contribute/add-content/) seemed appropriate for both a blog post and a task. But if it can't be in both sections I can remove the one the community think is less appropriate based on the content. Let me know

frankbu:
Since it's mostly about configuring authorization policy I think this would be a good place: https://istio.io/latest/docs/tasks/security/authorization/. Notice that there is already an ingress gateway task there.

Alternatively, it seems quite related to the tasks in this section as well: https://istio.io/latest/docs/tasks/traffic-management/egress/, so it might be even better.

Also note that if you make it a task, you need to add a test for the doc. https://github.com/istio/istio.io/blob/master/README.md#testing-document-content

If you make it a blog post, instead of a task, then a test is not required.

nauticalmike:
@frankbu I just deleted the task docs while I figure out how to do a more proper task doc and tests now that I get better the purpose, can we get an ok-to-test now for this as a blog post?

content/en/blog/2022/egress-policies/index.md
```diff
@@ -0,0 +1,1105 @@
...
+  - name: google
+    trafficPolicy:
+      portLevelSettings:
+      - port:
+          number: 80
+        tls:
+          mode: ISTIO_MUTUAL
+          sni: developers.google.com
+---
+apiVersion: networking.istio.io/v1alpha3
```
howardjohn, line 397: I know our existing examples do this, but IMO we should have 2 VSs. It is way more understandable and allows putting one in app namespace and one in egress ns (if desired).

I would be appreciative if all our examples were updated to this split model as well, although certainly not a block for this PR

content/en/blog/2022/egress-policies/index.md
```diff
@@ -0,0 +1,1105 @@
...
+expect-ct: max-age=31536000, report-uri="http://csp.yahoo.com/beacon/csp?src=yahoocom-expect-ct-report-only"
+x-xss-protection: 1; mode=block
+x-content-type-options: nosniff
+{{< /text >}}
+
+If you want you can test the other other address on the other `sleep` pod. We can confirm the pods have outbound access to Google and Yahoo.
+
+![Egress traffic](proxy-external.png)
+
+## Block outbound access
```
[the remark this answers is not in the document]
nauticalmike: ack

content/en/blog/2022/egress-policies/index.md
howardjohn: This is fine for now. But just to be clear - just mean splitting the 1 VS into 2 VS. They can even be in the same namespace if we really want

nauticalmike: was there another step to be done here?

nauticalmike: @howardjohn ^

craigbox (commented):
I reopened this, did I? :)

@nauticalmike would you still like this published?  Can we work on getting it live?

nauticalmike:
@craigbox sure, this was written before there was even any egress topics at all on our docs. Let me know if this still applies as a blog post or if we want to do something else.

ericvn:
Note that any changes in the release-1.14 branch will not show up in the docs. The cherry-pick to master will get it into preliminary.istio.io

craigbox:
If you start as a blog post we can figure out if it would be a good part of the docs.

PR #11485, opened 2022-06-24
fix 1.13 proxy-config example

yokoyang:
The example in [proxy-config](https://istio.io/latest/docs/reference/config/networking/proxy-config/) is invalid. Cause configuration error：

```
error validating data: [ValidationError(ProxyConfig.spec.image): unknown field "type" in io.istio.networking.v1beta1.ProxyConfig.spec.image, ValidationError(ProxyConfig.spec): unknown field "workloadSelector" in io.istio.networking.v1beta1.ProxyConfig.spec]; if you choose to ignore these errors, turn validation off with --validate=false

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/archive/v1.13/docs/reference/config/networking/proxy-config/index.html
+++ b/archive/v1.13/docs/reference/config/networking/proxy-config/index.html
@@ -42,11 +42,11 @@ metadata:
   namespace: example
 spec:
   selector:
-    labels:
+    matchLabels:
       app: ratings
   concurrency: 0
   image:
-    type: debug
+    imageType: debug
 </code></pre><p>If a <code>ProxyConfig</code> CR is defined that matches a workload it will merge with its <code>proxy.istio.io/config</code> annotation if present,
 with the CR taking precedence over the annotation for overlapping fields. Similarly, if a mesh wide <code>ProxyConfig</code> CR is defined and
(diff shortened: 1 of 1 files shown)


Chen-Xintong:
@yokoyang Please sign CLA

yokoyang:
> @yokoyang Please sign CLA

thanks, CLA has added

ericvn (changes requested):
Updating archive docs is not standard practice. They may get overriden if we ever archived the release again (which can be done) as we will simply replace this with what is in the 1.13 branch. If this change really needs to be made in 1.13 only, and is a change we should get published, the change should be made in the release-13 branch and we should reactive it to the main and 1.14 branches.

That said, should this change also be made in the 1.4 and main branches?

Also, just updating in the archive does not test this change at all against the 1.13 branch and releases.

yokoyang:
> Updating archive docs is not standard practice. They may get overriden if we ever archived the release again (which can be done) as we will simply replace this with what is in the 1.13 branch. If this change really needs to be made in 1.13 only, and is a change we should get published, the change should be made in the release-13 branch and we should reactive it to the main and 1.14 branches.
> 
> That said, should this change also be made in the 1.4 and main branches?
> 
> Also, just updating in the archive does not test this change at all against the 1.13 branch and releases.

content in  content/en/docs/reference/config/networking/proxy-config/index.html is correct. no need to change.

ericvn:
> content in content/en/docs/reference/config/networking/proxy-config/index.html is correct. no need to change.

PR #11510, opened 2022-06-27
[release-1.12] Update reference docs for upcoming 1.12 build

ericvn:

Prep run at updating the ref docs for last 1.12 build. Not sure we need ref doc updates prior to 1.14 as we won't really be rebuilding the archive to include it. Should the automation update it in multiple places?


craigbox (approved):

craigbox:
I think we should seek staffing for a larger reform of the docs structure and automation, and address reference docs at that point too?

craigbox (approved):

Issue #11516, opened 2022-06-29
Table width error


craigbox:
https://istio.io/latest/docs/releases/feature-stages/#feature-phase-definitions

<img width="922" alt="Screen Shot 2022-06-29 at 4 25 10 PM" src="https://user-images.githubusercontent.com/132510/176475030-9ddb48c0-903a-4032-8e5a-44d54d445bb5.png">

craigbox:
Ihor, I wonder if you had time to look at this?

Arhell:
> Ihor, I wonder if you had time to look at this?

This happened after edits on links that went beyond the block
https://github.com/istio/istio.io/pull/10501
But I'll see what I can do

PR #11522, opened 2022-06-29
fix table width error

Arhell:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [x] User Experience

resove https://github.com/istio/istio.io/issues/11516

### DIFF
--- a/src/sass/base/_base.scss
+++ b/src/sass/base/_base.scss
@@ -178,10 +178,10 @@ tr th:last-child {
 
 tr td:last-child {
     border-right: 1px solid $boxBorderColor;
-    word-break: break-word;
 }
 
-tr td:last-child code {
+tr td:last-child code,
+tr td:last-child a {
     word-break: break-word;
 }
 
--- a/src/sass/base/_base.scss
+++ b/src/sass/base/_base.scss
@@ -91,7 +91,8 @@ table {
     border-collapse: separate;
     border-spacing: 0;
     width: 100%;
-    display: table;
+    display: block;
+    overflow-x: auto;
     @media (max-width: $bp-sm) {
         display: block;
         overflow-x: auto;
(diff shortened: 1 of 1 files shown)


craigbox:
✅ https://deploy-preview-11522--preliminary-istio.netlify.app/latest/docs/releases/feature-stages/#feature-phase-definitions 
❌ https://deploy-preview-11522--preliminary-istio.netlify.app/latest/docs/reference/commands/pilot-agent/#envvars

Arhell:
@craigbox can make minimum width for pc?

craigbox:
I've had a quick play around with various options for `table-layout: fixed` and `word-break: break-word`, none of them seem to fix this. 

We could change Hugo to add [<wbr>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) tags to code blocks where underscores are?  Thoughts, @ericvn ?

Arhell:
> I've had a quick play around with various options for `table-layout: fixed` and `word-break: break-word`, none of them seem to fix this.
> 
> We could change Hugo to add tags to code blocks where underscores are? Thoughts, @ericvn ?

the problem is that there are just long words, not just blocks of code

craigbox:
we could make you scroll inside the table?

It looks like the layout engine seems to prefer using the first row as a guideline on how wide the columns should be, which doesn't make any sense.

I don't think we want to make users hint at which columns should be what width when they create tables either.

Arhell:
here is an example with scrolling
Screenshot_1
the same but without scrolling but there is a strip that scroll can be
Screenshot_2

Arhell:
@craigbox
Please take a look, it's not bad.
But some tables are not full width, I think this is not critical

craigbox:
Hey Ihor, I wasn't sure quite what to do with this one (I've had it open for a long time!)
I am not sure it was better to merge or not, but happy to take your feedback if you have any more.

Arhell:
> Hey Ihor, I wasn't sure quite what to do with this one (I've had it open for a long time!) I am not sure it was better to merge or not, but happy to take your feedback if you have any more.

Hello, so-so solution, when I made edits so that the content would not get out, I wrote like there might be small problems because there is no way to take into account everything. But in any case, what is done is better than it was when the content climbed off the screen

Unfortunately there is no perfect solution it all depends on the content

PR #11534, opened 2022-07-01
Test moving up a Hugo version and replace BlackFriday with GoldMark.

ericvn:

In the past there were issues with URLs in GoldMark and we kept reverting back to BlackFriday. BlackFriday has finally been removed from the latest Hugo versions, so let's see what's broken.


ericvn:
https://deploy-preview-11534--preliminary-istio.netlify.app/latest/about/ecosystem is one back with issues.

craigbox:
https://github.com/gohugoio/hugo/issues/6553

We have to fix some templates, probably by removing  `| markdownify` from shortcode blocks.

ericvn:
@craigbox I removed all amqrkdonwify stanzas and noted the about/ecosystem seemed to be good. However about/deployment had issues with links. I did not that the links had ../../*docs in them (which we don't have anywhere else (we say to use .docs) so I tried that change. That didn't help. I did add back the markdownify for centered blocks (only used in the /about tree) and it seemed to fix it.

Let's see how this looks..

ericvn:
https://deploy-preview-11534--preliminary-istio.netlify.app/latest/docs/ops/diagnostic-tools/istioctl/ has an issue the tab section...

ericvn:
So adding it back to tab-system fixes the https://deploy-preview-11534--preliminary-istio.netlify.app/latest/docs/ops/diagnostic-tools/istioctl/ and breaks the eco-system again.  Argh..

ericvn:
@craigbox Thoughts on these changes. Right now the output seems good for the few areas I had looked at it the past and seen issues.

I might just put the markdownify back in the places it was removed) or can likely remove those layouts entirely since I don't think they are used (haven't actually tried removing them).

ericvn:
Tried removing them and only layouts/partials/section_description.html was able to be removed.

ericvn:
Removing the DNM.

frankbu (approved):

PR #11552, opened 2022-07-08
Fix link to the documentation of SPIFFE trust domain federation

jewertow:
Signed-off-by: Jacek Ewertowski <jacek.ewertowski1@gmail.com>

This PR fixes link redirecting to a non-existent Google doc.

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn:
A couple of comments:

Updating the contents in the /archives folder may change them temporarily, but they could be overridden if a release gets re-archived (files re-copied from the branch back into the archive directory). The changes should really be made in the various release branches and then each branch re-archived to make the updates and prevent the potential for overriding the changes. Granted this change is across a lot of branches and this would really require a bunch of manual work and some of the older releases are very unlikely to get re-archived for a change, so doing them here may make sense. One just needs to be aware that things could get changed back later.

Combining changes for both the English docs and the translated docs is not usually done. There are different approvers for each language so combining them means the PR needs approval from both groups. However, In the case of some non-translated changes like links in this case, we do have some approvers that can cover both.

Given, the above, let's see if anyone else has any other comments, else I can merge this next week.

ericvn (approved):
No further comments so will merge. Note per my earlier comments, the archives changes may get reverted.

PR #11679, opened 2022-08-03
Fix lint errors with new build tools

ericvn:

Fix lint and gen errors

### DIFF
--- a/pkg/test/istioio/framework.go
+++ b/pkg/test/istioio/framework.go
@@ -1,8 +1,8 @@
+// Copyright 2020 Istio Authors
+//
 // Package istioio includes the framework for doc testing. This package will
 // first scan through all the docs to collect their information, and then
 // setup istio as appropriate to run each test.
-//
-// # Copyright 2020 Istio Authors
 //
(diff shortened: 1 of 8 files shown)


pkg/test/istioio/framework.go
```diff
@@ -2,13 +2,13 @@
 // first scan through all the docs to collect their information, and then
 // setup istio as appropriate to run each test.
 //
-// Copyright 2020 Istio Authors
+// # Copyright 2020 Istio Authors
```
frankbu, line 5: Is this right? Why the `#`?

ericvn: The doc format thinks it's a heading: The # syntax was added in Go 1.19. Before Go 1.19, headings were identified implicitly by single-line paragraphs satisfying certain conditions, most notably the lack of any terminating punctuation.

ericvn: Added a commit where I moved the copyright similar to other files where it is not surrounded by blank lines (now at the top).

As noted, we have gone to removing copyrights in files, so we could do a bigger cleanse of those later.

frankbu (approved):

PR #11680, opened 2022-08-04
clarify unmentioned constraint port.name equalness

tedli:

In our multi cluster mesh, we found that, even we meet the `namespace-sameness`, cross cluster calls to remote cluster through eastwest gateway never happen in some cases. (calls always invoke to primary cluster)
After some digging, we found that the `same` kubernetes `Service` had different `port.name`, and the code that aggregating eds reads the port.name to match endpoint with service.
And if we correct the `Service.spec.ports[_].name` to be same as what in primary cluster (literally equal, both name no specified consider equal "" == "" ), cross cluster calls act as expect.
So as `namespace-sameness`, the addition constraint should be documented.

https://github.com/istio/istio/blob/93e4ec424b52da807228bd7784c59d646ec73733/pilot/pkg/xds/endpoint_builder.go#L240-L242

```golang
for _, ep := range endpoints {
			// TODO(nmittler): Consider merging discoverability policy with cluster-local
			if !ep.IsDiscoverableFromProxy(b.proxy) {
				continue
			}
			if svcPort.Name != ep.ServicePortName {
				continue
			}
```

- [x] Docs
- [x] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/reference/glossary/namespace-sameness.md
```diff
@@ -5,5 +5,5 @@ test: n/a
 
 Within a multicluster mesh, [namespace sameness](https://github.com/kubernetes/community/blob/master/sig-multicluster/namespace-sameness-position-statement.md)
 applies and all namespaces with a given name are considered to be the same namespace. If multiple clusters contain a
-`Service` with the same namespaced name, they will be recognized as a single combined service. By default, traffic is
```
craigbox, line 8: Do you know what other pages you looked at and didn't find this information?

This is the glossary definition for namespace sameness.  What about "same namespaced name and port name"?

tedli: I mean `Service.spec.ports[_].name`, as pined a link https://github.com/istio/istio/blob/93e4ec424b52da807228bd7784c59d646ec73733/pilot/pkg/xds/endpoint_builder.go#L240-L242

if the multicluster `Service` have same port, but each with a different name, they are no longer `same`, only the service in primary cluster affect.

```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  selector:
    app.kubernetes.io/name: proxy
  ports:
  - name: name-of-service-port       # <-- I mean this, service.spec.ports[_].name
    protocol: TCP
    port: 80
    targetPort: http-web-svc
```

craigbox: I understand the situation, but I'm just trying to clarify: "namespace sameness" is a thing that exists in Kubernetes, and what Istio is doing is saying "service and namespace sameness" is required.  Let me propose an edit for this.

content/en/docs/reference/glossary/namespace-sameness.md
```diff
@@ -5,5 +5,5 @@ test: n/a
...
-load-balanced across all clusters in the mesh for a given service.
\ No newline at end of file
+`Service` with the same namespaced name, in addition same port within the same service should make sure `port.name` literally equal, then they will be recognized as a single combined service. By default, traffic is
```
craigbox, line 8:
```suggestion
`Service` with the same namespaced name and the same port name, then they will be recognized as a single combined service. By default, traffic is
```

craigbox (changes requested):

craigbox:
WG Network, PTAL

content/en/docs/reference/glossary/namespace-sameness.md
tedli: Thanks. My words didn't express well. It took me quit a long time to debug my cluster environment, so I though this PR can help others too. :smiling_face_with_three_hearts:

content/en/docs/reference/glossary/namespace-sameness.md
```diff
@@ -5,5 +5,5 @@ test: n/a
 
 Within a multicluster mesh, [namespace sameness](https://github.com/kubernetes/community/blob/master/sig-multicluster/namespace-sameness-position-statement.md)
 applies and all namespaces with a given name are considered to be the same namespace. If multiple clusters contain a
-`Service` with the same namespaced name, they will be recognized as a single combined service. By default, traffic is
-load-balanced across all clusters in the mesh for a given service.
\ No newline at end of file
+`Service` with the same namespaced name and the same port name, then they will be recognized as a single combined service. By default, traffic is
```
howardjohn, line 8: This seems a bit misleading. It seems to be saying that "service with same port name", which suggests a service has 1 port - it can have many. Additionally, ports not matching doesn't mean the whole _Service_ is not considered 'same'; rather, just the port is not considered.

tedli: Thanks for reply.
Sorry, my words did misleading. Let me give a example:

```yaml
# the one in primary cluster
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  selector:
    app.kubernetes.io/name: proxy
  ports:
  - name: primary-cluster-80
    protocol: TCP
    port: 80
    targetPort: http-web-svc

# the one in a remote cluster
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  selector:
    app.kubernetes.io/name: proxy
  ports:
  - name: remote-cluster-80
    protocol: TCP
    port: 80
    targetPort: http-web-svc
  - name: remote-cluster-irrelevant-port
    protocol: TCP
    port: 8080
    targetPort: irrelevant-port
```

These two service, in same namespace, had same name, both had 80 port, the one in remote cluster, also had an extra port, that's OK, service can have many ports.

But, for the 80, these two service had different `port.name`, one is `primary-cluster-80`, the other is `remote-cluster-80`.
If so, the eds of cds `outbound|80||nginx-service.default.svc.cluster.local` only contains endpoints in primary cluster, include the eastwest gateway. Pods in remote make request to this service, will always goes to workloads in primary cluster.

I will try to choose other words and ask for some help around, to express this more clear.

howardjohn: Yeah I argee, the poit I was trying to make is the granularity is Port, not Service -- if one of the port names matched they would still be counted

ericvn:
This PR is targeted for the release-1.14 branch which will shortly be archived and likely not archived again. Thus this PR will not show up in the 1.14 version of the documentation. This PR should be rebased on the main branch (and cherry-picked to 1.15 if needed).

craigbox:
@tedli could you please rebase this on master or release-1.15?

tedli:
> @tedli could you please rebase this on master or release-1.15?

Hi craigbox,

Thanks for reply. I cherry-picked this to master and release-1.15. Please have a look.

https://github.com/istio/istio.io/pull/11892
https://github.com/istio/istio.io/pull/11891

content/en/docs/reference/glossary/namespace-sameness.md
```diff
@@ -6,4 +6,6 @@ test: n/a
 Within a multicluster mesh, [namespace sameness](https://github.com/kubernetes/community/blob/master/sig-multicluster/namespace-sameness-position-statement.md)
 applies and all namespaces with a given name are considered to be the same namespace. If multiple clusters contain a
 `Service` with the same namespaced name, they will be recognized as a single combined service. By default, traffic is
-load-balanced across all clusters in the mesh for a given service.
\ No newline at end of file
+load-balanced across all clusters in the mesh for a given service.
+
+For one or many ports within the combined service, same port value should also have same port name.
```
craigbox, line 11:
```suggestion
Ports that match on _number_ must also have the same port _name_ to be considered as a combined service port.```

craigbox (changes requested):

craigbox (approved):

PR #11705, opened 2022-08-15
Test to verify if PR breaks test in 1.14

ericvn:
Test if PR broke tests as in the master branch.


ericvn:
Looks like fix worked, at least for 1.14

PR #11715, opened 2022-08-18
Update test ref to 1.15.0-beta.1

ericvn:


craigbox (approved):

ericvn:
Does kubectl sort on the pod name in each portion of the name (separated by -'s) or the whole name? Trying to figure out why the order was changed in a set of tests. will canary show up before istiod if the id (which is hex) starts with a d, e or f?

istiod-5a...
Istiod-canary...
isitod-de3...

ericvn:
Looking at one test:
```
received:
"NAME                            READY   STATUS    RESTARTS   AGE     REV
istiod-canary-d95b477b7-rs52r   1/1     Running   0          2m4s    canary
istiod-df8b6454d-vxwcv          1/1     Running   0          2m12s   default"
```
```
received:
"NAME                            READY   STATUS    RESTARTS   AGE     REV
istiod-canary-d95b477b7-6kzrg   1/1     Running   0          2m3s    canary
istiod-df8b6454d-n87lg          1/1     Running   0          2m12s   default"
```
```
"NAME                            READY   STATUS    RESTARTS   AGE     REV
istiod-canary-d95b477b7-kcgmp   1/1     Running   0          2m5s    canary
istiod-df8b6454d-tjzpk          1/1     Running   0          2m15s   default"
```

This means that our test should be updated since the two pods may be in a different order. Or need to verify the verify util code.

ericvn:
Note that the failure isn't on the order of the pod names as the verify only looks at the prefix, but the order of the revision value changes with the pod name order and that is failing.

PR #11718, opened 2022-08-19
The vendor directory should be in .gitignore

litong01:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@ generated
 # install generated files
 install/**
 
+# vendor dependency files
+vendor
+
 # contains samples to demo Istio
 samples/**
 


ericvn:
We use modules and don't use the vendor directory so it shouldn't need to be in the .gitignore. Why add it for no reason. Can you give me an example of what would create the vendor directory in istio.io?

litong01:
> We use modules and don't use the vendor directory so it shouldn't need to be in the .gitignore. Why add it for no reason. Can you give me an example of what would create the vendor directory in istio.io?

@ericvn there is this go.mod file, when import to VS code, it will be treated as go project to find dependencies, after running `go mod vendor` which produces vendor directory, then VS code will be happy. Without the vendor directory, VS code keeps searching, which makes vs code sluggish. I think adding this will help in that regard. and it wont hurt anything but help a lot in terms of IDE and git operations.

ericvn:
@litong01 I hadn't heard of vscode having that issue before and hadn't seen it myself.

I suspect you really want this to be `/vendor/` so you are only ignoring the vendor directory in the root folder?

litong01:
> @litong01 I hadn't heard of vscode having that issue before and hadn't seen it myself.
> 
> I suspect you really want this to be `/vendor/` so you are only ignoring the vendor directory in the root directory of the repo?

That is right. that should be same thing as vendor like how istio is doing. if it also refers the subdirectory, it should have been vendor/**/*

ericvn:
The way it is now, it will ignore any file or directory called vendor anywhere in the source tree.
ex:
```
a/vendor
a/vendor/
vendor
vendor/
```

I'm not saying istio/istio is correct.

ericvn:
Based on earlier discussion, can we close this one now?

litong01:
do not need it after discussion.

PR #11734, opened 2022-08-23
improve install gateway doc

zirain:

improve the doc, which may confuse user

fix: https://github.com/istio/istio/issues/38792

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/additional-setup/gateway/index.md
+++ b/content/en/docs/setup/additional-setup/gateway/index.md
@@ -64,7 +64,7 @@ spec:
   profile: empty # Do not install CRDs or the control plane
   components:
     ingressGateways:
-    - name: ingressgateway
+    - name: istio-ingressgateway
       namespace: istio-ingress
       enabled: true
       label:
--- a/content/en/docs/setup/additional-setup/gateway/index.md
+++ b/content/en/docs/setup/additional-setup/gateway/index.md
@@ -92,7 +92,7 @@ Install using standard `helm` commands:
 
 {{< text bash >}}
 $ kubectl create namespace istio-ingress
-$ helm install istio-ingress istio/gateway -n istio-ingress
+$ helm install istio-ingressgateway istio/gateway -n istio-ingress
 {{< /text >}}
 
 To see possible supported configuration values, run `helm show values istio/gateway`.
--- a/content/en/docs/setup/additional-setup/gateway/snips.sh
+++ b/content/en/docs/setup/additional-setup/gateway/snips.sh
@@ -50,7 +50,7 @@ istioctl install -f ingress.yaml
 
 snip_deploying_a_gateway_3() {
 kubectl create namespace istio-ingress
-helm install istio-ingress istio/gateway -n istio-ingress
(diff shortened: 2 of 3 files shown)


content/en/docs/setup/additional-setup/gateway/index.md
```diff
@@ -205,6 +205,10 @@ your `Gateway` selector matches these labels.
 For example, in the above deployments, the `istio=ingressgateway` label is set on the gateway pods.
 To apply a `Gateway` to these deployments, you need to select the same label:
 
+{{< tabset category-name="gateway-selector" >}}
```
frankbu, line 208: I don't think adding these 3 tabs is making this any more clear.

I would suggest to instead change line 205 to say:
```
For example, in the above deployments, the `istio=istio-ingress` label is set on the gateway pods.
```
and change the two other (not helm) deployments to match:
```
     label:
        # Set a unique label for the gateway. This is required to ensure Gateways
        # can select this workload
        istio: istio-ingress
```
This way, you don't need to explain that the helm deployment is different in the example.

zirain: sounds good, updated.

frankbu (approved):

content/en/docs/setup/additional-setup/gateway/index.md
frankbu: IIUC the helm install uses the deployment name for the label by default. If that's right, maybe a better idea is to leave the label as `istio: ingressgateway` and just change the helm command to use ingressgateway as the name. I'm not very familiar with helm, but it seems that the first argument for helm install is the name. If so, then I think that just changing the command to the following should work:
```
$ helm install ingressgateway istio/gateway -n istio-ingress
```
Then the gateway name will be the same in all 3 cases also. I wonder why it was different for helm in the first place.

frankbu: If you disagree, I've already approved, so you can remove the DNM label and merge as is.

zirain: I test with `helm install istio-ingressgateway istio/gateway -n istio-ingress`, which looks better

```
root@ecs-zirain:~# k get po -nistio-ingress
NAME                                   READY   STATUS    RESTARTS   AGE
istio-ingressgateway-685fcc78c-jltfj   1/1     Running   0          20s
root@ecs-zirain:~# k get po -nistio-ingress --show-labels
NAME                                   READY   STATUS    RESTARTS   AGE   LABELS
istio-ingressgateway-685fcc78c-jltfj   1/1     Running   0          28s   app=istio-ingressgateway,istio.io/rev=default,istio=ingressgateway,pod-template-hash=685fcc78c,service.istio.io/canonical-name=istio-ingressgateway,service.istio.io/canonical-revision=latest,sidecar.istio.io/inject=true

```

zirain:
@frankbu I updated the implement, named all the gateway deployment with same name/labels.

if you agree, please remove `DNM`.

PR #11736, opened 2022-08-23
Replace experimental uninstall command with uninstall command

litong01:

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/setup/upgrade/canary/index.md
```diff
@@ -158,13 +158,13 @@ $ istioctl tag set default --revision 1-10-0
...
 {{< text bash >}}
-$ istioctl x uninstall --revision 1-6-5
+$ echo y | istioctl uninstall --revision 1-6-5
 {{< /text >}}
 
 If the old control plane does not have a revision label, uninstall it using its original installation options, for example:
 
 {{< text bash >}}
-$ istioctl x uninstall -f manifests/profiles/default.yaml
+$ echo y | istioctl uninstall -f manifests/profiles/default.yaml
```
frankbu, line 167:
```suggestion
$ istioctl uninstall -f manifests/profiles/default.yaml -y
```

content/en/docs/tasks/security/authorization/authz-td-migration/index.md
```diff
@@ -176,7 +176,7 @@ as the old trust domain without you having to include the aliases.
 $ kubectl delete authorizationpolicy service-httpbin.default.svc.cluster.local
 $ kubectl delete deploy httpbin; kubectl delete service httpbin; kubectl delete serviceaccount httpbin
 $ kubectl delete deploy sleep; kubectl delete service sleep; kubectl delete serviceaccount sleep
-$ istioctl x uninstall --purge
+$ echo y | istioctl uninstall --purge
```
frankbu, line 179:
```suggestion
$ stioctl uninstall --purge -y
```

content/en/docs/setup/additional-setup/gateway/test.sh
```diff
@@ -20,7 +20,7 @@ set -o pipefail
 
 # @setup profile=none
 function rebuild() {
-    istioctl x uninstall --purge --skip-confirmation
+    echo y | istioctl uninstall --purge --skip-confirmation
```
frankbu, line 23:
```suggestion
    istioctl uninstall --purge --skip-confirmation
```

content/en/docs/setup/additional-setup/gateway/test.sh
```diff
@@ -62,6 +62,6 @@ _verify_like snip_canary_upgrade_advanced_2 "${snip_canary_upgrade_advanced_2_ou
 
 # @cleanup
 
-istioctl x uninstall --purge --skip-confirmation
+echo y | istioctl uninstall --purge --skip-confirmation
```
frankbu, line 65:
```suggestion
istioctl uninstall --purge --skip-confirmation
```

content/en/docs/tasks/security/authorization/authz-td-migration/index.md
```diff
@@ -176,7 +176,7 @@ as the old trust domain without you having to include the aliases.
 $ kubectl delete authorizationpolicy service-httpbin.default.svc.cluster.local
 $ kubectl delete deploy httpbin; kubectl delete service httpbin; kubectl delete serviceaccount httpbin
 $ kubectl delete deploy sleep; kubectl delete service sleep; kubectl delete serviceaccount sleep
-$ istioctl x uninstall --purge
+$ stioctl uninstall --purge -y
```
ericvn, line 179: You want istioctl, not stioctl

content/en/docs/tasks/security/authorization/authz-td-migration/snips.sh
```diff
@@ -138,7 +138,7 @@ snip_clean_up_1() {
 kubectl delete authorizationpolicy service-httpbin.default.svc.cluster.local
 kubectl delete deploy httpbin; kubectl delete service httpbin; kubectl delete serviceaccount httpbin
 kubectl delete deploy sleep; kubectl delete service sleep; kubectl delete serviceaccount sleep
-istioctl x uninstall --purge
+stioctl uninstall --purge -y
```
ericvn, line 141: misspelled here as well.

ericvn:
A couple comments.
- We use --skip-confirmation in a lot of the earlier pages, and then slip to using -y in later .md pages. We should be consistent in the .md pages. I don't know if @frankbu has a preference. While I like skip-confirmation, I don't think a lot of users would actually want to type the whole thing so may prefer the -y, but I don't know if we prefer using the more formal in the documentation.
- Does it make sense to simply use the --skip-confirmation/-y parameter instead of having to do the echo y? Not sure if we are testing that one gets prompted and we are entering y? Maybe we should have several of each to test both command paths?

litong01:
> A couple comments.
> 
>     * We use --skip-confirmation in a lot of the earlier pages, and then slip to using -y in later .md pages. We should be consistent in the .md pages. I don't know if @frankbu has a preference. While I like skip-confirmation, I don't think a lot of users would actually want to type the whole thing so may prefer the -y, but I don't know if we prefer using the more formal in the documentation.
> 
>     * Does it make sense to simply use the --skip-confirmation/-y parameter instead of having to do the echo y? Not sure if we are testing that one gets prompted and we are entering y? Maybe we should have several of each to test both command paths?

@ericvn yeah, I noticed that as well. That was the inconsistency in our docs. For document, I would prefer `--skip-confirmation` since it clearer than `-y`, having both in different places may also serve a purpose of testing both options, so I do not mind they appear at different locations.

For the pipe y thing, to be honest with you, I never liked use pipe to send a y to the command, in my view, that is really a hack, I would like to change it all to -y especially in the test scripts which no one sees, but Frank and I had a long discussion about this, where he won the argument, so that is what we had. but I can go with either.

content/en/docs/tasks/traffic-management/egress/wildcard-egress-hosts/index.md
```diff
@@ -253,5 +253,5 @@ $ kubectl delete destinationrule egressgateway-for-wikipedia
 * Uninstall Istio from your cluster:
 
     {{< text bash >}}
-    $ istioctl x uninstall --purge
+    $ echo y | istioctl uninstall --purge
```
frankbu, line 256: This should also use -y
```suggestion
    $ istioctl uninstall --purge -y
```

frankbu: Actually, thinking about it, why do any of the calls (in the test.sh files) use `echo y`, instead of `-y`? The only reason to use echo is when the command is in an md file but you don't want the doc to use -y, but rather let the user do it manually. Only in those cases does it make sense to use `echo y | snip_xxx` so that the test won't hang waiting for input.

frankbu:
> I would like to change it all to -y especially in the test scripts which no one sees, but Frank and I had a long discussion about this, where he won the argument, so that is what we had. but I can go with either.

I think you misunderstood my point. See [here](https://github.com/istio/istio.io/pull/11736#discussion_r953882576) for clarification.

frankbu:
> I don't know if @frankbu has a preference.

I agree that users will want to use the short form `-y`, so using it everywhere would be my preferences.

PR #11762, opened 2022-08-27
make helm install consistent with istioctl

zirain:
make helm install consistent with istioctl

fixes: https://github.com/istio/istio/issues/39049

cc @frankbu


ericvn:
Any feedback from @istio/wg-environments-maintainers ?

howardjohn (changes requested):
This is intended - the use of istio-system is an anti-pattern explicitly not recommended

howardjohn:
If they are confused we should improve documentation, etc. Not proliferate
bad practices

PR #11804, opened 2022-09-05
egress tls origination with credential name - clarify required RBAC

kfaseela:

When credential name is used at sidecar for egress tls origination, explicit RBAC is needed to allow access to the credential.
The doc currently talks about the need for having get and watch permissions, however in practice only list permission is needed.

- [X] Docs
- [ ] Installation
- [X] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn (approved):
Thanks.

PR #11806, opened 2022-09-06
zh:remove controlling-what-the-sidecar-reports.md

windsonsea:
Remove this file by following the en upstream:
```
content/zh/about/faq/metrics-and-logs/controlling-what-the-sidecar-reports.md
```

- [x] Docs


ericvn (approved):
This change does not involve translation as it's just removing a page.

PR #11828, opened 2022-09-07
Add sorting by title for Blogs posting on the same date

ericvn:

Does it make sense to sort on title, description, or do we need another field.

I've added a second commit copying some blogs but changing the date so we can see how this works. This commit will be removed before merging.


ericvn:
Looks like just adding weight: lets one sort. The lower weighted items are lower in the list for a given day.

PR #11876, opened 2022-09-09
Sort and dedupe .spelling

ericvn:

Sort and then deduce the .spelling file as cleanup


ericvn:
`1.9.x.` isn't in the current change so removing DNM label.

craigbox (approved):

PR #11888, opened 2022-09-11
A blog to introduce a light-weighted way to achieve layer 7 load balancing functions without per-pod sidecars.

soya3129:
We would like to introduce a light-weighted way to achieve layer 7 load balancing functions without per-pod sidecars. The work is 

- [X ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/blog/2022/centralized-east-west-gateway/index.md
```diff
@@ -0,0 +1,293 @@
...
+  - internalGateway
+  http:
+  - match:
+    - uri:
+        exact: /hello
+    route:
+    - destination:
+        host: helloworld.default.svc.cluster.local
+```
+Step 5: Execute the following command. You should find the IP address of the HelloWorld service changed to the gateway's IP.
```
howardjohn, line 152: how? Is there some third party code in play here? If so, is it open source? If so, where is the code?

soya3129: Thanks for the attention and comments!

Yes, additional code is needed to accomplish what we proposed here. We introduced a controller named Canal controller. Canal controller watches configuration changes in services and redirects traffic to the internal gateway through core DNS. For data planes, we use router mode Envoys, and envoy’s configuration goes through istiod as usual.

The code is still going through a company procedure before it can be open source. The earliest available date will be 9/23.  I will update this PR then. 

Thanks again!

soya3129: Here is a pointer to the open source code. Thanks for your attention! 

https://github.com/alibaba/alibaba-centralized-mesh-gateway

ericvn:
Looks like the originator has a number of spelling errors that need to be resolved; misspelled words, some need backticks per the guidelines, etc. You can click on the details of the lint test and look at the build log. Not that the build log is blank and white so it might be hard to identify the misspelled word on each line. If you can run `make lint`, requires Docker and make be installed, the output will take a bit but you'll see highlighted words for the misspellings.

louiscryan:
I'm worried about the positioning message of this blog.  In particular its positioning itself vs. the other mesh solutions when it is not really comparable as this solution loses isolation, identity (mTLS) and its related authorization capabilities which are viewed as intrinsic features of a mesh.

I'd much rather see this positioned as "How to use Istio's Gateway support to implement the 'internal L7 LB nexus' strategy" without trying to make it appear to compete with 'mesh' as a solution.



Its also likely that the resources savings relative to ambient may be marginal and so the loss of security properties for a limited benefit would not represent a useful tradeoff for the majority of users. 

A message like 

"If you're not ready to adopt mesh but you want some of Istio's power then you can use it as follows to implement the traditional internal L7 LB pattern. Its important to note that this solution does not provide any of the security or isolation guarantees that mesh solutions provide ..."

might be workable

@howardjohn 
@ZackButcher 
@linsun 

wdyt?

linsun:
+1 @louiscryan I am worried about the messaging of the blog as well. What is proposed in the blog is pretty much expanding Istio ingress gateway from edge to inter-services communication. While it may sound attractive, it is not a full mesh solution, and quite limited as having your service (no sidecar) connected to Istio ingress gateway at the edge today.

@soya3129 I noticed you are using a gateway resource that is different from Istio's or K8s's - is this open source code?  I think users would be more interested in how to do this with a released version of Istio without any proprietary code or config.

content/en/blog/2022/centralized-east-west-gateway/index.md
linsun: Looks like I am not alone asking this q :) Is it possible to write the blog highlighting what you had to change to build it on top of Istio?

yuri-lcy:
At present, the code of this idea has been completed.Based on version 1.15, In the deployment method of istio, a new acmg.profile is created. If you choose acmg deployment, acmg controller will be installed in the cluster (using coredns to hijack traffic). In addition, the lack of capabilities in tracing and tls must be resolved, and we are also considering how to solve these problems.

acmg profile source code:
https://github.com/yuri-lcy/istio
https://github.com/yuri-lcy/api

@soya3129 
@louiscryan 
@linsun 
@howardjohn 
@ZackButcher

howardjohn:
@yuri-lcy your goals seem to be aligned with Ambient mesh. Do you have any interest in collaborating there? We have an #ambient mesh channel on slack I am happy to discuss more on.

yuri-lcy:
@howardjohn of course, we are interested in cooperation. We have learned a lot about ambient, and some ideas on L4 and L7 are very consistent with ours. We also want to solve security and tracing problems at the L4 layer, but in terms of L7 agents, ambient tends to Service account granularity, our idea is cluster granularity.

content/en/blog/2022/centralized-east-west-gateway/index.md
```diff
@@ -0,0 +1,295 @@
...
+## Configuration examples
+
+
+Next, we introduce how to deploy centralized east-west traffic gateway.
+
+Prerequisite:
+
+1. Only one Gateway can be attached to InternalGateway.
+2.  The VirtualService managed by InternalGateway must be attached to the Gateway mention in 1.
+3. CoreDNS is required.
```
rootsongjc, line 90: What components and versions of Istio are required?

yuri-lcy: Traffic hijacking is based on coredns, so coredns is a required component, and the rest of the conditions are consistent with istio itself. The version of istio is only adapted to version 1.15, but all versions should be applicable.

hzxuzhonghu (commented):
This can be seen as a practice using istio. 

Not new except by hijacking a DNS resolution.  IIUC the use case is limited to access only HTTP service by domain

yuri-lcy:
> This can be seen as a practice using istio.
> 
> Not new except by hijacking a DNS resolution. IIUC the use case is limited to access only HTTP service by domain

Yes, we extended it as a deployment mode of istio. 
https://github.com/istio/istio/pull/42360

yuri-lcy:
> Hi @yuri-lcy and @soya3129,
> 
> Is this blog post still relevant in May 2023? Let's close it if not.



> Hi @yuri-lcy and @soya3129,
> 
> Is this blog post still relevant in May 2023? Let's close it if not.

ok, close it, we will continue to communicate if there is any relevant progress.

ericvn:
Closing per @yuri-lcy

PR #11927, opened 2022-09-17
zh:remove outdated files in /tasks/traffic-management/egress/

windsonsea:
Sync with the en upstream and remove the zh file:
```
content/zh/docs/tasks/traffic-management/egress/egress_sni_monitoring_and_policies/index.md
```
- [x] Docs


ericvn (changes requested):
The first file was removed as part of https://github.com/istio/istio.io/pull/10250 which has updates in other files like egress-gateway-tls-orifination/index.md (removing a section for example). That section of the file still exists in the zh side of the tree so just removing the -sds/index.md file doesn't seem like the correct solution.

ericvn:
The second removed file was removed as part of https://github.com/istio/istio.io/pull/8071 so makes sense to simply remove it.

windsonsea:
> The second removed file was removed as part of #8071 so makes sense to simply remove it.

Thanks for your careful check. The sds/index.md file is restored. 😃

windsonsea:
Seems remaining this file there doesn't hurt anything, so this PR for removal can be closed.

PR #11939, opened 2022-09-18
zh:sync /reference/config/security/normalization/

windsonsea:
Added zh text to:

```
content/zh/docs/reference/config/security/normalization/index.md
```

- [x] Docs

### DIFF
--- a/content/zh/docs/reference/config/security/normalization/index.md
+++ b/content/zh/docs/reference/config/security/normalization/index.md
@@ -1,0 +1,95 @@
+---
+title: 授权策略规范化
+description: 描述授权策略中支持的规范化操作。
+weight: 40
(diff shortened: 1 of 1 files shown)


zirain (changes requested):
reference docs are auto-generated from istio/api repo, I'd think we should translate them.

zirain:
@ericvn IIRC, there's a PR related this?

windsonsea:
> reference docs are auto-generated from istio/api repo, I'd think we should translate them.

Yes, all .html zh-files are generated but .md files are localized in this folder.

image

zirain:
open an new ticket to track that.

windsonsea:
This can be closed since we need not to maintain the reference docs that are generated automatically.

PR #12005, opened 2022-09-26
Clean zh script

zirain:
- add *.sh under content/zh
- add check-localization avoid mistake in future PR

cc @ericvn


ericvn (approved):
As noted in https://github.com/istio/istio.io/blob/master/pkg/test/istioio/framework.go#L45-L47 only the `en/docs` are looked at for running tests.

Issue #12040, opened 2022-10-03
Store the k8s gateway-api crds url in one place


frankbu:
We currently have instructions referencing the gateway-api crds in several places:

1. https://preliminary.istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#setup
2. https://preliminary.istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#cleanup
3. https://deploy-preview-12021--preliminary-istio.netlify.app/latest/docs/setup/additional-setup/getting-started/
4. [boilerplate](https://github.com/istio/istio.io/blob/master/content/en/boilerplates/gateway-api-support.md)
5. [gateway-api util](https://github.com/istio/istio.io/blob/5e89d73d81406993c14bb80f8271bf7c3d2da992/tests/util/gateway-api.sh#L18)

All these places need to be updated whenever we change the gateway-api version that Istio is using.

We should store the crd url in one place which we can then reference in the docs, e.g., `{{< gateway_api_crds >}}`, similar to how we use `{{< istio_full_version >}}` in various places in the docs.

PR #12042, opened 2022-10-03
use short_codes for gateway api version

ericvn:

Fix https://github.com/istio/istio.io/issues/12040

Note that I used a version ov 0.5.1 as it is current and was a way to verify I got all the changes.

### DIFF
--- a/content/en/boilerplates/gateway-api-version.md
+++ b/content/en/boilerplates/gateway-api-version.md
@@ -1,5 +1,5 @@
 ---
 ---
-{{< text syntax=plain snip_id=gateway_api_version >}}
+{{< text syntax=plain snip_id=value >}}
 {{< k8s_gateway_api_version >}}
 {{< /text >}}
--- a/content/en/boilerplates/gateway-api-version.md
+++ b/content/en/boilerplates/gateway-api-version.md
@@ -1,0 +1,5 @@
+---
+---
+{{< text syntax=plain snip_id=gateway_api_version >}}
+{{< k8s_gateway_api_version >}}
+{{< /text >}}
--- a/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
@@ -20,7 +20,7 @@ and [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) 
 
     {{< text bash >}}
     $ kubectl get crd gateways.gateway.networking.k8s.io || \
-      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.5.0" | kubectl apply -f -; }
+      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref={{< k8s_gateway_api_version >}}" | kubectl apply -f -; }
     {{< /text >}}
 
 1. Install Istio using the `minimal` profile:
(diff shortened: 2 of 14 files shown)


ericvn:
Making DNM while verifying this change works.

content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
```diff
@@ -247,7 +247,7 @@ This resource does not actually exist in the cluster and is only used to signal
 For example, to redirect calls to `example.com` to an in-cluster `Service` named `example`:
 
 {{< text yaml >}}
-apiVersion: gateway.networking.k8s.io/v1beta1
+apiVersion: gateway.networking.k8s.io/{{< k8s_gateway_api_type >}}
```
howardjohn, line 250: do we need to use k8s_gateway_api_type? it changes so rarely it seems like a hassle to have to use this.

Also only some types are beta, so we cannot actually use it universally

ericvn: We could leave this out. The issue didn't mention it, and I simply added both.

ericvn: Removed api_type

ericvn:
One additional thing to try is grep the version from the go.mod and update the arg.yaml as part of the make get. I'll look into that.

tests/util/gateway-api.sh
```diff
@@ -14,16 +14,16 @@
...
 GATEWAY_API="true"
 
 function install_gateway_api_crds() {
-    kubectl kustomize "${K8S_GATEWAY_API_CRDS}" | kubectl apply -f -
+    bpsnip_gateway_api_support_install_crds
 }
 
 function remove_gateway_api_crds() {
-    kubectl kustomize "${K8S_GATEWAY_API_CRDS}" | kubectl delete -f -
+    bpsnip_gateway_api_support_install_crds
```
frankbu, line 26: This boilerplate snip is installing the crds instead of deleting them. It is also not quite right for the above `install_gateway_api_crds` either, since it is checking if they exist before installing, instead of just installing. Maybe you should add another dummy boilerplate just for use by this function, e.g., something like this (`boilerplates/gateway-api-util.md`:
```
---
---

{{< text syntax=plain snip_id=crd-version >}}
{{< k8s_gateway_api_version >}}
{{< /text >}}

```

Then you can change gateway-api.sh to this:

```
K8S_GATEWAY_API_CRDS="github.com/kubernetes-sigs/gateway-api/config/crd?ref=${bpsnip_gateway_api_util_crd_version}"
```

frankbu: Alternatively, if it doesn't slow things down too much, can `gateway-api.sh` just read the version value from `data/args.yaml`?

content/en/boilerplates/gateway-api-version.md
```diff
@@ -0,0 +1,5 @@
+---
+---
+{{< text syntax=plain snip_id=gateway_api_version >}}
```
frankbu, line 3: nit: the generated env var name is pretty ugly `bpsnip_gateway_api_version_gateway_api_version`

How about: `snip_id=value` instead?

ericvn: Done.

tests/util/gateway-api.sh
ericvn: We could do that, but I think the former is a better precedence if we have to do something like this again.

frankbu (approved):
LGTM, thanks!

PR #12051, opened 2022-10-05
Add gateway-api instructions: secure-ingress

frankbu:
Added Gateway API instructions to secure ingress task.

Preview: https://deploy-preview-12051--preliminary-istio.netlify.app/latest/docs/tasks/traffic-management/ingress/secure-ingress/

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
@@ -542,7 +542,7 @@ WARNING: THIS DOES NOT CURRENTLY WORK!
 
 Because the Kubernetes Gateway API does not currently support mutual TLS termination in a
 [Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway),
-we use an Istio-specific option, `networking.istio.io/tls.mode: MUTUAL`,
+we use an Istio-specific option, `gateway.istio.io/tls-terminate-mode: MUTUAL`,
 to configure it:
 
 {{< text bash >}}
@@ -564,7 +564,7 @@ spec:
       certificateRefs:
       - name: httpbin-credential
       options:
-        networking.istio.io/tls.mode: MUTUAL
+        gateway.istio.io/tls-terminate-mode: MUTUAL
     allowedRoutes:
       namespaces:
         from: Selector
--- a/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
@@ -538,7 +538,7 @@ EOF
 
 {{< tab name="Gateway API" category-value="gateway-api" >}}
 
-WARNING: THIS DOES CURRENTLY WORK!
+WARNING: THIS DOES NOT CURRENTLY WORK!
 
 Because the Kubernetes Gateway API does not currently support mutual TLS termination in a
 [Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway),
--- a/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
@@ -540,7 +540,7 @@ EOF
 
 A [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway)
 performing mutual TLS termination is configured no differently from one using simple TLS. The only difference
-is that the secret referenced by the `certifcateRefs` field now includes a client certificate with which
+is that the secret referenced by the `certifcateRefs` field now includes a CA certificate with which
 the gateway will authenticate the client request.
 
 {{< text bash >}}
--- a/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
@@ -11,143 +11,225 @@ test: yes
 ---
 
 The [Control Ingress Traffic task](/docs/tasks/traffic-management/ingress/ingress-control)
-describes how to configure an ingress gateway to expose an HTTP
-service to external traffic. This task shows how to expose a secure HTTPS
-service using either simple or mutual TLS.
+describes how to configure an ingress gateway to expose an HTTP service to external traffic.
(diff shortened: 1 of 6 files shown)


content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
```diff
@@ -11,143 +11,256 @@ test: yes
...
+
+First, create a [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway):
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: mygateway
+  namespace: istio-system
```
howardjohn, line 168: I wonder if we should start encourage best practice here to _not_ deploy into istio-system, following https://istio.io/latest/docs/setup/additional-setup/gateway/?

Also - we should probably add Gateway API as an install mechanism for gateways as well on that page. Which is essentially: "Gateway API: Do nothing!"

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
```diff
@@ -336,43 +491,90 @@ retrieves unique credentials corresponding to a specific `credentialName`.
...
-The server uses the CA certificate to verify
-its clients, and we must use the name `cacert` to hold the CA certificate.
+You can extend your gateway's definition to support [mutual TLS](https://en.wikipedia.org/wiki/Mutual_authentication).
+
+1. Change the credentials of the ingress gateway by deleting its secret and creating a new one.
+    The server uses the CA certificate to verify its clients, and we must use the name `cacert` to hold the CA certificate.
+
+    {{< text bash >}}
+    $ kubectl -n istio-system delete secret httpbin-credential
+    $ kubectl create -n istio-system secret generic httpbin-credential \
```
howardjohn, line 501: very nit, outside of scope: instead of delete+create which has downtime, I wonder if we should do `kubectl create --dry-run | kubectl apply`

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
```diff
@@ -336,43 +491,90 @@ retrieves unique credentials corresponding to a specific `credentialName`.
...
-         mode: MUTUAL
-         credentialName: httpbin-credential # must be the same as secret
-       hosts:
-       - httpbin.example.com
-    EOF
-    {{< /text >}}
+{{< tab name="Gateway API" category-value="gateway-api" >}}
+
+A [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway)
+performing mutual TLS termination is configured no differently from one using simple TLS. The only difference
```
howardjohn, line 542: I wonder if "mutual TLS termination is configured no differently" is going to be an issue... tools like cert-manager may always stick the CA cert in the secret?

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
```diff
@@ -336,43 +491,90 @@ retrieves unique credentials corresponding to a specific `credentialName`.
...
+is that the secret referenced by the `certifcateRefs` field now includes a client certificate with which
```
howardjohn, line 543:
```suggestion
is that the secret referenced by the `certifcateRefs` field now includes a CA certificate with which
```

The client provides the client certificate in the TLS handshake

howardjohn (approved):

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
frankbu: Good question, it certainly surprised me.

howardjohn: Actually are you sure its true? https://github.com/istio/istio/blob/1c70653e0eec55aed47674590767cd2dd4e8ff64/pilot/pkg/config/kube/gateway/conversion.go#L1611

Istio cannot (or doesnt, anyways) tell what is in the secret at config gen time anyways... I think its only SIMPLE. I suspect the example would pass even if you didn't send client cert

frankbu: > I suspect the example would pass even if you didn't send client cert

No, it does fail without the client cert. Look at step 3, here: https://deploy-preview-12051--preliminary-istio.netlify.app/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-mutual-tls-ingress-gateway

howardjohn: :confused: I don't know how, but there is some bug somewhere I think...

frankbu: Oops, actually you are right. The failing curl command has the wrong value for cacert (`--cacert example.com.crt` should be `--cacert example_certs1/example.com.crt`), so it's failing for an unrelated reason.

So, Gateway API doesn't support mTLS currently? That seems like a serious omission.

howardjohn: Yes, not currently

frankbu: How about we add an Istio option for now? Something like this:
```
kind: Gateway
metadata:
  name: mygateway
  namespace: istio-system
spec:
  gatewayClassName: istio
  listeners:
  - name: https
    hostname: "httpbin.example.com"
    port: 443
    protocol: HTTPS
    tls:
      mode: Terminate
      certificateRefs:
      - name: httpbin-credential
      options:
        networking.istio.io/tls.mode: MUTUAL
    allowedRoutes:
      namespaces:
        from: Selector
        selector:
          matchLabels:
            kubernetes.io/metadata.name: default
```

content/en/docs/tasks/traffic-management/ingress/secure-ingress/index.md
frankbu: Yes, that would be good, but in followup PRs.

frankbu:
Depends on https://github.com/istio/istio/pull/41299

ericvn (approved):
LGTM. 

The required dependency should be merged with this weekend's istio update.

howardjohn:
its plausible, maybe add `experimental.` prefix to it though? since we will
likely drop it when first class support is added?

PR #12069, opened 2022-10-09
fix operator install

zirain:

fix: https://github.com/istio/istio.io/pull/12066

- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/operator/test.sh
+++ b/content/en/docs/setup/install/operator/test.sh
@@ -38,6 +38,5 @@ sleep 30s
 _verify_contains snip_kubectl_get_svc "egressgateway"
 
 # @cleanup
-istioctl operator remove
 snip_cleanup
 kubectl delete ns istio-operator --grace-period=0 --force


zirain:
@ericvn I test this with and without https://github.com/istio/istio.io/pull/12066, both pass all tests, I'd like to revert https://github.com/istio/istio.io/pull/12069/commits/ed4e323ac22be2996c89fce4eb0831a9f6c30563

ericvn:
I'll look into this some more, but it looks like the cleanup is leaving things around:
https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/12066/doc.test.profile_none_istio.io/1578948069506945024/artifacts/tests-setup-profile-none-624e57/TestDocs/setup/install/operator/test.sh/after_snapshot/_test_context/diff.txt
so not sure how removing the uninstall would help the situation.

zirain:
with this PR https://github.com/istio/istio/pull/41245 , you need add `-y` after `istioctl operator remove`
but, snip_cleanup contains `istioctl uninstall --purge -y` https://github.com/istio/istio.io/blob/c0ed39df4884dbe6a701b715cb6098202b3e56b3/content/en/docs/setup/install/operator/snips.sh#L188
actually, `istioctl operator remove` is redundant

ericvn:
@zirain https://github.com/istio/istio.io/pull/12066 is simply moving the test to using the current istio commit. If something is failing, either something happened in istio/istio to break the current documentation, or some change was made which requires a documentation update to go along with the change (and certainly an release note and change note in the istio/istio PR that caused the issue). I'll go back and look at the istio/istio PRs from the last week and see if one of them makes sense for causing the breakage unless someone does it first.

ericvn (approved):
Let this PR merge to fix the test.

PR #12087, opened 2022-10-12
gateway-api doc: ingress-sni-passthrough

frankbu:

Preview: https://deploy-preview-12087--preliminary-istio.netlify.app/latest/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/

Also updated all instructions to use experimental install of gateway-api crds.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/boilerplates/gateway-api-support.md
```diff
@@ -12,7 +12,7 @@ installed before using the Gateway API:
 
 {{< text syntax=bash snip_id=install_crds >}}
 $ kubectl get crd gateways.gateway.networking.k8s.io || \
-  { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref={{< k8s_gateway_api_version >}}" | kubectl apply -f -; }
+  kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/{{< k8s_gateway_api_version >}}/experimental-install.yaml
```
howardjohn, line 15: I am not sure its a good idea to reocmmend most users use the experimental install... it is experimental for a reason

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/index.md
```diff
@@ -165,65 +175,150 @@ to hold the configuration of the NGINX server:
...
+{{< /tab >}}
+
+{{< tab name="Gateway API" category-value="gateway-api" >}}
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: mygateway
```
howardjohn, line 215: do we want a namespace here? Or I guess we should do whatever we do in other places

content/en/boilerplates/gateway-api-support.md
frankbu: Is there an alternative for docs that use the v1alpha2 features? I was thinking that if the experimental install is just a superset then it would be fine for docs that only use v1beta1 features and just ignore the experimental v1alpha2 things. But maybe there are also experimental additions in the v1beta1 resources making them less stable too?

Maybe it would be better to use the standard install as the default (for most docs?) and then only in docs that use v1alpha2 things (e.g., TLSRoute) use the experimental install?

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/index.md
frankbu: For now, I'm trying to keep the Gateway API and Istio versions of these docs as similar as possible. Generally the Istio versions don't include explicit namespaces, just for simplicity. I do agree we need to somehow figure out a good way to start educating users on the proper separation of gateway and route resources in different namespaces, but I don't think that docs like this one, which are just trying to demonstrate a specific function, is the right place.

content/en/boilerplates/gateway-api-support.md
howardjohn: It also could include experiments in beta resources (although I am not sure there are any, currently).

I think using standard where we can, and experimental where not makes the most sense for now. And we can just put a note like "This is experimental"

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/index.md
howardjohn: Fair enough

content/en/boilerplates/gateway-api-support.md
frankbu: OK, will do.

One more question. There are two ways to install the CRDs:

1. `kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref={{< k8s_gateway_api_version >}}" | kubectl apply -f -;`
2. `kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/{{< k8s_gateway_api_version >}}/standard-install.yaml`

The first, using `kustomize`, was how you did it in the first gateway-api doc. The second way is what it says to use in the gateway-api getting-started docs.

I noticed the difference between the two approaches is that the first (using `kustomize` command) is only installing the CRDs, while the second way installs the CRDs and the webhook. I also noticed that one (just one - ingress-gateways - for some reason) of the doc integration tests had an internal error accessing the webhook when I was using the second command to install the webhooks in the test framework (kind) env. Running through the same instructions manually on `minikube` had no problem.

So, I'm using the `kustomize` approach for the integration tests now but I'm wondering which command should we show in the doc for users?

frankbu:
This documented procedure works fine running locally, but the Gateway API variant of the test is failing with mysterious behavior. I disabled the failing test for now. Will fix and reenable the test in a followup PR.

ericvn (approved):

PR #12059, opened 2022-10-07
'Because of this' semantic discrepancy

RachaelLuo:
Please provide a description for what this PR is for.
I think the purpose of the last sentence in this passage is to explain why the route goes to v3 (that is, yaml below), but the penultimate sentence of this passage has clarified that v3 satisfies the routing rules. Therefore, the meaning is repeated, and it is easy to produce ambiguity.
And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

- [ ] Configuration Infrastructure
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

PR #12093, opened 2022-10-13
'Because of this' semantic discrepancy

RachaelLuo:
ooptimization

'Because of this' semantic discrepancy

This is the result of the port [12059](https://github.com/istio/istio.io/pull/12059).

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/concepts/traffic-management/index.md
+++ b/content/en/docs/concepts/traffic-management/index.md
@@ -261,7 +261,7 @@ subset named v2. You’ll see how you define a service subset in the section on
 Routing rules are **evaluated in sequential order from top to bottom**, with the
 first rule in the virtual service definition being given highest priority. In
 this case you want anything that doesn't match the first routing rule to go to a
-default destination, specified in the second rule. Because of this, the second
+default destination, specified in the second rule. Here, the second
 rule has no match conditions and just directs traffic to the v3 subset.
 
 {{< text yaml >}}


content/en/docs/concepts/traffic-management/index.md
```diff
@@ -261,8 +261,8 @@ subset named v2. You’ll see how you define a service subset in the section on
 Routing rules are **evaluated in sequential order from top to bottom**, with the
 first rule in the virtual service definition being given highest priority. In
 this case you want anything that doesn't match the first routing rule to go to a
-default destination, specified in the second rule. Because of this, the second
-rule has no match conditions and just directs traffic to the v3 subset.
+default destination, specify that in a second rule which has no match conditions.
+Here, the rule directs traffic to the v3 subset.
```
ericvn, line 265: A thought would be to have a snippet here with two rules, the match plus the default behavior if no match. Then replace this sentence with something like `Here, the second rule redirects traffic to the v3 subset if the first rule is unmatched.`

frankbu: I think the problem is that the example needs to be fixed, something like this:
```
{{< text yaml >}}
- match:
   - headers:
       end-user:
         exact: jason
  route:
  - destination:
      host: reviews
      subset: v2
- route:
  - destination:
      host: reviews
      subset: v3
{{< /text >}}
```
Then I would change the wording:
```
Routing rules are **evaluated in sequential order from top to bottom**, with the
first rule in the virtual service definition being given highest priority.
In the following example, anything that doesn't match the first routing rule
will go to a default destination as specified in the second rule, which has no match
conditions and just directs traffic to the v3 subset.
```

RachaelLuo: In the original "In this case you want anything that doesn’t match the first routing rule to go to a default destination, specified in the second rule. ", it has been stated that anything that does not match the first routing rule goes to the default target specified in the second rule, so finally "if the first rule is unmatched" is not necessary.
image

RachaelLuo: Speaking of examples, I want to know why v1 and v2 are not used in this example. I think this is more intuitive.

RachaelLuo: In the "anything that doesn't match the first routing rule
will go to a default destination as specified in the second rule, which has no match
conditions and just directs traffic to the v3 subset.","which has no match
conditions and just directs traffic to the v3 subset",it is easy to see as a supplement to the default target rather than this example.

RachaelLuo: I think in the "anything that doesn't match the first routing rule
will go to a default destination as specified in the second rule, which has no match
conditions and just directs traffic to the v3 subset.","which has no match
conditions and just directs traffic to the v3 subset",it is easy to see as a supplement to the default target rather than this example.

RachaelLuo: The example is the unified style of the full text. I don't think there is much problem for the time being.

PR #12097, opened 2022-10-14
Add missing multiCluster.clusterName attributes

antoineco:
### What the PR solves

Without a `multiCluster.clusterName` configured inside the `IstioOperator` object for the remote cluster, `istioctl` creates the `istiod` Service with ClusterIP `None`, which prevents calling Kubernetes webhooks (e.g. sidecar injection).

Ref. https://discuss.istio.io/t/pods-not-created-with-headless-service/13940/3?u=antoineco

### Areas

- [X] Docs
- [X] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/multicluster/primary-remote/index.md
+++ b/content/en/docs/setup/install/multicluster/primary-remote/index.md
@@ -136,6 +136,8 @@ spec:
     istiodRemote:
       injectionPath: /inject/cluster/cluster2/net/network1
     global:
+      multiCluster:
+        clusterName: cluster2
       remotePilotAddress: ${DISCOVERY_ADDRESS}
 EOF
 {{< /text >}}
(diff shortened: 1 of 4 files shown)


ericvn:
@frankbu I believe you removed this in an earlier PR. Was that a change specific to 1.15 (which is what is shown on istio.io)?

@antoineco I don't see a version in the discuss. Are you maybe trying this on version 1.14 or earlier. The 1.14 version of the docs do have this attribute.

antoineco:
@ericvn you're right, apologies for not mentioning it. I was indeed using Istio 1.14 when I ran into this issue.
I didn't realize this had changed in 1.15.

I just deployed a new multi-cluster setup on Istio 1.15 and the issue is gone, even if I don't set `multiCluster.clusterName` explicitly.

Sorry for the noise!

PR #12169, opened 2022-10-31
Egress TLS origination: Modify the role to allow only TLS certificate secret access

cagataygurturk:
The role definition in the documentation suggests allowing a pod to read all the secrets in the namespaces and this is violating the least privilege principle.

It is possible to allow access only to the secret containing the TLS certificate and it is confirmed that with this narrow role, the sidecar proxy is still able to access the secret. 

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


cagataygurturk:
Closed as I made a mistake with branch

PR #12171, opened 2022-10-31
Egress TLS origination: Modify the role to allow only TLS certificate secret access

cagataygurturk:
The role definition in the documentation suggests allowing a pod to read all the secrets in the namespaces and this is violating the least privilege principle.

It is possible to allow access only to the secret containing the TLS certificate and it is confirmed that with this narrow role, the sidecar proxy is still able to access the secret. 

- [ ] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/egress/egress-tls-origination/index.md
+++ b/content/en/docs/tasks/traffic-management/egress/egress-tls-origination/index.md
@@ -238,7 +238,7 @@ Follow [these steps](/docs/tasks/traffic-management/egress/egress-gateway-tls-or
 1. Create required `RBAC` to make sure the secret created in the above step is accessible to the client pod, which is `sleep` in this case.
 
     {{< text bash >}}
-    $ kubectl create role client-credential-role --resource=secret --verb=list
+    $ kubectl create role client-credential-role --resource=secret --verb=get --resource-name=client-credential
     $ kubectl create rolebinding client-credential-role-binding --role=client-credential-role --serviceaccount=default:sleep
     {{< /text >}}
 


content/en/docs/tasks/traffic-management/egress/egress-tls-origination/index.md
```diff
@@ -238,7 +238,7 @@ Follow [these steps](/docs/tasks/traffic-management/egress/egress-gateway-tls-or
 1. Create required `RBAC` to make sure the secret created in the above step is accessible to the client pod, which is `sleep` in this case.
 
     {{< text bash >}}
-    $ kubectl create role client-credential-role --resource=secret --verb=list
+    $ kubectl create role client-credential-role --resource=secret --verb=get --resource-name=client-credential
```
kfaseela, line 241: Please check: https://github.com/istio/istio/blob/master/pilot/pkg/credentials/kube/secrets.go#L156 This won't work. I can run the documentation test for you on this PR and show that this will fail.

cagataygurturk: Thanks for pointing this out. However with this narrow role, my setup still works interestingly.
Could you run the tests and let's see if it will fail or not?

cagataygurturk: All right, I was able to reproduce the error myself. Closing the PR then.

Issue #11986, opened 2022-09-21
Performance doc charts are out of date.


ericvn:
https://istio.io/latest/docs/ops/deployment/performance-and-scalability/

Docs use {{< istio_release_name >}} so they state 1.15.

Chart svgs were last updated for 1.8 https://github.com/istio/istio.io/pull/8574

The results in the text are likely out of date as well.

zirain:
perf site has down for a long time.

ericvn:
Performance data is back.

PR #12176, opened 2022-11-02
Add performance expectations in ext-authz page

bochengchu:
Per https://github.com/istio/enhancements/pull/125,
adding performance benchmarking link.

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/security/authorization/authz-custom/index.md
+++ b/content/en/docs/tasks/security/authorization/authz-custom/index.md
@@ -240,3 +240,7 @@ The external authorizer is now ready to be used by the authorization policy.
     {{< /text >}}
 
 1. Remove the extension provider definition from the mesh config.
+
+## Performance expectations
+
+See [performance bencmarking](https://github.com/istio/tools/tree/master/perf/benchmark/configs/istio/ext-authz).


Arhell (commented):
/ok-to-test

ericvn:
Looking at he `lint logs` it shows a misspelling: `bencmarking`

bochengchu:
> Looking at he `lint logs` it shows a misspelling: `bencmarking`

Fixed.

ericvn:
Throw a DNM while we try and come up with an answer to Mitch's question.

First, we should notate the link using https://istio.io/latest/docs/releases/contribute/shortcodes/#add-links-to-content-on-github if we are keeping it.

Does putting the data in istio.io make a difference? I have an open issue on that currently: https://github.com/istio/istio.io/issues/11986.

We do have a number of links to other repo files in our documentation, but the majority of them are pointers to something with `samples` in the directory name. Most of those files may actually be maintained by the docs-maintainers.

So, I'm not sure which way to go with this. I think we have a good set of docs/tools to talk about this topic pointed to by the link. The question is should we wrapper it or copy it into istio.io? If we have a bigger Doc maintainership team, I would be inclined to say yes. But, since we don't, I'm thinking that if the referenced code/doc is good now, it will likely stay as good as it would if it was pulled in.

therealmitchconnors:
If Eric is happy, I'm happy. ;)

ericvn:
Thanks Mitch. I'll let this merge.

Issue #12013, opened 2022-09-27
Kubernetes Ingress manual - an outdated API version 


Bo-Zi:
Hi!

Please update the [Specifying IngressClass](https://istio.io/latest/docs/tasks/traffic-management/ingress/kubernetes-ingress/#specifying-ingressclass) section 

According to the [K8s Deprecated API Migration Guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingressclass-v122) :
> The networking.k8s.io/v1beta1 API version of IngressClass is no longer served as of v1.22.
>     Migrate manifests and API clients to use the networking.k8s.io/v1 API version, available since v1.19.

Using the  Kubernetes Ingress manual gives an **error**
> **the server could not find the requested resource**

Since Kubernetes 1.19 _apiVersion: networking.k8s.io/v1_ is recommended 
```yaml 
apiVersion: networking.k8s.io/v1
kind: IngressClass
metadata:
  name: istio
spec:
  controller: istio.io/ingress-controller
```

Thank you for guys  your great work!!!

ericvn:
Added label for the @istio/wg-networking-maintainers to take a look.

howardjohn:
+1, should be updated

PR #12200, opened 2022-11-08
Update ingress task to use networking.k8s.io/v1 apiVersion (#12013)

calebROlson:
**Description**
This PR updates the documentation page for the Ingress task to use `networking.k8s.io/v1` endpoint instead of the deprecated `networking.k8s.io/v1beta` endpoint (https://istio.io/latest/docs/tasks/traffic-management/ingress/ingress-control/).

Other fields are also updated for this new API endpoint (path/pathType, serviceName/servicePort)

Related issue: #12013 

I am working with @johnma14 to update this.
I tested this new specification locally and it seems to exhibit the same behavior as before the change, but please let me know if other fields should also be updated or of other issues. Thanks.

**Areas that this PR affects**
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/kubernetes-ingress/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/kubernetes-ingress/index.md
@@ -28,7 +28,7 @@ Let's see how you can configure a `Ingress` on port 80 for HTTP traffic.
 
     {{< text bash >}}
     $ kubectl apply -f - <<EOF
-    apiVersion: networking.k8s.io/v1beta1
+    apiVersion: networking.k8s.io/v1
     kind: Ingress
     metadata:
       annotations:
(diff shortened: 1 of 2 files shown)


johnma14:
@calebROlson thanks for the PR. Please review the https://github.com/istio/istio.io/blob/master/tests/README.md to see how to regenerate the bash script for the existing tests for this page. You will need to run the following commands:

```
make snips
make lint-fast
```

Thanks.

ericvn:
I only see snippet updates in the doc so far, so the `make gen` should be all that's needed. It should create a new snips.sh with the updated snippets and no function name changes. Since there are no function changes the test script should be fine. The actual test should verify the snippets.

johnma14:
tests passed and changes look good to me. Thanks for the PR.

ericvn (approved):

PR #12263, opened 2022-11-20
we are using the incorrect namespace and incorrect service name

luis-giraldo:
According to the docs provided in https://istio.io/latest/docs/examples/bookinfo/ , the ingress-gateway is created under the istio-ingress namespace, not under the istio-system namespace

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
@@ -225,7 +225,7 @@ Set the `INGRESS_HOST` and `INGRESS_PORT` environment variables according to the
 If you are unsure, you can determine if your Kubernetes cluster is running in an environment that supports external load balancers using the following command:
 
 {{< text bash >}}
-$ kubectl get svc istio-ingressgateway -n istio-system
+$ kubectl get svc istio-ingressgateway -n istio-ingress
 NAME                   TYPE           CLUSTER-IP       EXTERNAL-IP      PORT(S)   AGE
 istio-ingressgateway   LoadBalancer   172.21.109.129   130.211.10.121   ...       17h
 {{< /text >}}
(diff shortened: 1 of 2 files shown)


zirain:
you need run `make gen` to update test script

frankbu:
> According to the docs provided in https://istio.io/latest/docs/examples/bookinfo/ , the ingress-gateway is created under the istio-ingress namespace, not under the istio-system namespace

@luis-giraldo Where in this doc does it say to create the ingress-gateway in the istio-ingress namespace? As far as I know, all of the startup and example docs use the Istio `demo` or `default` profile to install Istio, which, by default, will install the ingress gatewaway in the istio-system namespace. Installing gateways separately in a different (e.g. istio-ingress) namespace is a recommended production pattern, but not what is done by default.

luis-giraldo:
make gen

kfaseela:
> make gen

@luis-giraldo : you will have to run `make gen` locally in your istio.io repo, and amend this commit with the generated files

luis-giraldo:
@frankbu this is what happens when you install via helm

ericvn:
Ref: https://github.com/istio/istio/issues/39049

frankbu:
> @frankbu this is what happens when you install via helm

@luis-giraldo It's unfortunate that helm and istioctl install things differently. We can't change the instructions to the helm way, it needs to accommodate both. The doc test won't pass with your change, since the doc framework does not use helm to install.

Here's a PR that updates the doc to handle ingress gateway locations more generally: #12308

ericvn:
I believe https://github.com/istio/istio.io/pull/12315 better solve this issue. Feel free to reopen if something else is needed.

luis-giraldo:
Yup, thanks!

PR #12295, opened 2022-11-24
Fix flaky test failures

frankbu:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/tasks/traffic-management/mirroring/test.sh
```diff
@@ -55,6 +54,9 @@ _wait_for_deployment default httpbin-v2
 snip_before_you_begin_4
 _wait_for_deployment default sleep
 
+# wait some more for the route config to be applied to the sleep pod
+sleep 30s # TODO proper wait for config update
```
kfaseela, line 58: can we instead do a proxy-config output and wait for the expected config to be available?

frankbu: Yes, something like that would be better. The current sleep 30s may be good enough for now. It looks like the latest test failure is the other flaky problem (timeout) even though it's now set to `DEFAULT_TIMEOUT=90m`. That seems to imply that there is more to it than just a combination of a busy build machine and lots of new tests??

PR #12320, opened 2022-12-01
docs update for policies graduation

aryan16:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/concepts/security/index.md
+++ b/content/en/docs/concepts/security/index.md
@@ -575,7 +575,7 @@ access the workloads with the `app: httpbin` and `version: v1` labels in the
 `foo` namespace when requests sent have a valid JWT token.
 
 {{< text yaml >}}
-apiVersion: security.istio.io/v1beta1
+apiVersion: security.istio.io/v1
 kind: AuthorizationPolicy
 metadata:
  name: httpbin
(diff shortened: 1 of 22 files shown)


kfaseela:
Not sure where this is used, but do we have to change here? https://github.com/istio/istio.io/blob/master/pkg/test/istioio/snapshot.go#L221

aryan16:
> Not sure where this is used, but do we have to change here? https://github.com/istio/istio.io/blob/master/pkg/test/istioio/snapshot.go#L221

not required, v1 is same as v1beta1. Under the hood we are still using v1beta1.

linsun (approved):
LGTM, exciting stuff!

kfaseela (approved):

ericvn:

dashboard flake.

ericvn (approved):

ericvn:
I was able to run the test locally successfully from the main branch, as well as cloning this test branch.

ericvn:
I ran this branch 3 times, with one failure if doing the git of gateway code, however that does not seem to be the problem here.

ericvn:

Try once more.

ericvn:
A few thoughts. Do we need a more recent istio version for something that this test uses? While I don't see a files that seem to imply that's an issue. If that was the case, I would have expected the local build to fail unless I moved to a more recent Istio SHA.

PR #9324, opened 2021-03-19
Add new istio.io design

Jakub-Sadowski:
Update master
[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience
[ ] Developer Infrastructure

PR #12366, opened 2022-12-14
Add warning banner for older blog posts

kfaseela:

Previously there used to be a warning banner for blog posts to indicate which target_release this was tested against, which seems to have broken ever since new istio.io design was introduced. (https://github.com/istio/istio.io/pull/9324)

For eg: see https://istio.io/v1.9/blog/2019/webhook/  vs https://istio.io/latest/blog/2019/webhook/

@ericvn This is basically because blog posts no longer are using [primary_top.html ](https://github.com/istio/istio.io/blob/master/layouts/partials/primary_top.html#L157). However adding this directly to the blog posts brings in too many other changes to the page layout, including layout changes and bringing back sidebar on the left hand side, which would have been purposefully avoided as per the new design(i am not sure, you may know the history better). Hence I have tried copying only the target_release section for the blog posts.

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/layouts/blog/single.html
+++ b/layouts/blog/single.html
@@ -5,6 +5,30 @@
         <p>{{ .Params.description }}</p>
       </div>
       <p class="post-author">{{ (.Params.publishdate).Format "Jan 2, 2006" }} <span> | </span> By {{ replace (replace .Params.attribution "(" "- ") ")" "" }}</p>
+      {{ if .Page.Params.target_release }}
+          {{ $page_parts := split .Page.Params.target_release "." }}
+          {{ $page_version := int (index $page_parts 0) }}
+          {{ $page_revision := int (index $page_parts 1) }}
+          {{ $page_normalized_version := add (add (mul $page_version 1000000) (mul $page_revision 1000)) 0 }}
+
+          {{ $site_parts := split .Site.Data.args.version "." }}
+          {{ $site_version := int (index $site_parts 0) }}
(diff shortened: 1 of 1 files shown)


ericvn:
@kfaseela Thanks for looking onto this and looking at the netlify output I can see the output now.

Two thoughts looking into the repo:
- Maybe it is an oversight, or maybe someone made a decision to not use this going forward. @craigbox, who was instrumental in the new design, may have some feedback of if this was simply missed in the redesign. I suspect it was the later, but want to confirm. As a FYI, the original work was done in https://github.com/istio/istio.io/pull/5134.
- It seems that for blog posts starting in 2020, we have neglected the `target_release:` field in blogs. I'm not sure if we should possibly go back and identify if some blogs point to a specific release and if this item should be set in some older blogs or not. Depending on the answer to the first item, we may think about going back and updating some.

craigbox:
This wasn't deliberately dropped, but when we built the plan we looked at what was current on the web site, and so probably didn't see any old posts to put it into the requirements.

I have no personal concerns about adding this back into the template, as the warning is significantly vague as to not say "this content is 100% out of date"

(See https://deploy-preview-12366--preliminary-istio.netlify.app/latest/blog/2019/webhook/ for render)

craigbox (approved):

PR #12249, opened 2022-11-16
Update supported versions with 1.16 info

jacob-delgado:
Please provide a description for what this PR is for.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

- [ ] Configuration Infrastructure
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

PR #12368, opened 2022-12-14
Update Supported Versions doc

ericvn:

Replacement for https://github.com/istio/istio.io/pull/12249 which is needing a lint fix.


content/en/docs/releases/supported-releases/index.md
```diff
@@ -79,7 +79,7 @@ Please keep up-to-date and use a supported version.
 
 | Minor Releases   | Patched versions with no known CVEs           |
 |------------------|-----------------------------------------------|
+| 1.16.x           | 1.16.0+                                       |
 | 1.15.x           | 1.15.3+                                       |
 | 1.14.x           | 1.14.5+                                       |
-| 1.13.x           | 1.13.9+                                       |
-| 1.12 and earlier | None, all versions have known vulnerabilities |
+| 1.13 and earlier | None, all versions have known vulnerabilities |
```
ericvn, line 85: Technically 1.13.9 doesn't have a known vulnerability, at least I don't see where we had a fix shown in https://istio.io/latest/news/ which called out 1.13.9. But 1.13 is EOL. We could call out releases with no CVEs that have reached EOL, 1.14 will be there soon, with some different text, such as:

```suggestion
| 1.13                    | End of life. A new CVE will NOT be patched.  |
| 1.12 and earlier | None, all versions have known vulnerabilities |
```

content/en/docs/releases/supported-releases/index.md
```diff
@@ -79,7 +79,7 @@ Please keep up-to-date and use a supported version.
 
 | Minor Releases   | Patched versions with no known CVEs           |
 |------------------|-----------------------------------------------|
+| 1.16.x           | 1.16.0+                                       |
```
kfaseela, line 82: we already released 1.16.1 , so should this be 1.16.1+ ?

ericvn: 1.16.0 doesn't have any CVE's so it's still 1.16.0. We only increase this value when we have an actual CVE. Example: 1.15.3 was determined as https://istio.io/latest/news/security/istio-security-2022-008/ says 1.15.2 has a CVE. 1.14 was determined by https://istio.io/latest/news/security/istio-security-2022-007/.

content/en/docs/releases/supported-releases/index.md
ericvn: Any feedback on if the second suggestion is better or worse would be helpful. I can make the change if needed.

kfaseela: Second suggestion looks better to me.

ericvn: Updated adding a separate 1.13 line.

kfaseela (approved):

PR #12387, opened 2022-12-16
fix grammar soft-multitenancy blog

kfaseela:

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/blog/2018/soft-multitenancy/index.md
+++ b/content/en/blog/2018/soft-multitenancy/index.md
@@ -28,7 +28,7 @@ instance. Separation between the tenants is provided by Kubernetes namespaces an
 One use case for this deployment model is a shared corporate infrastructure where malicious
 actions are not expected, but a clean separation of the tenants is still required.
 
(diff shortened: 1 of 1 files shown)


content/en/blog/2018/soft-multitenancy/index.md
```diff
@@ -28,7 +28,7 @@ instance. Separation between the tenants is provided by Kubernetes namespaces an
 One use case for this deployment model is a shared corporate infrastructure where malicious
 actions are not expected, but a clean separation of the tenants is still required.
 
-Potential future Istio multi-tenant deployment models are described at the bottom of this
+Potential future of Istio multi-tenant deployment models are described at the bottom of this
```
ericvn, line 31: I think the original text is correct. Below we talk about future models, not the future **of the** models. These are very close in definition, but I think the later implies more of a keep/remove of the models as a whole while the former is more of how the models might change or additional models added. But, this could just be my interpretation.

PR #12388, opened 2022-12-16
Change "-ppassword" > "-password"

bobbypower-asm:
It seems like this is a trivial typo, as I'm not seeing other uses of "-ppassword"

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn:
Looking at https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html it seems there should be a double hash on the options (password matches) and we should maybe expand out the -e to --execute?

@istio/wg-environments-maintainers owns this page. I'll let them determine what this should be.

bobbypower-asm (commented):
Additional improvements

ericvn:
You need to run `make gen` to create a new snippet file with the markdown file changes.

dhawton:
`-ppassword` isn't necessarily a typo. With mysql, when you combine `-p` and additional text, it is read as the password from the commandline. In this case, `-ppassword` means "password" is being passed in as the password instead of a password prompt being generated.

bobbypower-asm:
Good to know - thank you for that info!

ericvn:
> `-ppassword` isn't necessarily a typo. With mysql, when you combine `-p` and additional text, it is read as the password from the commandline. In this case, `-ppassword` means "password" is being passed in as the password instead of a password prompt being generated. `-ppassword` is the same as `--password=password` https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_password

Thanks @dhawton. So the text is correct. It is just using the short `-` options and `-p{password}`.

PR #12389, opened 2022-12-16
Typo fix ("namespaced" > "namespace")

bobbypower-asm:

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/reference/glossary/namespace-sameness.md
+++ b/content/en/docs/reference/glossary/namespace-sameness.md
@@ -5,7 +5,7 @@ test: n/a
 
 Within a multicluster mesh, [namespace sameness](https://github.com/kubernetes/community/blob/master/sig-multicluster/namespace-sameness-position-statement.md)
 applies and all namespaces with a given name are considered to be the same namespace. If multiple clusters contain a
-`Service` with the same namespaced name, they will be recognized as a single combined service. By default, traffic is
+`Service` with the same namespace name, they will be recognized as a single combined service. By default, traffic is
 load-balanced across all clusters in the mesh for a given service.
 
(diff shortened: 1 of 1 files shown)


ericvn (approved):

content/en/docs/reference/glossary/namespace-sameness.md
```diff
@@ -5,7 +5,7 @@ test: n/a
 
 Within a multicluster mesh, [namespace sameness](https://github.com/kubernetes/community/blob/master/sig-multicluster/namespace-sameness-position-statement.md)
 applies and all namespaces with a given name are considered to be the same namespace. If multiple clusters contain a
-`Service` with the same namespaced name, they will be recognized as a single combined service. By default, traffic is
+`Service` with the same namespace name, they will be recognized as a single combined service. By default, traffic is
```
frankbu, line 8: I don't think this change is correct. The service has to have the same namespace and name, which is what "same namespaced name" means. By saying the "same namespace name" it's just saying the name of the namespaces have to be the same, nothing about the service name itself.

bobbypower-asm: Understood! I will revert this change now.

ericvn: My thought was that the service would already have the same same so implying same namespace means both are equal. I'm not sure that namespaced really means "have the same namespace and name", although in this case that's how it was used. 

I would entertain that we extend this to say the same service name and namespace to be clear.

frankbu: I agree, let's make it unambiguous by saying `the same service name and namespace`.

PR #12390, opened 2022-12-16
Check for go package updates monthly

jlosito:

This will use dependabot to check for any updates against the go dependencies which this project uses on a monthly basis and submit pull requests with version bumps.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [x] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "monthly"


ericvn:
I'm not sure this makes sense, but I don't know a lot about dependabot. We have a weekly job that gets the latest istio.istio, and does `go mod tidy` into the go.mod which pulls along some other package updates. Will dependabot try and update some other packages during its monthly run that might get reverted by the weekly `go mod tidy` after the istio/istio update? Would it be better to do the dependabot on istio/istio and would that cause updates here during the weekly `go mod tidy`? Or maybe we need something in both?

ericvn:
Trying to better understand a --dry-run yielded:
```
INFO Results:
+---------------------------------------------------------+
|           Changes to Dependabot Pull Requests           |
+---------+-----------------------------------------------+
| created | k8s.io/apimachinery ( from 0.25.3 to 0.26.0 ) |
| created | k8s.io/client-go ( from 0.25.3 to 0.26.0 )    |
+---------+-----------------------------------------------+
```

ericvn:
I then did a `go get` on the two packages, followed by a `go mod tidy`. The package versions were not reverted. 

Also comparing to the go.mod in istio/istio, it was already updated in https://github.com/istio/istio/pull/42397.

I also did a go get istio/istio as we typically do weekly. This also picked up the 1.26.0 package update.

Summary, it seems that the weekly `go get istio/istio` and `go mod tidy` will pick up changes done in istio/istio. Will it pick up all required changes, I'm not sure, but in this case it did. It seems that a dependabot job on istio/istio might be a better thing and the changes it finds will moved here on a weekly basis.

ericvn:
Thanks @jlosito. Based on your last comment, I'll close this for now since I believe we will see the updates found in my sample run over the weekend as the istio/istio is pulled. 

As an aside, I did learn some things here and I'll see what other packages might need updating in istio/istio.

PR #12404, opened 2022-12-19
Clarify that Sidecar is not applicable to gateways

jewertow:
Signed-off-by: Jacek Ewertowski <jewertow@redhat.com>

To avoid confusion like in this [issue](https://github.com/istio/istio/issues/32041), I think it's worth to clarify in a warning that `Sidecar` is not applicable to gateways. Users may guess that `Sidecar` should work for gateways, because both gateway and sidecar proxies are the same istio-proxy.

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn (changes requested):
This file is copied from elsewhere (see the comment at the top of the file).

Looking it seems to be here: https://github.com/istio/api/blob/master/networking/v1alpha3/sidecar.proto#L61-L63

jewertow:
Thanks @ericvn.

PR #12422, opened 2022-12-28
bugfix(proxyConfig): for sidecar and gateway, default concurrency is …

j2gg0s:
Default concurrency for sidecar and gateway is different.

- [X ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


zirain (changes requested):
you should not update reference doc directly,
the right way is updating the comment in api repo, the Automater will sync to this repo.

PR #12439, opened 2023-01-10
Add `targetPort` to a Service for gateway injection

jewertow:
Signed-off-by: Jacek Ewertowski <jewertow@redhat.com>

An example of gateway injection is missing the `targetPort` that may be required in some environments. I experienced this problem on OpenShift with MetalLB.

The [documentation](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec) says that this field is ignored when `clusterIP=None`, but the `type: LoadBalancer` does not imply that `clusterIP` will be set to `None`.

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/additional-setup/gateway/index.md
+++ b/content/en/docs/setup/additional-setup/gateway/index.md
@@ -123,10 +123,12 @@ spec:
   selector:
     istio: ingressgateway
   ports:
-  - port: 80
-    name: http
-  - port: 443
-    name: https
+  - name: http
+    port: 80
+    targetPort: 8080
+  - name: https
+    port: 443
+    targetPort: 8443
 ---
 apiVersion: apps/v1
 kind: Deployment
--- a/content/en/docs/setup/additional-setup/gateway/snips.sh
+++ b/content/en/docs/setup/additional-setup/gateway/snips.sh
@@ -64,10 +64,12 @@ spec:
   selector:
     istio: ingressgateway
   ports:
-  - port: 80
-    name: http
-  - port: 443
-    name: https
+  - name: http
+    port: 80
+    targetPort: 8080
+  - name: https
+    port: 443
+    targetPort: 8443
 ---
 apiVersion: apps/v1
 kind: Deployment


ericvn:
@istio/wg-environments-maintainers Thoughts on this?

jewertow:
To avoid confusion, I would like to note that `targetPort`s are set by default in the ingress-gateway helm [chart](https://github.com/istio/istio/blob/master/manifests/charts/gateways/istio-ingress/values.yaml#L10).

frankbu:
>  may be required in some environments

It seems to me that targetPort is required in every environment that's using the default ingressgateway image, since the ingressgateway targetPort != port, so it won't be correct if not set.

Looks like the test for this doc just does the install, but doesn't check to see if the gateway is actually working. Maybe it would be good to add a curl command after each of the three installs to see if it's able to talk to the gateway? Here, for example: https://github.com/istio/istio.io/blob/9434bc1ee6d404fc4b030d1c98acb9fd7037627c/content/en/docs/setup/additional-setup/gateway/test.sh#L50

cc @howardjohn

howardjohn (changes requested):
This is instilling bad practices **that cannot be undone**.

We are pushing for https://github.com/istio/istio/blob/d5a828ebb79c69570cc80c1fead889fbec008d22/manifests/charts/gateway/templates/deployment.yaml#L41 to be the primary option to deploy this. This will also be the only option for gateway-api

> To avoid confusion, I would like to note that `targetPort`s are set by default in the ingress-gateway helm [chart](https://github.com/istio/istio/blob/master/manifests/charts/gateways/istio-ingress/values.yaml#L10).

That is the legacy chart.

frankbu:
@howardjohn, just to clarify, are you saying that the example should work without the targetPort set if it's using the latest version of the injector, or would there need to be additional configuration to make it use the new template?

Like I said above, it would be nice to have the doc test try to access the gateway after installing, to make sure it actually works.

howardjohn:
The injector does not inject it. It also does work out of the box, it just requires root; openshift doesn't allow that though.

dgn:
`securityContext.sysctls` should work fine on OpenShift, too as this one's considered a `safe` sysctl

jewertow:
Ok, I reinstalled everything from scratch and it turned out that it works fine out of the box. However, this is very strange, because I experienced this problem regularly, also on KinD. But I noticed that it depends on the order of changes applied to MetalLB and deployed services, e.g. removing `targetPort` after deploying service breaks routing, like it would be required.
Anyway, I'm closing this PR. Thanks!

PR #12481, opened 2023-01-17
Extend doc test timeout from 60 minutes to 75 minutes.

ericvn:

Just a PR to see if extending the timeout may help with some of the profile_default test failures. Looking at the logs, it seems that the tests are hanging so this isn't necessarily a good thing as we'll just end up waiting longer for it to fail. That's why this is a DNM.


ericvn:

try again

frankbu:
As I mentioned in the issue comment, I once tried bumping the timeout to 2 hours and it still failed, so it seems like somehow it gets into a deadlock state?

ericvn:
Agree Frank. In this case I think the networking fails to install correctly and the traffic just doesn't flow.

PR #12486, opened 2023-01-19
Update Gateway selector in ingress demo for Helm Installation

nshankar13:
Signed-off-by: nshankar <nshankar@microsoft.ghe.com>

When the Istio ingress-gateway is installed via the Gateway Helm chart, following the steps provided in the [Helm Install guide](https://istio.io/latest/docs/setup/install/helm/), the selector for the ingress-gateway is different from what it is when the ingress-gateway is installed via istioctl. This PR updates the ingress guide for the appropriate configuration for Helm installations. 

```
Labels:       app=istio-ingress
              istio=ingress
              istio.io/rev=default
              pod-template-hash=54464bcb87
              service.istio.io/canonical-name=istio-ingress
              service.istio.io/canonical-revision=latest
              sidecar.istio.io/inject=true
```

- [X] Docs
- [X] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
@@ -79,6 +79,30 @@ spec:
 EOF
 {{< /text >}}
 
+{{< tip >}}
+If you installed Istio using Helm, the selector for the `Gateway` is `istio=ingress`:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: httpbin-gateway
+spec:
+  selector:
+    istio: ingress # use Istio gateway installed with Helm CLI
+  servers:
+  - port:
+      number: 80
+      name: http
+      protocol: HTTP
+    hosts:
+    - "httpbin.example.com"
+EOF
+{{< /text >}}
+
+{{< /tip >}}
+
 Configure routes for traffic entering via the `Gateway`:
 
 {{< text bash >}}
--- a/content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
@@ -67,8 +67,10 @@ kind: Gateway
 metadata:
   name: httpbin-gateway
 spec:
+  # The selector matches the ingress gateway pod labels.
+  # If you installed Istio using Helm following the standard documentation, this would be "istio=ingress"
   selector:
-    istio: ingressgateway # use Istio default gateway implementation
+    istio: ingressgateway
   servers:
   - port:
       number: 80
@@ -78,30 +80,6 @@ spec:
     - "httpbin.example.com"
 EOF
 {{< /text >}}
-
-{{< tip >}}
-If you installed Istio using Helm, the selector for the `Gateway` is `istio=ingress`:
-
-{{< text bash >}}
-$ kubectl apply -f - <<EOF
-apiVersion: networking.istio.io/v1alpha3
-kind: Gateway
-metadata:
-  name: httpbin-gateway
-spec:
-  selector:
-    istio: ingress # use Istio gateway installed with Helm CLI
-  servers:
-  - port:
-      number: 80
-      name: http
-      protocol: HTTP
-    hosts:
-    - "httpbin.example.com"
-EOF
-{{< /text >}}
-
-{{< /tip >}}
 
 Configure routes for traffic entering via the `Gateway`:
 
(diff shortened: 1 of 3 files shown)


ericvn:
Since a set of snips got incremented by one, you may have to update the ones in the test.sh to match.

ericvn:
For point of reference, another way around this might have been to put a snippet title in the new text_bash so it would have gotten it's own tag, and the other ones may not have been incremented? I haven't tried the to see if it would have worked.

nshankar13:
@ericvn ok gotcha, should I go back and do that instead?

ericvn:
> @ericvn ok gotcha, should I go back and do that instead?

I don't think you have to, unless we find that that things are still broke and we need to update the test yet again.

kfaseela:
@howardjohn @frankbu : Hope this just needs documentation update as specified in the doc? Or is this something we need to fix in the actual code to make the labels uniform?

nshankar13:
@kfaseela @howardjohn  if needed, I'd be happy to open a PR in the Istio repo to update the code as well!

nshankar13:
@ericvn it's passing the tests now.

content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
```diff
@@ -79,6 +79,30 @@ spec:
 EOF
 {{< /text >}}
 
+{{< tip >}}
+If you installed Istio using Helm, the selector for the `Gateway` is `istio=ingress`:
```
howardjohn, line 83: This is not accurate, and one of the reasons we changed the label in the install doc :-) 

Its just a standard K8s label selector. There is nothing special about it, at all. Users need to know this, and not treat it as a magic value.

The label on the ingress is user defined. Just because our docs examples happens to set it to `istio=ingress` is not super important. A parallel would be a K8s doc deploying a Pod with app=foo, then making every Service doc say "set `selector: app=foo`" - it works for the example, but doesn't teach users what they need to know.

howardjohn: Gateway: arbitrary label selector.
Helm chart: users can set whatever labels they want.

I would be fine with saying "The selector matches pod labels; if you followed the standard helm docs <LINK> this would be `istio=ingress`". Just think we need to emphasize its not a magic value - just a label selector.

nshankar13: Ah ok yeah I see what you mean. Yeah my initial phrasing was poorly worded.

nshankar13:
@ericvn are we good to merge this?

kfaseela:
> /retest

@nshankar13 it is better rephrase as John suggested

nshankar13:
> @nshankar13 it is better to rephrase as John suggested

Right meant to do this but I forgot.

content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
```diff
@@ -79,6 +79,30 @@ spec:
 EOF
 {{< /text >}}
 
+{{< tip >}}
```
kfaseela, line 82: did you mean to have the <bash> section also inside the tip?

kfaseela: Take a look at the preview here, it is not looking that great to have the tip like this: https://deploy-preview-12486--preliminary-istio.netlify.app/latest/docs/tasks/traffic-management/ingress/ingress-control/ I would rather say you just add an additional comment to the existing yaml at line 71 instead of duplicating the whole stuff.

kfaseela:
@nshankar13 : you can check here how to run some of these tests locally, it saves your time: https://github.com/istio/istio.io/blob/master/tests/README.md#test-authoring-overview

kfaseela:
Please take a look at the rendered page by clicking on deploy/netlify job link in the jobs run.
You can also test the same locally by running "make serve"

nshankar13:
@kfaseela sorry for all the pushes - a few times I ran `make gen` but forgot to run the `lint` step, and then `lint` but forgot to run `make gen`.

content/en/docs/tasks/traffic-management/ingress/ingress-control/index.md
```diff
@@ -67,8 +67,10 @@ kind: Gateway
 metadata:
   name: httpbin-gateway
 spec:
+  # The selector matches the ingress gateway pod labels.
+  # If you installed Istio using Helm following the standard documentation, this would be "istio=ingress."
```
kfaseela, line 71:
```suggestion
  # If you installed Istio using Helm following the standard documentation, this would be "istio=ingress"
```

kfaseela (approved):

nshankar13:
@kfaseela are these CI failures intermittent? The last few runs it passed, but all I changed now was just removing the period at the end of the sentence and some of them are failing.

ericvn:
> @kfaseela are these CI failures intermittent? The last few runs it passed, but all I changed now was just removing the period at the end of the sentence and some of them are failing.

Yes there are. I have been chasing the flakiest test and should have a PR soon for that one.

Issue #12056, opened 2022-10-06
Doc test framework _wait_for_istio function doesn't work any more


frankbu:
Every call to the `_wait_for_istio` is failing lately and printing messages in the log like this:

```
Error: invalid character 'P' looking for beginning of value
Failed distribution of virtualservice httpbin in namespace default
NAME                                                    CLUSTER        CDS        LDS        EDS        RDS          ECDS         ISTIOD                      VERSION
httpbin-74fb669cc6-mvqr2.default                        Kubernetes     SYNCED     SYNCED     SYNCED     SYNCED       NOT SENT     istiod-5d5f6bd4c8-t9tb6     1.16-alpha.f8b4dc7bccc1fd2044c6014aea29368d46f086cc
istio-eastwestgateway-6d47fc4ddf-h7hwh.istio-system     Kubernetes     SYNCED     SYNCED     SYNCED     NOT SENT     NOT SENT     istiod-5d5f6bd4c8-t9tb6     1.16-alpha.f8b4dc7bccc1fd2044c6014aea29368d46f086cc
istio-egressgateway-5df9b4644c-bbsfn.istio-system       Kubernetes     SYNCED     SYNCED     SYNCED     NOT SENT     NOT SENT     istiod-5d5f6bd4c8-t9tb6     1.16-alpha.f8b4dc7bccc1fd2044c6014aea29368d46f086cc
istio-ingressgateway-79598f8d56-jqvfc.istio-system      Kubernetes     SYNCED     SYNCED     SYNCED     SYNCED       NOT SENT     istiod-5d5f6bd4c8-t9tb6     1.16-alpha.f8b4dc7bccc1fd2044c6014aea29368d46f086cc
TEST: wait for failed, but continuing.
Duration: 1 seconds
```

The tests continue to run after these failures and pass because the config is ready, just the function isn't working.

Maybe we can now rewrite these functions using k8s wait instead of the experimental istioctl wait?

These messages appear in the log files of every test run, of every PR.
Example: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/12051/doc.test.profile_default_istio.io/1578018304595857408/artifacts/tests-setup-profile-default-f42/TestDocs/tasks/traffic-management/ingress/ingress-control/test.sh/test.sh/_test_context/test.sh_output.txt

[...]

PR #12508, opened 2023-01-23
Update values so istioctl x wait works

ericvn:

Fixes https://github.com/istio/istio.io/issues/12056

Update some values required for `istioctl x wait`

### DIFF
--- a/content/en/docs/ops/configuration/mesh/config-resource-ready/snips.sh
+++ b/content/en/docs/ops/configuration/mesh/config-resource-ready/snips.sh
@@ -21,7 +21,7 @@
 ####################################################################################################
 
 snip_install_with_enable_status() {
-istioctl install --set PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING --set values.pilot.env.PILOT_ENABLE_STATUS=true --set values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING=true --set values.global.istiod.enableAnalysis=true
+istioctl install --set values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING=true --set values.pilot.env.PILOT_ENABLE_STATUS=true --set values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING=true --set values.global.istiod.enableAnalysis=true
 }
 
 snip_apply_and_wait_for_httpbin_vs() {
--- a/content/en/docs/ops/configuration/mesh/config-resource-ready/snips.sh
+++ b/content/en/docs/ops/configuration/mesh/config-resource-ready/snips.sh
@@ -21,7 +21,7 @@
 ####################################################################################################
 
 snip_install_with_enable_status() {
-istioctl install --set values.pilot.env.PILOT_ENABLE_STATUS=true --set values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING=true --set values.global.istiod.enableAnalysis=true
+istioctl install --set PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING --set values.pilot.env.PILOT_ENABLE_STATUS=true --set values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING=true --set values.global.istiod.enableAnalysis=true
 }
 
 snip_apply_and_wait_for_httpbin_vs() {
--- a/tests/setup/profile_default/doc_test.go
+++ b/tests/setup/profile_default/doc_test.go
@@ -41,10 +41,6 @@ func setupConfig(ctx resource.Context, cfg *istio.Config) {
 values:
   pilot:
     env:
-      PILOT_ENABLE_STATUS: true
       PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING: true
-  global:
-    istiod:
-      enableAnalysis: true
 `
 }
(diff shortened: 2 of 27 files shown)


tests/setup/profile_default/doc_test.go
```diff
@@ -34,3 +35,16 @@ func TestDocs(t *testing.T) {
...
 		Run(istioio.NewTestDocsFunc("profile=default"))
 }
+
+func setupConfig(ctx resource.Context, cfg *istio.Config) {
+	cfg.ControlPlaneValues = `
+values:
+  pilot:
+    env:
+      PILOT_ENABLE_STATUS: true
+      PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING: true
```
howardjohn, line 45: I think only this 1 is needed, the rest enables a bunch of experimental stuff

frankbu:
This looks like it's only fixing `wait_for_istio` calls in the `default` and `demo` profile tests. It will still fail in tests using `profile=none`.

tests/setup/profile_demo/doc_test.go
```diff
@@ -37,5 +37,11 @@ func TestDocs(t *testing.T) {
 }
 
 func setupConfig(ctx resource.Context, cfg *istio.Config) {
-	cfg.ControlPlaneValues = "profile: demo"
+	cfg.ControlPlaneValues = `
+profile: demo
+values:
+  pilot:
+    env:
+      PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING: true
```
kfaseela, line 45: Should we do the same for profile_none?

ericvn: Profile none doesn't install istio, so nothing to add there. We would have to add it to all the tests when they install it to all tests that install istio if the tests use `istioctl x wait`.

kfaseela (approved):

tests/setup/profile_demo/doc_test.go
ericvn: I did a quick search and the following tests look to be using _wait_for_istio_ and profile=none:
- observability/gateways
- observability/metrics/customize-metrics
- ops/config/traffic-management/network-topologies
- setup/getting-started
- traffic-management/egress/wildcard-egress
- traffic-management/egress/egress-control

frankbu: Seems kind of error prone to do this manually. I wonder if we can come up with a way for the framework to handle it automatically (e.g., change the default value of the config parameter)?

frankbu:
@ericvn instead of just merging a partial fix, why not fix it properly? How about updating the snippet generator to check for any snippet that generates `istioctl install ...` to add an extra set argument:
```
istioctl install --set PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING ...
```

ericvn:
I'm not sure that it makes sense to simply change all instances of `istio install` to add the new value, but here is a commit that does do that.

content/en/docs/ops/configuration/mesh/config-resource-ready/snips.sh
```diff
@@ -21,7 +21,7 @@
 ####################################################################################################
 
 snip_install_with_enable_status() {
-istioctl install --set values.pilot.env.PILOT_ENABLE_STATUS=true --set values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING=true --set values.global.istiod.enableAnalysis=true
+istioctl install --set values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING=true --set values.pilot.env.PILOT_ENABLE_STATUS=true --set values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING=true --set values.global.istiod.enableAnalysis=true
```
frankbu, line 24: This isn't ideal, since the test is already setting it. Since they both are set to true, it still works. You could make the snip generator a little smarter, by not doing the replace if the command already sets values.pilot.env.PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING, but it's not really necessary.

frankbu:
> I'm not sure that it makes sense to simply change all instances of istio install to add the new value

Why doesn't it make sense? Seems like a good way to fix it to me. With this addition, you can also put back the Fixes #12056 comment.

content/en/docs/ops/configuration/mesh/config-resource-ready/snips.sh
ericvn: I did try and code this up with anther match, but it didn't work on the first pass. And since it was setting it to true, we should be OK as mentioned.

ericvn:
> > I'm not sure that it makes sense to simply change all instances of istio install to add the new value
> 
> Why doesn't it make sense? Seems like a good way to fix it to me. With this addition, you can also put back the Fixes #12056 comment.

At first, it seems excessive, but when comparing to the demo and default profiles where it's done for all, it should be fine.

And fixes added.

ericvn:
I did try and not update the istio installs for the demo and default tests, but we are processing the .md file and it's the test scripts that contain the profile.

frankbu:
> I did try and not update the istio installs for the demo and default tests, but we are processing the .md file and it's the test scripts that contain the profile.

Not sure I follow. The demo and default tests should not be calling istioctl install, ever. If they did, they would be corrupting the cluster. So the way you handled demo and default is correct. Only none tests should be affected by the snip replace you now added.

frankbu (approved):

ericvn:
/retest 

tasks/traffic-management/tcp-traffic-shifting/gtwapi_test.sh/gtwapi_test.sh

Issue #12298, opened 2022-11-25
istio-dashboard test times out occassionally


kfaseela:
Reference: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/12295/doc.test.profile_default_istio.io/1595885674286813184

kfaseela:
Latest one: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/12496/doc.test.profile_default_istio.io/1617553037616746496/artifacts/tests-setup-profile-default-2ab/TestDocs/tasks/observability/metrics/using-istio-dashboard/test.sh/test.sh/_test_context/test.sh_debug.txt

Looking at the logs, the curl command is failing occassionally. @zirain thoughts?

```
++query_request_for_all_workloads(): curl -L -s 'http://localhost:3000/api/datasources/proxy/1/api/v1/query?query=label_join(sum(rate(istio_requests_total%7Breporter%3D%22destination%22%2C%20response_code%3D%22200%22%7D%5B1m%5D))%20by%20(destination_workload%2C%20destination_workload_namespace%2C%20destination_service)%2C%20%22destination_workload_var%22%2C%20%22.%22%2C%20%22destination_workload%22%2C%20%22destination_workload_namespace%22)'
++query_request_for_all_workloads(): jq -r '.data.result[].metric.destination_workload_var'
++query_request_for_all_workloads(): sort
+__verify_with_retry(): out=
+__verify_with_retry(): local funcret=0
++__verify_with_retry(): sed 's/[[:space:]]*$//g'
```

ericvn:
Yep. This is noted in the other test-flakes PR: https://github.com/istio/istio.io/issues/12464#issuecomment-1386081634. Wasn't as flakey (1/2 the tcpshift flake)

Issue #12464, opened 2023-01-14
doc.test.profile_default_istio.io test flaky


zirain:
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/12459/doc.test.profile_default_istio.io/1614012563471732736
https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/12455/doc.test.profile_default_istio.io/1613815126287192064

ericvn:
From the one job I looked at, it seems the job is taking to long so the whole test passes the list and fails:

```=== RUN   TestDocs/tasks/traffic-management/tcp-traffic-shifting/gtwapi_test.sh/gtwapi_test.sh
2023-01-14T02:14:12.430092Z	info	tf	=== BEGIN: Test: 'tests_setup_profile_default[TestDocs/tasks/traffic-management/tcp-traffic-shifting/gtwapi_test.sh/gtwapi_test.sh]' ===
2023-01-14T02:14:12.430153Z	info	tf	Running command script gtwapi_test.sh
panic: test timed out after 1h0m0s
```

ericvn:
debug.out has:
```
+++__verify_with_retry(): snip_apply_weightbased_tcp_routing_4
+++snip_apply_weightbased_tcp_routing_4(): for i in {1..20}
++++snip_apply_weightbased_tcp_routing_4(): kubectl get pod -l app=sleep -n istio-io-tcp-traffic-shifting -o 'jsonpath={.items..metadata.name}'
+++snip_apply_weightbased_tcp_routing_4(): kubectl exec sleep-bc9998558-6wbts -c sleep -n istio-io-tcp-traffic-shifting -- sh -c '(date; sleep 1) | nc 172.18.7.112 31400'
+++snip_apply_weightbased_tcp_routing_4(): for i in {1..20}
++++snip_apply_weightbased_tcp_routing_4(): kubectl get pod -l app=sleep -n istio-io-tcp-traffic-shifting -o 'jsonpath={.items..metadata.name}'
+++snip_apply_weightbased_tcp_routing_4(): kubectl exec '' -c sleep -n istio-io-tcp-traffic-shifting -- sh -c '(date; sleep 1) | nc 172.18.7.112 31400'
+++snip_apply_weightbased_tcp_routing_4(): for i in {1..20}
[...]

PR #12515, opened 2023-01-24
Test istio-dashboard fix of generating additional trace data

ericvn:

Fixes https://github.com/istio/istio.io/issues/12298

Related to https://github.com/istio/istio.io/issues/12464.

Is the dashboard test simply failing because some of the trace data is missing due to timing? 

See if generating trace data within the loop of testing for data helps at all.

### DIFF
--- a/content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
+++ b/content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
@@ -63,6 +63,7 @@ function query_request_count_to_productpage() {
 
 # This Prometheus query is from Mesh dashboard
 function query_request_for_all_workloads() {
+  bpsnip_trace_generation__1
   curl -L -s 'http://localhost:3000/api/datasources/proxy/1/api/v1/query?query=label_join(sum(rate(istio_requests_total%7Breporter%3D%22destination%22%2C%20response_code%3D%22200%22%7D%5B1m%5D))%20by%20(destination_workload%2C%20destination_workload_namespace%2C%20destination_service)%2C%20%22destination_workload_var%22%2C%20%22.%22%2C%20%22destination_workload%22%2C%20%22destination_workload_namespace%22)' \
   | jq -r '.data.result[].metric.destination_workload_var' | sort
 }


ericvn:

Run test again.

ericvn:

prior-run istio-dashboard passed, failed in /traffic-management/tcp-traffic-shifting/gtwapi_test.sh/gtwapi_test.sh

ericvn:
```
2023-01-25T03:28:40.791085Z	info	tf	Created mesh snapshot file /logs/artifacts/tests-setup-profile-default-e37/TestDocs/tasks/traffic-management/mirroring/gtwapi_test.sh/after_snapshot/_test_context/after snapshot.json
    snapshot_validator.go:105: timeout while waiting after 36 attempts (last error: snapshots are different: 
        @@ -15,16 +15,9 @@
                   "namespace": "default",
                   "services": [
        -            "httpbin-gateway",
                     "kubernetes"
                   ],
        -          "deployments": [
        -            "httpbin-gateway"
        -          ],
        -          "pods": [
        -            "httpbin-gateway-569d7d5dcb-87dsp"
        -          ],
        -          "replicaSets": [
        -            "httpbin-gateway-569d7d5dcb"
        -          ],
        +          "deployments": null,
        +          "pods": null,
        +          "replicaSets": null,
                   "daemonSets": null,
                   "istioOperators": [],
        )
2023-01-25T03:28:40.791486Z	info	tf	=== DONE (failed):  Test: 'tests_setup_profile_default[TestDocs/tasks/traffic-management/mirroring/gtwapi_test.sh/after_snapshot] (2m1.768351737s)' ===
```

ericvn:
tasks/traffic-management/tcp-traffic-shifting/gtwapi_test.sh/gtwapi_test.sh failed this time

ericvn:
istio-dashboard failed twice overnight in the post-submits but no failure here, so if this last run doesn't fail on istio-dashboard, I think this is a good fix.

I could possibly remove the earlier line so we aren't double running before the first check.

content/en/docs/tasks/observability/metrics/using-istio-dashboard/test.sh
```diff
@@ -63,6 +63,7 @@ function query_request_count_to_productpage() {
 
 # This Prometheus query is from Mesh dashboard
 function query_request_for_all_workloads() {
+  bpsnip_trace_generation__1
```
frankbu, line 66: This fix may work, but seems kind of kludgy.

There are 2 delays in this test that may cause the timing flake.

1. The delay after deploying the helloworld services before they are actually ready to handle requests and properly log them. This might be the problem without your fix, because the bsnip_trace_gneration__1 is called very soon after deploying the services.
2. The delay between calling the services and the actual metrics propogating to the logs. This is less likely the problem with the original code (without your fix), because the the test retries every second for up to 2 minutes.

Now with your fix, you're resending the requests, over and over, every second, but then checking the logs immediately after that. So there's 0 chance that the metrics will be there for the current requests, but it's relying on some previous requests finally coming through ... so that the kludgy fix works.

To do this properly, it would be better write some code to first send requests once and then wait/retry for the 2 mins (or maybe override to just 1 min would be better) for them to come through and if they don't, then retry the requests again. A new verify function that does something like that (given two functions), would be a nice way to do it, which could come in handy for some other tests as well.

ericvn: @frankbu Well, that's exactly what this is doing, running the trace steps, and then verifying they are there, and repeating until it happens (or the test fails). I could add a small delay to let the trace propagate, or it will happen after the failure.

The reason the prior test fails is the trace happens, some things propagate, and then re wait forever for all the trace to appear which it never will since we aren't sending new trace to propagate. We need to send more trace.

ericvn: We could throw some sleeps in the code, but that won't help if trace hasn't propagated. We just need to keep sending data, and verify that it propagates at some point.

frankbu: Yes, we're saying the same thing. My point is that the "proper" way to do it is to have two levels of retry, instead of just adding sleeps or otherwise just keep bombarding it with new requests until it it eventually sees what it's looking for :)

ericvn: Note that the current code simply keeps checking if they come through and it fails when they don't. That's the flake. We do need to send more trace to try and get them through, and this code seems to imply that works. I agree we could cycle a few times without trace and only send trace occasionally. I could code this up to send every nth time through the loop.

frankbu: Yeah, something like that would be better.

frankbu (approved):
Merging since this seems to help, but it would be good to revisit with a better solution in the future.

kfaseela (approved):

ericvn:
Likely going to flake on tasks/traffic-management/tcp-traffic-shifting/gtwapi_test.sh/gtwapi_test.sh this time.

PR #12520, opened 2023-01-25
WIP - Try moving gateway tests in default to it's own group

ericvn:

Are the test flakes solvable by moving the gateway tests, in the default profile) to their own group.

First need to verify that simply adding to the name removes them from default, or I may need to specify something else like gateway.

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/ingress-control/gtwapi_test.sh
+++ b/content/en/docs/tasks/traffic-management/ingress/ingress-control/gtwapi_test.sh
@@ -18,7 +18,7 @@
 source "tests/util/gateway-api.sh"
 install_gateway_api_crds
 
-# @setup profile=default
+# @setup profile=default-gateway
 source "content/en/docs/tasks/traffic-management/ingress/ingress-control/test.sh"
 
 # TODO fix cleanup approach and remove this temporary hack
--- a/content/en/docs/tasks/traffic-management/ingress/secure-ingress/gtwapi_test.sh
+++ b/content/en/docs/tasks/traffic-management/ingress/secure-ingress/gtwapi_test.sh
@@ -18,7 +18,7 @@
 source "tests/util/gateway-api.sh"
 install_gateway_api_crds
 
-# @setup profile=default
+# @setup profile=default-gateway
 source "content/en/docs/tasks/traffic-management/ingress/secure-ingress/test.sh"
 
 # TODO fix cleanup approach and remove this temporary hack
--- a/content/en/docs/tasks/traffic-management/mirroring/gtwapi_test.sh
+++ b/content/en/docs/tasks/traffic-management/mirroring/gtwapi_test.sh
@@ -18,7 +18,7 @@
 source "tests/util/gateway-api.sh"
 install_gateway_api_crds
 
-# @setup profile=default
+# @setup profile=default-gateway
 source "content/en/docs/tasks/traffic-management/mirroring/test.sh"
 
 # @cleanup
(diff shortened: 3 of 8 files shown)


ericvn:

dashboard hung out

frankbu:
This is a good experiment but I don't think we want to change the programming API that test writers use, just to control the organization of the test in the build system. Users know what profile the doc is using (e.g. default), so asking them to use profile=default-gateway instead of profile=default for gateway-api tests is an unnecessary complication. Note that there are also gateway-api tests that run with profile=none and profile=demo, but I don't think we want to add profile=none-gateway or profile=demo-gateway, so even more confusing for test writers.

ericvn:
Agree with the other gtwapi tests in the other profiles. Since they aren't flakey, I didn't try to spin them out into their own.

I don't know that there aren't many test writers, especially of gateway tests, right now, so this is a minimal impact to them.

Again, this is something to try and get some of the flakiness under control in the short-term. We have the issue for the work to limit the testing, and that's a great long term solution, but it's been out there for a while and no ones gotten the time todo that properly.

ericvn:

Lets kick default a few times to see if it flies elsewhere.

ericvn:
Looks like moving the Gateway API tests to their own group is still going to flake on tasks/traffic-management/tcp-traffic-shifting/gtwapi_test.sh/gtwapi_test.sh.

ericvn:

Lets kick default a few times to see if it flakes elsewhere.

ericvn:
/test doc.test.profile_default_gateway

Second test passed. Trying a third time.

frankbu:
> Again, this is something to try and get some of the flakiness under control in the short-term.

Short or long term we should not corrupt the programming model like this. If we need to do something to separate the tests, we can design a proper way without forcing the tests to explicitly be hacked like this.

That said, it seems like this experiment shows that it doesn't eliminate the flakes, so we won't need to do something like this anyway.

ericvn:
/test doc.test.profile_default_gateway

Again.

ericvn:
I agree that it doesn't seem that the flakes will be eliminated by moving them out of the other group and will have to see about another possible fix.

Not sure if we want this PR as something for further testing or not. At least it lets us run a subset of the tests.

ericvn:
After this run, I will bump the test memory from 3Gi to 8Gi to see if that will help.

ericvn:

After this run, I will bump the test memory from 3Gi to 8Gi to see if that will help.

ericvn:
As a status, I had 2 failures in 5 runs yesterday with the normal memory specified.

ericvn:
/test doc.test.profile_default_gateway

We've had 3 successes so far after increasing the memory.

ericvn:
/test doc.test.profile_default_gateway

We've had 5 success and no failures toady. One more success and I honk we try moving the default test to maybe 4Gi from 3Gi to see if that helps.

ericvn:
Shucks, 1 failure at 4Gi

ericvn:
I'm still seeing failures at 4Gi. Will try one debug PR to see if it helps identify things. Else, we could try going to 6Gi.

ericvn:
Unfortunately it still failed once on 6Gi: https://prow.istio.io/view/gs/istio-prow/logs/doc.test.profile_default_istio.io_postsubmit/1619033218516783104

ericvn:
Also this one: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/12520/doc.test.profile_default_gateway_istio.io/1619057140574457856

ericvn:
Closing this PR as the experiment didn't pan out. We are still running with 6Gi and will watch to see if we should remove that at some point.

PR #12531, opened 2023-01-27
Media resources

craigbox:
We have a hidden page from the old design:

[/about/media-resources](https://istio.io/latest/about/media-resources/)

This PR removes that page and replaces it with a link to the CNCF Artwork repo.

One of the removed files was the OpenGraph image; I've replaced that with a new one in the correct ratio. (Please note the recent redesign changed the spacing of the boat elements, so all old SVG/PNG files are technically incorrect.)

This does not address:

* embroidery instructions, which have the wrong spacing and were already orphaned
* favicons, which we could regenerate if we cared.


craigbox:
preview: https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fdeploy-preview-12531--preliminary-istio.netlify.app%2F

ericvn (approved):

PR #12538, opened 2023-01-27
Update revision tag diagrams

kfaseela:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/boilerplates/revision-tags-middle.md
```diff
@@ -2,8 +2,8 @@
 ---
 The resulting mapping between revisions, tags, and namespaces is as shown below:
 
-{{< image width="70%"
-link="/docs/setup/upgrade/canary/tags.png"
+{{< image
```
ericvn, line 5: I'm not sure I like the 100%, and maybe you can try going back to 70%. Just comparing the same page here vs https://istio.io/latest/docs/setup/upgrade/canary/#stable-revision-labels-experimental for example.

kfaseela: > I'm not sure I like the 100%, and maybe you can try going back to 70%. Just comparing the same page here vs https://istio.io/latest/docs/setup/upgrade/canary/#stable-revision-labels-experimental for example.

70% makes it too small.. I have made it 90% which looks okay to me.

ericvn:
Looks good slightly smaller. We should be ablate delete the old .png files as well.

kfaseela:
> Looks good slightly smaller. We should be ablate delete the old .png files as well.

I initially tried deleting them, but they seemed to be used in many older html files.. so i was not sure if we will end up in similar problem as we ended up when we updated the logo images... so just retained them..

ericvn:
> I initially tried deleting them, but they seem to be used in many older html files.. so i was not sure if we will end up in similar problem as we ended up when we updated the logo images... so just retained them..

As far as I could determine, all the references are to the zh versions of the file or the copies we have within all the archives so I don't think we should see any issues. There is always a prefix to `/docs/setup/upgrade/canary/tags.png`, with the exception of the changed file.

PR #12544, opened 2023-01-27
Flix flakey tcp-traffic-shifting/gateway test

ericvn:
You can see my discussion in remarks added to this PR.

Basically, there is a race condition where the first `nc` command in the loop of 20 causes the kubeconfig to be reset. Once it's reset, any further communication with the cluster won't work (kubeconfig points nowhere) and the test fails. It actually timeouts since we have a bunch of kubectl commands that will be run repeatedly with longer delays which blows past the test timeout.

Even in the case where the test passes, there are a number (my guess I 8 or so) of the nc calls that return an exit code of 1. Since the test only needs on nc to actually happen, and get back the `one`, the test will pass most times.

Adding a delay shows that the test will pass reliably. I expect there certainly could still be an issue since it's a delay instead of an appropriate wait (I couldn't find one that would work with local testing). For the last bit of testing I did use a 30 sec delay, but reducing that to 20s for this PR. I did make this as a TODO if someone can find a better wait that the time delay. I did try things like `kubectl wait -n istio-io-tcp-traffic-shifting --for=condition=ready tcproute tcp-echo --timeout=60s` but that didn't seem to happen.

### DIFF
--- a/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
+++ b/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
@@ -106,7 +106,7 @@ $ export TCP_INGRESS_PORT=$(kubectl get gtw tcp-echo-gateway -n istio-io-tcp-tra
     $ export SLEEP=$(kubectl get pod -l app=sleep -n istio-io-tcp-traffic-shifting -o jsonpath={.items..metadata.name})
     $ for i in {1..20}; do \
     cat /home/.kube/config; \
-    kubectl exec "$SLEEP" -c sleep -n istio-io-tcp-traffic-shifting -- sh -c "echo one"; \
+    kubectl exec "$SLEEP" -c sleep -n istio-io-tcp-traffic-shifting -- sh -c "(date; sleep 1) | nc $INGRESS_HOST $TCP_INGRESS_PORT"; \
     done
     one Mon Nov 12 23:24:57 UTC 2022
     one Mon Nov 12 23:25:00 UTC 2022
--- a/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
+++ b/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
@@ -104,7 +104,7 @@ $ export TCP_INGRESS_PORT=$(kubectl get gtw tcp-echo-gateway -n istio-io-tcp-tra
     {{< text bash >}}
     $ export SLEEP=$(kubectl get pod -l app=sleep -n istio-io-tcp-traffic-shifting -o jsonpath={.items..metadata.name})
     $ for i in {1..20}; do \
-    kubectl exec "$SLEEP" -c sleep -n istio-io-tcp-traffic-shifting -- sh -c "(date; sleep 1) | nc $INGRESS_HOST $TCP_INGRESS_PORT"; \
+    kubectl exec "$SLEEP" -v -c sleep -n istio-io-tcp-traffic-shifting -- sh -c "(date; sleep 1) | nc $INGRESS_HOST $TCP_INGRESS_PORT"; \
     done
     one Mon Nov 12 23:24:57 UTC 2022
     one Mon Nov 12 23:25:00 UTC 2022
--- a/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
+++ b/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
@@ -102,9 +102,12 @@ $ export TCP_INGRESS_PORT=$(kubectl get gtw tcp-echo-gateway -n istio-io-tcp-tra
 3)  Confirm that the `tcp-echo` service is up and running by sending some TCP traffic.
 
     {{< text bash >}}
-    $ export SLEEP=$(kubectl --v=8 get pod -l app=sleep -n istio-io-tcp-traffic-shifting -o jsonpath={.items..metadata.name})
+    $ cat /home/.kube/config
+    $ export SLEEP=$(kubectl get pod -l app=sleep -n istio-io-tcp-traffic-shifting -o jsonpath={.items..metadata.name})
+    $ /home/.kube/config
     $ for i in {1..20}; do \
-    kubectl --v=8 exec "$SLEEP" -c sleep -n istio-io-tcp-traffic-shifting -- sh -c "(date; sleep 1) | nc $INGRESS_HOST $TCP_INGRESS_PORT"; \
+    cat /home/.kube/config; \
+    kubectl exec "$SLEEP" -c sleep -n istio-io-tcp-traffic-shifting -- sh -c "(date; sleep 1) | nc $INGRESS_HOST $TCP_INGRESS_PORT"; \
     done
     one Mon Nov 12 23:24:57 UTC 2022
     one Mon Nov 12 23:25:00 UTC 2022
--- a/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
+++ b/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
@@ -102,9 +102,9 @@ $ export TCP_INGRESS_PORT=$(kubectl get gtw tcp-echo-gateway -n istio-io-tcp-tra
 3)  Confirm that the `tcp-echo` service is up and running by sending some TCP traffic.
 
     {{< text bash >}}
-    $ export SLEEP=$(kubectl get pod -l app=sleep -n istio-io-tcp-traffic-shifting -o jsonpath={.items..metadata.name})
+    $ export SLEEP=$(kubectl --v=8 get pod -l app=sleep -n istio-io-tcp-traffic-shifting -o jsonpath={.items..metadata.name})
     $ for i in {1..20}; do \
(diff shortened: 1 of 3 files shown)


ericvn:
Good verbose log:
```
I0131 05:07:30.380928  172345 round_trippers.go:463] POST https://127.0.0.1:36937/api/v1/namespaces/istio-io-tcp-traffic-shifting/pods/sleep-bc9998558-6nc75/exec?command=sh&command=-c&command=%28date%3B+sleep+1%29+%7C+nc+172.18.7.112+31400&container=sleep&stderr=true&stdout=true
I0131 05:07:30.380950  172345 round_trippers.go:469] Request Headers:
I0131 05:07:30.380960  172345 round_trippers.go:473]     X-Stream-Protocol-Version: v4.channel.k8s.io
I0131 05:07:30.380971  172345 round_trippers.go:473]     X-Stream-Protocol-Version: v3.channel.k8s.io
I0131 05:07:30.380982  172345 round_trippers.go:473]     X-Stream-Protocol-Version: v2.channel.k8s.io
I0131 05:07:30.380993  172345 round_trippers.go:473]     X-Stream-Protocol-Version: channel.k8s.io
I0131 05:07:30.381004  172345 round_trippers.go:473]     User-Agent: kubectl/v1.26.1 (linux/amd64) kubernetes/8f94681
I0131 05:07:30.401740  172345 round_trippers.go:574] Response Status: 101 Switching Protocols in 20 milliseconds
I0131 05:07:30.401819  172345 round_trippers.go:577] Response Headers:
I0131 05:07:30.401829  172345 round_trippers.go:580]     Connection: Upgrade
I0131 05:07:30.401836  172345 round_trippers.go:580]     Upgrade: SPDY/3.1
I0131 05:07:30.401843  172345 round_trippers.go:580]     X-Stream-Protocol-Version: v4.channel.k8s.io
I0131 05:07:30.401853  172345 round_trippers.go:580]     Date: Tue, 31 Jan 2023 05:07:30 GMT
command terminated with exit code 1
```

ericvn:
Bad verbose log:
```
I0131 15:44:07.727829  173161 round_trippers.go:463] POST https://127.0.0.1:40795/api/v1/namespaces/istio-io-tcp-traffic-shifting/pods/sleep-bc9998558-6s2js/exec?command=sh&command=-c&command=%28date%3B+sleep+1%29+%7C+nc+172.18.7.112+31400&container=sleep&stderr=true&stdout=true
I0131 15:44:07.727855  173161 round_trippers.go:469] Request Headers:
I0131 15:44:07.727867  173161 round_trippers.go:473]     User-Agent: kubectl/v1.26.1 (linux/amd64) kubernetes/8f94681
I0131 15:44:07.727878  173161 round_trippers.go:473]     X-Stream-Protocol-Version: v4.channel.k8s.io
I0131 15:44:07.727886  173161 round_trippers.go:473]     X-Stream-Protocol-Version: v3.channel.k8s.io
I0131 15:44:07.727894  173161 round_trippers.go:473]     X-Stream-Protocol-Version: v2.channel.k8s.io
I0131 15:44:07.727909  173161 round_trippers.go:473]     X-Stream-Protocol-Version: channel.k8s.io
I0131 15:44:07.749558  173161 round_trippers.go:574] Response Status: 101 Switching Protocols in 21 milliseconds
I0131 15:44:07.749626  173161 round_trippers.go:577] Response Headers:
I0131 15:44:07.749656  173161 round_trippers.go:580]     Connection: Upgrade
I0131 15:44:07.749694  173161 round_trippers.go:580]     Upgrade: SPDY/3.1
I0131 15:44:07.749717  173161 round_trippers.go:580]     X-Stream-Protocol-Version: v4.channel.k8s.io
I0131 15:44:07.749735  173161 round_trippers.go:580]     Date: Tue, 31 Jan 2023 15:44:07 GMT
I0131 16:14:18.191128  174201 loader.go:373] Config loaded from file:  /home/.kube/config
I0131 16:14:18.191930  174201 round_trippers.go:463] GET http://localhost:8080/api?timeout=32s
I0131 16:14:18.191964  174201 round_trippers.go:469] Request Headers:
I0131 16:14:18.191972  174201 round_trippers.go:473]     Accept: application/json;g=apidiscovery.k8s.io;v=v2beta1;as=APIGroupDiscoveryList,application/json
I0131 16:14:18.191977  174201 round_trippers.go:473]     User-Agent: kubectl/v1.26.1 (linux/amd64) kubernetes/8f94681
I0131 16:14:18.195327  174201 round_trippers.go:574] Response Status:  in 3 milliseconds
I0131 16:14:18.195379  174201 round_trippers.go:577] Response Headers:
E0131 16:14:18.195485  174201 memcache.go:238] couldn'\''t get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
I0131 16:14:18.195511  174201 cached_discovery.go:120] skipped caching discovery info due to Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
I0131 16:14:18.195591  174201 shortcut.go:100] Error loading discovery information: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
```

You will note that we do not see a `command terminated with exit code 1` in this log and it's followed by the created calls to the api/server.

ericvn:
Captured a failure and it looks like the kubeconfig in the failed case is becoming empty:
```
apiVersion: v1
kind: Config
preferences: {}
```

ericvn:
So I think something happens during the `nc` call which causes the failure @frankbu. It's always during the `nc`, so I don't think it's something external causing the cluster to die. After we return from the `nc` the kubeconfig is basically empty. Did the `nc` cause our kubecluster to die?  And it is always the first `nc`, and not on a later one in the loop.

I wonder if we can change the code to use something else.

ericvn:
Removing the `nc` sets to have reduced the failure as I've run the test 6 times with no failure (7th time is running now). I have been trying to determine a way to verify that things are set up, but `nc` is not part of the test container. It might be possible to curl, but I'm not sure that INGRESS HOST is correct for doing that (as it is with the sleep container).

ericvn:
Adding a delay before the first nc call (putting the call back in). Normally I was seeing about 8 failures before nc worked so with a 1 sec delay we should wait at least 8. Going with 30 secs.

ericvn:

Failed on the other falsely test: tasks/traffic-management/mirroring/gtwapi_test.sh/gtwapi_test.sh

ericvn:
1st run with the 30 sec did not show any of the initial return code of 1 errors

ericvn:
I wonder if the [route status](https://gateway-api.sigs.k8s.io/api-types/httproute/#status) would be sufficient? Or is there something else needed?

ericvn:

A few more runs, but the delay seems to be needed. Not sure if 30 secs is the correct value. I've tried a few kubctl wait or _wait_for_istio commands running test locally and haven't found anything that works.

ericvn:
@frankb @kfaseela Here is a fix for the flakiest test. I don't like adding a deli, but couldn't find a wait that worked. Left a TODO to revise at some point.

content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
```diff
@@ -102,8 +102,10 @@ $ export TCP_INGRESS_PORT=$(kubectl get gtw tcp-echo-gateway -n istio-io-tcp-tra
 3)  Confirm that the `tcp-echo` service is up and running by sending some TCP traffic.
 
     {{< text bash >}}
+    $ cat /home/.kube/config
```
frankbu, line 105: Remove these cat commands before merging.

content/en/docs/tasks/traffic-management/tcp-traffic-shifting/test.sh
```diff
@@ -42,6 +42,11 @@ if [ "$GATEWAY_API" == "true" ]; then
     snip_apply_weightbased_tcp_routing_2
     _wait_for_gateway istio-io-tcp-traffic-shifting tcp-echo-gateway
     snip_apply_weightbased_tcp_routing_3
+
+    # Make sure the nc command will work and not fail the cluster
+    # TODO proper wait for things being ready
+    # it seems we had 8 or so exits during the run of 20 so need 8 plus seconds
+    sleep 20s
```
frankbu, line 49: I wish we could understand what exactly nc does that corrupts the kubeconfig, and why it never happens when using Istio Gateway/VirtualService instead of gateway-api??

content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
ericvn: @frankbu Removed

content/en/docs/tasks/traffic-management/tcp-traffic-shifting/test.sh
```diff
@@ -42,6 +42,11 @@ if [ "$GATEWAY_API" == "true" ]; then
     snip_apply_weightbased_tcp_routing_2
     _wait_for_gateway istio-io-tcp-traffic-shifting tcp-echo-gateway
     snip_apply_weightbased_tcp_routing_3
+
+    # Make sure the nc command will work and not fail the cluster
```
frankbu, line 46:
```suggestion
    # Make sure the nc command will work and not fail the cluster
    # More info: https://github.com/istio/istio.io/pull/12544
```

frankbu (approved):

kfaseela:
thanks @ericvn !

PR #12564, opened 2023-01-31
Optimize doc testing based on PR changes

frankbu:

First step in optimizing the the doc testing to not run the entire suite of tests, with flakes, etc., for every doc PR. Most times we only need to run 1 test or no tests at all.

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


scripts/pr_tests.py
```diff
@@ -0,0 +1,69 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import sys
+from github import Github
```
frankbu, line 19: This dependency will need to be added in the build container. Maybe it would be better to try to do this without using pygithub?

@howardjohn, WDYT?

zirain:
changes like this https://github.com/istio/istio.io/pull/12566 should run a full test

ericvn:
One **BIG BIG** problem with the current regex put into the prow jobs, is that the test jobs are no longer required so that an approval will let the istio PR merge effectively breaking the test.

howardjohn:
@ericvn unless its configured incorrectly, github says "Not Required" since it doesn't know about the regex logic, but prow will not actually merge them

ericvn:
> @ericvn unless its configured incorrectly, github says "Not Required" since it doesn't know about the regex logic, but prow will not actually merge them

Thanks John. I was hoping that it would work that way. Just running a test to verify as well.

ericvn:
Seems to be working as hoped: https://github.com/istio/istio.io/pull/12571#issuecomment-1412485466

zirain:
> > changes like this #12566 should run a full test
> 
> @zirain Why should updating the reference docs cause all the tests to run? We don't use those changes within any of the tests.

link a wrong PR, should be https://github.com/istio/istio.io/pull/12564#issuecomment-1412281755

scripts/pr_tests.py
```diff
@@ -0,0 +1,69 @@
...
+    commits = pr.get_commits()
+    for commit in commits:
+        files = commit.files
+        for file in files:
+            filename = file.filename
+            if filename.startswith(doc_file_prefix):
+                if filename.endswith("test.sh") or filename.endswith("/snips.sh"):
+                    relative_file = filename[len(doc_file_prefix):]
+                    test_paths.add(relative_file.rsplit('/', 1)[0])
+            elif filename == istio_go_dependency or \
```
ericvn, line 54: @zirain This should cause all tests to run.

frankbu:
We are currently running this script without passing an access token for security reasons. Without an access token, the script will be subjected to github rate-limiting of 60 req/hr. The script will fall back to running all tests when this happens, so the question is how valuable will this optimization be?

Let's merge this now as-is and see how well it works. If it turns out to be relatively ineffective, we can either remove it entirely (we still get all-or-nothing optimization from prow regex), or we can potentially move this to istio/test-infra and then pass in an access token.

ericvn:
Is there anywhere where you enabled, or is there a way to enable the script output? I see in the build log:
```
Optimizing tests for pull nummber: 12564
++ python3 ./scripts/pr_tests.py --token= 12564
+ TESTS=ALL
```
Just wanting to verify that ALL is being returned here do to a failure or the second file in the prow directory.

frankbu:
> Just wanting to verify that ALL is being returned here do to a failure or the second file in the prow directory.

If it's ALL because of a failure, there will be another message in the log:
```
print("Exception: %s" % str(e), file=sys.stderr)
```

ericvn (approved):
Thanks Frank. Will watch and see how the testing goes. I suspect a short passing test in the Job History for a PR with changes that should cause tests would indicate a problem.

PR #12569, opened 2023-02-01
Add minimal profile for docs test setups

kfaseela:
Signed-off-by: Faseela K <faseela.k@est.tech>

Have been using minimal profile in some of the tests, so i think it makes sense to add this, and then migrate those tests to use this new setup.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

- [ ] Configuration Infrastructure
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

PR #12576, opened 2023-02-02
Migrate ingress-sni-passthrough test to minimal profile setup

kfaseela:

Make use of https://github.com/istio/istio.io/pull/12569

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
+++ b/content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
@@ -18,8 +18,7 @@
 source "tests/util/gateway-api.sh"
 install_gateway_api_crds
 
-# @setup profile=none
-istioctl install --set profile=minimal -y
+# @setup profile=minimal
 source "content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/test.sh"
 
 # @cleanup
--- a/content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
+++ b/content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
@@ -25,5 +25,4 @@ source "content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough
 snip_cleanup_2
 snip_cleanup_3
 snip_cleanup_4
-kubectl delete ns istio-system
 remove_gateway_api_crds
--- a/content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
+++ b/content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
@@ -25,6 +25,5 @@ source "content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough
 snip_cleanup_2
 snip_cleanup_3
 snip_cleanup_4
-istioctl uninstall --purge -y
 kubectl delete ns istio-system
 remove_gateway_api_crds
--- a/tests/setup/profile_minimal/doc_test.go
+++ b/tests/setup/profile_minimal/doc_test.go
@@ -43,5 +43,13 @@ values:
   pilot:
     env:
       PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING: true
+components:
+  egressGateways:
+  - enabled: false
+    name: istio-egressgateway
+  ingressGateways:
+  - enabled: false
+    name: istio-ingressgateway
 `
+	cfg.DeployEastWestGW = false
 }
--- a/tests/setup/profile_minimal/doc_test.go
+++ b/tests/setup/profile_minimal/doc_test.go
@@ -37,8 +37,9 @@ func TestDocs(t *testing.T) {
 }
 
 func setupConfig(ctx resource.Context, cfg *istio.Config) {
+	// FIXME: test framework does not honor profile=minimal config at present,
+	// hence we have to explicitly disable the gateways.
 	cfg.ControlPlaneValues = `
-profile: minimal
 values:
   pilot:
     env:


frankbu:
We can't use a new profile before we set up a new prow job for the minimal profile, otherwise these tests will not be running anymore.

That said, is this really worth adding another prow job for every build?

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
```diff
@@ -18,8 +18,7 @@
 source "tests/util/gateway-api.sh"
 install_gateway_api_crds
 
-# @setup profile=none
-istioctl install --set profile=minimal -y
```
frankbu, line 22: Also need to delete the uninstall from @cleanup

ericvn:
@frankbu Adding another job will help shorten some jobs as tests are moved from them to here. We have the regex in place to not run tests at all. I'm not sure how this affects your other code in determining the tests to run under each profile (if it matters as I think you can just not specify the profile, but would have to test).

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
```diff
@@ -18,8 +18,7 @@
...
+# @setup profile=minimal
 source "content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/test.sh"
```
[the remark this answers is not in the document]
kfaseela: There were tests outside gateway api as well. But this is not super-urgent.

frankbu:
>  many of the tests don't need the gateway, but are still simply using profile=default.. those can also be modified?

We have to be careful to try to make sure the test is using the profile that the doc says. If we optimize the test by using a "better profile" than the doc is suggesting, we won't be testing that the procedure really works as expected, for example, if having gateways installed might in the future interfere with the proper functioning of the test.

ericvn:
> > many of the tests don't need the gateway, but are still simply using profile=default.. those can also be modified?
> 
> We have to be careful to try to make sure the test is using the profile that the doc says. If we optimize the test by using a "better profile" than the doc is suggesting, we won't be testing that the procedure really works as expected, for example, if having gateways installed might in the future interfere with the proper functioning of the test.

I made this comment earlier this morning in Slack. Our README does say:
```
Choose the setup configuration that best matches the document prerequisites.
```

kfaseela:
> > many of the tests don't need the gateway, but are still simply using profile=default.. those can also be modified?
> 
> We have to be careful to try to make sure the test is using the profile that the doc says. If we optimize the test by using a "better profile" than the doc is suggesting, we won't be testing that the procedure really works as expected, for example, if having gateways installed might in the future interfere with the proper functioning of the test.

Understood.. this one is straight forward, but any other migration, will make sure to change the doc also accordingly...

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh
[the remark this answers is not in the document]
ericvn: Not sure you can specify for the gateway APIs to be installed in an IOP which is what the framework uses to setup istio. I don't know if there is a post install run this in the framework.

tests/setup/profile_minimal/doc_test.go
```diff
@@ -43,5 +43,13 @@ values:
   pilot:
     env:
       PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING: true
+components:
+  egressGateways:
+  - enabled: false
+    name: istio-egressgateway
+  ingressGateways:
+  - enabled: false
+    name: istio-ingressgateway
```
frankbu, line 52: These aren't needed because they are already set that way with profile=minimal.

tests/setup/profile_minimal/doc_test.go
```diff
@@ -43,5 +43,13 @@ values:
...
 `
+	cfg.DeployEastWestGW = false
```
frankbu, line 54: Also not needed since this one should be false by default

frankbu: Why did you add these? Am I missing something?

frankbu: Hmm, maybe there's a bug here: https://github.com/istio/istio/blob/7717d61314a9880fbd0c7742f1622f3070241944/pkg/test/framework/components/istio/kube.go#L422

I think it should also be checking for multicluster:
```
		if !i.cfg.DeployEastWestGW || !i.ctx.Environment().IsMultiCluster() {
```

ericvn: Yes Frank there is an issue with how the framework starts it seems. @kfaseela was going to look into it some more. The framework sets up the cluster using serveral iop files. Two are from the framework and the third is created by what we setup in our files like specific profiles. 

The first few files set the ingress and egressgteway values to true so even though the profile is minimal, the gateways are started. We need to specifically set them to false to prevent them from starting.

I'm not even convinced that we are starting things correctly in the other tests.

ericvn (approved):

tests/setup/profile_minimal/doc_test.go
ericvn: Here's what's being used:
```
	Installing Istio components on cluster cluster-0: InFilenames:      [/work/tests/integration/base.yaml /work/tests/integration/iop-integration-test-defaults.yaml /work/out/darwin_amd64/tests-setup-profile-minimal-d18/_suite_context/istio-deployment-3202578840/iop.yaml]
```
Even though the last yaml has the minimal profile specified, the gateways are installed. This might be an issue with Istio that needs to be fixed.

ericvn: Note that the tests/integration files come from istio/istio.

ericvn: Although the EastWest might not need to be called out specifically to disable.

kfaseela: @frankbu better to fix the framework than doing it this way?

frankbu: I would recommend fixing the framework. The suggested change for DeployEastWestGateway that I mentioned above seems like an obvious omission. I would also think that if the gateways are being installed with profile=minimal (that's probably just another omission in the framework's config file), then we should definitely fix that. Otherwise the framework is not testing what the user intends. Hacking in fixes like this is not a good approach.

kfaseela: Makes sense @frankbu ! @ericvn I will try to fix the framework first, and then get back to this PR

frankbu: It seems that the EastWestGW is also used for VM service testing, not just multicluster. Since the VM integration tests are not cleanly separated like the multicluster ones are, it doesn't look like we can easily change the default test cluster config to not include EastWestGW.

So, I think the best thing to do here is configure the profile_minimal explicitly, like you have done. But, remove `profile=minimal` (line 41) from the configuration, since it's confusing and doesn't work as expected anyway.

kfaseela: @frankbu done

frankbu (approved):

kfaseela:
@ericvn @frankbu : I see ingress-sni-passthrough test being run in both default and minimal tests on this PR.

PR #12608, opened 2023-02-08
Revert "Optimize doc testing based on PR changes"

ericvn:
Reverts istio/istio.io#12564

Looks like the post-submit jobs don't have PULL_NUMBER. Another option is to not run the jobs in post_submit, but not sure we want to give that up.

Example job list: https://prow.istio.io/job-history/gs/istio-prow/logs/doc.test.profile_demo_istio.io_postsubmit


ericvn:
Frank says he will fix the failure, so close this for now.

PR #12620, opened 2023-02-09
dependabot suggested updates

ericvn:

Just adding the dependabot suggested updates as one PR instead of seeing multiples arrive IF we had turned it on.

One item I'm curious about is if the gateway-api update will affect anything. I know it's updated in the main istio/istio branch but not in the `release-1.17` branch. However, comments in Slack seem to imply that it should not matter.


ericvn:
Closing, but may work on later. Linter is failing do to one of the updates.

PR #12639, opened 2023-02-12
fix lint

zirain:
xref: https://github.com/istio/istio.io/pull/12638 https://github.com/istio/istio.io/pull/12637


ericvn (approved):
I'm not sure why this lint warning is cropping up at this point, and that might need some more investigation, but does appear that the warning is holding out other PRs so approving this.  Thanks.

ericvn:
I would add that I'm also not seeing this running locally.

PR #12650, opened 2023-02-13
Termporarily remove the Performance page from the website

ericvn:

Discussion during today's TOC meeting lead to temporarily removing the Performance page since the results are [old](https://github.com/istio/istio.io/issues/11986). This PR renames the page so it is no longer visible. Once new data is generated, we can rename along with added the new data and the page will be back.

[Decision](https://docs.google.com/document/d/13lxJqtlaQhmV2EwsNnS6h-_O4pobZQZuMjrzOeMgVI0/edit#):
- Remove the perf results page, until new perf test results are ready.
- Mariam to work with [Sakari](https://github.com/sakari) to run the perf tests after releasing 1.17.


linsun (approved):

kfaseela:
> let me know if you need help reviving the fortio based continuous tests we used to have

cc @johnma14

howardjohn:
Do we have someone committed to running the tests for 1.17 or is it a "nice to have"?

spadger:
Hi,
Do you have any indication of how far the `0.5cores / 1000 requests` metric is from reality?

LithiumSR:
Hello everyone,
On top of restoring the page for 1.17 I think it would be nice to have the results for the versions that are currently in the support window. What do you think?
From my internal testing I see some performance degradation between Istio 1.15-16 for the traffic going through the ingressgw so it would be nice to compare.

howardjohn:
Generally our docs only include info about one version, but you can access previous version's: https://istio.io/archive/

ericvn:
Problem is that earlier versions 1.9-1.16 point to the data for 1.8 IIRC.

PR #12655, opened 2023-02-14
Nit: modify the path of `pilot-k8s.yaml`

LittleWat:
`pilot-k8s.yaml` is currently located at https://github.com/istio/istio/blob/master/operator/samples/pilot-k8s.yaml

so let me update the doc.

- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/additional-setup/customize-installation/index.md
+++ b/content/en/docs/setup/additional-setup/customize-installation/index.md
@@ -128,7 +128,7 @@ spec:
 Use `istioctl install` to apply the modified settings to the cluster:
 
 {{< text syntax="bash" repo="operator" >}}
-$ istioctl install -f samples/operator/pilot-k8s.yaml
+$ istioctl install -f operator/samples/pilot-k8s.yaml
 {{< /text >}}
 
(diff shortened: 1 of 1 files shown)


ericvn (changes requested):
This is where I see it in the current rc0 build:
```
 ~/work/go/src/istio.io/istio-1.17.0-rc.0  find . -name pilot-k8s.yaml                                                                            
./samples/operator/pilot-k8s.yaml
```

ericvn:
It may make sense to add a note that if you are running against a source branch as opposed to using an install, to use a different path.

LittleWat:
Oh, thank you for your quick reply!

Ah, sorry, I thought the directory structure of `releases`  is the same as that of GitHub 🤦‍♂️ but this was wrong.

> It may make sense to add a note that if you are running against a source branch as opposed to using an install, to use a different path.

Yes, it may make sense, but  I have to add the note to every docs, and this seems to be not so important so let me close this PR. 🙇

Issue #12672, opened 2023-02-15
Enable istioctl-analyze test


ericvn:
The istioctl analyze test was disables part of the release 1.17 branch cutting.

It seemed to have the wrong version (1.17) in the documentationURL (should be 1.18). I'm not sure if this is an issue upstream or in the test.

Investigate and enable the test again.

PR #12675, opened 2023-02-15
Enable istioctl-analyze

ericvn:

Fixes https://github.com/istio/istio.io/issues/12672

So for some reason, the line number change I did yesterday and thought was merged in the last PR never made it in. That's in the first commit.

The error I saw after that running locally is the documentationURL in the warning expected 1.18, which is correct, but was getting 1.17. This would indicate that something was maybe wrong upstream in the version being used or maybe something in istio.io with how we are creating istioctl.


content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -47,8 +47,8 @@ Analyze the current live cluster, simulating the effect of applying additional y
 
 {{< text syntax=bash snip_id=analyze_sample_destrule >}}
 $ istioctl analyze @samples/bookinfo/networking/bookinfo-gateway.yaml@ @samples/bookinfo/networking/destination-rule-all.yaml@
-Error [IST0101] (Gateway default/bookinfo-gateway samples/bookinfo/networking/bookinfo-gateway.yaml:7) Referenced selector not found: "istio=ingressgateway"
-Error [IST0101] (VirtualService default/bookinfo samples/bookinfo/networking/bookinfo-gateway.yaml:39) Referenced host not found: "productpage"
+Error [IST0101] (Gateway default/bookinfo-gateway samples/bookinfo/networking/bookinfo-gateway.yaml:9) Referenced selector not found: "istio=ingressgateway"
```
ericvn, line 50: I do know @frankbu mentioned we may be able to make this better by not looking for specific line numbers. I haven't seen a line number failure here before as I suspect the samples are pretty static.

PR #12760, opened 2023-02-22
Add (blank) ambient stub pages

howardjohn:
For https://github.com/istio/istio/issues/42750#issuecomment-1437387508

Add some blank pages so we can iterate on this with less merge conflicts etc

### DIFF
--- a/content/en/docs/ops/ambient/_index.md
+++ b/content/en/docs/ops/ambient/_index.md
@@ -1,0 +1,7 @@
+---
+title: Ambient Mesh [WIP]
+description: Information for setting up and operating Istio ambient mesh.
+weight: 60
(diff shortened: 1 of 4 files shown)


kfaseela:
leaving to @ericvn for merging. Just wondering is "Operations" the right place to add Ambient Mesh? May be it can be a separate section on its own?

howardjohn:
> leaving to @ericvn for merging. Just wondering is "Operations" the right place to add Ambient Mesh? May be it can be a separate section on its own?

FWIW the reason I put it here was it was the only place that really worked without making a whole new section

ericvn (approved):
Agree that there doesn't look to be a better place than this for now. We can always move things around at a later date.

PR #12775, opened 2023-02-24
Add Ericsson to Istio Ecosystem

kfaseela:


ericvn:
@craigbox I don't believe we do any randomization of ordering of companies. Is there any documentation that talks about ordering within the list?

kfaseela:
> @craigbox I don't believe we do any randomization of ordering of companies. Is there any documentation that talks about ordering within the list?

I did not see any alphabetical order followed atleast in the first page, so should I add to the end of the list? Or any other ordering guidelines?

craigbox:
The ordering is roughly:

- first 8 will show on the front page
- those should include, in order: 
  - as many brand-name Kubernetes distributions as possible
  - companies who have built a business around Istio
- the rest are in order of "noteworthiness"
  - the largest, most well known companies that are using Istio (possibly ordered in terms of contribution)
  - everyone else, in either alphabetical or order of addition

I think an acceptable ordering for row 3 would be Intel, Ericsson, Cisco, SAP.

Could you use an SVG image?  Is [this one](https://upload.wikimedia.org/wikipedia/commons/e/e9/Ericsson_logo.svg) OK?

kfaseela:
> The ordering is roughly:
> 
> * first 8 will show on the front page
> * those should include, in order:
>   
>   * as many brand-name Kubernetes distributions as possible
>   * companies who have built a business around Istio
> * the rest are in order of "noteworthiness"
>   
>   * the largest, most well known companies that are using Istio (possibly ordered in terms of contribution)
>   * everyone else, in either alphabetical or order of addition
> 
> I think an acceptable ordering for row 3 would be Intel, Ericsson, Cisco, SAP.
> 
> Could you use an SVG image? Is [this one](https://upload.wikimedia.org/wikipedia/commons/e/e9/Ericsson_logo.svg) OK?

have changed the order as suggested, and also updated the logo to use .svg

craigbox (approved):

PR #12779, opened 2023-02-24
Add more context to `externalTrafficPolicy=local` caveats.

bleggett:
Spawned by: https://github.com/istio/istio/issues/43397

Existing doc is correct but it confused myself and a few others, so attempting to add more context around why we make this recommendation when things still technically "work" without following it.

Signed-off-by: Benjamin Leggett <benjamin.leggett@solo.io>

- [x] Docs
- [ ] Installation
- [x] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -126,7 +126,7 @@ spec:
...
+If you are using a TCP/UDP network load balancer that preserves the client IP address (AWS Network Load Balancer, GCP External Network Load Balancer, Azure Load Balancer) or you are using Round-Robin DNS, then you can also preserve the client IP inside Kubernetes by bypassing kube-proxy and preventing it from sending traffic to other nodes.  **However, for production deployments it is strongly recommended to deploy an ingress gateway pod on every node** if you enable `externalTrafficPolicy: Local`. Any node that receives NLB ingress traffic and doesn’t have a local ingress gateway to handle it will drop the traffic if `externalTrafficPolicy: Local` is enabled. This creates a situation where **only** nodes with an active ingress gateway pod will be able to accept and distribute incoming NLB traffic to the rest of the cluster, creating potential ingress traffic bottlenecks. See [Source IP for Services with `Type=NodePort`](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport)
```
howardjohn, line 129: LBs health check the nodes though, right? So is the "deploy to all nodes" suggestion so that you don't have downtime while the healthchecks are propogating or something?

bleggett: Yeah lemme reword that, you're correct - the fundamental problem is that with `local` you can only accept LB traffic on the subset of nodes with a gateway, regardless of whether the LB actually tries to send traffic to them or not.

eghobo: sorry, but it's not practical (e.g. several replicas will address SPOF) even suggest to run gateway at every node ):  maybe better to suggest this only if user cannot setup `local`

bleggett: `several replicas will address SPOF` - that would address SPOF of the gateway pod on the node, but not of the node(s), correct?

`it's not practical to even suggest to run gateway at every node` - Agree, it sort of depends on how many nodes you have and what you consider "good enough" for your deployment. The existing docs are making that "every node" claim - if we want to modify that we can, but at the end of the day the answer for "how many" is "it depends on your deployment".

howardjohn: AFAIK you can run on a subset of nodes and just have it deployed to all of them there. The issue is only going from N -> 0 pods on a node.

That may be a GCP specific thing though

eghobo: we usualy deploy Gateways that spread across nodes that address node problem

bleggett: @eghobo removed the "all nodes" guidance in favor of "multiple nodes"

eghobo: @bleggett looks good, maybe "it is strongly recommended to deploy an ingress gateway pod on on different nodes"

bleggett:
@triptesh1212 and @subhankarc please take a look and double-check that  this makes more sense to you than the existing doc.

eghobo (approved):
+1

subhankarc:
@bleggett  The documentation looks ok with respect to our understanding as well. However, we saw an additional scenario where external traffic policy has another challenge with respect to K8s Cluster autoscaling. We tested this on AWS. 

1. If we start with two nodes and only one pod is deployed, for external traffic policy as local, both nodes become part of the target group but one of them having the gateway pod becomes healthy. In the case of Cluster-based external traffic policy, both nodes are healthy. This is also in line with the documentation. 
2. If we have one k8s node and one pod deployed on that node, on the AWS console, the target group shows one node and it is healthy. This is also expected and as per the documentation.
3. With external traffic policy as local, if we autoscale the k8s cluster to two nodes and deploy another pod in the second node, the target group still remains with one node. However, this is correctly updated if the traffic policy is mentioned as cluster. 

Should we also explain and mention this behavior also(point no 3) in the documentation with respect to the autoscaling of the Kubernetes cluster?

bleggett:
> With external traffic policy as local, if we autoscale the k8s cluster to two nodes and deploy another istio-ingressgateway pod in the second node, the target group still remains with one node. However, this is correctly updated if the traffic policy is mentioned as cluster.
> 
> 
> Should we also explain and mention this behavior also(point no 3) in the documentation with respect to the autoscaling of the Kubernetes cluster?

This sounds like a separate (potentially AWS-specific) issue to me. Keeping this PR focused on explaining the general problem rather than walking thru the specifics seems best.

eghobo:
> With external traffic policy as local, if we autoscale the k8s cluster to two nodes and deploy another istio-ingressgateway pod in the second node, the target group still remains with one node. However, this is correctly updated if the traffic policy is mentioned as cluster.

@subhankarc, how long have you waited? usually, all new nodes are registered in the target group after some time.

subhankarc:
> > With external traffic policy as local, if we autoscale the k8s cluster to two nodes and deploy another istio-ingressgateway pod in the second node, the target group still remains with one node. However, this is correctly updated if the traffic policy is mentioned as cluster.
> > Should we also explain and mention this behavior also(point no 3) in the documentation with respect to the autoscaling of the Kubernetes cluster?
> 
> This sounds like a separate (potentially AWS-specific) issue to me. Keeping this PR focused on explaining the general problem rather than walking thru platform specifics seems best.

I briefly tested on Azure as well, seems to have the same issue. I could create a different issue to discuss this autoscaling aspect. Let’s go ahead with this PR.

subhankarc:
> > With external traffic policy as local, if we autoscale the k8s cluster to two nodes and deploy another istio-ingressgateway pod in the second node, the target group still remains with one node. However, this is correctly updated if the traffic policy is mentioned as cluster.
> 
> @subhankarc, how long have you waited? usually, all new nodes are registered in the target group after some time.

I think I waited long enough, about 5
-10 mins but nothing changed. The same behaviour with external traffic policy as cluster adds the new node fast enough, so I guess it could be something to do with the external traffic policy set as local.

eghobo:
@subhankarc, can you try to wait longer? I am 100% confident that the cloud provider will register new nodes eventually ;)

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -126,7 +126,7 @@ spec:
...
+If you are using a TCP/UDP network load balancer that preserves the client IP address (AWS Network Load Balancer, GCP External Network Load Balancer, Azure Load Balancer) or you are using Round-Robin DNS, then you can also preserve the client IP inside Kubernetes by bypassing kube-proxy and preventing it from sending traffic to other nodes.  **However, for production deployments it is strongly recommended to deploy an ingress gateway pod to multiple nodes** if you enable `externalTrafficPolicy: Local`. Otherwise, this creates a situation where **only** nodes with an active ingress gateway pod will be able to accept and distribute incoming NLB traffic to the rest of the cluster, creating potential ingress traffic bottlenecks and reduced internal load balancing capability, or even complete loss of ingress traffic to the cluster if the subset of nodes with ingress gateway pods go down. See [Source IP for Services with `Type=NodePort`](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport)
 for more information. Update the ingress gateway to set `externalTrafficPolicy: Local` to preserve the
```
frankbu, line 130:
```suggestion
If you are using a TCP/UDP network load balancer that preserves the client IP address (AWS Network Load Balancer, GCP External Network Load Balancer, Azure Load Balancer) or you are using Round-Robin DNS, then you can use the `externalTrafficPolicy: Local` setting to also preserve the client IP inside Kubernetes by bypassing kube-proxy and preventing it from sending traffic to other nodes.

{{< warning >}}
For production deployments it is strongly recommended to **deploy an ingress gateway pod to multiple nodes** if you enable `externalTrafficPolicy: Local`. Otherwise, this creates a situation where **only** nodes with an active ingress gateway pod will be able to accept and distribute incoming NLB traffic to the rest of the cluster, creating potential ingress traffic bottlenecks and reduced internal load balancing capability, or even complete loss of ingress traffic to the cluster if the subset of nodes with ingress gateway pods go down. See [Source IP for Services with `Type=NodePort`](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport) for more information.
{{< /warning >}}

 Update the ingress gateway to set `externalTrafficPolicy: Local` to preserve the
```

frankbu (approved):

Issue #12759, opened 2023-02-22
Add cleanup instructions to `istioctl analyze` page


johnma14:
Add cleanup instructions  for [`istioctl analyze`](https://istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze/) page

johnma14:
Have few new contributors who can work on this.

johnma14:
Need to add the cleanup section to: https://github.com/istio/istio.io/blob/master/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md. (Similar to: https://istio.io/latest/docs/tasks/security/authentication/mtls-migration/#clean-up-the-example)

```
kubectl label namespace default istio-injection-
kubectl delete ns frod
kubectl delete deployment my-deployment
kubectl delete vs ratings
istioctl uninstall --purge -y
kubectl delete ns istio-system

hanxiaop:
Do we need clean up section for this page? Since `analyze` is analyzing existing configurations or configuration files, it's fine to use the feature anytime. Unlike 'Mutual TLS Migration,' which has specific configurations to work, analyze won't have such kinds of tasks to show, and doesn't need to apply any demo configs.

johnma14:
@hanxiaop I am closing this issue. Thank you for clarifying.

johnma14:
The `istioctl analyze` page does not install anything specifically but in the test.sh script it installs Istio (that install step is in that page) and deletes it by deleting the validating and mutating webhoook. That could be fixed, maybe.

PR #12782, opened 2023-02-24
Include cleanup section for istioctl analyse

sathishsw:
Subject: https://github.com/istio/istio.io/issues/12759 - Add cleanup instructions to istioctl analyze pag
Description:
This PR brings cleanup document section for ```istioctl analyse``` command

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
+++ b/content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
@@ -156,6 +156,18 @@ To ignore multiple codes for a resource, separate each code with a comma:
 $ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002
 {{< /text >}}
 
+### Cleanup
+You could also cleanup the deployments by running:
+
+{{< text syntax=bash snip_id=analyse_cleanup >}}
+$ kubectl label namespace default istio-injection-
+$ kubectl delete ns frod
+$ kubectl delete deployment my-deployment
+$ kubectl delete vs ratings
+$ istioctl uninstall --purge -y
+$ kubectl delete ns istio-system
+{{< /text >}}
+
 ## Helping us improve this tool
 
 We're continuing to add more analysis capability and we'd love your help in identifying more use cases.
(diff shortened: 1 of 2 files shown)


content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```diff
@@ -156,6 +156,18 @@ To ignore multiple codes for a resource, separate each code with a comma:
 $ kubectl annotate deployment my-deployment galley.istio.io/analyze-suppress=IST0107,IST0002
 {{< /text >}}
 
+### Cleanup
```
kfaseela, line 159: why do we need cleanup section for istioctl analyze?

frankbu: No, I don't think this makes sense. Cleanup instructions are only supposed to remove resources that are created in the task, but none of these things are created by instructions in this doc.

johnma14: I think I was looking at the test files for places where we were deleting Istio installation by just deleting the validating and mutating webhoook configuration and then deleting the istio-system ns. You are right, we should remove all those lines where resources are not created by this task. The page does have a line about installing istio ` istioctl install --set values.global.istiod.enableAnalysis=true`. But I agree it doesn't make sense to uninstall something that the task hasn't specifically required to install.

Arhell (commented):
/ok-to-test

content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
frankbu: > why do we need cleanup section for istioctl analyze?

We don't want a cleanup section in this doc because it isn't a task that tells the user to do things that need to be cleaned up. To test the snips, the test program sets up a test environment that needs to be cleaned up, but that's outside of the document.

kfaseela:
@sathishsw : hope we can close this PR, as the issue is closed by @johnma14

PR #12834, opened 2023-03-03
breakout support status table to yaml file

sspaeth-r7:
This PR defines all the data for the support status table in a YAML file, making it easily machine-readable. No user-facing changes are present. The only change for maintainers would be updating the YAML file instead of a markdown table when adding new support entries.

Being able to programmatically determine supported Kubernetes versions for a given Istio version would be extremely helpful for us. I propose an easier way to accomplish this than by creating an endpoint somewhere, or trying to parse the markdown table. The raw YAML file could be read straight from the repo. See similar: https://github.com/kiali/kiali.io/pull/633

Additionally, the proposed changes make it no longer necessary to update the `supported_kubernetes_versions` array in the `args.yml` file as it is now read from the appropriate entry in `supportStatus.yml`. I have removed the argument as a result. That said, the updated `supported_kubernetes_versions` shortcode requires a support status entry for the site's version to be present. For example, because the `preliminary` incarnation is on 1.18, and there's no 1.18 support entry yet, the shortcode will be empty while running the site in prelim mode until a 1.18 entry is added. Its output remains identical though when building for prod/main istio.io as its version - 1.17 - has a support entry. This ensures the K8s versions are those specifically entered as supported by the current release, rather than what may have been left over in the argument that's used now.

If there's an alternative way to populate the version used in the lookup that I'm not aware of, please advise. I'm happy to update the PR. I originally had it always using the second entry in the `supportStatus.yml` file (since first is master), but in that configuration, if the first element is ever removed, it would lookup the wrong version of Istio. Because of this I ultimately settled on looking up by version based off a Hugo condition of whether `preliminary` was the the current incarnation.

I'm guessing adding this shortcode depends on what point in the process entries are added to the current markdown table. Based off the current entries, it seems like they're added when/before a version is released, so I don't personally see any issue with it. If it's a problem though, I'm happy to not include any changes to the supported K8s versions shortcode since it doesn't affect my use case.

Lastly, even though quoting is generally optional in YAML, I ended up doing so on each property due to special characters, boolean meaning, and preventing versions from being seen as numbers (and therefore dropping trailing zeros).

I locally ensured this PR has zero visual changes to the docs. This is only a change to how the table is *created*, not presented.

Thank you for your consideration!

- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

> **Note**
> I don't know what all automations y'all have setup, so I can't say whether this should tick the "test and release box". I'm assuming not since I find it unlikely that markdown table is being created by an automation.


sspaeth-r7:
Hi Eric, thanks for your review. So do you want `supported_kubernetes_versions` to be generated from the row matching `.Site.Data.versions.main` if found, otherwise from master's row? I'd copy master's `k8sVersions` values the newest version with values right now - 1.17. Or are you suggesting another way to build the pick the version?

ericvn:
Maybe look at `source_branch_name` and use the values from that row. Right now it's `master`, so use the values there (you need to add them back). If it's pointing at a release branch, will be updated to release-1.18 after it's cut, then we'll use that from the json. It would be good if there was a command we could use to make a new row for the 1.18 data from the master row at branch time (when we update the source_branch_name).

sspaeth-r7:
@ericvn Pushed an update, let me know if that looks like it'd be good. I will work on a script that updates the version.

sspaeth-r7:
Are the scripts in the `scripts` directory run locally when prepping for a release or automated? Would a tool such as yq be acceptable to use in them if so?

ericvn:
In the case of release-branching (when we would update this source_branch_name to the release branch, see `scripts/prepare_release.sh`. You can see the sed for updating the release. We would likely want to add something here to make sure that there is a new line in the table for that release. There is a `create_version.sh` as well which will sett the source_branch_name back, but I don't think we need to worry about updating the supported versions for the master row (although it would likely good to set the master row to what is in the branch (in the advance master to next release)).

ericvn:
For where I expect to see the versions would be on a page like: https://deploy-preview-12834--preliminary-istio.netlify.app/latest/docs/setup/getting-started/ where where have the supported version listed in the third paragraph. Today the parends are empty. You could look at the same page on preliminary.istio.io to see the versions).

sspaeth-r7:
Yeah the master entry in the yaml file still has to be updated. I was going to do this at the same time as the script commit since I was going to use the new script to populate it. Previous commit shows working logic though - tested by updating the master entry temporarily.

I will take a look at the scripts you mentioned and try to put something together. But knowing the target environment that those scripts will be ran on to know what I have access to would be best. Obviously the sed idea could work but it'd be far more hacky than just using something like yq if that's available wherever this is running. Or to know whether to use macOS(BSD) sed or GNU sed.

ericvn:
Normally the make targets are run manually just after the branch cut and as part of release day. The scripts are what the make targets run. This is done manually and not part of automation and I can say the the running off the scripts isn't typically totally error free and sometimes the script's commands steps end up being run manually. So, having to make the table manually is fine, but it would still be better as part of the script as it will at least be noticed as something that needs to be done. Being in a json file as opposed to the markdown makes it a little less visible as something the person making the changes will see.

sspaeth-r7:
Doesn't look like I can add labels to this PR but this is still a **work in progress**.

sspaeth-r7:
I added a script (`update_support_status.py`) that will update the master entry with the latest release's `k8sVersions`. The script is run with no arguments.

I noticed that PyYAML was already being used by another script so figured it was installed on y'all's end and safe to use. Doing so should make this way safer than just using sed in-place mode to do the replacement.

The last two commits are arguably worse but the linter wouldn't approve unless I had tons of newlines around the classes. Not entirely sure why that is (I'm not very familiar with Python but haven't seen this before). Same goes for the two spaces at line end before I can do a comment.

Should I call on this script from another one? What's the best way to integrate it into your automation?

ericvn:
@frankbu is much better with python than I, but you could likely call this from scripts/create_version.sh in advance_master_to_next_release(). Maybe just before the `git status --porcelain`.

sspaeth-r7:
Sounds good, I added a commit. Looks like the scripts are meant to be run from the root so I'm using the path relative to that (`scripts/update_support_status.py` as opposed to just `update_support_status.py`).

@frankbu let me know if I should change anything with the Python. I'm by no means a Python developer and have never used the yaml package before so I'm just hacking it together from docs. I'm also curious why the linter was forcing me to put so many newlines after the comment and before the class definition as well as after lines and before comments.

frankbu:
> @frankbu let me know if I should change anything with the Python. I'm by no means a Python developer and have never used the yaml package before so I'm just hacking it together from docs. I'm also curious why the linter was forcing me to put so many newlines after the comment and before the class definition as well as after lines and before comments.

@sspaeth-r7 I've only ever used PyYAML to read files, so I really don't know anything about how to add representers (or sublass built-in types for that matter), which seems to be the source of the weird linter issues you're hitting. But if it works, I suppose it's fine.

That said, I think using python for this is overly complicated, esp. given that you have to reproduce comments.

You can use `yq` instead, it is available in the build tools: https://github.com/istio/tools/blob/96c21854bff423cb1c6da4b8a1b52cbf412757fd/docker/build-tools/Dockerfile#L224

sspaeth-r7:
I wish I knew that yq was available in the build tools before working on the Python script 😆 I had a feeling it might be which is why I asked before starting. Ended up using PyYAML because it wasn't clear whether I could use yq. I grepped for any use of it in the scripts dir and didn't find anything, but found PyYAML... On the upside now I'm a little more familiar with Python and that library!

I added a commit that uses a yq one-liner to accomplish the same goal as the ~70 line Python script 😂

sspaeth-r7:
Thanks for taking a look as well @frankbu!

frankbu:
> I added a commit that uses a yq one-liner to accomplish the same goal as the ~70 line Python script 😂

Ha ha, that's what I thought. On the bright side you and I both learned a little about how to add representers in pyYAML 😃

frankbu (approved):

ericvn:
Thanks @sspaeth-r7 for all the work on this.

sspaeth-r7:
Absolutely - thanks for working with me y'all!

Issue #12844, opened 2023-03-04
update hugo to 0.111.2


Arhell:
https://github.com/gohugoio/hugo/releases/tag/v0.111.2

PR #12846, opened 2023-03-05
update hugo to 0.111.2

Arhell:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

resolve https://github.com/istio/istio.io/issues/12844

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,7 @@
     publish = "public"
 
 [build.environment]
-    HUGO_VERSION = "0.110.0"
+    HUGO_VERSION = "0.111.2"
     NODE_VERSION = "18.12.1"
     BUILD_WITH_CONTAINER = "0"
     GO_VERSION = "1.19.3"


ericvn (changes requested):
I'm not sure which bug fix we are needing with the Hugo update, but 0.111.2 is the latest.

ericvn (approved):
Approve this update now, but in the future, we should have a reason to update the release. Something is fixed, or maybe needed for new function. There are a number of other issues like back level versions of other code that would be better suited (https://github.com/istio/istio.io/issues/11971)

Arhell:
> Approve this update now, but in the future, we should have a reason to update the release. Something is fixed, or maybe needed for new function. There are a number of other issues like back level versions of other code that would be better suited

Okay, I'm just doing my little contribution. I'll keep that in mind, thanks

PR #12850, opened 2023-03-07
add missing tests for ambient doc

kfaseela:
add test for l7 auth policy config

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/getting-started/test.sh
+++ b/content/en/docs/ops/ambient/getting-started/test.sh
@@ -59,13 +59,17 @@ _verify_contains snip_l7_authorization_policy_3 "Deployed waypoint proxy"
 snip_l7_authorization_policy_5
 _verify_contains snip_l7_authorization_policy_6 "$snip_l7_authorization_policy_6_out"
 _verify_contains snip_l7_authorization_policy_7 "$snip_l7_authorization_policy_7_out"
-_verify_contains snip_l7_authorization_policy_8 "$snip_l7_authorization_policy_8_out"
+_verify_contains snip_verify_traffic_sleep_to_productpage "$snip_verify_traffic_sleep_to_productpage_out"
 
 snip_control_traffic_control_1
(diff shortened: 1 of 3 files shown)


content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -59,13 +59,17 @@ _verify_contains snip_l7_authorization_policy_3 "Deployed waypoint proxy"
 snip_l7_authorization_policy_5
 _verify_contains snip_l7_authorization_policy_6 "$snip_l7_authorization_policy_6_out"
 _verify_contains snip_l7_authorization_policy_7 "$snip_l7_authorization_policy_7_out"
-_verify_contains snip_l7_authorization_policy_8 "$snip_l7_authorization_policy_8_out"
+_verify_contains snip_verify_traffic_sleep_to_productpage "$snip_verify_traffic_sleep_to_productpage_out"
```
frankbu, line 62: Why aren't you using the actual snip here? The purpose of the test is to make sure the snippets shown in the document continue to work. If somebody accidentally changes the code in snip_l7_authorization_policy_8 in the doc, so that it doesn't actually work anymore, the test wouldn't catch that.

kfaseela: ok, i thought it is all the same command, so why unnecessarily generate duplicated snips. I can revert it back

frankbu: Yeah, if we wanted to optimize it should be in the doc (i.e., tell the user to repeat a previous step). But in this case it's better to just have another copy of the command in the doc, so we also need to test the copy.

kfaseela: Done!

frankbu (approved):

PR #12842, opened 2023-03-04
[zh]sync deployment-models

my-git9:
Please provide a description for what this PR is for.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

- [ ] Configuration Infrastructure
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

PR #12863, opened 2023-03-09
Fix issues about glossary reference

windsonsea:
When reviewing #12842, it is found if an incomplete `{{< gloss >}}control plane{{< /gloss >}}` is translated into `{{< gloss >}}控制面板{{< /gloss >}}`, ci raised an error:

```
unrecognized character in shortcode action: U+003E '>'. Note: Parameters with non-alphanumeric args must be quoted
```
So have a try to change it to `{{< gloss "control plane" >}}control plane{{< /gloss >}}`. Hope it works.

- [x] Docs

### DIFF
--- a/content/en/docs/ops/deployment/deployment-models/index.md
+++ b/content/en/docs/ops/deployment/deployment-models/index.md
@@ -21,13 +21,13 @@ test: n/a
 ---
 
 When configuring a production deployment of Istio, you need to answer a number of questions.
-Will the mesh be confined to a single {{< gloss >}}cluster{{< /gloss >}} or distributed across
+Will the mesh be confined to a single {{< gloss "cluster" >}}clusters{{< /gloss >}} or distributed across
 multiple clusters? Will all the services be located in a single fully connected network, or will
 gateways be required to connect services across multiple networks? Is there a single
-{{< gloss >}}control plane{{< /gloss >}}, potentially shared across clusters,
+{{< gloss "control plane">}}control plane{{< /gloss >}}, potentially shared across clusters,
 or are there multiple control planes deployed to ensure high availability (HA)?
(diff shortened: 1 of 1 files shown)


frankbu:
You should not need to change the en docs because you want to use a different display string in zh; you should do this in the zh docs only.

frankbu (changes requested):
I'm not sure this will work, but regardless, it should only be done in zh docs.

windsonsea:
Those changes are mainly for fixing inconsistency, which may cause a trap for localization and also a part of bad doc, for example:

- line 27: {{< gloss >}}control plane{{< /gloss >}}
- line 54: {{< gloss "cluster" >}}clusters{{< /gloss >}}
- line 74: {{< gloss >}}cluster{{< /gloss >}}
- line 93: {{< gloss "cluster" >}}clusters{{< /gloss >}}
- line 124: {{< gloss "namespace sameness" >}}namespace sameness{{< /gloss >}}
- line 186: {{< gloss >}}workload instance{{< /gloss >}}s
- line 199: {{< gloss "workload instance" >}}workload instances{{< /gloss >}}
- and more

frankbu:
> Those changes are mainly for fixing inconsistency ...

There is no inconsistency. The consistent rule is:

1. If the display text is exactly the same as the glossary entry, then the reference is like this:
    `{{< gloss >}}gloss-entry{{< /gloss >}}`
2. If the display text is different from the glossary entry, then the reference is like this:
    `{{< gloss "gloss-entry" >}}some-different-text{{< /gloss >}}`

I'm guessing you will always want to use the second form in the zh docs, but there's no need to make the en ones unnecessarily more complicated by using the second form even when `gloss-entry == some-different-text`

windsonsea:
Well, then this PR can be closed though it may cause some problems to localization efforts.

PR #12869, opened 2023-03-09
gateway-api: add gateway api instructions to ingress access control doc

frankbu:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -60,16 +101,81 @@ What if the Pod that is handling traffic from the `NodePort` or `LoadBalancer` i
...
+
+{{< tab name="Gateway API" category-value="gateway-api" >}}
+
+{{< text yaml >}}
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: httpbin-gateway
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
```
frankbu, line 157: @howardjohn Does this look right? There is no test for this.

content/en/docs/tasks/security/authorization/authz-ingress/index.md
```diff
@@ -124,55 +262,76 @@ spec:
 
 {{< /tab >}}
 
-{{< tab name="Network Load Balancer" category-value="network" >}}
+{{< tab name="Gateway API" category-value="gateway-api" >}}
+
+{{< text yaml >}}
+apiVersion: gateway.networking.k8s.io/v1beta1
```
frankbu, line 268: @howardjohn Also, does this seem OK? This example is not complete (e.g., affinity rule)

howardjohn: I don't even get why we have affinity rules in the first place. As far as I can tell it is irrelevant to the doc and should be removed.

Samples, IMO, should have the minimum required to make things work, otherwise we are mislead to think you need the affinity setting. See logs comment below as well

Caveat: I am not 100% sure its not required, but I can't imagine why it would be

frankbu: > Samples, IMO, should have the minimum required to make things work ...

Agree 100%. I'll remove both the affinity and logging from both the gateway and the Istio examples. If it turns out something is needed, somebody will open an issue and we can add them back.

ericvn (approved):
Thanks for the update.

PR #12888, opened 2023-03-13
add docs for DNS Proxying for External TCP services

kfaseela:

- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -30,6 +30,12 @@ spec:
         ISTIO_META_DNS_CAPTURE: "true"
         # Enable automatic address allocation, optional
         ISTIO_META_DNS_AUTO_ALLOCATE: "true"
+    # The below configuration is only used in the External TCP services section of this document
+    # to clarify the external service behaviour in a simplified way.
+    # Otherwise this is not a mandatory setting for DNS proxying.
+    discoverySelectors:
+    - matchLabels:
+        istio-injection: enabled
```
frankbu, line 38: I think it would be better to leave this out here, and just keep it as simple as possible for using the feature. You will then need to repeat the install command in the later section, but that's usually preferable anyway vs telling the user to go back and follow instructions in an earlier section.

```suggestion
```

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -135,10 +141,76 @@ $ kubectl exec deploy/sleep -- curl -sS -v auto.internal
...
 
+## External TCP services without VIPs
+
+Istio by default has a limitation on routing external TCP traffic, because it is not able to distinguish between two different TCP services. This limitation especially impacts the use of third party databases, such as AWS Relational Database Service or any database setup with geographical redundancy. Similar, but different external TCP services, cannot be handled separately by default. For the sidecar to accurately distinguish traffic between two different TCP services that are outside the mesh, the services must be on different ports or they need to have a globally unique VIP. For example, if you have two external database services, `mysql-instance1` and `mysql-instance2`, and you create service entries for them, the sidecar will still have a single listener on `0.0.0.0:{port}` that looks up the IP address of only `mysql-instance1` from public DNS servers and forwards traffic to it. It cannot route traffic to `mysql-instance2` as it has no way of distinguishing whether traffic arriving at `0.0.0.0:{port}` is bound for `mysql-instance1` or `mysql-instance2`.
```
frankbu, line 146: This seems like just an extension of the previous section and repeating some of the same things as around line 98?

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -135,10 +141,76 @@ $ kubectl exec deploy/sleep -- curl -sS -v auto.internal
...
+
+The below example shows how we can make use of DNS proxying as a solution to this problem. A virtual IP address is assigned to the service entries automatically in this case, so that sidecar can clearly distinguish traffic bound for each external TCP service.
+
+1.  Enable DNS automatic address allocation as specified in the [Getting Started](#getting-started) section.
+
+1.  Bring up the client sleep pod as described in the [DNS Capture in Action](#dns-capture-in-action) section.
```
frankbu, line 152: Can't you just continue the previous example without needing to restart the sleep sample?

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -135,10 +141,76 @@ $ kubectl exec deploy/sleep -- curl -sS -v auto.internal
...
+
+The below example shows how we can make use of DNS proxying as a solution to this problem. A virtual IP address is assigned to the service entries automatically in this case, so that sidecar can clearly distinguish traffic bound for each external TCP service.
+
+1.  Enable DNS automatic address allocation as specified in the [Getting Started](#getting-started) section.
```
frankbu, line 150:
```suggestion
1.  Update the Istio configuration specified in the [Getting Started](#getting-started) section to also configure `discoverySelectors` that restrict the mesh to namespaces with `istio-injection` enabled. This will let us use any other namespaces in the cluster to run TCP services outside of the mesh.

{{< text bash >}}
$ cat <<EOF | istioctl install -y -f -
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  meshConfig:
    defaultConfig:
      proxyMetadata:
        # Enable basic DNS proxying
        ISTIO_META_DNS_CAPTURE: "true"
        # Enable automatic address allocation, optional
        ISTIO_META_DNS_AUTO_ALLOCATE: "true"
    discoverySelectors:
    - matchLabels:
        istio-injection: enabled
{{< /text >}}
```

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -135,10 +141,76 @@ $ kubectl exec deploy/sleep -- curl -sS -v auto.internal
...
+
+The below example shows how we can make use of DNS proxying as a solution to this problem. A virtual IP address is assigned to the service entries automatically in this case, so that sidecar can clearly distinguish traffic bound for each external TCP service.
```
frankbu, line 148:
```suggestion
The following example shows how DNS proxying can be used to solve this problem.
A virtual IP address will be assigned to the every service entry so that client sidecars can clearly distinguish traffic bound for each external TCP service.
```

content/en/docs/ops/configuration/traffic-management/dns-proxy/snips.sh
```diff
@@ -93,8 +99,59 @@ kubectl exec deploy/sleep -- curl -sS -v auto.internal
...
+  resolution: DNS
+EOF
+}
+
+snip_external_tcp_services_without_vips_4() {
+istioctl pc listener deploy/sleep | grep 9000
+}
+
+! read -r -d '' snip_external_tcp_services_without_vips_4_out <<\ENDSNIP
+240.240.105.94 9000  ALL                                                                                           Cluster: outbound|9000||tcp-echo.external-2.svc.cluster.local
```
frankbu, line 147: This line is very long and hard to read. I wonder if there's a better way to display?

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -135,10 +141,76 @@ $ kubectl exec deploy/sleep -- curl -sS -v auto.internal
...
+        protocol: TCP
+      resolution: DNS
+    EOF
+    {{< /text >}}
+
+1.  Verify listeners are configured separately for each service at the client side:
+
+    {{< text bash >}}
+    $ istioctl pc listener deploy/sleep | grep 9000
+    240.240.105.94 9000  ALL                                                                                           Cluster: outbound|9000||tcp-echo.external-2.svc.cluster.local
```
frankbu, line 204: This line is very long and hard to read. I wonder if there's a better way to display?

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
kfaseela: Done

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
kfaseela: changed it, can you check if the new way of printing is better?

frankbu: Yes, I think that's better.

Maybe overkill, but maybe using printf would be even better? For example:
```
awk '{printf "ADDRESS=%s, DESTINATION=%s %s\n", $1, $4, $5}'
```
to display something like this:
```
ADDRESS=240.240.105.94, DESTINATION=Cluster: outbound|9000||tcp-echo.external-2.svc.cluster.local
ADDRESS=240.240.69.138, DESTINATION=Cluster: outbound|9000||tcp-echo.external-1.svc.cluster.local
```

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -95,7 +95,8 @@ $ kubectl exec deploy/sleep -- curl -sS -v address.internal
...
 
-This is especially problematic with TCP traffic. Unlike HTTP requests, which are routed based on `Host` headers, TCP carries much less information; you can only route on the destination IP and port number. Because you don't have a stable IP for the backend, you cannot route based on that either, leaving only port number, which leads to conflicts when multiple `ServiceEntry`s for TCP services share the same port.
+This is especially problematic with TCP traffic. Unlike HTTP requests, which are routed based on `Host` headers, TCP carries much less information; you can only route on the destination IP and port number. Because you don't have a stable IP for the backend, you cannot route based on that either, leaving only port number, which leads to conflicts when multiple `ServiceEntry`s for TCP services share the same port. Refer
+[External TCP services without VIP](#external-tcp-services-without-vips) section for additional details.
```
frankbu, line 99:
```suggestion
to [the following section](#external-tcp-services-without-vips) for more details.
```

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
kfaseela: Modified

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -135,10 +136,95 @@ $ kubectl exec deploy/sleep -- curl -sS -v auto.internal
...
+
+For example, if you have two external database services, `mysql-instance1` and `mysql-instance2`, and you create service entries for both, client sidecars will still have a single listener on `0.0.0.0:{port}` that looks up the IP address of only `mysql-instance1`, from public DNS servers, and forwards traffic to it. It cannot route traffic to `mysql-instance2` because it has no way of distinguishing whether traffic arriving at `0.0.0.0:{port}` is bound for `mysql-instance1` or `mysql-instance2`.
+
+The following example shows how DNS proxying can be used to solve this problem.
+A virtual IP address will be assigned to every service entry so that client sidecars can clearly distinguish traffic bound for each external TCP service.
+
+1.  Update the Istio configuration specified in the [Getting Started](#getting-started) section to also configure `discoverySelectors` that restrict the mesh to namespaces with `istio-injection` enabled. This will let us use any other namespaces in the cluster to run TCP services outside of the mesh.
```
howardjohn, line 148: Not sure why discoverySelectors is required? Generally its best to keep examples as minimal as possible or it adds confusion about why it was set.

In this case, I  don't think its needed? Because we have auto-alloc we don't have issues with port conflicts.

content/en/docs/ops/configuration/traffic-management/dns-proxy/test.sh
```diff
@@ -33,6 +33,13 @@ _verify_first_line snip_dns_capture_in_action_3 "$snip_dns_capture_in_action_3_o
 snip_address_auto_allocation_1
 _verify_contains snip_address_auto_allocation_2 "*   Trying 240.240."
 
+# verify external tcp services without vips
+snip_external_tcp_services_without_vips_1
+snip_external_tcp_services_without_vips_2
+snip_external_tcp_services_without_vips_3
+snip_external_tcp_services_without_vips_4
+_verify_like snip_external_tcp_services_without_vips_5 "$snip_external_tcp_services_without_vips_5_out"
```
howardjohn, line 41: is this going to assume the IP output is stable? Might not be true

howardjohn (commented):
content looks good, just 1 question

content/en/docs/ops/configuration/traffic-management/dns-proxy/test.sh
kfaseela: @howardjohn "_verify_like" does not need exact IP... IP part can change. https://github.com/istio/istio.io/blob/master/tests/util/verify.sh#L428

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
kfaseela: If discoverySelectors is not present, cannot simulate the problem of external service, because tcp-echo service which is within the cluster is getting discovered by default.

howardjohn: I see. ExportTo or Sidecar may be a easier way to simulate it. And maybe we can mention its just for testing purposes, so we don't have to use an external site for testing, and not needed in real world?

content/en/docs/ops/configuration/traffic-management/dns-proxy/test.sh
howardjohn: nice

howardjohn:
Slight prefer exportTo/sidecar but lgtm!

content/en/docs/ops/configuration/traffic-management/dns-proxy/test.sh
frankbu, line 41: `_verify_lines` is only tolerant of varying IPs if it is a space-separated token. In this case it won't work because the token is `ADDRESS=x.x.x.x`. You could fix it by changing the output format to make it a token, but I think maybe it's better to go back to your previous implementation using `_verify_lines` instead:
```suggestion
_verify_lines snip_external_tcp_services_without_vips_5 "
+ outbound|9000||tcp-echo.external-2.svc.cluster.local
+ outbound|9000||tcp-echo.external-1.svc.cluster.local
"
```

PR #12932, opened 2023-03-23
Adds K8s Gateway API to Getting Started Guide

danehans:

This PR updates the Ambient getting started guide with steps for using Kubernetes Gateway API.

- [X] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/getting-started/gtwapi_test.sh
+++ b/content/en/docs/ops/ambient/getting-started/gtwapi_test.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2154,SC2034
+
+# Copyright 2023 Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+GATEWAY_API="true"
+
+# @setup profile=none
+source "content/en/docs/ops/ambient/getting-started/test.sh"
+
+# @cleanup
--- a/content/en/docs/ops/ambient/getting-started/gtwapi_test.sh
+++ b/content/en/docs/ops/ambient/getting-started/gtwapi_test.sh
@@ -15,9 +15,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# @setup profile=none
 GATEWAY_API="true"
 
-# @setup profile=none
 source "content/en/docs/ops/ambient/getting-started/test.sh"
 
 # @cleanup
+GATEWAY_API="true"
+
+source "tests/util/samples.sh"
+
+snip_uninstall_uninstall_1
+snip_uninstall_uninstall_2
+snip_uninstall_uninstall_3
+cleanup_bookinfo_sample
+snip_uninstall_uninstall_4
--- a/content/en/docs/ops/ambient/getting-started/gtwapi_test.sh
+++ b/content/en/docs/ops/ambient/getting-started/gtwapi_test.sh
@@ -23,10 +23,8 @@ source "content/en/docs/ops/ambient/getting-started/test.sh"
 # @cleanup
 GATEWAY_API="true"
 
-source "tests/util/samples.sh"
-
 snip_uninstall_uninstall_1
 snip_uninstall_uninstall_2
 snip_uninstall_uninstall_3
-cleanup_bookinfo_sample
+samples/bookinfo/platform/kube/cleanup.sh
 snip_uninstall_uninstall_4
--- a/content/en/docs/ops/ambient/getting-started/index.md
+++ b/content/en/docs/ops/ambient/getting-started/index.md
@@ -87,9 +87,15 @@ Follow these steps to get started with ambient:
     ztunnel          1         1         1       1            1           <none>                   82s
     {{< /text >}}
 
-1. This guide includes tasks that use a `Gateway`.
+1.  Install Kubernetes Gateway CRDs, which don’t come installed by default on most Kubernetes clusters:
 
-   {{< boilerplate gateway-api-support >}}
+    {{< text bash >}}
+    $ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
+      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.1" | kubectl apply -f -; }
+    {{< /text >}}
+
+    {{< boilerplate gateway-api-future >}}
+    {{< boilerplate gateway-api-choose >}}
 
 ## Deploy the sample application {#bookinfo}
 
@@ -138,31 +144,18 @@ $ export GATEWAY_SERVICE_ACCOUNT=ns/istio-system/sa/istio-ingressgateway-service
 
 {{< tab name="Gateway API" category-value="gateway-api" >}}
 
-Create a [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway),
-so you can access the bookinfo app from outside the cluster:
+Create a [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway)
+and [HTTPRoute](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute) so you can access
+the bookinfo app from outside the cluster:
 
 {{< text bash >}}
-$ kubectl apply -f - <<EOF
-apiVersion: gateway.networking.k8s.io/v1beta1
-kind: Gateway
-metadata:
-  name: bookinfo-gateway
-  namespace: istio-system
-spec:
-  gatewayClassName: istio
-  listeners:
-  - name: http
-    port: 80
-    protocol: HTTP
-    allowedRoutes:
-      namespaces:
-        from: All
-EOF
+$ sed -e 's/from: Same/from: All/'\
+    -e '/^  name: bookinfo-gateway/a\
+  namespace: istio-system\
+'   -e '/^  - name: bookinfo-gateway/a\
+    namespace: istio-system\
+' @samples/bookinfo/gateway-api/bookinfo-gateway.yaml@ | kubectl apply -f -
 {{< /text >}}
-
-The Gateway is configured to listen for HTTP traffic on port 80 and allow
-[HTTPRoutes](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRoute) from any
-namespace to bind to the listener.
 
 Creating a Kubernetes Gateway resource deploys an associated
 [proxy service](/docs/tasks/traffic-management/ingress/gateway-api/#automated-deployment), so run the following command
@@ -171,34 +164,6 @@ to wait for the gateway to be ready:
 {{< text bash >}}
 $ kubectl wait --for=condition=programmed gtw/bookinfo-gateway -n istio-system
 {{< /text >}}
-
-Route traffic through the Gateway to the `productpage` Service:
-
-{{< text bash >}}
-$ kubectl apply -f - <<EOF
-apiVersion: gateway.networking.k8s.io/v1beta1
-kind: HTTPRoute
-metadata:
-  name: productpage
-spec:
-  parentRefs:
-  - name: bookinfo-gateway
-    namespace: istio-system
-  hostnames:
-  - "bookinfo-gateway-istio.istio-system"
-  rules:
-  - matches:
-    - path:
-        type: PathPrefix
-        value: /productpage
-    backendRefs:
-    - name: productpage
-      port: 9080
-EOF
-{{< /text >}}
-
-You have now created an HTTPRoute configuration for the `productpage` service containing one route rule that allows
-traffic for path `/productpage`.
 
 Set the environment variables for the Kubernetes gateway:
 
@@ -412,7 +377,7 @@ $ kubectl apply -f @samples/bookinfo/networking/destination-rule-reviews.yaml@
 
 Create subsets of the bookinfo Services to control traffic to `reviews` v1 and v2 versions:
 
-{{< text syntax=bash snip_id=none >}}
+{{< text bash >}}
 $ kubectl apply -f @samples/bookinfo/platform/kube/bookinfo-versions.yaml@
 {{< /text >}}
 
@@ -434,15 +399,6 @@ $ kubectl exec deploy/sleep -- sh -c "for i in \$(seq 1 100); do curl -s http://
 
 ## Uninstall {#uninstall}
 
-To delete the Bookinfo sample application and its configuration, see [`Bookinfo` cleanup](/docs/examples/bookinfo/#cleanup).
-
-To remove the `sleep` and `notsleep` applications:
-
-{{< text bash >}}
-$ kubectl delete -f @samples/sleep/sleep.yaml@
-$ kubectl delete -f @samples/sleep/notsleep.yaml@
-{{< /text >}}
-
 To remove the `productpage-viewer` authorization policy, waypoint proxies and uninstall Istio:
 
 {{< text bash >}}
@@ -459,36 +415,17 @@ The label to instruct Istio to automatically include applications in the `defaul
 $ kubectl label namespace default istio.io/dataplane-mode-
 {{< /text >}}
 
-{{< tabset category-name="config-api" >}}
+To delete the Bookinfo sample application and its configuration, see [`Bookinfo` cleanup](/docs/examples/bookinfo/#cleanup).
 
-{{< tab name="Istio classic" category-value="istio-classic" >}}
+To remove the `sleep` and `notsleep` applications:
 
-To remove the `bookinfo` Gateway:
-
-{{< text syntax=bash snip_id=none >}}
-$ kubectl delete gw/bookinfo-gateway
+{{< text bash >}}
+$ kubectl delete -f @samples/sleep/sleep.yaml@
+$ kubectl delete -f @samples/sleep/notsleep.yaml@
 {{< /text >}}
-
-{{< /tab >}}
-
-{{< tab name="Gateway API" category-value="gateway-api" >}}
-
-To remove the `reviews` Service subsets and HTTPRoute:
-
-{{< text syntax=bash snip_id=none >}}
-$ kubectl delete -f @samples/bookinfo/platform/kube/bookinfo-versions.yaml@
-{{< /text >}}
-
-{{< text syntax=bash snip_id=none >}}
-$ kubectl delete -f httproute/reviews
-{{< /text >}}
-
-{{< /tab >}}
-
-{{< /tabset >}}
 
 If you installed the Gateway API CRDs, remove them:
 
-{{< text syntax=bash snip_id=none >}}
-$ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.1" | kubectl delete -f -
+{{< text bash >}}
+$ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.1" | kubectl delete -f -
 {{< /text >}}
--- a/content/en/docs/ops/ambient/getting-started/index.md
+++ b/content/en/docs/ops/ambient/getting-started/index.md
@@ -87,6 +87,10 @@ Follow these steps to get started with ambient:
     ztunnel          1         1         1       1            1           <none>                   82s
     {{< /text >}}
 
+1. This guide includes tasks that use a `Gateway`.
+
+   {{< boilerplate gateway-api-support >}}
+
 ## Deploy the sample application {#bookinfo}
 
 You’ll use the sample [bookinfo application](/docs/examples/bookinfo/), which is part of
@@ -111,17 +115,106 @@ $ kubectl apply -f @samples/sleep/notsleep.yaml@
 
 Note: `sleep` and `notsleep` are two simple applications that can serve as curl clients.
 
-Connect `productpage` to the Istio ingress gateway so you can access the bookinfo
-app from outside of the cluster:
+{{< tabset category-name="config-api" >}}
+
+{{< tab name="Istio classic" category-value="istio-classic" >}}
+
+Create an Istio [Gateway](/docs/reference/config/networking/gateway/) and
+[VirtualService](/docs/reference/config/networking/virtual-service/), so you can access the bookinfo app through the
+Istio ingress gateway:
 
 {{< text bash >}}
 $ kubectl apply -f @samples/bookinfo/networking/bookinfo-gateway.yaml@
 {{< /text >}}
 
-Test your bookinfo application, it should work with or without the gateway. Note: you can replace `istio-ingressgateway.istio-system` below with its load balancer IP (or hostname) if it has one:
+Set the environment variables for the Istio ingress gateway:
+
+{{< text bash >}}
+$ export GATEWAY_HOST=istio-ingressgateway.istio-system
+$ export GATEWAY_SERVICE_ACCOUNT=ns/istio-system/sa/istio-ingressgateway-service-account
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="Gateway API" category-value="gateway-api" >}}
+
+Create a [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.Gateway),
+so you can access the bookinfo app from outside the cluster:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: bookinfo-gateway
+  namespace: istio-system
+spec:
+  gatewayClassName: istio
+  listeners:
+  - name: http
+    port: 80
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: All
+EOF
+{{< /text >}}
+
+The Gateway is configured to listen for HTTP traffic on port 80 and allow
+[HTTPRoutes](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRoute) from any
+namespace to bind to the listener.
+
+Creating a Kubernetes Gateway resource deploys an associated
+[proxy service](/docs/tasks/traffic-management/ingress/gateway-api/#automated-deployment), so run the following command
+to wait for the gateway to be ready:
+
+{{< text bash >}}
+$ kubectl wait --for=condition=programmed gtw/bookinfo-gateway -n istio-system
+{{< /text >}}
+
+Route traffic through the Gateway to the `productpage` Service:
+
+{{< text bash >}}
+$ kubectl apply -f - <<EOF
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: productpage
+spec:
+  parentRefs:
+  - name: bookinfo-gateway
+    namespace: istio-system
+  hostnames:
+  - "bookinfo-gateway-istio.istio-system"
+  rules:
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /productpage
+    backendRefs:
+    - name: productpage
+      port: 9080
+EOF
+{{< /text >}}
+
+You have now created an HTTPRoute configuration for the `productpage` service containing one route rule that allows
+traffic for path `/productpage`.
+
+Set the environment variables for the Kubernetes gateway:
+
+{{< text bash >}}
+$ export GATEWAY_HOST=bookinfo-gateway-istio.istio-system
+$ export GATEWAY_SERVICE_ACCOUNT=ns/istio-system/sa/bookinfo-gateway-istio
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< /tabset >}}
+
+Test your bookinfo application, it should work with or without the gateway:
 
 {{< text syntax=bash snip_id=verify_traffic_sleep_to_ingress >}}
-$ kubectl exec deploy/sleep -- curl -s http://istio-ingressgateway.istio-system/productpage | grep -o "<title>.*</title>"
+$ kubectl exec deploy/sleep -- curl -s "http://$GATEWAY_HOST/productpage" | grep -o "<title>.*</title>"
 <title>Simple Bookstore App</title>
 {{< /text >}}
 
@@ -150,7 +243,7 @@ to the ambient mesh. The best part is that there was no need to restart or redep
 Send some test traffic:
 
 {{< text syntax=bash snip_id=none >}}
-$ kubectl exec deploy/sleep -- curl -s http://istio-ingressgateway.istio-system/productpage | grep -o "<title>.*</title>"
+$ kubectl exec deploy/sleep -- curl -s "http://$GATEWAY_HOST/productpage" | grep -o "<title>.*</title>"
 <title>Simple Bookstore App</title>
 {{< /text >}}
 
@@ -179,8 +272,7 @@ identities, but not at the L7 level, such as HTTP methods like `GET` and `POST`.
 
 ### L4 Authorization Policy
 
-Explicitly allow the `sleep` service account and `istio-ingressgateway` service accounts to call
- the `productpage` service:
+Explicitly allow the `sleep` and gateway service accounts to call the `productpage` service:
 
 {{< text bash >}}
 $ kubectl apply -f - <<EOF
@@ -197,7 +289,9 @@ spec:
  rules:
  - from:
    - source:
-       principals: ["cluster.local/ns/default/sa/sleep", "cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account"]
+       principals:
+       - cluster.local/ns/default/sa/sleep
+       - cluster.local/$GATEWAY_SERVICE_ACCOUNT
 EOF
 {{< /text >}}
 
@@ -205,7 +299,7 @@ Confirm the above authorization policy is working:
 
 {{< text syntax=bash snip_id=none >}}
 $ # this should succeed
-$ kubectl exec deploy/sleep -- curl -s http://istio-ingressgateway.istio-system/productpage | grep -o "<title>.*</title>"
+$ kubectl exec deploy/sleep -- curl -s "http://$GATEWAY_HOST/productpage" | grep -o "<title>.*</title>"
 <title>Simple Bookstore App</title>
 {{< /text >}}
 
@@ -224,12 +318,6 @@ command terminated with exit code 56
 ### L7 Authorization Policy
 
 Using the Kubernetes Gateway API, you can deploy a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} for the `productpage` service that uses the `bookinfo-productpage` service account. Any traffic going to the `productpage` service will be mediated, enforced and observed by the Layer 7 (L7) proxy.
-Install Kubernetes Gateway CRDs, which don’t come installed by default on most Kubernetes clusters:
-
-{{< text bash >}}
-$ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-  { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.1" | kubectl apply -f -; }
-{{< /text >}}
 
 Deploy a waypoint proxy for the `productpage` service:
 
@@ -239,15 +327,10 @@ waypoint default/bookinfo-productpage applied
 {{< /text >}}
 
 View the `productpage` waypoint proxy status; you should see the details of the gateway
-resource with `Ready` status:
+resource with `Programmed` status:
 
 {{< text bash >}}
 $ kubectl get gtw bookinfo-productpage -o yaml
-{{< /text >}}
-
-Verify that the waypoint proxy status is ready:
-
-{{< text syntax=plaintext snip_id=none >}}
 ...
 status:
   conditions:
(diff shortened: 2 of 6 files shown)


danehans:
Marked as a draft since the tests are failing locally. A snip from the failing test:
```
...
kind: Gateway
metadata:
  annotations:
    istio.io/for-service-account: bookinfo-productpage
  creationTimestamp: "2023-03-22T23:44:59Z"
  generation: 1
  name: bookinfo-productpage
  namespace: default
  resourceVersion: "8883"
  uid: ac333016-dabb-47c4-a4a6-e2319c497df8
spec:
  gatewayClassName: istio-waypoint
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: mesh
    port: 15008
    protocol: ALL
status:
  conditions:
  - lastTransitionTime: "2023-03-22T23:44:59Z"
    message: '\''Failed to assign to any requested addresses: hostname "bookinfo-productpage-istio-waypoint.default.svc.cluster.local"
      not found'\''
    observedGeneration: 1
    reason: NoResources
    status: "False"
    type: Accepted
  - lastTransitionTime: "2023-03-22T23:44:59Z"
    message: Resource programmed
    observedGeneration: 1
    reason: Programmed
    status: "True"
    type: Programmed
...
```
It appears that `istioctl x waypoint apply --service-account bookinfo-productpage` may have a bug where it's creating the waypoint with an incorrect configuration. Even after manually updating the protocol, I see:
```
$ kubectl get gtw/bookinfo-productpage -o yaml
kind: Gateway
metadata:
  annotations:
    istio.io/for-service-account: bookinfo-productpage
  creationTimestamp: "2023-03-22T23:59:45Z"
  generation: 2
  name: bookinfo-productpage
  namespace: default
  resourceVersion: "11127"
  uid: c890c731-cfc8-472d-9853-d12ea4769b20
spec:
  gatewayClassName: istio-waypoint
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    name: mesh
    port: 15008
    protocol: HBONE
status:
  conditions:
  - lastTransitionTime: "2023-03-22T23:59:45Z"
    message: 'Failed to assign to any requested addresses: hostname "bookinfo-productpage-istio-waypoint.default.svc.cluster.local"
      not found'
    observedGeneration: 2
    reason: NoResources
    status: "False"
    type: Accepted
...
```

It appears that the waypoint is expecting a Service that does not exist:
```
$ kubectl get svc/bookinfo-productpage-istio
Error from server (NotFound): services "bookinfo-productpage-istio" not found
```

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -135,6 +148,102 @@ $ kubectl exec deploy/notsleep -- curl -s http://productpage:9080/ | grep -o "<t
...
+  gatewayClassName: istio
+  listeners:
+  - name: http
+    port: 80
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: All
+EOF
+{{< /text >}}
```
frankbu, line 175: One of the goals of the gateway-api docs, is to try to make the delta between Istio config and Gateway-api be as small as possible so that users don't get the impression that switching APIs is a big deal.

There are already gateway-api yamls for the bookinfo sample that you should reuse in this doc:

```suggestion
{{< text bash >}}
$ kubectl apply -f @samples/bookinfo/gateway-api/bookinfo-gateway.yaml@
{{< /text >}}
```

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -135,6 +148,102 @@ $ kubectl exec deploy/notsleep -- curl -s http://productpage:9080/ | grep -o "<t
...
+to wait for the gateway to be ready.
+
+If you are using a cluster that supports a [load balancer Service](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer),
+wait until the Gateway is ready:
+
+{{< text bash >}}
+$ kubectl wait --for=condition=ready gtw/bookinfo-gateway -n istio-system
+{{< /text >}}
+
+Otherwise, wait until the Gateway deployment is available:
```
frankbu, line 192: Gateway API doesn't really work on a cluster without external LB.

frankbu: See https://istio.io/latest/docs/tasks/traffic-management/ingress/ingress-control/#determining-the-ingress-ip-and-ports

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -135,6 +148,102 @@ $ kubectl exec deploy/notsleep -- curl -s http://productpage:9080/ | grep -o "<t
...
+{{< /text >}}
+
+{{< text bash >}}
+$ kubectl exec deploy/notsleep -- curl -s http://productpage:9080/ | grep -o "<title>.*</title>"
+<title>Simple Bookstore App</title>
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< /tabset >}}
```
frankbu, line 245: I would move these curl calls outside of the tab, and just set an env variable to the gateway host, which can then be used to access the gateway (Istio or k8s) in the rest of the doc.

```suggestion

{{< text bash >}}
$ export GATEWAY_URL=...
{{< /text >}}

{{< /tab >}}

{{< /tabset >}}

Test your bookinfo application, it should work with or without the gateway:

{{< text syntax=bash snip_id=verify_traffic_sleep_to_gateway >}}
$ kubectl exec deploy/sleep -- curl -s ${GATEWAY_URL}/productpage | grep -o "<title>.*</title>"
<title>Simple Bookstore App</title>
{{< /text >}}

{{< text bash >}}
$ kubectl exec deploy/sleep -- curl -s http://productpage:9080/ | grep -o "<title>.*</title>"
<title>Simple Bookstore App</title>
{{< /text >}}

{{< text bash >}}
$ kubectl exec deploy/notsleep -- curl -s http://productpage:9080/ | grep -o "<title>.*</title>"
<title>Simple Bookstore App</title>
{{< /text >}}
```

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -164,13 +277,50 @@ $ kubectl exec deploy/notsleep -- curl -s http://productpage:9080/ | grep -o "<t
...
 
+{{< /tab >}}
+
+{{< tab name="Gateway API" category-value="gateway-api" >}}
+
+{{< image link="./kiali-ambient-gatewayapi-bookinfo.png" caption="Kiali dashboard" >}}
+
+{{< /tab >}}
+
+{{< /tabset >}}
```
frankbu, line 322: I don't think we need to have 2 tabs/diagrams here. The only difference is the name of the ingress gateway, and it's not even highlighted in the diagram. Let's not highlight trivial differences in the doc.

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -31,48 +31,92 @@ _wait_for_daemonset istio-system istio-cni-node
...
+  + reviews-v1
+  + reviews-v2
+  - reviews-v3
+  "
+fi
 
 # @cleanup
 cleanup_bookinfo_sample
-snip_uninstall_uninstall_1
+if [ "$GATEWAY_API" == "true" ]; then
```
frankbu, line 112: Cleanup instructions are special. The gateway-api instructions need to be moved to gtwapi_test.sh, which will need to be added. Also, you need to define GATEWAY_API at the top of test.sh (see other doc tests, for example).

ericvn:
/ok-to-test

Have the tests run when this is un-drafted.

content/en/docs/ops/ambient/getting-started/index.md
danehans: @frankbu thanks for the review and pointer. I'll update and test.

content/en/docs/ops/ambient/getting-started/index.md
danehans: OK, I'll remove.

content/en/docs/ops/ambient/getting-started/test.sh
danehans: Will do. Thanks for all the feedback and guidance.

content/en/docs/ops/ambient/getting-started/index.md
frankbu: I see. I think that since this is still in flux, and because we will need to change a bunch of docs for this, I would leave it out of here for now and we can update it, along with all the other docs, in a followup PR. WDYT?

danehans: > See https://istio.io/latest/docs/tasks/traffic-management/ingress/ingress-control/#determining-the-ingress-ip-and-ports

The ambient doc currently uses a standard kind cluster (no MetalLB) and due to istio/istio#44035, the Gateway only surfaces the load balancer address in status. Instead of bringing in MetalLB as a dependency of the doc, I will update the PR to wait for the Gateway deployment.

content/en/docs/ops/ambient/getting-started/index.md
danehans: @frankbu after further analysis, we can not use `samples/bookinfo/gateway-api/bookinfo-gateway.yaml` without disabling ambient for the Gateway pod due to https://github.com/istio/istio/issues/43587. Let's keep the inline manifest until this issue is fixed.

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -328,6 +584,96 @@ Confirm that roughly 10% traffic from the 100 requests go to reviews-v2:
...
+  - kind: Service
+    name: reviews
+  rules:
+  - backendRefs:
+    - name: reviews-v1
+      port: 9080
+      weight: 90
+    - name: reviews-v2
+      port: 9080
+      weight: 10
```
danehans, line 663: @frankbu do you prefer I add ^ to `samples/bookinfo/gateway-api` instead of inlining?

content/en/docs/ops/ambient/getting-started/index.md
frankbu: I'm not sure I understand the problem. If it's because the gateway needs to be in a different ns from the other services, won't `kubectl apply -n istio-system -f @samples/bookinfo/gateway-api/bookinfo-gateway.yaml@` work?

content/en/docs/ops/ambient/getting-started/index.md
frankbu: Yes, the plan for `samples/bookinfo/gateway-api` was to add more yaml files as needed by docs, rather than making mirror copies of everything in `samples/bookinfo/networking`.

content/en/docs/ops/ambient/getting-started/index.md
danehans: Until https://github.com/istio/istio/issues/43587 is fixed, Gateways are incapable of terminating HBONE, e.g. running in the ambient mesh. As a workaround, this PR deploys the Gateway in the istio-system namespace (just like the classic Istio ingress gateway) so the client ztunnel > Gateway uses TCP. If the Gateway is deployed in an ambient namespace, then ambient must be disabled for the Gateway pod.

content/en/docs/ops/ambient/getting-started/index.md
danehans: > I would leave it out of here for now and we can update it, along with all the other docs, in a followup PR. WDYT

IMO it's still a good idea to wait for the Gateway deployment to be ready (for those who are quick copy/paste'ers ;-))

content/en/docs/ops/ambient/getting-started/index.md
frankbu: That makes sense, but I still don't understand why you said:

> Let's keep the inline manifest until this issue is fixed.

Why can't you use the samples yaml but just install it in the istio-system ns, like I suggested above?

content/en/docs/ops/ambient/getting-started/index.md
frankbu: > IMO it's still a good idea to wait for the Gateway deployment to be ready ...

But IIUC, @howardjohn has removed the ready condition from the gateway status now, so every gateway will never be ready.

howardjohn: We can use Programmed

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -111,8 +119,13 @@ $ kubectl apply -f @samples/sleep/notsleep.yaml@
 
 Note: `sleep` and `notsleep` are two simple applications that can serve as curl clients.
 
-Connect `productpage` to the Istio ingress gateway so you can access the bookinfo
-app from outside of the cluster:
+{{< tabset category-name="config-api" >}}
```
howardjohn, line 122: We shouldn't even document this on ambient IMO

howardjohn: I mean "classic"

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -135,6 +148,102 @@ $ kubectl exec deploy/notsleep -- curl -s http://productpage:9080/ | grep -o "<t
...
+wait until the Gateway is ready:
+
+{{< text bash >}}
+$ kubectl wait --for=condition=ready gtw/bookinfo-gateway -n istio-system
+{{< /text >}}
+
+Otherwise, wait until the Gateway deployment is available:
+
+{{< text bash >}}
+$ kubectl wait --for=condition=available deploy/bookinfo-gateway-istio -n istio-system
```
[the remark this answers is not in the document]
howardjohn: And we should fix the implementation if that isn't possible

content/en/docs/ops/ambient/getting-started/index.md
danehans: OK, I'll do that.

content/en/docs/ops/ambient/getting-started/index.md
danehans: >But IIUC, @howardjohn has removed the ready condition from the gateway status now, so every gateway will never be ready.

Yes, but the Gateway's deployment will be ready and that's the `wait` that I was referring to.

content/en/docs/ops/ambient/getting-started/index.md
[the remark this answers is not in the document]
danehans: > -1, Deployment should be an implementation detail. We should only wait on Gateway status attributes

Agreed, I would normally just wait for the Gateway to be "Ready=true". I will update to use "Programmed=true".

content/en/docs/ops/ambient/getting-started/index.md
linsun: I disagree - nearly all users are still using classic Istio APIs, we don't want to force them to move to new k8s gateway API for ambient.

howardjohn: We should push them to - but not force, IMO. We should encourage best practices

frankbu: My first instinct was, same a @howardjohn, that we should only use gateway-api in this doc, since it's all new stuff. But @linsun reminded me that we need to be careful to make it clear that using ambient or not is independent of whether or not one uses gateway-api or classic config. We don't want the getting started with ambient doc to only show users how to do it if they also use an unrelated other new feature (gateway-api). To encourage users (esp new Istio users) to use gateway-api, we should add in this doc the gateway-api boilerplate tip that mentions that the doc shows how to do it with either API, but gateway-api is the future one (like we are currently doing in every other Istio doc using gateways/routes).

linsun: I think what @frankbu suggested is a reasonable middle ground ^^

content/en/docs/ops/ambient/getting-started/gtwapi_test.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
frankbu, line 1: Maybe a miscommunication, but only the @cleanup instructions need to be in this file. The bulk of the test should be included from test.sh, which has if checks around the parts that are different depending on gateway-api or not (as you had done previously).

Here's an example of what a typical gtwapi_test.sh looks like: https://github.com/istio/istio.io/blob/master/content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh

content/en/docs/ops/ambient/getting-started/index.md
danehans: @frankbu I created https://github.com/istio/istio/pull/44097 to add ^ manifest to Istio.

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -111,17 +119,106 @@ $ kubectl apply -f @samples/sleep/notsleep.yaml@
...
 {{< text bash >}}
 $ kubectl apply -f @samples/bookinfo/networking/bookinfo-gateway.yaml@
 {{< /text >}}
 
-Test your bookinfo application, it should work with or without the gateway. Note: you can replace `istio-ingressgateway.istio-system` below with its load balancer IP (or hostname) if it has one:
+Set the environment variables for the Istio ingress gateway:
+
+{{< text bash >}}
+$ export GATEWAY_HOST=istio-ingressgateway
+$ export GATEWAY_SERVICE_ACCOUNT="$GATEWAY_HOST"-service-account
```
frankbu, line 138: This isn't really the hostname (i.e., without the .namespace). Since we are planning to move the k8s gateway to the default ns in the future, you should really include the ns in the env vars anyway.

```suggestion
$ export GATEWAY_HOST=istio-ingressgateway.istio-system
$ export GATEWAY_SERVICE_ACCOUNT=ns/istio-system/sa/istio-ingressgateway-service-account
```

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -111,17 +119,106 @@ $ kubectl apply -f @samples/sleep/notsleep.yaml@
...
+{{< /text >}}
+
+You have now created an HTTPRoute configuration for the `productpage` service containing one route rule that allows
+traffic for path `/productpage`.
+
+Set the environment variables for the Kubernetes gateway:
+
+{{< text bash >}}
+$ export GATEWAY_HOST=bookinfo-gateway-istio
+$ export GATEWAY_SERVICE_ACCOUNT="$GATEWAY_HOST"
```
frankbu, line 211:
```suggestion
$ export GATEWAY_HOST=bookinfo-gateway-istio.istio-system
$ export GATEWAY_SERVICE_ACCOUNT=ns/istio-system/sa/bookinfo-gateway-istio
```

content/en/docs/ops/ambient/getting-started/index.md
danehans: I updated the doc to check for the gtw's "Programmed=true" condition.

content/en/docs/ops/ambient/getting-started/index.md
danehans: IMHO `samples/bookinfo/gateway-api/bookinfo-gateway.yaml` should be updated to only include the Gateway and a new `samples/bookinfo/gateway-api/bookinfo-httproute.yaml` should be introduced for the HTTPRoute. This will provide additional flexibility to work with the samples, e.g. cross namespace routing. Thoughts @frankbu?

danehans: https://github.com/istio/istio/pull/44101 adds a separate HTTPRoute for bookinfo.

danehans:
CI for this PR is failing b/c it requires [PR #44097](https://github.com/istio/istio/pull/44097) (merged last Friday) and CI builds Istio using SHA 2835a206ac (from March 18). I reviewed the Istio.io Makefile and Readme but I don't see how to update istio.io CI to use a newer Istio release. Any suggestions?

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -239,7 +327,7 @@ waypoint default/bookinfo-productpage applied
 {{< /text >}}
 
 View the `productpage` waypoint proxy status; you should see the details of the gateway
-resource with `Ready` status:
+resource with `Programmed=True` status:
```
danehans, line 330: The status change is due to https://github.com/istio/istio/pull/43975.

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -111,17 +115,106 @@ $ kubectl apply -f @samples/sleep/notsleep.yaml@
...
+The Gateway is configured to listen for HTTP traffic on port 80 and allow
+[HTTPRoutes](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRoute) from any
+namespace to bind to the listener.
+
+Creating a Kubernetes Gateway resource deploys an associated
+[proxy service](/docs/tasks/traffic-management/ingress/gateway-api/#automated-deployment), so run the following command
+to wait for the gateway to be ready:
+
+{{< text bash >}}
+$ kubectl wait --for=condition=programmed gtw/bookinfo-gateway -n istio-system
```
danehans, line 172: "Programmed" condition is due to https://github.com/istio/istio/pull/43975.

danehans:
I broke out changes from this PR into https://github.com/istio/istio.io/pull/12955 to get CI fixed.

danehans:
Moved to draft until https://github.com/istio/istio/pull/44139 is merged and CI bumps Istio dep to the tip of master.

danehans:
Rebased on https://github.com/istio/istio.io/pull/12955.

content/en/docs/ops/ambient/getting-started/index.md
danehans: - This PR needs to continue inlining the Gateway instead of using `samples/bookinfo/gateway-api` b/c the HTTPRoute in this doc is created in a different ns. This requires the allowed routes `from: All `for the Gateway instead of `from: Same`.
- This PR needs to continue inlining the HTTPRoute instead of using `samples/bookinfo/gateway-api` b/c the HTTPRoute in this doc is created in a different ns. This requires the parentRef to specify `namespace` as the ns of the Gateway (`istio-system`).

frankbu: In that case, I think the new gateway-httproute.yaml that you added in bookinfo samples is not really useful. Should we revert?

I still would prefer using a simple sed script to update `samples/bookinfo/gateway-api/bookinfo-gateway.yaml`, instead of adding all the noise in this doc, since it's really just a temporary fix until the ztunnel issue is fixed.
Here's a version of the command that will add the namespace in both places:
```
sed '/name: bookinfo-gateway/a\
  namespace: istio-system\
' @samples/bookinfo/gateway-api/bookinfo-gateway.yaml@ | kubectl apply -f -
```

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -111,17 +115,106 @@ $ kubectl apply -f @samples/sleep/notsleep.yaml@
 
 Note: `sleep` and `notsleep` are two simple applications that can serve as curl clients.
 
-Connect `productpage` to the Istio ingress gateway so you can access the bookinfo
-app from outside of the cluster:
+{{< tabset category-name="config-api" >}}
+
+{{< tab name="Istio classic" category-value="istio-classic" >}}
+
+Create an Istio [Gateway](/docs/reference/config/networking/gateway/) and
```
howardjohn, line 122: do they need to install the deployment as well? I think we turned it off by default in ambient (or did we not yet?)

howardjohn (commented):
looks good

content/en/docs/ops/ambient/getting-started/index.md
frankbu: Do you mean that the intent is to turn off managed gateways in ambient mode? Since the test is passing, I assume that isn't the case currently, but is that going to change?

content/en/docs/ops/ambient/getting-started/gtwapi_test.sh
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+GATEWAY_API="true"
+
+# @setup profile=none
+source "content/en/docs/ops/ambient/getting-started/test.sh"
+
+# @cleanup
```
frankbu, line 23: You need to add the full set of cleanup instructions for the gateway API test here, or nothing will be cleaned up in the premature exit (error) case.

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -15,12 +15,21 @@
...
 set -u
 set -o pipefail
 
+# Helper function to initialize KUBECONFIG_FILES and KUBE_CONTEXTS
+_set_kube_vars
+
+# Kubernetes Gateway API CRDs are required by waypoint proxy.
+source "tests/util/gateway-api.sh"
+install_gateway_api_crds "${KUBE_CONTEXTS[0]}"
+
```
frankbu, line 32: None of this should be needed if done in gtw_test.sh.
```suggestion
```

content/en/docs/ops/ambient/getting-started/index.md
howardjohn: No I mean we don't auto-deploy an istio-ingressgateway deploymnent in `istioctl install --set profile=ambient`. But maybe that was a "we should do this" not "we did".

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -351,3 +458,37 @@ The label to instruct Istio to automatically include applications in the `defaul
 {{< text bash >}}
 $ kubectl label namespace default istio.io/dataplane-mode-
 {{< /text >}}
+
+{{< tabset category-name="config-api" >}}
```
frankbu, line 462: This whole tab section should not be needed. It should be done in the bookinfo sample cleanup script used in the first cleanup instruction, above.

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -351,3 +458,37 @@ The label to instruct Istio to automatically include applications in the `defaul
...
+$ kubectl delete -f httproute/reviews
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< /tabset >}}
+
+If you installed the Gateway API CRDs, remove them:
+
+{{< text syntax=bash snip_id=none >}}
```
frankbu, line 492: You should call this in the not gateway-api cleanup.
```suggestion
{{< text syntax=bash snip_id=delete_crds >}}
```

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -76,3 +96,4 @@ cleanup_bookinfo_sample
 snip_uninstall_uninstall_1
 snip_uninstall_uninstall_2
 snip_uninstall_uninstall_3
+remove_gateway_api_crds "${KUBE_CONTEXTS[0]}"
```
frankbu, line 99: The @cleanup section should be only the not gateway-api cleanup and must be guarded because otherwise they will be called at the end of the test which is included in gtw_test.sh, instead of during the gtw_test.sh cleanup.
```
@cleanup
if [ "$GATEWAY_API" != "true" ]; then
    cleanup_bookinfo_sample
    snip_uninstall_uninstall_1
    snip_uninstall_uninstall_2
    snip_uninstall_uninstall_3
    snip_delete_crds
fi
```

content/en/docs/ops/ambient/getting-started/index.md
danehans: >In that case, I think the new gateway-httproute.yaml that you added in bookinfo samples is not really useful. Should we revert?

Yes, here's the PR: https://github.com/istio/istio/pull/44172

> I still would prefer using a simple sed script to update samples/bookinfo/gateway-api/bookinfo-gateway.yaml, instead of adding all the noise in this doc, since it's really just a temporary fix until the ztunnel issue is fixed.

This is not a temporary fix. The Gateway in this doc is not only running in a different ns than `samples/bookinfo/gateway-api/bookinfo-gateway.yaml`, the spec differs (allows xRoute binding from all ns's). The `samples/bookinfo/gateway-api/bookinfo-gateway.yaml` also creates an HTTPRoute in the istio-system ns, when the HTTPRoute needs to reside in the default namespace. It seems like the flow of this doc is consistent with the [existing Gateway API](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/) doc that inlines the Gateway and HTTPRoute.

content/en/docs/ops/ambient/getting-started/test.sh
danehans: Gateway API CRDs are installed regardless of what APIs (Istio Classic|K8s Gateway) are being used for ingress ([xref](https://preliminary.istio.io/latest/docs/ops/ambient/getting-started#l7-authorization-policy)). `cleanup_bookinfo_sample` has separate cleanup steps based on `GATEWAY_API` and the other snips are required regardless of what APIs (Istio Classic|K8s Gateway) are being used.

content/en/docs/ops/ambient/getting-started/index.md
danehans: As suggested in a different thread of this PR, the doc uses the gateway api [boilerplate](https://github.com/istio/istio.io/blob/master/content/en/boilerplates/gateway-api-support.md). Because of this, the test sources `tests/util/gateway-api.sh` to install the CRDs by calling `install_gateway_api_crds` and remove the CRDs by calling `remove_gateway_api_crds`.

content/en/docs/ops/ambient/getting-started/index.md
danehans: > do they need to install the deployment as well? I think we turned it off by default in ambient (or did we not yet?)

@howardjohn a deployment does not need to be cerated by the user since the Gateway auto creates the deployment and service. Note that the Gateway in this doc is not running on ambient (istio-system ns).

> No I mean we don't auto-deploy an istio-ingressgateway deploymnent in istioctl install --set profile=ambient

This would be a separate change to the doc since this PR is for adding Gateway API steps. Istio is still deploying ingress gateways for the Ambient profile:
```
$ echo $TAG
1.18-alpha.42c42563c082e6e4b457d5b94448be9cd6c43f26

$ istioctl install --set profile=ambient --set hub=$HUB --set tag=$TAG
This will install the Istio 1.18.0 ambient profile with ["Istio core" "Istiod" "CNI" "Ztunnel" "Ingress gateways"] components into the cluster. Proceed? (y/N) y
✔ Istio core installed
✔ Istiod installed
✔ Ingress gateways installed
✔ CNI installed
✔ Ztunnel installed
✔ Installation complete                                                                                                                                         Making this installation the default for injection and validation.
```

I created https://github.com/istio/istio/issues/44030 to address this issue but I don't see a PR against it. If you'd like me to work on it, please assign me.

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -224,12 +318,6 @@ command terminated with exit code 56
 ### L7 Authorization Policy
 
 Using the Kubernetes Gateway API, you can deploy a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} for the `productpage` service that uses the `bookinfo-productpage` service account. Any traffic going to the `productpage` service will be mediated, enforced and observed by the Layer 7 (L7) proxy.
-Install Kubernetes Gateway CRDs, which don’t come installed by default on most Kubernetes clusters:
-
-{{< text bash >}}
-$ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-  { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.1" | kubectl apply -f -; }
-{{< /text >}}
```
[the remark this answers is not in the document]
danehans: This is already accomplished at the beginning of the doc using the gateway api boilerplate:
```
1. This guide includes tasks that use a `Gateway`.

   {{< boilerplate gateway-api-support >}}
```

content/en/docs/ops/ambient/getting-started/test.sh
danehans: Gateway API CRDs are needed even when using Istio Classic APIs (required by waypoint).

content/en/docs/ops/ambient/getting-started/index.md
howardjohn: >a deployment does not need to be cerated by the user since the Gateway auto creates the deployment and service. Note that the Gateway in this doc is not running on ambient (istio-system ns).

I was specifically commenting on the "istio classic" tab

frankbu: Sorry for creating the confusion ... I didn't notice this comment was on the classic instruction :(

@howardjohn I thought that the ambient profile is temporary and will be removed in the future when ambient mode will be enabled by default. If that's true, users would use the minimal profile if they don't want the ingress gateway installed, same as non-ambient mode. Maybe, at some time in the future, the minimal profile will become default, but that seems like a breaking change?

content/en/docs/ops/ambient/getting-started/index.md
frankbu: > This is not a temporary fix.

Yes it is. As soon as the ztunnel issue is fixed we will just install both the gateway and route in the default ns by simply applying bookinfo-gateway.yaml. The fact that we can't do that today is a bug.

> It seems like the flow of this doc is consistent with the [existing Gateway API](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/) doc that inlines the Gateway and HTTPRoute.

The difference is that doc is teaching people about Gateway API so showing them the resources explicitly makes sense. In this doc, the bookinfo Gateway and HTTPRoute are just used for testing, so we don't need/want to dwell on the details here.

content/en/docs/ops/ambient/getting-started/index.md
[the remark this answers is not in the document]
frankbu: Problem is that the boilerplate is just a tip, telling the users that the Gateway API is optional.
I thing there are 2 choices to fix:

1. Change the earlier instruction to simply tell users to install the CRDs, period. Separately, add a tip about the option of using Gateway API or classic config for bookinfo.
2. My suggestion here

content/en/docs/ops/ambient/getting-started/test.sh
frankbu: Yes, I know, but my first instinct was to handle the classic case with explicit snips in the doc, which you would call instead of using the util functions.

That said, maybe fixing the util function to be usable for ambient-only, as you're doing, is a better since there will be more ambient docs coming which won't have explicit snips for the CRDs.

content/en/docs/ops/ambient/getting-started/index.md
danehans: > If that's true, users would use the minimal profile if they don't want the ingress gateway installed, same as non-ambient mode.

@frankbu ^ users would need to manually install CNI and ztunnel components. When https://github.com/istio/istio/issues/44030 is fixed, the ambient profile will install istiod, cni, and ztunnel. Users can then create an Istio Gateway/VirtualService or a k8s Gateway/HTTPRoute for ingress.

frankbu: > @frankbu ^ users would need to manually install CNI and ztunnel components.

By `If that's true, ...` I meant if we make everything in the current ambient profile enabled by default in the future.

> When https://github.com/istio/istio/issues/44030 is fixed, the ambient profile will install istiod, cni, and ztunnel. Users can then create an Istio Gateway/VirtualService or a k8s Gateway/HTTPRoute for ingress.

Not quite. If they want to use Gateway/VirtualService, they will also need to manually deploy the istio-ingressgateway.

content/en/docs/ops/ambient/getting-started/index.md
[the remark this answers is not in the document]
danehans: > Problem is that the boilerplate is just a tip, telling the users that the Gateway API is optional.

The `gateway-api-support` boilerplate is more than just a tip. It includes instructions for installing the CRDs. However, I will implement your feedback by including specific instructions to install the CRDs and use the `gateway-api-future` and `gateway-api-choose` boilerplates.

content/en/docs/ops/ambient/getting-started/test.sh
danehans: @frankbu thanks for the additional insight. It now makes sense to me.

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -52,27 +64,34 @@ _verify_contains snip_verify_traffic_sleep_to_ingress "$snip_verify_traffic_slee
...
 # @cleanup
-cleanup_bookinfo_sample
-snip_uninstall_uninstall_1
-snip_uninstall_uninstall_2
-snip_uninstall_uninstall_3
+if [ "$GATEWAY_API" != "true" ]; then
+    snip_uninstall_uninstall_1
+    snip_uninstall_uninstall_2
+    snip_uninstall_uninstall_3
+    cleanup_bookinfo_sample
```
frankbu, line 95:
```suggestion
    samples/bookinfo/platform/kube/cleanup.sh
```

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -87,6 +87,16 @@ Follow these steps to get started with ambient:
...
 
+1.  Install Kubernetes Gateway CRDs, which don’t come installed by default on most Kubernetes clusters:
+
+    {{< text bash >}}
+    $ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
+      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.1" | kubectl apply -f -; }
+    {{< /text >}}
+
+    {{< boilerplate gateway-api-future >}}
+    {{< boilerplate gateway-api-choose >}}
```
frankbu, line 98: I'd highlight this better by wrapping in a tip
```suggestion
    {{< tip >}}
    {{< boilerplate gateway-api-future >}}
    {{< boilerplate gateway-api-choose >}}
    {{< /tip >}}
```

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -87,6 +87,16 @@ Follow these steps to get started with ambient:
     ztunnel          1         1         1       1            1           <none>                   82s
     {{< /text >}}
 
+1.  Install Kubernetes Gateway CRDs, which don’t come installed by default on most Kubernetes clusters:
```
frankbu, line 90: Move this up to the first step?

danehans: The 1st step is to install the kind cluster, then install Istio, and verify the istio install. I think installing the CRDs is in the correct order.

frankbu (approved):

PR #12954, opened 2023-03-27
Reformat dates to use a consistent format

dhawton:

Continuing discussion from https://github.com/istio/istio.io/pull/12919#discussion_r1146431994

Reformatted dates in support status of Istio releases to use the same format. Formatting to a short month keeps things on a single line, except when (Expected) is added to the End of Life fields. Opted for short month names over yyyy-mm-dd as the it helps readability regardless of locale date formats, and maintains readability for individuals not used to reading that format.

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/data/compatibility/supportStatus.yml
+++ b/data/compatibility/supportStatus.yml
@@ -9,20 +9,20 @@
   testedK8sVersions:
 - version: "1.17"
   supported: "Yes"
-  releaseDate: "February 14, 2023"
+  releaseDate: "Feb 14, 2023"
   eolDate: "~Sept 2023 (Expected)"
   k8sVersions: ["1.23", "1.24", "1.25", "1.26"]
   testedK8sVersions: ["1.16", "1.17", "1.18", "1.19", "1.20", "1.21", "1.22"]
 - version: "1.16"
   supported: "Yes"
-  releaseDate: "November 15, 2022"
-  eolDate: "~June 2023 (Expected)"
+  releaseDate: "Nov 15, 2022"
+  eolDate: "~Jun 2023 (Expected)"
   k8sVersions: ["1.22", "1.23", "1.24", "1.25"]
   testedK8sVersions: ["1.16", "1.17", "1.18", "1.19", "1.20", "1.21"]
 - version: "1.15"
   supported: "Yes"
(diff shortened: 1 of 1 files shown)


data/compatibility/supportStatus.yml
```diff
@@ -9,20 +9,20 @@
   testedK8sVersions:
 - version: "1.17"
   supported: "Yes"
-  releaseDate: "February 14, 2023"
+  releaseDate: "Feb 14, 2023"
   eolDate: "~Sept 2023 (Expected)"
```
ericvn, line 13:
```suggestion
  eolDate: "~Sep 2023 (Expected)"
```

ericvn (commented):
Looks like on that didn't get shortened to 3 characters.

data/compatibility/supportStatus.yml
dhawton: Yup I missed that one. Thanks

ericvn:
Looks good to me at https://deploy-preview-12954--preliminary-istio.netlify.app/latest/docs/releases/supported-releases/ both on my phone and computer.

ericvn:
Adding a period does not seem to affect the flow (tried locally):
Screenshot 2023-03-27 at 1 23 04 PM

ericvn:
Unfortunately, looking at the deploy/netlify output, a few of the lines are now split with the year on the following line. I'm wondering what happens if someone extends the `Release date` in layouts/shortcodes/support_status_table.html with a space. Does that get picked up and widen the table column?

dhawton:
> Unfortunately, looking at the deploy/netlify output, a few of the lines are now split with the year on the following line. I'm wondering what happens if someone extends the `Release date` in layouts/shortcodes/support_status_table.html with a space. Does that get picked up and widen the table column?

What about leaving the periods off? It seems pretty common these days to not add the period with shortened month names. Might also be "better" than having a "hacky" space added to the column name. In my env without the periods, only 1.11's EOL date drops to another line (likely just because of font) or the `(Expected)` dates.

ericvn (approved):
Thanks for handling all the comments on this one.

PR #12955, opened 2023-03-27
Updates Gateway Status Checking

danehans:

Updates the Gateway status condition check for the ambient getting started guide due to https://github.com/istio/istio/pull/43975.

- [X] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


danehans:
xref: https://github.com/istio/istio.io/pull/12940

ericvn:
/ok-to-test

But you will need to run the `go get` and `go mod tidy` and submit the go.*  changes here.

danehans:
Looking into why `TestDocs/tasks/traffic-management/ingress/ingress-sni-passthrough/gtwapi_test.sh` is failing.

danehans:
Job `doc.test.profile_minimal_istio.io` is failing due to https://github.com/istio/istio.io/pull/12940#issuecomment-1485448799.

danehans:
For the `doc.test.profile_minimal_istio.io` failing test, I have repro'd the issue. The Gateway's service does not get an IP assigned. Since the `_wait_for_gateway` func only waits for the `Programmed=true" condition, this check passes and the `INGRESS_HOST` env var is unassigned.

```
 build-tools:/work$ kubectl get gtw/mygateway -o yaml
kind: Gateway
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"gateway.networking.k8s.io/v1beta1","kind":"Gateway","metadata":{"annotations":{},"name":"mygateway","namespace":"default"},"spec":{"gatewayClassName":"istio","listeners":[{"allowedRoutes":{"namespaces":{"from":"All"}},"hostname":"nginx.example.com","name":"https","port":443,"protocol":"TLS","tls":{"mode":"Passthrough"}}]}}
  creationTimestamp: "2023-03-27T18:53:38Z"
  generation: 1
  name: mygateway
  namespace: default
  resourceVersion: "2638"
  uid: 7bed5789-bb15-4931-8c54-c0318ffbdb73
spec:
  gatewayClassName: istio
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    hostname: nginx.example.com
    name: https
    port: 443
    protocol: TLS
    tls:
      mode: Passthrough
status:
  conditions:
  - lastTransitionTime: "2023-03-27T18:53:38Z"
    message: 'Assigned to service(s) mygateway-istio.default.svc.cluster.local:443,
      but failed to assign to all requested addresses: address pending for hostname
      "mygateway-istio.default.svc.cluster.local"'
    observedGeneration: 1
    reason: NoResources
    status: "False"
    type: Accepted
  - lastTransitionTime: "2023-03-27T18:53:38Z"
    message: Resource programmed
    observedGeneration: 1
    reason: Programmed
    status: "True"
    type: Programmed
...

build-tools:/work$ kubectl get svc
NAME              TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)                         AGE
kubernetes        ClusterIP      10.96.0.1      <none>        443/TCP                         19m
my-nginx          ClusterIP      10.96.45.85    <none>        443/TCP                         2m32s
mygateway-istio   LoadBalancer   10.96.68.111   <pending>     15021:31564/TCP,443:31096/TCP   2m28s
```

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -241,22 +241,22 @@ waypoint default/bookinfo-productpage applied
 View the `productpage` waypoint proxy status; you should see the details of the gateway
 resource with `Ready` status:
 
-{{< text bash >}}
+{{< text syntax=bash snip_id=none >}}
```
danehans, line 244: No longer needed since test is now using Test is now using `_wait_for_gateway`.

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/index.md
```diff
@@ -304,10 +304,15 @@ to set the `SECURE_INGRESS_PORT` and `INGRESS_HOST` environment variables.
 
 {{< tab name="Gateway API" category-value="gateway-api" >}}
 
+Wait for the Gateway to be accepted:
+
+{{< text syntax=bash snip_id=none >}}
```
danehans, line 309: Snip not needed since the test is using `_wait_for_gateway`.

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/snips.sh
```diff
@@ -235,7 +235,6 @@ EOF
 }
 
 snip_configure_an_ingress_gateway_5() {
-kubectl wait --for=condition=programmed gtw mygateway
```
danehans, line 238: Not needed since test calls `_wait_for_gateway`.

content/en/docs/ops/ambient/getting-started/index.md
frankbu: Since the primary purpose of doc tests (with side benefit of testing Istio)  is to test that the documented instructions work as documented, a general best practice is to call snips wherever we can. The util functions are intended for docs that don't include explicit instructions, e.g., docs that just say before-you-begin install bookinfo, or something like that.

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/index.md
```diff
@@ -304,10 +304,15 @@ to set the `SECURE_INGRESS_PORT` and `INGRESS_HOST` environment variables.
...
+$ kubectl wait --for=condition=accepted gtw mygateway
```
frankbu, line 310: The correct semantics of gateway accepted vs programmed vs ready is still not agreed in the upstream gateway-api project, so I think we should wait for resolution before updating any docs.
https://github.com/kubernetes-sigs/gateway-api/issues/1877

danehans: @frankbu PTAL at https://github.com/istio/istio.io/pull/12955#issuecomment-1485711217. The way that Gateway status is currently implemented `Programmed=True` can be surfaced when an address is unassigned. This is causing tests that set `INGRESS_HOST` to fail. As you mention, kubernetes-sigs/gateway-api#1877 is attempting to resolve ambiguity in the conditions. Maybe the best path forward is to check for the presence of both status conditions until this is resolved?

danehans:
Still seeing issues with the Gateway address not getting populated even though the status conditions are `True`. Converted to draft while I figure out why this is happening.

Xunzhuo (approved):
LGTM

content/en/docs/tasks/traffic-management/ingress/ingress-sni-passthrough/index.md
frankbu: @danehans The problem is that there are 14 other docs that are broken by the current implementation. Fixing the implementation (to at least align with the best guess of where the spec is going) seems like a better approach than hacking all the docs temporarily. (Btw, the fact that the current impl has accepted not true until the address is assigned is an obvious bug since according to the spec accepted just means the config is syntactically valid.)

ericvn:
Noting also that @howardjohn mentioned in Slack that the underlying change may be reverted so we may need to revert some of our changes as well. Biggest unknown is when and that drives when we can actually update the version of Istio we are testing against.

danehans:
xref: https://github.com/istio/istio/pull/44139 to remove wait conditions for Gateway status.

content/en/docs/ops/ambient/getting-started/snips.sh
```diff
@@ -119,18 +119,6 @@ snip_l7_authorization_policy_3() {
 kubectl get gtw bookinfo-productpage -o yaml
 }
 
-! read -r -d '' snip_l7_authorization_policy_4 <<\ENDSNIP
```
danehans, line 122: __Note:__ This snip is not currently being used by the test.

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -247,16 +247,16 @@ $ kubectl get gtw bookinfo-productpage -o yaml
 
 Verify that the waypoint proxy status is ready:
 
-{{< text plaintext >}}
+{{< text syntax=plaintext snip_id=none >}}
```
danehans, line 250: __Note:__ This snip is not currently being used by the test.

Xunzhuo (approved):
LGTM

ericvn (approved):

frankbu (approved):

PR #12962, opened 2023-03-28
Update the test reference and skywalking test

ericvn:

This PR will supersede https://github.com/istio/istio.io/pull/12940 updating the istio reference to the latest (there needs to be one more commit once https://github.com/istio/istio/pull/44139 merges (will need to also wait for the images to be created so the tests can use them)).

I ran the tests locally against a local build of the istio PR, and there was one test that needed minor updating.


ericvn:
Possibly a dup of https://github.com/istio/istio.io/pull/12955

ericvn:
Looks like a failure in `tasks/observability/distributed-tracing/skywalking/test.sh/test.sh` causing the test to timeout.

ericvn:
Will keep this PR around and try the commit after John's to verify it causes a timeout on the Skywalking test..

ericvn:
Move to an istio commit that includes the revert of the skywalking port and remove the port change from the test.

ericvn:

mirroring gateway test failed

ericvn:
Looks like we are good with the current istio commit.

Issue #12797, opened 2023-02-27
add e2e for https://github.com/istio/istio.io/pull/12743


zirain:
Consider introducing `loki` to make it easy to verify.

PR #13036, opened 2023-04-10
add tests for `tasks/observability/logs/telemetry-api`

zirain:
fixes: https://github.com/istio/istio.io/issues/12797


ericvn:
The go.* updates will prevent this from passing.

Issue #11720, opened 2022-08-19
Add proper cleanup instructions to all docs


frankbu:
The cleanup instructions in various docs are missing. Some docs do complete uninstall, some do partial, and some have no cleanup instructions at all. The ones that have partial or no instructions have cleanup commands in the test.sh (but many of those are only partially correct, just enough to pass the before/after snapshot diff).

We need do a pass through all of the docs to make sure they include proper cleanup instructions and then cleanup the tests to call the cleanup snips instead coding the cleanup commands themselves.

frankbu:
These are the docs that have no cleanup instructions at all:

```
content/en/docs/setup/install/multicluster/
content/en/docs/ops/configuration/mesh/config-resource-ready/index.md
content/en/docs/ops/configuration/security/security-policy-examples/index.md
content/en/docs/ops/configuration/telemetry/envoy-stats/index.md
content/en/docs/ops/diagnostic-tools/istioctl-analyze/index.md
```

Most are ops docs, which might be OK.

johnma14:
Linking related issues here for tracking purposes:

- [Add cleanup instructions to istioctl analyze page](https://github.com/istio/istio.io/issues/12759)
- [Add cleanup instructions for Plugin CA Certificates & Custom CA integration using Kubernetes CSR](https://github.com/istio/istio.io/issues/12695)
-

jaellio:
@frankbu Can this issue be closed?

PR #13076, opened 2023-04-20
fix(): add missing cleanup instructions

jaellio:
Part of #11720

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience


rvennam:
Personally, I don't think uninstalling all of Istio at the end makes sense for every task. Cleanup should just remove the resources created that were specific to that task. wdyt?

jaellio:
> Personally, I don't think uninstalling all of Istio at the end makes sense for every task. Cleanup should just remove the resources created that were specific to that task. wdyt?

Good point, I'll take another look through these cases and only include uninstall instructions when it makes sense and otherwise only include instructions to bring the environment back to its initial state.

keithmattix (approved):

ericvn (approved):

ericvn:
Thanks @jaellio

PR #13081, opened 2023-04-20
multi-cluster with multi-network setup flow fix

raviverma-ai:

Issue Description:
Installation of east-west gateway fails on remote cluster with primary-remote multi-network setup(Any combination of AKS, EKS and GKE).

Resolution:
This issue is caused by not having proper certificate on remote cluster. Hence, remote cluster secret to access remote API server need to be deployed on primary cluster before installing east-west gateway on remote cluster.

- [ ] Ambient
- [X] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/multicluster/primary-remote_multi-network/index.md
+++ b/content/en/docs/setup/install/multicluster/primary-remote_multi-network/index.md
@@ -185,34 +185,6 @@ Apply the configuration to `cluster2`:
 $ istioctl install --context="${CTX_CLUSTER2}" -f cluster2.yaml
 {{< /text >}}
 
-## Install the east-west gateway in `cluster2`
-
-As we did with `cluster1` above, install a gateway in `cluster2` that is dedicated
-to east-west traffic and expose user services.
-
-{{< text bash >}}
-$ @samples/multicluster/gen-eastwest-gateway.sh@ \
-    --mesh mesh1 --cluster cluster2 --network network2 | \
-    istioctl --context="${CTX_CLUSTER2}" install -y -f -
-{{< /text >}}
-
-Wait for the east-west gateway to be assigned an external IP address:
-
-{{< text bash >}}
-$ kubectl --context="${CTX_CLUSTER2}" get svc istio-eastwestgateway -n istio-system
-NAME                    TYPE           CLUSTER-IP    EXTERNAL-IP    PORT(S)   AGE
-istio-eastwestgateway   LoadBalancer   10.0.12.121   34.122.91.98   ...       51s
-{{< /text >}}
-
-## Expose services in `cluster2`
-
-As we did with `cluster1` above, expose services via the east-west gateway.
-
-{{< text bash >}}
-$ kubectl --context="${CTX_CLUSTER2}" apply -n istio-system -f \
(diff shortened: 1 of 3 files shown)


content/en/docs/setup/install/multicluster/primary-remote_multi-network/index.md
```diff
@@ -242,6 +214,34 @@ $ istioctl x create-remote-secret \
     kubectl apply -f - --context="${CTX_CLUSTER1}"
 {{< /text >}}
 
+## Install the east-west gateway in `cluster2`
```
frankbu, line 217: This change isn't necessary, as confirmed by the fact that this is a [tested document](https://github.com/istio/istio.io#testing-document-content). The e/w gateway can be created before or after the cluster is connected. The only difference is that the deployment will not be ready until the cluster is attached to the control plane.

Arhell (commented):
/ok-to-test

content/en/docs/setup/install/multicluster/primary-remote_multi-network/index.md
raviverma-ai: Hi Frank,
I had tested official mentioned steps on real cluster with all combination of GKE, EKS, AKS and DO, however, installation of east-west gateway on remote cluster always fails with following error.
image

and ReplicaSet of east-west gateway fails with following error.
image

However, if we deploy remote secret(on primary cluster) before installing east-west gateway on remote cluster, installation goes smoothly and works as expected.

Furthermore, we had been reported issues on similar lines with minikube setup as well, wherein, response was getting received only from primary cluster deployments. And this issue got resolved by applying remote secret before installing east-west gateway on remote cluster.

Tested with:
Istio version: 1.17.1 and 1.17.2
Kubernetes Version: 1.25 and 1.26

Please let me know if I am missing something.

frankbu: Hmm, the 5min timeout error seems to imply that the problem is that you are waiting too long between deploying the gateway and setting the remote secret - i.e., it only work if it's done quickly? If that's the case, maybe it is better to switch the order. Can you confirm that the problem is because there's too much delay between deploying the gateway deploy and the remote secret, but it does work if you do it faster.

raviverma-ai: Could it be related to real cluster vs kind/minikube setup?

frankbu: I have no idea what the explanation is, it really makes no sense.

Anyway, I think your change is probably worth making, regardless, because it doesn't have any expectation that the steps will happen within 5 mins.

You should, however, update the test.sh to change the order that we test the snips, so they are being called in the same order as doc readers will run them.

raviverma-ai: Hi Frank,

I have updated tesh.sh accordingly. Please review.

frankbu (approved):
LGTM, Thanks!

PR #13087, opened 2023-04-21
Create generic args.md with common args consumed by test scripts

### LINKED ISSUE #13085: Create a generic .md to access args
## Issue
Currently  https://github.com/istio/istio.io/blob/master/content/en/boilerplates/gateway-api-version.md is used is used as boilerplate snippet to get access to `k8s_gateway_api_version`. 

However, also `full_version` and `previous_version` args defined in https://github.com/istio/istio.io/blob/master/data/args.yml need to be accessed in `test.sh`. 

## Proposal
Update the filename to a more generic `args.md` and use the same approach to enable access to other args, including `version_full` and `version_previous`. So that they can be used in `test.sh` like `snip_versions_istio_full` etc. This allow to make necessary changes to prevent using hardcoded version values in the doc (see https://preliminary.istio.io/latest/docs/setup/upgrade/canary/), which would be updated automatically at every release.

## Reference
https://github.com/istio/istio.io/pull/13050  relates to the above proposal in preventing hardcoded version values (for [install/operator](https://preliminary.istio.io/latest/docs/setup/install/operator/) page). But I think a more generic approach is needed.

SRodi:
* create generic boilerplates/args.md
* include gw_api_version, full_version, previous_version
* update install/operator/test.sh to source and use new vars
* update test/util/gateway-api.sh to source and use new gw var
* remove gateway-api-version.md

Fixes #13085

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/operator/test.sh
+++ b/content/en/docs/setup/install/operator/test.sh
@@ -22,9 +22,11 @@ set -e
 set -u
 set -o pipefail
 
-fullVersion=$(type snip_canary_upgrade_init | sed '1,3d;$d' | sed 's/.*istio-\(.*\)\/bin.*/\1/') # 1.18.0
+source "content/en/boilerplates/snips/args.sh"
+
+fullVersion="${bpsnip_args_istio_full_version}"
 fullVersionRevision=$(echo "$fullVersion" | sed -r 's/[.]+/-/g' ) # 1-18-0
-previousVersion=$(type snip_deploy_operator_previous_version | sed '1,3d;$d' | sed 's/.*istio-\(.*\)\/bin.*/\1/')
+previousVersion="${bpsnip_args_istio_previous_version}.0"
 previousVersionMinorUpgrade="${previousVersion%.0}.1"
 
 function testOperatorDeployWatchNs(){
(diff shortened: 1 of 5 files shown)


content/en/docs/setup/install/operator/test.sh
```diff
@@ -22,9 +22,11 @@ set -e
 set -u
 set -o pipefail
 
-fullVersion=$(type snip_canary_upgrade_init | sed '1,3d;$d' | sed 's/.*istio-\(.*\)\/bin.*/\1/') # 1.18.0
+source "content/en/boilerplates/snips/args.sh"
+
+fullVersion="${bpsnip_args_istio_full_version}"
 fullVersionRevision=$(echo "$fullVersion" | sed -r 's/[.]+/-/g' ) # 1-18-0
```
frankbu, line 28: Why not a simpler (more obvious) sed command?
```suggestion
fullVersionRevision=$(echo "$fullVersion" | sed 's/\./-/g' ) # 1-18-0
```

SRodi: addressed in ce5401bc7428932de63cef0afe3adc24ace1f1e7

frankbu (approved):
LGTM. Thanks!

PR #13146, opened 2023-05-01
Use modern prow config default

howardjohn:
This matches istio/istio


ericvn:
I'm not sure there is a way to simply use the istio/istio files this earlier in the job setup.

ericvn:
https://github.com/istio/istio.io/pull/13152

PR #13214, opened 2023-05-18
Remove defaultRevision - should be set in values.yaml

jaymiracola:
Setting `defaultRevision` as described does not function as intended or described (not boolean). It should be set as seen below. 
```
values:
  defaultRevision: 1-16
```

- [ ] Ambient
- [ X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/reference/config/istio.operator.v1alpha1/index.html
+++ b/content/en/docs/reference/config/istio.operator.v1alpha1/index.html
@@ -125,18 +125,6 @@ This option is currently experimental.</p>
 No
 </td>
 </tr>
-<tr id="IstioOperatorSpec-defaultRevision">
-<td><code>defaultRevision</code></td>
-<td><code>bool</code></td>
-<td>
-<p>Identify whether this revision is the default revision for the cluster
-This option is currently experimental.</p>
-
-</td>
-<td>
-No
-</td>
-</tr>
 <tr id="IstioOperatorSpec-meshConfig">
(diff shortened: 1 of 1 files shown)


Arhell (commented):
/ok-to-test

content/en/docs/reference/config/istio.operator.v1alpha1/index.html
```diff
@@ -120,18 +120,6 @@ <h2 id="IstioOperatorSpec">IstioOperatorSpec</h2>
 <p>Identify the revision this installation is associated with.
 This option is currently experimental.</p>
 
-</td>
-<td>
-No
-</td>
-</tr>
-<tr id="IstioOperatorSpec-defaultRevision">
```
hanxiaop, line 128: I think this is generated and refered to the IstioOperator spec in the api repo, which should not edit here. IIRC this is not implemented and only has the api doc.

ericvn (changes requested):
Correct. See the top of the document:
```
---
source_repo: https://github.com/istio/api
title: IstioOperator Options
description: Configuration affecting Istio control plane installation version and shape.
```
This file is copied from the istio/api repo, and needs to be updated there.

ericvn:
Changes would be in this area: https://github.com/istio/api/blob/master/operator/v1alpha1/operator.proto#L94-L100

PR #13217, opened 2023-05-18
Update docs re CNI helm install

funkypenguin:
This PR clarifies documentation on a helm-based CNI install, highlighted in https://github.com/istio/istio/issues/43632, and complements https://github.com/istio/istio/pull/44984, which improves the UX of the istiod chart.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

PR #13265, opened 2023-05-30
Cleanup Istio CNI and Istiod chart installations

rrobb-aspen:
Updating documentation to indicate to users that the webhook annotation injection for CNI network annotations will overwrite existing annotations for other CNI networks.

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/additional-setup/cni/index.md
+++ b/content/en/docs/setup/additional-setup/cni/index.md
@@ -146,9 +146,10 @@ spec:
 
 {{< tab name="Red Hat OpenShift 4.2+" category-value="ocp" >}}
 
-{{< text yaml >}}
+{{< text bash >}}
 $ istioctl install --set profile=openshift
 {{< /text >}}
+
 {{< /tab >}}
 
 {{< /tabset >}}
--- a/content/en/docs/setup/additional-setup/cni/index.md
+++ b/content/en/docs/setup/additional-setup/cni/index.md
@@ -123,12 +123,20 @@ Some platforms required special installation settings.
           enabled: true
           namespace: kube-system
       values:
+        sidecarInjectorWebhook:
+          injectedAnnotations:
+            k8s.v1.cni.cncf.io/networks: istio-cni
         cni:
           cniBinDir: /var/lib/cni/bin
           cniConfDir: /etc/cni/multus/net.d
           cniConfFileName: istio-cni.conf
           chained: false
     {{< /text >}}
+    
+{{< tip >}}
+Note: Using `sidecarInjectorWebhook` to inject the `k8s.v1.cni.cncf.io/networks: istio-cni` annotation will overwrite pre-existing annotations of the same name.
+To prevent this in cases where you have more than one `k8s.v1.cni.cncf.io/networks` annotation value to manage the use of more than one CNI configuration, you will want to modify the value to include the additional CNI annotation values.  If you are installing Istio via Helm, you may leave out the `sidecarInjectorWebhook` configuration as Istio will automatically append `istio-cni` to the list of existing network annotations.
+{{< /tip >}}
 
 ## Operation details
 
--- a/content/en/docs/setup/additional-setup/cni/index.md
+++ b/content/en/docs/setup/additional-setup/cni/index.md
@@ -83,7 +83,8 @@ There are several commonly used install options:
 * `values.cni.cniBinDir` and `values.cni.cniConfDir` configure the directory paths to install the plugin binary and create plugin configuration.
 * `values.cni.cniConfFileName` configures the name of the plugin configuration file.
 * `values.cni.chained` controls whether to configure the plugin as a chained CNI plugin.
-
+  * `values.istio_cni.chained` should be set to the same value as `values.cni.chained` to manage network annotations when chaining CNI plugins.
+  
 {{< tip >}}
 There is a time gap between a node becomes schedulable and the Istio CNI plugin becomes ready on that node.
 If an application pod starts up during this time, it is possible that traffic redirection is not properly set up and traffic would be able to bypass the Istio sidecar.
--- a/content/en/docs/setup/additional-setup/cni/index.md
+++ b/content/en/docs/setup/additional-setup/cni/index.md
@@ -147,25 +147,8 @@ spec:
 {{< tab name="Red Hat OpenShift 4.2+" category-value="ocp" >}}
 
 {{< text yaml >}}
-apiVersion: install.istio.io/v1alpha1
-kind: IstioOperator
-spec:
-  components:
-    cni:
-      enabled: true
-      namespace: kube-system
-  values:
-    cni:
-      cniBinDir: /var/lib/cni/bin
-      cniConfDir: /etc/cni/multus/net.d
-      cniConfFileName: istio-cni.conf
-      chained: false
+$ istioctl install --set profile=openshift
 {{< /text >}}
-
-{{< tip >}}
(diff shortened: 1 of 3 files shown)


ericvn:
Also like some feedback from @istio/wg-networking-maintainers

content/en/docs/setup/additional-setup/cni/index.md
```diff
@@ -132,6 +132,11 @@ Some platforms required special installation settings.
           cniConfFileName: istio-cni.conf
           chained: false
     {{< /text >}}
+    
+{{< tip >}}
+Note: Using `sidecarInjectorWebhook` to inject the `k8s.v1.cni.cncf.io/networks: istio-cni` annotation will overwrite pre-existing annotations of the same name.
```
howardjohn, line 138: Why are we documenting it if its not the right way to do it?

jacob-delgado: We need to remove https://github.com/istio/istio/blob/e2c9358d20e56e03f7ce7c99e36690b17f5db607/manifests/profiles/openshift.yaml#L20 from the openshift profile which has existed for nearly 3 years since https://github.com/istio/istio/pull/28076 was merged

In the current docs for CNI it says you should set this up (all the way back to Istio 1.3). We do need to remove this reference as well in the OpenShift 4.2+ section @rrobb-aspen .

jacob-delgado:
So there is a usability in the helm charts for CNI that is probably using people that are using CNI via Helm to do so incorrectly, unless they are on a platform that fails when it is not setup correctly.

For instance, the `istio_cni` values are not present in the `istiod` chart. Users must set the values accordingly (whereas in the CNI chart they are under `cni` and not `istio_cni`).

Our hope is to clarify this as well for users.

One issue we found is that since `istio_cni` is not being set in the istiod chart and is not told to any user to set, the following annotation is always set `k8s.v1.cni.cncf.io/networks: istio-cni`. This is not correct if you are installed the CNI chart without modifying the `cni.chained` value to `false`.

content/en/docs/setup/additional-setup/cni/index.md
```diff
@@ -91,47 +107,71 @@ This race condition is mitigated by a "detect and repair" method.
...
-          cniConfFileName: istio-cni.conf
-          chained: false
-    {{< /text >}}
+{{< tabset category-name="cni-platform" >}}
+
+{{< tab name="Google Kubernetes Engine" category-value="gke" >}}
+
+{{< text yaml >}}
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
```
jacob-delgado, line 133: maybe we just do 

{{< text bash >}}
$ istioctl install -set profile=openshift
{{< /text >}}

content/en/docs/setup/additional-setup/cni/index.md
```diff
@@ -91,47 +107,71 @@ This race condition is mitigated by a "detect and repair" method.
...
+      enabled: true
+      namespace: kube-system
+  values:
+    cni:
+      cniBinDir: /home/kubernetes/bin
+{{< /text >}}
+
+{{< /tab >}}
+
+{{< tab name="Red Hat OpenShift 4.2+" category-value="ocp" >}}
```
jacob-delgado, line 146: maybe we just do

{{< text bash >}}
$ istioctl install -set profile=openshift
{{< /text >}}

content/en/docs/setup/additional-setup/cni/index.md
```diff
@@ -91,47 +107,67 @@ This race condition is mitigated by a "detect and repair" method.
 Please take a look at [race condition & mitigation](#race-condition--mitigation) section to understand the implication of this mitigation.
 {{< /tip >}}
 
+### Installing with Helm
+
+The Istio CNI and Istio discovery chart use different values that require you set the following either in an overrides values file or at your command prompt when installing the `istiod` chart.
+
+* `values.istio_cni.enabled` should be set to the same value as `values.cni.enabled` to manage network annotations when chaining CNI plugins.
+* `values.istio_cni.chained` should be set to the same value as `values.cni.chained` to manage network annotations when chaining CNI plugins.
```
frankbu, line 115:
```suggestion
The Istio CNI and Istio discovery chart use different values that require you set the following, either in an overrides values file or at your command prompt when installing the `istiod` chart, to manage network annotations when chaining CNI plugins:

* `values.istio_cni.enabled` should be set to the same value as `values.cni.enabled`
* `values.istio_cni.chained` should be set to the same value as `values.cni.chained`
```

rrobb-aspen:
These changes will make the documentation consistent with the changes being made in https://github.com/istio/istio/pull/45207

ericvn:
Note that nothing will pass on the istio.io main branch since we branched 1.18 and there are a few issues testing against the istio main branch instead of the 1.18 branch. I am working to get that fixed.

content/en/docs/setup/additional-setup/cni/index.md
rrobb-aspen: I've updated this tip so that we highlight that using the `sidecarInjectorWebhook` can cause overwrites of the appended cni values for Openshift users who have it set, in case they run into issues but have removed the `sidecarInjectorWebhook` configurations to make it consistent with https://github.com/istio/istio/pull/45207

rrobb-aspen:
@ericvn Now that https://github.com/istio/istio.io/pull/13217 has been merged, is there anything on this PR that we should revise?

content/en/docs/setup/additional-setup/cni/index.md
rrobb-aspen: With Openshift, do we expect users to be more likely to install using a different method?  The configuration values also work with Helm so I'm not sure if we would want to keep them for that use case?

howardjohn: +1 to profile=openshift

content/en/docs/setup/additional-setup/cni/index.md
```diff
@@ -91,47 +107,68 @@ This race condition is mitigated by a "detect and repair" method.
...
+  values:
+    cni:
+      cniBinDir: /var/lib/cni/bin
+      cniConfDir: /etc/cni/multus/net.d
+      cniConfFileName: istio-cni.conf
+      chained: false
+{{< /text >}}
+
+{{< tip >}}
+Note: Using `sidecarInjectorWebhook` to inject the `k8s.v1.cni.cncf.io/networks: istio-cni` annotation will overwrite preexisting annotations of the same name. The sidecar injection templates will automatically append the `istio-cni` network value to the `k8s.v1.cni.cncf.io/networks` annotation on pods if the following values are properly set for OpenShift: `istio_cni.enabled=true` and `istio_cni.chained=false`. If no network annotation currently exists, it will automatically create one for `istio-cni`.
```
howardjohn, line 166: TBH I am not sure this is needed... it seems useful now, but I feel like we should be saying what you _should_ do, not what you should not do.

Given we don't document `k8s.v1.cni.cncf.io/networks`, no users _reading this doc_ will ever set it. Instead, they will see this note and be confused.

Where this could be useful is a troubleshooting doc -- someone with `k8s.v1.cni.cncf.io/networks` has broken configuration. But even then, its completely broken, so not sure how they would ever get into this state?

rrobb-aspen: I can take that out if we don't think it's useful.  My thought was that if someone was using the webhook injection to add the annotation that the configuration could cause issues.

ericvn:
I am good with this. I'll wait a bit to see if there are any other comments.

jacob-delgado (approved):

ericvn (approved):

PR #13268, opened 2023-05-30
Update Hugo, Go, Node, and Ruby versions

ericvn:

Update of some versions on Netlify.


ericvn:
Hugo and Go versions match the build-tools versions. Node matches what is in the Bookinfo. Ruby is newer than the Bookinfo version which needs to be updated.

ericvn:
A quick run through of the netlify produced site, and looking at the build logs, nothing seems amiss.

frankbu (approved):

PR #13276, opened 2023-06-01
cleanup envoy filter use

kyessenov:
Prefer filter insertion over merge, see https://github.com/istio/istio/pull/45223 for test.


ericvn:
I assume this change requires changes that are only in 1.19 (current main branch). We will want to wait an merge until after we branch cut istio.io.

kyessenov:
@ericvn Nope, there is no associated code change. This is a better guidance.

ericvn (approved):

PR #13277, opened 2023-06-02
updates to dual stack blog to make following easier

ilrudie:
Makes a few small changes to the Mar 2023 dual stack blog post for ease of following your test steps.

- adds section about setting up KinD for the test
- add `-y` flag to istioctl command to avoid having the install fail when asking for user input
- remove `/master` from path to the sleep manifest which caused it to be invalid

As a side note, the tcp-echo files we're asking folks to use aren't in release-1.17 but as far as I can tell this should update when 1.18 is released. This is already reflected on preliminary.istio.io so I don't think it needs to be adjusted in the PR.

If there is interest I wouldn't mind adding in httpbin testing steps as well. I have some tests I've been using on my end which pretty closely model yours but use a slightly modified httpbin manifest to deploy. I did not add them in this PR but would be happy to if folks would like.

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [X] User Experience

### DIFF
--- a/content/en/blog/2023/experimental-dual-stack/index.md
+++ b/content/en/blog/2023/experimental-dual-stack/index.md
@@ -43,7 +43,7 @@ announce experimental support for dual stack in Istio 1.17!
 ## A Quick Experiment using Dual Stack
 
 {{< tip >}}
-If you want to use KinD for your test, you can setup a dual stack cluster with the following command:
+If you want to use KinD for your test, you can set up a dual stack cluster with the following command:
 
 {{< text bash >}}
 $ kind create cluster --name istio-ds --config - <<EOF
--- a/content/en/blog/2023/experimental-dual-stack/index.md
+++ b/content/en/blog/2023/experimental-dual-stack/index.md
@@ -42,10 +42,21 @@ announce experimental support for dual stack in Istio 1.17!
 
 ## A Quick Experiment using Dual Stack
 
+**note:** If you want to use KinD for your test you can setup a dual stack cluster with the following:
+
+    {{< text bash >}}
+    kind create cluster --name istio-ds --config - <<EOF
+    kind: Cluster
+    apiVersion: kind.x-k8s.io/v1alpha4
+    networking:
+    ipFamily: dual
+    EOF
+    {{< /text >}}
+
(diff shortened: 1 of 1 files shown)


content/en/blog/2023/experimental-dual-stack/index.md
```diff
@@ -42,10 +42,21 @@ announce experimental support for dual stack in Istio 1.17!
...
+
+    {{< text bash >}}
+    $ kind create cluster --name istio-ds --config - <<EOF
+    kind: Cluster
+    apiVersion: kind.x-k8s.io/v1alpha4
+    networking:
+    ipFamily: dual
+    EOF
+    {{< /text >}}
+
```
frankbu, line 55:
```suggestion
{{< tip >}}
If you want to use KinD for your test, you can setup a dual stack cluster with the following command:

{{< text bash >}}
$ kind create cluster --name istio-ds --config - <<EOF
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  ipFamily: dual
EOF
{{< /text >}}

{{< /tip >}}
```

ilrudie: Thanks!

ericvn:
Correct, this PR will update what is shown in preliminary.istio.io. When 1.18.0 is released (or at some point after that), istio.io will be updated to point at the present day preliminary.istio.io.

ericvn (commented):
LGTM. Will let @frankbu review again.

content/en/blog/2023/experimental-dual-stack/index.md
```diff
@@ -42,10 +42,23 @@ announce experimental support for dual stack in Istio 1.17!
 
 ## A Quick Experiment using Dual Stack
 
+{{< tip >}}
+If you want to use KinD for your test, you can setup a dual stack cluster with the following command:
```
frankbu, line 46: One small nit, sorry I missed it the first time. Otherwise LGTM.
```suggestion
If you want to use KinD for your test, you can set up a dual stack cluster with the following command:
```

frankbu (approved):

PR #13294, opened 2023-06-05
Add proxy-config secret command for verification

kfaseela:
Ingress/Egress TLS examples are enhanced by adding proxy-config secret command for better debugging


frankbu (changes requested):
See in-line comment

frankbu (approved):

PR #13297, opened 2023-06-06
Enhance mTLS origination example

kfaseela:
As described in https://github.com/istio/istio/issues/45259#issuecomment-1577044718 it is better to align the egress mtls origination example to align with the previous section of tls origination by using port 80 itself, to avoid confusions


ericvn (approved):
LGTM

PR #13275, opened 2023-06-01
Document CRL usage

kfaseela:
Document usage for CRL support.

Implementation:

https://github.com/istio/api/pull/2796
https://github.com/istio/istio/pull/45104

PR #13309, opened 2023-06-07
Revert "Document CRL usage"

kfaseela:
Reverts istio/istio.io#13275

Since 1.18 branch cut for docs has not yet happened, and this is a 1.19 only change


kfaseela:
Will revert on 1.18 branch

PR #13367, opened 2023-06-14
keep the same ip:port as above

zhengbingxian:

when i read setup/getting-started as my first guide,  i can see that:
INGRESS_HOST = 127.0.0.1, INGRESS_PORT = 80 
and then we set export GATEWAY_URL=$INGRESS_HOST:$INGRESS_PORT，
and next step we echo "$GATEWAY_URL"， but the result is 192.168.99.100:32194.
There may be a historical legacy, but It's very confusing to see for the  first time.

I think the result look more friendly if it is  "127.0.0.1:80" .

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/getting-started/index.md
+++ b/content/en/docs/setup/getting-started/index.md
@@ -332,7 +332,7 @@ $ export INGRESS_HOST=$(kubectl get po -l istio=ingressgateway -n istio-system -
 
     {{< text bash >}}
     $ echo "$GATEWAY_URL"
-    192.168.99.100:32194
+    127.0.0.1:80
     {{< /text >}}
 
 ### Verify external access {#confirm}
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/getting-started/index.md
```diff
@@ -332,7 +332,7 @@ $ export INGRESS_HOST=$(kubectl get po -l istio=ingressgateway -n istio-system -
 
     {{< text bash >}}
     $ echo "$GATEWAY_URL"
-    192.168.99.100:32194
+    127.0.0.1:80
```
ericvn, line 335: Since we default to the Minikube tab for this page, this would make sense. Not that if you use another tab, the value will likely be different.

ericvn (approved):

zhengbingxian:
@ericvn hi, how to fix the failed job of [gencheck_istio.io]

zirain:
run `make gen` locally

PR #13386, opened 2023-06-15
[zh] sync svg in traffic-management tls-configuration

windsonsea:
Sync  two SVGs in https://istio.io/latest/zh/docs/ops/configuration/traffic-management/tls-configuration/

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


zirain (changes requested):
prefer to not translate these pictures

People work with envoy should be farmiliar with these simply words.

windsonsea:
ok, seems this can be closed due to only image included.

PR #13409, opened 2023-06-19
Updated job names as they cannot use an underscore

ericvn:

Job test names changed to not include underscores. Need to update the directories to match.


kfaseela (approved):

howardjohn (commented):
thanks, sorry I didn't realize it was automated like this

Issue #13315, opened 2023-06-07
Re-enable test disabled after the 1.18 branching and moving back to main


ericvn:
Disabled:
istio.io/content/en/docs/tasks/observability/metrics/telemetry-api/test-disabled.sh

PR #13439, opened 2023-06-22
Re-enable telemetry-api test

ericvn:

Fixes https://github.com/istio/istio.io/issues/13315


ericvn:
Runs locally, just not in the pipeline. Will investigate this some more.

ericvn:
See https://github.com/istio/istio.io/pull/13442 for the same issue now appearing in the 1.18 stream...

ericvn:
The failure seems to be that we either run the commands too soon (before the stats are removed) and every follow on run will find the stats (they just don't go away). Injecting a delay does not seem to be the correct answer. It seems we would want to run once, if not found, restart the pods again , and then try again.

ericvn:
/restest

flaky tests. Passed in earlier runs.

ericvn:
/test doc.test.profile-none

flaky tests. Passed in earlier runs

nmittler (approved):

ericvn:
/test doc.test.profile-none

flaky test. Passed in earlier runs

PR #13504, opened 2023-07-04
[zh sync] delete docs/reference/config/security/constraints-and-properties/index.md

bzsuni:
sync docs/reference/config/security/constraints-and-properties/index.md


windsonsea:
Yes, the English version of the page is no longer available. I am unsure if it might be added back in the future. Alternatively, we could consider adding a postfix `.disable` to indicate its unavailability.

zirain:
I recall all reference docs should point to en version, because it's meaningless.

ericvn:
I believe we now copy all the automated updated reference docs to both the English an Chinese branches the updating so they should in sync (these have the DO NOT MODIFY in the header). The ones that are not automatically updated have nothing to keep them updated (they are like the remainder of the docs).

zirain (approved):

zirain:
> Yes, the English version of the page is no longer available. I am unsure if it might be added back in the future. Alternatively, we could consider adding a postfix `.disable` to indicate its unavailability.

If really need add this back, we can find it from git history.

PR #13516, opened 2023-07-06
update istio-ingress installation name in helm install instructions

### LINKED ISSUE #13497: default helm installation steps doesn't work for k8s ingress
https://istio.io/latest/docs/setup/install/helm/#installation-steps

```text
8. (Optional) Install an ingress gateway:

$ kubectl create namespace istio-ingress
$ helm install istio-ingress istio/gateway -n istio-ingress --wait
```

Following the above official steps to install istio doesn't work for k8s ingress, as the default values of `ingressService` and `ingressSelector` is `istio-ingressgateway` and `ingressgateway` respectly in [MeshConfig](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig).

We should at least change the example release name from `istio-ingress` to `istio-ingressgateway`, as did by [artifacthub example](https://artifacthub.io/packages/helm/istio-official/gateway/1.18.0#installing-the-chart) and [gateway setup example](https://istio.io/latest/docs/setup/additional-setup/gateway/#deploying-a-gateway), so the default helm installation steps works out of the box for k8s ingress.

A lot of people have stumbled upon this issue:

https://github.com/istio/istio/issues/36444#issuecomment-993105159
https://github.com/istio/istio/issues/37329
https://github.com/istio/istio/issues/44402

dhawton:

Mesh Config has a default of looking for `istio-ingressgateway`, which is the name elsewhere and by other methods for installation. This PR changes the installation instructions to match the name used elsewhere. For example: https://istio.io/latest/docs/setup/additional-setup/gateway/#deploying-a-gateway, https://istio.io/latest/docs/setup/install/istioctl/#check-whats-installed.

Fixes #13497 

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/helm/index.md
+++ b/content/en/docs/setup/install/helm/index.md
@@ -127,7 +127,7 @@ You can display the default values of configuration parameters using the `helm s
 
     {{< text syntax=bash snip_id=install_ingressgateway >}}
     $ kubectl create namespace istio-ingress
-    $ helm install istio-ingress istio/gateway -n istio-ingress --wait
+    $ helm install istio-ingressgateway istio/gateway -n istio-ingress --wait
     {{< /text >}}
 
     See [Installing Gateways](/docs/setup/additional-setup/gateway/) for in-depth documentation on gateway installation.
--- a/content/en/docs/setup/install/helm/snips.sh
+++ b/content/en/docs/setup/install/helm/snips.sh
@@ -98,7 +98,7 @@ ENDSNIP
 
 snip_install_ingressgateway() {
 kubectl create namespace istio-ingress
-helm install istio-ingress istio/gateway -n istio-ingress --wait
+helm install istio-ingressgateway istio/gateway -n istio-ingress --wait
 }
 
 snip_helm_ls() {
--- a/content/en/docs/setup/install/helm/test.sh
+++ b/content/en/docs/setup/install/helm/test.sh
@@ -27,7 +27,7 @@ _rewrite_helm_repo snip_install_discovery
 _wait_for_deployment istio-system istiod
 
 _rewrite_helm_repo snip_install_ingressgateway
-_wait_for_deployment istio-ingress istio-ingress
+_wait_for_deployment istio-ingress istio-ingressgateway
 
 # shellcheck disable=SC2154
 _verify_like snip_helm_ls "$snip_helm_ls_out"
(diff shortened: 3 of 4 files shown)


windsonsea (approved):
lgtm for the zh part if the en change is correct.

zirain:
FYI: https://github.com/istio/istio.io/pull/11762#pullrequestreview-1087913597

dhawton:
> FYI: [#11762 (review)](https://github.com/istio/istio.io/pull/11762#pullrequestreview-1087913597)

His comment is directed at the namespace rather than installation name. The installation name is consistent in this doc vs another doc with helm installation (ie, https://istio.io/latest/docs/setup/additional-setup/gateway/#deploying-a-gateway).

zirain:
@howardjohn ptal

howardjohn (changes requested):
This was intentional. Using the same names across samples, IMO, is actively harmful. users should understand what fields link things up and set them to match when appropriate.

We should not rely on a demo world where we _happened_ to name everything the same. It ends with users thinking `istio: ingressgateway` is some magic value, when really its a simple label selector. I have seen this countless times. I have never seen the same with Kubernetes Service -- but probably we would, if their docs always used `hello: world` in _every_ sample and treated it as a bug if it wasn't that.

dhawton:
> This was intentional. Using the same names across samples, IMO, is actively harmful. users should understand what fields link things up and set them to match when appropriate.
> 
> We should not rely on a demo world where we _happened_ to name everything the same. It ends with users thinking `istio: ingressgateway` is some magic value, when really its a simple label selector. I have seen this countless times. I have never seen the same with Kubernetes Service -- but probably we would, if their docs always used `hello: world` in _every_ sample and treated it as a bug if it wasn't that.

However the current documentation varies from the default value we specify in the meshConfig for ingressService, which is `istio-ingressgateway` (https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig). This makes it feel like a bug since a user could follow the documentation as-is, and not have a working Ingress controller without additional steps that are not documented in this page. For a new person coming in to the project, installation steps are a great way to get started and not always best practice. As far as the linked comment by @zirain, I agree, the changing of namespaces is a good move and we document that in the Deploying Gateway docs.

I guess an alternative could be to update the documentation to clarify that the name in this step differs from meshConfig's default value and additional steps are required. I'm trying to approach this as someone who knows Kubernetes and Helm, but doesn't know Istio. I would follow the guide to get a learning environment setup and play in, and nothing seems to indicate that this wouldn't work "out of the box" until they go to test the environment and see the ingress not working.

PR #13591, opened 2023-07-18
Bump gateway api version used in examples

keithmattix:
Bump gateway-api version in the examples to 0.7.1

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [X] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/getting-started/index.md
+++ b/content/en/docs/ops/ambient/getting-started/index.md
@@ -53,7 +53,7 @@ Follow these steps to get started with ambient:
 
     {{< text bash >}}
     $ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.1" | kubectl apply -f -; }
+      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.7.1" | kubectl apply -f -; }
     {{< /text >}}
 
     {{< tip >}}
@@ -471,5 +471,5 @@ $ kubectl delete -f @samples/sleep/notsleep.yaml@
 If you installed the Gateway API CRDs, remove them:
 
 {{< text bash >}}
-$ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.1" | kubectl delete -f -
+$ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.7.1" | kubectl delete -f -
 {{< /text >}}
--- a/content/en/docs/ops/ambient/getting-started/index.md
+++ b/content/en/docs/ops/ambient/getting-started/index.md
@@ -53,7 +53,7 @@ Follow these steps to get started with ambient:
 
     {{< text bash >}}
     $ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.7.1" | kubectl apply -f -; }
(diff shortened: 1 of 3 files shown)


dhawton (dismissed):

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -53,7 +53,7 @@ Follow these steps to get started with ambient:
 
     {{< text bash >}}
     $ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.1" | kubectl apply -f -; }
+      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.7.1" | kubectl apply -f -; }
```
frankbu, line 56: Why are these docs using a hard coded version, instead of the latest version like all the other Istio docs?
```suggestion
      { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref={{< k8s_gateway_api_version >}}" | kubectl apply -f -; }
```

frankbu (changes requested):
See inline comment

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -471,5 +471,5 @@ $ kubectl delete -f @samples/sleep/notsleep.yaml@
 If you installed the Gateway API CRDs, remove them:
 
 {{< text bash >}}
-$ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.6.1" | kubectl delete -f -
+$ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.7.1" | kubectl delete -f -
```
frankbu, line 474:
```suggestion
$ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref={{< k8s_gateway_api_version >}}" | kubectl delete -f -
```

content/en/docs/ops/ambient/getting-started/index.md
keithmattix: My guess is because we want to reference a particular semver release rather than the commit SHA that currently exists. I can create another arg (e.g. k8s_gateway_api_tag) if that would be better?

ericvn:
I did read the target to the `make gen` in https://github.com/istio/istio.io/pull/13593. Will see how the tests progress.

ericvn:
We also need to get the gateway-api update in istio/istio for which I created a PR as well. Once that merges, we can update istio.io to point to it, along with my other change and we will see 0.7.1 again.

ericvn:
Part of the equation is if we need some later commits from gateway-api for our istio repo. I see that John has a PR to pull in a later version which contains Frank's https://github.com/kubernetes-sigs/gateway-api/pull/2183.

ericvn:
In my istio PR it is noted that 0.7.1 is a downgrade of the gateway-api version that what is being used. Since preliminary.istio.io follows the main istio branch, we may use a commit off of HEAD, with the idea that we would be back to a normal gateway-api release for our releases. 

It seems that as such, we really want to stay on the current version, and likely will move up in the near future as there is an update pending in an istio PR. 

One thing that isn't automated with this is we won't automatically pick the commit up in our tests/docs since the make target was disabled (hard to get actual string needed when it's not just a release). I can look into doing that again.

frankbu:
@ericvn @keithmattix So IIUC, the correct fix for the docs in this PR, is to change them to use `ref={{< k8s_gateway_api_version >}}`, as I suggested, and the correct version will show up in the docs at some point in the near future when the related build automation is fixed up.

keithmattix:
Sounds good. Thanks @ericvn @frankbu

PR #13593, opened 2023-07-18
Renable the update-gateway-version

ericvn:

Renable the tag after it was removed when the main istio branch moved to a non GatewayApi release.

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -85,7 +85,7 @@ site:
 snips:
 	@scripts/gen_snips.sh
 
-gen: tidy-go format-go snips
+gen: tidy-go format-go update-gateway-version snips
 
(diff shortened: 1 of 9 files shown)


ericvn:
```
error: failed to run '/usr/bin/git fetch --depth=1 https://github.com/kubernetes-sigs/gateway-api v0.7.1-0.20230517171234-6b9b7346ca5b': fatal: couldn't find remote ref v0.7.1-0.20230517171234-6b9b7346ca5b
```

ericvn:
Since we are running on later commits on the HEAD of gateway-api, we can't update this at this point. We can and should at branch cut so that our releases use gateway-api releases.

Might be beneficial to update our logic when a sha is being used to follow that as part of the `make gen` to prevent us from missing an update there.

PR #13595, opened 2023-07-18
Re-enable processing of gateway-api version in `make gen`

ericvn:

Currently, we have removed the automation that updates the `k8s_gateway_api_version` in the data/args.yml.

This re-enables that code with the required addition that we use the version if specified in the go.mod or use the SHA if it is specified.

We need to get the full SHA which will require the GH_TOKEN and likely an update to the prow job.


ericvn:
`make gen` its failing with a 
```
gh: To use GitHub CLI in automation, set the GH_TOKEN environment variable.
```
in the newly created code. Creating a test-infra PR to add this to this job.

ericvn:
https://github.com/istio/test-infra/pull/4912 created

ericvn:
Updated to no longer use the full sha, but pull the base release, 0.7.1 in this case.

howardjohn (approved):

frankbu (approved):

PR #13624, opened 2023-07-25
[release-1.18] Try running tests with actual 1.18.2 images

ericvn:

Verify tests run with the 1.18.2 images. Update sha to the istio patched release branch and use the actual images (since the istio-testing ones were never created).

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -41,9 +41,9 @@ export ISTIO_LONG_SHA
 # we only need to export the pipeline HUB value.
 # If the images were built as part of the private pipeline (as for security releases),
 # we export the HUB and TAG for the images once they are published.
-HUB ?= gcr.io/istio-testing
-# export HUB := docker.io/istio
-# export TAG ?= 1.7.3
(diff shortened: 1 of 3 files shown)


ericvn:
/test doc.test.profile-none

Maybe profile-none tests don't follow the HUB/TAG and are failing to find items.

ericvn:
Looking at the tests, the helm related tests seem to want to use the sha related images which don't exist so those failures make sense.

The virtual machine tests have:
```
dpkg-deb: error: 'istio-sidecar.deb' is not a Debian format archive
dpkg: error processing archive istio-sidecar.deb (--install):
 dpkg-deb --control subprocess returned error exit status 2
Errors were encountered while processing:
 istio-sidecar.deb
```
which seem suspicious. Need to see how to verify that file.

I'm not sure about the distributed tracing. Seems the sample isn't starting?

ericvn:
More investigation and the deb file is being pulled from the wrong location. Looks like all tests are failing do to issues related to trying to use an actual build image and not using the dev/pere-release images.

Issue #13630, opened 2023-07-25
k3d cluster create command fails


totallyGreg:
In https://istio.io/latest/docs/setup/platform-setup/k3d/#installation the command to create the k3d cluster fails. 
```shell
❯ k3d cluster create --api-port 6550 -p "9080:80loadbalancer"  -p "9443:443loadbalancer" --agents 2 --k3s-arg '--disable=traefik@server:*'                                                                                                          17:14:31

INFO[0000] portmapping '9080:80loadbalancer' lacks a nodefilter, but there's more than one node: defaulting to [servers:*:proxy agents:*:proxy]
FATA[0000] failed to transform ports: error parsing port spec '9080:80loadbalancer': Invalid containerPort: 80loadbalancer
```

The nodefilters are missing the "@" before the loadbalancer group

Adding them allows k3d to create the cluster successfully. 
`k3d cluster create --api-port 6550 -p "9080:80@loadbalancer" -p "9443:443@loadbalancer" --agents 2 --k3s-arg '--disable=traefik@server:*' ` 

Editing the doc shows that the @ signs are present in port commands, but are wrapped in double quotes inside a text block which seems to render the command without the @ signs.  Changing these to single quotes should render the command properly but I have no way to test.

totallyGreg:
Digging deeper it seems with the original code:

```
$ k3d cluster create --api-port 6550 -p "9080:80@loadbalancer"  -p "9443:443@loadbalancer" --agents 2 --k3s-arg '--disable=traefik@server:*'
    {{< /text >}}
```

renders correctly using `hugo serve` from localhost (minus the images).

```
$ k3d cluster create --api-port 6550 -p "9080:80@loadbalancer"  -p "9443:443@loadbalancer" --agents 2 --k3s-arg '--disable=traefik@server:*'
```

Which is different that what I am seeing on https://istio.io/latest/docs/setup/platform-setup/k3d/#installation
Oddly, the page shows it correctly in html

```
<code class=language-bash data-expandlinks=true data-repo=istio>$ k3d cluster create --api-port 6550 -p &#34;9080:80@loadbalancer&#34;  -p &#34;9443:443@loadbalancer&#34; --agents 2 --k3s-arg &#39;--disable=traefik@server:*&#39;
                                </code>
```

This is incorrectly interpreting the text between the first two @ signs as a link

PR #13631, opened 2023-07-25
Wrap port options in single quotes so @ sign renders properly

totallyGreg:

Fix double quotes preventing @loadbalancer from rendering properly in k3d cluster create command by wrapping port commands in single quotes.  As described in https://github.com/istio/istio.io/issues/13630

- [ ] Ambient
- [X ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/platform-setup/k3d/index.md
+++ b/content/en/docs/setup/platform-setup/k3d/index.md
@@ -23,7 +23,7 @@ k3d makes it very easy to create single- and multi-node k3s clusters in docker, 
 1.  Create a cluster and disable `Traefik` with the following command:
 
     {{< text bash >}}
-    $ k3d cluster create --api-port 6550 -p "9080:80@loadbalancer"  -p "9443:443@loadbalancer" --agents 2 --k3s-arg '--disable=traefik@server:*'
+    $ k3d cluster create --api-port 6550 -p '9080:80@loadbalancer'  -p '9443:443@loadbalancer' --agents 2 --k3s-arg '--disable=traefik@server:*'
     {{< /text >}}
 
 1.  To see the list of k3d clusters, use the following command:


craigbox (approved):

content/en/docs/setup/platform-setup/k3d/index.md
```diff
@@ -23,7 +23,7 @@ k3d makes it very easy to create single- and multi-node k3s clusters in docker,
 1.  Create a cluster and disable `Traefik` with the following command:
 
     {{< text bash >}}
-    $ k3d cluster create --api-port 6550 -p "9080:80@loadbalancer"  -p "9443:443@loadbalancer" --agents 2 --k3s-arg '--disable=traefik@server:*'
+    $ k3d cluster create --api-port 6550 -p '9080:80@loadbalancer'  -p '9443:443@loadbalancer' --agents 2 --k3s-arg '--disable=traefik@server:*'
```
dhawton, line 26: Looking at the preliminary site, this is not resolved by using single quotes. I believe it's the way @'s are used as special links through the text shortcode.

dhawton (changes requested):

content/en/docs/setup/platform-setup/k3d/index.md
frankbu: Yes, I think the fix for this is to make the regex for processing `@` chars more specific here:
https://github.com/istio/istio.io/blob/2d14e2e08dc7dedefd619be80c6526a33d153405/src/ts/codeBlocks.ts#L189

The `/@(.*?)@/g` part should maybe make sure that the string between the `@` chars is not a quote.

Maybe something like this will work? `/@([^"']*?)@/g`

dhawton: Ya.. I'm discussing with Greg but I have the fix. There's two places that need changing.

dhawton: https://github.com/istio/istio.io/pull/13637 has the fix

dhawton:
Closing as this is included in https://github.com/istio/istio.io/pull/13637. Discussed with Greg via Slack.

PR #13640, opened 2023-07-26
add missed / to @@ link handling regex

dhawton:

Missed a / in the regex for @@ link handling. Didn't see this locally. @frankbu this should resolve the issue you highlighted.

Will do a manual cherry pick to combine the two

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/src/ts/codeBlocks.ts
+++ b/src/ts/codeBlocks.ts
@@ -186,7 +186,7 @@ function handleCodeBlocks() {
 
             if (cmd !== "") {
                 if (code.dataset.expandlinks === "true") {
-                    cmd = cmd.replace(/@([\w\.\-_]*?)@/g, "<a href='https://raw.githubusercontent.com/istio/" + code.dataset.repo + "/" + branchName + "/$1'>$1</a>");
+                    cmd = cmd.replace(/@([\w\/\.\-_]*?)@/g, "<a href='https://raw.githubusercontent.com/istio/" + code.dataset.repo + "/" + branchName + "/$1'>$1</a>");
                 }
 
                 let html = "<div class='command'>" + cmd + "</div>";
--- a/src/ts/codeBlocks.ts
+++ b/src/ts/codeBlocks.ts
@@ -186,7 +186,7 @@ function handleCodeBlocks() {
 
             if (cmd !== "") {
                 if (code.dataset.expandlinks === "true") {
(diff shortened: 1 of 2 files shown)


src/ts/codeBlocks.ts
```diff
@@ -186,7 +186,7 @@ function handleCodeBlocks() {
 
             if (cmd !== "") {
                 if (code.dataset.expandlinks === "true") {
-                    cmd = cmd.replace(/@([\w\.\-_]*?)@/g, "<a href='https://raw.githubusercontent.com/istio/" + code.dataset.repo + "/" + branchName + "/$1'>$1</a>");
+                    cmd = cmd.replace(/@([\w\/\.\-_]*?)@/g, "<a href='https://raw.githubusercontent.com/istio/" + code.dataset.repo + "/" + branchName + "/$1'>$1</a>");
```
frankbu, line 189: I'm not sure this covers all the characters that might appear in a URL path. For example `#` is possible. Also, `?` and `:`, among other things, but I'm not sure they would ever be used in URLs we want to be clickable links.

dhawton: https://deploy-preview-13640--preliminary-istio.netlify.app/latest/docs/tasks/traffic-management/ingress/gateway-api/#configuring-a-gateway looks fine now as does the original issue on the k3d platform setup page that highlighted the regex overreach

frankbu: You're probably right that we don't have any links to docs using other special characters. It would be nice if we could be sure, rather than rely on someone noticing a bad command in some doc. We could add lint checking for md files to make sure they only use the restricted URL syntax between the @ delimiters, but that's probably more work than it's really worth.

frankbu: Btw, I think \w includes _

```suggestion
                    cmd = cmd.replace(/@([\w\/\.\-]*?)@/g, "<a href='https://raw.githubusercontent.com/istio/" + code.dataset.repo + "/" + branchName + "/$1'>$1</a>");
```

dhawton: I did a search using `@[^\w/\.\-]+@` through all md and html files and there were not any matches that I didn't expect

frankbu (approved):

PR #13643, opened 2023-07-26
update links for cves

jacob-delgado:
Update links for published CVEs in Envoy


ericvn (approved):
Links behave as they have for earlier news items.

PR #13655, opened 2023-07-28
46223 Add missing step to apply destination rules after running virtual services

bedla:
Hi,
This is PR for https://github.com/istio/istio/issues/46223 ... see detials there.
Thx
Ivos


content/en/docs/tasks/traffic-management/request-routing/index.md
```diff
@@ -59,6 +59,12 @@ Run the following command to apply virtual services that will route all traffic
 $ kubectl apply -f @samples/bookinfo/networking/virtual-service-all-v1.yaml@
 {{< /text >}}
 
+And to apply `DestinationRule` items run following command.
```
frankbu, line 62: This isn't needed because the warning on line 46 tells the user to make sure it's already done.
```
If you haven't already, follow the instructions in [define the service versions](/docs/examples/bookinfo/#define-the-service-versions).
```

bedla: ahh, I see. It is kind of hidden.

I executed command from section https://istio.io/latest/docs/setup/getting-started/#bookinfo
image

And did not clicked to link (because it looks like that it is not necessary) that goes to https://istio.io/latest/docs/examples/bookinfo/ ... let me update the text, to make it more explicit.

bedla: I have force pushed changed, what do you think?

content/en/docs/setup/getting-started/index.md
```diff
@@ -110,7 +110,7 @@ Follow these steps to get started with Istio:
 
 ## Deploy the sample application {#bookinfo}
 
-1.  Deploy the [`Bookinfo` sample application](/docs/examples/bookinfo/):
+1.  Follow steps in the page [`Bookinfo` sample application](/docs/examples/bookinfo/) to deploy sample application:
```
frankbu, line 113: I don't think we want to change the getting started page to tell users to go to another document to do things. The getting started doc is supposed to be a simple, self contained, set of instructions to make it easy for users to start looking at Istio.

I see why the current request-routing instructions are a little confusing (hidden). My suggestion would be to remove the warning on line 46 and put the warning under each tab (`Istio APIs` and `Gateway API`) instead.

Then you can change the wording under the `Istio APIs` tab to make it clear that it's asking you to create the destination rules:
```
If you haven’t already applied destination rules, follow the instructions in [define the service versions](/docs/examples/bookinfo/#define-the-service-versions).
```

Arhell (commented):
/ok-to-test

craigbox (commented):
Please update the PR with Frank's suggestions.
(I definitely think the problem is on the `request-routing` page, not the getting started page.)

PR #13700, opened 2023-08-07
Add supported info for istio 1.10-1.12

jacob-delgado:
See https://github.com/istio/istio/pull/46213 for more information. We should document these w/r/t possible revision changes.

These releases have been EOL for over a year.


ericvn:
I'm not sure why we should add more to this table, and we should probably shrink the table some more (I think it currently shows 4 unsupported Istio releases).

PR #13712, opened 2023-08-10
Fix typo "AuthorizaionPolicy"

ramonssarmento:
As said just typo correction

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/reference/config/security/authorization-policy/index.html
+++ b/content/en/docs/reference/config/security/authorization-policy/index.html
@@ -143,7 +143,7 @@ in the <code>foo</code> namespace.</p>
 <p>{{<tabset category-name="example">}}
 {{<tab name="v1beta1" category-value="v1beta1">}}</p>
 <pre><code class="language-yaml">apiVersion: security.istio.io/v1beta1
-kind: AuthorizaionPolicy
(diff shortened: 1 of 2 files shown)


content/en/docs/reference/config/security/authorization-policy/index.html
```diff
@@ -143,7 +143,7 @@
 <p>{{<tabset category-name="example">}}
 {{<tab name="v1beta1" category-value="v1beta1">}}</p>
 <pre><code class="language-yaml">apiVersion: security.istio.io/v1beta1
-kind: AuthorizaionPolicy
+kind: AuthorizationPolicy
```
hanxiaop, line 146: I think this is not the right place to fix the typo, these are just copied yamls from `istio/api` and should be fixed there.

hanxiaop: See https://github.com/istio/api/blob/master/security/v1/authorization_policy.proto#L177

ramonssarmento: Nice point, ty

PR #12184, opened 2022-11-04
Istio Steering Committee Election Results

kfaseela:
Signed-off-by: Faseela K <faseela.k@est.tech>

Istio Steering Committee Election Results

PR #13736, opened 2023-08-16
2023 Steering Election Results

rvennam:
Blog to announce the 2023 Steering Election Results

### DIFF
--- a/content/en/blog/2023/steering-election-results/index.md
+++ b/content/en/blog/2023/steering-election-results/index.md
@@ -1,0 +1,25 @@
+---
+title: "Istio Announces Winners of 2023 Steering Committee Election"
+subtitle: Announcing the newest Istio Steering Committee members.
+description: Announcing the newest Istio Steering Committee members.
+publishdate: 2023-08-16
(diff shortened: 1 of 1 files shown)


dhawton (commented):
LGTM. One question, last year there was a thank you and list of departing Steering Committee members (https://github.com/istio/istio.io/pull/12184/files#diff-64053b8c0f8cace80f9b8826152d38aa5eadd12d37e7aeddd4e9f694ca455a78R17). Should that be included here as well?

linsun:
Thank you @rvennam and @cetezadi for leading the elections!  Congrats @craigbox @kfaseela @irisdingbj !

craigbox:
> LGTM. One question, last year there was a thank you and list of departing Steering Committee members

There were three people who are continuing and one who stood down in favour of a colleague, who won the same seat.  So maybe, but be mindful of how it is worded,

Though, if we're going to do that I think we should also thank the other candidates: any of them would have been a valuable member, and all of whom will remain leaders in the project

cetezadi (approved):

dhawton (approved):

PR #13747, opened 2023-08-18
Correct Istio Upgrade version step message

CathalC2011:

- [ ] Ambient
- [X] Docs
- [X] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [X] User Experience

The existing message implies that upgrading two minor versions is supported, and indeed this has historically been the case, at least anecdotally.

The Istio 1.17 control plane however, is not compatible with Istio 1.15 proxies, breaking that soft rule. That should be reflected here, to prevent users from incorrectly jumping multiple minor versions.

### DIFF
--- a/content/en/docs/setup/upgrade/_index.md
+++ b/content/en/docs/setup/upgrade/_index.md
@@ -7,5 +7,5 @@ test: n/a
 ---
 
 {{< warning >}}
-Upgrading across more than two minor versions (e.g., `1.6.x` to `1.9.x`) in one step is not officially tested or recommended.
+Upgrading across more than one minor version (e.g., `1.15.x` to `1.17.x`) in one step is not officially tested or recommended.
 {{< /warning >}}


ericvn:
Would you point to where this is discussed. Is this maybe related to this https://github.com/istio/istio/issues/43785? I believe there was an issue upgrading to 1.17.0 and 1.17.1 but it was fixed in 1.17.2.

content/en/docs/setup/upgrade/_index.md
```diff
@@ -7,5 +7,5 @@ test: n/a
 ---
 
 {{< warning >}}
-Upgrading across more than two minor versions (e.g., `1.6.x` to `1.9.x`) in one step is not officially tested or recommended.
+Upgrading across more than one minor version (e.g., `1.15.x` to `1.17.x`) in one step is not officially tested or recommended.
```
dhawton, line 10:
```suggestion
Upgrading across more than two minor version (e.g., `1.15.x` to `1.17.x`) in one step is not officially tested or recommended. Upgrades that skip a minor version should only be done when using [Canary Upgrades](https://istio.io/latest/docs/setup/upgrade/canary/).
```

Per John's guidance

dhawton (changes requested):

content/en/docs/setup/upgrade/_index.md
linsun: +1 should be supported for canary

ericvn: We do test some upgrade testing across version although I need to verify what it's doing:  https://github.com/istio/istio/tree/master/tests/integration/helm/upgrade

linsun: Should we say - Upgrading across more than two minor version (e.g., `1.15.x` to `1.17.x`) in one step is not officially tested or recommended for in place upgrade.

dhawton: Personally would like to specifically call out Canary Upgrades and link to doc to help guide toward a set procedure that is better practice.

Arhell (commented):
/ok-to-test

craigbox (commented):
Saying "Upgrading across" implies skipping a version, and the original line read "Upgrading across two minor versions (e.g. 1.6 to 1.9)".

I think we need to clarify the statement a little further.

PR #13753, opened 2023-08-21
Fix typo AuthorizaionPolicy to AuthorizationPolicy

zioproto:
Trivial, fix typo. Annoying when doing cut and paste of examples ...


kfaseela:
This was already fixed in the latest docs. Do you still see this? https://istio.io/latest/docs/reference/config/security/authorization-policy/

zioproto:
Yes I still see this at the following URL:
https://istio.io/latest/docs/reference/config/security/authorization-policy/
Screenshot 2023-08-21 at 12 35 31

kfaseela:
> Yes I still see this at the following URL: https://istio.io/latest/docs/reference/config/security/authorization-policy/ Screenshot 2023-08-21 at 12 35 31

Please fix it here then : https://github.com/istio/api/blob/master/security/v1/authorization_policy.pb.go#L179

Arhell (commented):
/ok-to-test

ericvn (changes requested):
We typically do not change archived versions as they can be overridden. To do this properly, you would need to update the istio/api registry in the 1.17 branch, and then we would need to do some manual work to get that copied to the 1.17 istio branch and finally to the archive.

zioproto:
I think we are good with the fix in 1.18, thanks

Issue #13574, opened 2023-07-14
F5 logo is too big


craigbox:
The largely horizontal nature of the images on our company logos (including front pages) makes the circular F5 logo stand out a bit too much.

<img width="1072" alt="Screenshot 2023-07-15 at 11 26 15 AM" src="https://github.com/istio/istio.io/assets/132510/f4880eda-3f81-4cd5-8c6f-cb8feb94b058">

Please use some CSS magic to make them visually equivalent.

Whatever we do also has to work well on https://istio.io/latest/about/ecosystem/.

wilsonwu:
Seems not only F5 has a large logo:
![image](https://github.com/istio/istio.io/assets/1269496/169bbb4e-6197-4517-917b-bda20f6cb027)

ericvn:
I suspect we limit the width so that the image doesn't pass horizontal or vertical bounds. A more square image will naturally scale larger in the square box. We likely need a more proportional images scaling. Not sure how to do the in our docs.

mike-weiner:
@ericvn's suspicion that we are limiting the width to prevent the image from spilling out of the gray tile behind it is correct. He is also correct that there probably needs to be more proportional logo image files.

The short answer is that I don't think there is any great solution.

Each gray tile is a `company_logo` partial layout. These tiles have a `.company-logo` class that have padding to keep the image "inside" the gray background and display the logos in this grid fashion.

I tried setting a `max-width` property of `4rem` on the actual `<img>` tag, but this resulted in the exact opposite problem. (See the screenshot below.) The larger logos looked reasonably sized, but some of the smaller logs were way to small.

<img width="1367" alt="Screenshot 2023-07-27 at 4 06 06 PM" src="https://github.com/istio/istio.io/assets/47571709/17055ba6-50bb-4a73-832e-4c20463f0fa8">

My second thought was to create a second CSS class that we could add to logos to add some extra padding to reduce the large logo sizes. However, this won't work either since the `companies` partial layout has no information on the underlying size of the logo image files. We wouldn't want to do this anyways since it would have to be manually maintained.

I don't see any way (at least currently) where we could look at the dimensions of the image and then based on the ratio of image width to image height set the padding to try and keep the logos all approximately the same size.

craigbox:
OK, thanks Mike - I think the action here is to upload an F5 logo with some padding built into the image, and then perhaps open an issue to revise the other logos later.  Some are much larger than others, but they don't immediately look out of place to me.

[...]

PR #13765, opened 2023-08-23
update intel company logo

Arhell:

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience

made the logo a little smaller


ericvn:
Is there some formula/process being used to figure out this resizing? I'm more concerned that someone would comeback and say why is our logo now seem smaller than others. Granted, a more square logo will be scaled down less than one that is wider in our square boxes. As noted in another issue, there are some formulas that try to take in account the height and with of pictures and scaling in square boxes to prevent some of this larger logo appearance. We don't seem to be able to apply the formulas ourselves in the code, but if we had a formula that said we would scale down the logo by `x` based on it's height/width ratio, then we have something to say that we are consistently doing this and not potentially favoring some logos.

Arhell:
> Is there some formula/process being used to figure out this resizing? I'm more concerned that someone would comeback and say why is our logo now seem smaller than others. Granted, a more square logo will be scaled down less than one that is wider in our square boxes. As noted in another issue, there are some formulas that try to take in account the height and with of pictures and scaling in square boxes to prevent some of this larger logo appearance. We don't seem to be able to apply the formulas ourselves in the code, but if we had a formula that said we would scale down the logo by `x` based on it's height/width ratio, then we have something to say that we are consistently doing this and not potentially favoring some logos.

Doing it because of this https://github.com/istio/istio.io/issues/13574

I want to do everything a little at a time in about 1 size, but the big ones definitely need to be reduced

ericvn:
Comment added in the issue to an article which talks about a fixed ratio. I'm not sure if that is easily applied to these images during manual resizing or not.

Arhell:
you can do everything through css but the pictures (content) should be suitable, but unfortunately not in this case.

ericvn (approved):

PR #13778, opened 2023-08-28
Fixed numbering for installation steps in the tip

archups:

- [ ] Ambient
- [ x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/operator/index.md
+++ b/content/en/docs/setup/install/operator/index.md
@@ -77,7 +77,7 @@ You can alternatively deploy the operator using Helm:
     $ kubectl create namespace istio-operator
     {{< /text >}}
 
-2. Install operator using Helm.
+2) Install operator using Helm.
 
     {{< text syntax=bash snip_id=deploy_istio_operator_helm >}}
     $ helm install istio-operator manifests/charts/istio-operator \
--- a/content/en/docs/setup/install/operator/index.md
+++ b/content/en/docs/setup/install/operator/index.md
@@ -77,7 +77,7 @@ You can alternatively deploy the operator using Helm:
     $ kubectl create namespace istio-operator
     {{< /text >}}
 
-1. Install operator using Helm.
+2. Install operator using Helm.
 
     {{< text syntax=bash snip_id=deploy_istio_operator_helm >}}
     $ helm install istio-operator manifests/charts/istio-operator \
--- a/content/zh/docs/setup/install/operator/index.md
+++ b/content/zh/docs/setup/install/operator/index.md
@@ -75,7 +75,7 @@ $ istioctl operator init --watchedNamespaces=istio-namespace1,istio-namespace2
     $ kubectl create namespace istio-operator
     {{< /text >}}
 
-1. 使用 Helm 安装 Operator。
+2. 使用 Helm 安装 Operator。
 
     {{< text syntax=bash snip_id=deploy_istio_operator_helm >}}
     $ helm install istio-operator manifests/charts/istio-operator \
--- a/content/zh/docs/setup/install/operator/index.md
+++ b/content/zh/docs/setup/install/operator/index.md
@@ -75,7 +75,7 @@ $ istioctl operator init --watchedNamespaces=istio-namespace1,istio-namespace2
     $ kubectl create namespace istio-operator
     {{< /text >}}
 
-2. 使用 Helm 安装 Operator。
+2) 使用 Helm 安装 Operator。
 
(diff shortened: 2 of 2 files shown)


ericvn (dismissed):
The original numbers are correct. All items should be labeled 1., and the will be automatically numbered,

Arhell (commented):
/ok-to-test

archups:
> The original numbers are correct. All items should be labeled 1., and the will be automatically numbered,

@ericvn Yes, it works usually that way, but not within tip. Please check the rendered number on the doc site itself https://istio.io/latest/docs/setup/install/operator/.
Hence PR raised for fix. Thanks for review!

ericvn:
If numbering doesn't work, I'd maybe recommend just using `-` instead of `#.` and just have bullets. @frankbu thoughts?

frankbu:
Using `1)` and `2)` for the list items, instead of `1.` and `2.`, might work.

ericvn:
> Using `1)` and `2)` for the list items, instead of `1.` and `2.`, might work.

I did try that locally, and it didn't seem to work for me. Also tried some indentation changes.

ericvn:
> > I did try that locally, and it didn't seem to work for me.
> 
> What did it display? It should work.

Using `1.` It showed both as `1.` which is the problem to solve. Changing to `1)` left them both showing as `1.` Changing to `-` showed each as a bullet. All this with no other indentation. Looking at the public index.html with the `)`s:
```
<div>
    <aside class="callout tip">
        <div class="type">
            <svg class="large-icon"><use xlink:href="/img/icons.svg#callout-tip"/></svg>
        </div>
        <div class="content"><p>You can alternatively deploy the operator using Helm:</p>
<ol>
<li>
<p>Create a namespace <code>istio-operator</code>.</p>
 <pre><code class='language-bash' data-expandlinks='true' data-repo='istio' >$ kubectl create namespace istio-operator
</li>
</ol>
<p></code></pre></p>
<ol>
<li>
<p>Install operator using Helm.</p>
 <pre><code class='language-bash' data-expandlinks='true' data-repo='istio' >$ helm install istio-operator manifests/charts/istio-operator \
 --set watchedNamespaces=&#34;istio-namespace1\,istio-namespace2&#34; \
 -n istio-operator
</li>
</ol>
<p></code></pre></p>
<p>Note that you need to <a href="/docs/setup/getting-started/#download">download the Istio release</a>
to run the above command.</p>
</div>
    </aside>
</div>
```

ericvn:
Looking through other parts of the doc, I do see #) used, most notably it seems when we need to restart numbering at some number. And it seems to be around `tabset`s. I really didn't see anything that related to numbering within a tip, and only see 2. used within a section header.

Looking at this particular page, I don't really see numbering used at all, except in this tip, so removing the numbering makes sense to me. Keeping a bullet (using `-`) does seem like an acceptable change if we wanted to call out the steps separately.

frankbu:
> Looking through other parts of the doc, I do see #) used, most notably it seems when we need to restart numbering at some number.

Markdown lets you continue a list that otherwise is disconnected from the previous items by specifying something other than 1 for the entry, but the lint rule we've enabled for the docs (MD029) requires all list entries to be 1, so it won't allow `2.`. The work around is that the lint rule only enforces the rule for the `.` syntax, but markdown allows either `.` or `)` for ordered lists. So using `2)` won't cause a lint error, and will continue the list at 2. Here's what I get when I tried:
image

content/en/docs/setup/install/operator/index.md
```diff
@@ -77,7 +77,7 @@ You can alternatively deploy the operator using Helm:
     $ kubectl create namespace istio-operator
     {{< /text >}}
 
-1. Install operator using Helm.
+2. Install operator using Helm.
```
frankbu, line 80:
```suggestion
2) Install operator using Helm.
```

content/zh/docs/setup/install/operator/index.md
```diff
@@ -75,7 +75,7 @@ $ istioctl operator init --watchedNamespaces=istio-namespace1,istio-namespace2
     $ kubectl create namespace istio-operator
     {{< /text >}}
 
-1. 使用 Helm 安装 Operator。
+2. 使用 Helm 安装 Operator。
```
frankbu, line 78:
```suggestion
2) 使用 Helm 安装 Operator。
```

frankbu (approved):

PR #13780, opened 2023-08-29
Add warning about enable alpha gateway API CRDs

keithmattix:
Since https://github.com/istio/istio/pull/44539, a special env var is required to enable the use of alpha Gateway API CRDs like TCPRoute. This PR adds a warning informing users about this before the TCP traffic shifting demo

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
+++ b/content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
@@ -23,6 +23,8 @@ weighted routing feature.
 
 {{< boilerplate gateway-api-experimental >}}
 
+{{< boilerplate gateway-api-enable-alpha-crds >}}
+
 ## Before you begin
 
 * Setup Istio by following the instructions in the [Installation guide](/docs/setup/).
--- a/content/en/boilerplates/gateway-api-enable-alpha-crds.md
+++ b/content/en/boilerplates/gateway-api-enable-alpha-crds.md
@@ -1,0 +1,7 @@
+---
+---
+{{< warning >}}
+This document uses alpha resources of the Kubernetes Gateway API. You must configure istiod to read these resources by setting the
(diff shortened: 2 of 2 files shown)


content/en/docs/tasks/traffic-management/tcp-traffic-shifting/index.md
```diff
@@ -23,6 +23,8 @@ weighted routing feature.
 
 {{< boilerplate gateway-api-experimental >}}
 
+{{< boilerplate gateway-api-enable-alpha-crds >}}
```
frankbu, line 26: Instead of this warning, should we add install tabs in the before you begin section, which includes the install instruction with the env var under the Gateway API tab?

Also, why is this doc's test not failing, given that we are running it in a cluster with the demo profile?

keithmattix: I'm actually not sure why the tests pass...running the demo itself didn't work for me

frankbu: Hmm, probably because the istio integration tests set it to true and the doc test framework is using the same base settings? https://github.com/istio/istio/blob/65ee971a8b57889c8e2126a51ade3f9668dc2bb8/tests/integration/base.yaml#L22

I think it needs to be default (false) in the istio.io tests or we're not really testing that docs work as written.

keithmattix: Oh wow I didn't realize the docs tests use the integration base settings 🤦🏾 That's actually kinda unexpected

keithmattix:
Closing in favor of 13782

PR #13782, opened 2023-08-30
gateway-api: doc tests should not run with alpha CRDs by default

frankbu:

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/boilerplates/gateway-api-experimental.md
+++ b/content/en/boilerplates/gateway-api-experimental.md
@@ -2,10 +2,19 @@
 ---
 {{< warning >}}
 This document uses [experimental features](https://gateway-api.sigs.k8s.io/concepts/versioning/#release-channels-eg-experimental-standard)
-of the Kubernetes Gateway API. Make sure to install the experimental CRDs before using the Gateway API:
+of the Kubernetes Gateway API which require the alpha version of the CRDs. Before proceeding with this task, make sure to:
 
-{{< text syntax=bash snip_id=install_experimental_crds >}}
-$ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref={{< k8s_gateway_api_version >}}" | kubectl apply -f -
-{{< /text >}}
+1) Install the alpha version of the Gateway API CRDs:
+
+    {{< text syntax=bash snip_id=install_experimental_crds >}}
+    $ kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref={{< k8s_gateway_api_version >}}" | kubectl apply -f -
+    {{< /text >}}
+
+2) Configure Istio to read the alpha resources by setting the `PILOT_ENABLE_ALPHA_GATEWAY_API` environment variable to `true`
+    when installing Istio:
+
+    {{< text syntax=bash snip_id=enable_alpha_crds >}}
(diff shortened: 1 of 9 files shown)


keithmattix:
*sigh* naturally when we try and fix the bad behavior, we can't merge it because of the bug. Should I just include this change in #13780?

content/en/boilerplates/gateway-api-experimental.md
```diff
@@ -2,10 +2,19 @@
 ---
 {{< warning >}}
 This document uses [experimental features](https://gateway-api.sigs.k8s.io/concepts/versioning/#release-channels-eg-experimental-standard)
-of the Kubernetes Gateway API. Make sure to install the experimental CRDs before using the Gateway API:
+of the Kubernetes Gateway API which require the alpha version of the CRDs. Before proceeding with this task, make sure to:
```
keithmattix, line 5: nit: `experimental` and alpha CRDs are two different concepts in Gateway API

frankbu: Yeah, I wondered how best to word this. I worded it to say it uses "experimental features which require the alpha CRDs", trying to imply that the experimental features used in this task are ones that require the alpha CRDs, but maybe it's confusing?

I wonder if we really need to mention that it uses experimental features, just say that it needs the alpha CRDs? Or maybe, like you had in your PR, we should have 2 separate warnings but the part about installing CRDs needs to be removed from the "experimental features" warning, and added in the second warning?

Thoughts?

keithmattix: Yeah from a Gateway API perspective there are two separate things the user needs to be aware of:
1. Experimental features like TCP support shouldn't be relied on in prod. This applies to all mesh-aware Gateway API examples 
2. Alpha gateway API _resources_ (like TCPRoute) are ignored by default so to do this particular tutorial, they need to set an env var. 

1. should have a warning everywhere Gateway API for mesh is used or where the doc is using TCP, UDP, GRPC, etc. I think the only doc that requires a warning for 2. is this one

frankbu: I just noticed that the doc tests are also always installing the experimental CRDs: https://github.com/istio/istio.io/blob/fb364e8ef148ed014c71be71053326189966e2e4/tests/util/gateway-api.sh#L19

Now I'm wondering if that's not what we want either. Isn't it true that we're supposed to use the beta CRDs for the experimental mesh features (HTTPRoute -> Service) now. So now we should really be using the standard CRDs for docs, unless they use TCP, UDP, etc? Is that right?

keithmattix: Experimental/standard != alpha/beta. Standard CRDs only include `Gateway`, `GatewayClass`, and `HTTPRoute`. Other resources like `TCPRoute`are still experimental. Both standard and experimental channels contain the beta CRDs for HTTPRoute, Gateway, and GatewayClass.

frankbu: I think the confusion is because there are two "Experimental" things in Gateway API:

1. Experimental Channel
2. Experimental Features

IIUC, Experimenal Features that involve new fields on beta resources (e.g., on HTTPRoute) will be available in both the Standard and Experimental Channels. If that's correct, I think the diagram [here](https://gateway-api.sigs.k8s.io/concepts/versioning/#release-channels-eg-experimental-standard) is wrong because it says that only Standard Fields are in the Standard Channel?

PR #13791, opened 2023-08-31
Bump gateway api to 0.8.0

jwendell:


ericvn (changes requested):
This will happen when the istio test reference gets updated automatically via the make gen. We keep the same version (actually just the tag) as the version that is in the istio commit's go.mod that is referenced.

jwendell:
> This will happen when the istio test reference gets updated automatically via the make gen. We keep the same version (actually just the tag) as the version that is in the istio commit's go.mod that is referenced. For the 1.19.0 release that is tentatively the rc.1 as noted here unless the istio build moves up. The PR is currently on hold until 1.19.1.

Ah ok, good to know this is automated. Thanks.

PR #13796, opened 2023-08-31
Fix Inconsistently Sized Company Logos

mike-weiner:
Related to: https://github.com/istio/istio.io/issues/13574#issuecomment-1693374423

This PR works to fix some of the inconsistent sizes of company logos. Changes include:

- `src/sass/base/_base.scss` and `src/sass/misc/_logo-gallery.scss` were cleaned up to remove some of the overly prescriptive CSS that was preventing us from somewhat normalizing the company logo sizes.
- `static/logos/google-cloud.png`, `static/logos/mulesoft.svg`, `static/logos/tencent-cloud.svg`, and `static/logos/tetrate.svg` were modified to remove the excess space around the logo.

**Moving forward we want any new logo added to the site to have zero extra spacing. The logos that were changed in this PR are good examples. The logo files don't need to have specific dimensions but we don't want the logo files to be just big enough to fit the logo and not any larger.**

@ericvn - I spent quite a bit of time working on seeing whether I could get the Javascript described in the article (https://danpaquette.net/read/automatically-resizing-a-list-of-icons-or-logos-so-theyre-visually-proportional/) that you linked [here](https://github.com/istio/istio.io/issues/13574#issuecomment-1693374423) to function for this use case. The short answer is that it probably won't work here. The longer answer is that, yes, I did get this partially working. I ran into issues since Javascript cannot modify image width and height until the image has been completely loaded by the browser. This introduced timing issues with certain images being cached and other images not being cached. I tried waiting for the entire window to be loaded, but then page execution is held up until every image is loaded...and that was way to slow. 

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn:
Comparing https://deploy-preview-13796--preliminary-istio.netlify.app/latest/about/ecosystem and the same page on istio.io I think this looks good. Some of the other large logos are reduced in line with what I would expect. The only weird one I noticed was F5 got a little bigger, but if it's in-line with the others now based on this change, I think that's good.

PR #13855, opened 2023-09-12
Document special case for gateway install on OpenShift

jwendell:
Documents the changes here: https://github.com/istio/istio/pull/46898


ericvn:
The additional instruction causes the snips to have their values increased requiring a small change in ../setup/additional-setup/gateway/test.sh.

jwendell:
Thanks @ericvn

ericvn (approved):

PR #13871, opened 2023-09-14
Add ambient helm installation guide

hanxiaop:

- [ ] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,102 @@
...
+$ helm install istio-base istio/base
+{{< /text >}}
+
+### Installing CNI Component
+
+The **CNI** chart installs the Istio CNI Plugin. There are some main roles of Istio CNI Plugin:
+
+1. Eliminates the need for the `istio-init` container that sets up traffic routing for sidecar proxies.
+
+1. In Ambient, it is responsible for detecting the pods that belong to the ambient mesh, and configuring
```
ericvn, line 40: Remove trailing space to fix lint
```suggestion
1. In Ambient, it is responsible for detecting the pods that belong to the ambient mesh, and configuring
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,102 @@
...
+
+{{< text bash >}}
+$ helm install istio-base istio/base
+{{< /text >}}
+
+### Installing CNI Component
+
+The **CNI** chart installs the Istio CNI Plugin. There are some main roles of Istio CNI Plugin:
+
+1. Eliminates the need for the `istio-init` container that sets up traffic routing for sidecar proxies.
```
jacob-delgado, line 38: Perhaps we should mention (with a link) that this is a safe, more secure security practice and link to the Istio Best Security practices page

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,102 @@
...
+configures the proxies to route traffic within the mesh.
+
+{{< text bash >}}
+$ kubectl create namespace istio-system
+$ helm install istiod istio/istiod --namespace istio-system \
+    --set defaultRevision="" \
+    --set meshConfig.defaultConfig.proxyMetadata.ISTIO_META_ENABLE_HBONE=true \
+    --set 'meshConfig.defaultProviders.metrics[0]=prometheus' \
+    --set 'meshConfig.extensionProviders[0].name=prometheus' \
+    --set 'meshConfig.extensionProviders[0].prometheus={}' \
```
jacob-delgado, line 63: We should probably remove these settings since not all of our users will have prometheus installed.

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,102 @@
...
+    --set 'pilot.env.VERIFY_CERTIFICATE_AT_CLIENT=true' \
+    --set 'pilot.env.ENABLE_AUTO_SNI=true' \
+    --set 'pilot.env.PILOT_ENABLE_HBONE=true' \
+    --set 'pilot.env.PILOT_ENABLE_AMBIENT_CONTROLLERS=true' \
+    --set 'pilot.env.CA_TRUSTED_NODE_ACCOUNTS=istio-system/ztunnel\,kube-system/ztunnel' \
```
jacob-delgado, line 68: Does it make sense for us to ship istiod with an `ambient-values.yaml` file for both cni and for istiod?

jacob-delgado: See https://github.com/istio/istio/pull/46994

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: Thanks, I was trying to figure out where's the trailing space :)

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: Sounds great

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,102 @@
...
+
+1. In Ambient, it is responsible for detecting the pods that belong to the ambient mesh, and configuring 
+  the traffic redirection between the ztunnels - which will be installed later.
+
+{{< text bash >}}
+$ helm install istio-cni istio/cni -n kube-system \
+    --set cni.ambient.enabled=true \
+    --set cni.logLevel=info \
+    --set cni.privileged=true \
+    --set 'cni.excludeNamespaces={kube-system}'
```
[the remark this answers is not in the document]
hanxiaop: CNI default value excludes `istio-system` as well, which should work in ambient.

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: Yes, it does. It looks like it's easier to maintain. Thanks for the value YAML files!

content/en/docs/ops/ambient/helm-installation/index.md
howardjohn: This is a bit confusing here since its for ambient where there is no sidecar. I mean technically its correct, since you can still use sidecars... but the motivation is not this; the motivation is that its a hard requirement for ambient.

hanxiaop: Thanks, I've removed the sidecar part to not confuse users.

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+{{< /text >}}
+
+### Installing CNI Component
+
+The **CNI** chart installs the Istio CNI Plugin. It is responsible for detecting the pods that belong to the ambient mesh,
+and configuring the traffic redirection between the ztunnels - which will be installed later.
+
+{{< text bash >}}
+$ helm install istio-cni istio/cni -n kube-system \
+  -f manifests/charts/istio-cni/ambient-values.yaml
```
jacob-delgado, line 48: This will only be for those that have downloaded the charts. If you are using the Helm repo you'll need to give the full URL to the ambient values file, such as `-f https://raw.githubusercontent.com/istio/istio/master/manifests/charts/istio-cni/ambient-values.yaml`

bleggett: Can also use the Istio.io code import: https://preliminary.istio.io/latest/docs/releases/contribute/code-blocks/#import-code-from-an-external-source-through-a-url and render (that version of the file) inline in the doc for copypasting, like so, which might be a little more obvious:


```
helm install istio-cni istio/cni -n kube-system -f - << EOF
cni:
  # Configuration log level of istio-cni binary
  # by default istio-cni send all logs to UDS server
  # if want to see them you need change global.logging.level with cni:debug
  logLevel: info

  # Allow the istio-cni container to run in privileged mode, needed for some platforms (e.g. OpenShift)
  privileged: true

  # Configure ambient settings
  ambient:
    # If enabled, ambient redirection will be enabled
    enabled: true

  # Default excludes istio-system; its actually fine to redirect there since we opt-out istiod, ztunnel, and istio-cni
  excludeNamespaces:
  - kube-system
EOF
```

bleggett: TBQH if base case requires a large values file for default/demo install, that's a good signal we have too many values - for instance in the above example it's not clear that `privileged` couldn't simply be implicitly set to true if `ambient.enabled` is true, and dropped from the values file.

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+In the following steps, we will use the downloaded Istio release package as an example, and make sure you are in the
+root directory of the Istio release package.
+
+### Installing Base Component
+
+The **Base** chart contains the basic CRDs and cluster roles required to set up Istio.
+This should be installed prior to any other Istio component.
+
+{{< text bash >}}
```
bleggett, line 37: Unrelated aside: I still think eventually we should/ought to get this down to 2 charts: `istio-dataplane` and `istio-controlplane` - installed separately, upgraded separately.

content/en/docs/ops/ambient/helm-installation/index.md
bleggett, line 48:
```suggestion
  -f manifests/charts/istio-cni/ambient-values.yaml
```

This specific `values.yaml` only has about 3 values overrides from the base case/default.

In terms of docs, my rule of thumb is 

1. If your `custom-values.yaml` override overrides 3 or fewer fields, docs should use individual `--set` flags and not a `-f file.yaml` flag.

2. If your `custom-values.yaml` override overrides 4 or more fields, docs should use a `-f file.yaml` flag and not individual `--set` flags. Also, if this is a common/base case, you might need to consider if you really need users to explicitly override that many things for the base case, versus implicitly setting some of the values based on other values, as a followup.

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Installing Istiod Component
+
+The **Istiod** chart installs a revision of Istiod. Istiod is the control plane component that manages and
+configures the proxies to route traffic within the mesh.
+
+{{< text bash >}}
+$ kubectl create namespace istio-system
+$ helm install istiod istio/istiod --namespace istio-system \
+  -f manifests/charts/istio-control/istio-discovery/ambient-values.yaml
```
bleggett, line 59: Yeah for `istiod` the number of overrides is larger so we are stuck with a `-f` flag file ref.

For now that's fine, maybe later we can streamline that or collapse multiple values behind a flag (since `ambient` is a base case, going forward).

srampal:
@hanxiaop   fyi (see PR below) I am adding a Helm User's guide and that currently includes an installation section with a placeholder for helm. We should co-ordinate and either consolidate to use your content or restructure the mapping of content to user guides.
https://github.com/istio/istio.io/pull/13635

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: > Can also use the Istio.io code import: https://preliminary.istio.io/latest/docs/releases/contribute/code-blocks/#import-code-from-an-external-source-through-a-url and render (that version of the file) inline in the doc for copypasting, like so, which might be a little more obvious

I'm going to try this to see if it works, thanks for the info!

hanxiaop: > TBQH if base case requires a large values file for default/demo install, that's a good signal we have too many values with unreasonable defaults - for instance in the above example it's not clear that privileged couldn't simply be implicitly set to true if ambient.enabled is true, and dropped from the ambient values file.

I think this makes the components easier to install, but it may potentially lack flexibility because users may want to change a subset of values under the 'grouped' `ambient.enabled`, rather than the entire set. Additionally, having separate values can help indicate the differences between the original mode and the ambient mode, which is useful for users who are migrating from the sidecar mode to ambient. Given this, IMO if Ambient is defined as a separate default configuration, it might be better to simplify the Helm install, and otherwise it may be better to keep the current settings.

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: So, in this case, where should `istio-cni` be placed? I thought it was required in ambient to configure data plane, but it seems more like a network helper in the default sidecar mode, which neither belongs to the data plane nor the control plane?

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: Not seeing this when I replied the early comment, so are we assuming `ambient` is a new base case compared to the default sidecar mode? If so I'd think we could optimize the values to provide better UX.

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: Looks like we cannot have embeded texts...
image
Maybe just revise to https://preliminary.istio.io/latest/docs/releases/contribute/code-blocks/#link-2-files?

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,83 @@
...
+### Installing Base Component
+
+The **Base** chart contains the basic CRDs and cluster roles required to set up Istio.
+This should be installed prior to any other Istio component.
+
+{{< text bash >}}
+$ helm install istio-base istio/base
+{{< /text >}}
+
+### Installing CNI Component
```
therealmitchconnors, line 34: Is CNI necessary for Ambient?  It isn't installed by istioctl install as far as I can tell....

bleggett: It should be installed with the `ambient` profile and yea it is.

content/en/docs/ops/ambient/helm-installation/index.md
bleggett: Right now `istio-cni` and `ztunnel` are effectively both part of the dataplane.

`istio-cni` isn't part of the dataplane at runtime, but it must be installed with (and upgraded with) the dataplane, in most cases, at the current time.

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Installing Istiod Component
+
+The **Istiod** chart installs a revision of Istiod. Istiod is the control plane component that manages and
+configures the proxies to route traffic within the mesh.
+
+{{< text bash >}}
+$ kubectl create namespace istio-system
+$ helm install istiod istio/istiod --namespace istio-system \
+  -f @manifests/charts/istio-control/istio-discovery/ambient-values.yaml@
```
bleggett, line 52:
```suggestion
$ helm install istiod istio/istiod --namespace istio-system --create-namespace \
  -f @manifests/charts/istio-control/istio-discovery/ambient-values.yaml@
```

bleggett (approved):
Minor nit/suggestion, otherwise LGTM as a documentation of current state, though I still think we can make this much simpler by automating the individual installs with a basic demo chart that rolls up the individual charts as deps.

Everyone can always install the individual charts separately one-by-one but that doesn't mean most people need or want to do that (e.g. it's unclear to me why most people wouldn't install ztunnel and CNI together as an `istio-dataplane` parent chart, 99% of the time, even in production environments, given current dataplane upgrade recommendations)

content/en/docs/ops/ambient/helm-installation/index.md
bleggett: > > TBQH if base case requires a large values file for default/demo install, that's a good signal we have too many values with unreasonable defaults - for instance in the above example it's not clear that privileged couldn't simply be implicitly set to true if ambient.enabled is true, and dropped from the ambient values file.
> 
> I think this makes the components easier to install, but it may potentially lack flexibility because users may want to change a subset of values under the 'grouped' `ambient.enabled`, rather than the entire set. Additionally, having separate values can help indicate the differences between the original mode and the ambient mode, which is useful for users who are migrating from the sidecar mode to ambient. Given this, IMO if Ambient is defined as a separate default configuration, it might be better to simplify the Helm install, and otherwise it may be better to keep the current settings.

If they have a legitimate need to do that we can add that later. It's much easier to expose something as an option after the fact when someone makes a case for it then it is to un-expose when it's clear no one really needs to individually set it.

We should expose the minimal set of things as `values` that are required to install Istio across the platforms and environments we support - everything beyond that, people can make a case for in a PR.

In the case of `privileged` here for the CNI there's no real case for us to let that be individually toggleable outside of the `ambient` flag - it has to be true for the `ambient` feature to work, and even if we change that as an implementation detail, there's no reason not to make it implicit, since users changing it to anything else if they set the other flag would be an explicit misconfiguration.

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,83 @@
...
+description: How to install Ambient Mesh with Helm.
+weight: 4
+owner: istio/wg-environments-maintainers
+test: n/a
+---
+
+This guide shows you how to install Ambient Mesh with Helm.
+Besides the demo in [Getting Started with Ambient Mesh](/docs/ops/ambient/getting-started/),
+we **encourage** you to follow this guide to install Ambient Mesh.
+Helm helps you manage components separately, and you can easily upgrade the components to the latest version.
```
bleggett, line 12: I feel like it should be obvious by this point, but we should maybe add a note that we only support Helm 3 and newer.

hanxiaop: Thanks, I've added the prerequisite step copied from the normal helm installation [https://istio.io/latest/docs/setup/install/helm/.](https://istio.io/latest/docs/setup/install/helm/#prerequisites)

hanxiaop:
@ericvn PTAL, thanks

hanxiaop:
Kindly ping @istio/wg-docs-maintainers-english

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+1. Check the [Requirements for Pods and Services](/docs/ops/deployment/requirements/).
+
+1. [Install the Helm client](https://helm.sh/docs/intro/install/), version 3.6 or above.
+
+1. Configure the Helm repository:
+
+{{< text bash >}}
+$ helm repo add istio https://istio-release.storage.googleapis.com/charts
+$ helm repo update
+{{< /text >}}
```
dhawton, line 27:
```suggestion
    {{< text bash >}}
    $ helm repo add istio https://istio-release.storage.googleapis.com/charts
    $ helm repo update
    {{< /text >}}
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+*See [helm repo](https://helm.sh/docs/helm/helm_repo/) for command documentation.*
+
+## Installing the Components
+
+### Installing Base Component
```
dhawton, line 33:
```suggestion
### Installing the Base Component
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+The **Base** chart contains the basic CRDs and cluster roles required to set up Istio.
```
dhawton, line 35:
```suggestion
The `base` chart contains the basic CRDs and cluster roles required to set up Istio.
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+This should be installed prior to any other Istio component.
+
+{{< text bash >}}
+$ helm install istio-base istio/base
+{{< /text >}}
+
+### Installing CNI Component
+
+The **CNI** chart installs the Istio CNI Plugin. It is responsible for detecting the pods that belong to the ambient mesh,
```
dhawton, line 44:
```suggestion
The `cni` chart installs the Istio CNI Plugin. It is responsible for detecting the pods that belong in the ambient mesh,
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Installing Ambient Mesh with Helm
+description: How to install Ambient Mesh with Helm.
+weight: 4
+owner: istio/wg-environments-maintainers
+test: n/a
+---
+
+This guide shows you how to install Ambient Mesh with Helm.
```
dhawton, line 9:
```suggestion
This guide shows you how to install ambient mesh with Helm.
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+Besides the demo in [Getting Started with Ambient Mesh](/docs/ops/ambient/getting-started/),
+we **encourage** you to follow this guide to install Ambient Mesh.
```
dhawton, line 11:
```suggestion
we **encourage** you to follow this guide to install ambient mesh.
```

content/en/docs/ops/ambient/helm-installation/index.md
dhawton: I feel like we should also mention its sidecar purpose here... since it isn't only for ambient. This, to me, eludes to the CNI only being used for ambient. Maybe link to the CNI page?

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+This should be installed prior to any other Istio component.
+
+{{< text bash >}}
+$ helm install istio-base istio/base
+{{< /text >}}
+
+### Installing CNI Component
+
+The **CNI** chart installs the Istio CNI Plugin. It is responsible for detecting the pods that belong to the ambient mesh,
+and configuring the traffic redirection between the ztunnels - which will be installed later.
```
dhawton, line 45:
```suggestion
and configuring the traffic redirection between the ztunnel daemonset, which will be installed later.
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+{{< text bash >}}
+$ helm install istio-cni istio/cni -n kube-system \
+  -f @manifests/charts/istio-cni/ambient-values.yaml@
+{{< /text >}}
+
+### Installing Istiod Component
+
+The **Istiod** chart installs a revision of Istiod. Istiod is the control plane component that manages and
```
dhawton, line 54:
```suggestion
The `istiod` chart installs a revision of Istiod. Istiod is the control plane component that manages and
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+configures the proxies to route traffic within the mesh.
+
+{{< text bash >}}
+$ helm install istiod istio/istiod --namespace istio-system --create-namespace \
+  -f @manifests/charts/istio-control/istio-discovery/ambient-values.yaml@
+{{< /text >}}
+
+### Installing Ztunnel Component
+
+The **Ztunnel** chart installs a ztunnel, which is the node-proxy component in Ambient.
```
dhawton, line 64:
```suggestion
The `ztunnel` chart installs the ztunnel daemonset, which is the node-proxy component of ambient.
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+The **Istiod** chart installs a revision of Istiod. Istiod is the control plane component that manages and
+configures the proxies to route traffic within the mesh.
+
+{{< text bash >}}
+$ helm install istiod istio/istiod --namespace istio-system --create-namespace \
+  -f @manifests/charts/istio-control/istio-discovery/ambient-values.yaml@
+{{< /text >}}
+
+### Installing Ztunnel Component
```
dhawton, line 62:
```suggestion
### Installing the ztunnel component
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+The **CNI** chart installs the Istio CNI Plugin. It is responsible for detecting the pods that belong to the ambient mesh,
+and configuring the traffic redirection between the ztunnels - which will be installed later.
+
+{{< text bash >}}
+$ helm install istio-cni istio/cni -n kube-system \
+  -f @manifests/charts/istio-cni/ambient-values.yaml@
+{{< /text >}}
+
+### Installing Istiod Component
```
dhawton, line 52:
```suggestion
### Installing the Istiod Component
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,90 @@
...
+### Installing Base Component
+
+The **Base** chart contains the basic CRDs and cluster roles required to set up Istio.
+This should be installed prior to any other Istio component.
+
+{{< text bash >}}
+$ helm install istio-base istio/base
+{{< /text >}}
+
+### Installing CNI Component
```
dhawton, line 42:
```suggestion
### Installing the CNI Component
```

content/en/docs/ops/ambient/helm-installation/index.md
dhawton: Indenting keeps it "under" the ordered list item instead of breaking out of the list, and allows for further additions without needing to worry about numbering in the future.

dhawton:
Sorry for the delay, believe both Eric and I are on vacation.. I'm just sitting in the lounge at the airport so gave it a review.

content/en/docs/ops/ambient/helm-installation/index.md
dhawton: It is mandatory.. the CNI component handles all the redirection logic for both the node and pods. Without it, the pods won't be put into the ambient mesh.

content/en/docs/ops/ambient/helm-installation/index.md
linsun: +1 on helm 3.x plus

content/en/docs/ops/ambient/helm-installation/index.md
linsun: yep worth noting this is the same cni from before but with expanded function to configure traffic redirection from app to ztunnel

linsun:
Hi @hanxiaop thx for putting this together, looks good in general. Do you plan to add quick step (or link) to install ingress gw?  And perhaps worth adding a link to the get started guide as next step?  I think the goal is after running through steps here, users can jump to the get started guide somewhere (without needing to install istio), agree?

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: Thanks for clarifying this, makes sense.

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: I added the sidecar mode explaination before, however I see the motivation of ambient here https://github.com/istio/istio.io/pull/13871#discussion_r1327736697, which makes sense to me.

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: Right now we have the requirement of helm version to be 3.6+ in the helm installation guide https://istio.io/latest/docs/setup/install/helm/#prerequisites, shouldn't we use the same value for consistency?

hanxiaop:
> And perhaps worth adding a link to the get started guide as next step? I think the goal is after running through steps here, users can jump to the get started guide somewhere (without needing to install istio), agree?

Sounds good, I'm going to add some more content about the demo part. 

> Do you plan to add quick step (or link) to install ingress gw?

Maybe mention and link to the gateway api page, which is currently preferred?

ericvn:
This guide really needs a test.

hanxiaop:
> This guide really needs a test.

I'll try to add some.

content/en/docs/ops/ambient/helm-installation/index.md
dhawton: The suggested change here still is outstanding.

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+    {{< text syntax=bash snip_id=delete_system_namespace >}}
+    $ kubectl delete namespace istio-system
+    {{< /text >}}
+
+## Uninstall stable revision label resources
+
+If you decide to continue using the old control plane, instead of completing the update,
+you can uninstall the newer revision and its tag by first issuing
+`helm template istiod istio/istiod -s templates/revision-tags.yaml --set revisionTags={prod-canary} --set revision=canary -n istio-system | kubectl delete -f -`.
```
dhawton, line 162:
```suggestion
you can uninstall the newer revision and its tag by first running:

{{< text syntax=bash >}}
$ helm template istiod istio/istiod -s templates/revision-tags.yaml --set revisionTags={prod-canary} --set revision=canary -n istio-system | kubectl delete -f -
{{< /text >}}

```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,175 @@
...
+You must them uninstall the revision of Istio that it pointed to by following the uninstall procedure above.
```
dhawton, line 163:
```suggestion
You must then uninstall the revision of Istio by following the uninstall procedure above.
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+If you installed the gateway(s) for this revision using in-place upgrades, you must also reinstall the gateway(s) for the previous revision manually,
```
dhawton, line 165:
```suggestion
If you installed the gateway(s) for this revision using in-place upgrades, you must also reinstall the gateway(s) for the previous revision manually.
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,175 @@
...
+Removing the previous revision and its tags will not automatically revert the previously in-place upgraded gateway(s).
```
dhawton, line 166:
```suggestion
Removing the previous revision and its tags will not automatically revert the previously upgraded gateway(s).
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+### (Optional) Deleting CRDs installed by Istio
+
+Deleting CRDs permanently removes any Istio resources you have created in your
+cluster. To permanently delete Istio CRDs installed in your cluster:
```
dhawton, line 171:
```suggestion
Deleting CRDs removes any Istio resources you have created in your cluster. To delete Istio CRDs installed in your cluster:
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Verifying the deployments status
+
+After installing all the components, you can check the helm deployment status:
+
+{{< text syntax=bash snip_id=show_components >}}
+$ helm list -n istio-system
+{{< /text >}}
+
+You can check the status of the pods deployed:
```
dhawton, line 88:
```suggestion
You can check the status of the deployed pods with:
```

content/en/docs/ops/ambient/helm-installation/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+{{< text syntax=bash >}}
+$ helm show values istio/istiod
+{{< /text >}}
+
+## Verifying the Installation
+
+### Verifying the deployments status
+
+After installing all the components, you can check the helm deployment status:
```
dhawton, line 82:
```suggestion
After installing all the components, you can check the Helm deployment status with:
```

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: Thanks for all comments, revised.

hanxiaop:
@ericvn I've looked into the test:
```
++_rewrite_helm_repo(): helm install --set global.tag=1.20-alpha.486c6ca3836df5c27e9b1f6b0dc4a79c87c4f71d istio-cni istio/cni -n istio-system -f manifests/charts/istio-cni/ambient-values.yaml
...
+_wait_for_daemonset(): kubectl --context= -n istio-system rollout status daemonset istio-cni-node --timeout 5m
+_wait_for_daemonset(): echo 'Failed rollout of daemonset istio-cni-node in namespace istio-system'
+_wait_for_daemonset(): exit 1
```
It looks like we don't have `1.20.0-alpha.0` in our docker registry. However, we do have beta.0 and rc.0, so not sure why though. I think whenever the tag is updated, the test I wrote can pass.

content/en/docs/ops/ambient/helm-installation/index.md
hanxiaop: This cannot be revised, otherwise we will violate some lint rules. This is actually copied from https://istio.io/latest/docs/setup/install/helm/#uninstall-stable-revision-label-resources.

ericvn:
> It looks like we don't have `1.20.0-alpha.0` in our docker registry. However, we do have beta.0 and rc.0, so not sure why though. I think whenever the tag is updated, the test I wrote can pass.

You need to set the hub to gcr.io/istio-testing to find the images. The istio.io testing typically references normal istio commits, and not releases. Those images are built in a post-submit to Istio-testing.

srampal:
We are adding multiple user guides into a dedicated folder as can be seen from the current master branch and the link below. Please move this doc into that folder also.

https://istio.io/latest/docs/ops/ambient/usage/

hanxiaop:
> > It looks like we don't have `1.20.0-alpha.0` in our docker registry. However, we do have beta.0 and rc.0, so not sure why though. I think whenever the tag is updated, the test I wrote can pass.
> 
> You need to set the hub to gcr.io/istio-testing to find the images. The istio.io testing typically references normal istio commits, and not releases. Those images are built in a post-submit to Istio-testing.

Thanks, I'll have a try.

hanxiaop:
> We are adding multiple user guides into a dedicated folder as can be seen from the current master branch and the link below. It would be more consistent to move this doc also into that folder (into the Installation Guide folder within the Usage folder).
> 
> https://istio.io/latest/docs/ops/ambient/usage/

Sounds good, I'll rebase and see what is the current structure.

hanxiaop:
> > > It looks like we don't have `1.20.0-alpha.0` in our docker registry. However, we do have beta.0 and rc.0, so not sure why though. I think whenever the tag is updated, the test I wrote can pass.
> > 
> > 
> > You need to set the hub to gcr.io/istio-testing to find the images. The istio.io testing typically references normal istio commits, and not releases. Those images are built in a post-submit to Istio-testing.
> 
> Thanks, I'll have a try.

I think it's using the correct repository. However, I'm not quite sure why the test fails because `istio-cni-node` cannot be successfully rolled out in the test. I've looked into other tests where istio-cni is deployed, and they seem fine. Are there any clues where I can gather more information, besides the one at https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/13871/doc.test.profile-none_istio.io/1726852176258535424/artifacts/tests-setup-profile-none-f7a832/TestDocs/ops/ambient/usage/helm-installation/test.sh/test.sh/_test_context/test.sh_debug.txt?

hanxiaop:
> > > > It looks like we don't have `1.20.0-alpha.0` in our docker registry. However, we do have beta.0 and rc.0, so not sure why though. I think whenever the tag is updated, the test I wrote can pass.
> > > 
> > > 
> > > You need to set the hub to gcr.io/istio-testing to find the images. The istio.io testing typically references normal istio commits, and not releases. Those images are built in a post-submit to Istio-testing.
> > 
> > 
> > Thanks, I'll have a try.
> 
> I think it's using the correct repository. However, I'm not quite sure why the test fails because `istio-cni-node` cannot be successfully rolled out in the test. I've looked into other tests where istio-cni is deployed, and they seem fine. Are there any clues where I can gather more information, besides the one at https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/13871/doc.test.profile-none_istio.io/1726852176258535424/artifacts/tests-setup-profile-none-f7a832/TestDocs/ops/ambient/usage/helm-installation/test.sh/test.sh/_test_context/test.sh_debug.txt?

Looks like `_rewrite_helm_repo` is missing components like istio-cni and ztunnel... Hopefully this fix will pass the test.

hanxiaop:
test flake /test doc.test.profile-default

hanxiaop:
@ericvn Could you PTAL? Tests are added. Thanks!

hanxiaop:
@istio/wg-docs-maintainers-english can I get an approval for this?

content/en/docs/ops/ambient/usage/helm-installation/index.md
```diff
@@ -0,0 +1,176 @@
+---
+title: Installing Ambient Mesh with Helm
```
kfaseela, line 2:
```suggestion
title: Install Ambient Mesh with Helm
```

kfaseela: Can we have a separate Install section under Operations>Ambient Mesh, and then move this under there. Also it would make sense to remove the other empty Installation Guide page: https://deploy-preview-13871--preliminary-istio.netlify.app/latest/docs/ops/ambient/usage/install/

content/en/docs/ops/ambient/usage/helm-installation/index.md
```diff
@@ -0,0 +1,176 @@
...
+$ kubectl get pods -n istio-system
+{{< /text >}}
+
+### Verifying with the Sample Application
+
+After installing ambient with Helm, you can follow
+[Deploy the sample application](/docs/ops/ambient/getting-started/#bookinfo)
+guide to deploy the sample application and ingress gateways, and follow the
+[Adding your application to ambient](/docs/ops/ambient/getting-started/#addtoambient)
+to add your application to ambient.
```
kfaseela, line 100:
```suggestion
guide to deploy the sample application and ingress gateways, and then you can
[add your application to ambient](/docs/ops/ambient/getting-started/#addtoambient).
```

content/en/docs/ops/ambient/usage/helm-installation/index.md
```diff
@@ -0,0 +1,176 @@
...
+    $ helm delete istio-base -n istio-system
+    {{< /text >}}
+
+1. Delete the `istio-system` namespace:
+
+    {{< text syntax=bash snip_id=delete_system_namespace >}}
+    $ kubectl delete namespace istio-system
+    {{< /text >}}
+
+## Uninstall stable revision label resources
```
kfaseela, line 158: Do we have to specify revision related commands here? I cannot see anything related to this in the install section. and isn't this same as the usual helm revision commands, and becomes a repitition

content/en/docs/ops/ambient/usage/helm-installation/index.md
hanxiaop: Makes sense, I'll fix it. Thanks!

content/en/docs/ops/ambient/usage/helm-installation/index.md
hanxiaop: I think we can just remove this section.

content/en/docs/ops/ambient/usage/helm-installation/index.md
hanxiaop: @kfaseela Could you check this new section format? https://deploy-preview-13871--preliminary-istio.netlify.app/latest/docs/ops/ambient/install/

content/en/docs/ops/ambient/install/helm-installation/index.md
```diff
@@ -0,0 +1,164 @@
...
+    $ helm delete istio-base -n istio-system
+    {{< /text >}}
+
+1. Delete the `istio-system` namespace:
+
+    {{< text syntax=bash snip_id=delete_system_namespace >}}
+    $ kubectl delete namespace istio-system
+    {{< /text >}}
+
+### (Optional) Deleting CRDs Installed by Istio
```
kfaseela, line 157:
```suggestion
### Deleting CRDs Installed by Istio (Optional) 
```

kfaseela, line 157: i think this can be a bullet point in the previous uninstall section after "Delete Istio base chart"

hanxiaop: Do you mean a red warning message like https://istio.io/latest/docs/setup/getting-started/#install?

content/en/docs/ops/ambient/install/helm-installation/index.md
```diff
@@ -0,0 +1,164 @@
...
+
+Deleting CRDs removes any Istio resources you have created in your cluster.
+To delete Istio CRDs installed in your cluster:
+
+{{< text syntax=bash >}}
+$ kubectl get crd -oname | grep --color=never 'istio.io' | xargs kubectl delete
+{{< /text >}}
```
kfaseela, line 164:
```suggestion
1. Delete CRDs Installed by Istio (Optional)

{{< text syntax=bash >}}
$ kubectl get crd -oname | grep --color=never 'istio.io' | xargs kubectl delete
{{< /text >}}

1. Delete the `istio-system` namespace:

    {{< text syntax=bash snip_id=delete_system_namespace >}}
    $ kubectl delete namespace istio-system
    {{< /text >}}
```

content/en/docs/ops/ambient/install/helm-installation/index.md
kfaseela: just added a suggestion, you can check. It can be the next step after deleting the base chart, and before deleting the system namespace?

hanxiaop: The suggestion seems to be better organized, thanks!

content/en/docs/ops/ambient/install/helm-installation/index.md
```diff
@@ -147,18 +147,14 @@ installed above.
     {{< text syntax=bash snip_id=delete_base >}}
     $ helm delete istio-base -n istio-system
     {{< /text >}}
+1. Delete CRDs Installed by Istio (Optional)
+
+{{< text syntax=bash >}}
```
kfaseela, line 152: i think i had not aligned it properly :D

hanxiaop: Fixed, PTAL again, thanks!

kfaseela (approved):

kfaseela:
Is it okay to cherry-pick to 1.20?

hanxiaop:
> Is it okay to cherry-pick to 1.20?

@kfaseela IMO cherry-picking this allows us to gather more feedback before ambient beta for necessary improvements, and this doc already includes tests to verify the procedures. However, I'm not sure if docs like this typically got cherry-picked before.

PR #13873, opened 2023-09-14
Small correction in dual-stack blog

kfaseela:


ericvn:
@jacob-delgado @zhlsunshine Do these changes make sense?

jacob-delgado (approved):
The unification of the envs was done in Istio 1.19. We should also change that section as well. There's the 1.17-1.18 envs and the 1.19+. Maybe we should make it a tab view?

ericvn:
Thanks @jacob-delgado. I'll make it DNM and approve and @kfaseela can merge or update to a tab view if she wants.

ericvn (approved):

kfaseela:
> The unification of the envs was done in Istio 1.19. We should also change that section as well. There's the 1.17-1.18 envs and the 1.19+. Maybe we should make it a tab view?

The env variable unification is already done. and you can see it in istio.io from 1.19. See : https://istio.io/latest/blog/2023/experimental-dual-stack/#a-quick-experiment-using-dual-stack 

If tab view is needed, we can do it in a separate PR

PR #13879, opened 2023-09-15
Update index.md

eitansuez:
corrects example httproute for mesh traffic in k8s gw api ingress task.  the backendRefs should be part of the one rule that applies the filters.

- [x] Docs

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
@@ -334,7 +334,7 @@ spec:
         add:
         - name: my-added-header
           value: added-value
-  - backendRefs:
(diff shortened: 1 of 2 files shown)


frankbu:
@eitansuez Thanks for the fix. You'll need to run `make gen` to regenerate the corresponding snip file, or since it's a trivial change you can just make the equivalent change on line 235 of the corresponding snip.sh file: https://github.com/istio/istio.io/blob/209f70dffe90a37b2cff8e2e185a36d361336a55/content/en/docs/tasks/traffic-management/ingress/gateway-api/snips.sh#L235

eitansuez:
thank you @frankbu .  snips.sh is now updated.

ericvn (approved):

PR #13892, opened 2023-09-18
Update istio test ref, hardcode sha from Gateway API to non-release

ericvn:

Update istio test ref, do not strip sha from gateway version

The latest istio code is using a mid release version of the Gateway API, and thus we need a sha.


content/en/boilerplates/snips/args.sh
```diff
@@ -21,7 +21,7 @@
 ####################################################################################################
 
 ! read -r -d '' bpsnip_args_gateway_api_version <<\ENDSNIP
-v0.8.0
+f98e94a5305e3a08cb14f8a03470f8f3bdf6d54c
```
frankbu, line 24: I don't recall why we need to strip anything here. We really just want the docs to use the same version that the istio/istio is using, currently `v0.8.2-0.20230915021212-f98e94a5305e`

ericvn: @frankbu That's exactly what I hardcoded. That last part of `v0.8.2-0.20230915021212-f98e94a5305e` is the commit sha, and `f98e94a5305e3a08cb14f8a03470f8f3bdf6d54c` is the full sha. I do not currently have the makefile doing that conversion.

ericvn: If I try and strip the trailing items off (what the code used to do), you get `v0.8.2` and `github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.8.2` and since that release doesn't exist, the customize fails.

frankbu: Ah, right. So the problem is that when it's an official release we want to use it instead? Maybe the makefile can check if it's a proper release, use it, otherwise use the commit sha?

ericvn: A full blown solution might be if there are `-`'s, then we know we have to do some special logic. For RC's we strip of what's after the RC since that will exist). For others we have to strip of what's after the - (so here it would be `v0.8.2`) and then decrement the version (to v0.8.1). And we need special logic on case we have to go from v0.8.0 to 0.7.x (whatever the last one was (we can't just cement to 9).

frankbu: Yeah, but then we couldn't doc new things until after they're officially released. Seems that whenever there are `-` chars, the last part is a prefix of the sha we need. Isn't there a way to automatically find the full sha, given the prefix (i.e., `f98e94a5305e` -> `f98e94a5305e3a08cb14f8a03470f8f3bdf6d54c`)?

ericvn: The only way I know to get the full sha from the short sha is to git fetch the repo, git checkout the short sha and then get the full sha (we do this in bin/init.sh), but we are getting the repo so we can build istioctl for the appropriate commit. This seems excessive for here, but we could do it.

frankbu (approved):

content/en/boilerplates/snips/args.sh
dhawton: You can do it without cloning the repo which might be faster using the GitHub API. https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit -- If the repo is public, you do not need to provide an Authorization key and `REF` in this case accepts both short and full SHAs. The returned sha will always be full. For example:

```
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/kubernetes-sigs/gateway-api/commits/f98e94a5305e | jq -r .sha
```

Will return `f98e94a5305e3a08cb14f8a03470f8f3bdf6d54c`

ericvn: Thanks Daniel, let me try that.

ericvn:
Test failure is:
```
VERIFY FAILED snip_dns_capture_in_action_3 (timeout after 120s):
received:
"*   Trying 198.51.100.1:80...
* Connected to address.internal (198.51.100.1) port 80
> GET / HTTP/1.1
> Host: address.internal
> User-Agent: curl/8.3.0
> Accept: */*
>
< HTTP/1.1 503 Service Unavailable
upstream connect error or disconnect/reset before headers. retried and the latest reset reason: connection timeout< content-length: 114
< content-type: text/plain
< date: Mon, 18 Sep 2023 17:19:39 GMT
< server: envoy
<
{ [114 bytes data]
* Connection #0 to host address.internal left intact"
expected:
"* processing: address.internal
*   Trying 198.51.100.1:80..."
```

ericvn:
Moved the logic to a script as I was running into issues with if/else usage in a rule and things running earlier.

frankbu:
> The `processing: address.internal` is no longer in the output for some reason.

Yes, but the call is also timing out and failing. I don't think that's right either.
```
< HTTP/1.1 503 Service Unavailable
upstream connect error or disconnect/reset before headers. retried and the latest reset reason: connection timeout< content-length: 114
```

ericvn:
Also interesting that the specific test doesn't seem to have been run in earlier passing runs.

frankbu:
> The 503 seems to be the correct response.

If true, that's a pretty bad test since it will just hang until it times out and repeat, which will slow down the test. Maybe it used to be a IP that would respond, but things have changed?

Maybe the test should be changed to use an IP like `198.35.26.96`, similar to this egress test: https://preliminary.istio.io/latest/docs/tasks/traffic-management/egress/egress-kubernetes-services/#use-a-kubernetes-service-with-endpoints-to-access-an-external-service

PR #13942, opened 2023-09-28
1.19.1 Release Notes

ksubrmnn:

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [X] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/news/releases/1.19.x/announcing-1.19.1/index.md
+++ b/content/en/news/releases/1.19.x/announcing-1.19.1/index.md
@@ -1,0 +1,48 @@
+---
+title: Announcing Istio 1.19.1
+linktitle: 1.19.1
+subtitle: Patch Release
+description: Istio 1.19.1 patch release.
+publishdate: 2023-10-02
+release: 1.19.1
+---
+
+This release note describes what’s different between Istio 1.19.0 and 1.19.1.
+
+{{< relnote >}}
+
+## Changes
+
+- **Added** the ability to install the Gateway Helm chart with a dual-stack service definition.
+
+- **Added** a new configuration to `ProxyConfig` and `ProxyHeaders`. This allows customization of headers like `server`, `x-forwarded-client-cert`, etc. Most notably, these can now be disabled so that they are not modified.
+
+- **Added** a new configuration to `ProxyHeaders` and `MetadataExchangeHeaders`. The `IN_MESH` mode ensures `x-envoy-peer-metadata` and `x-envoy-peer-metadata-id`
+headers will not be added to outbound requests from sidecars to destination services considered mesh external.
+  ([Issue #17635](https://github.com/istio/istio/issues/17635))
+
+- **Fixed** an issue where the upgrade warning is given incorrectly between default and revisioned control planes.
+
+- **Fixed** an issue where ambient pods are incorrectly processed when the ambient namespace label is changed.
+
+- **Fixed** an issue where the Istio CNI plugin was not writing IPv6 iptables rules for dual stack clusters.  ([Issue #46625](https://github.com/istio/istio/issues/46625))
+
+- **Fixed** an issue where `meshConfig.defaultConfig.sampling` is ignored when there's only default providers.  ([Issue #46653](https://github.com/istio/istio/issues/46653))
+
+- **Fixed** SDS fetching timeout when we do not push back invalid certificate to Envoy.
+  ([Issue #46868](https://github.com/istio/istio/issues/46868))
+
+- **Fixed** an issue where the installation process was failing due to failed verification of the `NetworkAttachmentDefinition` resource.
+  ([Issue #46859](https://github.com/istio/istio/issues/46859))
+
+- **Fixed** metric `DNSNoEndpointClusters` not working.
+  ([Issue #46960](https://github.com/istio/istio/issues/46960))
+
+- **Fixed** the output of `istioctl proxy-config all` to include EDS configuration when the `--json` or `--yaml` flags are used.
+
+- **Fixed** an issue in control plane metrics causing gauge types to emit zero values without labels in addition to the expected metrics.
+  ([Issue #46977](https://github.com/istio/istio/issues/46977))
+
+## Security updates
+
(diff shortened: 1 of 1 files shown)


content/en/news/releases/1.19.x/announcing-1.19.1/index.md
```diff
@@ -0,0 +1,48 @@
+---
+title: Announcing Istio 1.19.1
+linktitle: 1.19.1
+subtitle: Patch Release
+description: Istio 1.19.1 patch release.
+publishdate: 2023-08-28
```
ericvn, line 6: Need a correct date

content/en/news/releases/1.19.x/announcing-1.19.1/index.md
```diff
@@ -0,0 +1,48 @@
...
+headers will not be added to outbound requests from sidecars to destination services considered mesh external.
+  ([Issue #17635](https://github.com/istio/istio/issues/17635))
+
+- **Fixed** an issue where the upgrade warning is given incorrectly between default and revisioned control planes.
+
+- **Fixed** an issue where ambient pods are incorrectly processed when the ambient namespace label is changed.
+
+- **Fixed** an issue where the Istio CNI plugin was not writing IPv6 iptables rules for dual stack clusters.  ([Issue #46625](https://github.com/istio/istio/issues/46625))
+
+- **Fixed** and issue where `meshConfig.defaultConfig.sampling` is ignored when there's only default providers.  ([Issue #46653](https://github.com/istio/istio/issues/46653))
```
ericvn, line 30:
```suggestion
- **Fixed** an issue where `meshConfig.defaultConfig.sampling` is ignored when there's only default providers.  ([Issue #46653](https://github.com/istio/istio/issues/46653))
```

content/en/news/releases/1.19.x/announcing-1.19.1/index.md
```diff
@@ -0,0 +1,48 @@
...
+
+- **Fixed** SDS fetching timeout when we do not push back invalid certificate to envoy.
```
ericvn, line 32:
```suggestion
- **Fixed** SDS fetching timeout when we do not push back invalid certificate to Envoy.
```

ericvn:
Note that there are several ways to do the istio patch release docs. I'm only mentioning this for information, not that I think the way this PR is doing things is wrong (I think this is actually correct).
1) Merging into the main branch. The problem is since 1.19 is the release shown at istio.io, we can't simply cherry-pick this PR to the `release-1.19` branch. It will fail as `https://github.com/istio/istio.io/blob/release-1.19/data/args.yml#L5` will need to be updated to 1.19.1. Note that after 1.20.1 ships, the `data/args.yml` points to the 1.20 version, so the PR can be cherry-picked from then forward.
2) Merge into the branch which is displayed by istio.io, which is currently the `release-1.19` branch. This will also not cherry-pick back to the main branch since you don't want the ``data/args.yml`.

Basically, until 1.20.0 releases, the 1.19 release notes PRs cannot be cherry-picked. I would recommend going with 1) above simply because other than having to manually do the cherry-pick for a while, the process doesn't change. 2) also requires the manual cherry-pick, but then you will have to change which release you check into first over time.

ericvn:
Only making DNM to hold until we want this to merge.

ksubrmnn:
> Note that there are several ways to do the istio patch release docs. I'm only mentioning this for information, not that I think the way this PR is doing things is wrong (I think this PR is actually correct).
> 
> 1. Merging into the main branch. The problem is since 1.19 is the release shown at istio.io, we can't simply cherry-pick this PR to the `release-1.19` branch. It will fail as `https://github.com/istio/istio.io/blob/release-1.19/data/args.yml#L5` will need to be updated to 1.19.1. Note that after 1.20.1 ships, the `data/args.yml` points to the 1.20 version, so the PR can be cherry-picked from then forward.
> 2. Merge into the branch which is displayed by istio.io, which is currently the `release-1.19` branch. This will also not cherry-pick back to the main branch since you don't want the ``data/args.yml`.
> 
> Basically, until 1.20.0 releases, the 1.19 release notes PRs cannot be cherry-picked. I would recommend going with 1) above simply because other than having to manually do the cherry-pick for a while, the process doesn't change. 2) also requires the manual cherry-pick, but then you will have to change which release you check into first over time.

Sounds good. For more information, I got a linting error when I tried to do this change off the release-1.19 branch.
```
ERROR Can't have a release note with a greater version then the site's version (1.19.1 vs 1.19.0)
```

This error went away when I worked off master.

ericvn:
> Sounds good. For more information, I got a linting error when I tried to do this change off the release-1.19 branch.
> 
> ```
> ERROR Can't have a release note with a greater version then the site's version (1.19.1 vs 1.19.0)
> ```
> 
> This error went away when I worked off master.

The fix for that lint error is to update the yaml file line I mentioned above (which is what prevents the cherry-pick from working).

content/en/news/releases/1.19.x/announcing-1.19.1/index.md
```diff
@@ -0,0 +1,48 @@
...
+
+- **Fixed** metric `DNSNoEndpointClusters` not working.
+  ([Issue #46960](https://github.com/istio/istio/issues/46960))
+
+- **Fixed** the output of `istioctl proxy-config all` to include EDS configuration when the `--json` or `--yaml` flags are used.
+
+- **Fixed** an issue in control plane metrics causing gauge types to emit zero values without labels in addition to the expected metrics.
+  ([Issue #46977](https://github.com/istio/istio/issues/46977))
+
+## Security update
```
ericvn, line 46: Minor nit (and may have been around a long time):
```suggestion
## Security updates
```
We use Changes and have multiple, but here our title expects a single security update while the following text has there are no updates. It seems the section title should be plural or the text singular for consistency. I went with the title since that seems consistent with the earlier title and we have multiple security updates a lot (most of the time?).

ericvn (approved):

PR #13983, opened 2023-10-09
Changing to the correct environment variable

ymesika:

Changing to the correct environment variable to get the home folder of the user who runs the command and not root.

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn:
> Changing to the correct environment variable to get the home folder of the user who runs the command and not root.

SUDO_USER is set to the login name of the user who invoked sudo, so it seems that what is currently in the docs is what you desire. This change would cause root to be used.

ymesika:
My bad, you are correct.

PR #14029, opened 2023-10-19
Add note to ambient quickstart mentioning minikube workaround

bleggett:
Document minikube-specific workaround config flag introduced in https://github.com/istio/istio/pull/47444

This may or may not be final, but it's current state and it's better to mention it.

- [x] Ambient
- [x] Docs
- [x] Installation
- [x] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/getting-started/index.md
+++ b/content/en/docs/ops/ambient/getting-started/index.md
@@ -64,6 +64,12 @@ Follow these steps to get started with ambient:
     Install Istio with the `ambient` profile on your Kubernetes cluster, using
     the `istioctl` command downloaded above:
 
+{{< tip >}}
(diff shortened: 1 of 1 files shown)


ericvn:
Another comment is that this change requires https://github.com/istio/istio/pull/47444 which at this point isn't in the release-1.20 branch while this documentation is now 1.20 specific (until 1.20.0 is released). If that's the case, we either need to wait with this documentation or get the function into the 1.20 branch.

bleggett:
Referenced changes have been cherrypicked into 1.20 so we should be consistent - ty @ericvn for catching that.

ericvn (approved):
Since the change is now in 1.20, we can merge this. If we find that we need more tips in this spot, I would like to make this message more generic and move the tips to the platform specific pages.

PR #14038, opened 2023-10-20
Use updated gateway-api -rc build versions

frankbu:

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/scripts/get_gateway_api_version.sh
+++ b/scripts/get_gateway_api_version.sh
@@ -18,16 +18,12 @@ set -e
 
 GATEWAY_VERSION=$(grep gateway-api go.mod | awk '{ print $2 }')
 #echo "GATEWAY_VERSION=${GATEWAY_VERSION}"
-if [[ $GATEWAY_VERSION == *"-rc"* ]]; then
-  GATEWAY_VERSION=$(echo "$GATEWAY_VERSION" | awk -F '.0.202' '{ print $1 }')
-#  echo "Found -rc, GATEWAY_VERSION=${GATEWAY_VERSION}"
-else
-  if [[ $GATEWAY_VERSION == *"-"* ]]; then
+if [[ $GATEWAY_VERSION == *"-"* ]]; then
+  #echo "Found -, GATEWAY_VERSION=${GATEWAY_VERSION}"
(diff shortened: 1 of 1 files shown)


ericvn:
This code doesn't do anything the case where there is an `-rc` in the name. As an example with the current istio master go.mod, I believe you will find GATEWAY_VERSION=v1.0.0-rc1.0.20231018214119-37d81593e5a9. When you run this through the process, you will end up trying to do a, which fails:
```
> kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v1.0.0-rc1.0.20231018214119-37d81593e5a9"                    
error: failed to run '/usr/bin/git fetch --depth=1 https://github.com/kubernetes-sigs/gateway-api v1.0.0-rc1.0.20231018214119-37d81593e5a9': fatal: couldn't find remote ref v1.0.0-rc1.0.20231018214119-37d81593e5a9
: exit status 128
```

frankbu:
With this fix, the current `go.mod`, with `v1.0.0-rc1.0.20231018214119-37d81593e5a9`, results in `github.com/kubernetes-sigs/gateway-api/config/crd?ref=37d81593e5a965ed76582dbc1a2f56bbd57c0622`, which works.

Btw, we're using this same algorithm in istio/istio now: https://github.com/istio/istio/pull/47503

ericvn (approved):

ericvn:
I believe the rc check as written can be removed as the versions will not start with ~rc and so you are just using the short sha in all the cases.

frankbu:
> I believe the rc check as written can be removed, have to look into this some more, as the versions will not start with -rc and so you are just using the short sha in all the cases.

The way it works now:

  If the version is `vN.N.N` or `vN.N.N-rcN` then use `ref=vN.N.N[-rcN]`, otherwise use `ref=sha`.

This way, we use the readable name for any "official" build and only use the sha if it's a daily build.

frankbu:
> the versions will not start with -rc and so you are just using the short sha in all the cases.

No, the check is a regex and it doesn't start with `^` so the `-rc` doesn't need to be at the start.

PR #14073, opened 2023-10-30
Fix incorrect link about code-blocks

kfaseela:
@ericvn : Not sure if this is correct. I am wondering why we have "Contribute to Documentation" under the Releases section. Should I move that outside?

https://preliminary.istio.io/latest/docs/releases/

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,5 +261,5 @@ checking errors that emerge.
 while section headings should use first letter capitalization only (e.g. "This is a valid heading").
 
 - Ensure that preformatted text blocks that reference files from the Istio GitHub repos use the @@ syntax
-to produce links to the content. See [here](https://istio.io/about/contribute/creating-and-editing-pages/#links-to-github-files)
+to produce links to the content. See [here](https://preliminary.istio.io/latest/docs/releases/contribute/code-blocks/#link-2-files)
 for context.


ericvn:
I agree that it doesn't make the most sense where it is. I would have to go back and see if it got moved as part of the new istio.io changes and what discussion we might have had at that point.

kfaseela:
> I agree that it doesn't make the most sense where it is. I would have to go back and see if it got moved as part of the new istio.io changes and what discussion we might have had at that point.

Ok, until then, let us atleast fix the README to point to the right place

ericvn:
https://github.com/istio/istio.io/pull/9324 was the change where it was moved. No discussion there.

README.md
```diff
@@ -261,5 +261,5 @@ checking errors that emerge.
 while section headings should use first letter capitalization only (e.g. "This is a valid heading").
 
 - Ensure that preformatted text blocks that reference files from the Istio GitHub repos use the @@ syntax
-to produce links to the content. See [here](https://istio.io/about/contribute/creating-and-editing-pages/#links-to-github-files)
+to produce links to the content. See [here](https://preliminary.istio.io/latest/docs/releases/contribute/code-blocks/#link-2-files)
```
ericvn, line 264: I wouldn't use preliminary:
```suggestion
to produce links to the content. See [here](https://istio.io/latest/docs/releases/contribute/code-blocks/#link-2-files)
```

kfaseela: Would this work for README as well? I was doubtful about that, as there are other places in the same file where local is not used

Issue #14094, opened 2023-11-03
update hugo to 0.120.3


Arhell:

ericvn:
Please wait to do this until after we release 1.20.0 and we have completed the istio.io branch cut.

ericvn:
Note that one reason not to always update Hugo is that we don't test everything, and some things are only done every 3 months.. As an example. I cannot do the istio.io branch cut with Hugo 1.20.? in the master branch tooling.  I specifically had to revert to the 1.15.x version in the release-1.19 branch tooling to do the branch cut.

PR #14098, opened 2023-11-05
update hugo to 0.120.4

Arhell:
resolve https://github.com/istio/istio.io/issues/14094
Let me know when the thread can be updated branch.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,7 @@
     publish = "public"
 
 [build.environment]
-    HUGO_VERSION = "0.120.3"
+    HUGO_VERSION = "0.120.4"
     NODE_VERSION = "18.18.2"
     BUILD_WITH_CONTAINER = "0"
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -2,7 +2,7 @@
     publish = "public"
 
 [build.environment]
-    HUGO_VERSION = "0.119.0"
+    HUGO_VERSION = "0.120.3"
```
craigbox:
```suggestion
    HUGO_VERSION = "0.120.4"
```

craigbox (changes requested):

netlify.toml
Arhell: Thank you, already

ericvn:
Just adding the DNM label to be removed once we have branched.

ericvn (approved):
LGTM, and we have branched.

craigbox:
LGTM!

PR #14100, opened 2023-11-06
Auto sort .spelling with `make gen`

hanxiaop:

When I try to add release notes, I spend some time finding the correct position for the words. Are there any concerns with not sorting it using `make gen`? Do we have to keep some words in a specific order? 

/cc @ericvn

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ilrudie (commented):
This looks like a good idea. It might be helpful to use `--ignore-case` on the sort command. So that `Abc` and `abc` would be located next to each other in the file.

Makefile.core.mk
```diff
@@ -85,7 +85,12 @@ site:
 snips:
 	@scripts/gen_snips.sh
 
-gen: tidy-go format-go update-gateway-version snips
+format-spelling:
+	@echo "Sorting the .spelling file..."
+	@sort .spelling -o .spelling
```
dhawton, line 90:
```suggestion
	@sort --ignore-case .spelling -o .spelling
```

Agree with Ian on this one. Otherwise, I don't think this is an issue. I'll defer to @frankbu and/or @ericvn for this but don't see an issue with having a sort (and this would be enforced via the gencheck test)

hanxiaop: Revised, thanks for the suggestion!

.spelling
```diff
@@ -1,15 +1,3 @@
-# markdown-spellcheck spelling configuration file
-# Format - lines beginning with # are comments
-# global dictionary is at the start, file overrides afterwards
-# one word per line, to define a file override use ' - filename'
-# followed by file specific exceptions, one per line
-# filename is relative to this configuration file
```
frankbu, line 6: Would be nice to retain these comments at the top of the file.

Also, even though we don't use file overrides currently, I think this sorting will mess those entries up. If we don't ever plan to use file overrides, I suppose the auto sort is a good thing.

dhawton: Could that be better served in the README potentially? Though it could also be done with a bit of shell script foo.

hanxiaop: > Would be nice to retain these comments at the top of the file.

Fixed.

> Also, even though we don't use file overrides currently, I think this sorting will mess those entries up.

I agree. However, I don't see any current examples, and not sure about its usage...

frankbu: >  I'll be happy to revise the script if we have any cases.

Even though we don't have any file overrides currently, I think it would be good to support it in the script so that it won't break if someone ever adds one.

I think the script just needs to `sed` for the first line in the file that matches ` - filename`, then only sort the lines up to that line and append the remaining lines without sorting. WDYT?

dhawton: He added a grep filter that does something similar but ignores all comments. LGTM but this is more infra so wanted to defer to you or Eric.

frankbu: > He added a grep filter that does something similar but ignores all comments.

Yes, I saw that. That's great. I'm suggesting he also handle the second problem (file overrides), probably pretty easy to do in his script using `sed`.

frankbu: Thinking about this some more, there are still 2 problems:

1. If we add any comments further down in the file (i.e., comment in front of some specific word, for example), it will get moved to the top of the file.
2. Any file overrides will not work.

Since comments don't really work, and the current comments are describing things (comments and file overrides) that don't work anyway, if we want to go with this auto sort feature, I think we need to remove all the comments (like it was before), and just use the basic list of words only, i.e., no comments or file overrides in the file.

@hanxiaop I suggest you revert your latest commit, since it doesn't really help, and then otherwise LGTM.

hanxiaop: @frankbu Done, thanks!

frankbu (approved):

ericvn:
I'm not sure why this code behaves differently when running in a build container on my local machine versus in the pipeline. If I run `make format-spelling` on my local machine is updates the files with many changes, which if I check in fails the `make gen-check` in the pipeline. I would have expected the `sort` to be the same on both platforms since it should be run inside a container using the same container image.

dhawton:
@ericvn what's the behavior you're seeing versus pipeline?

hanxiaop:
I'm curious as well. I got same results when doing `make format-spelling` on Mac/Ubuntu with & without BUILD_WITH_CONTAINER=1.

ericvn:
It appears the issue is probably related to the locale. I suspect my Ubuntu VM inherits the MacOS locale. Locale in build-container:
```
make shell
build-tools:/work$ locale
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=/home/ericvn/work/istio.io/diff.txt
build-tools:/work$
```
Here's a pointer to the difference: https://github.com/istio/istio.io/pull/14153/commits/f752dce14ed5ae64365b7b35b49a35a4689a9efe

It certainly looks like a lowercase letter before uppercase issue as well and maybe some numbering issues.

ericvn:
Verified that on my Mac, `LC_ALL=C make format-spelling` does not change the .spelling file.

ericvn:
Submitted https://github.com/istio/istio.io/pull/14165

hanxiaop:
> Verified that on my Mac, LC_ALL=C make format-spelling does not change the .spelling file.

Thanks for the fix.

PR #10648, opened 2021-12-19
chore(PR-template): add reference links area

Xunzhuo:
Please provide a description for what this PR is for.

And to help us figure out who should review this PR, please 
put an X in all the areas that this PR affects.

- [ ] Configuration Infrastructure
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Policies and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure

PR #14118, opened 2023-11-09
Change PR template

craigbox:
Make the PR template a little more obvious, and merge in the request change in #10648 to add a link to translation.


kfaseela (approved):

hanxiaop:
Looks like we also need the bot to automatically add a label? I see people manually adding `translation/chinese` to their PRs.

craigbox:
That would have to work on if the box is ticked, yes?  Or should it be set on all changes to the zh/docs dir?

PR #14165, opened 2023-11-15
Fix make format-spelling difference on MacOS

ericvn:
## Description

Fix make format-spelling differnce on MacOS. Running `make gen` on a Mac will result in a .spelling file update that if checked in fails the `make gen-check` in the pipeline. This is because MacOS and Linux use different locales by default, and sort uses the locale to determine its sorting. This change forces the C locale to be used on both platforms for the sort.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -85,7 +85,7 @@ site:
 snips:
 	@scripts/gen_snips.sh
 
-# Force locale since MacOS and Linux use differnet locales. Else MacOS users updates will
+# Force locale, since macOS and Linux use different locales. Otherwise updates from macOS users will
 # fail make gen-check with an incorrect (for the pipeline) .spelling.
 format-spelling:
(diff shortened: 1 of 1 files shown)


Makefile.core.mk
```diff
@@ -85,9 +85,11 @@ site:
 snips:
 	@scripts/gen_snips.sh
 
+# Force locale since MacOS and Linux use differnet locales. Else MacOS users updates will
```
craigbox, line 88:
```suggestion
# Force locale, since macOS and Linux use different locales. Otherwise updates from macOS users will
```

craigbox (approved):
LGTM with typo change I made to comment.

PR #14170, opened 2023-11-16
Improve some text in /ambient/usage/ztunnel/

windsonsea:
Have a try to improve the readability.

> One suggestion is to tweak the long lines around col 100-120.
> Currently, it's hard to find what have been changed in such a long sentence if no any wrapping.

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/usage/ztunnel/index.md
+++ b/content/en/docs/ops/ambient/usage/ztunnel/index.md
@@ -116,7 +116,7 @@ As noted earlier, some ambient functions may change as the project moves to beta
 
 It was noted earlier that traffic is always sent to a destination pod by first sending it to the ztunnel proxy on the same node as the destination pod. But what if the sender is either completely outside the Istio ambient mesh and hence does not initiate HBONE tunnels to the destination ztunnel first? What if the sender is malicious and trying to send traffic directly to an ambient pod destination, bypassing the destination ztunnel proxy?
 
-There are two scenarios here both of which are depicted in the following figure:
+There are two scenarios here, both of which are depicted in the following figure:
 
 1. Traffic stream B1 is being received by node W2 outside of any HBONE tunnel and directly addressed to ambient pod S1's IP address for some reason (possibly because the traffic source is not an ambient pod). As shown in the figure, the ztunnel traffic redirection logic intercepts such traffic and redirects it via the local ztunnel proxy for destination-side proxy processing and possible filtering based on AuthorizationPolicy prior to sending it into pod S1.
 1. Traffic stream G1 is being received by the ztunnel proxy of node W2 (possibly over an HBONE tunnel). However, the ztunnel proxy checks that the destination service requires waypoint processing and yet the source sending this traffic is not a waypoint or is not associated with this destination service. In this case, the ztunnel proxy hairpins the traffic towards one of the waypoints associated with the destination service from where it can then be delivered to any pod implementing the destination service (possibly to pod S1 itself, as shown in the figure).
--- a/content/en/docs/ops/ambient/usage/ztunnel/index.md
+++ b/content/en/docs/ops/ambient/usage/ztunnel/index.md
@@ -116,9 +116,10 @@ As noted earlier, some ambient functions may change as the project moves to beta
 
 It was noted earlier that traffic is always sent to a destination pod by first sending it to the ztunnel proxy on the same node as the destination pod. But what if the sender is either completely outside the Istio ambient mesh and hence does not initiate HBONE tunnels to the destination ztunnel first? What if the sender is malicious and trying to send traffic directly to an ambient pod destination, bypassing the destination ztunnel proxy?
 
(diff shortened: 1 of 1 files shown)


ericvn:
The lines are still very long. Just splitting the line, for example 119, into multiple shorter lines will not have any effect on the document. A blank like will create a paragraph.

content/en/docs/ops/ambient/usage/ztunnel/index.md
```diff
@@ -293,14 +295,19 @@ $ kubectl -n istio-system logs -l app=ztunnel | grep -E "inbound|outbound"
...
+
+* [Prometheus installation](/docs/ops/integrations/prometheus/#installation)
+* [Kiali installation](/docs/ops/integrations/kiali/#installation)
+* [Istio metrics](/docs/reference/config/metrics/)
+* [Querying Metrics from Prometheus](/docs/tasks/observability/metrics/querying-metrics/)
 
-One point to note is that in case of a service that is only using ztunnel and L4 networking, the Istio metrics reported will currently only be the L4/ TCP metrics (namely `istio_tcp_sent_bytes_total`, `istio_tcp_received_bytes_total`, `istio_tcp_connections_opened_total`, `istio_tcp_connections_closed_total`). The full set of Istio and Envoy metrics will be reported when a Waypoint proxy is involved.
+One point to note is that in case of a service that is only using ztunnel and L4 networking, the Istio metrics reported will currently only be the L4 or TCP metrics (namely `istio_tcp_sent_bytes_total`, `istio_tcp_received_bytes_total`, `istio_tcp_connections_opened_total`, `istio_tcp_connections_closed_total`). The full set of Istio and Envoy metrics will be reported when a Waypoint proxy is involved.
```
ilrudie, line 310: It's probably better without either `/` or `or` I think. "Layer 4 TCP metrics" makes sense, yeah?
```suggestion
One point to note is that in case of a service that is only using ztunnel and L4 networking, the Istio metrics reported will currently only be the L4 TCP metrics (namely `istio_tcp_sent_bytes_total`, `istio_tcp_received_bytes_total`, `istio_tcp_connections_opened_total`, `istio_tcp_connections_closed_total`). The full set of Istio and Envoy metrics will be reported when a Waypoint proxy is involved.
```

content/en/docs/ops/ambient/usage/ztunnel/index.md
```diff
@@ -116,7 +116,9 @@ As noted earlier, some ambient functions may change as the project moves to beta
...
+There are two scenarios here both of which are depicted in the following figure. In the first scenario, traffic stream B1 is being received by node W2 outside of any HBONE tunnel and directly addressed to ambient pod S1's IP address for some reason (possibly because the traffic source is not an ambient pod). As shown in the figure, the ztunnel traffic redirection logic intercepts such traffic and redirects it via the local ztunnel proxy for destination-side proxy processing and possible filtering based on AuthorizationPolicy prior to sending it into pod S1.
+
+In the second scenario, traffic stream G1 is being received by the ztunnel proxy of node W2 (possibly over an HBONE tunnel). However, the ztunnel proxy checks that the destination service requires waypoint processing and yet the source sending this traffic is not a waypoint or is not associated with this destination service. In this case, the ztunnel proxy hairpins the traffic towards one of the waypoints associated with the destination service from where it can then be delivered to any pod implementing the destination service (possibly to pod S1 itself, as shown in the figure).
```
ilrudie, line 121: Does it make more sense to enumerate these scenarios as numbered bullet points?
```suggestion
There are two scenarios here both of which are depicted in the following figure:
1. Traffic stream B1 is being received by node W2 outside of any HBONE tunnel and directly addressed to ambient pod S1's IP address for some reason (possibly because the traffic source is not an ambient pod). As shown in the figure, the ztunnel traffic redirection logic intercepts such traffic and redirects it via the local ztunnel proxy for destination-side proxy processing and possible filtering based on AuthorizationPolicy prior to sending it into pod S1.
1. Traffic stream G1 is being received by the ztunnel proxy of node W2 (possibly over an HBONE tunnel). However, the ztunnel proxy checks that the destination service requires waypoint processing and yet the source sending this traffic is not a waypoint or is not associated with this destination service. In this case, the ztunnel proxy hairpins the traffic towards one of the waypoints associated with the destination service from where it can then be delivered to any pod implementing the destination service (possibly to pod S1 itself, as shown in the figure).
```

ilrudie (approved):
Overall looks reasonable. Couple of suggestions but they're non-blocking IMO.

content/en/docs/ops/ambient/usage/ztunnel/index.md
windsonsea: Yes, it's clearer now.

content/en/docs/ops/ambient/usage/ztunnel/index.md
```diff
@@ -116,7 +116,10 @@ As noted earlier, some ambient functions may change as the project moves to beta
...
+There are two scenarios here both of which are depicted in the following figure:
```
craigbox, line 119:
```suggestion
There are two scenarios here, both of which are depicted in the following figure:
```

craigbox (approved):
LGTM, thanks!  I also have suggestions on improving this page but we'll do it iteratively.

Issue #14126, opened 2023-11-09
Add proper SNI+wildcard based egress rouging.


libesz:
I've been working on to fill this gap after `forward_downstream_sni` and a few other Istio specific envoy features were dropped. See https://github.com/istio/proxy/pull/4958

I assume many Istio users will face this once the missing filters breaks their egress traffic more or less silently. My use-case is exactly the one which was dropped from the official docs without any replacement after 1.13.

The use-case: route TLS traffic to a subset of hostnames (including wildcard hostnames) to the exact destination that they are targeting. This requires dynamic routing, which the first class Istio resources cannot really express. The original docs suggested to use a dedicated nginx container to do the dynamic SNI dispatch part per connection and the actual forwarding + proxying. This allows the egress traffic restriction of the application namespace, so that targeting the non-listed hostnames can be blocked with i.e. NetworkPolicies. 
See: https://istio.io/v1.13/docs/tasks/traffic-management/egress/wildcard-egress-hosts/

The idea now is to use the built-in SNI forwarder in Envoy instead of a standalone container: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/sni_dynamic_forward_proxy/v3/sni_dynamic_forward_proxy.proto

I developed two PoCs around this:

### Double listener:
One using two independent Envoy listeners to work around the problem that TLS inspection can inspect only the outer TLS handshake, which is Istios mTLS. Without the `forward_downstream_sni` filter, the sidecar sends the egress GW service name in the mTLS session, so a second listener+inspection is needed to determine the final destination.

PoC code: https://gist.github.com/libesz/961ac221f210f8a2a3ff3b8e69ad9cc2

![istio-sni-double](https://github.com/istio/istio.io/assets/311062/e44aa2fc-c21a-4456-9459-9216d9465bce)

### Single listener:
The other one is using the new Envoy set-filter-state extension that is able to do the job of `forward_downstream_sni`, so a single (heavily modified) Istio `Gateway` can do but the mesh mTLS termination and the SNI forwarding. New filter docs: https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/set_filter_state#config-network-filters-set-filter-state

PoC code: https://gist.github.com/libesz/804ad714d47d57ac44c8f2d8dd7ba1f6

![istio-sni-single](https://github.com/istio/istio.io/assets/311062/7b471507-932c-4404-95da-59087be0f2ea)

---
[...]

PR #14188, opened 2023-11-20
Blog post for egress SNI routing.

libesz:
## Description

Blog post for https://github.com/istio/istio.io/issues/14126

## Reviewers

TBD

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,348 @@
...
+# VirtualService that will instruct sidecars in the mesh to route the outgoing
+# traffic to the egress gateway Service if the SNI target hostname matches
+  - apiVersion: networking.istio.io/v1alpha3
+    kind: VirtualService
+    metadata:
+      name: direct-wildcard-through-egress-gateway
+      namespace: istio-system
+    spec:
+      hosts:
+        - "*.foo.com"
```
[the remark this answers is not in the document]
libesz: The problem with wikipedia is that I could not find any second service on their domains which are served by a different set of IPs, so the use-case is not valid with it unfortunately 🤷 . Not sure what else should I use which is generic enough for an example. I could certainly use IBM targets for example...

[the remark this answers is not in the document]
frankbu: True that it's not needed for wekipedia, but wikipedia can still be used to show the config is set up properly and working, and it shows that it can be used for any wildcard host without needing to care about what underlying servers are involved. The old doc used wikipedia to demonstrate. That said, you could use something as general as `*.com` and show that you can access `google.com` and `ibm.com` to demonstrate, but if you want a more realistic example, I don't see anything wrong with just picking a couple `ibm.com` subdomains to demonstrate.

Arhell (commented):
/ok-to-test

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,453 @@
...
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: direct-wildcard-through-egress-gateway
+  namespace: istio-system
+spec:
+  hosts:
+    - "*.wikipedia.org"
+  gateways:
+  - mesh
```
frankbu, line 184: This isn't needed since `mesh` is the default.
```suggestion
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,453 @@
...
+  tls:
+  - match:
+    - gateways:
+      - mesh
```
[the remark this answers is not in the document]
libesz: good idea... I believe there was a revision earlier where these were combined :)
now combined again

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,442 @@
...
+      port: 443
+      sniHosts:
+        - "*.wikipedia.org"
+    route:
+    - destination:
+        host: egressgateway.istio-egress.svc.cluster.local
+        subset: wildcard
+# Dummy routing instruction. If omitted, no reference will point to the Gateway
```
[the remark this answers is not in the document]
howardjohn: I dont. All EnvoyFilter usage is kludgy and is defaulting to "only bad ways" :slightly_smiling_face:

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,442 @@
...
+title: "[WIP] Routing HTTPS/TLS traffic to arbitrary remote destinations"
+description: "A generic approach to set up egress gateways and use them to route traffic to a restricted set of target remote hosts dynamically, supporting wildcard domains."
+publishdate: 2023-12-10
+attribution: "Gergő Huszty (IBM)"
+keywords: [traffic-management,gateway,mesh,mtls,egress,remote]
+---
+
+If you are using Istio to handle application originated outbound traffic, you are probably familiar with the concept of
+egress gateways. Those are specific in a way that they are forwarding and routing mesh application originated traffic
+and not exposing their services to the external world. This is a useful feature if your system is operating in a restricted
```
frankbu, line 11:
```suggestion
If you are using Istio to handle application-originated traffic to destinations outside of the mesh, you're probably familiar with the concept of egress gateways.
Egress gateways can be used to monitor and forward traffic from mesh-internal applications to locations outside of the mesh.
This is a useful feature if your system is operating in a restricted
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,442 @@
...
+environment and you want to control what can be reached on the public internet from your mesh.
+
+## Background
+
+The use-case is documented in the official Istio docs until 1.13, but parts related to arbitrary target domains got removed
+as it caused unfortunate confusions.
+However, the old solution was still usable until now, when Istio 1.20 dropped some code from Envoy which made the approach work.
+This post attempts to describe how we resolved the issue and fill the gap with an as-similar-as-possible approach. We will use
+generic and Istio version independent components and Envoy features. That should allow users of the old solution to seamlessly
+migrate before their systems are facing the breaking changes in Istio 1.20.
```
frankbu, line 21:
```suggestion
The use-case of configuring an egress gateway to handle arbitrary wildcard domains had been included in the [official Istio docs](https://istio.io/v1.13/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#wildcard-configuration-for-arbitrary-domains) up until version 1.13, but was subsequently removed because the documented solution was not officially supported or recommended and was subject to breakage in future versions of Istio.
Nevertheless, the old solution was still usable with Istio versions before 1.20. Istio 1.20, however, dropped some Envoy functionality that was required for the approach to work.

This post attempts to describe how we resolved the issue and filled the gap with a similar approach using Istio version-independent components and Envoy features, but without the need for a separate Nginx SNI proxy.
Our approach allows users of the old solution to seamlessly migrate configurations before their systems face the breaking changes in Istio 1.20.
```

content/en/blog/2023/egress-sni/index.md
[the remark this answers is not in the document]
libesz: Thanks! Will leave it like that.

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+---
+
+If you are using Istio to handle application-originated traffic to destinations outside of the mesh, you're probably familiar with the concept of egress gateways.
+Egress gateways can be used to monitor and forward traffic from mesh-internal applications to locations outside of the mesh.
+This is a useful feature if your system is operating in a restricted
+environment and you want to control what can be reached on the public internet from your mesh.
+
+## Background
+
+The use-case of configuring an egress gateway to handle arbitrary wildcard domains had been included in the [official Istio docs](/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#wildcard-configuration-for-arbitrary-domains) up until version 1.13, but was subsequently removed because the documented solution was not officially supported or recommended and was subject to breakage in future versions of Istio.
```
frankbu, line 16: I think it's better to link to the old doc here:
```suggestion
The use-case of configuring an egress gateway to handle arbitrary wildcard domains had been included in the [official Istio docs](https://archive.istio.io/v1.13/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#wildcard-configuration-for-arbitrary-domains) up until version 1.13, but was subsequently removed because the documented solution was not officially supported or recommended and was subject to breakage in future versions of Istio.
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+(the hostname) is statically determined by a `VirtualService`, that tells Envoy in the egress gateway pod where to TCP proxy
+the matching outbound connections. You can use multiple and even wildcard DNS names to match the routing criteria, but you
+are not able to route the traffic exactly where the application wanted to. For example you can match traffic that targets
+`*.wikipedia.org`, but you can direct everything into a single final target, i.e. `en.wikipedia.org`. If there is a
+service on `anyservice.wikipedia.org` that is not hosted by the same server(s) as `en.wikipedia.org`, your traffic to that
+arbitrary host will fail. In fact the connection would go to `en.wikipedia.org` servers, while the target hostname in the
+TLS handshake and the HTTP payload would contain `anyservice.wikipedia.org`, which may not be served by the same servers.
+The solution on high level is to inspect the original server name (SNI extension) in the application TLS handshake (that is sent
+in plain-text, so no TLS termination or other man-in-the-middle operation is needed) in every new connection and use it as
+a target dynamically to TCP proxy the traffic when leaves the gateway.
```
frankbu, line 34:
```suggestion
The currently documented egress gateway use-cases rely on the fact that the target of the traffic
(the hostname) is statically configured in a `VirtualService`, telling Envoy in the egress gateway pod where to TCP proxy
the matching outbound connections. You can use multiple, and even wildcard, DNS names to match the routing criteria, but you
are not able to route the traffic to the exact location specified in the application request. For example you can match traffic for targets
`*.wikipedia.org`, but you then need to forward the traffic to a single final target, e.g., `en.wikipedia.org`. If there is another
service, e.g., `anyservice.wikipedia.org` that is not hosted by the same server(s) as `en.wikipedia.org`, traffic to that host will fail because, even though the target hostname in the
TLS handshake of the HTTP payload contains `anyservice.wikipedia.org`, the `en.wikipedia.org` servers will not be be able to serve the request.

The solution to this problem at a high level is to inspect the original server name (SNI extension) in the application TLS handshake (which is sent
in plain-text, so no TLS termination or other man-in-the-middle operation is needed) in every new gateway connection and use it as
the target to dynamically TCP proxy the traffic leaving the gateway.
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+## Further security measures
+
+Now that we are restricting the traffic, we should be able to lock down the egress gateways in a way, that they can be used
+only from within the mesh. This is achieved by enforcing the `ISTIO_MUTUAL` or mTLS peer authentication between the application
+sidecar and the gateway. That means, there will be two layers of TLS on the application L7 payload. One that is the application
+originated end-to-end TLS session terminated by the final remote target, and another one that is the Istio mTLS session.
+
+Also, in order to mitigate application pod corruption, not only the application sidecar performs hostname list checks, but
+the gateway as well. So that any compromised application pod can only still access the allowed targets, nothing more.
```
frankbu, line 44:
```suggestion
When restricting egress traffic via egress gateways, we need to lock down the egress gateways so that they can only be used
by clients within the mesh. This is achieved by enforcing `ISTIO_MUTUAL` (mTLS peer authentication) between the application
sidecar and the gateway. That means that there will be two layers of TLS on the application L7 payload. One that is the application
originated end-to-end TLS session terminated by the final remote target, and another one that is the Istio mTLS session.

Another thing to keep in mind is that in order to mitigate any potential application pod corruption, the application sidecar and the gateway should both perform hostname list checks.
This way, any compromised application pod will still only be able to access the allowed targets and nothing more.
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+attribution: "Gergő Huszty (IBM)"
+keywords: [traffic-management,gateway,mesh,mtls,egress,remote]
+---
+
+If you are using Istio to handle application-originated traffic to destinations outside of the mesh, you're probably familiar with the concept of egress gateways.
+Egress gateways can be used to monitor and forward traffic from mesh-internal applications to locations outside of the mesh.
+This is a useful feature if your system is operating in a restricted
+environment and you want to control what can be reached on the public internet from your mesh.
+
+## Background
```
frankbu, line 14: I think this heading isn't needed.

```suggestion
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+`EnvoyFilter`s by introducing internal Envoy listeners and clusters that will make it happen. The last thing is to patch the
+internal destination of the gateway implemented TCP proxy to the internal SNI forwarder.
+
+## Solution
+
+{{< image width="90%" link="./egress-sni-flow.png" alt="Egress SNI routing with arbitrary domain names" title="Egress SNI routing with arbitrary domain names" caption="Egress SNI routing with arbitrary domain names" >}}
+
+In order to deploy the sample configuration, create the `istio-egress` namespace and deploy the gateway along with some RBAC
+and it's `Service`. We use the gateway injection method in the example. Depending on your install method, you may want to
+deploy it differently (for example through an `IstioOperator` CR, or through Helm).
```
frankbu, line 64:
```suggestion
## Deploy the sample

In order to deploy the sample configuration, start by creating the `istio-egress` namespace and then use the follwowing YAML to deploy an egress gateway, along with some RBAC
and its `Service`. We use the gateway injection method to create the gateway in this example but, depending on your install method, you may want to
deploy it differently (for example, using an `IstioOperator` CR or using Helm).
```

content/en/blog/2023/egress-sni/index.md
libesz: I see! Sorry I removed the beginning as the linter complained about the full URL. It insisted to shorten it. Will try to figure out something that passes the linter...

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+internal destination of the gateway-implemented TCP proxy to the internal SNI forwarder.
+
+The end-to-end request flow is shown in the following diagram:
+
+{{< image width="90%" link="./egress-sni-flow.png" alt="Egress SNI routing with arbitrary domain names" title="Egress SNI routing with arbitrary domain names" caption="Egress SNI routing with arbitrary domain names" >}}
+
+## Deploy the sample
+
+In order to deploy the sample configuration, start by creating the `istio-egress` namespace and then use the following YAML to deploy an egress gateway, along with some RBAC
+and its `Service`. We use the gateway injection method to create the gateway in this example but, depending on your install method, you may want to
```
frankbu, line 63:
```suggestion
and its `Service`. We use the gateway injection method to create the gateway in this example. Depending on your install method, you may want to
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+      value:
+        name: envoy.tcp_proxy
+        typed_config:
+          "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
+          stat_prefix: tcp
+          cluster: sni_cluster
+{{< /text >}}
+
+Check the logs of `istiod` and the gateway pod for any errors of warnings. If all goes well, your mesh sidecars are now routing
+`*.wikipedia.org` towards your gateway pod, while the gateway pod is routing to the exact remote host based on the application
```
frankbu, line 356:
```suggestion
Check the `istiod` and gateway logs for any errors or warnings. If all went well, your mesh sidecars are now routing
`*.wikipedia.org` requests to your gateway pod while the gateway pod is then forwarding them to the exact remote host specified in the application
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+request.
+
+## Testing
```
frankbu, line 359:
```suggestion
## Try it out
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+Following other examples, we will use the sleep pod. Assuming there is automatic sidecar injection in your default namespace, deploy
+the test app:
```
frankbu, line 362:
```suggestion
Following other Istio egress examples, we will use the
[sleep]({{< github_tree >}}/samples/sleep) pod as a test source for sending requests.
Assuming automatic sidecar injection is enabled in your default namespace, deploy
the test app using the following command:
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+{{< text bash >}}
+$ kubectl apply -f {{< github_file >}}/samples/sleep/sleep.yaml
+{{< /text >}}
+
+Get your pod and gateway name:
```
frankbu, line 368:
```suggestion
Get your sleep and gateway pods:
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+{{< text bash >}}
+$ export SOURCE_POD=$(kubectl get pod -l app=sleep -o jsonpath={.items..metadata.name})
+$ export GATEWAY_POD=$(kubectl get pod -n istio-egress -l istio=egressgateway -o jsonpath={.items..metadata.name})
+{{< /text >}}
+
+Test that you are able to connect to a `wikipedia.org` site:
```
frankbu, line 375:
```suggestion
Run the following command to confirm that you are able to connect to the `wikipedia.org` site:
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+{{< text bash >}}
+$ kubectl exec "$SOURCE_POD" -c sleep -- sh -c 'curl -s https://en.wikipedia.org/wiki/Main_Page | grep -o "<title>.*</title>"; curl -s https://de.wikipedia.org/wiki/Wikipedia:Hauptseite | grep -o "<title>.*</title>"'
+<title>Wikipedia, the free encyclopedia</title>
+<title>Wikipedia – Die freie Enzyklopädie</title>
+{{< /text >}}
+
+We could reach both English and German Wikipedia, great! To have something to compare, let's check another site:
```
frankbu, line 383:
```suggestion
We could reach both English and German `wikipedia.org` subdomains, great!

Normally, in a production environment, we would [block external requests](/docs/tasks/traffic-management/egress/egress-control/#change-to-the-blocking-by-default-policy) that are not configured to redirect through the egress gateway, but since we didn't do that in our test environment, let's access another external site for comparison:
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+{{< text bash >}}
+$ kubectl exec "$SOURCE_POD" -c sleep -- sh -c 'curl -s https://cloud.ibm.com/login | grep -o "<title>.*</title>"'
+<title>IBM Cloud</title>
+{{< /text >}}
+
+Now that we turned on access logging globally (with the `Telemetry` CR in the manifest), we can double-check what happened in the proxies.
```
frankbu, line 390:
```suggestion
Since we have access logging turned on globally (with the `Telemetry` CR in the manifest), we can now inspect the logs to see how the above requests were handled by the proxies.
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+First, check the gateway:
```
frankbu, line 392:
```suggestion
First, check the gateway logs:
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+[2023-11-24T13:21:53.000Z] "- - -" 0 - - - "-" 821 92848 49 - "-" "-" "-" "-" "185.15.59.224:443" dynamic_forward_proxy_cluster 172.17.5.170:48278 envoy://sni_listener/ envoy://internal_client_address/ de.wikipedia.org -
+[2023-11-24T13:21:53.000Z] "- - -" 0 - - - "-" 1539 93646 50 - "-" "-" "-" "-" "envoy://sni_listener/" sni_cluster envoy://internal_client_address/ 172.17.5.170:8443 172.17.34.35:55108 outbound_.443_.wildcard_.egressgateway.istio-egress.svc.cluster.local -
+{{< /text >}}
+
+There are four log entries, representing two of the three curl requests. Each pair shows how a single request is flowing through the envoy traffic processing pipeline twice.
+They are printed in reverse order, but we can see the 2nd and the 4th line saying connection arrived to the gateway service and passed through the internal `sni_cluster` target.
+The 1st and 3rd line shows the final target is determined from the inner SNI header. This is the one set by the original application.
+Then it is forwarded to `dynamic_forward_proxy_cluster` which is the final entity in envoy that is talking to the final remote target.
```
frankbu, line 406:
```suggestion
There are four log entries, representing two of our three curl requests. Each pair shows how a single request flows through the envoy traffic processing pipeline.
They are printed in reverse order, but we can see the 2nd and the 4th line show that the requests arrived at the gateway service and were passed through the internal `sni_cluster` target.
The 1st and 3rd line show that the final target is determined from the inner SNI header, i.e., the target host set by the application.
The request is forwarded to `dynamic_forward_proxy_cluster` which finally sends on the request from Envoy to the remote target.
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+{{< text bash >}}
+$ kubectl logs $SOURCE_POD -c istio-proxy
+[...]
+[2023-11-24T13:21:52.793Z] "- - -" 0 - - - "-" 813 111152 61 - "-" "-" "-" "-" "172.17.5.170:8443" outbound|443|wildcard|egressgateway.istio-egress.svc.cluster.local 172.17.34.35:55102 208.80.153.224:443 172.17.34.35:37020 en.wikipedia.org -
+[2023-11-24T13:21:52.994Z] "- - -" 0 - - - "-" 821 92848 55 - "-" "-" "-" "-" "172.17.5.170:8443" outbound|443|wildcard|egressgateway.istio-egress.svc.cluster.local 172.17.34.35:55108 208.80.153.224:443 172.17.34.35:37030 de.wikipedia.org -
+[2023-11-24T13:21:55.197Z] "- - -" 0 - - - "-" 805 15199 158 - "-" "-" "-" "-" "104.102.54.251:443" PassthroughCluster 172.17.34.35:45584 104.102.54.251:443 172.17.34.35:45582 cloud.ibm.com -
+{{< /text >}}
+
+As you can see, Wikipedia requests were sent through the gateway, while the request towards IBM Cloud went straight out from the application pod to the internet. This is indicated with the `PassthroughCluster` log.
```
frankbu, line 418:
```suggestion
As you can see, Wikipedia requests were sent through the gateway while the request to IBM Cloud went straight out from the application pod to the internet, as indicated by the `PassthroughCluster` log.
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+That was all! I hope it was helpful.
+If you are an existing user of the previous nginx based solution,
+you are now free to migrate to this approach before upgrading to Istio 1.20, which would disrupt your current setup.
+
+Happy SNI routing!
+
+## References
+
+* [Envoy docs for the SNI forwarder](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/sni_dynamic_forward_proxy_filter)
+* [Previous solution with nginx as an SNI proxy container in the gateway](/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#wildcard-configuration-for-arbitrary-domains)
```
frankbu, line 441:
```suggestion
* [Previous solution with Nginx as an SNI proxy container in the gateway](https://archive.istio.io/v1.13/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#wildcard-configuration-for-arbitrary-domains)
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+{{< /text >}}
+
+As you can see, Wikipedia requests were sent through the gateway, while the request towards IBM Cloud went straight out from the application pod to the internet. This is indicated with the `PassthroughCluster` log.
+
+## Conclusion
+
+We implemented a selective routing for egress HTTPS/TLS traffic with egress gateways, supporting arbitrary and wildcard domain names. Of course the example
+can be extended with HA requirements (i.e. with adding zone aware gateway `Deployment`s, etc.). From here you are free to restrict the direct external
+network access of your application, for example with network policies that denies traffic to tcp port 443 in the application namespace.
+Once that is in place, the application cannot directly talk to the public network, only through the gateway, which is limited to a predefined set of remote hostnames.
```
frankbu, line 425:
```suggestion
We implemented controlled routing for egress HTTPS/TLS traffic using egress gateways, supporting arbitrary and wildcard domain names. In a production environment, the example shown in this post
would be extended to support HA requirements (e.g., adding zone aware gateway `Deployment`s, etc.) and to restrict the direct external
network access of your application so that the application can only access the public network through the gateway, which is limited to a predefined set of remote hostnames.
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+
+The solution scales easily. You can pick up multiple domain names to the list, and they are allow-listed as soon as you roll it out!
+No need to set up per domain `VirtualService`s or other routing details. Beware the list appears at multiple places, so if you use
+some tooling for CI/CD (i.e. Kustomize), the domain name list is better to be extracted to a single place, from where you render into the final
+places.
```
frankbu, line 430:
```suggestion
The solution scales easily. You can include multiple domain names in the configuration, and they will be allow-listed as soon as you roll it out!
No need to configure per domain `VirtualService`s or other routing details. Be careful, however, as the domain names are listed in multiple places in the config. If you use
tooling for CI/CD (e.g., Kustomize), it's best to extract the domain name list into a single place from which you can render into the required configuration resources.
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
+---
+title: "[WIP] Routing HTTPS/TLS traffic to arbitrary remote destinations"
+description: "A generic approach to set up egress gateways and use them to route traffic to a restricted set of target remote hosts dynamically, supporting wildcard domains."
```
frankbu, line 3:
```suggestion
title: "Routing egress gateway traffic to wildcard destinations"
description: "A generic approach to set up egress gateways that can route traffic to a restricted set of target remote hosts dynamically, including wildcard domains."
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+are not able to route the traffic to the exact location specified in the application request. For example you can match traffic for targets
+`*.wikipedia.org`, but you then need to forward the traffic to a single final target, e.g., `en.wikipedia.org`. If there is another
+service, e.g., `anyservice.wikipedia.org`, that is not hosted by the same server(s) as `en.wikipedia.org`, traffic to that host will fail because, even though the target hostname in the
+TLS handshake of the HTTP payload contains `anyservice.wikipedia.org`, the `en.wikipedia.org` servers will not be be able to serve the request.
+
+The solution to this problem at a high level is to inspect the original server name (SNI extension) in the application TLS handshake (which is sent
+in plain-text, so no TLS termination or other man-in-the-middle operation is needed) in every new gateway connection and use it as
+the target to dynamically TCP proxy the traffic leaving the gateway.
+
+## Further security measures
```
frankbu, line 34:
```suggestion
```

content/en/blog/2023/egress-sni/index.md
```diff
@@ -0,0 +1,441 @@
...
+Our approach allows users of the old solution to seamlessly migrate configurations before their systems face the breaking changes in Istio 1.20.
+
+## Problem to solve
+
+The currently documented egress gateway use-cases rely on the fact that the target of the traffic
+(the hostname) is statically configured in a `VirtualService`, telling Envoy in the egress gateway pod where to TCP proxy
+the matching outbound connections. You can use multiple, and even wildcard, DNS names to match the routing criteria, but you
+are not able to route the traffic to the exact location specified in the application request. For example you can match traffic for targets
+`*.wikipedia.org`, but you then need to forward the traffic to a single final target, e.g., `en.wikipedia.org`. If there is another
+service, e.g., `anyservice.wikipedia.org`, that is not hosted by the same server(s) as `en.wikipedia.org`, traffic to that host will fail because, even though the target hostname in the
```
windsonsea, line 27:
```suggestion
service, e.g., `anyservice.wikipedia.org`, that is not hosted by the same server(s) as `en.wikipedia.org`, the traffic to that host will fail. This is because, even though the target hostname in the
```
It would be better to separate the long sentence into multiple shorter

content/en/blog/2023/egress-sni/index.md
libesz: Shouldn't `Routing egress gateway traffic` be just `Routing egress traffic`?

frankbu (approved):
Nice article, LGTM. Thanks @libesz!

PR #14196, opened 2023-11-27
Small update to the dual-stack documentation

kfaseela:
As per @jacob-delgado's comments, we need to remove the global flag because it is currently unused. But eventually may get used as a flag to enable dual-stack globally. Until then it is better to hide it from the docs.


ericvn (approved):

jacob-delgado:
LGTM. Thanks!

Issue #14210, opened 2023-11-30
Re-enable tasks/observability/logs/telemetry-api/test.sh


ericvn:
The test is failing when moving up to a recent istio/istio commit. This test should be re-enabled before we ship 1.21.

See https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/14200/doc.test.profile-none_istio.io/1730032695670476800 for a failure.

PR #14248, opened 2023-12-04
Re-enable tasks/observability/logs/telemetry-api

zirain:
fix: https://github.com/istio/istio.io/issues/14210

xhref: https://github.com/istio/istio/pull/48136


zirain:
looks like a new flaky

/retest

PR #14255, opened 2023-12-04
[zh] Localize security/id-prov image

my-git9:
## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


windsonsea:
It's would be better to include all SVGs related to a single file into a PR, which may save much time.

windsonsea:
While it is great to localize images and make them more accessible to Chinese readers, engaging in intentional "spam" behavior is strongly discouraged. This kind of behavior can undermine the value of contributions made by contributors.

PR #14292, opened 2023-12-08
[release-1.20] Update to use 1.20.1 for testing

ericvn:
## Description

Update to use 1.20.1 for testing.

I notice that we moved to a later GatewayAPI commit (I suspect we needed a fix post the 1.0.0 release and 1.0.1 is not yet available) and will look into making that 1.0.0.


Makefile.core.mk
```diff
@@ -92,7 +92,7 @@ format-spelling:
 	@LC_ALL=C sort .spelling --ignore-case -o .spelling
 	@echo ".spelling file sorted."
 
-gen: tidy-go format-go update-gateway-version snips format-spelling
```
ericvn, line 95: Remove `update-gateway-version` target so the version isn't updated to a commit.

PR #14303, opened 2023-12-11
Rework header animation to use WebGL render and WGSL shaders

### LINKED ISSUE #10172: High CPU usage on istio.io homepage
Seeing at least 60% CPU usage on the Istio.io homepage on Safari, Chrome and Firefox browsers on my Mac.  No CPU spike on the other pages on the site, only the main page.  Similar .io website rarely hit double digits.

leosarra:
## Description

This represents a comprehensive rework our original header animation, there are multiple reasons that led to these extensive changes.

First of all, the previous animation relied on a deprecated Canvas 2D render in Three.JS which is no longer available starting from 2018. This led to the need of switching to the 3D WebGL render which is more performant (and actively supported by Three.JS). On the other end,  the old animation  could have not be ported to the new render without major changes.

Further complicating matters, the old animation involved the generation of 130*25 sprites, with their positions continually calculated through JavaScript. This approach, unfortunately, caused a significant spike in browser CPU usage when accessing the istio.io homepage.
In response, the revamped animation adopts a different approach. It generates a plane geometry whose vertices mirror the characteristics of the old sprites. The position, size, and color of these vertices are now dynamically determined by WGSL shaders, executing on the GPU.
This ensure the CPU impact to remain minimal (Fixes #10172) and it also reduces the amount of RAM used to render the page.

One more minor detail to note: in the old animation, the particle positions were calculated using an internal counter that was bound to overflow if the webpage remained open for an extended period. This issue caused the particle animation to eventually come to a halt. In the new animation, I have taken a different approach by computing the state based on the time interval between each frame, ensuring a smooth and robust particle movement.

Keep in mind that, the new animation has a different aesthetic when compared to the old one.
The new animation is wider and with less particles overall, I tried to emulate a 3d wave-like effect reminiscent of old animation, finding a balance between visual appeal and resource utilization.
Nonetheless I left multiple knobs in the code to tinker with the waves animation including: frequency, amplitude, number of particles, distance between particles, particles size and transparency.

Screenshot from 2023-12-11 15-19-02

Old Animation CPU/MEM Utilization:
image
New Animation CPU/MEM Utilization:
image

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/src/ts/headerAnimation.js
+++ b/src/ts/headerAnimation.js
@@ -78,9 +78,9 @@ function createRenderer(rendererProps = {}) {
     const container = document.getElementById("banner");
     const renderer = new THREE.WebGLRenderer({ ...rendererProps, alpha: true });
     renderer.domElement.id = "banner-animation";
+    renderer.setPixelRatio(window.devicePixelRatio);
     renderer.setSize(container.offsetWidth, container.offsetHeight);
     renderer.setClearColor(0xffffff, 0);
-    renderer.setPixelRatio(window.devicePixelRatio);
     renderer.domElement.id = "banner-animation";
     return renderer;
 }


leosarra:
Pinging the original reporter of the issue @GregHanson

ericvn:
Also looking for feedback from @craigbox as one of the original architects behind the overall rewrite.

kfaseela:
Thanks @leosarra

GregHanson:
Thanks for figuring this one out @leosarra !

ericvn:
LGTM

GregHanson (commented):
Screenshot 2023-12-11 at 10 53 23 AM

Not sure how big a concern it is, but I noticed in Safari the rotating particles appear bigger than what they do in Chrome.  Minor nit, but I feel it overpowers the `Simplify observability, traffic ...` banner

leosarra:
> Screenshot 2023-12-11 at 10 53 23 AM
> 
> Not sure how big a concern it is, but I noticed in Safari the rotating particles appear bigger than what they do in Chrome. Minor nit, but I feel it overpowers the `Simplify observability, traffic ...` banner

This is quite weird, from the screenshot it seems the computation of the size lacks precision on the webgl implementation of Safari.
I will take a look but I will need some time to figure out how to emulate Safari :cry:

craigbox:
LGTM on the concept.  (I don't see the same error in Safari; it looks much the same as Chrome to me.) I wouldn't spend too much extra effort here twiddling the settings ([see here](https://www.folklore.org/StoryView.py?story=Calculator_Construction_Set.txt)) because we will do an overhaul of the front page in Q1.

Perhaps apply some styling to the hero text?

I tried `text-shadow: 1px 1px 1px black;` just to get an idea, and it helps make the headline more visible.

Screenshot 2023-12-12 at 11 19 14 AM

Or, some further styling to decrease the opacity of the dots.

kfaseela:
LGTM. Thanks @leosarra

GregHanson:
@leosarra so I do see a difference between Safari technology preview (left) and base Safari (right, on Sonoma) 
Screenshot 2023-12-12 at 9 47 04 AM

But I agree - definitely not a blocker!

ericvn:
Sounds like we are good to merge this into at least preliminary, and then we can cherry-pick to istio.io.

ericvn (approved):

PR #14388, opened 2023-12-27
Fix example in Gateway with TLS termination common problems doc

### LINKED ISSUE #14386: Documentation incomplete and not accurate when giving an example of a VirtualService HTTP routing with a TLS Terminated Gateway
The [Gateway with TLS termination](https://istio.io/latest/docs/ops/common-problems/network-issues/#gateway-with-tls-termination) documentation gives the following example to solve a situation where the Gateway is doing TLS Termination and the VirtualService is doing the same using the `sniHosts` attribute for the `match` condition:

```
To fix this problem, you should switch the virtual service to specify http routing, instead of tls:

spec:
  ...
  http:
  - match: ...
```

The problem is there's no `sniHosts` attribute for the HTTP match, as the [HTTPMatchRequest](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest) documentation states. This attribute only exists for TLS, as the [TLSMatchAttributes](https://istio.io/latest/docs/reference/config/networking/virtual-service/#TLSMatchAttributes) documentation states.

If someone just change from TLS to HTTP using `sniHosts`, as the reported documentation suggests, the `Object type invalid` is thrown to the user.

So I suggested this part of documentation should be fixed with a more clear example on how to do it.

AlexonOliveiraRH:
## Description

Fixes example in common problems doc of Gateway with TLS Termination.

Closes #14386 

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [x] Networking
- [x] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/common-problems/network-issues/index.md
+++ b/content/en/docs/ops/common-problems/network-issues/index.md
@@ -432,7 +432,10 @@ To fix this problem, you should switch the virtual service to specify `http` rou
 spec:
   ...
   http:
-  - match: ...
+  - match:
+    - headers:
+        ":authority":
+          regex: "*.example.com"
 {{< /text >}}
 
 #### Gateway with TLS passthrough


kfaseela (approved):
Thanks!

PR #14409, opened 2024-01-02
Add ist0167 message description

hanxiaop:
## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/content/en/docs/reference/config/analysis/ist0167/index.md
+++ b/content/en/docs/reference/config/analysis/ist0167/index.md
@@ -6,7 +6,7 @@ test: n/a
 ---
 
 This message occurs when a policy applied in your Istio service mesh has no impact. This might be due to the
-policy's configuration not correctly targeting any workloads, namespaces in your service mesh.
+policy's configuration not correctly targeting any workloads or namespaces in your service mesh.
 
 ## Example
 
--- a/content/en/docs/reference/config/analysis/ist0167/index.md
+++ b/content/en/docs/reference/config/analysis/ist0167/index.md
@@ -10,7 +10,7 @@ policy's configuration not correctly targeting any workloads, namespaces in your
 
 ## Example
 
-You might encounter a message like this:
+You will receive a message like this:
 
 {{< text plain >}}
 Warning [IST0167] (Sidecar ns-ambient/namespace-scoped testdata/sidecar-default-selector.yaml:84) The policy has no
--- a/content/en/docs/reference/config/analysis/ist0167/index.md
+++ b/content/en/docs/reference/config/analysis/ist0167/index.md
@@ -1,0 +1,39 @@
+---
+title: IneffectivePolicy
+layout: analysis-message
+owner: istio/wg-user-experience-maintainers
+test: n/a
+---
+
+This message occurs when a policy applied in your Istio service mesh does not have any impact. This might be due to the
+policy's configuration not correctly targeting any workloads, namespaces in your service mesh.
+
+## Example
+
+You might encounter a message like this:
+
+{{< text plain >}}
+Warning [IST0167] (Sidecar ns-ambient/namespace-scoped testdata/sidecar-default-selector.yaml:84) The policy has no
+impact: namespace is in ambient mode, the policy has no impact.
+{{< /text >}}
+
+or this:
+
+{{< text plain >}}
+Warning [IST0167] (Sidecar ns-ambient/pod-scoped testdata/sidecar-default-selector.yaml:90) The policy has no impact:
+selected workload is in ambient mode, the policy has no impact.
+{{< /text >}}
+
+These messages indicate that the `Sidecar` resource in the example is targeting a workload or namespace which is in
+ambient mode, meaning that the policy specified in the `Sidecar` resource does not have any effect.
+
+## How to resolve
+
+To resolve this issue, you first need to check the reason. Currenly the policy is ineffective for the following
+reasons:
+
+1. The `Sidecar` resource is targeting a workload or namespace which is in ambient mode.
+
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/config/analysis/ist0167/index.md
```diff
@@ -0,0 +1,39 @@
...
+owner: istio/wg-user-experience-maintainers
+test: n/a
+---
+
+This message occurs when a policy applied in your Istio service mesh does not have any impact. This might be due to the
+policy's configuration not correctly targeting any workloads, namespaces in your service mesh.
+
+## Example
+
+You might encounter a message like this:
```
windsonsea, line 13:
```suggestion
You will receive a message like this:
```
For consistency with previous messages like ist0164 or ist0166

windsonsea (commented):
a text suggestion for consistency

content/en/docs/reference/config/analysis/ist0167/index.md
hanxiaop: Looks good, thanks!

content/en/docs/reference/config/analysis/ist0167/index.md
```diff
@@ -0,0 +1,39 @@
+---
+title: IneffectivePolicy
+layout: analysis-message
+owner: istio/wg-user-experience-maintainers
+test: n/a
+---
+
+This message occurs when a policy applied in your Istio service mesh does not have any impact. This might be due to the
```
ericvn, line 8:
```suggestion
This message occurs when a policy applied in your Istio service mesh has no impact. This might be due to the
```

content/en/docs/reference/config/analysis/ist0167/index.md
```diff
@@ -0,0 +1,39 @@
...
+policy's configuration not correctly targeting any workloads, namespaces in your service mesh.
```
ericvn, line 9:
```suggestion
policy's configuration not correctly targeting any workloads or namespaces in your service mesh.
```

content/en/docs/reference/config/analysis/ist0167/index.md
```diff
@@ -0,0 +1,39 @@
...
+{{< /text >}}
+
+or this:
+
+{{< text plain >}}
+Warning [IST0167] (Sidecar ns-ambient/pod-scoped testdata/sidecar-default-selector.yaml:90) The policy has no impact:
+selected workload is in ambient mode, the policy has no impact.
+{{< /text >}}
+
+These messages indicate that the `Sidecar` resource in the example is targeting a workload or namespace which is in
```
ericvn, line 27:
```suggestion
These messages indicate that the `Sidecar` resource is targeting a workload or namespace which is in
```

content/en/docs/reference/config/analysis/ist0167/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+These messages indicate that the `Sidecar` resource is targeting a workload or namespace which is in
+ambient mode, meaning that the policy specified in the `Sidecar` resource does not have any effect.
+
+## How to resolve
+
+To resolve this issue, you first need to check the reason. Currently, the policy is ineffective for the following
+reasons:
+
+1. The `Sidecar` resource is targeting a workload or namespace which is in ambient mode.
```
windsonsea, line 35: Seems we need to add more reasons. Only `1` is incomplete, normally 1-2-3.

hanxiaop: Right now, we only have one, but we may eventually add more. How should we display them in this case?

windsonsea: We may need to rewrite these 3 paras like:

```
To resolve this issue, it is necessary to investigate the cause. If the currently ineffective policy is
caused by the `Sidecar` resource targeting a workload or namespace that is in ambient mode,
the following actions can be taken:

- Ensure that the policy is defined correctly or determine if it is necessary.
- If the namespace/pod was recently added to the ambient mesh, remove
  the policy that is no longer needed.
- Update the policy to target the correct workload or namespace if necessary.
```

hanxiaop: Makes sense to me, thanks!

hanxiaop:
@istio/wg-docs-maintainers-english PTAL, thanks!

content/en/docs/reference/config/analysis/ist0167/index.md
```diff
@@ -0,0 +1,39 @@
+---
+title: IneffectivePolicy
+layout: analysis-message
+owner: istio/wg-user-experience-maintainers
+test: n/a
+---
+
+This message occurs when a policy applied in your Istio service mesh has no impact. This might be due to the
+policy's configuration not correctly targeting any workloads or namespaces in your service mesh.
```
ericvn, line 9: My first thought was to change `not correctly` to `incorrectly`.  Reading it a few more times, I'm wondering if we can remove `correctly`, so it's `not targeting`? My concern with the later is if there is something planned in the future where correctness might be checked. We use correctness at the end of the page, so going with my first thought.
```suggestion
policy's configuration incorrectly targeting any workloads or namespaces in your service mesh.
```

hanxiaop: Thanks for the suggestion. I prefer using `incorrectly` because there may be situations where users target workloads, but the method they use is incorrect. This message is intended to verify correctness.

ericvn (approved):

PR #14421, opened 2024-01-03
Remove 1.17 from CVE table, update Envoy version section

ericvn:
Remove 1.17 from CVE table. I don't think we updated it with the last security work, but it's been EOL for awhile now.

I also updated Envoy version section since we don't specifically include specific Envoy versions, but commits in the various branches.


content/en/docs/releases/supported-releases/index.md
```diff
@@ -73,19 +73,17 @@ Please keep up-to-date and use a supported version.
...
-| 1.19.x        | 1.27.x        |
-| 1.18.x        | 1.26.x        |
-| 1.17.x        | 1.25.x        |
+| Istio version | Envoy release branch |
+| ------------- | -------------------- |
+| 1.20.x        | release/v1.28        |
+| 1.19.x        | release/v1.27        |
+| 1.18.x        | release/v1.26        |
 
-In general, Istio releases tend to map one to one with Envoy releases. You can find the precise Envoy commit used by Istio in [`istio/proxy`](https://github.com/istio/proxy/blob/master/WORKSPACE#L38).
```
ericvn, line 91: We don't truly map to Envoy versions, like 1.28.1 since we move to the latest commit in the related branch on a daily basis.

dhawton (approved):

PR #14362, opened 2023-12-20
add san-validation documentation

kfaseela:

PR #14443, opened 2024-01-05
Add dependency computation for scripts in presubmit job

### LINKED ISSUE #14432: Update to snips did not really run all affected docs tests in CI
See: https://github.com/istio/istio.io/pull/14362

This PR updated a snip, however the CI did not run another test file which was using the same snip and the PR got merged, breaking the automated tests. This had to be fixed in a separate PR later then.

https://github.com/istio/istio.io/pull/14415

Would be good to analyze the framework and see why the test was skipped.

leosarra:
## Description

Fixes #14432
See https://github.com/istio/istio.io/issues/14432#issuecomment-1877044976 for details

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [x] Test and Release
- [ ] User Experience

### DIFF
--- a/scripts/pr_tests.py
+++ b/scripts/pr_tests.py
@@ -25,7 +25,7 @@ def get_script_dependencies(script_path, folder_prefix):
     with open(script_path, 'r') as script_file:
         script_content = script_file.read()
 
-        # Using regular expression to find 'source' or '.' followed by a script name
+        # Using regular expression to find 'source' followed by a script path
         pattern = r'^source\s+["\']?(.+?)["\']?$'
         matches = (match for match in re.findall(pattern, script_content, re.MULTILINE) if match.startswith(folder_prefix))
         dependencies.update(matches)
--- a/scripts/pr_tests.py
+++ b/scripts/pr_tests.py
@@ -34,12 +34,12 @@ def get_script_dependencies(script_path, folder_prefix):
     return dependencies
 
 
-def get_script_dependencies_graph(folder_prefix, file_suffix="test.sh"):
-    matching_files = [os.path.join(root, file_name) for root, _, files in os.walk(folder_prefix) for file_name in files if file_name.endswith(file_suffix)]
+def get_script_dependencies_graph(path_prefix, file_suffix="test.sh"):
+    matching_files = [os.path.join(root, file_name) for root, _, files in os.walk(path_prefix) for file_name in files if file_name.endswith(file_suffix)]
     package_dependencies = {}
 
     for file in matching_files:
-        dependencies = get_script_dependencies(file, folder_prefix)
+        dependencies = get_script_dependencies(file, path_prefix)
         for dep in dependencies:
             dep = os.path.dirname(dep)
             package_dependencies.setdefault(dep, set()).add(os.path.dirname(file))
--- a/scripts/pr_tests.py
+++ b/scripts/pr_tests.py
@@ -34,8 +34,8 @@ def get_script_dependencies(script_path, folder_prefix):
     return dependencies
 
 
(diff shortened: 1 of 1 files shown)


leosarra:
Please note the following details:
- Dependency tracking for scripts that use (for example) 'source "$(dirname "$0")/A.sh"' is not covered. This means that if a script imports another script using this method, the dependency will not be detected by the presubmit script as the dirname command is not executed. As of now that is not a problem as our scripts declare the path starting from the root of the repo.
- Dependency tracking for scripts in /contents/en/docs is currently limited to dependencies between test modules within that directory. Dependencies between test modules and scripts in other directories (such as tests/util/verify.sh) are ignored. This means that the old behavior is retained, e.g.: changes to the test-framework will still trigger all tests.

ericvn:
@frankbu Input since you did the original changes to limit which tests were run.

scripts/pr_tests.py
```diff
@@ -16,8 +16,37 @@
...
+        # Using regular expression to find 'source' followed by a script path
+        pattern = r'^source\s+["\']?(.+?)["\']?$'
+        matches = (match for match in re.findall(pattern, script_content, re.MULTILINE) if match.startswith(folder_prefix))
+        dependencies.update(matches)
+
+    return dependencies
+
+
+def get_script_dependencies_graph(folder_prefix):
+    matching_files = [os.path.join(root, file_name) for root, _, files in os.walk(folder_prefix) for file_name in files if file_name.endswith(".sh")]
```
frankbu, line 38: Only need to check *test.sh because snips.sh files never source other ones.
```suggestion
    matching_files = [os.path.join(root, file_name) for root, _, files in os.walk(folder_prefix) for file_name in files if file_name.endswith("test.sh")]
```

leosarra: Ah good to know, I see that snips.sh only imports the .sh from the boilerplate folder whose changes will anyway trigger a complete test-rerun thanks to
https://github.com/istio/istio.io/blob/568245276f54cc9da3861a3bffe1bed64e1419aa/scripts/pr_tests.py#L95-L100
Changed to filter to *test.sh only files.

frankbu:
LGTM, but are we sure this actually works? How was it tested? Maybe a followup PR (that we don't actually merge) with a dummy change to make sure it's working?

leosarra:
> LGTM, but are we sure this actually works? How was it tested? Maybe a followup PR (that we don't actually merge) with a dummy change to make sure it's working?

I tested by relying on the PR #14362 which updated a snips.sh which was in turn imported by another test that was not changed.

```
build-tools:/work$ python3 scripts/pr_tests.py 14362
tasks/traffic-management/egress,tasks/traffic-management/egress/egress-gateway-tls-origination
```

The code that is currently on master returns only the test whose snips.sh was changed:
```
build-tools:/work$ python3 scripts/pr_tests.py 14362
tasks/traffic-management/egress/egress-gateway-tls-origination
```

frankbu (approved):
Thanks @leosarra!

leosarra:
Thank you @frankbu for the review :)

ericvn:
Thanks @leosarra for the work in this.

leosarra:
Sorry I just realized that I overlooked something.
The script is returning the dependency directories one level too high.

```
build-tools:/work$ python3 scripts/pr_tests.py 14362
tasks/traffic-management/egress,tasks/traffic-management/egress/egress-gateway-tls-origination
```

It should have returned _tasks/traffic-management/egress/egress-tls-origination_ instead of _tasks/traffic-management/egress_. I will fix it right away.

PR #14486, opened 2024-01-17
Designate master and Istio 1.20 as being supported on Kubernetes 1.29

keithmattix:
## Description

Designate master and Istio 1.20 as being supported on Kubernetes 1.29 after running acceptance tests (via GKE rapid channel) and configuring test infra to run against those versions as well in istio/test-infra#5139

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/data/compatibility/supportStatus.yml
+++ b/data/compatibility/supportStatus.yml
@@ -5,13 +5,13 @@
   supported: "No, development only"
   releaseDate:
   eolDate:
-  k8sVersions: ["1.25", "1.26", "1.27", "1.28", "1.29"]
-  testedK8sVersions: ["1.23", "1.24"]
+  k8sVersions: ["1.26", "1.27", "1.28", "1.29"]
+  testedK8sVersions: ["1.23", "1.24", "1.25"]
 - version: "1.20"
   supported: "Yes"
   releaseDate: "Nov 14, 2023"
   eolDate: "~Jul 2024 (Expected)"
-  
+
   k8sVersions: ["1.25", "1.26", "1.27", "1.28", "1.29"]
   testedK8sVersions: ["1.23", "1.24"]
 - version: "1.19"
--- a/data/compatibility/supportStatus.yml
+++ b/data/compatibility/supportStatus.yml
@@ -5,13 +5,14 @@
   supported: "No, development only"
   releaseDate:
   eolDate:
-  k8sVersions: ["1.25", "1.26", "1.27", "1.28"]
+  k8sVersions: ["1.25", "1.26", "1.27", "1.28", "1.29"]
   testedK8sVersions: ["1.23", "1.24"]
 - version: "1.20"
   supported: "Yes"
   releaseDate: "Nov 14, 2023"
   eolDate: "~Jul 2024 (Expected)"
-  k8sVersions: ["1.25", "1.26", "1.27", "1.28"]
+  
+  k8sVersions: ["1.25", "1.26", "1.27", "1.28", "1.29"]
   testedK8sVersions: ["1.23", "1.24"]
 - version: "1.19"
   supported: "Yes"


SanyaKochhar (approved):

data/compatibility/supportStatus.yml
```diff
@@ -5,13 +5,14 @@
...
-  k8sVersions: ["1.25", "1.26", "1.27", "1.28"]
+  k8sVersions: ["1.25", "1.26", "1.27", "1.28", "1.29"]
   testedK8sVersions: ["1.23", "1.24"]
 - version: "1.20"
   supported: "Yes"
   releaseDate: "Nov 14, 2023"
   eolDate: "~Jul 2024 (Expected)"
-  k8sVersions: ["1.25", "1.26", "1.27", "1.28"]
+  
+  k8sVersions: ["1.25", "1.26", "1.27", "1.28", "1.29"]
```
hanxiaop, line 15: I think this doesn't seem correct to me. 1.20 has announced support for k8s 1.25 to 1.28, and this will lead to 1.20 supporting five k8s versions.

hanxiaop: See https://deploy-preview-14486--preliminary-istio.netlify.app/latest/docs/releases/supported-releases/#support-status-of-istio-releases

data/compatibility/supportStatus.yml
```diff
@@ -5,13 +5,14 @@
   supported: "No, development only"
   releaseDate:
   eolDate:
-  k8sVersions: ["1.25", "1.26", "1.27", "1.28"]
+  k8sVersions: ["1.25", "1.26", "1.27", "1.28", "1.29"]
```
hanxiaop, line 8: I think the change should be something like
```suggestion
  k8sVersions: ["1.26", "1.27", "1.28", "1.29"]
```

keithmattix: Yeah looks like 1.25 is EOL. I haven't removed it from testing yet, but can do that in a follow up.

data/compatibility/supportStatus.yml
keithmattix: Do we have a limit to number of k8s versions we support? Even though k8s 1.25 has now hit EOL, it doesn't feel like we can remove it from the table. The difference in k8s releases policy means that we'll soon have a version of k8s that doesn't have a supported Istio version. Supporting 5 versions for a bit seems preferable to that outcome

hanxiaop: We usually support four versions. If 1.29 is listed in Istio 1.20, it might lead to the misconception that Istio 1.20.0 - 1.20.2 also support Kubernetes 1.29, but they have not been tested. Therefore, I think for Istio 1.20,  we will stick to 1.25-1.28, and the upcoming Istio 1.21 is likely to support 1.26-1.29. I think @ericvn knows much about this.

keithmattix: 1.20.2 has had acceptance tests run on k8s 1.29: https://istio.slack.com/archives/C05SW44NP32/p1704400785770819

We're going to run into this every year because we have 4 releases and k8s only has 3. Istio 1.21 won't be released until 2/19 (assuming no delays) while k8s 1.29 was released 12/20/2023. The holidays help us out a bit but still we need to have an Istio version out sooner that supports 1.29 IMO

data/compatibility/supportStatus.yml
keithmattix: Eh actually, we still declare support for it, so we should probably continue to test it

ericvn: Although, we should remove 1.25 from the supported versions for master moving it to tested.

data/compatibility/supportStatus.yml
```diff
@@ -5,13 +5,14 @@
...
   testedK8sVersions: ["1.23", "1.24"]
```
ericvn, line 9: Based on my prior remark
```suggestion
  k8sVersions: ["1.26", "1.27", "1.28", "1.29"]
  testedK8sVersions: ["1.23", "1.24", "1.25"]
```

keithmattix: Ack - fixed

data/compatibility/supportStatus.yml
keithmattix: Yep makes sense to me 👍🏾

hanxiaop: Good to know this, thanks!

hanxiaop (approved):

ericvn (approved):

PR #14491, opened 2024-01-17
Initial inplace ambient upgrade guide

bleggett:
## Description

Basic ambient upgrade guide to go with the install guide. Fixes https://github.com/istio/istio/issues/48387

- I'm only documenting the in-place Helm upgrade 
  - Tested going from 1.19.6 -> 1.20.2
  - This is the minimum viable upgrade with brief traffic disruption for workload pods so it's worth documenting as a baseline even if we improve/change it - this at least should always _work_ and when in-place upgrade is completed, app pods should still be part of the mesh and still function without requiring a restart. 
  - This includes the previously-agreed caveats around using node cordons and blue-green node pools for production blast radius control - I am _not_ documenting that in detail because how it should work and be orchestrated depends on your specific cloud provider/env, and there's no clean way to simulate it with a baseline of `kind` or similar that I know of, and I don't feel good about replicating cloud provider docs.
  - I did not document revision-based installs for `istiod`/gateways because that's still basically an `in-place` upgrade for the datapath in all the ways that users care about, and whether or not you use revisions to install `istiod`/gateway, it would look the same to your apps either way (upgrading ztunnel creates traffic disruption) - using revisions to upgrade ambient doesn't buy the user much of anything right now, because of ztunnel.
  - In current `master`, upgrading Istio CNI in-place also creates app traffic disruption. The proposed [inpod ambient capture mechanism](https://github.com/istio/istio/pull/48253) will allow the Istio CNI agent to be upgraded in-place without disrupting app traffic. But, as mentioned, ztunnel is the last remaining problem in that regard - it doesn't usefully support revisions and cannot be upgraded without traffic disruption or leveraging cloud-provider node pools.

This will need to be slightly revised when `inpod` hits.

## Reviewers

- [x] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/install/helm-installation/index.md
+++ b/content/en/docs/ops/ambient/install/helm-installation/index.md
@@ -80,7 +80,7 @@ See [Installing Gateways](/docs/setup/additional-setup/gateway/) for in-depth do
 
 ## Configuration
 
-To view support configuration options and documentation, run:
+To view supported configuration options and documentation, run:
 
 {{< text syntax=bash >}}
 $ helm show values istio/istiod
--- a/content/en/docs/ops/ambient/install/helm-installation/index.md
+++ b/content/en/docs/ops/ambient/install/helm-installation/index.md
@@ -65,6 +65,19 @@ The `ztunnel` chart installs the ztunnel DaemonSet, which is the node-proxy comp
 $ helm install ztunnel istio/ztunnel -n istio-system
 {{< /text >}}
 
+### (Optional) Install an ingress gateway
+
+{{< warning >}}
+The namespace the gateway is deployed in must not have a `istio-injection=disabled` label.
+See [Controlling the injection policy](/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy) for more info.
+{{< /warning >}}
+
+{{< text syntax=bash snip_id=install_ingress >}}
+$ helm install istio-ingress istio/gateway -n istio-ingress --wait --create-namespace
+{{< /text >}}
+
+See [Installing Gateways](/docs/setup/additional-setup/gateway/) for in-depth documentation on gateway installation.
+
 ## Configuration
 
 To view support configuration options and documentation, run:
@@ -112,7 +125,7 @@ installed above.
 
 1. (Optional) Delete any Istio gateway chart installations:
 
-    {{< text syntax=bash >}}
+    {{< text syntax=bash snip_id=delete_ingress >}}
     $ helm delete istio-ingress -n istio-ingress
     $ kubectl delete namespace istio-ingress
     {{< /text >}}
@@ -152,7 +165,7 @@ installed above.
     This will delete all created Istio resources.
     {{< /warning >}}
 
-    {{< text syntax=bash >}}
+    {{< text syntax=bash snip_id=delete_crds >}}
     $ kubectl get crd -oname | grep --color=never 'istio.io' | xargs kubectl delete
     {{< /text >}}
 
--- a/content/en/docs/ops/ambient/upgrade/_index.md
+++ b/content/en/docs/ops/ambient/upgrade/_index.md
@@ -1,0 +1,7 @@
+---
+title: Upgrade Guide
+description: Upgrade guide for Istio Ambient mesh.
+weight: 10
+owner: istio/wg-environment-maintainers
+test: n/a
+---
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -78,7 +78,7 @@ Node cordoning and blue/green node pools are recommended to mitigate blast radiu
 $ helm upgrade ztunnel istio/ztunnel -n istio-system
 {{< /text >}}
 
-### Upgrading the CNI Component
+### Upgrade the CNI Component
 
 The Istio CNI agent is responsible for detecting the pods that belong to the ambient mesh, and configuring the traffic redirection between pods and the ztunnel DaemonSet. It is not part of the data plane or control plane.
 
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -92,7 +92,7 @@ Upgrading the Istio CNI agent will reconfigure networking on the node, and as su
 $ helm upgrade istio-cni istio/cni -n istio-system
 {{< /text >}}
 
-### (Optional) Upgrading the Gateway component
+### (Optional) Upgrade the Gateway component
 
 Gateway components manage east-west and north-south dataplane traffic between ambient mesh boundaries, as well as some aspects of the L7 dataplane.
 
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -1,0 +1,129 @@
+---
+title: Upgrade with Helm
+description: How to upgrade Ambient Mesh with Helm.
+weight: 5
+owner: istio/wg-environments-maintainers
+test: yes
+---
+
+Follow this guide to upgrade and configure ambient mesh using
+[Helm](https://helm.sh/docs/).  This guide assumes you have already performed an ambient mesh
+[installation with Helm](/docs/ops/ambient/install/helm-installation/) with a previous minor or patch version of Istio.
+
+{{< boilerplate ambient-alpha-warning >}}
+
+{{< warning >}}
+`Ambient` does not currently support canary upgrades of the dataplane. This means that ambient mesh dataplane upgrades will disrupt application pod traffic. Node cordoning and blue/green node pools are recommended to mitigate blast radius risk during production upgrades. See your Kubernetes provider documentation for details.
+{{< /warning >}}
+
+## Prerequisites
+
+1. Install ambient mesh with Helm by following [Install with Helm](/docs/ops/ambient/install/helm-installation/) and meeting all prerequisites in that guide.
+
+1. Update the Helm repository:
+
+    {{< text syntax=bash snip_id=update_helm >}}
+    $ helm repo update istio
+    {{< /text >}}
+
+## In-place upgrade
+
+You can perform an in place upgrade of Istio in your cluster using the Helm
+upgrade workflow.
+
+Before upgrading Istio, it is recommended to run the `istioctl x precheck` command to make sure the upgrade is compatible with your environment.
+
+{{< text syntax=bash snip_id=istioctl_precheck >}}
+$ istioctl x precheck
+✔ No issues found when checking the cluster. Istio is safe to install or upgrade!
+  To get started, check out <https://istio.io/latest/docs/setup/getting-started/>
+{{< /text >}}
+
+{{< warning >}}
+[Helm does not upgrade or delete CRDs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations) when performing an upgrade. Because of this restriction, an additional step is required when upgrading Istio with Helm.
+{{< /warning >}}
+
+### Manually upgrading the CRDs and Istio base chart
+
+1. Upgrade the Kubernetes custom resource definitions ({{< gloss >}}CRDs{{</ gloss >}}):
+
+    {{< text syntax=bash snip_id=manual_crd_upgrade >}}
+    $ kubectl apply -f manifests/charts/base/crds
+    {{< /text >}}
+
+1. Upgrade the Istio base chart:
+
+    {{< text syntax=bash snip_id=upgrade_base >}}
+    $ helm upgrade istio-base manifests/charts/base -n istio-system --skip-crds
+    {{< /text >}}
+
+### Upgrading the Istio discovery Component
+
+Istiod is the control plane component that manages and configures the proxies to route traffic within the ambient mesh.
+
+{{< text syntax=bash snip_id=upgrade_istiod >}}
+$ helm upgrade istiod istio/istiod -n istio-system
+{{< /text >}}
+
+### Upgrading the ztunnel component
+
+The ztunnel DaemonSet is the L4 node-proxy component of ambient.
+
+{{< warning >}}
+Upgrading ztunnel in-place will briefly disrupt all ambient mesh traffic on the node.
+Node cordoning and blue/green node pools are recommended to mitigate blast radius risk during production upgrades. See your Kubernetes provider documentation for details.
+{{< /warning >}}
+
+{{< text syntax=bash snip_id=upgrade_ztunnel >}}
+$ helm upgrade ztunnel istio/ztunnel -n istio-system
+{{< /text >}}
+
+### Upgrading the CNI Component
+
+The Istio CNI agent is responsible for detecting the pods that belong to the ambient mesh, and configuring the traffic redirection between pods and the ztunnel DaemonSet. It is not part of the data plane or control plane.
+
+The Istio CNI agent at version 1.x is compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the Istio CNI agent and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
+
+{{< warning >}}
+Upgrading the Istio CNI agent will reconfigure networking on the node, and as such will momentarily disrupt node traffic. To manage blast radius to application pods during Istio CNI agent upgrade, node cordons are recommended.
+{{< /warning >}}
+
+{{< text syntax=bash snip_id=upgrade_cni >}}
+$ helm upgrade istio-cni istio/cni -n istio-system
+{{< /text >}}
+
+### (Optional) Upgrading the Gateway component
+
+Gateway components manage east-west and north-south dataplane traffic between ambient mesh boundaries, as well as some aspects of the L7 dataplane.
+
+{{< text syntax=bash snip_id=upgrade_gateway >}}
+$ helm upgrade istio-ingress istio/gateway -n istio-ingress
+{{< /text >}}
+
+## Configuration
+
+To view support configuration options and documentation, run:
+
+{{< text syntax=bash snip_id=show_istiod_values >}}
+$ helm show values istio/istiod
+{{< /text >}}
+
+## Verifying the Installation
+
+### Verifying the workload status
+
+After installing all the components, you can check the Helm deployment status with:
+
+{{< text syntax=bash snip_id=show_components >}}
+$ helm list -n istio-system
+{{< /text >}}
+
+You can check the status of the deployed Istio pods with:
+
+{{< text syntax=bash snip_id=check_pods >}}
+$ kubectl get pods -n istio-system
+{{< /text >}}
+
+## Uninstall
+
+Please refer to the uninstall section in our [Helm ambient install guide](/docs/ops/ambient/install/helm-installation/#uninstall).
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -14,7 +14,9 @@ Follow this guide to upgrade and configure ambient mesh using
 {{< boilerplate ambient-alpha-warning >}}
 
 {{< warning >}}
-`Ambient` does not currently support canary upgrades of the dataplane. This means that ambient mesh dataplane upgrades will disrupt application pod traffic. Node cordoning and blue/green node pools are recommended to mitigate blast radius risk during production upgrades. See your Kubernetes provider documentation for details.
+In contrast to sidecar mode, `Ambient` supports moving application pods to an upgraded data plane without a mandatory restart or reschedule of running application pods. However, upgrading the data plane **will** briefly disrupt all workload traffic on the upgraded node, and ambient does not currently support canary upgrades of the data plane.
+
+Node cordoning and blue/green node pools are recommended to control blast radius of application pod traffic disruption during production upgrades. See your Kubernetes provider documentation for details.
 {{< /warning >}}
 
 ## Prerequisites
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -4,6 +4,7 @@ description: How to upgrade Ambient Mesh with Helm.
 weight: 5
 owner: istio/wg-environments-maintainers
 test: yes
+status: Experimental
 ---
 
 Follow this guide to upgrade and configure ambient mesh using
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -102,7 +102,7 @@ $ helm upgrade istio-ingress istio/gateway -n istio-ingress
 
 ## Configuration
 
-To view support configuration options and documentation, run:
+To view supported configuration options and documentation, run:
 
 {{< text syntax=bash snip_id=show_istiod_values >}}
 $ helm show values istio/istiod
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -57,7 +57,7 @@ $ istioctl x precheck
     $ helm upgrade istio-base manifests/charts/base -n istio-system --skip-crds
     {{< /text >}}
 
-### Upgrading the Istio discovery Component
+### Upgrade the Istio discovery Component
 
 Istiod is the control plane component that manages and configures the proxies to route traffic within the ambient mesh.
 
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -108,7 +108,7 @@ To view support configuration options and documentation, run:
 $ helm show values istio/istiod
 {{< /text >}}
 
-## Verifying the Installation
+## Verify the Installation
 
 ### Verify the workload status
 
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -110,7 +110,7 @@ $ helm show values istio/istiod
 
 ## Verifying the Installation
 
-### Verifying the workload status
+### Verify the workload status
 
 After installing all the components, you can check the Helm deployment status with:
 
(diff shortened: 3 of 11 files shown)


content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,135 @@
...
+Istiod is the control plane component that manages and configures the proxies to route traffic within the ambient mesh.
+
+{{< text syntax=bash snip_id=upgrade_istiod >}}
+$ helm upgrade istiod istio/istiod -n istio-system
+{{< /text >}}
+
+### Upgrading the ztunnel component
+
+The ztunnel DaemonSet is the L4 node-proxy component of ambient.
+{{< warning >}}
```
dhawton, line 79:
```suggestion
The ztunnel DaemonSet is the L4 node-proxy component of ambient.

{{< warning >}}
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+{{< text syntax=bash snip_id=upgrade_gateway >}}
+$ helm upgrade istio-ingress istio/gateway -n istio-ingress
+{{< /text >}}
+
+### (Optional) Upgrading the CNI Component
+
+The Istio CNI agent is responsible for detecting the pods that belong to the ambient mesh, and configuring the traffic redirection between pods and the ztunnel DaemonSet. It is not part of the data plane or control plane, and so does not need to be upgraded on the same cadence as the control plane and data plane.
+
+// TODO BML - so, on what cadence SHOULD it be upgraded, if it should not be upgraded with the rest of Istio?
```
dhawton, line 99: Might be worth referencing https://istio.io/latest/docs/setup/additional-setup/cni/#upgrade, it can be upgraded in any order provided it is within 1 minor version of the control plane. So while we don't say before/after, this could be different for ambient.

content/en/docs/ops/ambient/install/helm-installation/index.md
```diff
@@ -65,6 +65,19 @@ The `ztunnel` chart installs the ztunnel DaemonSet, which is the node-proxy comp
 $ helm install ztunnel istio/ztunnel -n istio-system
 {{< /text >}}
 
+### (Optional) Install an ingress gateway
+
+{{< text syntax=bash snip_id=install_ingress >}}
+$ helm install istio-ingress istio/gateway -n istio-ingress --wait --create-namespace
+{{< /text >}}
```
ericvn, line 72: I think it makes more sense to have the snippet after the See... and warning.

content/en/docs/ops/ambient/upgrade/_index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Upgrade Guide
+description: Upgrade guide for Istio Ambient mesh.
+weight: 2
```
ericvn, line 4: This conflicts with install which is also at 2. I suspect we want this to be after that, so maybe 3 or more. But that will conflict with architecture which is currently 3, so need to bump the architecture page to a higher number as well.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,135 @@
...
+title: Upgrade with Helm
+description: How to upgrade Ambient Mesh with Helm.
+weight: 4
+owner: istio/wg-environments-maintainers
+test: yes
+---
+
+Follow this guide to upgrade and configure ambient mesh using
+[Helm](https://helm.sh/docs/).  This guide assumes you have already performed an ambient mesh
+[installation with Helm](/docs/ops/ambient/install/helm-installation/) for a previous minor or patch version of Istio.
```
ericvn, line 11: I think we should change this (noting that you did copy this from the other page):
```suggestion
[installation with Helm](/docs/ops/ambient/install/helm-installation/) with a previous minor or patch version of Istio.
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+{{< warning >}}
```
ericvn, line 13: Replace with ambient-alpha-warning template.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,135 @@
+---
+title: Upgrade with Helm
+description: How to upgrade Ambient Mesh with Helm.
+weight: 4
+owner: istio/wg-environments-maintainers
+test: yes
```
ericvn, line 6: This says `yes` which generates the snippets, but I don't see the test.

ericvn (commented):
Quick pass:

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: I'll copy the wording around the version constraint as-is, ty!

content/en/docs/ops/ambient/install/helm-installation/index.md
bleggett: It does, and since I copied this block as-is from `content/en/docs/setup/install/helm/index.md` I'll fix it there as well.

content/en/docs/ops/ambient/upgrade/_index.md
bleggett: Right, I noticed this and then forgot to fix it.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: Found another spot that should use this too, will roll that up as well.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
dhawton: Just need to build some test.sh scripts that call the snips to test them. https://github.com/istio/istio.io/blob/master/tests/README.md

kfaseela: or you can make it `no` for the time being and I can pick up the tests . Anyways I was thinking of doing the integration tests, so docs test can help build an understanding. let me know

bleggett: I'll do the basic docs tests since it's mostly copy-paste from the existing ambient install tests, we still need proper integ tests with traffic verification.

bleggett: Done, PTAL.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,132 @@
...
+
+{{< warning >}}
+`Ambient` does not currently support canary upgrades of the dataplane. This means that ambient mesh dataplane upgrades will disrupt application pod traffic. Node cordoning and blue/green node pools are recommended to mitigate blast radius risk during production upgrades. See your Kubernetes provider documentation for details.
+{{< /warning >}}
+
+{{< boilerplate helm-preamble >}}
+
+{{< boilerplate helm-prereqs >}}
+
+## Prerequisites
```
frankbu, line 23: Two `Prerequisites` sections.
image

If you want to just append two items to the boilerplate list:

```suggestion
{{< boilerplate helm-prereqs >}}
```

You'll also need to indent the helm command in step 4 of the boilerplate or the numbering will be off/

bleggett: Even with the indent it's not numbering correctly - I'm just going to remove the Helm boilerplates as duplicative from the upgrade guide, since we already say to follow the install guide first and the boilerplates are present there already.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,132 @@
...
+    {{< /text >}}
+
+### Upgrading the Istio discovery Component
+
+Istiod is the control plane component that manages and configures the proxies to route traffic within the ambient mesh.
+
+{{< text syntax=bash snip_id=upgrade_istiod >}}
+$ helm upgrade istiod istio/istiod -n istio-system
+{{< /text >}}
+
```
linsun, line 71: Does it make sense to upgrade istio-cni here after control plane? or before control plane?  Feels it is part of control plane and we are asking users to update it after dataplane atm.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,128 @@
...
+### Upgrading the ztunnel component
+
+The ztunnel DaemonSet is the L4 node-proxy component of ambient.
+
```
linsun, line 71: should we mention node cordon here to minimize disruption?

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,128 @@
...
+### (Optional) Upgrading the Gateway component
+
+Gateway components manage east-west and north-south dataplane traffic between ambient mesh boundaries, as well as some aspects of the L7 dataplane.
+
+{{< text syntax=bash snip_id=upgrade_gateway >}}
+$ helm upgrade istio-ingress istio/gateway -n istio-ingress
+{{< /text >}}
+
+### (Optional) Upgrading the CNI Component
```
linsun, line 88: Seems this is required?

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: It does not matter (or rather it should not matter) as long as the skew is respected as per what @dhawton pointed out.

The ordering logic here is

- Required steps
- Optional steps

It is, strictly speaking, optional to upgrade the CNI, so it goes after the CP (which is not optional) for that reason. 

Would be good to have it both ways in docs tests to catch issues anyway.

linsun: Note we have some existing guideline here - https://istio.io/latest/docs/setup/additional-setup/cni/#operation-details

linsun: iirc, our recommendation is to upgrade istio-cni before or after istiod.

linsun (commented):
LG - left some comments around order of istio-cni and whether it should be optional.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: For sidecar it matters, and I don't want to change sidecar ordering reqs.

For ambient, it does not, we have no direct control plane dependency in CNI. So I don't want to convey to people that there is a preferred ordering, because there isn't. Both should work and be supported.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: Yeah I should to reinforce.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
linsun: Understood it may not matter but we may not have tested all these permutations in the past and I'm a bit cautious on recommending new orderings with unnecessary benefits to users.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: Not strictly speaking. You can upgrade the data plane and control plane and leave the CNI and everything will work just fine if you respect the skew requirements.

That's why I marked it as optional. You can do this now, later, or wait til your next upgrade, depending, as the text (hopefully) makes clear.

bleggett: Removed `(Optional)` and put CNI next to ztunnel, and after `istiod` (which is where it will need to sit with `inpod` anyway)

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: For inpod, the recommendation will be to upgrade `istio-cni` when `ztunnel` is upgraded, if it is upgraded at all. Istiod isn't a factor.

dhawton (commented):
LGTM. As @ericvn had some comments previous I'll defer final approval to him.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+{{< warning >}}
+Upgrading the Istio CNI agent will reconfigure networking on the node, and as such will momentarily disrupt node traffic. To manage blast radius to application pods during Istio CNI agent upgrade, node cordons are recommended.
+{{< /warning >}}
+
+{{< text syntax=bash snip_id=upgrade_cni >}}
+$ helm upgrade istio-cni istio/cni -n istio-system
+{{< /text >}}
+
+### (Optional) Upgrading the Gateway component
```
linsun, line 95: Should we also mention waypoint here?

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
linsun: Thanks! Just thought of it is simpler not offering a lot of options.

linsun (commented):
LGTM from env maintainer perspective.

Asked a small question about waypoint - may want to clarify it. or can wait till waypoint reaching beta.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: We do need it but it's not present in the ambient install guide yet either, and I was trying to keep those at rough parity right now.

We can do as a followup, @kfaseela might want to do that as well since the integration tests should add/include waypoints in the mix, I'd expect.

linsun:
@kfaseela or @ericvn Daniel is on vacationf or 3 days, can you check if this is ready to be merged?

kfaseela:
Can we have an approval from any of the ambient maintainers?

linsun (approved):

linsun:
> Can we have an approval from any of the ambient maintainers?

Me... sorry thought I approved it already.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+{{< /text >}}
+
+### Upgrading the CNI Component
+
+The Istio CNI agent is responsible for detecting the pods that belong to the ambient mesh, and configuring the traffic redirection between pods and the ztunnel DaemonSet. It is not part of the data plane or control plane.
+
+The Istio CNI agent at version 1.x is compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the Istio CNI agent and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
+
+{{< warning >}}
+Upgrading the Istio CNI agent will reconfigure networking on the node, and as such will momentarily disrupt node traffic. To manage blast radius to application pods during Istio CNI agent upgrade, node cordons are recommended.
```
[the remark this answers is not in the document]
linsun: Correct if you want to apply node cordons, you'll need at least 1 nodes, and possibly add extra nodes during upgrade.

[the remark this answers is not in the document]
bleggett: Yep. This is all highly dependent on your K8S setup, and it's a bit crude, but it's what we've settled on for the time being.

One of the goals in writing it down in black and white was to clarify what the actual current process for users is, so it's easier for people to reason about future improvements.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+Gateway components manage east-west and north-south dataplane traffic between ambient mesh boundaries, as well as some aspects of the L7 dataplane.
+
+{{< text syntax=bash snip_id=upgrade_gateway >}}
+$ helm upgrade istio-ingress istio/gateway -n istio-ingress
+{{< /text >}}
+
+## Configuration
+
+To view support configuration options and documentation, run:
```
kfaseela, line 105: Did you mean, "supported"configuration?

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+{{< text syntax=bash snip_id=show_istiod_values >}}
+$ helm show values istio/istiod
+{{< /text >}}
+
+## Verifying the Installation
```
kfaseela, line 111:
```suggestion
## Verify the Installation
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Verifying the workload status
```
kfaseela, line 113:
```suggestion
### Verify the workload status
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+$ istioctl x precheck
+✔ No issues found when checking the cluster. Istio is safe to install or upgrade!
+  To get started, check out <https://istio.io/latest/docs/setup/getting-started/>
+{{< /text >}}
+
+{{< warning >}}
+[Helm does not upgrade or delete CRDs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations) when performing an upgrade. Because of this restriction, an additional step is required when upgrading Istio with Helm.
+{{< /warning >}}
+
+### Manually upgrading the CRDs and Istio base chart
```
kfaseela, line 46:
```suggestion
### Manually upgrade the CRDs and Istio base chart
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+    $ kubectl apply -f manifests/charts/base/crds
+    {{< /text >}}
+
+1. Upgrade the Istio base chart:
+
+    {{< text syntax=bash snip_id=upgrade_base >}}
+    $ helm upgrade istio-base manifests/charts/base -n istio-system --skip-crds
+    {{< /text >}}
+
+### Upgrading the Istio discovery Component
```
kfaseela, line 60:
```suggestion
### Upgrade the Istio discovery Component
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+Istiod is the control plane component that manages and configures the proxies to route traffic within the ambient mesh.
+
+{{< text syntax=bash snip_id=upgrade_istiod >}}
+$ helm upgrade istiod istio/istiod -n istio-system
+{{< /text >}}
+
+### Upgrading the ztunnel component
```
kfaseela, line 68:
```suggestion
### Upgrade the ztunnel component
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+{{< warning >}}
+Upgrading ztunnel in-place will briefly disrupt all ambient mesh traffic on the node.
+Node cordoning and blue/green node pools are recommended to mitigate blast radius risk during production upgrades. See your Kubernetes provider documentation for details.
+{{< /warning >}}
+
+{{< text syntax=bash snip_id=upgrade_ztunnel >}}
+$ helm upgrade ztunnel istio/ztunnel -n istio-system
+{{< /text >}}
+
+### Upgrading the CNI Component
```
kfaseela, line 81:
```suggestion
### Upgrade the CNI Component
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
kfaseela, line 95:
```suggestion
### (Optional) Upgrade the Gateway component
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
+---
+title: Upgrade with Helm
+description: How to upgrade Ambient Mesh with Helm.
+weight: 5
+owner: istio/wg-environments-maintainers
+test: yes
```
howardjohn, line 6:
```suggestion
test: yes
status: Experimental
```

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+---
+
+Follow this guide to upgrade and configure ambient mesh using
+[Helm](https://helm.sh/docs/).  This guide assumes you have already performed an ambient mesh
+[installation with Helm](/docs/ops/ambient/install/helm-installation/) with a previous minor or patch version of Istio.
+
+{{< boilerplate ambient-alpha-warning >}}
+
+{{< warning >}}
+`Ambient` does not currently support canary upgrades of the dataplane. This means that ambient mesh dataplane upgrades will disrupt application pod traffic. Node cordoning and blue/green node pools are recommended to mitigate blast radius risk during production upgrades. See your Kubernetes provider documentation for details.
```
howardjohn, line 16: maybe nitpick, but fwiw in sidecar Istio "in place" has no disruption or downtime either. So we may want to make it a bit more clear this is not just losing the benefits of canary, but is also disruptive?

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+1. Upgrade the Istio base chart:
+
+    {{< text syntax=bash snip_id=upgrade_base >}}
+    $ helm upgrade istio-base manifests/charts/base -n istio-system --skip-crds
+    {{< /text >}}
+
+### Upgrade the Istio discovery Component
+
+Istiod is the control plane component that manages and configures the proxies to route traffic within the ambient mesh.
```
howardjohn, line 62: We do have the ability to canary with control plane ztunnel uses, no? Do we want to encourage that/document it?

Its really like https://istio.io/latest/docs/setup/additional-setup/gateway/#in-place-upgrade -- its "in place" from data plane, but not from control plane

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+{{< text syntax=bash snip_id=upgrade_ztunnel >}}
+$ helm upgrade ztunnel istio/ztunnel -n istio-system
+{{< /text >}}
+
+### Upgrade the CNI Component
+
+The Istio CNI agent is responsible for detecting the pods that belong to the ambient mesh, and configuring the traffic redirection between pods and the ztunnel DaemonSet. It is not part of the data plane or control plane.
+
+The Istio CNI agent at version 1.x is compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the Istio CNI agent and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
```
howardjohn, line 85: This doesn't seem true with ambient anymore, at least in its current stability? Especially with inpod, etc

howardjohn: inpod specifically since it was just released, not fundamental long term issue

howardjohn (commented):
Mostly lgtm, just a few small comments. OK to merge

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: Yes, and it's wrong in the install doc too, ty!

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: Yep good point - in sidecar that's because we literally don't upgrade anything and let the sidecar + CP fall into version skew for $period_of_time so it's a bit hidden.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: If we do, I would prefer to add that later on top of this as an alt flow.

Moving the ztunnel to a different CP endpoint requires a manual update of the ztunnel manifest IIRC, and cannot be accomplished thru Helm without rescheduling or weird workarounds where you "upgrade" the old version of the chart with the old version of the chart and hope the resulting manifest diff doesn't cause ztunnel pods to be recycled.

Also, you have to move the node's ztunnel to the new CP revision, which moves the entire node to the new revision - we currently do not have a way in Helm or otherwise to partially upgrade the ztunnel on some nodes, so calling that "canary" is a bit confusing.

We could probably fix that.

Additionally, the ztunnel update itself is still purely inplace and disruptive, and the baseline guide should/must explain how to upgrade everything, at a minimum.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: Correct, this will need to be revised a bit when inpod is fully merged (e.g. https://github.com/istio/istio/pull/48253 - given that is still pending, `master` still is not using `inpod`)

We can hold this PR open until that merges and do it in one go, or we can merge this now and do that later, and I'll followup - no real pref. on ordering.

bleggett: But yeah long term with inpod the only real hard compatibility requirement is between ztunnel and the CNI agent, not between the CNI agent and the CP.

How we surface that to people if we want to tolerate skew there is TDB.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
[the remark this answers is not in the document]
zirain: can you point out the detail in the doc? At latest we should highlight it.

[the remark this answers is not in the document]
bleggett: I mention it at the top of the page here: https://github.com/istio/istio.io/pull/14491/files#diff-1e0e52b46d7e94ba4f52d12e3a86d71d14b68b0ade7c78618ef54d81f23d6d18R19

if that's what you're asking.

As I mentioned in the PR description, I don't want to get into the weeds of how this works operationally in this doc, because it is different for every cloud provider, and not easily demo-able locally with `kind` or similar.

That is an inherent downside of this approach, it's implicitly a punt and very cloud-provider dependent.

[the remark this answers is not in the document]
zirain: ok, it's ok for now.

kfaseela (approved):

PR #14516, opened 2024-01-22
Add wait flag to istioctl waypoint apply cmd

leosarra:
## Description

The https://github.com/istio/istio/pull/48769 commit added a --wait flag to `istioctl x waypoint apply` cmd.
With that flag set, istioctl will wait until the gateway has the "Programmed" status set to `true`.

This commit updates the documentation to use that flag and removes the logic in test.sh that was checking for the "Programmed" status.

## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/getting-started/test.sh
+++ b/content/en/docs/ops/ambient/getting-started/test.sh
@@ -75,6 +75,7 @@ _verify_contains snip_l4_authorization_policy_3 "$snip_l4_authorization_policy_3
 _verify_failure snip_l4_authorization_policy_4
 
 _verify_contains snip_l7_authorization_policy_1 "$snip_l7_authorization_policy_1_out"
+_verify_contains snip_l7_authorization_policy_2 "Resource programmed, assigned to service"
 snip_l7_authorization_policy_3
 _verify_contains snip_l7_authorization_policy_4 "$snip_l7_authorization_policy_4_out"
 _verify_contains snip_l7_authorization_policy_5 "$snip_l7_authorization_policy_5_out"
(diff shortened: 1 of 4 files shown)


leosarra:
This needs #14508 in order to pass the tests

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -75,7 +75,6 @@ _verify_contains snip_l4_authorization_policy_3 "$snip_l4_authorization_policy_3
 _verify_failure snip_l4_authorization_policy_4
 
 _verify_contains snip_l7_authorization_policy_1 "$snip_l7_authorization_policy_1_out"
-_verify_contains snip_l7_authorization_policy_2 "Resource programmed, assigned to service"
```
hanxiaop, line 78: Why this is being removed?

hanxiaop (approved):
Nice!

content/en/docs/ops/ambient/getting-started/test.sh
leosarra: I removed it because that step was checking that the bookinfo gateway was programmed.
It felt it was no longer needed considering that istioctl will now perform that check instead.

In retrospect, there is still usefulness in preserving it, even if it is only to confirm that the command we describe in the md functions properly, no matter how redundant that step may seem.

The alternative would be to remove the following text from the markdown itself, but I feel it gives useful context to the reader on what is going on.

```
resource with `Programmed` status:

{{< text bash >}}
$ kubectl get gtw bookinfo-productpage -o yaml
...
```

I will made a commit to restore it

hanxiaop: Had no preference but was curious. Thanks for the information!

leosarra:
Rebased to include the recent commit that updated istio.io/istio dependency, otherwise I think the test would have continued failing

kfaseela (approved):

PR #14520, opened 2024-01-23
Fix revision test

hanxiaop:
## Description

This is try to fix https://github.com/istio/istio.io/pull/14508

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


hanxiaop:
/cc @ericvn 

The webhook generated by `istioctl tag set` needs to be cleaned up separately instead of `istioctl uninstall`

content/en/docs/setup/upgrade/canary/revision_tags_test.sh
```diff
@@ -56,5 +56,7 @@ _verify_contains snip_usage_8 "istiod-$fullVersionRevision"
 
 # @cleanup
 snip_uninstall_old_control_plane_1
+kubectl delete mutatingwebhookconfiguration istio-revision-tag-prod-canary
```
kfaseela, line 59: why is this not part of uninstall? is it a new change coming in? If so can you point to the istio PR which changes this?

hanxiaop: It's not a new change: a tag webhook should be created with `istioctl tag set` and removed with `istioctl tag remove`, since these are user-created items not generated by the installer. `istioctl uninstall` should not remove these, to avoid issues like https://github.com/istio/istio/issues/47423.

kfaseela: was wondering why the test started failing all of a sudden now

kfaseela:
Ok, so looks like this change came in as part of https://github.com/istio/istio/pull/48780/files ?

hanxiaop:
> Ok, so looks like this change came in as part of https://github.com/istio/istio/pull/48780/files ?

Yes it is. However I don't know why initially the test was passed, as `istioctl uninstall` shouldn't cleanup the user created webhook.

kfaseela:
> > Ok, so looks like this change came in as part of https://github.com/istio/istio/pull/48780/files ?
> 
> Yes it is. However I don't know why initially the test was passed before the regression was introduced, as `istioctl uninstall` shouldn't cleanup the user created webhook.

Do we have to change the documentation as well to add this in the clean-up? instead of adding this only in the test? https://istio.io/latest/docs/setup/upgrade/canary/#cleanup

hanxiaop:
> > > Ok, so looks like this change came in as part of https://github.com/istio/istio/pull/48780/files ?
> > 
> > 
> > Yes it is. However I don't know why initially the test was passed before the regression was introduced, as `istioctl uninstall` shouldn't cleanup the user created webhook.
> 
> Do we have to change the documentation as well to add this in the clean-up? instead of adding this only in the test? https://istio.io/latest/docs/setup/upgrade/canary/#cleanup

@kfaseela Makes sense to me, I'll add some contents to the cleanup.

content/en/docs/setup/upgrade/canary/index.md
```diff
@@ -261,6 +261,13 @@ with the canary uninstall.
 
 ## Cleanup
 
+1. Clean up created revisioned tags:
+
+   {{< text bash >}}
+   $ istioctl tag remove prod-stable
+   $ istioctl tag remove prod-canary
+   {{< /text >}}
```
wilsonwu:
```suggestion
    {{< text bash >}}
    $ istioctl tag remove prod-stable
    $ istioctl tag remove prod-canary
    {{< /text >}}
```

wilsonwu (commented):
Suggest to keep same indent with other content.

content/en/docs/setup/upgrade/canary/index.md
hanxiaop: Thanks, the linter told me the same, and I think this has been fixed.

hanxiaop:
From the test, it looks like `istioctl` is not working at all (`istioctl uninstall --purge` and `istioctl tag remove`):
```
    snapshot_validator.go:105: timeout while waiting after 98 attempts (last error: snapshots are different: 
        @@ -34,6 +34,15 @@
                 }
               ],
        -      "mutatingWebhookConfigurations": null,
        +      "mutatingWebhookConfigurations": [
        +        "istio-revision-tag-default",
        +        "istio-revision-tag-prod-canary",
        +        "istio-revision-tag-prod-stable",
        +        "istio-sidecar-injector-1-20-1",
        +        "istio-sidecar-injector-1-21-0"
        +      ],
               "validatingWebhookConfigurations": [
        +        "istio-validator-1-20-1-istio-system",
        +        "istio-validator-1-21-0-istio-system",
        +        "istiod-default-validator",
                 "metallb-webhook-configuration"
               ]
        )
```
will have a try

hanxiaop:
@ericvn @kfaseela I think I've passed the test. However, I'm not sure why `istioctl tag remove`  and `istioctl uninstall --prune` fails to remove the webhook in the test environment, even though it works perfectly in my actual environment. Therefore, I added a logic to manually deleting the webhook, using a method similar to the other test:
https://github.com/istio/istio.io/blob/69cdf19243b5d02dcf64a1167dd5fa09deca301f/content/en/docs/setup/upgrade/helm/revlabels_test.sh#L32-L34

content/en/docs/setup/upgrade/canary/canary_upgrade_test.sh
```diff
@@ -55,4 +55,4 @@ _verify_like snip_uninstall_old_control_plane_3 "$snip_uninstall_old_control_pla
 
 # @cleanup
 snip_uninstall_canary_control_plane_1
-snip_cleanup_1
\ No newline at end of file
+snip_cleanup_2
```
kfaseela, line 58: why removed 1? I think we should have all 3 snip cleanups?

ericvn:
Note that the istio.io tests in the master branch are getting updated to use the commits from the release-1.21 branch of istio/istio (PR coming soon). I expect it to initially fail the test as well (as this one is). If there are any changes in istio/istio needed, we will need to make sure they are CP'd the the release-1.21 branch as well.

content/en/docs/setup/upgrade/canary/canary_upgrade_test.sh
hanxiaop: This is not removed, cleanup_2 is the old cleanup_1, as I've added a new cleanup_1 which is doing `tag remove`. In this test, there's nothing related to the tag, so I didn't add the tag cleanup.

hanxiaop:
> Note that the istio.io tests in the master branch are getting updated to use the commits from the release-1.21 branch of istio/istio (PR coming soon). I expect it to initially fail the test as well (as this one is). If there are any changes in istio/istio needed, we will need to make sure they are CP'd the the release-1.21 branch as well.

Makes sense, I think it is expected to have the same error and will be fixed here.

content/en/docs/setup/upgrade/canary/canary_upgrade_test.sh
dhawton: I see 1 and 3 being called in revision_tags_test so all 3 are being called at some point.

dhawton (approved):
lgtm

kfaseela (approved):

ericvn:

flake

PR #14500, opened 2024-01-19
sync new format of labels and annotations

zirain:
manual sync labels and annotations page

part of https://github.com/istio/istio.io/pull/14476

cc @ericvn

PR #14525, opened 2024-01-24
[zh] sync config pages

zirain:
follow up: https://github.com/istio/istio.io/pull/14500#issuecomment-1900521456

cc @ericvn


windsonsea (approved):
Thanks.

PR #14539, opened 2024-01-25
Add KubeCon EU Istio sessions to the blog

kfaseela:


linsun (commented):
Thanks!!  If you search istio, i think we have a few more talks.  Worth adding them?

kfaseela:
> Thanks!! If you search istio, i think we have a few more talks. Worth adding them?

will take a look

dhawton:
> > Thanks!! If you search istio, i think we have a few more talks. Worth adding them?
> 
> will take a look

I'll hold off reviewing until you can look

kfaseela:
@linsun I have added the missing sessions, you can take a look now.

dhawton (approved):
LGTM. Threw a DNM as you asked @linsun for a review. Feel free to remove when ready to publish.

linsun (approved):
Thanks Faseela

PR #14542, opened 2024-01-26
Fix in-place upgrade doc to be more accurate

hanxiaop:
## Description

The upgrade is an alias of install command, with no additional check.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/upgrade/in-place/index.md
+++ b/content/en/docs/setup/upgrade/in-place/index.md
@@ -7,10 +7,7 @@ owner: istio/wg-environments-maintainers
 test: no
 ---
 
-The `istioctl upgrade` command performs an upgrade of Istio. Before performing
-the upgrade, it checks that the Istio installation meets the upgrade eligibility
-criteria. Also, it alerts the user if it detects any changes in the profile
-default values between Istio versions.
+The `istioctl upgrade` command performs an upgrade of Istio.
 
 {{< tip >}}
 [Canary Upgrade](/docs/setup/upgrade/canary/) is safer than doing an in-place upgrade and is the recommended upgrade method.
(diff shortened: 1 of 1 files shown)


linsun:
Does istioctl upgrade still work? While I don't object to use istioctl install for upgrade, I do think istioctl upgrade is more clear to users and keep the doc unchanged can reduce questions from users.

hanxiaop:
> Does istioctl upgrade still work? While I don't object to use istioctl install for upgrade, I do think istioctl upgrade is more clear to users and keep the doc unchanged can reduce questions from users.

It works but it's just an alias to the install command, with no additional check. Since you have the preference to it, I can just revise the upgrade description to be more accurate instead of chaning the command.

dhawton (approved):
Agree that upgrade might be more clear.. LGTM

rvennam:
If `upgrade` is just an alias and doesn't do anything different than `install`, it feels misleading to use it. Can't we just make it `install`?

dhawton:
> If `upgrade` is just an alias and doesn't do anything different than `install`, it feels misleading to use it. Can't we just make it `install`?

I think upgrade is more clear, tbh.

hanxiaop:
> If `upgrade` is just an alias and doesn't do anything different than `install`, it feels misleading to use it. Can't we just make it `install`?

I actually don't know if it has been deprecated or not. What I do know is that `upgrade` was different from `install`, and it may have been changed to be an alias of install due to bugs. After that, there should have been no changes for at least two years. When I read the doc, I thought it's quite not accurate now.

dhawton:
It is basically an alias: https://github.com/istio/istio/blob/master/operator/cmd/mesh/upgrade.go#L30...

though https://github.com/istio/istio/blob/master/operator/cmd/mesh/upgrade.go#L39C6-L39C54 says it does perform checks.

hanxiaop:
> It is basically an alias: https://github.com/istio/istio/blob/master/operator/cmd/mesh/upgrade.go#L30...
> 
> though https://github.com/istio/istio/blob/master/operator/cmd/mesh/upgrade.go#L39C6-L39C54 says it does perform checks. I don't see any checks being performed, so this could be needing changed as well.

Yes, they are exactly same commands for years. The description is not accurate as well.

hanxiaop:
I'd like to merge this one since the old text is not accurate and misleading, and the command description has been corrected https://github.com/istio/istio/pull/49012. We can discuss further if there's a need to revise 'upgrade' to 'install'.

PR #14559, opened 2024-01-29
add additional breakpoints for blog content width

### LINKED ISSUE #14556: istio.io blogs real estate issue
When reviewing our blogs on istio.io, i noticed that we have too many spaces on the left and right side.  Seems wasting a lot of real estate. This makes blog a little hard to read. See screenshot:

image

dhawton:
## Description

Adds additional breakpoints to increase max-width for larger screens. They are much more common now than when originally written, which leaves a lot of whitespace to the left and right of blog posts.

Closes #14556 

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -129,7 +129,7 @@ lint-md: clean_public build_nominify
 	@SKIP_LINK_CHECK=true scripts/lint_site.sh en
 
 serve: site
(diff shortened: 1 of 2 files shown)


dhawton:
before
after

Values based on bootstrap 4 definitions, which is where the original was taken.

rvennam (approved):

PR #14562, opened 2024-01-29
Slightly tweak upgrade guide in the wake of inpod merge

bleggett:
Note the (potentially) controversial addition of a ztunnel skew requirement.

- With inpod, the explicit upgrade skew concern is now between CNI agent and ztunnel, not between CNI agent and control plane. So, duplicated the CNI agent skew requirement for ztunnel, so that they are kept together. This is not perfect but it will do.

### DIFF
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -72,6 +72,8 @@ $ helm upgrade istiod istio/istiod -n istio-system
 
 The ztunnel DaemonSet is the L4 node-proxy component of ambient.
 
+The ztunnel at version 1.x is compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the ztunnel DaemonSet and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
+
 {{< warning >}}
 Upgrading ztunnel in-place will briefly disrupt all ambient mesh traffic on the node.
 Node cordoning and blue/green node pools are recommended to mitigate blast radius risk during production upgrades. See your Kubernetes provider documentation for details.
@@ -83,12 +85,12 @@ $ helm upgrade ztunnel istio/ztunnel -n istio-system
 
 ### Upgrade the CNI Component
 
-The Istio CNI agent is responsible for detecting the pods that belong to the ambient mesh, and configuring the traffic redirection between pods and the ztunnel DaemonSet. It is not part of the data plane or control plane.
+The Istio CNI agent is responsible for detecting the pods that belong in the ambient mesh, informing ztunnel that proxy ports should be established within added pods, and configuring traffic redirection within the pod network namespace. It is not part of the data plane or control plane.
 
 The Istio CNI agent at version 1.x is compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the Istio CNI agent and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
 
 {{< warning >}}
-Upgrading the Istio CNI agent will reconfigure networking on the node, and as such will momentarily disrupt node traffic. To manage blast radius to application pods during Istio CNI agent upgrade, node cordons are recommended.
+Upgrading the Istio CNI agent to a compatible version in-place will not disrupt networking for running pods already successfully added to ambient mesh, but new pods will fail to schedule on the node until the upgrade is complete and the upgraded Istio CNI agent on the node passes readiness checks. If this is a significant disruption concern, or stricter blast radius controls are desired for CNI upgrades, node taints and/or node cordons are recommended.
 {{< /warning >}}
 
 {{< text syntax=bash snip_id=upgrade_cni >}}
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -73,7 +73,7 @@ $ helm upgrade istiod istio/istiod -n istio-system
 The ztunnel DaemonSet is the L4 node-proxy component of ambient.
 
 {{< warning >}}
-As ambient is not yet stable, the following statement is not a compatibility guarantee and is subject to change, or removal.
+As ambient is not yet stable, the following statement is not a compatibility guarantee and is subject to change, or removal. Prior to reaching stable status, this component and/or the control plane may receive breaking changes that prevent compatibility between minor versions.
 {{< /warning >}}
 
 The ztunnel at version 1.x is generally compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the ztunnel DaemonSet and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
@@ -92,13 +92,13 @@ $ helm upgrade ztunnel istio/ztunnel -n istio-system
 The Istio CNI agent is responsible for detecting pods added to the ambient mesh, informing ztunnel that proxy ports should be established within added pods, and configuring traffic redirection within the pod network namespace. It is not part of the data plane or control plane.
 
 {{< warning >}}
-As ambient is not yet stable, the following statement is not a compatibility guarantee and is subject to change, or removal.
+As ambient is not yet stable, the following statement is not a compatibility guarantee and is subject to change, or removal. Prior to reaching stable status, this component and/or the control plane may receive breaking changes that prevent compatibility between minor versions.
 {{< /warning >}}
 
 The Istio CNI agent at version 1.x is generally compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the Istio CNI agent and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
 
 {{< warning >}}
-Upgrading the Istio CNI agent to a compatible version in-place will not disrupt networking for running pods already successfully added to ambient mesh, but new pods will fail to schedule on the node until the upgrade is complete and the upgraded Istio CNI agent on the node passes readiness checks. If this is a significant disruption concern, or stricter blast radius controls are desired for CNI upgrades, node taints and/or node cordons are recommended.
+Upgrading the Istio CNI agent to a compatible version in-place will not disrupt networking for running pods already successfully added to ambient mesh, but no ambient-captured pods will be successfully scheduled (or rescheduled) on the node until the upgrade is complete and the upgraded Istio CNI agent on the node passes readiness checks. If this is a significant disruption concern, or stricter blast radius controls are desired for CNI upgrades, node taints and/or node cordons are recommended.
 {{< /warning >}}
 
 {{< text syntax=bash snip_id=upgrade_cni >}}
--- a/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
+++ b/content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
@@ -72,7 +72,11 @@ $ helm upgrade istiod istio/istiod -n istio-system
 
 The ztunnel DaemonSet is the L4 node-proxy component of ambient.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -72,6 +72,8 @@ $ helm upgrade istiod istio/istiod -n istio-system
 
 The ztunnel DaemonSet is the L4 node-proxy component of ambient.
 
+The ztunnel at version 1.x is compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the ztunnel DaemonSet and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
```
bleggett, line 75: This might be controversial, but we need to say _something_, and with inpod, ztunnel and CNI need to be (largely, for now) in lockstep with each other in terms of upgrade skew range.

dhawton: Overall PR looks fine to me.  This part may need some discussion, though. With inpod stuff, can we really say ztunnel is compatible with N-1.. IE, would 1.20's ztunnel function as expected with a 1.21 control plane or vice versa?

cc: @howardjohn what do you think here?

bleggett: I can be more explicit and say that ztunnel and CNI must be upgraded together, and should as a unit respect the version skew, rather than saying that each one individually does.

ilrudie: 🤔 probably (at least more) OK for alpha feat but definitely something to consider more thoroughly for Beta. Seems like a safer way forward if we can tie the components together in lock step but we'll want to consider failure modes and how they get bubbled up especially since failing to maintain this lock step might be the difference between "the ambient mesh is working" and "the ambient mesh is failing (hopefully at least very loudly)"

howardjohn: I am a bit worried about making any compatibility claims. Part of being "alpha" means we can make breaking changes. In particular, this isn't even true right now, as N-1 doesn't work (N-1 had no 'inpod' at all :slightly_smiling_face: ), and I am worried about committing to N+1 support since we have no feedback on inpod yet.

Maybe we can say we strive to support this range, but given its feature status this is best effort? Later, we can make a more concrete commitment
I would say "no skew allowed" but that is hard to do

linsun: Not sure why we need to support N-1 and add more test matrix for us vs just telling users to update Istiod (control plane) first, then data plane (ztunnel and waypoints).

linsun: I'd in favor of dropping N-1, note, we'll likely be asked to support N, N+1 and N+2 CPs because istio supports skip version update.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -83,12 +85,12 @@ $ helm upgrade ztunnel istio/ztunnel -n istio-system
 
 ### Upgrade the CNI Component
 
-The Istio CNI agent is responsible for detecting the pods that belong to the ambient mesh, and configuring the traffic redirection between pods and the ztunnel DaemonSet. It is not part of the data plane or control plane.
+The Istio CNI agent is responsible for detecting the pods that belong in the ambient mesh, informing ztunnel that proxy ports should be established within added pods, and configuring traffic redirection within the pod network namespace. It is not part of the data plane or control plane.
```
linsun, line 88:
```suggestion
The Istio CNI agent is responsible for detecting the pods are added to the ambient mesh, informing ztunnel that proxy ports should be established within added pods, and configuring traffic redirection within the pod network namespace. It is not part of the data plane or control plane.
```

linsun: use add to ambient mesh for consistency.

Also, i am not sure users need to know proxy ports other than in a troubleshooting docs and detailed "how it works" docs.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -83,12 +85,12 @@ $ helm upgrade ztunnel istio/ztunnel -n istio-system
...
 
 The Istio CNI agent at version 1.x is compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the Istio CNI agent and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
 
 {{< warning >}}
-Upgrading the Istio CNI agent will reconfigure networking on the node, and as such will momentarily disrupt node traffic. To manage blast radius to application pods during Istio CNI agent upgrade, node cordons are recommended.
+Upgrading the Istio CNI agent to a compatible version in-place will not disrupt networking for running pods already successfully added to ambient mesh, but new pods will fail to schedule on the node until the upgrade is complete and the upgraded Istio CNI agent on the node passes readiness checks. If this is a significant disruption concern, or stricter blast radius controls are desired for CNI upgrades, node taints and/or node cordons are recommended.
```
linsun, line 93: +1 on this important clarification

linsun: may be worth mention this will impact recovered/restarted pods as well, i think?

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -72,6 +72,12 @@ $ helm upgrade istiod istio/istiod -n istio-system
 
 The ztunnel DaemonSet is the L4 node-proxy component of ambient.
 
+{{< warning >}}
+As ambient is not yet stable, the following statement is not a compatibility guarantee and is subject to change, or removal.
```
dhawton, line 76:
```suggestion
As ambient is not yet stable, the following statement is not a compatibility guarantee and is subject to change, or removal. During alpha stage, ztunnel and/or the control plane may receive breaking changes that prevent compatibility between minor versions.
```

I'd probably add a note like this to specifically spell out the breaking changes.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
dhawton: Echo Lin's comment about recovered/restarted pods prior to CNI passing readiness checks.. but maybe expand that once the readiness checks are complete, pods will be added/removed from ambient mesh as appropriate.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: Yeah I wanted to not mention `alpha` so that we wouldn't have to update this part when switching to beta, since, strictly speaking, we are able to change this during beta as well.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: I'll think about how to word this better. "New pod" by definition covers any new pod on the node, whether it comes from a new resource being deployed, or a restart of pods belonging to an existing resource - it's still scheduling a new pod on the node.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
dhawton: True. I don't think it's a huge ask to remove this during beta, but I also see the ambient-alpha-warning boilerplate so it may not be necessary now that I think more on it as we're already stating that at the top of the document.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: I took a stab at it, LMK if it's worse or better.

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
bleggett: Anyway I used something like this that I think captures the gist, LMK

content/en/docs/ops/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -72,6 +72,12 @@ $ helm upgrade istiod istio/istiod -n istio-system
 
 The ztunnel DaemonSet is the L4 node-proxy component of ambient.
 
+{{< warning >}}
+As ambient is not yet stable, the following statement is not a compatibility guarantee and is subject to change, or removal. Prior to reaching stable status, this component and/or the control plane may receive breaking changes that prevent compatibility between minor versions.
+{{< /warning >}}
+
+The ztunnel at version 1.x is generally compatible with control plane at version 1.x-1, 1.x, and 1.x+1, which means the ztunnel DaemonSet and the Istio control plane can be upgraded independently and in any order, as long as their version difference is within one minor version.
```
linsun, line 79: I'm still concerned about `can be upgraded independently and in any order` as this is not something we test in the past.  I'd prefer to only have the following, unless we can test 1.x-1 and have automated tests:

The ztunnel at version 1.x is generally compatible with control plane at version 1.x, and 1.x+1

dhawton: It is something that is generally followed with other components, IE CNI. I think until after it's in beta, it is sufficient as is and can be discussed maybe during the Ambient meeting what compatibility requirements we need on ztunnel/CNI.

linsun: I 'd think istio-cni is different comparing with ztunnel as ztunnel is data plane.  we always recommend users to update CP first then DP, I'd rather avoid sending confusion messages. 

I'm tagging more folks to comment @istio/wg-environments-maintainers.

bleggett: The goal with WDS/ztunnel and ambient in general is to make the dataplane and CP independent. Part of that independence means they do not have an upgrade ordering dependency.

If our tests assume a specific order, they probably shouldn't - but I don't think we have any ambient upgrade tests, so it's likely moot.

When we do have them, they should test either order, for robustness.

linsun: The challenge is we do change WDS API sometimes, may be not a whole lot, but if we do change it from n to n+1, how would we ensure that n still work with n+1's ztunnel?

howardjohn: +1, prefer not to have n+1 support, at least until we are more stabilized

linsun: yeah and if users request for it.

bleggett: KK, will update to N, N-1 for now.

I do think we should definitely start to think about having tests that do both permutations tho - even if users don't ask for it, it will catch bugs, incompatibilities, and assumptions in CI that we are likely to otherwise miss.

bleggett: Also just to clarify - if we want to enforce CP first then zt/CNI, it's `N+1` on the control plane side actually:

> The ztunnel at version 1.x is generally compatible with _control plane at version_ 1.x+1 and 1.x, which means the control plane must be upgraded first.

linsun: That is correct. Your update LG

linsun: It is not only about testing, it is also about ensure ztunnel code works with older CP, which is additional work and baggage to carry when there is WDS change.  We don't support updating DP first with sidecars, and I don't recall anyone complains.  People usually complains who they follow the doc and stuff doesn't work.

dhawton (approved):

PR #14617, opened 2024-02-15
Note on DR workloadSelector

NaturezzZ:
## Description

Supplementary explanation based on https://github.com/istio/istio/issues/49111#issuecomment-1921674872.
The same field name WorkloadSelector has a different effect/behavior in DR and SE. WorkloadSelector in ServiceEntry indicates destination workloads of traffic, while in DR workloadSelector reveals which workloads the traffic emanating from uses the DR. 

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [X] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/reference/config/networking/destination-rule/index.html
+++ b/content/en/docs/reference/config/networking/destination-rule/index.html
@@ -136,6 +136,11 @@ spec:
 <p>Destination Rules can be customized to specific workloads as well.
 The following example shows how a destination rule can be applied to a
 specific workload using the workloadSelector configuration.</p>
+<p><strong>Note:</strong> The workloadSelector configuration in 
+Destination Rules reveals which workloads the traffic emanating from 
+uses this Destination Rule, unlike the workloadSelector configuration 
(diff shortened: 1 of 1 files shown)


kfaseela (changes requested):
The source for this is not in istio.io repo, rather: https://github.com/istio/api/blob/master/networking/v1alpha3/destination_rule.proto#L167

NaturezzZ:
> The source for this is not in istio.io repo, rather: https://github.com/istio/api/blob/master/networking/v1alpha3/destination_rule.proto#L167

Thanks, I see. So is the right place to apply this PR is [istio/istio](https://github.com/istio/api/)?

kfaseela:
> > The source for this is not in istio.io repo, rather: https://github.com/istio/api/blob/master/networking/v1alpha3/destination_rule.proto#L167
> 
> Thanks, I see. So is the right place to apply this PR is [istio/api](https://github.com/istio/api/)?

yes

PR #14643, opened 2024-02-22
Fix make server failure after hugo update

ericvn:
## Description

Fix make server failure after Hugo update. remove extra ()'s.


zirain (approved):

ericvn:
Running into an issue trying to do a dry-build for the 1.21 branching. Looks like we need this fix back in 1.20

PR #14662, opened 2024-02-27
gateway-api: fix broken links and use v1beta1 in examples for now

frankbu:
## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/blog/2022/getting-started-gtwapi/index.md
+++ b/content/en/blog/2022/getting-started-gtwapi/index.md
@@ -127,7 +127,7 @@ EOF
 {{< /text >}}
 
 The controller that will implement a Gateway is selected by referencing a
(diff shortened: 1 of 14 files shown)


ericvn:
See https://github.com/istio/istio/pull/49583 for `v1` vs `v1beta1` discussion. I'm not sure we really need to change the docs back to `v1beta`, but suspect this is just to keep them in sync with the samples?

frankbu:
> I'm not sure we really need to change the docs back to v1beta, but suspect this is just to keep them in sync with the samples?

Problem is that currently if somebody tries to run the steps in these docs with v1 resources on a GKE cluster (maybe some other platforms as well), they will fail because GKE has only the v1beta1 resources pre-installed. So for now we'll just use the beta1 resources in all the samples and docs.

ericvn (approved):
Frank, thanks for the answer.

PR #14670, opened 2024-02-29
Change Istio documentation license to Creative Commons Attribution 4.0 International.

craigbox:
The Istio documentation repo is currently published under the Apache 2.0 license.  It should be published under the Creative Commons 4.0 according to point 11.f of the CNCF Charter.  This wasn’t ever a deliberate choice, just an oversight which wasn’t caught earlier.

(See also https://github.com/kubernetes/website/issues/1762. It always amuses me when I search for something on Google and the first hit is Past Me, having completely forgotten I did this.)

Hold merging on approval from the LF on this action, being sought in https://cncfservicedesk.atlassian.net/servicedesk/customer/portal/1/CNCFSD-2189

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,202 +1,395 @@
+Attribution 4.0 International
 
-                                Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+=======================================================================
 
-   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+Creative Commons Corporation ("Creative Commons") is not a law firm and
+does not provide legal services or legal advice. Distribution of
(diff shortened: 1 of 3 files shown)


ericvn:
Hmm.  I think the LICENSE file get's updated in a common-files update, so we will want to add some logic to prevent that from overriding this file moving forward. I know we have some logic to prevent CONTRIBUTING from getting updated as well.

ericvn:
https://github.com/istio/common-files/pull/999 created to not update the CCA 4.0 license file during `make update-common`

craigbox:
I understand the conclusion here is "too hard".  CNCF advice is we don't have the right to relicense without the permission of all the previous authors, and we have taken our contributions under the Apache license since the beginning.  Closing with regret.

PR #14697, opened 2024-03-07
Improve ambient mode docs

craigbox:
Initial edits, mostly around clarifying that you can install Istio in ambient mode, and you then get an ambient mesh: the feature is not *called* ambient mesh.


ericvn (approved):
Thanks for the improvements.

PR #14736, opened 2024-03-13
Release branching for 1.21.0

ericvn:
## Description

Release branching for 1.21.0.


craigbox (approved):
Imma only read the 259 removed lines if that's OK, I don't have time to read 2603473 added lines

PR #14753, opened 2024-03-15
Add Gateway API instructions to gateway TLS origination doc

frankbu:
## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/gtwapi_mtls_test.sh
+++ b/content/en/docs/tasks/traffic-management/egress/egress-gateway-tls-origination/gtwapi_mtls_test.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC1090,SC2154
+
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source "tests/util/gateway-api.sh"
+install_gateway_api_crds
+
+# @setup profile=none
+istioctl install --set profile=minimal --set values.pilot.env.PILOT_ENABLE_ALPHA_GATEWAY_API=true --set meshConfig.accessLogFile=/dev/stdout -y
+_wait_for_deployment istio-system istiod
+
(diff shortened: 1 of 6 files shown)


frankbu:
The failing test (`TestDocs/tasks/traffic-management/egress/egress-tls-origination/mtls_test.sh/mtls_test.sh`) is unrelated to the changes in this PR. It looks to me like it is broken with the current version of istio being used for testing. I ran it locally, and get the same failure.

@ericvn any idea how we managed to update the test istio, without running this failing test? Maybe something to do with the release update not running all tests?

frankbu:
Waiting for new istio test ref to merge, then `/retest` should fix failing tests

ericvn:
I think a fix for the profile-none test is to update Step 5 in the Istio tab for https://preliminary.istio.io/latest/docs/tasks/traffic-management/tcp-traffic-shifting/#apply-weight-based-tcp-routing to have `apiVersion: networking.istio.io/v1`

ericvn:
Still not sure why `istioctl proxy-config secret deploy/sleep | grep client-credential` seems to be empty for the `https://preliminary.istio.io/latest/docs/tasks/traffic-management/egress/egress-tls-origination/#configure-mutual-tls-origination-for-egress-traffic-at-sidecar` test

frankbu:
The profile-none failure in this PR is my new test, which also fails locally.

I think both failing tests in this PR might be related to https://github.com/envoyproxy/envoy/issues/24373. When running locally I saw the same error in the logs as described in that issue.

ericvn:
This is what the output looks like:
```
istioctl pc secret deploy/sleep                                                                                                    ✔  18:51:08
RESOURCE NAME     TYPE           STATUS     VALID CERT     SERIAL NUMBER                        NOT AFTER                NOT BEFORE
default           Cert Chain     ACTIVE     true           12a949149769ff00f212cb9062271483     2024-03-20T18:49:56Z     2024-03-19T18:47:56Z
ROOTCA            CA             ACTIVE     true           b19a466d83fe9e409cf6a6c4c99fda4d     2034-03-17T18:49:34Z     2024-03-19T18:49:34Z
```
and the name isn't `kubernetes://client-credential`

frankbu:
The second failing test is a little different. Instead of missing, the proxy has the secret but the status is not ACTIVE:
```
kubernetes://client-credential                           WARMING     false
```
It seems that both tests have the same fundamental issue: the secret is not properly configured in the proxy. That's why I'm thinking it might be related to https://github.com/envoyproxy/envoy/issues/24373.

keithmattix:
Yeah, now that we're using delta xDS on master, it wouldn't surprise me that some of our docs would fail since any SDS removal will result in a NACK

ericvn:
```
VERIFY FAILED snip_configure_mutual_tls_origination_for_egress_traffic_10 (timeout after 120s):
received:
"kubernetes://client-credential                           WARMING     false
kubernetes://client-credential-cacert                    WARMING     false"
expected:
"kubernetes://client-credential            Cert Chain     ACTIVE"
```

ericvn (approved):

PR #14778, opened 2024-03-25
Grafana: Use latest dashboard revision in import reference script

tungbq:
## Description
Use latest dashboard revision in import reference script:
In the `content/en/docs/ops/integrations/grafana/index.md` file, there is an issue with the `REVISION` variable in the script obtained from the `curl` command, as it returns a list. This causes subsequent steps to fail.  To make it works, we should specify revision in `REVISION` then pass into the download command. May be the latest one is the best candidate.

## Reviewers

- [ ] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn:
Is it possible to create a simple set to verify this script? There is very little on this page, so adding the istio.io test may be fairly easy and get another page shown as tested.

ericvn (approved):

tungbq:
> Is it possible to create a simple set to verify this script? There is very little on this page, so adding the istio.io test may be fairly easy and get another page shown as tested.

I'm new to the test framework of istio.io, so it might take time to figure out and writing the test case for this one.
I think we could have a separate task to improve this page.

Issue #14355, opened 2023-12-19
update ambient get started guide for the inpod cni stuff once merged


linsun:
after https://github.com/istio/istio/issues/48212 PRs are merged, the cni limitations should be removed.

linsun:
Bhavya Sachdeva pinged me offline that she is interested, I'll wait for her to comment in GitHub before assign to her.

Note: this is not merged yet.

linsun:
This should be unblocked, please try latest master or release 1.21 build, or wait for the official 1.21 release. see https://github.com/istio/istio/issues/48212.

linsun:
Sorry, just got confirmation from Eric that the change is not in beta 0 yet, but should be in -beta.1 since it was merged after beta.0 was released. I expect that the new build will be available in a day or 2.

linsun:
beta 1 is already avail - anyone interested in helping this out?

christian-posta:
What exactly is needed here? The current getting started correctly calls out that ambient mode can run on any cni that is supported by sidecar mode:

https://preliminary.istio.io/latest/docs/ops/ambient/getting-started/

linsun:
I think this is already updated, and could probably be closed. cc @bleggett, if you agree, feel free to close it.LinSent from my iPhone On Mar 26, 2024, at 1:47 AM, Christian Posta ***@***.***> wrote:﻿
What exactly is needed here? The current getting started correctly calls out that ambient mode can run on any cni that is supported by sidecar mode:
https://preliminary.istio.io/latest/docs/ops/ambient/getting-started/

christian-posta:
Actually, @nauticalmike had some suggestion to improve the guide based on the listing outputs. I'll let him comment here. Otherwise, I think we're in a good state for this issue.

nauticalmike:
I'm working on this @linsun FYI

nauticalmike:
https://github.com/istio/istio.io/pull/14788

linsun:
Saw #14788 also merged, going to close this out, please reopen if I'm wrong. Thank you @nauticalmike for updating the tests!!

PR #14788, opened 2024-03-26
Ambient beta get started guide patch

nauticalmike:
## Description

adjusting ambient getting started guide to include in-pod ztunnel log proxy status #14355 

## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn:
`make snips` is used to generate the file from the index.md.

nauticalmike:
@ericvn what's the next step here?

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -65,8 +65,7 @@ _verify_contains snip_verify_traffic_notsleep_to_productpage "$snip_verify_traff
 snip_adding_your_application_to_the_ambient_mesh_1
 
 # test traffic after ambient mode is enabled
-_verify_contains snip_adding_your_application_to_the_ambient_mesh_2 "$snip_adding_your_application_to_the_ambient_mesh_2_out"
-_verify_contains snip_adding_your_application_to_the_ambient_mesh_3 "$snip_adding_your_application_to_the_ambient_mesh_3_out"
+_verify_contains snip_adding_your_application_to_the_ambient_mesh_2 "namespace/default labeled"
```
frankbu, line 68: Why aren't you testing snips 1, 3, 5, and 6?

Also, since you're just testing the doc, not the function, you don't need to `verify_` on commands that the doc doesn't show what to expect:
```suggestion
snip_adding_your_application_to_the_ambient_mesh_2
```

nauticalmike: @frankbu I generated the snips using `make gen` and just adjusted for the content but the content I want to add is the actual ztunnel log verification to educate the user on the new in-pod redirection so the output to be matched is not always the same being the reason why I skipping some and some of those where skipped from the previous version of the doc. Let me know if I'm missing anything.

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -245,7 +243,21 @@ $ export GATEWAY_SERVICE_ACCOUNT=ns/istio-system/sa/bookinfo-gateway-istio
 
 ## Adding your application to the ambient mesh {#addtoambient}
 
-You can enable all pods in a given namespace to be part of an ambient mesh
+When an application pod is part of an ambient mesh, you can check the ztunnel proxy logs to confirm the mesh is redirecting traffic.
+Before we label the namespace to be part of an ambient mesh, check the ztunnel logs related to `inpod` which indicate that in-pod redirection mode is enabled:
```
frankbu, line 247: What exactly in the output should the user look for to check that in-pod redirection mode is enabled? `inpod_enabled: true`? The doc test should also check for the same thing.

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -255,6 +267,20 @@ $ kubectl label namespace default istio.io/dataplane-mode=ambient
 Congratulations! You have successfully added all pods in the default namespace
 to the mesh. Note that you did not have to restart or redeploy anything!
 
+Check once again the ztunnel logs for the proxy has received the network namespace (netns) information about an ambient application pod, and has started proxying for it:
+
+{{< text syntax=bash snip_id=none >}}
```
frankbu, line 272: Instead of snip_id=none, you should generate this snip and then call it in test.sh and confirm that the output contains what the doc says it should, e.g., `received netns, starting proy`

content/en/docs/ops/ambient/getting-started/index.md
nauticalmike: I agree, I thought this is what I was doing exactly for that before and after labeling the ns on my snips:
```
snip_adding_your_application_to_the_ambient_mesh_1() {
kubectl logs ds/ztunnel -n istio-system  | grep inpod
}

! IFS=$'\n' read -r -d '' snip_adding_your_application_to_the_ambient_mesh_1_out <<\ENDSNIP
Found 3 pods, using pod/ztunnel-jrxln
inpod_enabled: true
inpod_uds: /var/run/ztunnel/ztunnel.sock
inpod_port_reuse: true
inpod_mark: 1337
2024-03-26T00:02:06.161802Z  INFO ztunnel::inpod::workloadmanager: handling new stream
2024-03-26T00:02:06.162099Z  INFO ztunnel::inpod::statemanager: pod received snapshot sent
ENDSNIP

snip_adding_your_application_to_the_ambient_mesh_2() {
kubectl label namespace default istio.io/dataplane-mode=ambient
}
```

Do you mean like actually also checking after is labeled? because that is exactly I understood from you before that we didn't need to check so I added this to the one after labeling:

```
$ kubectl logs ds/ztunnel -n istio-system  | grep inpod
Found 3 pods, using pod/ztunnel-jrxln
inpod_enabled: true
inpod_uds: /var/run/ztunnel/ztunnel.sock
inpod_port_reuse: true
inpod_mark: 1337
2024-03-26T00:02:06.161802Z  INFO ztunnel::inpod::workloadmanager: handling new stream
2024-03-26T00:02:06.162099Z  INFO ztunnel::inpod::statemanager: pod received snapshot sent
2024-03-26T00:41:05.518194Z  INFO ztunnel::inpod::statemanager: pod WorkloadUid("7ef61e18-725a-4726-84fa-05fc2a440879") received netns, starting proxy
{{< /text >}}
```

now that I think about it, on the before labeling I should just verify for `inpod_enabled: true` and after labeling we just care about `received netns, starting proxy` what do you think?

content/en/docs/ops/ambient/getting-started/test.sh
nauticalmike: I agree. Understood.

content/en/docs/ops/ambient/getting-started/index.md
nauticalmike: missed this one. We are on the same page, I will update this soon.

nauticalmike:
/retest

For some reason now some other tests that I didn't touch are failing and they were not before:
`FAIL: TestDocs/tasks/security/authorization/authz-ingress/gtwapi_test.sh/gtwapi_test.sh (35.57s)`
```
Waiting for deployment "httpbin" rollout to finish: 0 of 1 updated replicas are available...
deployment "httpbin" successfully rolled out
gateway.gateway.networking.k8s.io/httpbin-gateway created
httproute.gateway.networking.k8s.io/httpbin created
error: timed out waiting for the condition on gateways/httpbin-gateway
```

`FAIL: TestDocs/tasks/security/authorization/authz-tcp/test.sh/test.sh (127.89s)`
```
Waiting for deployment "tcp-echo" rollout to finish: 0 of 1 updated replicas are available...
deployment "tcp-echo" successfully rolled out
deployment "sleep" successfully rolled out
VERIFY FAILED snip_before_you_begin_4 (timeout after 120s):
received:
```

nauticalmike:
@frankbu do I need to do anything else in this step?

content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -255,6 +261,13 @@ $ kubectl label namespace default istio.io/dataplane-mode=ambient
 Congratulations! You have successfully added all pods in the default namespace
 to the mesh. Note that you did not have to restart or redeploy anything!
 
+Check once again the ztunnel logs for the proxy has received the network namespace (netns) information about an ambient application pod, and has started proxying for it:
+
+{{< text bash >}}
+$ kubectl logs ds/ztunnel -n istio-system | grep "starting proxy"
+received netns, starting proxy
```
frankbu, line 268: Is this really the expected output? Isn't there text before this string?
```suggestion
... received netns, starting proxy
```

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -65,7 +65,7 @@ _verify_contains snip_verify_traffic_notsleep_to_productpage "$snip_verify_traff
 snip_adding_your_application_to_the_ambient_mesh_1
 
 # test traffic after ambient mode is enabled
-_verify_contains snip_adding_your_application_to_the_ambient_mesh_2 "$snip_adding_your_application_to_the_ambient_mesh_2_out"
+snip_adding_your_application_to_the_ambient_mesh_2
 _verify_contains snip_adding_your_application_to_the_ambient_mesh_3 "$snip_adding_your_application_to_the_ambient_mesh_3_out"
 _verify_contains snip_adding_your_application_to_the_ambient_mesh_4 "$snip_adding_your_application_to_the_ambient_mesh_4_out"
```
frankbu, line 70: Lines 65-70 need to be updated to properly call all the snips:
```
_verify_same snip_adding_your_application_to_the_ambient_mesh_1 "$snip_adding_your_application_to_the_ambient_mesh_1_out"

# test traffic after ambient mode is enabled
snip_adding_your_application_to_the_ambient_mesh_2
_verify_like snip_adding_your_application_to_the_ambient_mesh_3 "$snip_adding_your_application_to_the_ambient_mesh_3_out"
_verify_same snip_adding_your_application_to_the_ambient_mesh_4 "$snip_adding_your_application_to_the_ambient_mesh_4_out"
_verify_same snip_adding_your_application_to_the_ambient_mesh_5 "$snip_adding_your_application_to_the_ambient_mesh_5_out"
_verify_same snip_adding_your_application_to_the_ambient_mesh_6 "$snip_adding_your_application_to_the_ambient_mesh_6_out"
```

content/en/docs/ops/ambient/getting-started/snips.sh
```diff
@@ -308,20 +324,36 @@ kubectl exec deploy/sleep -- sh -c "for i in \$(seq 1 100); do curl -s http://$G
...
+kubectl delete namespace istio-system
+}
+
+snip_uninstall_4() {
 kubectl delete -f samples/sleep/sleep.yaml
 kubectl delete -f samples/sleep/notsleep.yaml
 }
 
-snip_uninstall_4() {
+snip_uninstall_5() {
```
frankbu, line 357: You need to add a call to snip_uninstall_5 in the @cleanup section of both test.sh and gtwapi_test.sh

content/en/docs/ops/ambient/getting-started/index.md
nauticalmike: right looks better with the `...` I originally wanted to showcase the whole entry but the `_verify_contains` and `_verify_like` just were not working for the type of output

content/en/docs/ops/ambient/getting-started/snips.sh
nauticalmike: I added the uninstall calls, thanks

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -65,7 +65,7 @@ _verify_contains snip_verify_traffic_notsleep_to_productpage "$snip_verify_traff
 _verify_contains snip_adding_your_application_to_the_ambient_mesh_1 "$snip_adding_your_application_to_the_ambient_mesh_1_out"
 
 # test traffic after ambient mode is enabled
-_verify_contains snip_adding_your_application_to_the_ambient_mesh_3 "$snip_adding_your_application_to_the_ambient_mesh_3_out"
+_verify_contains snip_adding_your_application_to_the_ambient_mesh_3 "received netns, starting proxy"
```
nauticalmike, line 68: @frankbu I decided to do the validation directly on the test script as the snips keeps on failing. I see someone did the same before here on line 77

nauticalmike: test failed again... running out of ideas here...

dhawton:
```VERIFY FAILED snip_adding_your_application_to_the_ambient_mesh_3 (timeout after 120s):
received:
""
expected:
"received netns, starting proxy"```

Comment (dhawton) [reply content/en/docs/ops/ambient/getting-started/test.sh:68]: Might be worth troubleshooting by dropping the grep in that snip and just having it dump the logs, just to see what is being logged in the ztunnel pods.

Comment (nauticalmike) [reply content/en/docs/ops/ambient/getting-started/test.sh:68]: thanks @dhawton I found that one, I tried several combinations to get that snippet of the doc in but the test keeps on failing for different reasons. The idea here is to show the logs after the ns labeling, but when I use the whole output like:
```
kubectl logs ds/ztunnel -n istio-system  | grep inpod
Found 3 pods, using pod/ztunnel-jrxln
inpod_enabled: true
inpod_uds: /var/run/ztunnel/ztunnel.sock
inpod_port_reuse: true
inpod_mark: 1337
2024-03-26T00:02:06.161802Z  INFO ztunnel::inpod::workloadmanager: handling new stream
2024-03-26T00:02:06.162099Z  INFO ztunnel::inpod::statemanager: pod received snapshot sent
2024-03-26T00:41:05.518194Z  INFO ztunnel::inpod::statemanager: pod WorkloadUid("7ef61e18-725a-4726-84fa-05fc2a440879") received netns, starting proxy
```

the test `_verify_like` or `_verify_contains` doesn't match but I can see a similar print on the artifacts output, my last attempt just removed the snips.sh out reference and match the contains directly like:
```
_verify_contains snip_adding_your_application_to_the_ambient_mesh_3 "received netns, starting proxy"
```
but as you can see the output is empty

Comment (nauticalmike) [reply content/en/docs/ops/ambient/getting-started/test.sh:68]: that's the thing, on my previous validations I was able to see an output like described above but one pod instead of 3 and obviously different time stamps and workloadid but neither the `_verify_like` or `_verify_contains` matched anything...

Comment (dhawton) [reply content/en/docs/ops/ambient/getting-started/test.sh:68]: I can't see the output of the log really, I can only tell grep had no match... the received netns statement is ~2 months old, so it's in the commit of Istio that the docs is using atm. Wonder if it's not starting up, or hasn't finished initializing when the test is being run?

File: content/en/docs/ops/ambient/getting-started/test.sh [hunk 8]
```diff
@@ -62,10 +62,9 @@ _verify_contains snip_verify_traffic_sleep_to_ingress "$snip_verify_traffic_slee
 _verify_contains snip_verify_traffic_sleep_to_productpage "$snip_verify_traffic_sleep_to_productpage_out"
 _verify_contains snip_verify_traffic_notsleep_to_productpage "$snip_verify_traffic_notsleep_to_productpage_out"
 
-snip_adding_your_application_to_the_ambient_mesh_1
+_verify_contains snip_adding_your_application_to_the_ambient_mesh_1 "$snip_adding_your_application_to_the_ambient_mesh_1_out"
```
frankbu, line 65: I'm not surprised it's failing now, since you're not calling snip_adding_your_application_to_the_ambient_mesh_2

Why didn't you change to the code I suggested in my previous comment? https://github.com/istio/istio.io/pull/14788#pullrequestreview-1964009799 
```suggestion
_verify_contains snip_adding_your_application_to_the_ambient_mesh_1 "$snip_adding_your_application_to_the_ambient_mesh_1_out"
snip_adding_your_application_to_the_ambient_mesh_2
```

nauticalmike: I'm not calling it because when I generate the snips there is no `snip_adding_your_application_to_the_ambient_mesh_2` and that is because as pointed out before there is no need to validate this section:
```
$ kubectl label namespace default istio.io/dataplane-mode=ambient
{{< /text >}}
```
as I'm not referencing any output on the document.

Are you saying that even though in this case I should remove the `snip_id=none ` let it generate the `snip_adding_your_application_to_the_ambient_mesh_2` and just not call any validation from the test.sh?

frankbu: I said that you don't need to validate the output of the snip, not that you don't need to call it. If you don't call the snip that adds the label, how could it possibly get labelled so that the following snips will work?

Just put it back to generate all the snips the way you had it previously and change the test to what I said yesterday.
```
_verify_same snip_adding_your_application_to_the_ambient_mesh_1 "$snip_adding_your_application_to_the_ambient_mesh_1_out"

# test traffic after ambient mode is enabled
snip_adding_your_application_to_the_ambient_mesh_2
_verify_like snip_adding_your_application_to_the_ambient_mesh_3 "$snip_adding_your_application_to_the_ambient_mesh_3_out"
_verify_same snip_adding_your_application_to_the_ambient_mesh_4 "$snip_adding_your_application_to_the_ambient_mesh_4_out"
_verify_same snip_adding_your_application_to_the_ambient_mesh_5 "$snip_adding_your_application_to_the_ambient_mesh_5_out"
_verify_same snip_adding_your_application_to_the_ambient_mesh_6 "$snip_adding_your_application_to_the_ambient_mesh_6_out"
```
Either that, or disable the test (rename it to `test-disable.sh`) and I'll fix it in a followup.

content/en/docs/ops/ambient/getting-started/test.sh
```diff
@@ -62,12 +62,14 @@ _verify_contains snip_verify_traffic_sleep_to_ingress "$snip_verify_traffic_slee
...
 
-snip_adding_your_application_to_the_ambient_mesh_1
+_verify_same snip_adding_your_application_to_the_ambient_mesh_1 "$snip_adding_your_application_to_the_ambient_mesh_1_out"
 
 # test traffic after ambient mode is enabled
-_verify_contains snip_adding_your_application_to_the_ambient_mesh_2 "$snip_adding_your_application_to_the_ambient_mesh_2_out"
-_verify_contains snip_adding_your_application_to_the_ambient_mesh_3 "$snip_adding_your_application_to_the_ambient_mesh_3_out"
-_verify_contains snip_adding_your_application_to_the_ambient_mesh_4 "$snip_adding_your_application_to_the_ambient_mesh_4_out"
+snip_adding_your_application_to_the_ambient_mesh_2
+_verify_like snip_adding_your_application_to_the_ambient_mesh_3 "$snip_adding_your_application_to_the_ambient_mesh_3_out"
```
frankbu, line 69: Actually, verify_like won't work here because of token count mismatch. Just change it to explicitly check for the required output for now, and I'll think about how to improve in a followup.
```suggestion
_verify_contains snip_adding_your_application_to_the_ambient_mesh_3 "received netns, starting proxy"
```

frankbu: right, that part will work, but not ideal. The test had other problems previously.

content/en/docs/ops/ambient/getting-started/test.sh
frankbu: `_verify_same` and `_verify_contains` will both work in these cases, but `_verify_same` is a better check since it really is checking that the output is "exactly the same as expected", which it is in this case. `_verify_contains` checks that the expected string is part of the output, which could include more text.

frankbu: https://github.com/istio/istio.io/blob/master/tests/README.md#verify-functions

nauticalmike: understood, I think I tested with that and succeed locally but failed here. I will send one like this and we'll see, thanks.

frankbu (approved):

content/en/docs/ops/ambient/getting-started/index.md
bleggett: `inpod_enabled` will always be true with our charts - we offer no way to turn it off, it is the only mode. So people should not need to check this.

nauticalmike: ha, thanks @bleggett I didn't know that. The intention here was to show the before and after labeling logs to show some action without having to get into the actual mechanics as this is intended to be a getting started guide.

As this validation is being done before the ns labeling, do you think it would be better just to check for the string `ztunnel::inpod`?

bleggett: There's no other mode and I don't want to teach people the term `inpod` - 

the checks are "is ambient enabled/do you have a ztunnel, did you label something". There are no other checks people using our charts need.

nauticalmike: I will open another PR for this, thanks

PR #14816, opened 2024-04-02
Remove the reference to installed-state CR

### LINKED ISSUE #14811: remove references to "install-state" IOP CR from the "install with istioctl" docs.
Despite having [removed this feature](https://github.com/istio/istio/pull/47860/files) there are still [docs telling users to expect it](https://istio.io/latest/docs/setup/install/istioctl/#check-whats-installed). We should remove them to [avoid confusion](https://istio.slack.com/archives/C01BAFJLEAD/p1711967275403699).

adityasamant25:
## Description
Fixes #14811

This PR removes the [reference](https://istio.io/latest/docs/setup/install/istioctl/#check-whats-installed) to the `installed-state` CR in the [Install with Istioctl](https://istio.io/latest/docs/setup/install/istioctl/) docs. 

This is necessary as this CR is no longer available and the following command mentioned in the docs is failing:
`kubectl -n istio-system get IstioOperator installed-state -o yaml > installed-state.yaml `

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/istioctl/index.md
+++ b/content/en/docs/setup/install/istioctl/index.md
@@ -106,28 +106,6 @@ to install the `demo` profile:
 $ istioctl install --set profile=demo
 {{< /text >}}
 
-## Check what's installed
-
-The `istioctl` command saves the `IstioOperator` CR that was used to install Istio in a copy of the CR named `installed-state`.
-Instead of inspecting the deployments, pods, services and other resources that were installed by Istio, for example:
-
-{{< text bash >}}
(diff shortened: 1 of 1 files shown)


hanxiaop (commented):
/ok-to-test

Since in later section we also have `istioctl verify-install`, I think it's reasonable to remove this section entirely.

PR #14839, opened 2024-04-05
Mark Gateway API as stable, beta for mesh

howardjohn:
Mesh is declared as Beta because it just graduated upstream, so it has
much less soak time. Additionally, we fail one conformance test. I
would like to note the failing conformance test somewhere, but not sure
where.

### DIFF
--- a/data/features.yaml
+++ b/data/features.yaml
@@ -87,14 +87,22 @@ features:
       maturity: Beta
       nextExpectedPromotion: "1.13"
     area: Traffic Management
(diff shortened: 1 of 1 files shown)


dhawton (approved):

ericvn:
@howardjohn Would you make these changes in https://github.com/istio/enhancements/blob/master/features.yaml. This change will be overwritten by the nightly reference doc update when we grab the enhancements file.

Todo for me is to figure out why we don't flag this being overridden by the pre-submit tests.

PR #14846, opened 2024-04-05
add notes

dhawton:
## Description

In prep for releases for 1.21.1, 1.20.5, and 1.19.9 targeting Monday. Will manually cherrypick to `master` after merge.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

PR #14863, opened 2024-04-09
Fix CVSS Vector Calculator Link Gen

dhawton:
## Description

The vector calculator does not support `CVSS:x.x/` as part of the vector to define the version to use. It's a seperate query parameter.

Issue found here: https://github.com/istio/istio.io/pull/14846#discussion_r1555964752

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/news/security/istio-security-2019-001/index.md
+++ b/content/en/news/security/istio-security-2019-001/index.md
@@ -4,7 +4,8 @@ subtitle: Security Bulletin
 description: Incorrect access control.
 cves: [CVE-2019-12243]
 cvss: "8.9"
-vector: "CVSS:3.0/AV:A/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:N/E:H/RL:O/RC:C"
+vector: "AV:A/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:N/E:H/RL:O/RC:C"
+cvss_version: "3.0"
 releases: ["1.1 to 1.1.6"]
 publishdate: 2019-05-28
 keywords: [CVE]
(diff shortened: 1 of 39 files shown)


kfaseela:
I did not understand much of it :D May be @craigbox knows better?

dhawton:
> I did not understand much of it :D May be @craigbox knows better?

The short of it... if we put CVSS/3.1: in the vector param, we link to something NVD's calculator won't parse:

[https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:L/A:L](https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:L/A:L)

Versus if we remove it from the vector and pass it as a version query param instead, which seems to be how NVD uses the tool:

[https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:L/A:L&version=3.1](https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:L/A:L&version=3.1)

ericvn:
This LGTM. I'm thinking we might have been able to parse the version out of the existing value, as well as the vector and used that when building the URL. But since you have already manually done that, I'm not sure if we need to actually write additional code. I'll let Craig weigh in as well.

dhawton:
> This LGTM. I'm thinking we might have been able to parse the version out of the existing value, as well as the vector and used that when building the URL. But since you have already manually done that, I'm not sure if we need to actually write additional code. I'll let Craig weigh in as well.

Yeah, I wasn't sure which approach would be best. I'm not strongly opinionated either direction, so if there's a preference to pull it out and extract from the vector string I can do that.

craigbox:
I'm happy with the work having been done already.  Thanks!

craigbox (approved):

PR #14866, opened 2024-04-10
Fix incorrect issue link in releasenotes

wilsonwu:
## Description

Fix bad issue link in releasenotes, thanks @hanxiaop, and already fixed in the ZH translation PR.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


kfaseela (approved):

ericvn:
Might want to see if this needs to be fixed in the actual release note, else it may creep into the 1.22 notes (I haven't searched back to when this was introduced).

dhawton:
> Might want to see if this needs to be fixed in the actual release note, else it may creep into the 1.22 notes (I haven't searched back to when this was introduced).

It is an error in the release note yaml, so it'll need to get fixed in there.

ericvn:
> It is an error in the release note yaml, so it'll need to get fixed in there.

Updating the yaml was what I meant to say.

PR #14876, opened 2024-04-13
add tips for CEL expression

zirain:
fix: https://github.com/istio/istio/issues/49853

### DIFF
--- a/content/en/docs/tasks/observability/logs/telemetry-api/index.md
+++ b/content/en/docs/tasks/observability/logs/telemetry-api/index.md
@@ -110,6 +110,10 @@ $ kubectl apply -f @samples/open-telemetry/loki/otel.yaml@ -n istio-system
     EOF
     {{< /text >}}
 
+   {{< tip >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/observability/logs/telemetry-api/index.md
```diff
@@ -110,6 +110,10 @@ $ kubectl apply -f @samples/open-telemetry/loki/otel.yaml@ -n istio-system
     EOF
     {{< /text >}}
 
+    {{< tip >}}
+    There's no `response.code` attribute at all when connection fails, in that case, you should use CEL expression `!has(response.code) || response.code >= 500`.
```
ericvn, line 114:
```suggestion
    There's no `response.code` attribute when connections fail. In that case, you should use the CEL expression `!has(response.code) || response.code >= 500`.
```

ericvn:
@howardjohn Is this a good tip to include? I know it solves the noted issue, but you also mention doing some more investigation. Thanks.

PR #14884, opened 2024-04-16
ambient: comment no gateway as default

daixiang0:
## Description

Fix https://github.com/istio/istio/issues/50219

## Reviewers

- [x] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/getting-started/index.md
+++ b/content/en/docs/ops/ambient/getting-started/index.md
@@ -66,6 +66,10 @@ Follow these steps to get started with Istio's ambient mode:
 $ istioctl install --set profile=ambient --set "components.ingressGateways[0].enabled=true" --set "components.ingressGateways[0].name=istio-ingressgateway" --skip-confirmation
 {{< /text >}}
 
+{{< tip >}}
+Note that this command includes `--set "components.ingressGateways[0].enabled=true"` because the ambient profile does not install the ingress gateway by default.
+{{< /tip >}}
+
 After running the above command, you’ll get the following output that indicates
 five components (including {{< gloss "ztunnel" >}}ztunnel{{< /gloss >}}) have been installed successfully!
 


content/en/docs/ops/ambient/getting-started/index.md
```diff
@@ -101,6 +101,11 @@ four components (including {{< gloss "ztunnel" >}}ztunnel{{< /gloss >}}) have be
 
 {{< /tabset >}}
 
+{{< tip >}}
+Note that ambient mode does not include ingress gateway as default.
+{{< /tip >}}
+
+
```
wilsonwu, line 108:
```suggestion
{{< tip >}}
Note that ambient mode does not include ingress gateway as default.
{{< /tip >}}

```

ericvn:
I'm not sure that we need this called out here. We already document that there is in ingress gateway or egress gateway for the ambient profile on the profile page: https://istio.io/latest/docs/setup/additional-setup/config-profiles/

dhawton:
> I'm not sure that we need this called out here. We already document that there is in ingress gateway or egress gateway for the ambient profile on the profile page: https://istio.io/latest/docs/setup/additional-setup/config-profiles/

+1 we don't need to call it out here. We don't call out in our demos when installing the `default` profile that it does not have ingress gateway. Referencing the config-profiles doc is the best action and reduces information duplication.

frankbu:
> +1 we don't need to call it out here. We don't call out in our demos when installing the default profile that it does not have ingress gateway. Referencing the config-profiles doc is the best action and reduces information duplication.

The default profile DOES have ingress gateway.

I think it's helpful if the tip is changed as I suggested above (https://github.com/istio/istio.io/pull/14884#discussion_r1567360541), it's just a tip to point out that the ingress gateway needs to be explicitly enabled if using the bookinfo services in ambient.

That said, I'm also OK with not highlighting anything about installing gateways along with istiod, since we now recommend gateways be installed separately.

daixiang0:
@frankbu thank you, I have updated.

frankbu (approved):

PR #14888, opened 2024-04-16
Update ambient helm install index.md to be consistent

linsun:
## Description
1. pods and services requirement are mainly for sidecar, so no need to ask users to check it.
2. discovery component -> istiod as we don't explain discovery component here in install yet.
3. remove `--wait` to be consistent with the rest of cmds on this page.
4. Use similar words as https://istio.io/latest/docs/setup/install/istioctl/ without `ing`.

### DIFF
--- a/content/en/docs/ops/ambient/install/helm-installation/snips.sh
+++ b/content/en/docs/ops/ambient/install/helm-installation/snips.sh
@@ -42,7 +42,7 @@ helm install ztunnel istio/ztunnel -n istio-system
 }
 
 snip_install_ingress() {
-helm install istio-ingress istio/gateway -n istio-ingress --wait --create-namespace
+helm install istio-ingress istio/gateway -n istio-ingress --create-namespace
 }
 
 snip_configuration_1() {
--- a/content/en/docs/ops/ambient/install/helm-installation/snips.sh
+++ b/content/en/docs/ops/ambient/install/helm-installation/snips.sh
@@ -26,23 +26,23 @@ helm repo update
 }
 
 snip_install_base() {
-helm install istio-base istio/base -n istio-system --create-namespace
+helm install istio-base istio/base -n istio-system --create-namespace --wait
 }
 
 snip_install_cni() {
-helm install istio-cni istio/cni -n istio-system --set profile=ambient
+helm install istio-cni istio/cni -n istio-system --set profile=ambient --wait
 }
 
 snip_install_discovery() {
-helm install istiod istio/istiod --namespace istio-system --set profile=ambient
+helm install istiod istio/istiod --namespace istio-system --set profile=ambient --wait
 }
 
 snip_install_ztunnel() {
-helm install ztunnel istio/ztunnel -n istio-system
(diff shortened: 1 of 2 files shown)


linsun:
@dhawton and @bleggett  can you pls review?

content/en/docs/ops/ambient/install/helm-installation/snips.sh
```diff
@@ -42,7 +42,7 @@ helm install ztunnel istio/ztunnel -n istio-system
 }
 
 snip_install_ingress() {
-helm install istio-ingress istio/gateway -n istio-ingress --wait --create-namespace
```
bleggett, line 45: Can we just consistently use `--wait` everywhere?

It's useful IMO, helps make sure people following the steps get stuck at that step, if something is wrong.

bleggett (approved):
Would prefer to keep the `--wait`s and use them everywhere, but I'm fine if people don't care.

Rest LGTM.

content/en/docs/ops/ambient/install/helm-installation/snips.sh
linsun: One issue with the `--wait` is that it doesn't work on my Mac with kind, because LB external IP is pending so the cmd appears hang while the pod is already up.

```
 ~/Downloads/istio-1.21.0/ k get svc -n istio-ingress2
NAME            TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                                      AGE
istio-ingress   LoadBalancer   10.96.181.157   <pending>     15021:30903/TCP,80:30580/TCP,443:31328/TCP   2m58s
```

bleggett: That makes sense, but the ingress isn't much use without that IP anyway.

We should probably either remove the ingress installation, or specifically mention that you need something like metallb in `kind` if you want an ingress, otherwise we should recommend people to skip the ingress install.

linsun: > That makes sense, but the ingress isn't much use without that IP anyway.

I'd agree with that in production, but in testing, it is still useful as I can hit istio-ingressgateway svc and perform all types of testing as we documented in our get started guide.

dhawton (approved):
I agree with keeping `--wait`, for the test at least this should fail if an IP doesn't get assigned and the test times out. I see you added text for users in this situation, but ideally it shouldn't be an issue since we also recommend MetalLB for kind clusters on its [Platform Setup](https://istio.io/latest/docs/setup/platform-setup/kind/#setup-metallb-for-kind) page.

linsun:
> I agree with keeping `--wait`, for the test at least this should fail if an IP doesn't get assigned and the test times out. I see you added text for users in this situation, but ideally it shouldn't be an issue since we also recommend MetalLB for kind clusters on its [Platform Setup](https://istio.io/latest/docs/setup/platform-setup/kind/#setup-metallb-for-kind) page.

Maybe it was just Mac and me, last time I tried MetalLB it didn't work for my Mac and crashed in a day.  Could be fixed recently tho.

Issue #14597, opened 2024-02-08
Istio injection doc could use a few clarifications


linsun:
Doc: https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

What is not clear when it says `If either label is enabled, the pod is injected`, does the label include istio-injection, sidecar.istio.io/inject and 	istio.io/rev here? for example, if i have istio-injection=disabled on NS, and istio.io/rev=1.20 on the pod, would the injection happen?

linsun:
cc @howardjohn @peterj

howardjohn:
No, it would not due to the "If either label is disabled, the pod is not injected" part. Agree its a bit confusing with the 3 labels though

linsun:
Thanks, we can clarify it. The other one is confusing is:

i also found odd you had to remove istio-injection=enabled on the NS when you have istio.io/rev configured on the pod.

Thinking it over, this is as designed because istio-injection=enabled implies the default CP to inject

peterj:
Here are the scenarios is tested:

1. `istio-injection=enabled` & `istio.io/rev` set ==> default revision is used (not necessarily the one set in istio.io/rev)
2. No `istio-injection` label & `istio.io/rev` set ==> the revision set in `istio.io/rev` is used
3. `istio-injection=disabled` & `istio.io/rev` set ==> no injection happens

howardjohn:
Here is basically every combination possible: https://github.com/istio/istio/blob/44c2f6312201a4172f138b1953e763ced3170cff/operator/cmd/mesh/manifest-generate_test.go#L1060

The reason things behave as they do is largely for backwards compatibility

PR #14889, opened 2024-04-16
Clarification on injection

linsun:
hope to fixes: https://github.com/istio/istio.io/issues/14597

I don't think you could use istio.io/rev=disabled, so didn't list that label in the first bullet.


linsun:
@howardjohn can you PTAL? Thanks!

howardjohn (approved):

ericvn (approved):

PR #14922, opened 2024-04-19
Ambient get started update

linsun:
## Description

Updated ambient get started to latest change on waypoint, also:

fixes: https://github.com/istio/istio/issues/50145
## Reviewers

- [x ] Ambient
- [x ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/ambient/getting-started/index.md
+++ b/content/en/docs/ops/ambient/getting-started/index.md
@@ -349,32 +349,30 @@ command terminated with exit code 56
 
 ### Layer 7 authorization policy
 
-Using the Kubernetes Gateway API, you can deploy a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} for the `productpage` service that uses the `bookinfo-productpage` service account. Any traffic going to the `productpage` service will be mediated, enforced and observed by the Layer 7 (L7) proxy.
-
-Deploy a waypoint proxy for the `productpage` service:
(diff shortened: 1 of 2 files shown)


linsun:
@howardjohn can you take a quick look to see if there is any other simplification we can apply?

I kept L7 authz policy for productpage as it is nice to show the waypoint proxy worked on multiple services, just didn't call it from ingress-gw.

I removed the inpod stuff checking, think it is best to have it in a debug/ztunnel section vs getting started as inpod is the only mode we support so no need to check it. :)

dhawton (approved):
Changes look good to me on first pass. I put a DNM to allow others to look as well.

bleggett (commented):
In light of https://github.com/istio/istio/pull/50572 tweaked slightly

linsun:
> In light of [istio/istio#50572](https://github.com/istio/istio/pull/50572) tweaked slightly

Yep, thank you for the suggestions!

ilrudie (approved):
looks good, left a few minor tweaks but no show stoppers

linsun:
> looks good, left a few minor tweaks but no show stoppers

Thanks! I forgot to remove the extra cmd, good catch!

linsun:
remove don't merge as this is better than what we have - we can always iterate on the content.

PR #14941, opened 2024-04-23
Enable the ambient mode getting started tests

ericvn:
## Description

Test to see the ambient mode getting started failures.

PR #14953, opened 2024-04-24
starting fixup for ambient getting-stated tests

ilrudie:
## Description

Enable and fix ambient mode getting started tests

fixes #14941 

## Reviewers

- [X] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


ericvn:
Note you also need to rename the disabled gtw-test.sh file.

ilrudie:
I was thinking it might be easier to work through the failures one at a time but maybe that's not the case and I should dive in on the other tests too

ericvn:
> I was thinking it might be easier to work through the failures one at a time but maybe that's not the case and I should dive in on the other tests too

Either way works. If the failure in the second test is caused by missing cleanup in the first test, it may make sense to wait. Otherwise, if you are actually using the PR's testing to test changes instead of locally, I find it easier to debug both and change both with each commit, since it'll run both tests each time.

dhawton:
just popping DNM so it doesn't accidentally get merged while you're working on the tests

ilrudie:
@dhawton and @ericvn, I will remove the istio api bits shortly but I think this is looking much better now. LMK if it looks like I still have something broken which just isn't failing

dhawton:
> @dhawton and @ericvn, I will remove the istio api bits shortly but I think this is looking much better now. LMK if it looks like I still have something broken which just isn't failing

Looks LGTM. Nothing stands out.

ericvn (approved):
LGTM. Feel free to remove the DNM whenever you are done.

PR #14968, opened 2024-04-25
Tweak the LHS moving ambient and slight rename for legacy

ericvn:
## Description

From the discussion in yesterday's ambient call, here is a slight tweak to the LHS naming/organization for discussion.

I moved the Ambient Mode item up.
I wanted to rename Setup to something else so went with `Legacy`. Maybe there is a better word.
I removed Kubernetes from the description as we have VM under that topic.

Look at the left hand column as well as text for items on the right:
https://deploy-preview-14968--preliminary-istio.netlify.app/latest/docs/

I believe @linsun and @justinpettit had some thoughts on this during the call.

### DIFF
--- a/content/en/docs/ambient/_index.md
+++ b/content/en/docs/ambient/_index.md
@@ -1,7 +1,7 @@
 ---
 title: Ambient Mode
 description: Information for setting up and operating Istio in ambient mode.
-weight: 25
+weight: 17
 aliases:
   - /docs/ops/ambient
   - /latest/docs/ops/ambient
--- a/content/en/docs/ambient/_index.md
+++ b/content/en/docs/ambient/_index.md
@@ -1,5 +1,5 @@
 ---
-title: Ambient Mode (Sidecar-less)
+title: Ambient Mode
 description: Information for setting up and operating Istio in ambient mode.
 weight: 17
 aliases:
--- a/content/en/docs/ambient/_index.md
+++ b/content/en/docs/ambient/_index.md
@@ -1,5 +1,5 @@
 ---
-title: Ambient Mode
+title: Ambient Mode (Sidecar-less)
 description: Information for setting up and operating Istio in ambient mode.
 weight: 17
 aliases:
--- a/content/en/docs/setup/_index.md
+++ b/content/en/docs/setup/_index.md
@@ -1,6 +1,6 @@
 ---
-title: Setup
-description: Instructions for installing the Istio control plane on Kubernetes.
+title: Legacy Setup
+description: Instructions for installing the Istio control plane using sidecars.
 weight: 15
 aliases:
     - /docs/tasks/installing-istio.html
--- a/content/en/docs/setup/_index.md
+++ b/content/en/docs/setup/_index.md
@@ -1,6 +1,6 @@
 ---
-title: Legacy Setup
-description: Instructions for installing the Istio control plane using sidecars.
+title: Sidecar Mode
+description: Information for setting up and operating Istio in sidecar mode.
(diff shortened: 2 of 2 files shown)


dhawton:
Not sure how I like calling it Legacy before ambient is stable. Maybe sidecar setup?

justinpettit:
Overall, I like this placement better.

I agree with Daniel that "Legacy" sounds like we're planning to move away from sidecars.  I'd suggested using similar titles like "Sidecar Mode" and "Ambient Mode", so it's clearer that they're similar.  On the top level page of each of those sections, I think it would be good to have a little bit more description of what the mode is and maybe link to a page that describes why one would choose one mode over the other.  I'd be happy to help write that up if it would be helpful.

content/en/docs/setup/_index.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Setup
```
craigbox, line 2: +1 to "Sidecar Mode"

content/en/docs/setup/_index.md
```diff
@@ -1,6 +1,6 @@
...
-description: Instructions for installing the Istio control plane on Kubernetes.
+title: Legacy Setup
+description: Instructions for installing the Istio control plane using sidecars.
```
craigbox, line 3:
```suggestion
description: Information for setting up and operating Istio in sidecar mode.
```

content/en/docs/setup/_index.md
linsun: +1 Sidecar Setup

content/en/docs/ambient/_index.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Ambient Mode
 description: Information for setting up and operating Istio in ambient mode.
-weight: 25
+weight: 17
```
linsun, line 4: For Ambient Mode - Should we call it Ambient (Sidecar-less)?  Many folks don't know ambient is sidecar-less, why not make it clear for them.

linsun:
@ericvn thank you for starting this! I've commented my thoughts.

ericvn:
> and maybe link to a page that describes why one would choose one mode over the other. I'd be happy to help write that up if it would be helpful.

@justinpettit (or maybe we already have a good page to borrow from) I think this new comparison page would be good if you want to work on it. Maybe under `Concepts` but not sure. I'll see if I can come up with a blurb for the top of each page, unless you want to be able to tie that it with the comparison.

content/en/docs/ambient/_index.md
craigbox: I'm going to take this PR without that change, and then we can figure out if we're going to want to use that terminology throughout. and how.  (It will be more than just here, and it causes the heading to be uncomfortably long)

Screenshot 2024-04-30 at 2 31 46 PM

craigbox:
> > and maybe link to a page that describes why one would choose one mode over the other. I'd be happy to help write that up if it would be helpful.
> 
> @justinpettit (or maybe we already have a good page to borrow from) I think this new comparison page would be good if you want to work on it. Maybe under `Concepts` but not sure. I'll see if I can come up with a blurb for the top of each page, unless you want to be able to tie that it with the comparison.

I'm adding some new content in an "overview/introduction" section in a new PR, as per [the planning sheet](https://docs.google.com/spreadsheets/d/1ag5zhMOm2flEK7w5D7Sv3pbpHsx7afG2hs74-h2MHCY/edit#gid=0).  Justin, I'll tag you in!

craigbox (approved):

PR #14994, opened 2024-04-29
Refresh ambient architecture doc

linsun:
## Description

adding a section for ambient APIs, and refresh rest of docs.

## Reviewers

- [ x ] Ambient
- [ x ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,63 @@ owner: istio/wg-networking-maintainers
...
+
+### Labels
+
+You can use the following labels to enroll your namespace to ambient, enroll your resource to use a waypoint and specify the capture scope of your waypoint.
+
+|  Name  | Feature Status | Scope | Description |
+| --- | --- | --- | --- | --- |
+| `istio.io/dataplane-mode` | Beta | `Namespace` |  Specifies the data plane mode, valid value: `ambient`. |
+| `istio.io/use-waypoint` | Beta | `Namespace` or `Service` or `Pod` or `WorkloadEntry` or `ServiceEntry` | Enrolls your resource to use a given waypoint, valid value: `#none` or `{namespace}/{waypoint-name}` |
+| `istio.io/waypoint-for` | Alpha | `Gateway` | Specifies the waypoint's capture scope, valid value: `service` or `none` or `workload` or `all`. The default value is `service`. |
```
linsun, line 22: I didn't add GatewayClass here, I don't think it is implemented yet?  @ilrudie thoughts?

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,63 @@ owner: istio/wg-networking-maintainers
...
 1. **Waypoint enabled:** this is a pod that is "Captured" *and* has a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} deployed.
-  A waypoint will, by default, apply to all pods in the same namespace.
-  It can optionally be set to apply to only a specific service account with the `istio.io/for-service-account` annotation on the `Gateway`.
-  If there is both a namespace waypoint and service account waypoint, the service account waypoint takes precedence.
+  If a namespace is labeled with `istio.io/use-waypoint` with its default waypoint for the namespace, the waypoint will apply to all pods in the namespace.
+  The `istio.io/use-waypoint` label can optionally be set to apply to only a specific service or pod with its desired waypoint.
+  If the `istio.io/use-waypoint` label exists on the namespace and/or service and/or pod, the pod waypoint takes precedence over the service waypoint, which takes precedence over the namespace waypoint.
```
linsun, line 68: Is this correct? @ilrudie or @howardjohn

linsun:
Hi @louiscryan @howardjohn @ilrudie could you pls review and see if these update are correct? Thanks!

content/en/docs/ambient/architecture/index.md
```diff
@@ -33,9 +81,8 @@ However, depending on the destination's capabilities, different behavior will oc
 If the destination is also captured, or otherwise has Istio proxy capabilities (such as a sidecar), the request will be upgraded to an encrypted {{< gloss "HBONE" >}}HBONE tunnel{{< /gloss >}}.
 If the destination has a waypoint proxy, in addition to being upgraded to HBONE, the request will instead be forwarded to that waypoint.
 
-Note that in the case of a request to a `Service`, a specific endpoint will be selected to determine if it has a waypoint.
-However, if it *has* a waypoint, the request will be sent with a target destination of the `Service`, not the selected endpoint.
-This allows the waypoint to apply service-oriented policies to the traffic.
+Note that in the case of a request to a `Service`, if the service *has* a waypoint, the request will be sent to its waypoint to enforce service-oriented policies to the traffic.
```
MorrisLaw, line 84: @ilrudie is it true that if a service has a waypoint, the traffic will be sent to the waypoint if and only if the waypoint is capable of handling `service` or `all` traffic? If so, should we clarify that here?

Similarly, below, w.r.t. Pod/Workload traffic.

content/en/docs/ambient/architecture/index.md
```diff
@@ -33,9 +81,8 @@ However, depending on the destination's capabilities, different behavior will oc
...
+Similarly, in the case of a request to a `Pod` IP, if the pod *has* a waypoint, the request will be sent to its waypoint to enforce pod-oriented policies to the traffic.
```
MorrisLaw, line 85: Same as above comment. Especially since the waypoint capture scope defaults to service unless specified otherwise. I'm not sure that `use-waypoint` will work as intended if the waypoint only handles service traffic.

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,63 @@ owner: istio/wg-networking-maintainers
...
+to the waypoint named `waypoint` for the `default` namespace:
+
+{{< text yaml >}}
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: viewer
+  namespace: default
+spec:
+  targetRef:
```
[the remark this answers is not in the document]
linsun: updated

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,63 @@ owner: istio/wg-networking-maintainers
...
+    kind: Gateway
```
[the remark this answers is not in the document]
linsun: updated manually

content/en/docs/ambient/architecture/index.md
linsun: Hi @MorrisLaw I think what you said is true to my understanding. cc @howardjohn @stevenctl to weigh in as @ilrudie is going out of office for the next few days.

content/en/docs/ambient/architecture/index.md
linsun: Agreed, if the waypoint capture scope isn't workload, i'd think the label of use-waypoint on the pod won't be useful.

content/en/docs/ambient/architecture/index.md
linsun: My understanding is derived from the [waypoint capture doc](https://docs.google.com/document/d/1DAM_6SajCyC2hD8wBUraaZWZJqUVYPyemzw0-xpHmdY/edit):

```
The system assumes that any Gateway specified by istio.io/use-waypoint is capable of handling redirected traffic subject to this annotation and the binding specification. Since istio.io/use-waypoint plays a role in configuration generation too, it is useful to disable or more finely control redirection while allowing for testing of the generated config. Setting istio.io/waypoint-for: none we allow for this testing scenario.
```

craigbox:
We need to have a "technical overview of how ambient works" page.  It can be based on the blogs we've done. This page is at risk of being "things we haven't put anywhere else" - I thought the stuff about traffic routing should be on its own page.  But we can add things as they come in and I can reorganise them as we build it out!

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,63 @@ owner: istio/wg-networking-maintainers
 test: n/a
 ---
 
+## Ambient APIs
+
+### Labels
+
+You can use the following labels to enroll your namespace to ambient, enroll your resource to use a waypoint and specify, the capture scope of your waypoint.
+
+|  Name  | Feature Status | Scope | Description |
```
wilsonwu, line 18: The table format can be improved, suggest to use: http://markdowntable.com/ to format it.

linsun:
> We need to have a "technical overview of how ambient works" page. It can be based on the blogs we've done. This page is at risk of being "things we haven't put anywhere else" - I thought the stuff about traffic routing should be on its own page. But we can add things as they come in and I can reorganise them as we build it out!

Sounds good, thank you so much for offering to organize them so we can focus on getting the contents out first. :)

content/en/docs/ambient/architecture/index.md
linsun: I tried to use it, but i think it is harder to view or make change in the source view.  BTW, the table renders fine in preliminary, which is what users would see.

content/en/docs/ambient/architecture/index.md
ilrudie: Technically, wl and svc attachment are equal weight because they don't actually compete with each other. there is no need to know which overwrites the other. How the traffic is addressed determines which waypoint would be used. service addressed traffic will not use a wl-attached waypoint and likewise wl addressed traffic will not use a svc-attached waypoint.

linsun: The prob is what if my namespace says `use-waypoint: A`, and my service has `use-waypoint: B` and my workload has `use-waypoint: C` or if my service has `use-waypoint: #none` and my workload has `use-waypoint: C`?

content/en/docs/ambient/architecture/index.md
MorrisLaw: > if user tries to attach a waypoint that is svc-only to a workload that will be more or less a noop

good point

linsun: Yep that is what i have observed, which makes sense. basically, if I label my pod use-waypoint: C, the pod won't use waypoint C until waypoint C is configured for workload (or all).

content/en/docs/ambient/architecture/index.md
linsun: Updated to make this clear, PTAL @MorrisLaw !!

content/en/docs/ambient/architecture/index.md
linsun: Updated to make this clear, PTAL @MorrisLaw or @ilrudie  !!

linsun:
I removed SE and WE from the supported value of use-waypoint, given we don't claim to support VM or external service, I don't feel comfortable to claim them beta.

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+| `istio.io/use-waypoint` | Beta | `Namespace` or `Service` or `Pod` | Enrolls your resource to use a given waypoint, valid values: `{waypoint-name}`, `{namespace}/{waypoint-name}`, or `#none` |
+| `istio.io/waypoint-for` | Alpha | `Gateway` | Specifies the waypoint's capture scope, valid value: `service` or `none` or `workload` or `all`. This label is optional and the default value is `service`. |
+
+In order for your `istio.io/use-waypoint` label value to be effective, you have to ensure the used waypoint is configured with the corresponding capture scope which is `service` by default. For example, when you label a pod to use a specific waypoint via the `istio.io/use-waypoint` label, the waypoint's capture scope should be `workload` or `all`.
+
+### Policy attachment to waypoints
+
+You can attach Layer 7 policies (such as `AuthorizationPolicy`, `RequestAuthentication`, `Telemetry`, `WasmPlugin`, etc) to your waypoint using `targetRef`.
+
+- To attach the entire waypoint, set `Gateway` as the `targetRef` value. The example below shows how to attach the `viewer` policy
```
bleggett, line 30:
```suggestion
- To attach L7 policy to the entire waypoint, set `Gateway` as the `targetRef` value. The example below shows how to attach the `viewer` policy
```

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+      name: viewer
+      namespace: default
+    spec:
+      targetRefs:
+      - kind: Gateway
+        group: gateway.networking.k8s.io
+        name: waypoint
+    {{< /text >}}
+
+- To attach a given service within the waypoint, set `Service` as the `targetRef` value. The example below shows how to attach
```
bleggett, line 46:
```suggestion
- To attach L7 policy to a specific service within the waypoint, set `Service` as the `targetRef` value. The example below shows how to attach
```

linsun:
> I removed SE and WE from the supported value of use-waypoint, given we don't claim to support VM or external service, I don't feel comfortable to claim them beta.

tagging @craigbox and @zirain as we should document this in release notes:

`ServiceEntry` and `WorkloadEntry` for the `istio.io/use-waypoint` label support is under development and should land in a future release soon.

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+    spec:
+      targetRefs:
+      - kind: Service
+        group: ""
+        name: productpage
+    {{< /text >}}
+
 ## Traffic routing
 
 In {{< gloss "ambient" >}}ambient mode{{< /gloss >}}, workloads can fall into 3 categories:
```
bleggett, line 64: It feels odd to talk about traffic capture _after_ we talk about L7 policy enforcement via waypoints, which requires traffic capture to work.

Feels like this whole section needs to be at the top.

content/en/docs/ambient/architecture/index.md
```diff
@@ -48,18 +98,18 @@ By default, both will be accepted by ztunnel.
...
+Currently, traffic from sidecars and gateways won't go through any waypoint proxy either and they will be made aware of waypoint proxies
+in a future release.
 
 ### Waypoint routing
 
 A waypoint exclusively receives HBONE requests.
 Upon receiving a request, the waypoint will ensure it is targeting either a `Pod` that it manages or a `Service` that contains a `Pod` it manages.
 
-For either type of request, the waypoint will enforce policies (such as `AuthorizationPolicy`, `WasmPlugin`, `Telemetry`, etc) before forwarding.
+For either type of request, the waypoint will enforce policies (such as `AuthorizationPolicy`, `RequestAuthentication`, `WasmPlugin`, `Telemetry`, etc) before forwarding.
```
bleggett, line 112:
```suggestion
For either type of request, the waypoint will enforce L7 policies (such as `AuthorizationPolicy`, `RequestAuthentication`, `WasmPlugin`, `Telemetry`, etc) before forwarding.
```

content/en/docs/ambient/architecture/index.md
```diff
@@ -48,18 +98,18 @@ By default, both will be accepted by ztunnel.
...
-When these direct calls are made, the ztunnel will "hairpin" the request to its own waypoint to ensure policies are properly enforced.
+When the destination is waypoint enabled, if the source is `captured` by its ztunnel, the ztunnel ensures the request **must** go through
+the waypoint where policy is enforced. However, a workload outside of the mesh doesn't know anything about waypoint proxies therefore it sends
+requests directly to the destination without going through any waypoint proxy even if the destination is waypoint enabled.
+Currently, traffic from sidecars and gateways won't go through any waypoint proxy either and they will be made aware of waypoint proxies
+in a future release.
 
 ### Waypoint routing
 
 A waypoint exclusively receives HBONE requests.
```
bleggett, line 109:
```suggestion
A waypoint exclusively receives HBONE requests.
```

TODO this should link back to a HBONE doc

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+        name: productpage
+    {{< /text >}}
+
 ## Traffic routing
 
 In {{< gloss "ambient" >}}ambient mode{{< /gloss >}}, workloads can fall into 3 categories:
 1. **Uncaptured:** this is a standard pod without any mesh features enabled.
-1. **Captured:** this is a pod that has traffic intercepted by {{< gloss >}}ztunnel{{< /gloss >}}. Pods can be captured by setting the `istio.io/dataplane-mode=ambient` label on a namespace.
+1. **Captured:** this is a pod that has traffic intercepted by {{< gloss >}}ztunnel{{< /gloss >}}. A pod's captured mode can be enabled by setting the `istio.io/dataplane-mode=ambient` label on its namespace, which enables all pods' captured mode for that namespace.
 1. **Waypoint enabled:** this is a pod that is "Captured" *and* has a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} deployed.
```
bleggett, line 67:
```suggestion
1. **Waypoint enabled:** this is a pod that is "Captured" *and* has a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} deployed. With a waypoint, L7 policy can be enforced for pod traffic. Waypoints are enabled for a workload by setting the `istio.io/use-waypoint` label.
```

content/en/docs/ambient/architecture/index.md
MorrisLaw: Lgtm

content/en/docs/ambient/architecture/index.md
ilrudie: Correct, gw class isn't checked yet :(

linsun: Thanks both, marking as resolved

content/en/docs/ambient/architecture/index.md
linsun: Cool, marking as resolved

content/en/docs/ambient/architecture/index.md
linsun: Thanks, marking as resolved

content/en/docs/ambient/architecture/index.md
linsun: Oops, I am so sorry!!! @MorrisLaw

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+| --- | --- | --- | --- | --- |
+| `istio.io/dataplane-mode` | Beta | `Namespace` |  Specifies the data plane mode, valid value: `ambient`. |
+| `istio.io/use-waypoint` | Beta | `Namespace` or `Service` or `Pod` | Enrolls your resource to use a given waypoint, valid values: `{waypoint-name}`, `{namespace}/{waypoint-name}`, or `#none` |
+| `istio.io/waypoint-for` | Alpha | `Gateway` | Specifies the waypoint's capture scope, valid value: `service` or `none` or `workload` or `all`. This label is optional and the default value is `service`. |
+
+In order for your `istio.io/use-waypoint` label value to be effective, you have to ensure the used waypoint is configured with the corresponding capture scope which is `service` by default. For example, when you label a pod to use a specific waypoint via the `istio.io/use-waypoint` label, the waypoint's capture scope should be `workload` or `all`.
+
+### Policy attachment to waypoints
+
+You can attach Layer 7 policies (such as `AuthorizationPolicy`, `RequestAuthentication`, `Telemetry`, `WasmPlugin`, etc) to your waypoint using `targetRef`.
```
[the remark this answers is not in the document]
linsun: I prefer write it starting with a person, which is the reader :)

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
 test: n/a
 ---
 
+## Ambient APIs
+
+### Labels
+
+You can use the following labels to enroll your namespace to ambient, enroll your resource to use a waypoint and specify, the capture scope of your waypoint.
```
louiscryan, line 16: Suggestion for consistent terms.....

"add your [namespace | service | workload | pod ...] to the mesh"
"enable a waypoint for traffic to your [namespace | service | workload | pod ...] "
"control what traffic is sent to the waypoint"

I'd avoid the use of the term 'enroll' as it's harder to conjugate with 'to' making statements harder to form.

content/en/docs/ambient/architecture/index.md
linsun: noted when that doc is avail :)

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+
+|  Name  | Feature Status | Scope | Description |
+| --- | --- | --- | --- | --- |
+| `istio.io/dataplane-mode` | Beta | `Namespace` |  Specifies the data plane mode, valid value: `ambient`. |
+| `istio.io/use-waypoint` | Beta | `Namespace` or `Service` or `Pod` | Enrolls your resource to use a given waypoint, valid values: `{waypoint-name}`, `{namespace}/{waypoint-name}`, or `#none` |
```
louiscryan: replace "Enrolls your resource to use a given waypoint, valid values"

"Use a waypoint for traffic to the labelled resource, valid values"

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+| `istio.io/waypoint-for` | Alpha | `Gateway` | Specifies the waypoint's capture scope, valid value: `service` or `none` or `workload` or `all`. This label is optional and the default value is `service`. |
+
+In order for your `istio.io/use-waypoint` label value to be effective, you have to ensure the used waypoint is configured with the corresponding capture scope which is `service` by default. For example, when you label a pod to use a specific waypoint via the `istio.io/use-waypoint` label, the waypoint's capture scope should be `workload` or `all`.
+
+### Policy attachment to waypoints
```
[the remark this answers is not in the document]
linsun: I think attachment to waypoints is important here, I'll add L7

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+
+### Labels
+
+You can use the following labels to enroll your namespace to ambient, enroll your resource to use a waypoint and specify, the capture scope of your waypoint.
+
+|  Name  | Feature Status | Scope | Description |
+| --- | --- | --- | --- | --- |
+| `istio.io/dataplane-mode` | Beta | `Namespace` |  Specifies the data plane mode, valid value: `ambient`. |
+| `istio.io/use-waypoint` | Beta | `Namespace` or `Service` or `Pod` | Enrolls your resource to use a given waypoint, valid values: `{waypoint-name}`, `{namespace}/{waypoint-name}`, or `#none` |
+| `istio.io/waypoint-for` | Alpha | `Gateway` | Specifies the waypoint's capture scope, valid value: `service` or `none` or `workload` or `all`. This label is optional and the default value is `service`. |
```
louiscryan, line 22: If we want to define a reusable term like 'capture scope' when we need to be a bit more verbose

"Specifies the waypoint's 'capture scope', the class of addressed endpoints it is willing to accept, valid values: `service` or `none` or `workload` or `all`. This label is optional and the default value is `service`. |

or we can avoid trying to define the term and rely on the natural use of 'for' to explain

Specifies what types of endpoints the  waypoint will process traffic for, valid values: `service` or `none` or `workload` or `all`. This label is optional and the default value is `service`. |

content/en/docs/ambient/architecture/index.md
linsun: I don't think it is correct to jump in to discuss pod/workload so fast given it is not enabled by default and is not the most common scenario.  Most common scenario is to enable waypoint for service.

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+
+In order for your `istio.io/use-waypoint` label value to be effective, you have to ensure the used waypoint is configured with the corresponding capture scope which is `service` by default. For example, when you label a pod to use a specific waypoint via the `istio.io/use-waypoint` label, the waypoint's capture scope should be `workload` or `all`.
```
louiscryan, line 24: If avoiding the term 'capture scope' ...

In order for your `istio.io/use-waypoint` label value to be effective, you have to ensure  waypoint is configured for the endpoint which is using it. By default waypoints accept traffic for `service` endpoints. For example, when you label a pod to use a specific waypoint via the `istio.io/use-waypoint` label, the waypoint should be labelled  `istio.io./waypoint-for' with the values `workload` or `all`.

content/en/docs/ambient/architecture/index.md
louiscryan: s/To attach the entire waypoint/To attach to the entire waypoint

content/en/docs/ambient/architecture/index.md
linsun: Maybe :) @craigbox said he would reorganize things once contents are in.  It is possible we may want Ambient APIs to have its own page.  I'll defer this organization as a follow up.

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+      targetRefs:
+      - kind: Service
+        group: ""
+        name: productpage
+    {{< /text >}}
+
 ## Traffic routing
 
 In {{< gloss "ambient" >}}ambient mode{{< /gloss >}}, workloads can fall into 3 categories:
 1. **Uncaptured:** this is a standard pod without any mesh features enabled.
```
louiscryan, line 65: I worry 'capture' is a little too technical here.  We could just as easily say

"Out of Mesh"
"In Mesh"
"Waypoint Enabled"

content/en/docs/ambient/architecture/index.md
bleggett: Yeah the last sentence is not meant to refer to workload-attached waypoints directly, but is just intended to be a generic lead into "here are the ways in which waypoints can be enabled for _your workloads_"

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+1. **Captured:** this is a pod that has traffic intercepted by {{< gloss >}}ztunnel{{< /gloss >}}. A pod's captured mode can be enabled by setting the `istio.io/dataplane-mode=ambient` label on its namespace, which enables all pods' captured mode for that namespace.
 1. **Waypoint enabled:** this is a pod that is "Captured" *and* has a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} deployed.
-  A waypoint will, by default, apply to all pods in the same namespace.
-  It can optionally be set to apply to only a specific service account with the `istio.io/for-service-account` annotation on the `Gateway`.
-  If there is both a namespace waypoint and service account waypoint, the service account waypoint takes precedence.
+  If a namespace is labeled with `istio.io/use-waypoint` with its default waypoint for the namespace, the waypoint will apply to all pods in the namespace.
+  The `istio.io/use-waypoint` label can optionally be set to apply to only a specific service or pod with its desired waypoint.
```
louiscryan: The `istio.io/use-waypoint` label can also be set on individual services or pods when using a waypoint for the entire namespace is not desired.

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+  If the `istio.io/use-waypoint` label exists on the namespace and service, the service waypoint takes precedence over the namespace waypoint as long as the service waypoint can capture `service` or `all` traffic.
```
louiscryan: If the `istio.io/use-waypoint` label exists on both a namespace and a service, the service waypoint takes precedence over the namespace waypoint as long as the service waypoint can handle `service` or `all` traffic.

content/en/docs/ambient/architecture/index.md
bleggett: Yes this is a TODO for me

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,15 +9,66 @@ owner: istio/wg-networking-maintainers
...
+  If the `istio.io/use-waypoint` label exists on the namespace and pod, the pod waypoint takes precedence over the namespace waypoint as long as the pod waypoint can capture `workload` or `all` traffic.
```
louiscryan: Can elide if merging above comment

content/en/docs/ambient/architecture/index.md
```diff
@@ -33,9 +84,8 @@ However, depending on the destination's capabilities, different behavior will oc
 If the destination is also captured, or otherwise has Istio proxy capabilities (such as a sidecar), the request will be upgraded to an encrypted {{< gloss "HBONE" >}}HBONE tunnel{{< /gloss >}}.
 If the destination has a waypoint proxy, in addition to being upgraded to HBONE, the request will instead be forwarded to that waypoint.
```
louiscryan: s/instead/also

content/en/docs/ambient/architecture/index.md
```diff
@@ -33,9 +84,8 @@ However, depending on the destination's capabilities, different behavior will oc
...
 
-Note that in the case of a request to a `Service`, a specific endpoint will be selected to determine if it has a waypoint.
-However, if it *has* a waypoint, the request will be sent with a target destination of the `Service`, not the selected endpoint.
-This allows the waypoint to apply service-oriented policies to the traffic.
+Note that in the case of a request to a `Service`, if the service *has* a waypoint and the waypoint can capture `service` or `all` traffic, the request will be sent to its waypoint to enforce service-oriented policies to the traffic.
```
louiscryan, line 87: I don't think we need to keep qualifying that the waypoint is correctly configured to habdle 'service' or 'all' traffic. We've covered it above so we can elide this here.


Also I would s/service-oriented policies/L7 policies  here and elsewhere

Same for 'pod oriented' below, I think these terms are likely to confuse

content/en/docs/ambient/architecture/index.md
```diff
@@ -33,9 +84,8 @@ However, depending on the destination's capabilities, different behavior will oc
...
+Similarly, in the case of a request to a `Pod` IP, if the pod *has* a waypoint and the waypoint can capture `workload` or `all` traffic, the request will be sent to its waypoint to enforce pod-oriented policies to the traffic.
 In the rare case that a `Service` has a mix of waypoint enabled and non-enabled endpoints, some requests would be sent to a waypoint while other requests to the same service would not.
```
louiscryan, line 89: This is incorrect, a 'Service' either has a waypoint or it does not atomically

I think you may be conflating the definition of K8S Service with the common parlance for a bunch of related worklaods being a 'service'. This case can only happen for 'pods' and to-workload traffic but it's pretty unlikely. If we were going to say anything it would be...

"Since it is possible to vary the labels associated with pods in a Deployment it is technically possible for some pods to *use* a waypoint while others do not. Users are generally recommended to avoid this advanced use-case:"

content/en/docs/ambient/architecture/index.md
```diff
@@ -33,9 +84,8 @@ However, depending on the destination's capabilities, different behavior will oc
 If the destination is also captured, or otherwise has Istio proxy capabilities (such as a sidecar), the request will be upgraded to an encrypted {{< gloss "HBONE" >}}HBONE tunnel{{< /gloss >}}.
```
louiscryan, line 84: I think we should swap HBONE with TLS here, we can have a separate doc on write protocol, TLS & HBONE but I don't think we're trying to sell 'HBONE" here

content/en/docs/ambient/architecture/index.md
```diff
@@ -48,18 +98,18 @@ By default, both will be accepted by ztunnel.
 Because plaintext requests will have no peer identity when Authorization Policies are evaluated,
```
louiscryan, line 98: This feels like an abrupt statement and its odd to lead with an edge case rather than the more common stuff which you cover below.

We should contextualize 'plaintext' as 'traffic from sources not yet in the mesh' to be consistent with our terminology earlier.

content/en/docs/ambient/architecture/index.md
linsun: cc @MorrisLaw I had enroll earlier because we have the `istioctl x waypoint --enroll-namespace` cmd. thoughts?  `enable` does sound easier to understand.

content/en/docs/ambient/architecture/index.md
```diff
@@ -48,18 +98,18 @@ By default, both will be accepted by ztunnel.
...
+When the destination is waypoint enabled, if the source is `captured` by its ztunnel, the ztunnel ensures the request **must** go through
+the waypoint where policy is enforced. However, a workload outside of the mesh doesn't know anything about waypoint proxies therefore it sends
+requests directly to the destination without going through any waypoint proxy even if the destination is waypoint enabled.
+Currently, traffic from sidecars and gateways won't go through any waypoint proxy either and they will be made aware of waypoint proxies
+in a future release.
 
 ### Waypoint routing
 
 A waypoint exclusively receives HBONE requests.
 Upon receiving a request, the waypoint will ensure it is targeting either a `Pod` that it manages or a `Service` that contains a `Pod` it manages.
```
louiscryan, line 110: Upon receiving a request, the waypoint will ensure that the traffic is for a `Pod` or `Service`  which *use* it.

content/en/docs/ambient/architecture/index.md
louiscryan: Not sure what is meant by 'type' of request. I think you covered this in the previous sentence so you could just skip. Alternatively can say something like

"Having accepted the traffic the waypoint will then enforce policies (such as `AuthorizationPolicy`, `RequestAuthentication`, `WasmPlugin`, `Telemetry`, etc) before forwarding.

content/en/docs/ambient/architecture/index.md
bleggett: https://github.com/istio/istio.io/pull/15004 - I'll fix that after this merges.

content/en/docs/ambient/architecture/index.md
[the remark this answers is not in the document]
linsun: updated!

content/en/docs/ambient/architecture/index.md
linsun: updated

content/en/docs/ambient/architecture/index.md
linsun: updated!

content/en/docs/ambient/architecture/index.md
linsun: updated, took your 2nd suggestion :)

content/en/docs/ambient/architecture/index.md
linsun: agreed, updated!

content/en/docs/ambient/architecture/index.md
linsun: updated

content/en/docs/ambient/architecture/index.md
linsun: updated

content/en/docs/ambient/architecture/index.md
linsun: like it!

content/en/docs/ambient/architecture/index.md
linsun: Updated

content/en/docs/ambient/architecture/index.md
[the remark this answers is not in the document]
linsun: updated

content/en/docs/ambient/architecture/index.md
linsun: updated!

content/en/docs/ambient/architecture/index.md
linsun: updated

content/en/docs/ambient/architecture/index.md
linsun: updated

content/en/docs/ambient/architecture/index.md
linsun: Defer to this change as we currently have a glossary for hbone but nothing for TLS.

content/en/docs/ambient/architecture/index.md
linsun: Yes I was thinking about the service is labelled use-waypoint and some of the workloads belong to the service are labelled as use a different waypoint or use 0 waypoint.  I think this matters in the old design but not in the current design where it relies on traffic to service or workload to decide if waypoint enabled.

accepted your suggestions!

content/en/docs/ambient/architecture/index.md
linsun: Updated the `plaintext` wording. Not exactly sure how to fix the abrupt statement seems okay for me....

content/en/docs/ambient/architecture/index.md
linsun: updated!

content/en/docs/ambient/architecture/index.md
linsun: Updated

content/en/docs/ambient/architecture/index.md
linsun: Could update when https://github.com/istio/istio.io/pull/15004/files is merged from @bleggett

content/en/docs/ambient/architecture/index.md
linsun: Thanks for the clarification. I reworded it to not focus on pod:

```
To enforce L7 policies, add the `istio.io/use-waypoint` label to your resource to use waypoint for the labeled resource.
```

content/en/docs/ambient/architecture/index.md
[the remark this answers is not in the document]
linsun: SG, cc @craigbox defer this as a follow up!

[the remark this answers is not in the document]
linsun: opened https://github.com/istio/istio.io/issues/15005 to track this

linsun:
Hi @ericvn could you take a look? All comments have addressed and @bleggett is also waiting on this for his https://github.com/istio/istio.io/pull/15004/files

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,57 +9,109 @@ owner: istio/wg-networking-maintainers
...
+If the destination has a waypoint proxy, in addition to being upgraded to HBONE, the request will also be forwarded to that waypoint for L7 policy enforcement.
 
-Note that in the case of a request to a `Service`, a specific endpoint will be selected to determine if it has a waypoint.
-However, if it *has* a waypoint, the request will be sent with a target destination of the `Service`, not the selected endpoint.
-This allows the waypoint to apply service-oriented policies to the traffic.
-In the rare case that a `Service` has a mix of waypoint enabled and non-enabled endpoints, some requests would be sent to a waypoint while other requests to the same service would not.
+Note that in the case of a request to a `Service`, if the service *has* a waypoint, the request will be sent to its waypoint to enforce L7 policies to the traffic.
+Similarly, in the case of a request to a `Pod` IP, if the pod *has* a waypoint, the request will be sent to its waypoint to enforce L7 policies to the traffic.
```
linsun, line 89: @bleggett i think you had a comment earlier to clarify if the use-waypoint label for both service and pod, what happens... It is explained here imho as it is for outbound, not inbound.

linsun:
Hi @bleggett can you take another review and approve if all good?

content/en/docs/ambient/architecture/index.md
MorrisLaw: We can call the `--enroll-namespace` command to enable a waypoint for traffic to your namespace or use the `use-waypoint` annotation to enable a waypoint for traffic to a [service | workload | pod ...]. That kind of wording makes sense to me here for readability and better understanding. I like it, good call!

content/en/docs/ambient/architecture/index.md
bleggett: Yep fair enough - we could make it a bit more explicit but this will do.

bleggett (approved):

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,57 +9,109 @@ owner: istio/wg-networking-maintainers
...
+      name: viewer
+      namespace: default
+    spec:
+      targetRefs:
+      - kind: Gateway
+        group: gateway.networking.k8s.io
+        name: waypoint
+    {{< /text >}}
+
+- To attach L7 policy to a specific service within the waypoint, set `Service` as the `targetRefs` value. The example below shows how to attach
```
ericvn, line 46:
```suggestion
- To attach a L7 policy to a specific service within the waypoint, set `Service` as the `targetRefs` value. The example below shows how to attach
```

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,57 +9,109 @@ owner: istio/wg-networking-maintainers
...
-1. **Waypoint enabled:** this is a pod that is "Captured" *and* has a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} deployed.
-  A waypoint will, by default, apply to all pods in the same namespace.
-  It can optionally be set to apply to only a specific service account with the `istio.io/for-service-account` annotation on the `Gateway`.
-  If there is both a namespace waypoint and service account waypoint, the service account waypoint takes precedence.
+1. **Out of Mesh:** this is a standard pod without any mesh features enabled.
+1. **In Mesh:** this is a pod that has traffic intercepted at the Layer 4 level by {{< gloss >}}ztunnel{{< /gloss >}}. In this mode, L4 policy can be enforced for pod traffic. This mode can be enabled for a pod by setting the `istio.io/dataplane-mode=ambient` label on the pod's namespace. This will enable *in mesh* mode for all pods in that namespace.
+1. **Waypoint enabled:** this is a pod that is "In Mesh" *and* has a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} deployed. To enforce L7 policies, add the `istio.io/use-waypoint` label to your resource to use waypoint for the labeled resource.
```
ericvn, line 67: You didn't uppercase in the prior line. Lower case here for consistency:
```suggestion
1. **Waypoint enabled:** this is a pod that is "in mesh" *and* has a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} deployed. To enforce L7 policies, add the `istio.io/use-waypoint` label to your resource to use waypoint for the labeled resource.
```

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,57 +9,109 @@ owner: istio/wg-networking-maintainers
...
-1. **Uncaptured:** this is a standard pod without any mesh features enabled.
-1. **Captured:** this is a pod that has traffic intercepted by {{< gloss >}}ztunnel{{< /gloss >}}. Pods can be captured by setting the `istio.io/dataplane-mode=ambient` label on a namespace.
-1. **Waypoint enabled:** this is a pod that is "Captured" *and* has a {{< gloss "waypoint" >}}waypoint proxy{{< /gloss >}} deployed.
-  A waypoint will, by default, apply to all pods in the same namespace.
-  It can optionally be set to apply to only a specific service account with the `istio.io/for-service-account` annotation on the `Gateway`.
-  If there is both a namespace waypoint and service account waypoint, the service account waypoint takes precedence.
+1. **Out of Mesh:** this is a standard pod without any mesh features enabled.
+1. **In Mesh:** this is a pod that has traffic intercepted at the Layer 4 level by {{< gloss >}}ztunnel{{< /gloss >}}. In this mode, L4 policy can be enforced for pod traffic. This mode can be enabled for a pod by setting the `istio.io/dataplane-mode=ambient` label on the pod's namespace. This will enable *in mesh* mode for all pods in that namespace.
```
ericvn, line 66:
```suggestion
1. **In Mesh:** this is a pod that has traffic intercepted at the Layer 4 level by {{< gloss >}}ztunnel{{< /gloss >}}. In this mode, L4 policies can be enforced for pod traffic. This mode can be enabled for a pod by setting the `istio.io/dataplane-mode=ambient` label on the pod's namespace. This will enable *in mesh* mode for all pods in that namespace.
```

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,57 +9,109 @@ owner: istio/wg-networking-maintainers
...
+  Similarly, a label on a pod will take precedence over a namespace label.
 
 Depending on which category a workload is in, the request path will be different.
 
 ### Ztunnel routing
 
 #### Outbound
 
-When a captured pod makes an outbound request, it will be transparently redirected to ztunnel which will determine where and how to forward the request.
+When a pod in ambient mesh makes an outbound request, it will be transparently redirected to ztunnel which will determine where and how to forward the request.
```
ericvn, line 80: I'd say either a `pod in ambient mode` or `pod in an ambient mesh`. I just picked one
```suggestion
When a pod in an ambient mesh makes an outbound request, it will be transparently redirected to ztunnel which will determine where and how to forward the request.
```

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,57 +9,109 @@ owner: istio/wg-networking-maintainers
...
-In the rare case that a `Service` has a mix of waypoint enabled and non-enabled endpoints, some requests would be sent to a waypoint while other requests to the same service would not.
+Note that in the case of a request to a `Service`, if the service *has* a waypoint, the request will be sent to its waypoint to enforce L7 policies to the traffic.
+Similarly, in the case of a request to a `Pod` IP, if the pod *has* a waypoint, the request will be sent to its waypoint to enforce L7 policies to the traffic.
+Since it is possible to vary the labels associated with pods in a `Deployment` it is technically possible for
+some pods to use a waypoint while others do not. Users are generally recommended to avoid this advanced use-case.
 
 #### Inbound
 
-When a captured pod receives an inbound request, it will be transparently redirected to ztunnel.
+When a pod in ambient mesh receives an inbound request, it will be transparently redirected to ztunnel.
```
ericvn, line 95:
```suggestion
When a pod in an ambient mesh receives an inbound request, it will be transparently redirected to ztunnel.
```

ericvn (commented):
A few minor comments.

content/en/docs/ambient/architecture/index.md
```diff
@@ -9,57 +9,109 @@ owner: istio/wg-networking-maintainers
...
+
+### Labels
+
+You can use the following labels to add your resource to the mesh, use a waypoint for traffic to your resource, and control what traffic is sent to the waypoint.
+
+|  Name  | Feature Status | Resource | Description |
+| --- | --- | --- | --- | --- |
+| `istio.io/dataplane-mode` | Beta | `Namespace` |  Add your resource to the mesh, valid value: `ambient`. |
+| `istio.io/use-waypoint` | Beta | `Namespace` or `Service` or `Pod` | Use a waypoint for traffic to the labeled resource for L7 policy enforcement, valid values: `{waypoint-name}`, `{namespace}/{waypoint-name}`, or `#none` |
+| `istio.io/waypoint-for` | Alpha | `Gateway` | Specifies what types of endpoints the waypoint will process traffic for, valid value: `service` or `none` or `workload` or `all`. This label is optional and the default value is `service`. |
```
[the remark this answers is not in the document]
linsun: I've considered it but thought it may be harder to edit things because some columns have lots of texts!

ericvn (approved):

Issue #14983, opened 2024-04-26
Refresh Ztunnel docs for Istio 1.22 Ambient Beta release


MorrisLaw:
In preparation for the Ambient beta release in Istio 1.22, we'll need to ensure docs are up to date. This ticket addresses Ztunnel docs:

- [ ] Ensure that current Ztunnel documentation is refreshed/up-to-date:
  - [ ] [docs/ambient/usage/ztunnel](https://preliminary.istio.io/latest/docs/ambient/usage/ztunnel/)
  - [ ] [docs/ops/ambient/usage/traffic-redirection](https://istio.io/latest/docs/ops/ambient/usage/traffic-redirection/)
  - [ ] any other relevant Ztunnel docs [here](https://istio.io/latest/search/?q=ztunnel&site=docs)

bleggett:
Would advocate that `docs/ops/ambient/usage/traffic-redirection` go under an architecture section, it is 100% an architectural explainer and 0% a usage guide.

bleggett:
Would additionally advocate that we take the `HBONE` section out of `docs/ops/ambient/usage/traffic-redirection` and spin it out as the start of

bleggett:
> Would additionally advocate that we take the `HBONE` section out of `docs/ambient/usage/ztunnel/` and spin it out as the start of an "HBONE" doc
> 
> #13482
> 
> (I may do this unless you want/intend to)

https://github.com/istio/istio.io/pull/15004 all right, there. 

Shouldn't impact you for this issue much, except it removes the HBONE section from the ztunnel doc.

linsun:
A couple of things I'd like to see us clean up on the ztunnel page(https://preliminary.istio.io/latest/docs/ambient/usage/ztunnel/):

- remove alpha warning
- remove caveats - should be clear in release blog/note but no need to emphasize it in each doc.
- CP/DP overview - should move to architecture overview page.
- we should use the new istioctl x ztunnel-config cmd instead proxy-config imho.
- https://preliminary.istio.io/latest/docs/ambient/usage/ztunnel/#pod-selection-logic-for-ambient-and-sidecar-modes we don't support sidecar co-exist with sidecarless, worth update it to be clear.
[...]

PR #14995, opened 2024-04-29
Remove ztunnel hairpinning from docs

MorrisLaw:
## Description
Related to #14983 (this is one of a few PRs that will be made to refresh the docs in an effort to ensure continuous progress is made as opposed to holding everything up on one big documentation PR)

Removes ztunnel hairpinning from docs since it has now been removed from the codebase. See [Ztunnel PR #913](https://github.com/istio/ztunnel/pull/913).

## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -475,7 +457,7 @@ In the use cases so far, the traffic source and destination pods are both ambien
...
+In the example below, the same `httpbin` service which has already been set up in the prior examples is accessed via client `sleep` pods that are running in a separate namespace that is not part of the Istio mesh. This example shows that East-west traffic between ambient mesh pods and non mesh pods is seamlessly supported. Since the non-mesh pods initiate traffic directly to the backend pods without going through HBONE or ztunnel. Note that in this case, you have control over whether or not the waypoint gets bypassed too depending on how you configure policy.
```
linsun, line 460: I don't think user can control whether waypoint gets bypassed.  The traffic won't be sent to waypoint if it is from out of mesh.

linsun (commented):
LGTM on removal of hairpin.

Made a comment on your clarification for out of mesh client which I think is wrong.

content/en/docs/ambient/usage/ztunnel/index.md
ilrudie: +1, I think it's just a wording/clarity issue though. You can control if bypass is allowed or denied using istio L4 policy but the way it is worded seems like it could mean that you can define a policy that makes the traffic hairpin to the waypoint which we do not offer.

content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -457,7 +457,7 @@ In the use cases so far, the traffic source and destination pods are both ambien
...
 
-In the example below, the same `httpbin` service which has already been set up in the prior examples is accessed via client `sleep` pods that are running in a separate namespace that is not part of the Istio mesh. This example shows that East-west traffic between ambient mesh pods and non mesh pods is seamlessly supported. Since the non-mesh pods initiate traffic directly to the backend pods without going through HBONE or ztunnel. Note that in this case, you have control over whether or not the waypoint gets bypassed too depending on how you configure policy.
+In the example below, the same `httpbin` service which has already been set up in the prior examples is accessed via client `sleep` pods that are running in a separate namespace that is not part of the Istio mesh. This example shows that East-west traffic between ambient mesh pods and non mesh pods is seamlessly supported. Since the non-mesh pods initiate traffic directly to the backend pods without going through HBONE or ztunnel. Note that in this case, Istio L4 policy can control whether traffic is denied or allowed to bypass the waypoint.
```
MorrisLaw, line 460: Does this read better @ilrudie @linsun ?

content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -475,7 +457,7 @@ In the use cases so far, the traffic source and destination pods are both ambien
...
```
ilrudie, line 460: Yeah, seems clearer to me. I'd remove the preceding sentence at this point too though.
```suggestion
In the example below, the same `httpbin` service which has already been set up in the prior examples is accessed via client `sleep` pods that are running in a separate namespace that is not part of the Istio mesh. This example shows that East-west traffic between ambient mesh pods and non mesh pods is seamlessly supported. Note that in this case, Istio L4 policy can control whether traffic is denied or allowed to bypass the waypoint.
```

linsun: Hi @MorrisLaw @ilrudie I think I'm confused with `Istio L4 policy can control whether traffic is denied or allowed to bypass the waypoint.`  My mental model is this is controlled by source ztunnel based on the source (out of mesh or ingress or sidecar or ztunnel captured pod).  You have L4 policy which is enforced at destination ztunnel thus it is confusing to me that it controls bypass waypoint. I think it should be:

Note that in this case, Istio L4 policy can control whether traffic is denied or allowed from the source (which is sleep here I think?).

content/en/docs/ambient/usage/ztunnel/index.md
linsun: answered a bit below.

content/en/docs/ambient/usage/ztunnel/index.md
MorrisLaw: Is this policy part, specifically, documented somewhere @linsun ?

>  Istio L4 policy can control whether traffic is denied or allowed from the source

I'm getting a bit confused myself now. I think a reference might be helpful.

content/en/docs/ambient/usage/ztunnel/index.md
louiscryan: nits:

- no need to say 'Istio' 
- s/East-west/east-west
- s/through HBONE or ztunnel/through ztunnel  (we can elaborate on HBONE elsewhere)

content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -111,24 +111,6 @@ link="ztunnel-waypoint-datapath.png"
...
-
-{{< image width="100%"
-link="ztunnel-hairpin.png"
-caption="Ztunnel traffic hair-pinning"
->}}
-
 ### Note on HBONE {#hbonesection}
 
 HBONE (HTTP Based Overlay Network Encapsulation) is an Istio-specific term. It refers to the use of standard HTTP tunneling via the [HTTP CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) method to transparently tunnel application packets/ byte streams. In its current implementation within Istio, it transports TCP packets by tunneling these transparently using the HTTP CONNECT method, uses [HTTP/2](https://httpwg.org/specs/rfc7540.html), with encryption and mutual authentication provided by [mutual TLS](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/) and the HBONE tunnel itself runs on TCP port 15008. The overall HBONE packet format from IP layer onwards is depicted in the following figure.
```
louiscryan, line 116: HBONE (HTTP Based Overlay Network Encapsulation) is an Istio-specific term. It refers to the use of [mutual TLS](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/) and HTTP tunnels via the [HTTP CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) method to transparently carry TCP byte streams. In its current implementation within Istio, this is done with [HTTP/2](https://httpwg.org/specs/rfc7540.html) to enable efficient multiplexing and TLS connection reuse. HBONE is served on the reserved TCP port 15008. The overall HBONE packet format from IP layer onwards is depicted in the following figure.

linsun: cc @bleggett who is looking at hbone doc.

content/en/docs/ambient/usage/ztunnel/index.md
MorrisLaw: That makes sense, thank you for clarifying!

content/en/docs/ambient/usage/ztunnel/index.md
MorrisLaw: Updated, PTAL @louiscryan @bleggett

content/en/docs/ambient/usage/ztunnel/index.md
MorrisLaw: Updated, PTAL @linsun

content/en/docs/ambient/usage/ztunnel/index.md
louiscryan: Thanks but I think we can now just update to refer to https://github.com/istio/istio.io/pull/15004

content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -475,7 +457,7 @@ In the use cases so far, the traffic source and destination pods are both ambien
...
+In the example below, the same `httpbin` service which has already been set up in the prior examples is accessed via client `sleep` pods that are running in a separate namespace that is not part of the Istio mesh. This example shows that East-west traffic between ambient mesh pods and non mesh pods is seamlessly supported. Note that in this case, Istio L4 policy can control whether traffic is denied or allowed from the source.
```
linsun, line 460: Hi @MorrisLaw LGTM.  You may want to address @louiscryan 's comments - https://github.com/istio/istio.io/pull/14995#discussion_r1585317353

content/en/docs/ambient/usage/ztunnel/index.md
bleggett: (my bad I missed this) - yeah just leave this as is and the other PR will remove this section entirely

content/en/docs/ambient/usage/ztunnel/index.md
ilrudie: Source ztunnel does control if traffic is sent to a waypoint or not but it's not via an authz policy. The destination ztunnel can enforce things like "this wl can only be reached from my waypoint's identity" or conversely it can be permissive of traffic without an identity or it may allow an identity of a non-waypoint wl to reach the destination.

I was describing that as allowing the waypoint to be "bypassed". I can see how that's a confusing way to think about/describe what's happening though.

MorrisLaw:
@bleggett @linsun @louiscryan I think all comments have been addressed now. PTAL again, thank you!

content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -475,7 +457,7 @@ In the use cases so far, the traffic source and destination pods are both ambien
...
+In the example below, the same `httpbin` service which has already been set up in the prior examples is accessed via client `sleep` pods that are running in a separate namespace that is not part of the mesh. This example shows that east-west traffic between ambient mesh pods and non mesh pods is seamlessly supported. Since the non-mesh pods initiate traffic directly to the backend pods without going through ztunnel. Note that in this case, Istio L4 policy can control whether traffic is denied or allowed from the source.
```
linsun, line 460:
```suggestion
In the example below, the same `httpbin` service which has already been set up in the prior examples is accessed via client `sleep` pods that are running in a separate namespace that is not part of the mesh. This example shows that east-west traffic between ambient mesh pods and non mesh pods is seamlessly supported. The non-mesh pods initiate traffic directly to the destination pods without going through the source ztunnel, while the destination ztunnel enforces any L4 policy to control whether traffic should be allowed or denied.
```

linsun: How about this?

MorrisLaw: I agree, clarifying `destination ztunnel` and `source ztunnel` is more clear. Will update.

linsun (approved):

linsun:
LGTM, thank you @MorrisLaw !!

bleggett (approved):

ericvn (approved):

Issue #13482, opened 2023-06-28
Need a page to explain HBONE


linsun:
We have some contents out there from existing ambient blog but may need additional contents

@justinpettit thoughts? Would you be a good owner for this?

srampal:
@linsun I would be interested in helping or working on this .. possibly in collaboration with @justinpettit or anyone else. Let me know.

srampal:
@linsun @justinpettit  Open to helping with this if Justin needs any help.

linsun:
no one asked about hbone doc outside of Istio contributors imho, could be deferred imho

PR #15004, opened 2024-04-30
Move HBONE section into its own doc

bleggett:
## Description

We've had a glossary description that sort of handwaves at what HBONE is, but given how often we use the acronym and the fact that it's a nontrivial part of the ambient security model, it probably needs its own Architecture section that can be referenced from other spots.

This mostly just moves the existing HBONE section out of the ztunnel architecture doc into its own, smaller, (hopefully) clearer thing.

Fixes: https://github.com/istio/istio.io/issues/13482

## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [x] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/ambient/architecture/hbone/index.md
```diff
@@ -0,0 +1,39 @@
+---
+title: HBONE
```
louiscryan, line 2: Consider renaming to "TLS & Tunneling" and define HBONE within the content here.

When we link to this doc from elsewhere we will just mention TLS

content/en/docs/ambient/architecture/hbone/index.md
```diff
@@ -0,0 +1,39 @@
...
+description: Understanding Istio's secure tunneling protocol.
+weight: 2
+owner: istio/wg-networking-maintainers
+test: no
+---
+
+
+## Understanding the HBONE protocol
+
+HBONE (HTTP Based Overlay Network Encapsulation) is an Istio-specific term. It is a mechanism to transparently multiplex packets related to many different application connections over a single, mTLS encrypted network connection - an encrypted tunnel.
```
louiscryan, line 12: 'packets' or 'TCP streams' ?

similarly in the "... many different application connections..." section. 

No need to pressage QUIC :)

bleggett: yeah `packets` is left over from the previous doc. Agree tho.

bleggett:
DNM-ing until https://github.com/istio/istio.io/pull/14994 goes in

louiscryan (approved):
LGTM

MorrisLaw (approved):

craigbox:
Both are now in.

Don't we use HBONE between sidecars?

bleggett:
> Both are now in.
> 
> Don't we use HBONE between sidecars?

Yep, resynced this against those two, should be good to go.

bleggett:
@craigbox note that this also does https://github.com/istio/istio.io/issues/15005, PTAL

bleggett:
oops, forgot this would merge even if I added @craigbox directly as a reviewer. if he doesn't want some/all of this it I'll back it out.

craigbox:
i'm fine with treating master as a dev branch until we release!  Merge early and often please

PR #15007, opened 2024-05-01
initial waypoint guide

linsun:
## Description

attempt to fix https://github.com/istio/istio.io/issues/14984

## Reviewers

- [x ] Ambient
- [x ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

Issue #15005, opened 2024-04-30
Organize ambient architecture doc


linsun:
Some discussion from https://github.com/istio/istio.io/pull/14994#discussion_r1585314434 where we may likely need to have multiple pages under ambient architecture.  Open this to capture thoughts and actions.

Architecture:
- Overview
- API
  - [Authorization Policy](https://github.com/istio/istio.io/issues/15006)
- Deep Dive

cc @craigbox @bleggett

bleggett:
https://github.com/istio/istio.io/pull/15004 does this as a matter of course.

I think we do need at least 2 subheadings based on content we already have, as @craigbox 's doc already pointed out:

- Architecture
  - Traffic Redirection
  - HBONE

craigbox:
* **Labels**: should move under a "Adding your workload to the ambient mesh" page under /docs/ambient/usage/
* **Layer 7 policy attachment to waypoints**: should move to an "Attaching policy to waypoints" under /docs/ambient/usage/waypoints/, but I'm confused as I would expect to attach policy to workloads, not waypoints
* **Traffic routing**: make this a sub-page of architecture?

bleggett:
> Labels: should move under a "Adding your workload to the ambient mesh" page under /docs/ambient/usage/ ?

Yes please.

Something like

```
### Labels
#### Adding Workloads To Ambient Dataplane
(talk about adding `dataplane-mode` labels)
#### Enforcing L7 Policies In Ambient
(talk about adding `use-waypoint` labels) -> link to `Layer 7 Policy Attachment` subdoc
```

> Layer 7 policy attachment to waypoints: should move to an "Attaching policy to waypoints" under /docs/ambient/usage/waypoints/, but I'm confused as I would expect to attach policy to workloads, not waypoints

I think if we breadcrumb from "for L7 policy, add a waypoint with this label" -> "then see this doc for applying L7 policy (to that waypoint)" it will be clear.

linsun:
Maybe complete now?

craigbox:
https://istio.io/latest/docs/ambient/architecture/ is now only a table of contents with subpages, a good base to build from.

PR #15010, opened 2024-05-02
move architecture docs

### LINKED ISSUE #14993: Move `Ztunnel traffic redirection` guide under `Concepts/Architecture` headings
It is a conceptual explainer, so IMO it makes the most sense there.

- Probably just needs a light once-over after the move.
- Make sure we don't say "inpod mode" or similar anywhere.
- Probably should move from `/docs/ops/ambient/usage/traffic-redirection/` to `/docs/ops/concepts/architecture/traffic-capture/`

craigbox:
* Helps with #15005 and improves #14994
* Fixes #14993 (traffic redirection)
* Improves #13482 if @louiscryan doesn't mind the rename
* Orphans some text that I have [left here](https://github.com/istio/istio.io/pull/15007#issuecomment-2089366377) for consideration in the waypoint guide


content/en/docs/ambient/architecture/_index.md
```diff
@@ -11,15 +11,22 @@ test: n/a
 
 ## Ambient APIs
 
+To enforce L7 policies, add the `istio.io/use-waypoint` label to your resource to use waypoint for the labeled resource.
```
bleggett, line 14: It still feels weird to immediately jump into `use-waypoint` when that won't work if you haven't done `dataplane-mode` yet.

1. Add to ambient dataplane
2. `use-waypoint` if you then want L7

Which in practice probably looks like

```
### Labels
#### Adding Workloads To Ambient Dataplane
#### Enforcing L7 Policies In Ambient

File: content/en/docs/ambient/architecture/_index.md [hunk 2]
```diff
@@ -11,15 +11,22 @@ test: n/a
```
bleggett, line 11: We can leave the path as `/architecture` but can we do:


`description: A deep dive into the data plane of ambient mode.`

content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -94,17 +90,17 @@ caption="Basic ztunnel L4-only datapath"
 
 The figure depicts ambient pod workloads running on two nodes W1 and W2 of a Kubernetes cluster. There is a single instance of the ztunnel proxy on each node. In this scenario, application client pods C1, C2 and C3 need to access a service provided by pod S1 and there is no requirement for advanced L7 features such as L7 traffic routing or L7 traffic management so no Waypoint proxy is needed.
```
bleggett, line 91: I would suggest deleting the entire `## Data Plane Overview` section out of the "usage" guide, and putting the contents into your new `architecture/traffic-routing/index.md` doc (and linking to that from the usage guide)

content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -321,7 +317,7 @@ $ kubectl -n istio-system logs -l app=ztunnel | grep -E "inbound|outbound"
...
+Here note the logs from the ztunnel proxies first indicating the http CONNECT request to the new destination pod (10.240.1.11) which indicates the setup of the HBONE tunnel to ztunnel on the node hosting the additional destination service pod. This is then followed by logs indicating the client traffic being sent to both 10.240.1.11 and 10.240.2.10 which are the two destination pods providing the service. Also note that the datapath is performing client-side load balancing in this case and not depending on Kubernetes service load balancing. In your setup these numbers will be different and will match the pod addresses of the httpbin pods in your cluster.
```
bleggett, line 320:
```suggestion
Here note the logs from the ztunnel proxies first indicating the http CONNECT request to the new destination pod (10.240.1.11) which indicates the setup of the HBONE tunnel to ztunnel on the node hosting the additional destination service pod. This is then followed by logs indicating the client traffic being sent to both 10.240.1.11 and 10.240.2.10 which are the two destination pods providing the service. Also note that the dataplane is performing client-side load balancing in this case and not depending on Kubernetes service load balancing. In your setup these numbers will be different and will match the pod addresses of the httpbin pods in your cluster.
```

Suggestion - settle on `dataplane` or `datapath` (or `data plane`/`data path`)- but not both.

content/en/docs/ambient/usage/ztunnel/index.md
```diff
@@ -183,7 +179,7 @@ This indicates the traffic path is working. The next section looks at how to mon
 
 ## Monitoring the ztunnel proxy & L4 networking {#monitoringzt}
 
-This section describes some options for monitoring the ztunnel proxy configuration and data path. This information can also help with some high level troubleshooting and in identifying information that would be useful to collect and provide in a bug report if there are any problems. Additional advanced monitoring of ztunnel internals and advanced troubleshooting is out of scope for this guide.
+This section describes some options for monitoring the ztunnel proxy configuration and datapath. This information can also help with some high level troubleshooting and in identifying information that would be useful to collect and provide in a bug report if there are any problems. Additional advanced monitoring of ztunnel internals and advanced troubleshooting is out of scope for this guide.
```
bleggett, line 182:
```suggestion
This section describes some options for monitoring the ztunnel proxy configuration and dataplane. This information can also help with some high level troubleshooting and in identifying information that would be useful to collect and provide in a bug report if there are any problems. Additional advanced monitoring of ztunnel internals and advanced troubleshooting is out of scope for this guide.
```

content/en/docs/ambient/architecture/_index.md
craigbox: This page will eventually be a header for the whole section, and not all of it will be about the data plane - some of it will be about the control plane implementation, I would think

bleggett: Ambient should not have any unique control plane docs (or we are doing something wrong) - those should be shared across sidecar and ambient.

content/en/docs/ambient/architecture/_index.md
craigbox: This PR only moves content; there are still a lot of gaps that need to be filled.

content/en/docs/ambient/usage/ztunnel/index.md
craigbox: (agreed - but this needs to be done with the creating of an "About ztunnel" concept page, and then a "Using ztunnel" page.  this PR only touches this page because I needed to change some links.

content/en/docs/ambient/usage/ztunnel/index.md
craigbox: I think "data path" is different to "data plane", especially in the sidecar case.  And yes I would love consistency; I did a quick Google to try and figure out what is most popular

bleggett: I might be thinking about it weirdly - to me they seem pretty synonymous in istio

- Dataplane in sidecar is HTTPS + sidecar envoy
- Dataplane in ambient is HBONE + ztunnel

the rest is redirection, which isn't relevant for the context here.

IDK too much which one tho, if we can use one everywhere let's do it.

content/en/docs/ambient/architecture/_index.md
bleggett: Yeah, the "labels you set" feels like they belong more in "usage" than "architecture" anyway tho - architecture is just "things you do not _absolutely_ need to know to use or follow our usage guides but which you or your compliance teams might want to have an understanding of" in my head.

bleggett (approved):
Move LGTM, outstanding comments are followups/NA

kfaseela (approved):

PR #15027, opened 2024-05-03
Add more ztunnel architectural context

bleggett:
## Description

As @craigbox pointed out we have https://github.com/istio/istio/blob/master/architecture/ambient/ztunnel.md

That doc is more for people working on the code, but it does have some relevant callouts we aren't really covering in our architectural docs, that I've expanded the HBONE and L4 dataplane docs with.

## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/architecture/data-plane/index.md
+++ b/content/en/docs/ambient/architecture/data-plane/index.md
@@ -63,9 +63,11 @@ Ztunnel will request certificates for all identities on the node. It determines 
 Ztunnel additionally will handle the rotation of these certificates as they approach expiry.
 
 ##### Telemetry
+
 Ztunnel emits the full set of [Istio Standard TCP Metrics](docs/reference/config/metrics/).
 
 ##### Dataplane example for Layer 4 traffic
+
 The L4 ambient dataplane between is depicted in the following figure.
 
 {{< image width="100%"
--- a/content/en/docs/ambient/architecture/data-plane/index.md
+++ b/content/en/docs/ambient/architecture/data-plane/index.md
@@ -63,7 +63,7 @@ Ztunnel will request certificates for all identities on the node. It determines 
 Ztunnel additionally will handle the rotation of these certificates as they approach expiry.
 
 ##### Telemetry
-Ztunnel emits the full set of [Istio Standard TCP Metrics](https://istio.io/latest/docs/reference/config/metrics/).
+Ztunnel emits the full set of [Istio Standard TCP Metrics](docs/reference/config/metrics/).
 
 ##### Dataplane example for Layer 4 traffic
 The L4 ambient dataplane between is depicted in the following figure.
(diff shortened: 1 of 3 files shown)


content/en/docs/ambient/architecture/data-plane/index.md
```diff
@@ -46,8 +46,26 @@ requests directly to the destination without going through any waypoint proxy ev
 Currently, traffic from sidecars and gateways won't go through any waypoint proxy either and they will be made aware of waypoint proxies
 in a future release.
 
-#### Dataplane details
+#### L4 dataplane details
```
craigbox, line 49: I would prefer not to say "L4 dataplane".

Istio has a control plane and a data plane. How the data plane is implemented can be one of at least three ways.

Data plane for Layer 4 traffic?

First thought, not perfect.

It's wordier but may be easier to make consistent.

bleggett: fair 'nough.

content/en/docs/ambient/architecture/data-plane/index.md
```diff
@@ -46,8 +46,26 @@ requests directly to the destination without going through any waypoint proxy ev
...
 
+##### Identity
+
+All inbound and outbound traffic between workloads in the ambient mesh is secured by the data plane, using mTLS via {{< gloss >}}HBONE{{< /gloss >}}, ztunnel, and x509 certificates.
+
+However, the certificates used will be bound to the workloads themselves, not ztunnel. This means ztunnel will hold multiple distinct certificates at a time, one for each unique identity (service account) running on its node.
```
ilrudie: This is a tough bit of information to convey. I'm not certain the suggestion should be accepted as is but perhaps it will contain inspiration for a revision at least.

```suggestion
All inbound and outbound traffic between workloads in the ambient mesh is secured by the data plane, using mTLS via {{< gloss >}}HBONE{{< /gloss >}}, ztunnel, and x509 certificates. The x509 certificates will be used by to sign traffic on behalf of node-local pods as well as validate traffic arriving at these pods. In order to achieve a zero-trust posture the certificates used will not bear the identity of a trusted node-proxy component instead containing the attested identity of each node-local pod respectively. This requires the ztunnel to manage multiple distinct certificates at a time, one for each unique identity (service account) for every node-local pod.
```

ilrudie (commented):
nice work, one awkward (imo at least) section I left a suggestion for but overall looks good to me

content/en/docs/ambient/architecture/data-plane/index.md
bleggett: Yeah I'm kinda trying to avoid getting too much into "how mTLS works" here since we cover that more in the HBONE arch bit (which we breadcrumb to via the glossary ref).

bleggett: Reworded a bit using some of your suggestion.

bleggett: Reworded a bit using some of your suggested text.

ilrudie: nice, good middle ground

ilrudie (approved):
lgtm

craigbox (approved):

PR #15055, opened 2024-05-09
Add revision and a note on automatic gateways

costinm:
## Description

Revision and using K8S Gateway. 

We may want to expand on the later - since Waypoints also use Gateway resource, not clear why we would
want to document the old method and have users deal with both.

Also, as we already have the 'demo' install via istioctl ( which uses the default istiod revision ) it may be worth
having this section just show the revision-based install, as it will help with migrations and co-existence too.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


craigbox:
can you rebase please and I can fix up!

windsonsea:
You can commented our all irrelevant commits when you perform rebase action like:

```
# pick 1
# pick 2
# pick 3
your-commit
# pick 4
# pick 5
```

costinm:
Ok, no clue what happened here - I tried to rebase on my branch and somehow it shows this was merged upstream ? Will open a new PR and try to revert if something go changed.

PR #15075, opened 2024-05-14
bump api and istio refs

dhawton:
## Description

Resolve lint failure due to changed api

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


go.mod
```diff
@@ -195,7 +195,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	helm.sh/helm/v3 v3.14.4 // indirect
-	istio.io/api v1.22.0-beta.0 // indirect
```
craigbox, line 198: are these sensible changes? reads like a move backwards

craigbox (approved):

go.mod
dhawton: beta.0's tag is actually older than 09344006f961. So while it looks odd "going back" to alpha, the commit used is newer. I don't know how Go does tags like this.

dhawton:
/test doc.test.profile-none

This is an odd failure.. the same bookinfo startup func is used in other tests without issue...

craigbox:
/test doc.test.profile-none

I can do this all day.

kfaseela (approved):

kfaseela:
@zirain any idea why the tracing test fails?

kfaseela:
Does not look like the usual flakes

dhawton:
> @zirain any idea why the tracing test fails?

There was 1 or 2 runs where it was something else... on this, it is productpage-v1 timing out deploying, for some reason, even though the same deployment succeeds elsewhere. I don't see any info in the other logs to know why it's not deploying.

dhawton:
Some branch protections now exist so I can't push to this branch anymore... reopening from fork.

PR #15136, opened 2024-05-16
Wasm Configuration for Ambient Mode - Updates

adilmp1:
This PR's modification addresses the resolution of an issue identified in Issue [#51014](https://github.com/istio/istio/issues/51014) by John. Since the issue is fixed, we can now apply the WasmPlugin directly to the service instead of applying it to the gateway. Previously, despite providing credentials, requests to other services bypassed authentication, prompting the decision against applying the WasmPlugin directly to the service. Relevant tests have been updated to reflect these changes, although they will encounter failures until the new Istio image incorporating John's fixes is deployed. 

Adds documentation as per : https://github.com/istio/istio/issues/42337


craigbox (approved):
/retest-required

kfaseela (changes requested):
we need to remove the service specific waypoint proxy from the document. With the fix from @howardjohn it is not needed anymore. And then let us wait till the CI has the latest test image to retest this and see if the it passes.

kfaseela (approved):

PR #15143, opened 2024-05-18
remove stackdriver reference

zirain:
xref: https://github.com/istio/istio.io/pull/15123

cc @kyessenov @lei-tang


kfaseela:
why is lint still failing? is there any more reference which is to be removed?

zirain:
wait https://github.com/istio/api/pull/3204

lei-tang (approved):

data/features.yaml
```diff
@@ -142,14 +142,6 @@ features:
       maturity: Stable
       nextExpectedPromotion: ""
     area: Observability
-  - name: "Stackdriver Integration"
```
ericvn, line 145: This change needs to be made in the istio/enhancements repo as it's pulled from there: 

https://github.com/istio/istio.io/blob/master/scripts/grab_reference_docs.sh#L114

ericvn (changes requested):

data/features.yaml
zirain: https://github.com/istio/enhancements/pull/179

PR #15152, opened 2024-05-20
Updates Ambient Getting Started Guide

### LINKED ISSUE #15119: Confusing Steps in Ambient Getting Started
Overall, I was able to successfully complete the steps in the [ambient getting started](https://istio.io/latest/docs/ambient/getting-started/) guide. However, a few stpes provided confusion that I would like to bring to the community's attention:

> Deploy an ingress gateway so you can access the bookinfo app from outside the cluster:
To get IP address assignment for Loadbalancer service types in kind, you may need to install a tool like [MetalLB](https://metallb.universe.tf/). Please consult [this guide](https://kind.sigs.k8s.io/docs/user/loadbalancer/) for more information.
Create a [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1.Gateway) and [HTTPRoute](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1.HTTPRoute):
$ kubectl apply -f [samples/bookinfo/gateway-api/bookinfo-gateway.yaml](https://raw.githubusercontent.com/istio/istio/release-1.22/samples/bookinfo/gateway-api/bookinfo-gateway.yaml)

None of the proceeding steps actually test connectivity from outside the cluster. By default, the bookinfo Gateway creates a service of type `LoadBalancer` and therefore the Gateway status never reaches `Programmed=true` without running in a cloud environment or using a tool such as MetalLB to populate the Service status with a LoadBalancer IP. Istio supports annotating a Gateway with `networking.istio.io/service-type: ClusterIP` to create a ClusterIP service instead of a LoadBalancer service. I suggest updating the guide to use this annotation to simplify the UX.

>Update your AuthorizationPolicy to explicitly allow the sleep service to GET the productpage service, but perform no other operations:

The proceeding yaml removes the `- cluster.local/$GATEWAY_SERVICE_ACCOUNT` source from the L4 example that is being updated but makes no mention why. I believe this was done in error and should be re-added.

danehans:
## Description

- Adds a step to annotate the ingress gateway to use a ClusterIP service.
- Removes the reference to external tooling for managing a LoadBalancer service type.

Fixes #15119

## Reviewers

- [X] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/getting-started/index.md
+++ b/content/en/docs/ambient/getting-started/index.md
@@ -112,17 +112,19 @@ Make sure the default namespace does not include the label `istio-injection=enab
 
     `sleep` and `notsleep` are two simple applications that can serve as curl clients.
 
-1. Deploy an ingress gateway so you can access the bookinfo app from outside the cluster:
-
-    {{< tip >}}
-    To get IP address assignment for `Loadbalancer` service types in `kind`, you may need to install a tool like [MetalLB](https://metallb.universe.tf/). Please consult [this guide](https://kind.sigs.k8s.io/docs/user/loadbalancer/) for more information.
-    {{</ tip >}}
+1. Deploy an ingress gateway:
 
     Create a [Kubernetes Gateway](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1.Gateway)
     and [HTTPRoute](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1.HTTPRoute):
 
     {{< text bash >}}
     $ kubectl apply -f @samples/bookinfo/gateway-api/bookinfo-gateway.yaml@
+    {{< /text >}}
+
+    By default, Istio creates a `LoadBalancer` service for a gateway. Change the service type to `ClusterIP` by annotating the gateway.
+
+    {{< text bash >}}
+    $ kubectl annotate gateway bookinfo-gateway networking.istio.io/service-type=ClusterIP --namespace=default
     {{< /text >}}
 
     Set the environment variables for the Kubernetes Gateway:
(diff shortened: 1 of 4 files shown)


content/en/docs/ambient/getting-started/index.md
```diff
@@ -112,11 +112,7 @@ Make sure the default namespace does not include the label `istio-injection=enab
 
     `sleep` and `notsleep` are two simple applications that can serve as curl clients.
 
-1. Deploy an ingress gateway so you can access the bookinfo app from outside the cluster:
-
-    {{< tip >}}
-    To get IP address assignment for `Loadbalancer` service types in `kind`, you may need to install a tool like [MetalLB](https://metallb.universe.tf/). Please consult [this guide](https://kind.sigs.k8s.io/docs/user/loadbalancer/) for more information.
-    {{</ tip >}}
+2. Deploy an ingress gateway:
```
dhawton, line 115:
```suggestion
1. Deploy an ingress gateway:
```

dhawton (commented):
You'll need to run `make gen` to update the snippets used in testing since you updated text blocks. May also need to update test.sh to call the correct snippets

content/en/docs/ambient/getting-started/snips.sh
```diff
@@ -59,6 +59,10 @@ kubectl apply -f samples/bookinfo/gateway-api/bookinfo-gateway.yaml
 }
 
 snip_deploy_the_sample_application_4() {
+kubectl annotate gateway bookinfo-gateway networking.istio.io/service-type=ClusterIP --namespace=default
+}
+
+snip_deploy_the_sample_application_5() {
```
dhawton, line 65: You created a new snip, test is failing because the exports from here aren't available when snip_verify_traffic_sleep_to_ingress is called in test.sh. You'll need to call _5 as well.

dhawton:
The failing test is pulling in snips from the getting started, and the function it ran changed which is causing the fail. You'll have to, unfortunately, fix that as well to get the test to pass. https://github.com/istio/istio.io/blob/master/content/en/docs/ambient/usage/extend-waypoint-wasm/test.sh#L38-L43 you changed the `snip_deploy_the_sample_application_4` function.

dhawton (approved):

PR #15174, opened 2024-05-23
Update index.html

johnzheng1975:
## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


johnzheng1975:
Actually, we do support <code>/dev/null</code>  besides stdout and stderr.
I think /dev/null is very useful for some case, since /dev/stderr still print many logs.
FYI

zirain (changes requested):
this is generated from api repo, please do not modify it manually.

johnzheng1975:
Thanks :-)

PR #15184, opened 2024-05-24
image showing the historical statistics of the project would be most …

chaltenio:
…appropriate for this celebration.

## Description

Image showing the historical statistics of the project -instead of the last year only- would be most appropriate for this celebration. 

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


kfaseela:
cc @linsun @craigbox

Arhell (commented):
/ok-to-test

kfaseela (changes requested):
Agree with @linsun

chaltenio:
Thanks @linsun.

I agree that displaying statistics for the last twelve months is valuable for recognizing recent active contributors.

Additionally, it is important to publicly share the historical contribution statistics of the project to honor and acknowledge the efforts of individuals and organizations that have invested their time and dedication over the years in the project.

Two graphs showing contributions over the past 12 months and the past 7 years would be sufficient and provide transparency. I do not think it is necessary to break down the data further.

linsun:
I've read the discussion in slack, I believe given this is a birthday blog (which is yearly), pretty much everyone (@kfaseela @ericvn @howardjohn @keithmattix @justinpettit @craigbox) who serves a leadership position in Istio in the thread agrees that 12 months contribution is appropriate for the context of this blog.  @kfaseela suggested last decade contribution could be useful for a 10th birthday which I think it is a good idea.

Note: I don't fully understand @hzxuzhonghu as he said he agreed with Eric, which I assume he meant if the blog talks about Istio as a whole (not a birthday), showing all 7 years make sense.

chaltenio:
I discussed this with Craig Box and will be closing this PR. 

Note that I still don't fully understand why the following proposal to add a second graph with the project's entire history of stats to this blog - adding more information and transparency - is rejected.

PR #15210, opened 2024-05-31
Add ambient profile for docs test

kfaseela:
## Description

Now that we have more documentation and document tests coming with ambient profile, I think it makes better sense to have an ambient profile created separately?

Wait on https://github.com/istio/test-infra/pull/5374

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/tests/setup/profile-ambient/doc_test.go
+++ b/tests/setup/profile-ambient/doc_test.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Istio Authors
+// Copyright 2024 Istio Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


kfaseela:
The cleanup section of the ambient test will fail until we have https://github.com/istio/istio/pull/51329/files

kfaseela:
@dhawton i am separating out the test changes and profile creation, so that this PR can be merged first. https://github.com/istio/istio.io/pull/15220/files seems to have issues and may be delayed, so the test will otherwise fail until that gets merged.

craigbox (approved):

Issue #15200, opened 2024-05-30
test scripts break when a folder with snips is created and removed in the same PR


craigbox:
In #15168 we created `ambient/getting-started/setup/snips.sh`, and it was removed in a later commit.

Now, tests won't pass, because `scripts/pr_tests.py`is looking at all the places a snips.sh existed in all commits, not just in the end state.

```
+ echo 'Optimizing tests for pull number: 15168'
Optimizing tests for pull number: 15168
++ python3 ./scripts/pr_tests.py --token= 15168
+ TESTS=ambient/usage/extend-waypoint-wasm,ambient/getting-started/setup,ambient/getting-started/cleanup,ambient/getting-started/manage-traffic,ambient/getting-started/secure-and-visualize,ambient/getting-started/enforce-auth-policies,ambient/getting-started/deploy-sample-app,ambient/getting-started
```

PR #15168, opened 2024-05-22
docs: break apart the ambient getting start guide (WIP)

peterj:
## Description

[WIP] Breaking up the Ambient Getting Started section into multiple smaller sections.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

TODOs:

- [x] make the tests work
- [x] update the bookinfo screenshot (once new docker images are published)
- [ ] show L7 metrics (for some reason I don't see them in Kiali)

PR #15227, opened 2024-06-03
Remove empty folder with snips

craigbox:
## Description

In #15168 we had, once, created a folder with an index.md/`test: yes`/snips.sh combo.

It was removed in a later commit, but the script checked directories that had existed _in any commit_ (#15200). Thus, this file needed to be created with a boilerplate (and otherwise empty) snips.sh to have the tests pass. It is now being removed.

## Reviewers

If the tests pass, @kfaseela.  If they do not, then I think it will need to be forced, and I summon @howardjohn assuming it's rude to ask Eric, and I'm not sure who else can do such things (Louis? MItch? Anyone on the TOC?)

- [x] Docs
- [x] Test and Release


craigbox:
As suspected, all tests are failing:

> 2024-06-03T10:06:37.429272Z	fatal	tf	Error occurred while traversing content: lstat /home/prow/go/src/istio.io/istio.io/content/en/docs/ambient/getting-started/setup: no such file or directory

and this will need force-merging.  

Tests won't fail _after_, they will only fail _during_.

kfaseela:
i did not get this error on https://github.com/istio/istio.io/pull/15226

craigbox:
The error is only caused in PRs that _remove_ a folder with snips.

The test scripts look at folders that were touched by any commit, and
contain a snips.sh file.

craigbox:
@howardjohn are you OK to force-merge this?

howardjohn (approved):

Issue #15232, opened 2024-06-03
consider a better general term than "ztunnel" for ambient's node proxy


ilrudie:
ztunnel is really an implementation detail in the ambient architecture. Presently there isn't really another implementation of the layer 4 node proxy but that is potentially not always true. Additionally, at present ztunnel really only acts as the node proxy but that again is hot a hard and fast rule. Nothing really prevents it from being deployed as a sidecar sometime down the road if we find an interesting use for that architecture for instance.

To those outside of the Istio development stream ztunnel also doesn't mean all that much. Consider trying to describe sidecar Istio to someone who may not know that much about the implementation of a service mesh but the term you have available is just "envoy".

craigbox:
**(Bikeshed warning)**

I think the problem here is one of there not being a way to describe the two different potential traffic flows in "ambient mode"

- ztunnel only
- ztunnel and waypoint

We sometimes use "secure L4 overlay" mode to describe the first one.  It's two words too many.

If we standardised on something like

- ambient mode
- waypoint mode

we would then have to talk about the "node proxy" when you're referring to the thing that enforces the L4 policy, for example.

(noting that @louiscryan would prefer "ambient/waypoint mode") we should almost never have to say "ztunnel" except when talking about configuring/using Istio's reference implementation of the node proxy

ilrudie:
Ambient `enabled` == no waypoint
Ambient `enhanced` == with a waypoint

bleggett:
:+1:  on not using "ztunnel" in architecture documents.

- `node proxy` (preferred, it describes <location> <function> which is enough)
- `L4 proxy`

ilrudie:
Some general thoughts translated from the ambient wg meeting discussion

- `node proxy` was not really strong disliked and gels pretty well with other terminology
- `l4 proxy` had some push back

`istio-` prefix could be added but in the context of Istio documentation that is probably not necessary
`ambient-` prefix could be added but it may imply that a waypoint somehow isn't ambient and is also likely redundant so was not well received

PR #15247, opened 2024-06-05
Expand Ambient Canary Upgrade Instructions

therealmitchconnors:
## Description

Expand Ambient Upgrade Docs to specify Revision-Based Upgrades with Tags, increasing stability.

## Reviewers

- [x] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,101 @@ description: Upgrading an ambient mode installation with Helm.
...
+The ztunnel DaemonSet is the node proxy component. The ztunnel at version 1.x is compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before ztunnel, as long as their version difference is within one minor version.
 
 {{< warning >}}
 Upgrading ztunnel in-place will briefly disrupt all ambient mesh traffic on the node.
 Node cordoning and blue/green node pools are recommended to mitigate blast radius risk during production upgrades. See your Kubernetes provider documentation for details.
 {{< /warning >}}
 
 {{< text syntax=bash snip_id=upgrade_ztunnel >}}
-$ helm upgrade ztunnel istio/ztunnel -n istio-system
+$ helm upgrade ztunnel istio/ztunnel -n istio-system --set revision="$REVISION"
```
howardjohn, line 94: nit: do we need to tell them to pass the original values? Is that even needed or does helm automagically do it

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,101 @@ description: Upgrading an ambient mode installation with Helm.
...
+## Upgrade the CNI DaemonSet
 
 The Istio CNI agent is responsible for detecting pods added to the ambient mesh, informing ztunnel that proxy ports should be established within added pods, and configuring traffic redirection within the pod network namespace. It is not part of the data plane or control plane.
 
-{{< warning >}}
-As ambient mode is not yet Stable, the following statement is not a compatibility guarantee, and is subject to change, or removal. Prior to reaching Stable status, this component and/or the control plane may receive breaking changes that prevent compatibility between minor versions.
-{{< /warning >}}
-
-The CNI at version 1.x is generally compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before Istio CNI, as long as their version difference is within one minor version.
+The CNI at version 1.x is  compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before Istio CNI, as long as their version difference is within one minor version.
```
howardjohn, line 101: out of scope: we should update https://istio.io/latest/docs/releases/supported-releases/#control-planedata-plane-skew to include CNI. And maybe for ambient in general

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,101 @@ description: Upgrading an ambient mode installation with Helm.
...
 
 {{< warning >}}
 Upgrading the Istio CNI agent to a compatible version in-place will not disrupt networking for running pods already successfully added to an ambient mesh, but no ambient-captured pods will be successfully scheduled (or rescheduled) on the node until the upgrade is complete and the upgraded Istio CNI agent on the node passes readiness checks. If this is a significant disruption concern, or stricter blast radius controls are desired for CNI upgrades, node taints and/or node cordons are recommended.
```
howardjohn, line 104: is this true? I think they will schedule and not have ambient capabilities @bleggett ?(unless you have taints, which is not documented here)

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -135,6 +150,18 @@ You can check the status of the deployed Istio pods with:
...
+$ helm delete istiod-"$REVISION" -n istio-system
+{{< /text >}}
+
+## Upgrade the CRDs
+
+Due to the difficulty of downgrading custom resource definitions, ({{< gloss >}}CRDs{{</ gloss >}}) should be upgraded only once you are confident you do not need to rollback to the previous version:
 
-Please refer to the uninstall section in the [Helm installation guide](/docs/ambient/install/helm-installation/#uninstall).
+    {{< text syntax=bash snip_id=manual_crd_upgrade >}}
+    $ kubectl apply -f manifests/charts/base/crds
```
howardjohn, line 166: fwiw https://deploy-preview-15247--preliminary-istio.netlify.app/latest/docs/setup/upgrade/helm/#in-place-upgrade does it the other order. I don't have a strong preference one way or another, but it seems like they ought to be consistent

howardjohn: Really it seems like most of the doc could be shared (or at least very similar) with just some additional steps for ambient

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: Does the CNI interact with the control plane?  isn't it's job just redirecting traffic to the ztunnel?

bleggett: Yeah and I think I wrote this, it is slightly wrong - it should read:

"Upgrading the Istio CNI agent to a compatible version in-place will not disrupt networking for running pods already successfully added to an ambient mesh, but any new pods scheduled on the node _during_ the `istio-cni` upgrade process may not be properly captured and could leak traffic during the upgrade process, until the upgrade is complete and the upgraded Istio CNI agent on the node passes readiness checks. If this is a significant disruption concern, or stricter blast radius controls are desired for CNI upgrades, node taints and/or node cordons are recommended."

content/en/docs/ambient/upgrade/helm-upgrade/index.md
bleggett: Looking at it, putting it under `Operations` and linking to it as an "advanced upgrade strategy" makes sense, yeah.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
howardjohn: My comment about 'not true' was about "but no ambient-captured pods will be successfully scheduled" btw

To your question - 
It **indirectly** interacts with the control plane because the things it does needs to be compatible with how the proxies are configured. it is not directly communicating though

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,101 @@ description: Upgrading an ambient mode installation with Helm.
...
-{{< /text >}}
+## Prerequisites
 
-{{< warning >}}
-[Helm does not upgrade or delete CRDs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations) when performing an upgrade. Because of this restriction, an additional step is required when upgrading Istio with Helm.
-{{< /warning >}}
+1. Organize your Tags and Revisions
 
-### Manually upgrade the CRDs and Istio base chart
+    In order to safely upgrade Ambient, your Waypoints, Gateways, and Namespaces should be using the istio.io/rev label to specify an Istio Tag which will control the version of the Istio Proxy that is running. We recommend dividing your production cluster into multiple tags to organize your upgrade. All members of a given tag will be upgraded simultaneously, so it is wise to begin your upgrade with your lowest risk applications. We do not recommend referencing revisions directly via labels for upgrades, as this process can easily result in the accidental upgrade of a large number of proxies, and is difficult to segment. To see what tags and revisions you are using in your cluster, see the section on upgrading tags.
```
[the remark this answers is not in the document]
bleggett: I am also fine with documenting the current tag semantics, but putting a warning around the bits that interact with MWCs - we _cannot_ carry that forward in ambient, and an ambient upgrade guide should either not recommend the use of tags, or say "this is how tags work today, in ambient they will not work this way, TBD"

content/en/docs/ambient/upgrade/helm-upgrade/index.md
bleggett: It indirectly interacts with ztunnel (dataplane) , which directly interacts with the controlplane, yep

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -14,7 +14,7 @@ Follow this guide to upgrade and configure an ambient mode installation using
 [Helm](https://helm.sh/docs/). This guide assumes you have already performed an [ambient mode installation with Helm](/docs/ambient/install/helm-installation/) with a previous version of Istio.
 
 {{< warning >}}
-In contrast to sidecar mode, ambient mode supports moving application pods to an upgraded data plane without a mandatory restart or reschedule of running application pods. However, upgrading the data plane **will** briefly disrupt all workload traffic on the upgraded node, and ambient mode does not currently support canary upgrades of the data plane.
+In contrast to sidecar mode, ambient mode supports moving application pods to an upgraded data plane without a mandatory restart or reschedule of running application pods. However, upgrading the ztunnel **will** briefly disrupt all workload traffic on the upgraded node, and ambient mode does not currently support canary upgrades of the ztunnel.
```
bleggett, line 17: I almost regret asking, but should we follow https://github.com/istio/istio.io/issues/15232 and say `node proxy` instead of `ztunnel` here? Or maybe say `ztunnel` instead of `the ztunnel`?
 
 At least if we do namedrop it, let's use `<gloss>` so people can figure out what the heck we're talking about (I know we explicitly say this later on, but we should have a central definition for jargon and define them once, or use non-jargon)

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -83,7 +85,7 @@ $ helm install istiod-"$REVISION" istio/istiod -n istio-system --set revision="$
 
 ## Upgrade the ztunnel DaemonSet
 
-The ztunnel DaemonSet is the node proxy component. The ztunnel at version 1.x is compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before ztunnel, as long as their version difference is within one minor version.
+The ztunnel DaemonSet is the node proxy component. The ztunnel at version 1.x is compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before ztunnel, as long as their version difference is within one minor version. If you have previously customized your ztunnel installation, you can reuse the `values.yaml` file from previous upgrades or installs to keep your data planes consistent.
```
bleggett, line 88: `data plane` singular (there is by definition only one)

and if we use `data plane` let's `{{< gloss >}}data plane{{< /gloss >}}` like we do in some other spots.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: I've added a gloss reference to the ztunnel step.  I also have links to the ambient architecture page where these terms are explained at length.  While I support using the term L4 Node Proxy, I don't think this PR is the right place to do that work.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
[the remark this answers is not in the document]
therealmitchconnors: To be clear, we have already carried forward MWCs into the Ambient Beta, both for revisions and for tags.  Ambient Waypoints will not function without MWCs present in the cluster.  I am all for improving this experience, but we should not hold up documentation of current functionality in hopes of better functionality later on...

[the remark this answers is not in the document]
bleggett: > To be clear, we have already carried forward MWCs into the Ambient Beta, both for revisions and for tags. Ambient Waypoints will not function without MWCs present in the cluster. I am all for improving this experience, but we should not hold up documentation of current functionality in hopes of better functionality later on...

Okay, so that it's clear and I understand it, let me repeat this back to you:

- We are signed up to support the revisions feature in Ambient Beta (true)
- We are signed up to support the tags feature in Ambient Beta (true)
- We are signed up to support MWCs  in ambient beta, only indirectly, as a requirement for using tags and revisions as non-unstable features (hopefully true)

If we have to ship MWCs with ambient just support tags and revisions we probably shouldn't sign up for that - I am fine with saying "we support tags and revisions, and those are _currently_ implemented with MWCs, the implementation may change later"

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
 {{< /warning >}}
 
+## Understanding Ambient Upgrades
+
+All Istio Upgrades involve upgrading the control plane, data plane, and Istio CRDs. Because the Ambient Data Plane is separated into [two components](/docs/ambient/architecture/data-plane), the ztunnel and waypoints, Ambient Upgrades involve separate steps for these components. Upgrading the control plane and CRDs will be covered here in brief, but is essentially identical to the process for these components in sidecar mode, which is covered extensively [here](docs/setup/upgrade/canary/).
+
+Like sidecar mode, ambient mode upgrades make use of Tags and Revisions to allow fine-grained control over gateway upgrades, including waypoints, with simple controls for rolling back at any point. However, unlike sidecar mode, the Ambient ztunnel runs as a DaemonSet, or per-node proxy, meaning that ztunnel upgrades affect, at minimum, one entire node at a time. For this reason, we recommend using Node cordoning and draining before upgrading the ztunnel component for a given node. For the sake of simplicity, this document will demonstrate in-place upgrades of the ztunnel, which are less safe, and may involve some downtime.
```
bleggett, line 26: I really don't think we should imply that you should only use tags and revisions to upgrade, or even mention that first.

We support _2_ upgrade modes - inplace of all components (simple, fewer blast radius controls), tag + rev (complex, requires more operational planning, not even required if you A/B clusters for instance)

I think we have to document both, and start with the simpler one (inplace). You _may_ have operational needs that make the more complex tag+rev model more appealing, see here for reference, inplace upgrade of everything has these drawbacks, etc etc.

Especially since ztunnel _has_ to be an inplace upgrade anyway no matter which mode you pick, it's not really a choice between "safe" and "unsafe", it's two different models that might be better or worse dependent on your operational setup and needs.

I would rather have an advanced upgrade guide (tags and revisions) under Operations - I don't want to obscure the simplest upgrade instructions, and I don't want to imply everyone should be using tag and rev - most people do not, even today, and it's even less critical that they do for ambient than for sidecars, given the non-requirement for pod restarts.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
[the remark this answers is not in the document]
therealmitchconnors: The only direct references to MWCs in this document are for listing tags and revisions.  We could, instead, instruct the user to make use of istioctl to list tags and revs, which should continue to work no matter what API we eventually come up with.  Do you think that would be better?  I have avoided that so far because typically helm users don't want to make use of istioctl...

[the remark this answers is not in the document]
bleggett: No, I agree with the approach here - it's ugly we make you grep thru MWCs, but we do.

Putting in a disclaimer or something that inspecting MWCs directly is subject to change and not a stable API is probably enough.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
bleggett, line 26:
```suggestion
Like sidecar mode, ambient mode upgrades make use of Tags and Revisions to allow fine-grained control over gateway upgrades, including waypoints, with simple controls for rolling back at any point. However, unlike sidecar mode, the Ambient ztunnel runs as a DaemonSet, or per-node proxy, meaning that ztunnel upgrades affect, at minimum, one entire node at a time. For this reason, we recommend using Node cordoning and draining before upgrading the ztunnel component for a given node. For the sake of simplicity, this document will demonstrate in-place upgrades of the ztunnel, which are less safe, and may involve some downtime.
```

It wouldn't be clear to me as a reader/operator not familiar with Istio or ambient what the implications of "revisions work for gateways, and also waypoints" means.

Should we say "revisions only affect north-south and east-west  gateways, but not the Istio dataplane as a whole"?

At least, it is probably worth linking back to some other doc that explains what a "waypoint" is, or using a glossary ref here - as an operator, I want to understand how my application will be disrupted or not by the use of revisions - which (unfortunately) requires that I understand the different kinds of Istio gateways, and what they do.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
-[Helm](https://helm.sh/docs/). This guide assumes you have already performed an [ambient mode installation with Helm](/docs/ambient/install/helm-installation/) with a previous minor or patch version of Istio.
+[Helm](https://helm.sh/docs/). This guide assumes you have already performed an [ambient mode installation with Helm](/docs/ambient/install/helm-installation/) with a previous version of Istio.
 
 {{< warning >}}
-In contrast to sidecar mode, ambient mode supports moving application pods to an upgraded data plane without a mandatory restart or reschedule of running application pods. However, upgrading the data plane **will** briefly disrupt all workload traffic on the upgraded node, and ambient mode does not currently support canary upgrades of the data plane.
+In contrast to sidecar mode, ambient mode supports moving application pods to an upgraded data plane without a mandatory restart or reschedule of running application pods. However, upgrading the ztunnel **will** briefly disrupt all workload traffic on the upgraded node, and ambient mode does not currently support canary upgrades of the ztunnel.
```
howardjohn, line 17: Non blocking, especially since its existing wording, but "briefly disrupt" could use some expansion. Something like https://github.com/istio/istio/pull/51805 tuned for users.

tl;dr would be "all long lived TCP connections will be reset"

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
+
+All Istio Upgrades involve upgrading the control plane, data plane, and Istio CRDs. Because the Ambient Data Plane is separated into [two components](/docs/ambient/architecture/data-plane), the ztunnel and waypoints, Ambient Upgrades involve separate steps for these components. Upgrading the control plane and CRDs will be covered here in brief, but is essentially identical to the process for these components in sidecar mode, which is covered extensively [here](docs/setup/upgrade/canary/).
+
+Like sidecar mode, ambient mode upgrades make use of Tags and Revisions to allow fine-grained control over ({{< gloss >}}gateway{{</ gloss >}}) upgrades, including ({{< gloss >}}waypoint{{</ gloss >}})s, with simple controls for rolling back at any point. However, unlike sidecar mode, the Ambient ztunnel runs as a DaemonSet, or per-node proxy, meaning that ztunnel upgrades affect, at minimum, one entire node at a time. For this reason, we recommend using Node cordoning and draining before upgrading the ztunnel component for a given node. For the sake of simplicity, this document will demonstrate in-place upgrades of the ztunnel, which are less safe, and may involve some downtime.
```
howardjohn, line 26:
```suggestion
Like sidecar mode, ambient mode upgrades can make use of Tags and Revisions to allow fine-grained control over ({{< gloss >}}gateway{{</ gloss >}}) upgrades, including ({{< gloss >}}waypoint{{</ gloss >}})s, with simple controls for rolling back at any point. However, unlike sidecar mode, the Ambient ztunnel runs as a DaemonSet, or per-node proxy, meaning that ztunnel upgrades affect, at minimum, one entire node at a time. For this reason, we recommend using Node cordoning and draining before upgrading the ztunnel component for a given node. For the sake of simplicity, this document will demonstrate in-place upgrades of the ztunnel, which are less safe, and may involve some downtime.
```
maybe? No strong preference.

howardjohn: I would like the messaging to be something along these lines:
* Ztunnel upgrade provides semantics that are good enough for 99.99% of use cases, so we **recommend** in place upgrades for most users
* If you are a rare user, you can do it with cordoning, its just harder

The current wording suggests you actually really need cordoning, but its too complex, so we will only document the demo way.

Note: since some of the original discussions around this, Ztunnel has been shown to be/made to be far less impactful during upgrade which may skew some original decisions.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
+
 ## Prerequisites
 
-1. Update the Helm repository:
+1. Organize your Tags and Revisions
 
-    {{< text syntax=bash snip_id=update_helm >}}
-    $ helm repo update istio
-    {{< /text >}}
+    In order to safely upgrade Ambient, your Gateways and Namespaces should be using the istio.io/rev label to specify an Istio Tag which will control the version of the Istio Proxy that is running. We recommend dividing your production cluster into multiple tags to organize your upgrade. All members of a given tag will be upgraded simultaneously, so it is wise to begin your upgrade with your lowest risk applications. We do not recommend referencing revisions directly via labels for upgrades, as this process can easily result in the accidental upgrade of a large number of proxies, and is difficult to segment. To see what tags and revisions you are using in your cluster, see the section on upgrading tags.
```
howardjohn, line 42:
```suggestion
    In order to safely upgrade Ambient, your Gateways and Namespaces should be using the `istio.io/rev` label to specify an Istio Tag which will control the version of the Istio Proxy that is running. We recommend dividing your production cluster into multiple tags to organize your upgrade. All members of a given tag will be upgraded simultaneously, so it is wise to begin your upgrade with your lowest risk applications. We do not recommend referencing revisions directly via labels for upgrades, as this process can easily result in the accidental upgrade of a large number of proxies, and is difficult to segment. To see what tags and revisions you are using in your cluster, see the section on upgrading tags.
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -103,38 +110,38 @@ Upgrading the Istio CNI agent to a compatible version in-place will not disrupt
...
-{{< text syntax=bash snip_id=show_components >}}
-$ helm list -n istio-system
+{{< text syntax=bash snip_id=rollback_tag >}}
+$ helm template istiod istio/istiod -s templates/revision-tags.yaml --set revisionTags="{$MYTAG}" --set revision="$OLD_REVISION" -n istio-system | kubectl apply -f -
 {{< /text >}}
 
-You can check the status of the deployed Istio pods with:
+### Upgrade the Gateway component (optional)
+
+Gateway components using the `networking.istio.io/*/Gateway` types must be upgraded individually using helm.
```
howardjohn, line 135: nit: I would call this the "using the [manual gateway deployment mode](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment)". Also on L127

content/en/docs/ambient/upgrade/helm-upgrade/index.md
howardjohn: I agree with this. I don't know how we want to express that on the page (tabs, two pages, ...), but it does seem likely we are incorrectly implying (by omission of non-revision docs) you must use revisions to upgrade ambient, which may give the perception ambient is more complex.

howardjohn (approved):
LGTM, we can follow up on https://github.com/istio/istio.io/pull/15247#discussion_r1630511929

content/en/docs/ambient/upgrade/helm-upgrade/index.md
bleggett: I mentioned this over in https://github.com/istio/istio.io/pull/15279#discussion_r1643545398

(which also reworks the upgrade docs a bit)

I am fine fixing this in a followup if we want to merge this now, but we simply shouldn't recommend the use of revisions by default/as a matter of first principles for ambient IMO - for install or upgrade, simply because it's a lot of extra work and doesn't actually give you of the benefits you might expect or even need in all cases, if you are coming from sidecars.

We can mention that as a supported install/upgrade topology you _may_ want, depending on your env,  and put it under `/ops`.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
 
 {{< warning >}}
-In contrast to sidecar mode, ambient mode supports moving application pods to an upgraded data plane without a mandatory restart or reschedule of running application pods. However, upgrading the data plane **will** briefly disrupt all workload traffic on the upgraded node, and ambient mode does not currently support canary upgrades of the data plane.
+In contrast to sidecar mode, ambient mode supports moving application pods to an upgraded ztunnel without a mandatory restart or reschedule of running application pods. However, upgrading the ztunnel **will** cause all long-lived TCP connections on the upgraded node to reset, and ambient mode does not currently support canary upgrades of the ztunnel.
 
-Node cordoning and blue/green node pools are recommended to control blast radius of application pod traffic disruption during production upgrades. See your Kubernetes provider documentation for details.
+Node cordoning and blue/green node pools are recommended to limit the blast radius of resets on application traffic during production upgrades. See your Kubernetes provider documentation for details.
 {{< /warning >}}
 
+## Understanding Ambient Upgrades
```
ericvn, line 22: I'm going to say that I don't know the current capitalization guidelines for ambient, but I still believe they are lower case as is done in other places on this page. I have a few items where I lower case some instances. Feel free to take the advice, or keep what is there.
```suggestion
## Understanding ambient upgrades
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
+
+All Istio Upgrades involve upgrading the control plane, data plane, and Istio CRDs. Because the Ambient Data Plane is separated into [two components](/docs/ambient/architecture/data-plane), the ztunnel and waypoints, Ambient Upgrades involve separate steps for these components. Upgrading the control plane and CRDs will be covered here in brief, but is essentially identical to the process for these components in sidecar mode, which is covered extensively [here](docs/setup/upgrade/canary/).
```
ericvn, line 24:
```suggestion
All Istio Upgrades involve upgrading the control plane, data plane, and Istio CRDs. Because the ambient Data Plane is separated into [two components](/docs/ambient/architecture/data-plane), the ztunnel and waypoints, ambient upgrades involve separate steps for these components. Upgrading the control plane and CRDs will be covered here in brief, but is essentially identical to the process for these components in sidecar mode, which is covered extensively [here](docs/setup/upgrade/canary/).
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
+
+Like sidecar mode, ambient mode upgrades can make use of Tags and Revisions to allow fine-grained control over ({{< gloss >}}gateway{{</ gloss >}}) upgrades, including ({{< gloss >}}waypoint{{</ gloss >}})s, with simple controls for rolling back at any point. However, unlike sidecar mode, the Ambient ztunnel runs as a DaemonSet, or per-node proxy, meaning that ztunnel upgrades affect, at minimum, one entire node at a time. While this disruptions should be trivial for most use cases, applications with long-lived TCP connections may be disrupted.  In these cases, we recommend using Node cordoning and draining before upgrading the ztunnel component for a given node. For the sake of simplicity, this document will demonstrate in-place upgrades of the ztunnel, which are less safe, and may involve some downtime.
+
+With that in mind, this guide will expand on the following steps to upgrade Istio in ambient mode:
+
+1. Prerequisites
+1. Upgrade the CRDs
+1. Install the New Control Plane
```
ericvn, line 32: Match case for old.
```suggestion
1. Install the new Control Plane
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
-{{< /warning >}}
+    {{< text syntax=bash snip_id=update_helm >}}
+    $ helm repo update istio
+    {{< /text >}}
 
-### Manually upgrade the CRDs and Istio base chart
+1. Choose a Revision Name
+    Revisions identify unique instances of the Istio control plane, allowing you to run multiple distinct versions of the control plane simultaneously in a single mesh. It is recommended that revisions stay immutable, that is, that once a control plane is installed with a particular revision name, that installation not be modified, and that revision name not be reused. Tags, on the other hand, are mutable pointers to revisions. This enables a cluster operator to effect data plane upgrades without the need to adjust any workload labels, simply by moving a tag from one revision to the next. All data planes will connect only to one control plane, specified by the istio.io/rev label (pointing to either a revision or tag), or by the default revision if no istio.io/rev label is present. Upgrading a data plane consists of simply changing the control plane it is pointed to via modifying labels or editing tags.
```
ericvn, line 61:
```suggestion
    Revisions identify unique instances of the Istio control plane, allowing you to run multiple distinct versions of the control plane simultaneously in a single mesh. It is recommended that revisions stay immutable, that is, that once a control plane is installed with a particular revision name, that installation not be modified, and that the revision name not be reused. Tags, on the other hand, are mutable pointers to revisions. This enables a cluster operator to effect data plane upgrades without the need to adjust any workload labels, simply by moving a tag from one revision to the next. All data planes will connect only to one control plane, specified by the istio.io/rev label (pointing to either a revision or tag), or by the default revision if no istio.io/rev label is present. Upgrading a data plane consists of simply changing the control plane it is pointed to via modifying labels or editing tags.
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
+## Upgrade the Kubernetes custom resource definitions ({{< gloss >}}CRDs{{</ gloss >}})
 
-    {{< text syntax=bash snip_id=upgrade_base >}}
-    $ helm upgrade istio-base manifests/charts/base -n istio-system --skip-crds
+    {{< text bash >}}
+    $ kubectl apply -f manifests/charts/base/crds
     {{< /text >}}
 
-### Upgrade the Istio discovery Component
+## Install the New Control Plane
```
ericvn, line 78:
```suggestion
## Install the new Control Plane
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
dhawton: I think in titles we keep it capitalized, but I haven't seen any hard and fast rules... other than it's lowercase in typical use, but follows the language capitalization rules elsewhere (ie, beginning of sentences and titles).

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
+
 ## Prerequisites
 
-1. Update the Helm repository:
+1. Organize your Tags and Revisions
 
-    {{< text syntax=bash snip_id=update_helm >}}
-    $ helm repo update istio
-    {{< /text >}}
+    In order to safely upgrade Ambient, your Gateways and Namespaces should be using the `istio.io/rev` label to specify an Istio Tag which will control the version of the Istio Proxy that is running. We recommend dividing your production cluster into multiple tags to organize your upgrade. All members of a given tag will be upgraded simultaneously, so it is wise to begin your upgrade with your lowest risk applications. We do not recommend referencing revisions directly via labels for upgrades, as this process can easily result in the accidental upgrade of a large number of proxies, and is difficult to segment. To see what tags and revisions you are using in your cluster, see the section on upgrading tags.
```
ericvn, line 42:
```suggestion
    In order to safely upgrade ambient, your Gateways and Namespaces should be using the `istio.io/rev` label to specify an Istio Tag which will control the version of the Istio Proxy that is running. We recommend dividing your production cluster into multiple tags to organize your upgrade. All members of a given tag will be upgraded simultaneously, so it is wise to begin your upgrade with your lowest risk applications. We do not recommend referencing revisions directly via labels for upgrades, as this process can easily result in the accidental upgrade of a large number of proxies, and is difficult to segment. To see what tags and revisions you are using in your cluster, see the section on upgrading tags.
```

ericvn (commented):
I'm also not sure  on the capitalization of Gateway, Namespace and Tags in some places. There seems to be some inconsistency.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
dhawton: Agreed, gateways and namespaces here should be lowercase.

craigbox: And "data plane/control plane" throughout.

I can fix uses of "a mesh with support for ambient mode" etc.
This is just weird while it's gated and not on by default.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
craigbox: "ambient mode" in text, "Ambient mode" at the start of a sentence, whatever works in headings

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
     {{< /text >}}
 
-1. Upgrade the Istio base chart:
+## Upgrade the Kubernetes custom resource definitions ({{< gloss >}}CRDs{{</ gloss >}})
 
-    {{< text syntax=bash snip_id=upgrade_base >}}
-    $ helm upgrade istio-base manifests/charts/base -n istio-system --skip-crds
+    {{< text bash >}}
+    $ kubectl apply -f manifests/charts/base/crds
     {{< /text >}}
```
craigbox, line 76:
```suggestion
The cluster-wide Custom Resource Definitions (CRDs) must be upgraded prior to the deployment of a new version of the control plane:

{{< text bash >}}
$ kubectl apply -f manifests/charts/base/crds
{{< /text >}}
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
 
-### Upgrade the Istio discovery Component
+## Install the new Control Plane
```
craigbox, line 78:
```suggestion
## Install the new control plane
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
 
-Istiod is the control plane component that manages and configures the proxies to route traffic within an ambient mesh.
+Istiod is the control plane component that manages and configures the proxies to route traffic within an ambient mesh. The following command will install a new instance of this control plane alongside the old, but will not introduce any new proxies, or take over control for existing proxies. If you have previously customized your istiod installation, you can reuse the `values.yaml` file from previous upgrades or installs to keep your control planes consistent.
```
craigbox, line 80:
```suggestion
The [Istiod](/docs/ops/deployment/architecture/#istiod) control plane manages and configures the proxies that route traffic within the mesh. The following command will install a new instance of the control plane alongside the current, but will not introduce any new proxies, or take over control of existing proxies. 

If you have customized your istiod installation, you can reuse the `values.yaml` file from previous upgrades or installs to keep your control planes consistent.
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
+## Upgrade the ztunnel DaemonSet
 
-The ztunnel DaemonSet is the node proxy component.
-
-{{< warning >}}
-As ambient mode is not yet Stable, the following statement is not a compatibility guarantee, and is subject to change, or removal. Prior to reaching Stable status, this component and/or the control plane may receive breaking changes that prevent compatibility between minor versions.
-{{< /warning >}}
-
-The ztunnel at version 1.x is generally compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before ztunnel, as long as their version difference is within one minor version.
+The {{< gloss >}}ztunnel{{< /gloss >}} DaemonSet is the node proxy component. The ztunnel at version 1.x is compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before ztunnel, as long as their version difference is within one minor version. If you have previously customized your ztunnel installation, you can reuse the `values.yaml` file from previous upgrades or installs to keep your {{< gloss >}}data plane{{< /gloss >}} consistent.
```
craigbox, line 88:
```suggestion
The ztunnel `DaemonSet` is the node proxy component. ztunnel at version 1.x is compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before ztunnel, as long as their version difference is within one minor version. 

If you have previously customized your ztunnel installation, you can reuse the `values.yaml` file from previous upgrades or installs to keep your data plane consistent.
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
 
 {{< text syntax=bash snip_id=upgrade_ztunnel >}}
-$ helm upgrade ztunnel istio/ztunnel -n istio-system
+$ helm upgrade ztunnel istio/ztunnel -n istio-system --set revision="$REVISION" --wait
 {{< /text >}}
 
-### Upgrade the CNI component
+## Upgrade the CNI DaemonSet
 
 The Istio CNI agent is responsible for detecting pods added to the ambient mesh, informing ztunnel that proxy ports should be established within added pods, and configuring traffic redirection within the pod network namespace. It is not part of the data plane or control plane.
```
craigbox, line 101: This is not your edit, but surely it's part of the control plane?

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,103 @@ description: Upgrading an ambient mode installation with Helm.
...
 
-{{< warning >}}
-As ambient mode is not yet Stable, the following statement is not a compatibility guarantee, and is subject to change, or removal. Prior to reaching Stable status, this component and/or the control plane may receive breaking changes that prevent compatibility between minor versions.
-{{< /warning >}}
-
-The CNI at version 1.x is generally compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before Istio CNI, as long as their version difference is within one minor version.
+The CNI at version 1.x is  compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before Istio CNI, as long as their version difference is within one minor version.
```
craigbox, line 103:
```suggestion
Similar to ztunnel, the CNI agent at version 1.x is compatible with the control plane at version 1.x+1 and 1.x. This means the control plane must be upgraded before the CNI, agent as long as their version difference is within one minor version.
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
craigbox: Why not just `base` btw?  Do roles ever change?

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -103,38 +110,38 @@ Upgrading the Istio CNI agent to a compatible version in-place will not disrupt
 $ helm upgrade istio-cni istio/cni -n istio-system
 {{< /text >}}
 
-### Upgrade the Gateway component (optional)
+## Upgrade waypoints and gateways using tags
 
-Gateway components manage east-west and north-south dataplane traffic between ambient mesh boundaries, as well as some aspects of the L7 dataplane.
+If you have followed best practices, all of your gateway, workload, and namespaces use either the default revision (which is effectively a tag named default), or the `istio.io/rev` label with the value set to a tag name. We will now upgrade all of these data planes to the new version of Istio by moving their tags to point to the new version, one at a time. To list all tags in your cluster, run:
```
craigbox, line 115:
```suggestion
If you have followed best practices, all of your gateways, workloads, and namespaces use either the default revision (effectively, a tag named `default`), or the `istio.io/rev` label with the value set to a tag name. You can now upgrade all of these to the new version of the Istio data plane by moving their tags to point to the new version, one at a time. To list all tags in your cluster, run:
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -103,38 +110,38 @@ Upgrading the Istio CNI agent to a compatible version in-place will not disrupt
...
+{{< text syntax=bash snip_id=upgrade_tag >}}
+$ helm template istiod istio/istiod -s templates/revision-tags.yaml --set revisionTags="{$MYTAG}" --set revision="$REVISION" -n istio-system | kubectl apply -f -
 {{< /text >}}
 
-## Verify the installation
-
-### Verify the workload status
-
-After installing all the components, you can check the Helm deployment status with:
+This will upgrade all dataplanes referencing that tag, except for those using [manual gateway deployment mode](docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment), which are dealt with below, and sidecars, which are not used in Ambient mode. It is recommended that you closely monitor the health of applications using the upgraded dataplane before upgrading the next tag. If you detect a problem, you can rollback a tag, resetting it to point to the name of your old revision:
```
craigbox, line 127:
```suggestion
This will upgrade all objects referencing that tag, except for those using [manual gateway deployment mode](docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment), which are dealt with below, and sidecars, which are not used in ambient mode. 

It is recommended that you closely monitor the health of applications using the upgraded data plane before upgrading the next tag. If you detect a problem, you can rollback a tag, resetting it to point to the name of your old revision:
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -103,38 +110,38 @@ Upgrading the Istio CNI agent to a compatible version in-place will not disrupt
...
-{{< text syntax=bash snip_id=show_components >}}
-$ helm list -n istio-system
+{{< text syntax=bash snip_id=rollback_tag >}}
+$ helm template istiod istio/istiod -s templates/revision-tags.yaml --set revisionTags="{$MYTAG}" --set revision="$OLD_REVISION" -n istio-system | kubectl apply -f -
 {{< /text >}}
 
-You can check the status of the deployed Istio pods with:
+### Upgrade the Gateway component (optional)
+
+Gateway components using [manual gateway deployment mode](docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment) must be upgraded individually using helm.
```
craigbox, line 135:
```suggestion
### Upgrade manually deployed gateways (optional)

`Gateway`s that were [deployed manually](docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment) must be upgraded individually using Helm:
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -103,38 +110,38 @@ Upgrading the Istio CNI agent to a compatible version in-place will not disrupt
...
-$ kubectl get pods -n istio-system
+{{< text syntax=bash snip_id=upgrade_gateway >}}
+$ helm upgrade istio-ingress istio/gateway -n istio-ingress
 {{< /text >}}
 
-## Uninstall
+## Uninstall the previous control plane
 
-Please refer to the uninstall section in the [Helm installation guide](/docs/ambient/install/helm-installation/#uninstall).
+If you have upgraded all dataplane components to use the new version of Istio, and are satisfied that you do not need to rollback, you can remove the previous version of the control plane by running:
```
craigbox, line 143:
```suggestion
If you have upgraded all data plane components to use the new version of Istio, and are satisfied that you do not need to roll back, you can remove the previous version of the control plane by running:
```

content/en/docs/ambient/upgrade/helm-upgrade/index.md
craigbox: Mitch, should we make this a separate page?

(We should also think about how we rationalise all these docs when ambient is GA/on by default/etc)

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: helm skips applying CRDs, so these have to be directly applied.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: I can't tell from your comment what the 'it' is that you are asking about.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: IIRC We do not yet support ztunnel skew.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: This is incorrect.  This step is only for gateways that are not managed by istiod.  Others are upgraded above in the tag step.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: This gloss was requested by a previous reviewer, so I'm hesitant to remove it.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,98 @@ description: Upgrading an ambient mode installation with Helm.
...
 
-1. Update the Helm repository:
+Like sidecar mode, ambient mode upgrades can make use of [revision tags](https://istio.io/latest/docs/setup/upgrade/canary/#stable-revision-labels) to allow fine-grained control over ({{< gloss >}}gateway{{</ gloss >}}) upgrades, including waypoints, with simple controls for rolling back at any point. However, unlike sidecar mode, the ztunnel runs as a DaemonSet — a per-node proxy — meaning that ztunnel upgrades affect, at minimum, an entire node at a time. While this may be acceptable in many cases, applications with long-lived TCP connections may be disrupted.  In such cases, we recommend using node cordoning and draining before upgrading the ztunnel for a given node. For the sake of simplicity, this document will demonstrate in-place upgrades of the ztunnel, which may involve a short downtime.
```
linsun, line 26: Like sidecar mode, ambient mode upgrades

to

Like sidecar mode, waypoint proxy upgrades

?

linsun: because you can't use revision for cni or ztunnel.

linsun: `which may involve a short downtime.`

- we should frame it as similarly as when you update the node... and you should budget for deploying your apps on more than 1 nodes.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -4,96 +4,98 @@ description: Upgrading an ambient mode installation with Helm.
...
+## Prerequisites
 
-    {{< text syntax=bash snip_id=update_helm >}}
-    $ helm repo update istio
-    {{< /text >}}
+### Organize your tags and revisions
 
-## In-place upgrade
+In order to safely upgrade a mesh in ambient mode, your gateways and namespaces should be using the `istio.io/rev` label to specify a revision tag which controls the version of the proxy that is running. We recommend dividing your production cluster into multiple tags to organize your upgrade. All members of a given tag will be upgraded simultaneously, so it is wise to begin your upgrade with your lowest risk applications. We do not recommend referencing revisions directly via labels for upgrades, as this process can easily result in the accidental upgrade of a large number of proxies, and is difficult to segment. To see what tags and revisions you are using in your cluster, see the section on upgrading tags.
```
linsun, line 31: Agree with some other the other comments, we should not remove in-place upgrade. A good portion of our users are still using it for sidecars as long as they are okay with the limitations.

therealmitchconnors: Yes, John is going to add an in-place upgrade tab after this PR merges.  I don't have time to add that, as this PR has been in review for almost two months.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: The downtime language has already been revised three times by three other reviewers, and I'm hesitant to change it again and risk breaking consensus...  I'll get to the first comment in my next commit.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -103,38 +107,40 @@ Upgrading the Istio CNI agent to a compatible version in-place will not disrupt
...
-### Upgrade the Gateway component (optional)
+## Upgrade waypoints and gateways using tags
 
-Gateway components manage east-west and north-south dataplane traffic between ambient mesh boundaries, as well as some aspects of the L7 dataplane.
+If you have followed best practices, all of your gateways, workloads, and namespaces use either the default revision (effectively, a tag named `default`), or the `istio.io/rev` label with the value set to a tag name. You can now upgrade all of these to the new version of the Istio data plane by moving their tags to point to the new version, one at a time. To list all tags in your cluster, run:
 
-{{< text syntax=bash snip_id=upgrade_gateway >}}
-$ helm upgrade istio-ingress istio/gateway -n istio-ingress
+{{< text syntax=bash snip_id=list_tags >}}
+$ kubectl get mutatingwebhookconfigurations -l 'istio.io/tag' -L istio\.io/tag,istio\.io/rev
```
linsun, line 115: No equivalent istioctl cmd for it? :-(. would `istioctl tag list` work?

content/en/docs/ambient/upgrade/helm-upgrade/index.md
```diff
@@ -103,38 +107,40 @@ Upgrading the Istio CNI agent to a compatible version in-place will not disrupt
...
-{{< text syntax=bash snip_id=show_istiod_values >}}
-$ helm show values istio/istiod
+{{< text syntax=bash snip_id=upgrade_tag >}}
+$ helm template istiod istio/istiod -s templates/revision-tags.yaml --set revisionTags="{$MYTAG}" --set revision="$REVISION" -n istio-system | kubectl apply -f -
 {{< /text >}}
 
-## Verify the installation
-
-### Verify the workload status
+This will upgrade all objects referencing that tag, except for those using [manual gateway deployment mode](docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment), which are dealt with below, and sidecars, which are not used in ambient mode.
```
linsun, line 124: I assume this is in place upgrade for waypoints and gateways?  worth clarifying as you mentioned revision based upgrade earlier.

linsun (commented):
Thank you @therealmitchconnors for proposing the change and add revision based CP instruction in.

Could I propose to have 2 upgrade instructions similar as sidecars, both using helm like you have? one for in place, one for revision?

https://istio.io/latest/docs/setup/upgrade/

content/en/docs/ambient/upgrade/helm-upgrade/index.md
linsun: Ok SG, you can ignore my new comment about in-place upgrade, before seeing yours. :)

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: Since this is a helm-based guide, we chose to reference istioctl in as few locations as possible.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
therealmitchconnors: I'm not sure I understand this question.  This is a canary upgrade of all gateways referencing the tag, and does not require any manual restart of workloads...

therealmitchconnors:
> Thank you @therealmitchconnors for proposing the change and add revision based CP instruction in.
> 
> Could I propose to have 2 upgrade instructions similar as sidecars, both using helm like you have? one for in place, one for revision?
> 
> https://istio.io/latest/docs/setup/upgrade/

Yes, John is working on the other instruction set.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
bleggett: We are stuck `getting` a MWH config because `istioctl` does that under the hood - it is A Problem that `istioctl` masks for us, and any other tool will not.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
craigbox: I will have done that because you should gloss on first mention.

content/en/docs/ambient/upgrade/helm-upgrade/index.md
craigbox: "The CNI component"... "is not part of the data plane or control plane"

craigbox (approved):

Issue #9859, opened 2021-05-31
Instructions for kOps setup are unnecessary 


johngmyers:
The instructions for configuring kOps given in content/en/docs/setup/platform-setup/kops/index.md are no longer necessary as of kOps 1.20.

See https://kops.sigs.k8s.io/operations/service_account_token_volumes/

howardjohn:
thanks,ill update

Issue #15234, opened 2024-06-04
improve kind installation page


craigbox:
https://istio.io/latest/docs/setup/platform-setup/kind/

The new target audience for this page is a Getting Started user.

We should rewrite:

- to not need to know about your Go version
- s/MetalLB/Cloud Provider kind/, and only if you _really_ need it
- without concern for the Kubernetes Dashboard

and/or we should use Minikube instead.

peterj:
A couple of common themes across all platforms:
- run a cluster large enough to run Istio + bookinfo (e.g. 8gb mem requirement from docker desktop, minikube, etc.)
- different ways of exposing ingress gateway, getting the IP, etc.

Other than that, the instructions are not in any way specific to Istio and they are mostly about setting up a k8s cluster.

| Platform | Notes | 
| --- | --- |
| Alibaba cloud | Outdated (i.e. "hasn't been tested on 1.9") + no Istio specific config |
| Amazon EKS |  no Istio specific config, links to EKS blueprints |
| Azure | no Istio specific config |
| Docker Desktop | no Istio specific config |
| GKE | Has a note for private GKE cluster + section for multi-cluster communication (which isn't necessarily istio specific) |
| Huawei | Notes on resources, but otherwise no Istio specific config |
| IBM | no Istio specific config |
| k3d | command to set up a k3d cluster without traefik and to expose certain ports (I think) |
| kind | no Istio specific config |
| Kops | no Istio specific config (unless you're  k8s < 1.22) |
| Kubernetes Gardener | no Istio specific config |
| KubeSphere Container Platform | Outdated (i.e. "hasn't been tested on 1.9") + no Istio specific config |
| MicroK8s | no Istio specific config |
| Minikube | Memory requirements, no Istio specific config |
| Openshift | no Istio specific config |
| OCI | Outdated, no Istio specific config |
| Tencent | no Istio specific config |

craigbox:
I've been wanting to collapse these down to one page.  Let's do that.

peterj:
All right, I'll collapse this down.

peterj:
Before I go further down the hole, would the FAQ layout work well for this page, @craigbox? 


![Uploading Screenshot 2024-06-05 at 1.26.09 PM.png…]()

craigbox:
Maybe?  Depends how big the titles are under each section, I guess.  
Hopefully it's just a small toggle to change each and we'll see what they look like

peterj:
[...]

PR #15258, opened 2024-06-06
Update stale kops docs

howardjohn:
Fixes https://github.com/istio/istio.io/issues/9859

No need to document unsupported versions


craigbox:
All these are being removed/condensed in #15234.

Issue #15167, opened 2024-05-22
Redo ambient install guides to use ambient wrapper charts for 1.23


bleggett:
Context: https://github.com/istio/istio/pull/51174

PR #15279, opened 2024-06-13
Document use of ambient wrapper chart in 1.23 install/upgrade guides

bleggett:
## Description
Fixes https://github.com/istio/istio.io/issues/15167

Note that until at least a 1.23 prerelease build is published, the `ambient` chart will not show up in the documented repo, see https://github.com/istio/release-builder/pull/1849

To test locally in the meantime, you can use the local copies of the chart, so instead of doing:

```
helm install istio-ambient istio/ambient -n istio-system --create-namespace --wait
```
like the doc says, do 

```
helm install ambient ./manifests/charts/ambient -n istio-system --create-namespace --wait
```
(you may need to run `helm dep update ./manifests/charts/ambient` first to populate the depcache)

I left both the individual and wrapper modes in both docs.

## Reviewers

- [x] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/getting-started/_index.md
+++ b/content/en/docs/ambient/getting-started/_index.md
@@ -10,62 +10,123 @@ skip_list: true
 test: yes
 ---
 
-This guide lets you quickly evaluate Istio's {{< gloss "ambient" >}}ambient mode{{< /gloss >}}. You'll need a Kubernetes cluster to proceed. If you don't have a cluster, you can use [kind](/docs/setup/platform-setup/kind) or any other [supported Kubernetes platform](/docs/setup/platform-setup).
+This guide shows you how to install Istio in ambient mode with Helm, in a non-production quick-start context.
 
-These steps require you to have a {{< gloss >}}cluster{{< /gloss >}} running a
-[supported version](/docs/releases/supported-releases#support-status-of-istio-releases) of Kubernetes ({{< supported_kubernetes_versions >}}).
+If you are looking to install Istio's ambient mode in a production context, we recommend following the more [advanced Helm install guide](/docs/ambient/install/helm-installation/).
 
-## Download the Istio CLI
+We generally encourage the use of Helm to install Istio when using ambient mode. Helm helps you manage components separately, is widely compatible with other tooling, and allows you to independently install and upgrade individual Istio components if desired.
 
-Istio is configured using a command line tool called `istioctl`.  Download it, and the Istio sample applications:
+## Prerequisites
 
-{{< text syntax=bash snip_id=none >}}
-$ curl -L https://istio.io/downloadIstio | sh -
-$ cd istio-{{< istio_full_version >}}
-$ export PATH=$PWD/bin:$PATH
-{{< /text >}}
+1. Check the [Platform-Specific Prerequisites](/docs/ambient/install/platform-prerequisites).
 
-Check that you are able to run `istioctl` by printing the version of the command. At this point, Istio is not installed in your cluster, so you will see that there are no pods ready.
+1. [Install the Helm client](https://helm.sh/docs/intro/install/), version 3.6 or above.
 
-{{< text syntax=bash snip_id=none >}}
-$ istioctl version
-no ready Istio pods in "istio-system"
-{{< istio_full_version >}}
-{{< /text >}}
+1. Configure your Helm client to pull from the Istio Helm repository:
 
-## Install Istio on to your cluster
+    {{< text syntax=bash snip_id=configure_helm >}}
+    $ helm repo add istio https://istio-release.storage.googleapis.com/charts
+    $ helm repo update
+    {{< /text >}}
 
-`istioctl` supports a number of [configuration profiles](/docs/setup/additional-setup/config-profiles/) that include different default options, and can be customized for your production needs. Support for ambient mode is included in the `ambient` profile. Install Istio with the following command:
+*See [the helm repo documentation](https://helm.sh/docs/helm/helm_repo/) for command documentation.*
 
-{{< text syntax=bash snip_id=install_ambient >}}
-$ istioctl install --set profile=ambient --skip-confirmation
-{{< /text >}}
+Istio is made up of multiple components which belong to either the {{< gloss >}}data plane{{< /gloss >}} or the {{< gloss >}}control plane{{< /gloss >}}. Different components may have different effects on your applications and cluster networking when upgraded.
 
-It might take a minute for the Istio components to be installed. Once the installation completes, you’ll get the following output that indicates all components have been installed successfully.
-
-{{< text syntax=plain snip_id=none >}}
-✔ Istio core installed
-✔ Istiod installed
-✔ CNI installed
-✔ Ztunnel installed
-✔ Installation complete
-{{< /text >}}
+For installing Istio with support for the {{< gloss >}}ambient{{< /gloss >}} data plane mode quickly in a non-production cluster, we provide a simple Helm chart which bundles all the required Istio component Helm charts.
 
 {{< tip >}}
-You can verify the installed components using the command `istioctl verify-install`.
+For production deployments, it is strongly recommended to install the components separately via Helm to allow for more control during Istio upgrades. Consult the [full Helm install guide](/docs/ambient/install/helm-installation/) as well as the [operations guidelines](/docs/ops/best-practices/deployment/) for production deployment considerations.
 {{< /tip >}}
 
-## Install the Kubernetes Gateway API CRDs
+## Install
 
-You need to install the Kubernetes Gateway API CRDs, which don’t come installed by default on most Kubernetes clusters:
+The `ambient` Helm chart composes all the components that enable the use of Istio's ambient data plane mode:
 
-{{< text syntax=bash snip_id=install_k8s_gateway_api >}}
-$ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-  { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v1.1.0" | kubectl apply -f -; }
+{{< text syntax=bash snip_id=install_ambient_chart >}}
+$ helm install istio-ambient istio/ambient -n istio-system --create-namespace --wait
 {{< /text >}}
 
-You will use the Kubernetes Gateway API to configure traffic routing.
+Proceed to [verifying the installation](#verify-the-installation) or [installing an ingress gateway (optional)](#install-an-ingress-gateway-optional)
 
-## Next steps
+## Install an ingress gateway (optional)
 
-Congratulations! You've successfully installed Istio with support for ambient mode. Continue to the next step to [install the demo application and add it to the ambient mesh](/docs/ambient/getting-started/deploy-sample-app/).
+To install an ingress gateway, run the command below:
+
+{{< text syntax=bash snip_id=install_ingress >}}
+$ helm install istio-ingress istio/gateway -n istio-ingress --create-namespace --wait
+{{< /text >}}
+
+If your Kubernetes cluster doesn't support the `LoadBalancer` service type (`type: LoadBalancer`) with a proper external IP assigned, run the above command without the `--wait` parameter to avoid the infinite wait. See [Installing Gateways](/docs/setup/additional-setup/gateway/) for in-depth documentation on gateway installation.
+
+## Verify the installation
+
+### Verify the workload status
+
+After installing all the components, you can check the Helm deployment status with:
+
+{{< text syntax=bash snip_id=show_components >}}
+$ helm ls -n istio-system
+NAME            NAMESPACE       REVISION    UPDATED                                 STATUS      CHART           APP VERSION
+istio-base      istio-system    1           2024-04-17 22:14:45.964722028 +0000 UTC deployed    base-{{< istio_full_version >}}     {{< istio_full_version >}}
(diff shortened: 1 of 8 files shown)


content/en/docs/ambient/install/helm-installation/index.md
```diff
@@ -28,7 +28,21 @@ we encourage the use of Helm to install Istio for use in ambient mode. Helm help
 
 *See [helm repo](https://helm.sh/docs/helm/helm_repo/) for command documentation.*
 
-## Install the components
+{{< gloss "ambient" >}}Ambient{{< /gloss >}} is made up of multiple components - for production deployments, it is strongly recommended to install the components separately to allow for more control during upgrades. Consult the [operations guidelines](/docs/ops/deployment) for production deployment considerations.
```
craigbox, line 31:
```suggestion
Istio comprises many components (control plane, data plane, custom resources, etc). For production deployments, it is strongly recommended to install the components separately to allow for fine grained control during upgrades. Consult the [operations guidelines](/docs/ops/deployment) for production deployment considerations.
```

so is sidecar mode.  At least three of the charts are shared?

Does the operations docs actually talk about what you should do differently?  Is it a scaling suggestion? (You wouldn't scale istio-cni and I doubt you'd scale ztunnel)

content/en/docs/ambient/install/helm-installation/index.md
```diff
@@ -28,7 +28,21 @@ we encourage the use of Helm to install Istio for use in ambient mode. Helm help
...
+
+If you just want to try out Istio ambient in a non-production cluster, to keep things simple it is recommended to use the `ambient` Helm wrapper chart which installs the minimum set of required components together.
```
craigbox, line 33:
```suggestion
For installing Istio with support for the {{< gloss "ambient" >}}ambient data plane mode{{< /gloss >}} in a non-production cluster, we provide a simple Helm chart which bundles all the required components.
```

content/en/docs/ambient/install/helm-installation/index.md
```diff
@@ -28,7 +28,21 @@ we encourage the use of Helm to install Istio for use in ambient mode. Helm help
...
+
+## Install the components together
```
craigbox, line 35:
```suggestion
## Simple install
```

(or similar)

content/en/docs/ambient/install/helm-installation/index.md
```diff
@@ -28,7 +28,21 @@ we encourage the use of Helm to install Istio for use in ambient mode. Helm help
...
+
+The `ambient` Helm wrapper chart composes all the components that make up an ambient installation into one Helm chart:
```
craigbox, line 37: is "wrapper chart" a term of art?  I don't see it in the Helm docs

craigbox, line 37:
```suggestion
The `ambient` Helm chart composes all the components that support Istio with the ambient data plane:
```

content/en/docs/ambient/install/helm-installation/index.md
```diff
@@ -114,8 +138,48 @@ After installing ambient mode with Helm, you can follow the [Deploy the sample a
 
 ## Uninstall
 
-You can uninstall Istio and its components by uninstalling the charts
-installed above.
+If you installed with the `ambient` wrapper chart above, you can uninstall Istio and its components by uninstalling that chart:
```
craigbox, line 141:
```suggestion
If you installed with the simple `ambient` chart above, you can uninstall Istio and its components by uninstalling that chart:
```

maybe don't put this before the gateway instructions?

content/en/docs/ambient/install/helm-installation/index.md
```diff
@@ -114,8 +138,48 @@ After installing ambient mode with Helm, you can follow the [Deploy the sample a
...
+    [By design](https://github.com/helm/community/blob/main/hips/hip-0011.md#deleting-crds),
+    deleting a chart via Helm doesn't delete the installed Custom
+    Resource Definitions (CRDs) installed via the chart.
+    {{< /tip >}}
+
+    {{< text syntax=bash snip_id=delete_ambient_wrapper >}}
+    $ helm delete istio-ambient -n istio-system --wait
+    {{< /text >}}
+
+1. Delete the `istio-system` namespace:
```
craigbox, line 164: this isn't done automatically? 🤔

content/en/docs/ambient/install/helm-installation/index.md
bleggett: > so is sidecar mode. At least three of the charts are shared?

Yeah, this is sort of boilerplate. I'll enhance to mention that some components have different impacts on app traffic and node health than others (data plane vs control plane, really) and this impacts upgrade cadence considerations. 


> Does the operations docs actually talk about what you should do differently? Is it a scaling suggestion? (You wouldn't scale istio-cni and I doubt you'd scale ztunnel)

Scaling, and upgrade cadence.

Really, this should link to

https://preliminary.istio.io/latest/docs/ops/best-practices/deployment/

and that should be built out. I'll do that as part of this.

content/en/docs/ambient/install/helm-installation/index.md
bleggett: very much not.

This is a relatively common pattern (used in Argo, etc) that doesn't really have a specific Helm definition.

We could say "parent chart" or "composite chart" but to Helm it's just a Helm chart with Helm dependencies (and nothing else)

content/en/docs/ambient/install/helm-installation/index.md
bleggett: It would be, except for the whole "you have to delete CRDs manually" thing, which Helm categorically refuses to do automatically, because kubernetes effectively provides no way to do it safely and unconditionally

content/en/docs/ambient/install/helm-installation/index.md
bleggett: Yeah the idea was to exactly reverse the install directions but I dorked the wording here.

content/en/docs/ambient/install/helm-installation/index.md
bleggett: Actually - what I was kinda hoping is that we could just link to wherever this https://github.com/istio/istio.io/pull/15247 ends up (under `/ops/` or wherever) as the "recommended/opinionated production setup".

content/en/docs/ambient/install/helm-installation/index.md
bleggett: I've just removed `wrapper` as a term, and talked around the informal term. Open to better ideas if anyone has them.

linsun:
Do we want to have this in get started guide as the wrapper chart is mainly to show how simple it is to install Istio for demo purpose?

Just to recap some discussion from Monday's TOC, unfortunately the attendance is relative low, but folks on the call @therealmitchconnors @keithmattix @howardjohn and myself prefer to have the simple 1 helm install for demo purpose (or consider using [helmfile](https://helmfile.readthedocs.io/en/latest/) as an alternative if we choose not to publish the wrapper chart). One primary concern is not to show people multiple ways to do Install within 1 install method (which is helm here) and the existing helm install instructions offer more flexibility than the wrapper chart.  Please correct me if I am wrong here in capturing the discussion.

bleggett:
> Do we want to have this in get started guide as the wrapper chart is mainly to show how simple it is to install Istio for demo purpose?

:+1: I'll make that change.

howardjohn:
The other key part of the TOC discussion was to move this chart wrapper under `samples/`. We can still publish to OCI (and probably should!), but under "samples" in some way (could be gcr.io/istio-release/samples/...)

bleggett:
> The other key part of the TOC discussion was to move this chart wrapper under `samples/`. We can still publish to OCI (and probably should!), but under "samples" in some way (could be gcr.io/istio-release/samples/...)

:+1: I assume that's `release-builder` changes, will do.

bleggett:
@linsun @craigbox et al - PTAL:

- Moved the `simple Helm` instructions to `gettting started`
- Left the `full Helm` instructions in `helm install`

tweaked wording and such.

content/en/docs/ambient/getting-started/_index.md
craigbox, line 1: Oh hello, I don't think we should be using `helm` in the Getting Started guide.  This has to be zero to installed in as quick a time as possible.

I'd love to get updates to the Helm page in before making some updates of my own though.  Could we remove this for now and just get the others?

craigbox:
Test failure is because the extend-wasm guide uses the Gateway API install docs from the getting started page, which were removed when we went to Helm.

content/en/docs/ambient/getting-started/_index.md
bleggett: >  This has to be zero to installed in as quick a time as possible.

Yes, that was the intent behind this PR. `helm install istio-ambient istio/samples/ambient -n istio-system --create-namespace --wait` being analogous to `istioctl install --profile=ambient`

I was looking at the `getting started` doc as the simplest possible/most normative set of steps for the non-Istio user (who probably already uses Helm).


I had the wrapper chart in the regular Helm install guide before, and was asked to move it out: https://github.com/istio/istio.io/pull/15279#issuecomment-2220493410

The options here are:

1. Drop the wrapper chart people have asked for, and  (IMO) continue to have a needlessly substandard Helm experience, because we don't recommend/test/use/compose Helm charts the way our users do in any capacity.
2. Put the wrapper chart guide in `quickstart`
3. Put the wrapper chart guide in `Helm install`

@linsun @howardjohn and @craigbox you all decide, and I'll wait for that.

bleggett:
> Test failure is because the extend-wasm guide uses the Gateway API install docs from the getting started page, which were removed when we went to Helm.

Yeah I fixed some of this in https://github.com/istio/istio.io/pull/15474 - that one should probably go in before this one, and I'll resync this one.

content/en/docs/ambient/getting-started/_index.md
howardjohn: I think @craigbox's comment was just not to put anything 'helm' in quickstart, likely on a disagreement that its the 'simplest set of steps for the non-Istio user'.

I do tend to agree for a _quickstart_ Helm is likely to only cause friction. `istioctl` is already something a user will install as part of the getting started (used in other parts), so its a decent choice -- as is `kubectl apply` which is what we used to do. 

I don't think that means we need to drop the wrapper chart, though, we can just say you can install with helm in a aggregated or decomposed manner based on your needs

craigbox: There is a disconnect here between the TOC and the "docs WG" (me), who is asleep when TOC meetings happen. I'm sorry to Ben for it having dragged him in two directions.

The goal of the "Getting Started" page is the easiest possible experience for someone looking at Istio for the first time.  We're debating on Slack if this is `helm` or if it is `istioctl`, but until now we've assumed it was `istioctl`.  Peter and I have put some effort into improving this recently as unfortunately it is a measure by which the "complexity of Istio" has been reported.

There is also a meme of "Istio has 7 Helm charts", which I am all in favour of resolving.  I think that the Helm install page should point out that you can install everything with one chart, or use separate charts, and lay out the case for why you should actually do the more complex thing.  

I _would_ like the TOC (or whoever; Environments WG, Ben?) to decide on the order of recommendation for helm and istioctl, and that might dictate the choice in the Getting Started guide.

howardjohn: The question is -- recommendation for _what goal_? 

For a serious production install, where we are optimizing for reliability and safety, Helm is very likely the best choice at this point. `istioctl` isn't bad or anything, just doesn't integrate with tooling.

For "just give me an env with it deployed so I can try it out", helm is probably not the best...

To put it another way -- when I develop on Istio, I don't use helm since it is tedious to use.

bleggett: > To put it another way -- when I develop on Istio, I don't use helm since it is tedious to use.

I don't use it _with Istio_ either because it is tedious to use _with Istio_ for reasons that have to do _with Istio_ - I have and do use it in dev flows for other things just fine - Helm probably has more daily users than Istio does.

But all this brings up a very good point - the people developing Istio day-to-day will shave the edges of of _any_ install mechanism we pick, no matter which is picked, in much the same way that water wears down rocks. 

If we don't subject our Helm to that direct daily wear it will continue to be an afterthought and tedious to use. The same is true of istioctl.

I am convinced that the people developing Istio day-to-day (and I am including myself in this) are probably not qualified to make this decision for those reasons, and are too close to the problem to make a good assessment.

linsun: Looking at the PR again, there are more lines (2X roughly i think) to read for users in the helm instruction with wrapper chart than istioctl... so I'd definitely continue to vote for istioctl.  After reading the helm instruction, I lean towards to put helm get started instruction using wrapper chart as a link to be linked from the get started with ambient page.

linsun:
FYI, I added this PR to ambient meeting agenda for tmr.

bleggett:
Aight so current consensus: https://docs.google.com/document/d/1SMlwliEnthgq7r2PjpLl1kCq3t8rAMbgu6r_lDAXJ0w/edit#heading=h.ttmnrtyxr98j

- we aren't going to document the wrapper chart (gonna close this PR)
- The wrapper chart is published under `samples`: https://github.com/istio/release-builder/pull/1901 but I am no longer comfortable publishing an artifact we do not document or doctest (which is why this PR was opened) - for now there seems to be a desire to leave it published, so I will leave it.

PR #15315, opened 2024-06-23
remove wasm telemetry

zirain:
this's unnecessary any more, cc @kyessenov @lei-tang


kfaseela:
This is an old blog post though, do we want to change?

kfaseela (approved):

zirain:
istio stats moved to native cpp filter for a long time.

Issue #15319, opened 2024-06-24
Add FAQ on "which api version should I use"


howardjohn:
See https://blog.howardjohn.info/posts/crd-versioning/

tl;dr: it usually doesn't matter as much as people think. Always use the latest if you are a user; use the oldest version you need to support for integration authors

kfaseela:
i am wondering why we still have community/good first issue label.. I had actually removed it as part of https://github.com/istio/bots/pull/762

priyanshsao:
hey @howardjohn , i would like to work on this issue , since i am  new to open source , can you provide me some guidance for solving this issue.

kfaseela:
@priyanshsao : There is already a PR up for review for this from @darshan174

craigbox:
@priyanshsao Please see https://github.com/istio/istio.io/pull/15329#issuecomment-2191280978.

I don't actually think this is a good first issue, because you need to understand (a) something pretty complex about Kubernetes (b) how Istio implements it (c) how to explain it trivially to a new user.

I should be able to do all 3, but I suspect it will take me an hour or so of thought to get something that is well crafted.

PR #15329, opened 2024-06-25
Added an Faq on api versioning

Darshan174:
## Description

Added an Faq on api versioning.

Fixes https://github.com/istio/istio.io/issues/15319

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [X] Networking
- [ ] Security
- [ ] Test and Release
- [X] User Experience

### DIFF
--- a/content/en/about/faq/traffic-management/api.md
+++ b/content/en/about/faq/traffic-management/api.md
@@ -3,8 +3,5 @@ title: Which API version should I use?
 weight: 10
 ---
 
-Always use the primary/latest API version declared in the CRD, as this will give you the most up-to-date functionality and the system will handle any necessary version conversions transparently.
+Any version will give you the most up-to-date functionality and it doesn't really matter which version is used. There is no reason to use a specific version (unless you are an integration author, then checkout this [link](https://istio.io/latest/docs/ops/integrations/)), and hence there no need to spend time migrating existing resources or anything like that.
 
-One can also use the oldest API version that you need to support, to ensure compatibility with all your users for integration authors, the system will handle converting the requests/responses to the primary/latest version behind the scenes.
-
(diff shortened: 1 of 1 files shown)


kfaseela:
Hello @Darshan174 please sign the cla

Darshan174:
> Hello @Darshan174 please sign the cla

Hey @kfaseela! I've signed the cla

kfaseela:
@howardjohn let me know if this is good to go

content/en/about/faq/traffic-management/api.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Which API version should I use?
+weight: 10
+---
+
+Always use the primary/latest API version declared in the CRD, as this will give you the most up-to-date functionality and the system will handle any necessary version conversions transparently.
```
howardjohn, line 6: This is not quite true, any version gives up-to-date functionality. 

I think we want to convey a message that it _it really doesn't matter_. There is no reason not to use the latest (unless you are an integration author, then see below), but there is also no strong need to spending time migrating existing resources or anything like that

Darshan174: Hey @howardjohn, is this [link](https://istio.io/latest/docs/ops/integrations/) sufficient for displaying info on integration author in the FAQ

craigbox:
@Darshan174 Thanks for contributing.  Docs are sometimes an easy place to start, but you really have to understand a system to write something in the language a new user to it will be able to comprehend.

I don't think what we have is going to help someone understand the problem, and I think that it is because you, as the author, don't understand the problem.

I would recommend you consider getting started with code contributions; perhaps try looking at the `istio/istio` repo for good first issues there.

Issue #15148, opened 2024-05-20
migrate from markdownlint (ruby) to markdownlint (node JS)


craigbox:
Either of these:

* https://github.com/igorshubovych/markdownlint-cli
* https://github.com/DavidAnson/markdownlint-cli2

use the engine https://github.com/DavidAnson/markdownlint, which is based on https://github.com/markdownlint/markdownlint but seems more actively maintained.

Replacing our existing implementation with one of them will allow to access useful behaviour such as [ignoring rules](https://discourse.gohugo.io/t/ignore-a-line-with-markdownlint/44269).

ayushrakesh:
@craigbox I want to work in this issue, can you please elaborate more about how this can be done. I am not able to understand properly.

rashmi43:
I would like to take up this issue

rashmi43:
https://github.com/istio/istio.io/pull/15360

craigbox:
Please note that the issue here refers to changing the engine used to verify all the istio.io docs.  It's a systems engineering/integration task, not a configuration or docs task.

rashmi43:
todos:
You would need to learn how the current script/build tools use the Ruby version of markdownlint, update them to package and use the Node.JS version, and then update the configuration to match our existing rules.

You may also need to do a mass update of the docs to fix any behaviour we can't configure around.

rashmi43:
figured out how the ruby version of markdownlint is used, trying to replace with nodejs version

kassvl:
I'd like to pick this up. I did some digging before writing any code, and the good news is that the build-tools image already ships markdownlint-cli2 v0.17.2, so this doesn't need any istio/tools changes.

I ported mdl.rb to an equivalent .markdownlint.yml and ran markdownlint-cli2 against content transformed the same way lint_site.sh does it (shortcode to fence replacement etc). Results: en 156, zh 158, uk 204 findings.

About 40% is MD029 (ordered list prefixes). That one is parser strictness, not real breakage: kramdown treats numbered steps separated by unindented text as one list, CommonMark splits them, and Hugo renders these fine since goldmark keeps the start number. I would disable MD029 in the config with a comment. The rest are real issues the old parser missed: mixed list markers (MD004), 2-space nesting (MD007), bare URLs (MD034), plus a few structural ones. Around 300 fixes across the three languages.

[...]

PR #15360, opened 2024-06-28
update markdownlint headers to match doc

rashmi43:
## Description

Fixes: https://github.com/istio/istio.io/issues/15148

## Reviewers

- [ ] Ambient
- [X ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/mdl.rb
+++ b/mdl.rb
@@ -1,5 +1,5 @@
 all
-rule 'MD002', :level => 2
+rule 'MD002', :level => 1
 rule 'MD007', :indent => 4
 rule 'MD013', :line_length => 160, :code_blocks => false, :tables => false
 rule 'MD026', :punctuation => ".,;:!"


craigbox:
@rashmi43 Thanks for this, but the issue in question was asking to change the engine we use, not the configuration for the current one.

The change youve proposed would  actually break every doc in istio.io, as they would all expect to start with a single `#` heading, where they currently start with a `##` heading.

rashmi43:
thanks for your comments, how can I go about engine change?

craigbox:
You would need to learn how the current script/build tools use the Ruby version of markdownlint, update them to package and use the Node.JS version, and then update the configuration to match our existing rules. 

You may also need to do a mass update of the docs to fix any behaviour we can't configure around.

If you are interested in doing this, then consider dropping by #docs on the Istio slack.

rashmi43:
> You would need to learn how the current script/build tools use the Ruby version of markdownlint, update them to package and use the Node.JS version, and then update the configuration to match our existing rules.
> 
> You may also need to do a mass update of the docs to fix any behaviour we can't configure around.
> 
> If you are interested in doing this, then consider dropping by #docs on the Istio slack.

sure I will take this up

PR #15394, opened 2024-07-08
Fix docs on customTags

taer:
it has a typo and that was the one example I used to copy/pasta

Just fixes the docs on custom-tags.

### DIFF
--- a/content/en/docs/tasks/observability/distributed-tracing/telemetry-api/index.md
+++ b/content/en/docs/tasks/observability/distributed-tracing/telemetry-api/index.md
@@ -153,7 +153,7 @@ You can customize the tags using any of the three supported options below.
         - providers:
             - name: "zipkin"
           randomSamplingPercentage: 100.00
-          custom_tags:
+          customTags:
             my_tag_header:
               header:
                 name: <CLIENT-HEADER>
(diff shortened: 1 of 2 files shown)


kfaseela:
cc @zirain

dhawton (changes requested):
This should be made against master branch. We can cherry-pick to release-1.22 after.

taer:
> This should be made against master branch. We can cherry-pick to release-1.22 after.

OMG.  No idea how I did that. Will fixing soon

taer:
@dhawton There are a few more instances of custom_tags in the docs. mostly in the meshConfig/defaultConfig.

Think those might be deprecated? They are note referenced here: https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#Tracing

that tracing section reached via: https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig

zirain (approved):
LGTM, can you make CI happy.

you may need to run `make gen` locally.

taer:
> LGTM, can you make CI happy.
> 
> you may need to run `make gen` locally.

done

dhawton (approved):

PR #15408, opened 2024-07-12
Update shared drive location

craigbox:
By updating the link from the root of the Shared Drive to a second-level folder I have created for this purpose, people will be able to see the Istio community drive without having had to first join a group.


justinpettit:
LGTM

kfaseela (approved):

PR #15414, opened 2024-07-15
New Google Analytics property IDs

craigbox:
This PR replaces the Universal Analytics accounts that were restricted to only be accessed by users at google.com, and adds new accounts for Google Analytics 4 that are owned by admin@istio.io.

- [x] Docs


craigbox (commented):
/cherry-pick release-1.22

kfaseela (approved):

PR #15443, opened 2024-07-18
prepare for v1.23

dhawton:
## Description

Prepare for 1.23 release

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

PR #15450, opened 2024-07-19
Fixup some waypoint docs issues

bleggett:
## Description

https://github.com/istio/istio.io/pull/15443 should go in first.

This fixes a few issues I found redoing the ambient getting started guide in https://github.com/istio/istio.io/pull/15279.

This:

1. Fixes a bad ref in the `waypoint wasm` doc that points back to a "how to configure waypoints" ref that no longer exists.
2. Actually puts the "install the K8S gateway CRDS" step where it belongs in the waypoint guide
3. Starts to add a few basic tests to the waypoint guide.

## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/usage/waypoint/index.md
+++ b/content/en/docs/ambient/usage/waypoint/index.md
@@ -46,7 +46,7 @@ After the waypoint is deployed, the entire namespace (or whichever services or p
 
 Before you deploy a waypoint proxy for a specific namespace, confirm the namespace is labeled with `istio.io/dataplane-mode: ambient`:
 
-{{< text bash snip_id=check_ns_label >}}
+{{< text syntax=bash snip_id=check_ns_label >}}
 $ kubectl get ns -L istio.io/dataplane-mode
 NAME              STATUS   AGE   DATAPLANE-MODE
 istio-system      Active   24h
@@ -55,7 +55,7 @@ default           Active   24h   ambient
 
 `istioctl` can generate a Kubernetes Gateway resource for a waypoint proxy. For example, to generate a waypoint proxy named `waypoint` for the `default` namespace that can process traffic for services in the namespace:
 
-{{< text bash snip_id=gen_waypoint_resource >}}
+{{< text syntax=bash snip_id=gen_waypoint_resource >}}
 $ istioctl waypoint generate --for service -n default
 kind: Gateway
 metadata:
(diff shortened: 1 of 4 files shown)


content/en/docs/ambient/usage/waypoint/index.md
```diff
@@ -31,13 +31,22 @@ When you configure redirection to a waypoint, traffic will be forwarded by ztunn
...
 
-Waypoint proxies are deployed declaratively using Kubernetes Gateway resources. You can use istioctl experimental subcommands to generate, apply or list these resources.
+Waypoint proxies are deployed using standard Kubernetes Gateway resources. 
+
+You need to install the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/) CRDs, which don’t come installed by default on most Kubernetes clusters:
+
+{{< text syntax=bash snip_id=install_k8s_gateway_api >}}
+$ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
+  { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v1.1.0" | kubectl apply -f -; }
+{{< /text >}}
```
craigbox, line 41: There's a boilerplate for this

bleggett: @craigbox there is (and I see a few other spots we aren't using it), but it includes a step that

- We didn't include here before
- Isn't necessary for _ambient_ specifically
- Has a WARN text


e.g. https://preliminary.istio.io/latest/docs/tasks/traffic-management/tcp-traffic-shifting/

Should we create a new boilerplate, tweak the existing one to remove e.g. `PILOT_ENABLE_ALPHA_GATEWAY_API`, or just stick with this (which was copied from the existing doc) for now?

bleggett: Actually, it seems like we do not even need the `experimental` version of the CRDs anyway, for ambient waypoints.

Lemme test and confirm and if so I will leave this non-boilerplated - we should _not_ need to mention this outside of the actual waypoint doc (what this PR does) so it maybe shouldn't be boilerplated for ambient docs.

bleggett: Aight @craigbox updated this to use the non-experimental version.

PTAL. I think I would prefer if we just link to this doc in Ambient docs when we need to mention waypoints, rather than stick the boilerplate in a bunch of places.

bleggett:
I realized I pushed this to `upstream` (which oddly wasn't blocked, but which is now `protected` meaning I can't push to it - so I am going to reopen this as a proper fork PR)

PR #15460, opened 2024-07-22
Remove docs on removed PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING feature

howardjohn:


dhawton:
Note: This is already in #15443

craigbox:
I only pulled out the link to the page that I removed - there is more in this PR that we need to keep.

dhawton:
> I only pulled out the link to the page that I removed - there is more in this PR that we need to keep.

I updated the rest of the markdown.. the other bits got removed with the make update_ref_docs.

craigbox:
Can we close this then @dhawton?

PR #15518, opened 2024-08-06
Adjust fonts for code snippets

craigbox:
* `<a>` tags inside `<pre>` were not rendering in monotype
* code was always just a tiny bit too small, so bumped `<code>` to the same size as `<pre>`

see https://istio.io/latest/docs/examples/bookinfo/#cleanup for a good before and after


craigbox:
cf. https://deploy-preview-15518--preliminary-istio.netlify.app/latest/docs/examples/bookinfo/#cleanup

looks good with a sniff test

kfaseela (approved):

PR #15532, opened 2024-08-07
bump hugo to match tools

### LINKED ISSUE #15524: Update Hugo to match what is in the tools image
Currently istio.io has Hugo set to 0.130.0 in the `master` branch which matches the current istio/tools `master` branch, although it is being updated to 0.131.0 in https://github.com/istio/tools/pull/2997. 

However, the current istio.io master branch is mirroring the 1.23 release, which is using Hugo 0.128.0: https://github.com/istio/tools/blob/release-1.23/docker/build-tools/Dockerfile#L59

We should update the Hugo in the 1.23 tooling to be the same 0.130.0, move istio.io back to 0.128.0, or we could move both to 0.131.0.

This is because we will create the 1.23 branch of istio.io when 1.23.0 is released and then this branch is using a later Hugo than the tooling. We just want them to be in sync.

dhawton:
## Description

Closes #15524 

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,7 @@
     publish = "public"
 
 [build.environment]
-    HUGO_VERSION = "0.130.0"
+    HUGO_VERSION = "0.131.0"
     NODE_VERSION = "18.18.2"
     BUILD_WITH_CONTAINER = "0"
     GO_VERSION = "1.21.3"


ericvn:
The tools build-image change is in https://github.com/istio/common-files/pull/1077 with the tools update in https://github.com/istio/tools/pull/3002

craigbox (approved):

PR #15536, opened 2024-08-08
Update index.md

tesla59:
## Description

This PR updated release name of the istio gateway installed using helm. While in the gateway [docs](https://istio.io/latest/docs/setup/additional-setup/gateway/) the release name is `istio-ingressgateway`, in the helm installation doc for istio the release name is `istio-ingress`.

Due to this change, anyone following the helm installation doc will have gateway label as `istio: gateway` instead of `istio: ingressgateway`. Other installation method sets the labels as `istio: ingressgateway` and the other docs also assume the label to be `istio: ingressgateway`. 

image
image

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


dhawton:
This has been requested before and been blocked: https://github.com/istio/istio.io/pull/13516#pullrequestreview-1522988286

tesla59:
> users should understand what fields link things up and set them to match when appropriate.

While i agree that the user should be aware of how the selector is working, i think we should mention it in docs then. if installing using istioctl sets `istio: ingressgateway`, other installation method should at least mention it. We are doing it at one place but not doing at other. we should make it clear in the doc itself that the `istio: ingressgateway` is just a label selector and it can even be `foo: bar`

howardjohn:
> we should make it clear in the doc itself that the `istio: ingressgateway` is just a label selector and it can even be `foo: bar`

:+1:

dhawton:
/ok-to-test

You're going to need to run `make gen` as well to update the tests

tesla59:
@dhawton i have run `make gen` and pushed the commit. LMK if anything else is needed

Issue #15548, opened 2024-08-12
update hugo to v0.132.0


Arhell:

dhawton:
Delay until after branch cut or bump build-tools first: https://github.com/istio/tools/blob/release-1.23/docker/build-tools/Dockerfile#L59

PR #15553, opened 2024-08-13
update hugo to v0.133.0

Arhell:
resolve https://github.com/istio/istio.io/issues/15548

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,7 @@
     publish = "public"
 
 [build.environment]
-    HUGO_VERSION = "0.132.1"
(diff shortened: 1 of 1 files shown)


dhawton (changes requested):
https://github.com/istio/tools/blob/release-1.23/docker/build-tools/Dockerfile#L59

We should keep this in sync with the build-tools.

craigbox (approved):

dhawton:
https://github.com/istio/tools/pull/3013 should merge first so we don't get too far ahead of the tooling

dhawton (approved):

Issue #15458, opened 2024-07-22
set noindex on all content outside of https://istio.io/


craigbox:
Google is confused about our duplicate content.  We think it's all in the search index in order to power the [custom search engine](https://developers.google.com/custom-search), but it looks like at least some that content is still available even though I've delisted it from the main search index.

That, and the facets don't actually correctly list versioned content. 

My suggestion at this point is that we break site search for old versions, and then we add everything back, with a new engine if needed.

So, the first thing to do is to **add `noindex` to the meta tags of every page on preliminary and archive**.

https://developers.google.com/search/docs/crawling-indexing/block-indexing

Please do this in order:

(a) preliminary.istio.io
(b) await validation
(c) archive

milinddethe15:
Hi @craigbox , we can include the noindex meta tag globally by updating the `/layouts/_default/baseof.html` file. This ensures that the tag is applied across all pages that extend from this base template.
the logic that can be added to the <head> section of baseof.html:
```
    {{ if .Site.Data.args.preliminary }}
        <meta name="robots" content="noindex">
    {{ else if .Params.noindex }}
        <meta name="robots" content="noindex">
    {{ end }}
```
Let me know if this works?

craigbox:
You would have to tell me if it works.

https://istio.io/v1.20/ should get `noindex`, but it's not preliminary.  How does this work in the context of our current archiving, and how will it work in a future case?

milinddethe15:
> https://istio.io/v1.20/ should get `noindex`, but it's not preliminary. How does this work in the context of our current archiving, and how will it work in a future case?

For previous versions, it can be re-builded with noindex tags (we have source code in respective version branches) or add manually, lots of work.
I have created PR for adding noindex tag for preliminary. Later we can add tags in older versions. once the tags works for preliminary.

PR #15592, opened 2024-08-21
Add conditional noindex meta tag for preliminary and archive

milinddethe15:
## Description
#15458

Added noindex meta tag,  for preliminary and upcoming archive, by conditional logic in < head > tag of `/layouts/_default/baseof.html` 

For preliminary.istio.io :: set `.Site.Data.args.preliminary` = true
For archive.istio.io / in future, when creating snapshot of current version :: set `.Site.Data.args.archive` = true

Value of args are set here:
https://github.com/istio/istio.io/blob/1a036575a64fbf4eabe6f94eae49639fada3575d/data/args.yml#L19-L20
## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -124,6 +124,13 @@
 
         <!-- set the color theme -->
         <script src="/js/themes_init.min.js"></script>
+
+        <!-- noindex meta tag -->
+        {{ if .Site.Data.args.preliminary }}
+        <meta name="robots" content="noindex">
+        {{ else if .Site.Data.args.archive }}
+        <meta name="robots" content="noindex">
+        {{ end }}
     </head>
 
     {{ $site_mode := ""}}


Arhell (commented):
/ok-to-test

milinddethe15:
@craigbox  PTAL :)

layouts/_default/baseof.html
```diff
@@ -124,6 +124,13 @@
 
         <!-- set the color theme -->
         <script src="/js/themes_init.min.js"></script>
+
+        <!-- noindex meta tag -->
+        {{ if .Site.Data.args.preliminary }}
+        <meta name="robots" content="noindex">
+        {{ else if .Site.Data.args.archive }}
+        <meta name="robots" content="noindex">
+        {{ end }}
```
craigbox, line 133: This is the same content, you don't need to include it twice.  Use `{{ if or }}`

PR #15622, opened 2024-08-27
update hugo to v0.133.1

Arhell:
resolve https://github.com/istio/istio.io/issues/15613


craigbox:
I don't think we need to track every minor Hugo version?  If that were the case we could automate this,

Issue #15624, opened 2024-08-28
In-doc header links broken in "tip" block


jaellio:
I found several instances where links in "tip" formatted blocks do not correctly redirect. This is specific to links to in doc headers. For example, the link to `optimizing metrics collection with recording rules` in the tip block on https://istio.io/latest/docs/ops/best-practices/observability/ is broken.

```
{{< tip >}}
The recording rules above only aggregate across pods and instances. They still preserve the full set of
[Istio Standard Metrics](/docs/reference/config/metrics/), including all Istio dimensions. While this
will help with controlling metrics cardinality via federation, you may want to further optimize the recording rules
to match your existing dashboards, alerts, and ad-hoc queries.

For more information on tailoring your recording rules, see the section on
[Optimizing metrics collection with recording rules](#optimizing-metrics-collection-with-recording-rules).
{{< /tip >}}
```

The link redirects to https://istio.io/latest/#optimizing-metrics-collection-with-recording-rules rather than https://istio.io/latest/docs/ops/best-practices/observability/#optimizing-metrics-collection-with-recording-rules. 

I am not positive this is related to the links being defined in a "tip" formatting block, but I have found several examples including `compatibility with CNIs` in https://istio.io/latest/docs/setup/additional-setup/cni/.

craigbox:
https://discourse.gohugo.io/t/link-to-page-anchor/5013

PR #15635, opened 2024-08-30
in-doc header links broken under <tip>

kfaseela:
## Description

Fixes https://github.com/istio/istio.io/issues/15624

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/best-practices/observability/index.md
+++ b/content/en/docs/ops/best-practices/observability/index.md
@@ -159,7 +159,7 @@ will help with controlling metrics cardinality via federation, you may want to f
 to match your existing dashboards, alerts, and ad-hoc queries.
 
 For more information on tailoring your recording rules, see the section on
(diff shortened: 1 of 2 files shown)


kfaseela:
@craigbox @dhawton aware of any other better ways to fix this?

craigbox:
https://discourse.gohugo.io/t/link-to-page-anchor/5013

because we pass in `baseURL` when we build/run the site, it seems it's going to be added in front of everything inside another template file.  We won't be able to use relative links inside `{{< tip >}}` blocks.

kfaseela:
> https://discourse.gohugo.io/t/link-to-page-anchor/5013
> 
> because we pass in `baseURL` when we build/run the site, it seems it's going to be added in front of everything inside another template file.  We won't be able to use relative links inside `{{< tip >}}` blocks.

So we merge this PR then?

craigbox (approved):

PR #15656, opened 2024-09-05
update hugo to v0.134.1

Arhell:
resolve https://github.com/istio/istio.io/issues/15652


dhawton:
The changes in this patch release don't affect Istio. We don't necessarily have to bump every release.

Issue #15648, opened 2024-09-04
Documentation unreadable in Chrome


cefn:
### Is this the right place to submit this?

- [X] This is not a security vulnerability or a crashing bug
- [X] This is not a question about how to use Istio

### Bug Description

The layout of tabular data within e.g. https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig prevents reading it. The fixed width is narrower than the contained table even on a very wide screen. Zooming in and out delivers no benefit since it consistently zooms all scaling, including both the fixed width and the font size. 

![image](https://github.com/user-attachments/assets/22f87172-8e7b-45fc-bb29-09969013e885)

A workaround on my Chrome was to open the inspector and edit the style for `.docs.container` to be `2000px`

![image](https://github.com/user-attachments/assets/69d15de1-d1fb-4441-9688-8efb0b3e49b7)

### Version

```Text
N/A
```

kfaseela:
cc @Arhell

Arhell:
@cefn 
I didn't find any problems, there is a table with horizontal scrolling, if you remove it it will be even worse.

The only thing I found is a very long title
https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ExtensionProvider-EnvoyExternalAuthorizationRequestBody

I use chrome to

cefn:
Can you send me a screenshot of the MeshConfig table, then? Here is my screenshot 

<img width="1712" alt="image" src="https://github.com/user-attachments/assets/da86c7c5-ed23-4823-ba4a-f958c0cb06a7">

Arhell:
> Can you send me a screenshot of the MeshConfig and ProxyConfig tables, then? Here are screenshots.
> 
> The fact this table is too wide for the viewport makes this reference documentation unusable in my opinion (for my browser config). Take the ProxyConfig table. There is a whole invisible 'Requirements' column and no control to get to it.
> 
[...]

PR #15666, opened 2024-09-07
fix title

Arhell:
https://github.com/istio/istio.io/issues/15648#issuecomment-2334247557


src/sass/base/_base.scss
```diff
@@ -554,3 +554,9 @@ sup {
 iframe {
     max-width: 100%;
 }
+
+.article-container {
+    h2 {
+        word-wrap: break-word;
+    }
+}
```
craigbox, line 562:
```suggestion
}

```

Arhell: Done, thanks

dhawton (approved):

PR #15727, opened 2024-09-24
1.24 performance updates

Stevenjin8:
## Description

Update performance and scalability docs for 1.24

xref: https://github.com/istio/istio/issues/52977 

## Reviewers

- [x] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

Need someone to update the chinese version


content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -90,45 +92,45 @@ request rate doesn't affect the memory consumption.
 
 ### Latency
 
-Since Istio injects a sidecar proxy on the data path, latency is an important
+Since Istio injects a sidecar proxy or ztunnel proxy on the data path, latency is an important
```
linsun, line 95: we don't inject ztunnel proxy... consider reword it to the request traverse through sidecar or ztunnel or/and waypoint proxy....

content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -90,45 +92,45 @@ request rate doesn't affect the memory consumption.
...
 consideration.
 Every feature Istio adds also adds to the path length inside the proxy and potentially affects latency.
 
 The Envoy proxy collects raw telemetry data after a response is sent to the
```
linsun, line 99: should we add ztunnel here as well?

linsun:
Hi @Stevenjin8 thanks for updating the page!  I have 1 question regarding the sidecar numbers, I saw it is more than ambient + L7 - just wanted to confirm it is the default profile without HBONE being used?

Note the numbers from 1.21 on sidecar is much better: https://istio.io/v1.21/docs/ops/deployment/performance-and-scalability/

Stevenjin8:
> Hi @Stevenjin8 thanks for updating the page! I have 1 question regarding the sidecar numbers, I saw it is more than ambient + L7 - just wanted to confirm it is the default profile without HBONE being used?
> 
> Note the numbers from 1.21 on sidecar is much better: https://istio.io/v1.21/docs/ops/deployment/performance-and-scalability/

@linsun I believe it is the default profile without HBONE. when I look at the traffic, none of it goes to port 15008.

content/en/docs/ops/deployment/performance-and-scalability/index.md
Stevenjin8: I'm not sure what you mean

linsun: line 95 starts with sidecar or ztunnel, and line 99 only covers Envoy proxy. Should it also cover ztunnel as well in the explanation.

linsun:
> > Hi @Stevenjin8 thanks for updating the page! I have 1 question regarding the sidecar numbers, I saw it is more than ambient + L7 - just wanted to confirm it is the default profile without HBONE being used?
> > Note the numbers from 1.21 on sidecar is much better: https://istio.io/v1.21/docs/ops/deployment/performance-and-scalability/
> 
> @linsun I believe it is the default profile without HBONE. when I look at the traffic, none of it goes to port 15008.

hmm, do you think we regressed sidecar numbers so bad?  Or the difference lies in setup and measurement?

content/en/docs/ops/deployment/performance-and-scalability/index.md
Stevenjin8: I see

Stevenjin8:
> > > Hi @Stevenjin8 thanks for updating the page! I have 1 question regarding the sidecar numbers, I saw it is more than ambient + L7 - just wanted to confirm it is the default profile without HBONE being used?
> > > Note the numbers from 1.21 on sidecar is much better: https://istio.io/v1.21/docs/ops/deployment/performance-and-scalability/
> > 
> > 
> > @linsun I believe it is the default profile without HBONE. when I look at the traffic, none of it goes to port 15008.
> 
> hmm, do you think we regressed sidecar numbers so bad? Or the difference lies in setup and measurement?



> > > Hi @Stevenjin8 thanks for updating the page! I have 1 question regarding the sidecar numbers, I saw it is more than ambient + L7 - just wanted to confirm it is the default profile without HBONE being used?
> > > Note the numbers from 1.21 on sidecar is much better: https://istio.io/v1.21/docs/ops/deployment/performance-and-scalability/
> > 
> > 
> > @linsun I believe it is the default profile without HBONE. when I look at the traffic, none of it goes to port 15008.
> 
> hmm, do you think we regressed sidecar numbers so bad? Or the difference lies in setup and measurement?

I'll do some digging, but its hard to tell because there isn't a lot of information about the test setup. Then again, we're not doing anything crazy so it shouldn't matter too much?

Stevenjin8:
@linsun Is it even possible for sidecar to sidecar traffic to be over HBONE? I couldn't find any docs on it. All HBONE documentation seems to be under Ambient mode.

Also, I do think the sidecar numbers are right. John ran similar tests (albeit locally on a kind cluster) and got similar numbers

```
DEST   CLIENT  QPS   CONS  DUR  PAYLOAD  SUCCESS  THROUGHPUT  P50      P90      P99
hyper  fortio  1000  1     5    0        5000     999.75qps   0.391ms  0.523ms  0.818ms
hyper  fortio  1000  4     5    0        5000     999.30qps   0.412ms  0.623ms  1.552ms
hyper  fortio  1000  16    5    0        4992     996.77qps   0.445ms  0.651ms  1.535ms
hyper  fortio  1000  64    5    0        4992     987.42qps   0.494ms  0.818ms  1.895ms
hyper  fortio  1000  128   5    0        4992     975.06qps   0.480ms  0.841ms  1.904ms
hyper  fortio  1000  265   5    0        4770     947.34qps   0.459ms  0.768ms  1.909m
```

Stevenjin8:
@linsun I reran the tests with 1.21 and go the same results

Stevenjin8:
going to mark this as ready for review so that the pipelines run

content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -78,55 +78,56 @@ The latency, throughput, and the proxies' CPU and memory consumption are measure
 ### Sidecar CPU and memory usage
 
 Since the sidecar proxy performs additional work on the data path, it consumes CPU
-and memory. In Istio 1.22, a sidecar proxy consumes about 0.5 vCPU per 1000
-requests per second.
+and memory. In Istio 1.24, with 1000 http requests per second containing 1 KB of payload each,j
```
zirain, line 81: nit?

Stevenjin8: Oops good catch

Stevenjin8:
@linsun The difference is likely that the 1.21 tests ran without http proxying (only tcp), while the current ones run with http-proxying.

content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -78,55 +78,56 @@ The latency, throughput, and the proxies' CPU and memory consumption are measure
 ### Sidecar CPU and memory usage
 
 Since the sidecar proxy performs additional work on the data path, it consumes CPU
-and memory. In Istio 1.22, a sidecar proxy consumes about 0.5 vCPU per 1000
-requests per second.
+and memory. In Istio 1.24, with 1000 http requests per second containing 1 KB of payload each
+- a sidecar proxy with 4 worker threads consumes about 27% CPU and 61 MB of memory.
```
howardjohn, line 82: Can we avoid using % and instead use 0.25 vCPU (note I also rounded -- its not like our measurements are so precise we can really commit to that level of granularity).

We can also likely round memory to 50mb.

Also, why 4 worker threads? The default (and 99.999% of users) run 2 threads

content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -78,55 +78,56 @@ The latency, throughput, and the proxies' CPU and memory consumption are measure
...
+- a waypoint proxy with 2 worker threads consumes about 41% CPU and 57 MB of memory
+- a ztunnel proxy consumes about 7% CPU and 12 MB of memory.
```
howardjohn, line 84: Nit: this is odd on "Sidecar CPU and memory usage" section.

howardjohn: Also this implies ambient uses ~2x CPU (7+41=48% vs 27%). Is the sidecar only looking at 1 sidecar? If so we should make that more clear

content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -78,55 +78,56 @@ The latency, throughput, and the proxies' CPU and memory consumption are measure
...
 
 The memory consumption of the proxy depends on the total configuration state the proxy holds.
 A large number of listeners, clusters, and routes can increase memory usage.
-In a large namespace with [namespace isolation](/docs/reference/config/networking/sidecar/) enabled, the proxy consumes approximately 50 MB of memory.
 
 Since the proxy normally doesn't buffer the data passing through,
```
howardjohn, line 89: Not for this PR but this is not an accurate statement. It allocates buffers for each connection

content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -78,55 +78,56 @@ The latency, throughput, and the proxies' CPU and memory consumption are measure
...
 Every feature Istio adds also adds to the path length inside the proxy and potentially affects latency.
 
 The Envoy proxy collects raw telemetry data after a response is sent to the
-client. The time spent collecting raw telemetry for a request does not contribute
-to the total time taken to complete that request. However, since the worker
-is busy handling the request, the worker won't start handling the next request
-immediately. This process adds to the queue wait time of the next request and affects
-average and tail latencies. The actual tail latency depends on the traffic pattern.
+client.
+The time spent collecting raw telemetry for a request does not contribute to the total time taken to complete that request.
```
howardjohn, line 100: Arguably this isn't true for ztunnel

howardjohn: But that is pretty pedantic, I don't mind if it stays as-is

howardjohn (approved):
Left a few comments but this is clearly an improvement. LGTM!

content/en/docs/ops/deployment/performance-and-scalability/index.md
howardjohn: Is ztunnel adding up both ztunnels or just 1?

howardjohn: For ztunnel I got 40m CPU and 2Mb memory each so hopefully that is adding them

Stevenjin8: These are all per sidecar/ztunnel. Will clarify.

content/en/docs/ops/deployment/performance-and-scalability/index.md
Stevenjin8: I think this was a formatting change from my part, so I will keep as is.

content/en/docs/ops/deployment/performance-and-scalability/index.md
Stevenjin8: I'll update

linsun:
> @linsun The difference is likely that the 1.21 tests ran without http proxying (only tcp), while the current ones run with http-proxying.
> 
> So I think the tests as they are are correct. The question is whether we want include http proxying in tests or revert back to the 1.21 testing style. I think it makes sense to have http proxying if we are doing http.

Yep agreed, thank you so much for looking into the differences and analyzing why!

content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -75,58 +75,56 @@ Data plane performance depends on many factors, for example:
 
 The latency, throughput, and the proxies' CPU and memory consumption are measured as a function of said factors.
 
-### Sidecar CPU and memory usage
+### Sidecar and Ztunnel resource usage
```
craigbox, line 78:
```suggestion
### Sidecar and ztunnel resource usage
```

craigbox (approved):

PR #15733, opened 2024-09-26
Clarify subjectAltNames requires tls mode MUTUAL

zioproto:
## Description

The user might be confused about the usage of `subjectAltNames`, trying to use it to add more hostnames to an endpoint.
Make clear that it requires TLS mode mutual, and the the alternative names are those presented by the client for authentication.

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [X] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/reference/config/networking/gateway/index.html
+++ b/content/en/docs/reference/config/networking/gateway/index.html
@@ -546,7 +546,7 @@ No
 <td><code>string[]</code></td>
 <td>
 <p>A list of alternate names to verify the subject identity in the
-certificate presented by the client.</p>
+certificate presented by the client. Requires TLS mode to be set to <code>MUTUAL</code>.</p>
 
 </td>
 <td>


content/en/docs/reference/config/networking/gateway/index.html
```diff
@@ -546,7 +546,7 @@ <h2 id="ServerTLSSettings">ServerTLSSettings</h2>
 <td><code>string[]</code></td>
 <td>
 <p>A list of alternate names to verify the subject identity in the
-certificate presented by the client.</p>
+certificate presented by the client. Requires tls mode MUTUAL.</p>
```
nshankar13, line 549: Nit - "requires TLS mode to be set to MUTUAL" ?

dhawton (changes requested):
```
```

This needs to be changed in the api repo.

zioproto:
@dhawton can you help me to understand where to patch exactly ?

I found here:
https://github.com/istio/api/blob/eb7df4d4b474d2aa19f89fcece31c60893a7eb67/networking/v1alpha3/gateway.pb.html#L542-L547

but that also seems to be generated.

Do I have to patch the `.go` file here in the `v1alpha3` folder ?
https://github.com/istio/api/blob/eb7df4d4b474d2aa19f89fcece31c60893a7eb67/networking/v1alpha3/gateway.pb.go#L789-L790

Also the `v1beta1` and `v1` API folders seem not to have the full `gateway.pb.go` file. I am not familiar with this repo. Could you please advice ? thanks

dhawton:
It comes from the proto files, that's where you'll need to edit it. The `.pb.go` files are generated from the protos, and the docs are also generated from the protos.

zioproto:
Closing in favour of https://github.com/istio/api/pull/3319

PR #15740, opened 2024-09-27
first pass for ambient api page

linsun:
Part of https://github.com/istio/istio/issues/52698

## Reviewers

- [x] Ambient
- [ x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


linsun:
Anything else should be added to this page @howardjohn @ilrudie @keithmattix @louiscryan ?

content/en/docs/ambient/overview/APIs/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+TODO: add link to annotations page for each when 3307 is merged.
+
+## Kubernetes Gateway API resources
+
+Below are the resources you can use to configure your Ingress gateway or waypoint proxies:
+
+|  Name  | Status |
+| --- | --- |
+| [`HTTPRoute`](https://gateway-api.sigs.k8s.io/guides/http-routing/) | Beta |
```
keithmattix, line 40: HTTPRoute is GA and so is Istio's support for it generally. If this is documentation for Ambient GA, shouldn't this status be GA as well?

content/en/docs/ambient/overview/APIs/index.md
```diff
@@ -0,0 +1,56 @@
...
+TODO: add link to labels page for each when 3307 is merged.
+
+## Annotations
+
+Below are annotations that can be used to customize your waypoint proxies:
+
+| Name | Status |
+| ------------------------------- | -------------------------- |
+| `gateway.istio.io/service-account` | Alpha |
+| `ambient.istio.io/waypoint-inbound-binding` | Alpha |
```
keithmattix, line 29: I actually have no idea what this does

content/en/docs/ambient/overview/APIs/index.md
```diff
@@ -0,0 +1,56 @@
...
+test: no
+---
+
+This page lists all the user facing APIs relevant to the ambient mode and its feature status.
+
+## Labels
+
+| Name | Status |
+| ------------------------------- | -------------------------- |
+| `istio.io/dataplane-mode` | Beta |
```
keithmattix, line 15: Similar to comment below: if this is GA documentation, these labels should be GA

content/en/docs/ambient/overview/APIs/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+TODO: add link to labels page for each when 3307 is merged.
+
+## Annotations
+
+Below are annotations that can be used to customize your waypoint proxies:
+
+| Name | Status |
+| ------------------------------- | -------------------------- |
+| `gateway.istio.io/service-account` | Alpha |
```
keithmattix, line 28: Why are these annotations and not labels? IIRC, @bleggett was pushing for a convention of annotations == controller-set and labels == user-intent and I agree.

content/en/docs/ambient/overview/APIs/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+|  Name  | Status |
+| --- | --- |
+| [`HTTPRoute`](https://gateway-api.sigs.k8s.io/guides/http-routing/) | Beta |
+| [`TLSRoute`](https://gateway-api.sigs.k8s.io/guides/tls) | Alpha |
+| [`TCPRoute`](https://gateway-api.sigs.k8s.io/guides/tcp/) | Alpha |
+
+## Istio resources
+
+Below are the Istio resources you can use to build networking or security or other policies for your services in the ambient mode:
```
keithmattix, line 46: Maybe add a sentence here about how the project recommends using Gateway API as the default and only using these resources when you need to. And maybe sort the resources in order of safety just for subliminal messaging (e.g. AuthZ policy first, then ServiceEntry, then DestinationRule, etc.).

content/en/docs/ambient/overview/APIs/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+|  Name  | Status |
+| --- | --- |
+| [`VirtualService`](https://gateway-api.sigs.k8s.io/guides/http-routing/) | Alpha |
+| [`DestinationRule`](https://gateway-api.sigs.k8s.io/guides/tls) | Beta |
```
keithmattix, line 51: These 3 (DR, SE, and AP) should be GA

content/en/docs/ambient/overview/APIs/index.md
howardjohn: I would drop this and annotations, its already documented on a dedicated page (https://github.com/istio/api/pull/3307 is important though)

content/en/docs/ambient/overview/APIs/index.md
bleggett: PR: https://github.com/istio/istio/pull/43103/files predates that.

I'll sign up to move them to labels if there's agreement we should.

content/en/docs/ambient/overview/APIs/index.md
bleggett: It's undocumented/WIP/experimental AFAIK: https://github.com/istio/istio/pull/50235

I'd be in favor of simply not mentioning it in this doc at all at this time.

content/en/docs/ambient/overview/APIs/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Annotations
+
+Below are annotations that can be used to customize your waypoint proxies:
+
+| Name | Status |
+| ------------------------------- | -------------------------- |
+| `gateway.istio.io/service-account` | Alpha |
+| `ambient.istio.io/waypoint-inbound-binding` | Alpha |
+| `gateway.istio.io/name-override` | Alpha |
```
bleggett, line 30: This also comes from: https://github.com/istio/istio/pull/43103/files and also should be a label.

content/en/docs/ambient/overview/APIs/index.md
howardjohn: I am not super convinced; there are many many label and annotations that do not align with this, so we would need to change 100s

bleggett: I don't think there are 100s in ambient specifically.

That being said - this _should_ probably be something we explicitly address before moving an annotation/label from `Alpha`

howardjohn: `gateway.istio.io/service-account` isn't really ambient, it existed before ambient and impacts GW in general. A large portion of annotations probably apply to sidecar+ambient

content/en/docs/ambient/overview/APIs/index.md
keithmattix: Oh yeah I'd leave this completely undocumented; I don't even know if I see sandwich on istio.io anywhere

content/en/docs/ambient/overview/APIs/index.md
keithmattix: Can we duplicate the ambient annotations into labels and gradually deprecate the annotations?

bleggett (commented):
Is this designed to document 

- labels/annos that _are_ stable/GA (and thus likely already documented elsewhere)?
- labels/annos that are _not_ stable/GA (and thus likely not already documented elsewhere)?
- both?

Perhaps we simply should _not_ document unstable annos/labels. If they're important enough to surface to end-users, that's maybe a good signal they should be stabilized and documented.

(if we just want an inventory of all the unstable candidate annos/labels for our own purpsoes, that maybe can just be in a dev wiki and not user-facing docs).

Otherwise, mentioning them here probably just helps codify them as "supported" when they may in fact be entirely unused, and we may not want that? In general I feel like documenting `alpha` anything in user-facing docs is a bad habit we should always avoid.

content/en/docs/ambient/overview/APIs/index.md
keithmattix: Or, alternatively, use a completely different label for ambient (though this would result in more confusion IMO)

bleggett: > `gateway.istio.io/service-account` isn't really ambient, it existed before ambient and impacts GW in general. A large portion of annotations probably apply to sidecar+ambient

Fair enough - but if it's Alpha, its ambient-ness kinda doesn't matter. It shouldn't graduate to beta or beyond and be documented until we make it what it should be.

linsun:
> Is this designed to document
> 
>     * labels/annos that _are_ stable/GA (and thus likely already documented elsewhere)?
> 
>     * labels/annos that are _not_ stable/GA (and thus likely not already documented elsewhere)?
> 
>     * both?
> 
> 
> Perhaps we simply should _not_ document unstable annos/labels. If they're important enough to surface to end-users, that's maybe a good signal they should be stabilized and documented.
> 
> (if we just want an inventory of all the unstable candidate annos/labels for our own purposes, that maybe can just be in a dev wiki and not user-facing docs).
> 
> Otherwise, mentioning them here probably just helps codify them as "supported" when they may in fact be entirely unused/associated with WIP features, and we may not want that? In general I feel like documenting `alpha` anything in user-facing docs is a bad habit we should always avoid.

The intention is to have 1 single page on ambient related APIs so user can easily understand what APIs are for ambient and their feature status. Yes, we probably have the labels/annotations documented elsewhere but I felt would be good to have everything in one page.  Some of them could be alpha as that is the current state so users are well informed when they choose to use an alpha API.

content/en/docs/ambient/overview/APIs/index.md
linsun: My intent is to have everything in one page while linking to the official API reference page, so users don't need to go through 100 labels/annotations to identify which ones (in fact only a few of them) are ambient related.

bleggett:
> Some of them could be alpha as that is the current state so users are well informed when they choose to use an alpha API.

I don't think any alpha annotations or labels are required for ambient GA usage.

I think, based on the [actual feature definitions we always have had](https://istio.io/latest/docs/releases/feature-stages/), we simply should never document any labels or annotations that are not Beta or higher.

If this causes a problem, the affected non-beta labels or annotations should be promoted to beta or better, and documented at that time.

keithmattix:
> I don't think any alpha annotations or labels are required for ambient GA usage, correct?

Put another way, if we required alpha labels/annotations for ambient GA usage, then Ambient's not at GA quality.

bleggett:
Yeah and since 

- https://github.com/istio/api/blob/master/label/labels.yaml
- https://github.com/istio/api/blob/master/annotation/annotations.yaml

already exist as "unstable" documentation for Alpha-or-below labels, I don't see the need to mention Alpha-or-below labels in the user facing docs at all. If we want to refer to them in user-facing docs they should be Beta or higher.

content/en/docs/ambient/overview/APIs/index.md
bleggett: > Or are they already stable?

already stable. so ship has sailed on these. In general though I _would_ strongly suggest we don't promote labels or annos out of Alpha until we make sure they're in the right bucket, going forward.

That seems like a pretty minimally reasonable quality gate and just because we haven't been consistent in the past doesn't mean we should avoid starting now.

linsun:
Thanks to everyone for the great feedback, I decided to leave the annotations out as they are alpha and many are not user facing anyway. Still waiting for label reference to show up in our doc to refer to add links to it, otherwise should be good to go.

content/en/docs/ambient/overview/APIs/index.md
linsun: Updated!

content/en/docs/ambient/overview/APIs/index.md
linsun: Sounds good, I've removed annotations out.

content/en/docs/ambient/overview/APIs/index.md
linsun: Decided to just keep the few important labels. As far as status for the label, I'll add a discussion item for today's meeting. I do feel they are GA also but John had beta in https://github.com/istio/api/pull/3307 so I started with beta.

content/en/docs/ambient/overview/APIs/index.md
linsun: Yep good thought!

linsun:
> Thanks to everyone for the great feedback, I decided to leave the annotations out as they are alpha and many are not user facing anyway. Still waiting for label reference to show up in our doc to refer to add links to it, otherwise should be good to go.

I don't see specific label reference [shown up yet](https://preliminary.istio.io/latest/docs/reference/config/labels/), so going to get this one in first and update it later on when the specific labels are available.

bleggett (approved):

dhawton:
For `GA`, we don't seem to define that anywhere but we do define `Stable` (https://istio.io/latest/docs/releases/feature-stages/). Should we use Stable over "GA" since it's documented?

linsun:
> For `GA`, we don't seem to define that anywhere but we do define `Stable` (https://istio.io/latest/docs/releases/feature-stages/). Should we use Stable over "GA" since it's documented?

Good point, though my personal preference is GA as that is what everyone cares about :)

linsun:
@craigbox can you review/approve?  Thanks!

craigbox:
* putting the file where you have it [makes the overview look odd](https://deploy-preview-15740--preliminary-istio.netlify.app/latest/docs/ambient/overview/). 
* I think we should use "stable" because features are GA, labels are not. 
* What people really want to see is a list of what features are GA.  Could that be on this page?
* probably want to lowercase `apis`
* I can't believe that @keithmattix  is letting `EnvoyFilter` onto this page
* likewise, we need a better story about whether `VirtualService` is supported forever, until a migration is complete, not if also used with `HTTPRoute`, etc

I would be inclined for this content to live on https://istio.io/latest/docs/releases/feature-stages/ if it's to be encyclopaedic

linsun:
We are preparing ambient GA blog, i think this content may be best to host on the ambient GA blog. Then we can decide if we should use stable or GA and where to host it. any objections?

linsun:
Hi @dhawton okay to merge this as a first pass? I need to mention it in the GA blog.

linsun:
> > putting the file where you have it [makes the overview look odd](https://deploy-preview-15740--preliminary-istio.netlify.app/latest/docs/ambient/overview/).
> 
> good catch, let me look info it :)

This should be resolved now :)

craigbox:
Second pass, see https://github.com/istio/istio.io/pull/15879.  I suggest we take that one instead.

linsun:
closing this as Craig is reworking on the new PR.

PR #15743, opened 2024-09-27
Make Cilium net pol safe

howardjohn:
Before: If a workload has no other policies, this makes them ONLY allow the
  snat IP

After: This only allows more traffic (the snat source) without denying
anything additional.


bleggett:
Yep, think I got this in https://github.com/istio/istio.io/pull/15742

LMK if not.

howardjohn:
too fast

PR #15744, opened 2024-09-27
Fixup Helm CRD docs in light of istio/istio#53318 

bleggett:
## Description

1. Add a new boilerplate for 1.23 CRD manual fix (should be dropped when 1.23 goes out of support).
2. Fix docs to indicate the correct approach. 

Additionally, pulls in stuff from: https://github.com/istio/istio.io/pull/15756

1. Automator is broken and hasn't updated the gateway/istio refs in a few weeks, this has a manual bump to get some CI passing.
2. Tests were failing due to the old `httpbin` breaking.
3. Tests were still failing with the new `httpbin`, because of differing implementation behaviors related to ambiguous language in https://datatracker.ietf.org/doc/html/rfc2324#section-2.3.2
4. `gateway-api` regen script uses a commit SHA for the version - not the _published_ version. This is fine for dev but it's a no-go for docs, and it breaks the docs that use/expect published refs (which should be all of them, going forward).

## Reviewers

- [x] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/boilerplates/crd-upgrade-123.md
```diff
@@ -0,0 +1,16 @@
...
+{{< warning >}}
+If upgrading CRDs via Helm from an Istio release 1.23 or older, you may encounter an error such as the following
+
+`Error: rendered manifests contain a resource that already exists. Unable to continue with update: CustomResourceDefinition "wasmplugins.extensions.istio.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata`
+
+You can resolve this with a one-time migration using the following `kubectl` commands:
+
+    {{< text syntax=bash snip_id=install_experimental_crds >}}
+    $ kubectl label crds -l chart=istio "app.kubernetes.io/managed-by=Helm"
+    $ kubectl annotate crds -l chart=istio "meta.helm.sh/release-name=istio-base"
```
howardjohn, line 12: do we need a '`# replace with actual release name` comment here

bleggett: Yeah I dithered - we don't have such a comment for the actual chart install/upgrade tho - all these docs assume you are following our release naming instructions.

If you're confident enough to deviate from our docs as-written, you maybe should understand you need to do this etc etc.

I'll add it tho, I don't feel strongly.

bleggett:
I think these doctests are failing (and will fail) until the 1.24 SHA gets bumped:
```
make[1]: Leaving directory '/home/prow/go/src/istio.io/istio'
Installing go-junit-report...
/usr/bin/go-junit-report
ISTIO_LONG_SHA=4e0fb1d3b617aea3967af841c759ad0f0da0f861
TAG=1.24-alpha.4e0fb1d3b617aea3967af841c759ad0f0da0f861
BASE_VERSION=master-2024-09-04T19-00-52
```

is from ~2 weeks ago, and won't work here.

bleggett:
Aside from this, docstests seem generally broken, looking at that over in https://github.com/istio/istio.io/pull/15756

dhawton (approved):

bleggett:

this is just flaky

bleggett:
/test doc.test.profile-demo

the flakes, they are ever-present

kfaseela:
Thanks @bleggett :)

PR #15756, opened 2024-10-03
[DNM] [test] Fix doctests against latest Istio 1.24 pre

bleggett:
1. Automator is broken and hasn't updated the gateway/istio refs in a few weeks, this is a manual bump to get some CI passing.
2. Tests were failing due to the old `httpbin` breaking.
3. Tests were still failing with the new `httpbin`, because of differing implementation behaviors related to ambiguous language in https://datatracker.ietf.org/doc/html/rfc2324#section-2.3.2
4. `gateway-api` regen script uses a commit SHA for the version - not the _published_ version. This is fine for dev but it's a no-go for docs, and it breaks the docs that use/expect published refs (which should be all of them, going forward).


bleggett:
Everything here is in https://github.com/istio/istio.io/pull/15744 so closing

PR #15729, opened 2024-09-24
Add blog post: "The Istio Service Mesh for People Who Have Stuff To Do"

lucavallin:
## Description

I have been invited to re-post [my article](https://www.lucavall.in/blog/the-istio-service-mesh-for-people-who-have-stuff-to-do) here. Let me know if you need more details!

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

PR #15770, opened 2024-10-09
Add link to Luca Cavallin's post

craigbox:
I propose adding a type of blog post to istio.io which is a link to a post made elsewhere.

We can then link to other good content about Istio without rehosting it.

This could be achieved by only having metadata in the front matter, which marks it as link: true.  Then, the RSS can have a link to the actual post, and not to the blog page on istio.io.

Until we decide if we want to do that, we can do things like this:

- title starts with "Link: " to make it obvious
- pages are link-page-something.md 

This is an example of such, bringing #15729 onto the site.


hzxuzhonghu:
LGTM

kfaseela:
cc @clucavallin

kfaseela:
The idea LGTM. Is there something else other than "Link" which we can use in the title? Not looking great in the preview.

image

craigbox:
We need to (a) differentiate it somehow in the title, as that goes to RSS feeds (b) imply it's not an Istio post and it's a link to outside istio.io. The latter will actually become a direct link in a later PR.

Updated now to "featured post", but I'd like the same prefix for videos as blogs if possible.  We can bikeshed on that later, if this one is ok?

dhawton:
> We need to (a) differentiate it somehow in the title, as that goes to RSS feeds (b) imply it's not an Istio post and it's a link to outside istio.io. The latter will actually become a direct link in a later PR.
> 
> Updated now to "featured post", but I'd like the same prefix for videos as blogs if possible. We can bikeshed on that later, if this one is ok?

Maybe instead of "Featured Post" we just use something similar to "External Blog" to signify it's not Istio-hosted.

craigbox:
Updated, but I'm not sure that's the right word for long term.   Kicked off a Slack thread and we can change it later if we wish.

kfaseela (approved):

PR #15774, opened 2024-10-09
Add back ambient helm inplace upgrade docs

bleggett:
## Description

Previous to https://github.com/istio/istio.io/pull/15247 we (only) had an "in-place" Helm upgrade guide.

After that we had complete a "revisioned" upgrade guide but that guide contains several steps that are not necessary if you do not care about using revisions in your environment.

This adds both flavors back as tabs (both with doctests) and tidies up a few bits of language/warnings, which was originally suggested as a followup in https://github.com/istio/istio.io/pull/15247.

cc @therealmitchconnors and @howardjohn 

## Reviewers

- [x] Ambient
- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/upgrade/helm/index.md
+++ b/content/en/docs/ambient/upgrade/helm/index.md
@@ -15,22 +15,18 @@ Follow this guide to upgrade and configure an ambient mode installation using
 [Helm](https://helm.sh/docs/). This guide assumes you have already performed an [ambient mode installation with Helm](/docs/ambient/install/helm/) with a previous version of Istio.
 
 {{< warning >}}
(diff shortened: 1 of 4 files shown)


dhawton (commented):
Missed period. There's also a duplicated header in the markdown that lint found.

bleggett:
> Missed period. There's also a duplicated header in the markdown that lint found.

Yes, unfortunately it doesn't seem to be smart enough to ignore dupe headers within tabs.

dhawton:
> > Missed period. There's also a duplicated header in the markdown that lint found.
> 
> Yes, unfortunately it doesn't seem to be smart enough to ignore dupe headers within mutually-exclusive tabs.

It's because headers are used to generate anchor refs using a stub of the text of the header. Can't create anchor refs with dup titles, at least easily.

dhawton (approved):

PR #15802, opened 2024-10-16
scale test blog

therealmitchconnors:
## Description

Blog post comparing Istio Ambient and Cilium performance at scale.


content/en/blog/2024/ambient-vs-cilium/index.md
```diff
@@ -0,0 +1,67 @@
...
+title: "Scaling in the Clouds: Istio Ambient vs. Cilium"
+description: A Deep Dive into Performance at Scale.
+publishdate: 2024-10-16
+attribution: "Mitch Connors (Microsoft)"
+keywords: [istio,cilium,analysis]
+---
+
+A common question from prospective Istio users is "how does Istio compare to Cilium?"  While Cilium originally only provided L3/L4 Network Policy, recent releases have added service mesh functionality with Envoy, as well as Wireguard Encryption. Like Istio, Cilium is a CNCF Graduated project, and has been around in the community for many years.
+
+Despite offering a similar feature set on the surface, the two projects have substantially different architectures, most notably cilium’s eBPF program for processing L4 in the kernel, contrasted with Istio’s ztunnel component for L4 in user space. These differences have resulted in substantial speculation about how Istio will perform at scale compared to Cilium.
```
howardjohn, line 11:
```suggestion
Despite offering a similar feature set on the surface, the two projects have substantially different architectures, most notably Cilium’s eBPF programs for processing L4 in the kernel, contrasted with Istio’s ztunnel component for L4 in user space. These differences have resulted in substantial speculation about how Istio will perform at scale compared to Cilium.
```

content/en/blog/2024/ambient-vs-cilium/index.md
```diff
@@ -0,0 +1,67 @@
...
+{{< image width="60%"
+    link="./scale-scenario.png"
+    alt="Scaling to 500 services with 50,000 pods."
+    >}}
+
+For the Istio test, we used Istio’s Ambient Mode, with a Waypoint in every service namespace, and default install parameters. In order to make our test scenarios similar, we had to turn on a few non-default features in Cilium, including Wireguard Encryption, L7 Proxies, and Node Init. We also created a Cilium Network Policy in each namespace, with HTTP path-based rules. In both scenarios, we generated churn by scaling one service to between 85 and 115 instances at random every second, and relabeling one namespace every minute. To see the precise settings we used, and to reproduce our results, see this site.
+
+## Scalability Scorecard
+
+{{< image width="80%"
```
howardjohn, line 28: I think its unfair to give a checkmark for "Total cores". If I were to add a 'sleep(2s)' call, it would dominate both of these, which is a bit of a sign its not a great measurement. CPU should only be compared wrt load, which is what we already have with QPS/core.

Similar applies to memory and latency, tbh, but to a _much_ lesser degree (and would need to be measured by testing at fixed load vs computing QPS/memory or something), so I think those are likely fine.

content/en/blog/2024/ambient-vs-cilium/index.md
```diff
@@ -0,0 +1,67 @@
...
+    link="./scale-scorecard.png"
```
howardjohn, line 29: How does Istio have ~1.7x QPS but ~3.5x bandwidth?

content/en/blog/2024/ambient-vs-cilium/index.md
```diff
@@ -0,0 +1,67 @@
...
+    alt="Scalability Scorecard: Istio vs. Cilium!"
```
howardjohn, line 30: somewhat nitpick: the bars being side-by-side instead of stacked make it hard to compare. Maybe an illusion but Istio's total cores looks shorter despite being a higher value, for instance

content/en/blog/2024/ambient-vs-cilium/index.md
```diff
@@ -0,0 +1,67 @@
...
+    >}}
+
+* **The Cilium Slowdown** Cilium, while boasting impressive low latency with default install parameters, slows down substantially when Istio’s baseline features such as L7 policy and encryption are turned on. Additionally, cilium’s memory and CPU utilization remained high even when no traffic was flowing in the mesh. This can impact the overall stability and reliability of your cluster, especially as it grows.
```
howardjohn, line 33: > while boasting impressive low latency with default install parameters

Given we don't show any data, test setup, explanation, etc of this, it feels out of place

content/en/blog/2024/ambient-vs-cilium/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Hidden Costs
+
+While Istio operates entirely in user space, Cilium’s L4 dataplane runs in the Linux Kernel using eBPF. Prometheus metrics for resource consumption only measure user space resources, meaning that all Kernel resources used by Cilium are not accounted for in this test.
+
+## Recommendations: Choosing the Right Tool for the Job
+
+So, what's the verdict? Well, it depends on your specific needs and priorities. If low latency is your absolute top priority and you have a smaller cluster, Cilium might still be a viable option. However, for larger clusters and a focus on stability, scalability, and advanced features, Istio Ambient Mode, along with an alternate NetworkPolicy implementation is the way to go.
```
howardjohn, line 61: I am worried about this recommendation of low latency -> Cilium.

1. We are comparing Cilium L7 vs Istio L7 _only_ in this test. We have not proven any claims that any setup outside of this scope is faster or slower.
2. If we talk all about "Cilium L7", then say "use Cilium", the assumption is "Use Cilium [L7 mode]", which is not what the intended message here is (AFAIK)
3. I disagree with the assertion that Cilium L4 Wireguard is lower latency than Istio L4 TLS based on past testing.
4. Even if we are comparing apples-to-oranges of Cilium unencrypted to Ztunnel TLS, I still think the recommendation is wrong. Its like if a grilled hotdog has a 0.00000002% chance of causing health issues while a raw hotdog has a 0.00000001% chance --- the raw one may be 2x better, but _who cares_. Similar, IMO, applies to the latency overheads we are seeing with Ztunnel TLS. Adding 0.05ms to p90s is simply irrelevant to 99.99% of users

howardjohn: I also don't know that we _need_ to recommend you use a networkpolicy on top of ambient...?

content/en/blog/2024/ambient-vs-cilium/index.md
howardjohn: Also this says "higher is better" but its not the case for CPU usage, memory, latency? Maybe it refers to the length of the bars?

therealmitchconnors: Correct.  Lower-is-better metrics have been transformed with 1/n so that the longest bar always represents the winner.

content/en/blog/2024/ambient-vs-cilium/index.md
therealmitchconnors: For one thing, QPS is load balancing 500 clients against 50k servers.  Bandwidth is point-to-point.  QPS may include connection and handshake semantics, while bandwidth only includes the initial connection.  I'm sure there are other differences too.

content/en/blog/2024/ambient-vs-cilium/index.md
therealmitchconnors: I have substantially reworked this paragraph, which may address some of your concerns.  Without netpol, Ambient has no egress controls, which seem important from a defense standpoint...

content/en/blog/2024/ambient-vs-cilium/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Not all Encryption is Created Equal
+
+While Cilium offers IPsec for FIPS-compatible encryption, most other Cilium features such as L7 Policy and Load Balancing are incompatible with IPsec. Cilium has much better feature compatibility when using WireGuard encryption, but WireGuard cannot be used in FIPS-compliant environments. Istio, on the other-hand, uses FIPS-compliant mTLS by default, which has the added benefit of complying with Zero-Trust principles.
+
+### Hidden Costs
+
+While Istio operates entirely in user space, Cilium’s L4 dataplane runs in the Linux Kernel using eBPF. Prometheus metrics for resource consumption only measure user space resources, meaning that all Kernel resources used by Cilium are not accounted for in this test.
```
bleggett, line 57:
```suggestion
While Istio operates entirely in user space, Cilium’s L4 dataplane runs in the Linux kernel using eBPF. Prometheus metrics for resource consumption only measure user space resources, meaning that all kernel resources used by Cilium are not accounted for in this test.
```

Is `Kernel` typically capitalized? AFAICT not typically?

content/en/blog/2024/ambient-vs-cilium/index.md
howardjohn: I don't get it. A request is, say 1kb. If I send 1000 requests that's 1mb bandwidth. 

Or is it counting waypoint requests as effectively 2000 requests to send 1000 requests (since we have to waypoint and from waypoint) while Cilium doesn't (even though they have the same number of hops, one is node local)? that would align with bandwidth being roughly 2x?

content/en/blog/2024/ambient-vs-cilium/index.md
howardjohn: (1) still remains but it's a bit pedantic so probably fine

I'd note that with the new framing basically any CNI/kube-proxy/etc fits as well as cilium (just a note to myself really, not expecting a change to the blog...)

content/en/blog/2024/ambient-vs-cilium/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+While the objective of this project is to compare Istio and Cilium scalability, several constraints make a direct comparison difficult.
+
+### Layer 4 Isn’t always Layer 4
+
+While Istio and Cilium both offer L4 policy enforcement, their APIs and implementation differ substantially. Cilium implements Kubernetes NetworkPolicy, which uses labels and namespaces to block or allow access to and from IP Addresses. Istio offers an AuthorizationPolicy API, and makes allow and deny decisions based on the TLS identity used to sign each request. Most defense-in-depth strategies will need to make use of both NetworkPolicy and TLS-based policy for comprehensive security.
+
+### Not all Encryption is Created Equal
+
+While Cilium offers IPsec for FIPS-compatible encryption, most other Cilium features such as L7 Policy and Load Balancing are incompatible with IPsec. Cilium has much better feature compatibility when using WireGuard encryption, but WireGuard cannot be used in FIPS-compliant environments. Istio, on the other-hand, uses FIPS-compliant mTLS by default, which has the added benefit of complying with Zero-Trust principles.
```
bleggett, line 53:
```suggestion
While Cilium offers IPsec for FIPS-compatible encryption, most other Cilium features such as L7 Policy and Load Balancing are incompatible with IPsec. Cilium has much better feature compatibility when using WireGuard encryption, but WireGuard cannot be used in FIPS-compliant environments. Istio, on the other-hand, because it strictly complies with TLS protocol standards, always uses FIPS-compliant mTLS by default.
```

`which has the added benefit of complying with Zero-Trust principles.` reads as vague/handwavy opinion to me. That may be the intent, in which case ignore.

It might also be clearer and more readable to just change the comma to a period and drop the coda, eg:

```
Istio, on the other-hand, uses FIPS-compliant mTLS by default.
```

content/en/blog/2024/ambient-vs-cilium/index.md
therealmitchconnors: That may account for the some of the difference as well.

content/en/blog/2024/ambient-vs-cilium/index.md
craigbox: That's deliberate, the metric here is actually 1/cores in order that higher is always better.

content/en/blog/2024/ambient-vs-cilium/index.md
therealmitchconnors: If you're referring to the comment that cilium is fast with default params, I think this is generally assumed, and I do have test data to support it.  It's not the focus of this blog post, but I think it's important for fairness to note that Cilium is quite fast when not using encryption and L7...

craigbox (approved):

PR #15853, opened 2024-10-31
Change the typeface used on istio.io

craigbox:
I find our current body typeface, Barlow, a bit hard to read, at least at the weights we use it.
It also has unfixed kerning issues.

This PR experiments with changing it - I have picked Inter as a modern web-friendly option, so we can start seeing what it might look like, and consider what weights we would need to adjust.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -120,7 +120,7 @@
         <link rel="stylesheet" href="/css/all.css">
         <link rel="preconnect" href="https://fonts.googleapis.com">
         <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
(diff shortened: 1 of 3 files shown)


ilrudie:
subtle difference but I like it, TY

Andygol:
🤞 I hope new font supports Cyrillic

Andygol:
According to <https://fonts.google.com/specimen/Barlow/glyphs>, Barlow has no support for scripts other than Latin. Inter (https://fonts.google.com/specimen/Inter/glyphs) supports Latin, Greek, and Cyrillic.

craigbox:
We're already using Barlow, and you have a site in Ukrainian, which uses the Cyrillic alphabet...?

Andygol:
If you compare English version and Ukrainian, you will see differences in glyphs such as "e" (Lat) and "е" (Cyr) on the screenshots, width of letters is different and so on. And, yes, Ukrainian version use some fallback font.

istio.io | istio.io/latest/uk/
---|---
image |  image

PR #15881, opened 2024-11-07
ambient ga blog

linsun:
## Reviewers

- [ x ] Ambient
- [ x ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

PR #15883, opened 2024-11-07
Copy edits for ambient launch blogs

craigbox:
Update our 2022 launch blogs to clarify the difference between "ambient mesh" (a pattern) and Istio's ambient mode (the reference implementaiton of such), and link to the fact ambient mode is now GA.

(Will require #15881 to merge for the lint to pass)


craigbox:
told you so.

/test lint

kfaseela (approved):

PR #15884, opened 2024-11-07
Small fix in 1.24 release notes

kfaseela:


kfaseela:
Already covered in https://github.com/istio/istio.io/pull/15881/files

PR #15894, opened 2024-11-07
Update our company logo

linsun:
## Reviewers

- [ ] Ambient
- [ x ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


data/companies.yml
```diff
@@ -20,7 +20,7 @@ providers:
     logo: "/logos/tetrate.svg"
     url: "https://www.tetrate.io/tetrate-service-bridge/"
   - name: "Gloo Mesh by Solo.io"
-    logo: "/logos/solo.png"
+    logo: "/logos/gloomesh.svg"
```
dhawton, line 23:
```suggestion
    logo: "/logos/solo.svg"
```

The companies should use the company's logo

dhawton (changes requested):

data/companies.yml
linsun: That is intentional because this is gloo mesh in the context of the page.  Not solo.io.

dhawton: Under the `providers` page we use the company's logo rather than a product logo: https://deploy-preview-15894--preliminary-istio.netlify.app/latest/about/ecosystem

linsun: Ok, i see that now, a bit confusing as name is gloo mesh. :). pushed a change for it.

PR #15907, opened 2024-11-08
Fix revision in uninstall previous control plane

sridhargaddam:
The upgrade documentation uses references to both `OLD_REVISION` as well as `REVISION`, but as part of "Uninstall the previous control plane" section instead of using "OLD_REVISION", it uses "REVISION". This PR fixes it.

### DIFF
--- a/content/en/docs/ambient/upgrade/helm/index.md
+++ b/content/en/docs/ambient/upgrade/helm/index.md
@@ -30,7 +30,7 @@ Like sidecar mode, gateways can make use of [revision tags](/docs/setup/upgrade/
 
 ### Prepare for the upgrade
 
-Before upgrading Istio, we recommend downloading the new version of istioctl, and running `istioctl x precheck` to make sure the upgrade is compatible with your environment. The output should looks something like this:
+Before upgrading Istio, we recommend downloading the new version of istioctl, and running `istioctl x precheck` to make sure the upgrade is compatible with your environment. The output should look something like this:
 
 {{< text syntax=bash snip_id=istioctl_precheck >}}
 $ istioctl x precheck
@@ -214,7 +214,7 @@ $ helm upgrade istio-ingress istio/gateway -n istio-ingress
 If you have upgraded all data plane components to use the new revision of the Istio control plane, and are satisfied that you do not need to roll back, you can remove the previous revision of the control plane by running:
 
 {{< text syntax=bash snip_id=none >}}
-$ helm delete istiod-"$REVISION" -n istio-system
+$ helm delete istiod-"$OLD_REVISION" -n istio-system
 {{< /text >}}
 
 {{< /tab >}}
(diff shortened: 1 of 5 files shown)


kfaseela:
errors are unrelated, should be fixed by https://github.com/istio/istio.io/pull/15909

content/en/docs/ambient/upgrade/helm/index.md
```diff
@@ -214,7 +214,7 @@ $ helm upgrade istio-ingress istio/gateway -n istio-ingress
 If you have upgraded all data plane components to use the new revision of the Istio control plane, and are satisfied that you do not need to roll back, you can remove the previous revision of the control plane by running:
 
 {{< text syntax=bash snip_id=none >}}
```
bleggett, line 216: Really these should use doctests, which would prevent a lot of this stuff being out of sync.

content/en/docs/ambient/upgrade/helm/revision_test.sh
```diff
@@ -49,5 +49,5 @@ _rewrite_helm_repo snip_rollback_tag
 # upgrading a tag creates an MWC, let's clean it up 
 export REVISION=istio-1-22-1
 helm template istiod istio/istiod -s templates/revision-tags.yaml --set revisionTags="{tagname}" --set revision="$OLD_REVISION" -n istio-system | kubectl delete -f -
-helm delete istiod-"$REVISION" -n istio-system
+helm delete istiod-"$OLD_REVISION" -n istio-system
```
bleggett, line 52: I think we need to delete everything here in `cleanup` - $REVISION as well as $OLD_REVISION.

I think this won't remove $REVISION istiod anymore?

sridhargaddam: TBH, I'm not familiar with the doctests. Let me add cleanup for $REVISION as well.

sridhargaddam: Thanks for reviewing @bleggett. All tests are passing, hope its all good :D

content/en/docs/ambient/upgrade/helm/index.md
bleggett, line 216: You don't need to manually keep the tests and the docs in sync - make this

```suggestion
{{< text syntax=bash snip_id=delete_old_revision >}}
```

and this will create a `bash` function with the same name you can call in `revision_test.sh`, so the docs and the tests stay in sync.

content/en/docs/ambient/upgrade/helm/revision_test.sh
```diff
@@ -48,6 +48,8 @@ _rewrite_helm_repo snip_rollback_tag
 
 # upgrading a tag creates an MWC, let's clean it up 
 export REVISION=istio-1-22-1
+export OLD_REVISION=istio-1-21-2
```
bleggett, line 51: As mentioned above, you can generate `delete_old_revision` from tthe docs and call it here, so you don't have to hardcode the same commands in 2 spots.

craigbox (approved):

PR #15922, opened 2024-11-11
Include platform details while installing ztunnel on OpenShift

sridhargaddam:
On OpenShift, installing the ztunnel chart requires specifying the global.platform details. This PR includes that update.

### DIFF
--- a/content/en/docs/ambient/install/platform-prerequisites/index.md
+++ b/content/en/docs/ambient/install/platform-prerequisites/index.md
@@ -200,17 +200,21 @@ For example:
 
 OpenShift requires that `ztunnel` and `istio-cni` components are installed in the `kube-system` namespace, and that you set `global.platform=openshift` for all charts.
 
-If you use `helm`, you can set the target namespace and `global.platform` values directly.
-
-If you use `istioctl`, you must use a special profile named `openshift-ambient` to accomplish the same thing.
-
 {{< tabset category-name="install-method" >}}
 
 {{< tab name="Helm" category-value="helm" >}}
 
+    You must `--set global.platform=openshift` for **every** chart you install, for example with the `istiod` chart:
+
     {{< text syntax=bash >}}
-    $ helm install istio-cni istio/cni -n kube-system --set profile=ambient --set global.platform=openshift --wait
-    $ helm install ztunnel istio/ztunnel -n kube-system --set global.platform=openshift --wait
+    $ helm install istiod istio/istiod -n istio-system --set profile=ambient --set global.platform=openshift --wait
+    {{< /text >}}
+
+    In addition, you must install `istio-cni` and `ztunnel` in the `kube-system` namespace, for example:
+
+    {{< text syntax=bash >}}
+    $ helm install istio-cni istio/istio-cni -n kube-system --set profile=ambient --set global.platform=openshift --wait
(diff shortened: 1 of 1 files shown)


content/en/docs/ambient/install/platform-prerequisites/index.md
```diff
@@ -210,6 +210,7 @@ If you use `istioctl`, you must use a special profile named `openshift-ambient`
 
     {{< text syntax=bash >}}
     $ helm install istio-cni istio/cni -n kube-system --set profile=ambient --set global.platform=openshift --wait
```
bleggett, line 212: (ty for catching)

Actually, you also need to `--set global.platform=openshift` for `istiod` as well, so this is still incorrect.

In the docs, `platform` should always be a blanket statement for all charts. 

We should not say "use `--set global.platform=openshift` for the charts listed here", we should _always_ say "set `--set global.platform=openshift` for all charts you install".
 
Can you rework this section say that `--set global.platform=openshift` should _always_ be set for _all_ charts, full-stop, and remove the list of charts here?

bleggett: Something like 

```
<tab name="Helm">

You must `--set global.platform=openshift` for **every** chart you install, for example with the `istiod` chart:

$ helm install istiod istio/istiod -n istio-system --set profile=ambient --set global.platform=openshift --wait

You must install `istio-cni` and `ztunnel` to the `kube-system` namespace, for example:

$ helm install ztunnel istio/ztunnel -n kube-system --set profile=ambient --set global.platform=openshift --wait
$ helm install istio-cni istio/istio-cni -n kube-system --set profile=ambient --set global.platform=openshift --wait

</tab>
```

would also work.

sridhargaddam: Good catch, let me update the PR. Thank you @bleggett

sridhargaddam: @bleggett I see that some of the sentences appear to be a bit redundant. For example, we already have these notes captured here - https://github.com/istio/istio.io/blob/37ca41844bc1d2d72127b1644dd40bda2dac14f6/content/en/docs/ambient/install/platform-prerequisites/index.md?plain=1#L201

Is it okay, or do you have any suggestions for improving it?

bleggett: @sridhargaddam 

Looking at it - let's maybe drop these two sentences:

```
If you use `helm`, you can set the target namespace and `global.platform` values directly.

If you use `istioctl`, you must use a special profile named `openshift-ambient` to accomplish the same thing.
```

since they're just restating what the example shows.

sridhargaddam: Since the ["Red Hat OpenShift" section](https://istio.io/latest/docs/ambient/install/platform-prerequisites/#red-hat-openshift) already includes the two points, shall we just include the missing istiod and the ztunnel chart in the helm section?

```
    {{< text syntax=bash >}}
    $ helm install istiod istio/istiod -n istio-system --set profile=ambient --set global.platform=openshift --wait
    $ helm install istio-cni istio/istio-cni -n kube-system --set profile=ambient --set global.platform=openshift --wait
    $ helm install ztunnel istio/ztunnel -n kube-system --set profile=ambient --set global.platform=openshift --wait
    {{< /text >}}
```

Issue #15891, opened 2024-11-07
FAQ in Ukrainian doesn't work as expected


Andygol:
I was already mentioned the issue in the PR https://github.com/istio/istio.io/pull/15677

>  I couldn't get the desired result locally with the FAQ page in Ukrainian, while the English and Chinese versions work as expected. I have an issue (locally), when I click on a question, no action is taken, the answer section is not displayed.

I was hoping that the automation scripts didn't work at the local level. However, even after the Ukrainian translation was merged with the main documentation, the situation did not change for the better.

https://preliminary.istio.io/latest/uk/about/faq

I kindly request your consideration of this matter and any necessary corrections. 🙏

PR #15930, opened 2024-11-13
Fix FAQ in Ukrainian doesn't work as expected

Andygol:
## Description

Fixes: #15891

This pull request includes a minor change to the `handleEvents` function in `src/ts/events.ts`. The change ensures that the `banner` variable is checked for both existence and length before proceeding.

This fix allows you to get the expected behavior on the FAQ page when clicking on a question leads to toggle the answer to it. 

It would be great to cherry-pick this to v[1.24](https://github.com/istio/istio.io/tree/release-1.24).

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/src/ts/events.ts
+++ b/src/ts/events.ts
@@ -161,7 +161,7 @@ function handleEvents(): void {
     function toggleActiveHeader(): void {
         const top = window.scrollY;
 
-        if (!banner) {
+        if (!banner || !banner.length) {
             return;
         }
 


kfaseela:
And it works: https://deploy-preview-15930--preliminary-istio.netlify.app/latest/uk/about/faq

kfaseela:
Though I don't understand which banner value was empty in `uk` case, but it is not so for `en` ?

Andygol:
> Though I don't understand which banner value was empty in `uk` case, but it is not so for `en` ?

Me neither, but it was an error in a console log that helped to make this fix.

kfaseela:
LGTM, but I will leave to @craigbox for merging, as he knows this area better

craigbox:
I guess the error is if there is an EN banner, but the translation for it in UK is missing, so zero length?

craigbox (approved):

craigbox:
As a rule you can backport any and all docs changes, except those that are only relevant to features in 1.25, to `release-1.24`

Issue #15846, opened 2024-10-30
V2 IP Autoallocation Controller Documentation Gaps


ilrudie:
# Tracking gaps which come up from [feature feedback](https://github.com/istio/istio/issues/53596)

- [ ] Make sure it is clear that IPs will change when migrating from ephemeral to status-based IP assignement

PR #15931, opened 2024-11-13
Document DNS Auto Allocation Version 2

kfaseela:
## Description

Attempt to document https://github.com/istio/istio/pull/51828/

Fixes https://github.com/istio/istio.io/issues/15846

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
+++ b/content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
@@ -281,7 +281,6 @@ $ kubectl exec deploy/curl -- curl -sS -v auto.internal
 * shutting down connection #0
 {{< /text >}}
 
-
 ## Cleanup
 
 {{< text bash >}}
--- a/content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
+++ b/content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
@@ -101,7 +101,7 @@ to [the following section](#external-tcp-services-without-vips) for more details
 To work around these issues, the DNS proxy additionally supports automatically allocating addresses for `ServiceEntry`s that do not explicitly define one. This is configured by the `ISTIO_META_DNS_AUTO_ALLOCATE` option.
 
 {{< tip >}}
-Please see [DNS Auto Allocation V2](/docs/ops/configuration/traffic-management/dns-proxy/#dns-auto-allocation-v2) for a new enhanced implementation of auto allocation supported by Istio from 1.23 onwards. DNS Auto Allocation V2 is the recommended for sidecar mode and required for ambient mode.
+Please see [DNS Auto Allocation V2](/docs/ops/configuration/traffic-management/dns-proxy/#dns-auto-allocation-v2) for a new enhanced implementation of auto allocation supported by Istio from 1.23 onwards. DNS Auto Allocation V2 is recommended for sidecar mode and required for ambient mode.
 {{< /tip >}}
 
 When this feature is enabled, the DNS response will include a distinct and automatically assigned address for each `ServiceEntry`. The proxy is then configured to match requests to this IP address, and forward the request to the corresponding `ServiceEntry`. When using `ISTIO_META_DNS_AUTO_ALLOCATE`, Istio will automatically allocate non-routable VIPs (from the Class E subnet) to such services as long as they do not use a wildcard host. The Istio agent on the sidecar will use the VIPs as responses to the DNS lookup queries from the application. Envoy can now clearly distinguish traffic bound for each external TCP service and forward it to the right target. For more information check respective [Istio blog about smart DNS proxying](/blog/2020/dns-proxy/#automatic-vip-allocation-where-possible).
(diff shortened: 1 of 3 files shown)


kfaseela:
cc @ilrudie @howardjohn

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -219,6 +223,64 @@ A virtual IP address will be assigned to every service entry so that client side
     ADDRESS=240.240.69.138, DESTINATION=Cluster: outbound|9000||tcp-echo.external-1.svc.cluster.local
     {{< /text >}}
 
+## DNS Auto Allocation V2
+
+Istio now offers an enhanced implementation of DNS Auto Allocation. To use the new feature, replace the feature flag `ISTIO_META_DNS_AUTO_ALLOCATE`, which was used in the previous example, with `PILOT_ENABLE_IP_AUTOALLOCATE` while installing Istio. All examples given so far would work as is.
```
ilrudie, line 228: This is a bit strange because the enablement also moves from meshConfig to a pilot environment variable. Should we mention that somehow? A quick reading of this without looking at your yaml might assume it is still in meshConfig

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
```diff
@@ -100,6 +100,10 @@ to [the following section](#external-tcp-services-without-vips) for more details
 
 To work around these issues, the DNS proxy additionally supports automatically allocating addresses for `ServiceEntry`s that do not explicitly define one. This is configured by the `ISTIO_META_DNS_AUTO_ALLOCATE` option.
 
+{{< tip >}}
+Please see [DNS Auto Allocation V2](/docs/ops/configuration/traffic-management/dns-proxy/#dns-auto-allocation-v2) for a new enhanced implementation of auto allocation supported by Istio from 1.23 onwards.
```
bleggett, line 104: Should we explicitly mention that `ISTIO_META_DNS_AUTO_ALLOCATE` is deprecated/not preferred, if in fact that is the case? Implying that it will be removed in N releases?

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
kfaseela: Sure, let me clarify this part

content/en/docs/ops/configuration/traffic-management/dns-proxy/index.md
kfaseela: Have modified the tip to include this, please take a look.

kfaseela: Should I revert the deprecation part then? Please decide and let me know.

ilrudie: I would probably vote to not mention deprecation in our documentation until that is the official decision. We should probably bring it up in the community WG I suppose. Easy enough to adjust the note later. @bleggett, would you be opposed to leaving it out until we know for sure it is our plan to remove the legacy impl?

bleggett: > 1 sort of related topic though is that there is near as makes no different no support for v1 in ambient. If you are using ambient then you need to use v2.

We should, at least, mention this explicitly here.

kfaseela: > I'm fine waiting on a deprecation intent decision - but in general I don't think we should be documenting features that exactly and fully replace other features (as this does AFAICT) if we don't at least articulate an opinion on what we _plan_ to do with the old one to users.
> 
> "We have two things that do exactly the same thing. Here they are." is not ideal messaging, and provides no guidance.

In this case then how about replacing the old content with the new flag, and not mention anything about the old flag?

bleggett: > In this case then how about replacing the old content with the new flag, and not mention anything about the old flag?

Probably too confusing/disruptive to do that until we have confirmation the old one is going to be deprecated (and we actually deprecate it), like @ilrudie said.

```
Please see [DNS Auto Allocation V2](/docs/ops/configuration/traffic-management/dns-proxy/#dns-auto-allocation-v2) for an enhanced implementation of auto allocation supported by Istio from 1.23 onwards. This enhanced implementation is recommended for sidecar mode and required for ambient mode.
```

ilrudie: @bleggett suggested wording reads nicely to me. big +1

kfaseela: @bleggett : Done

content/en/docs/ops/configuration/traffic-management/dns-proxy/test.sh
```diff
@@ -27,7 +27,7 @@ snip_getting_started_1
 # deploy test application
 snip_dns_capture_in_action_2
 
-# configure service entries and verify
+# configure service entries #and verify
```
bleggett, line 30:
```suggestion
# configure service entries and verify
```

bleggett, line 30:
```suggestion
# configure service entries and verify
```

bleggett (approved):

kfaseela:
@bleggett @ilrudie now that 1.24 docs are updated, how do we want to refactor this page for 1.25 where the new flag is enabled by default?

ilrudie:
Good question. That PR which changed the feature to on by default included creating a compatibility profile so the existence of the profile and how to use it to get the 1.24 behavior is probably worth considering.

PR #15936, opened 2024-11-14
Fix annotation name in release notes

howardjohn:


bleggett (approved):

wilsonwu:
`ambient.istio.io/bypass-inbound-capture"` is that a typo? Should be `ambient.istio.io/bypass-inbound-capture`?

PR #15979, opened 2024-11-22
Removing deprecated `istioctl verify-install` from "Getting Started" with ambient

grnmeira:
## Description

This PR removes a reference to the deprecated command `istioctl verify-install`. It uses a `kubectl` check instead.

## Reviewers

- [X] Ambient
- [X] Docs
- [X] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/getting-started/_index.md
+++ b/content/en/docs/ambient/getting-started/_index.md
@@ -51,10 +51,6 @@ It might take a minute for the Istio components to be installed. Once the instal
 ✔ Installation complete
 {{< /text >}}
 
-{{< tip >}}
-You can verify the installed components using the command `kubectl get all -n istio-system`.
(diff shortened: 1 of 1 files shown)


Arhell (commented):
/ok-to-test

kfaseela:
@howardjohn @bleggett I hope this is not a bug to be fixed, and really something that has been removed. If so, we can merge this PR

howardjohn (commented):
I would just remove the command entirely TBH. `kubectl get all` doesn't really verify the install worked. The install command will fail if the install doesn't work

craigbox (approved):

craigbox:
I'll do a pass of this whole doc set for 1.24 too, so no need to cherry pick this.

PR #16010, opened 2024-11-26
Change the typeface used on istio.io, but a little less

craigbox:
Adjust weights and heights a little to see if we should stick with what we have. Compare with #15853.

### DIFF
--- a/assets/sass/base/_base.scss
+++ b/assets/sass/base/_base.scss
@@ -556,9 +556,9 @@ iframe {
 }
 
 .article-container {
+    line-height: 1.75rem;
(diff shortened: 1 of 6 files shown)


craigbox:
Code within pre rendering odd.  See [here](https://deploy-preview-16010--preliminary-istio.netlify.app/latest/docs/setup/install/istioctl/) for an example.  see also [the red box](https://deploy-preview-16010--preliminary-istio.netlify.app/latest/docs/setup/install/istioctl/#generate-a-manifest-before-installation) looking off

Issue #16012, opened 2024-11-26
bug: the blog year selectors overflow


craigbox:
Istio is too old - we have too many years in our blog!

under 1200px:

<img width="1174" alt="Screenshot 2024-11-26 at 10 33 38 PM" src="https://github.com/user-attachments/assets/ded914c0-d589-4e8d-93a8-9c48b67e4a2a">

above 1200px:

<img width="953" alt="Screenshot 2024-11-26 at 10 34 39 PM" src="https://github.com/user-attachments/assets/cad15e0e-9733-4748-a7c0-fcef89aa30c7">

Perhaps we make these lovely pill-shaped buttons into a CSS grid?

mike-weiner:
Hey, @craigbox! 👋🏻 I've got a fairly small PR out to fix this. https://github.com/istio/istio.io/pull/16022

PR #16022, opened 2024-11-28
Move Blog Year Buttons to Grid Layout

mike-weiner:
## Description

Fixes: https://github.com/istio/istio.io/issues/16012

This PR:

- Moves the year buttons on https://istio.io/latest/blog/ to a flex layout that allows for wrapping.
- Remove the overflow support and instead use an `@media` query to adjust button sizing.
- Reduces the spacing between the year buttons.
- Change the cursor to the hover cursor type when a person hovers on the year buttons.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/src/sass/misc/_blog.scss
+++ b/src/sass/misc/_blog.scss
@@ -29,20 +29,17 @@
     }
 
     .year-buttons {
-        display: flex;
-        overflow: auto;
-        white-space: nowrap;
-        @media (min-width: $bp-xl) {
-            white-space: normal;
-            overflow: unset;
-            justify-content: center;
+        display: grid;
+        grid-template-columns: repeat(1, 1fr);
+        @media (min-width: $bp-md) {
+            grid-template-columns: repeat(4, 1fr);
         }
 
         .year-button {
             padding: .5em 3.5em;
             min-width: 168px;
             height: 36px;
-            margin: 2.5em .5em 3.75em;
+            margin: 1.0em;
             border-radius: 28px;
             border: solid 2px $inputBorderColor;
             background-color: #fff;
@@ -51,8 +48,6 @@
             letter-spacing: normal;
             color: $linkActiveColor;
             @media (min-width: $bp-md) {
-                margin: 2.5em 1.15em 3.75em 0;
-
                 &:hover {
                     outline: 0;
                     -webkit-box-shadow: 2px 2px 15px 0 rgba(101, 126, 186, 1);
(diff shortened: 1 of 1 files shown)


craigbox:
thanks for taking this on! I'm a bit worried about the article so far below the fold on mobile, given there are so many buttons. Might it be better to think of some kind of drop-down for that situation

mike-weiner:
> thanks for taking this on! I'm a bit worried about the article so far below the fold on mobile, given there are so many buttons. Might it be better to think of some kind of drop-down for that situation

@craigbox - Yea... I didn't love that. 

A dropdown is one idea. I could also work to make the buttons a bit smaller on mobile, something like this:

Screenshot 2024-11-27 at 11 57 43 PM

craigbox:
That's better, thanks.
tbh I doubt people regularly want to filter by year. Regular pagination is probably plenty.

Arhell (commented):
https://deploy-preview-16022--preliminary-istio.netlify.app/latest/news/

mike-weiner:
I'm really liking how https://deploy-preview-16022--preliminary-istio.netlify.app/latest/blog/ looks and adapts to smaller screen sizes now. 

I'm less of a fan of how https://deploy-preview-16022--preliminary-istio.netlify.app/latest/news/ looks and adapts to smaller screen sizes.

Thoughts?

craigbox:
I tried your first version and I think there was space between the pills; I would often end up zooming into the page when I was trying to tap a year.  The second version actually fixes that, which is great!

Nits, all of which probably predate your change:

- the pills aren't the same size 
Screenshot 2024-11-30 at 10 34 05 PM

- will this look weird when we add one more pill?

- the margin only being on the right of each pill means they don't line up visually
Screenshot 2024-11-30 at 10 36 55 PM

(seems to work OK if you change `margin-right: 1rem` to `margin-left: 0.5rem; margin-right: 0.5rem`)

mike-weiner:
@craigbox -

Yep, good call on the margin. I've fixed that and the buttons being unequal sizes in 8e191f197b1e94999dc4fdab782e83eb679b7cc9. 

> will this look weird when we add one more pill?

Here is what the Blog page will look like when we get to next year and add a `2025` button :) I'm trying to get away from CSS that sets a fixed height and width in pixels... since that works until it doesn't.... 

I think these all look pretty decent.

Screenshot-2024-11-30-08 38 50AM
Screenshot-2024-11-30-08 38 58AM
Screenshot-2024-11-30-08 39 09AM

craigbox (approved):

PR #16028, opened 2024-12-01
move sass to Hugo pipes

craigbox:
build CSS with Hugo, rather than with our own custom makefile. This means that changes to the .scss files can be previewed in the regular Hugo workflow.


craigbox:
Previews well on Netlify, which uses the extended build of Hugo.
Fails lint, which doesn't.

https://github.com/istio/tools/pull/3074 will fix the lint error by changing which version of Hugo we bundle.

craigbox:
/test lint
as https://github.com/istio/test-infra/pull/5528 has now landed
and we watch the change propagate through the repos

dhawton (approved):

PR #16030, opened 2024-12-02
update resource attachment

### LINKED ISSUE #15952: feature status: Gateway resource attachment & scaling
> Resource attachment is currently experimental.

I don't see how this can currently be. Can we please have a judgment on if this needs to remain.
We should also update the docs to refer to the fact that Gateways have the `gatewayClass` suffix, but waypoints do not.

craigbox:
Reflect that resource attachment is stable in upstream Gateway API, and the generated name used by Istio uses the gateway class suffix for Gateways, but not Waypoints.

Fixes #15952

### DIFF
--- a/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
+++ b/content/en/docs/tasks/traffic-management/ingress/gateway-api/index.md
@@ -224,13 +224,12 @@ Note: only one address may be specified.
 
 #### Resource Attachment and Scaling
 
-{{< warning >}}
-Resource attachment is currently experimental.
-{{< /warning >}}
-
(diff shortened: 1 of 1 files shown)


mikemorris (approved):
Would consider mentioning the label as preferred because it is standardized, refs https://gateway-api.sigs.k8s.io/geps/gep-1762/#resource-attachment

craigbox:
I added a link to the GEP for clarity.  As for the preference, I wasn't sure how to get that across, and left it out in the end. I don't think it's really relevant given these are the Istio docs, documenting Istio's behaviour.  As long as we set the labels/names as we claim we do, I think we're OK.

dhawton (approved):

PR #16039, opened 2024-12-03
Add warning for injection customization

howardjohn:
See https://github.com/istio/istio/issues/51187#


mikebell90:
Thanks @howardjohn I was about to suggest such a warning. TBH is the pod "skeleton" idea ever wise given the unpredictability? Should Istio generally recommend the custom-template option in almost all cases?

kfaseela (approved):

Issue #15755, opened 2024-10-03
Allow overriding prev and next in Hugo frontmatter


craigbox:
/Users/craigbox/Documents/Projects/istio.io-second/layouts/partials/posts_navigation.html

akashsawan1:
Hi @craigbox , I would like to work on this. Please Assign this to me.

craigbox:
If you want to work on something, please send a PR that addresses it. We don't have a culture of claiming issues with assignment in the Istio community. Thanks!

mike-weiner:
@craigbox - Hey! It looks like https://github.com/istio/istio.io/pull/16044 addressed this issue. Can we close this issue out?

PR #16044, opened 2024-12-04
feat: allow overriding prev and next links in frontmatter (#15755)

AdeshGhadage:
## Description

This pull request introduces a feature allowing content authors to override the prev and next navigation links directly in the front matter of a Hugo page. This PR addresses issue #15755

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [X] User Experience


dhawton:
There was a build failure here, so this PR is not yet ready for review.

AdeshGhadage:
I have solved build failure, I am new to istio. any more changed needed?

craigbox:
A brief look at this suggests it isn't going to work how we need.
We need to have the original `.NextInSection`/`.PrevInSection` behavior (including the same formatting) unless `prev` or `next` is set in the front matter,, in which case it should be overridden.

Also, we don't have a `reversePagination` setting.

craigbox:
Thanks adesh!  It looks like it might work but I suspect you've changed  which of "next" and "prev" are left and right.
(That's a really awkward thing about Hugo, so don't worry, it's hard to get right.  Or left!)

To make the testing easier, it would be worth setting up a couple of next: and prev: in the front matter of some pages.  Where I would explicitly suggest you put them:
/docs/ambient/getting-started/ should `next:` to `/docs/ambient/getting-started/deploy-sample-app`, and the latter should `prev:` to the former.

then `/docs/ambient/getting-started/cleanup/` could optionally `next:` to `/docs/ambient/install/` and vice versa.

layouts/partials/posts_navigation.html
```diff
@@ -1,15 +1,23 @@
 {{ $skipPageNav := .Scratch.Get "skipPageNav" }}
 {{ if not (or .Params.skip_pagenav $skipPageNav) }}
-    {{ if or .NextInSection .PrevInSection }}
+    {{$prev := .PrevInSection}}
+    {{$next := .NextInSection}}
```
craigbox, line 4:
```suggestion
    {{ $prev := .PrevInSection }}
    {{ $next := .NextInSection }}
```

craigbox: standard is spaces after/before the moustaches
(yes, that's what they are called!)

layouts/partials/posts_navigation.html
```diff
@@ -1,15 +1,23 @@
...
+    {{ with .Params.prevInSection }}
```
craigbox, line 5: This would have us use the parameter `prevInSection` in the front matter, but we just want `prev`.)

layouts/partials/posts_navigation.html
```diff
@@ -1,15 +1,23 @@
...
+        {{$prev = .}}
+    {{ end }}
+    {{ with .Params.nextInSection }}
```
craigbox, line 8: and here

layouts/partials/posts_navigation.html
```diff
@@ -1,15 +1,23 @@
...
+        {{$next = .}}
+    {{ end }}
+    {{ if or $prev $next }}
```
craigbox, line 11: We've always had the pagenav display even if there's nothing in it.
More worrying is you didn't seem to close the `{{ if }}` block.

layouts/partials/posts_navigation.html
```diff
@@ -1,15 +1,23 @@
 {{ $skipPageNav := .Scratch.Get "skipPageNav" }}
 {{ if not (or .Params.skip_pagenav $skipPageNav) }}
-    {{ if or .NextInSection .PrevInSection }}
+    {{$prev := .PrevInSection}}
+    {{$next := .NextInSection}}
+    {{ with .Params.prevInSection }}
+        {{$prev = .}}
```
craigbox, line 6: `prevInSection` [returns a Page object](https://gohugo.io/methods/page/previnsection/).  
The front matter contains a text string pointing to a page.

You need to get the object for that page with a `{{- with $.Site.GetPage . -}}` block, and inside _that_ you can assign `.` to `$prev`.

Look at the Hextra example again if you need to.

AdeshGhadage:
@craigbox The difficulty I faced stemmed from not fully understanding Hugo's multilingual mode. Initially, I tried to provide the URL for a specific language version directly to Site.GetPage, but this didn't work because Hugo resolves paths based on the current language. After learning about multilingual support, I realized I needed to handle language-specific pages properly by including the correct language code in the path. Once I figured this out, I was able to resolve the issue and successfully implement the functionality. I enjoyed learning a lot while solving this challenge!

AdeshGhadage:
> LGTM; it's there in EN, it's not there in ZH, it does what we wanted.
> Thanks!
> 
> I'll approve this once the requested change is made. but I would like to encourage you to think about if there is a way to automate this: if you're on a page with children, then 'next' should be the first child, and the 'prev' should be the last page in the bundle from before.
> 
> Don't think too hard, but this demonstrates good pre-work for an upcoming mentorship :)

I currently out on vacation i will do the changes as soon as i get back

craigbox:
No rush, thanks.

craigbox (approved):
/cherry-pick release-1.24

Issue #16047, opened 2024-12-04
The OpenShift installation instructions don't work


wallrj:
The OpenShift installation instructions don't work
* https://istio.io/latest/docs/setup/platform-setup/openshift/

The suggested command returns an error:

```
[crc@crc-4-17 ~]$ istioctl install --set global.platform=openshift
        |\
        | \
        |  \
        |   \
      /||    \
     / ||     \
    /  ||      \
   /   ||       \
  /    ||        \
 /     ||         \
/______||__________\
____________________
  \__       _____/
     \_____/

Error: generate config: could not unmarshal: json: unknown field "global"
```

```
[crc@crc-4-17 ~]$ istioctl version
client version: 1.24.1
control plane version: 1.24.1
data plane version: 1.24.1 (1 proxies)
```

kfaseela:
cc @sridhargaddam

sridhargaddam:
@wallrj the documentation needs update. Will push a fix shortly.
In the meantime, please use the following command.

```
$: istioctl install --set profile=openshift
        |\          
        | \         
        |  \        
        |   \       
      /||    \      
     / ||     \     
    /  ||      \    
   /   ||       \   
  /    ||        \  
 /     ||         \ 
/______||__________\
____________________
  \__       _____/  
     \_____/        

This will install the Istio 1.24.0 profile "openshift" into the cluster. Proceed? (y/N) y
✔ Istio core installed ⛵️                                                                                                                                                                     
✔ Istiod installed 🧠                                                                                                                                                                         
✔ Ingress gateways installed 🛬                                                                                                                                                               
✔ CNI installed 🪢                                                                                                                                                                            
✔ Installation complete 
```

PR #16049, opened 2024-12-04
Fix istioctl install command for OpenShift

sridhargaddam:
Fixes: https://github.com/istio/istio.io/issues/16047

### DIFF
--- a/content/en/docs/setup/platform-setup/openshift/index.md
+++ b/content/en/docs/setup/platform-setup/openshift/index.md
@@ -16,7 +16,7 @@ Follow these instructions to prepare an OpenShift cluster for Istio.
 Install Istio using the OpenShift profile:
 
 {{< text bash >}}
-$ istioctl install --set global.platform=openshift
+$ istioctl install --set profile=openshift
 {{< /text >}}
 
 After installation is complete, expose an OpenShift route for the ingress gateway.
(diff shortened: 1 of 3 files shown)


kfaseela (approved):

bleggett (commented):
this isn't correct, `profile` and `platform` are not the same.

You can for instance have something like `--profile=ambient --platform=openshift`

bleggett:
@sridhargaddam can you correct? `platform` and `profile` are 2 different things, and you can have both.

`openshift` is a platform, not a profile.

content/en/docs/setup/platform-setup/openshift/index.md
```diff
@@ -16,7 +16,7 @@ Follow these instructions to prepare an OpenShift cluster for Istio.
 Install Istio using the OpenShift profile:
 
 {{< text bash >}}
-$ istioctl install --set global.platform=openshift
+$ istioctl install --set profile=openshift
```
bleggett, line 19: This should be `istioctl install --set values.global.platform=openshift` - using the Helm values.

istioctl-only profiles via `--set profile` are/should be deprecated, and `openshift` is now a `platform`, and not a `profile`.

This is so we can compose like so: `istioctl install --set values.global.platform=openshift  --set values.global.profile=ambient`

bleggett:
NVM, fixed in https://github.com/istio/istio.io/pull/16051

content/en/docs/setup/platform-setup/openshift/index.md
sridhargaddam: Okay Ben. We can continue with your PR - https://github.com/istio/istio.io/pull/16051/files

Issue #16053, opened 2024-12-05
Netlify won't build our site any more


craigbox:
as of ~18 hours ago all our Netlify deploys are failing. 

The errors are in one of the `babel --source-maps --minified --no-comments --presets minify` step of `gen_site.sh` but they don't seem consistent.

No changes to: node version, module version, any of the JS/TS files.

Some examples:

https://app.netlify.com/sites/preliminary-istio/deploys/67505e7ce5d55a00086ffb15

```
Error: /opt/build/repo/tmp/js/themes.js: Binding not found for ReferencedIdentifier "handleThemes" present in "CallExpression". Please report this at https://github.com/babel/minify/issues/new
2:53:49 AM:     at ReferencedIdentifier (/opt/build/repo/node_modules/babel-plugin-minify-mangle-names/lib/index.js:199:19)
2:53:49 AM:     at newFn (/opt/buildhome/.nvm/versions/node/v18.18.2/lib/node_modules/@babel/core/node_modules/@babel/traverse/lib/visitors.js:205:17)
2:53:49 AM:     at bfsTraverse (/opt/build/repo/node_modules/babel-plugin-minify-mangle-names/lib/bfs-traverse.js:38:43)
2:53:49 AM:     at Mangler.collect (/opt/build/repo/node_modules/babel-plugin-minify-mangle-names/lib/index.js:235:7)
2:53:49 AM:     at Mangler.run (/opt/build/repo/node_modules/babel-plugin-minify-mangle-names/lib/index.js:60:12)
2:53:49 AM:     at PluginPass.exit (/opt/build/repo/node_modules/babel-plugin-minify-mangle-names/lib/index.js:546:19)
2:53:49 AM:     at newFn (/opt/buildhome/.nvm/versions/node/v18.18.2/lib/node_modules/@babel/core/node_modules/@babel/traverse/lib/visitors.js:170:14)
2:53:49 AM:     at NodePath._call (/opt/buildhome/.nvm/versions/node/v18.18.2/lib/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:50:20)
2:53:49 AM:     at NodePath.call (/opt/buildhome/.nvm/versions/node/v18.18.2/lib/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:40:18)
[...]

PR #16054, opened 2024-12-05
test babel fix for netlify

craigbox:
Try holding `babel/traverse` back to a known good version while we wait for upstream fix.

https://github.com/istio/istio.io/issues/16053


craigbox:
yep, builds fine; that wasit

kfaseela (approved):

PR #16060, opened 2024-12-05
Do the thing we suggest instead of documenting the hard way

bleggett:
## Description

This repeatedly annoys me so fixing it.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

PR #16071, opened 2024-12-09
fix Helm test error introduced in #16060

craigbox:
#16060 removed a load-bearing snippet: the `doc.test.profile-none_istio.io` test, which obviously wasn't invoked in #16060 itself, relied on this snippet existing via the `_install_istio_helm` common function.


kfaseela (approved):

bleggett:
ugh, ty.

Issue #16011, opened 2024-11-26
bug: "contents" button to make docs TOC appear broken at <1200px


craigbox:
Up to 1200px, there is a floating "Contents" button which is missing its hamburger icon:

<img width="445" alt="Screenshot 2024-11-26 at 6 11 00 PM" src="https://github.com/user-attachments/assets/84b8032f-62b8-4fb6-8da6-81077edbd6c0">

When you click it to expand the contents, you can't make the pane go away again.

<img width="427" alt="Screenshot 2024-11-26 at 6 16 25 PM" src="https://github.com/user-attachments/assets/b2efe790-5127-43a7-ab72-bc9887dab647">

craigbox:
The problem seems to be related to SVG sprites and Chrome.  It works fine in Safari. I tried adding `href` to the SVG template as well as `xlink:href`, as is required by SVG 2.0, and I also tried setting a doctype of SVG 1.1.  Neither changed anything locally.

craigbox:
If we want to solve this properly, we have a similar problem with the close icon on the top level menu (works on Safari, not on Chrome):

<img width="439" alt="Screenshot 2024-12-12 at 1 59 12 PM" src="https://github.com/user-attachments/assets/9cffe752-b0ba-47eb-89b9-000534cbb6cc" />

<img width="454" alt="Screenshot 2024-12-12 at 1 59 28 PM" src="https://github.com/user-attachments/assets/0ed67152-b809-4897-9c89-28d6afef09ea" />

craigbox:
`hamburger-sidebar` did not render, but `menu-hamburger` did.

This is the code, as in the sprite sheet:

```
[...]

PR #16077, opened 2024-12-10
fix menu icon

Arhell:
## Description

resolve https://github.com/istio/istio.io/issues/16011

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience


Arhell (commented):
@craigbox

As it was written here https://github.com/istio/istio.io/issues/16011

I also tried different options but it didn't work.
Replaced it with a new svg and it worked, but I decided to install what already exists and works.

craigbox:
LGTM.  Though I think we should get to the bottom of this, because I've just seen that the close button doesn't appear on the main menu in Chrome either!

craigbox (approved):

PR #16087, opened 2024-12-12
Expand config profiles doc to mention helm + istioctl usage, and also mention platform profiles

bleggett:
## Description

- Explicitly mention profiles can be used by `helm` and `istioctl`
- Explain the difference between `platform profiles` and `regular profiles` - I have dubbed the latter `deployment profiles`, better names welcome.

There are still some UX gubbins here I do not like (mostly having to get into the weeds that even though they're directly shared with `istioctl`, with `istioctl` they also select specific components, which is not true in our helm impl ATM) - but better to document how it currently works than not.

## Reviewers

- [ ] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/additional-setup/config-profiles/index.md
+++ b/content/en/docs/setup/additional-setup/config-profiles/index.md
@@ -26,13 +26,13 @@ There are 2 kinds of configuration profiles: _deployment_ profiles and _platform
 - _deployment_ profiles are intended to provide good defaults for a given deployment topology (`default`, `remote`, `ambient`, etc).
 - _platform_ profiles are intended to provide necessary platform-specific defaults, for a given target platform (`eks`, `gke`, `openshift`, etc).
 
-For example, if you are installing `default` sidecar dataplane on `gke`, we recommend using the following deployment and platform profiles to get started.
+For example, if you are installing `default` sidecar dataplane on `gke`, we recommend using the following deployment and platform profiles to get started:
 
 {{< tabset category-name="install-method" >}}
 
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/additional-setup/config-profiles/index.md
```diff
@@ -10,12 +10,60 @@ test: n/a
...
+{{< tip >}}
+Configuration profiles compose with other values overrides or flags, so any individual value a configuration profile sets can be manually overridden by specifying a `--set` flag after it in the command.
+{{< /tip >}}
+
+There are 2 kinds of configuration profiles: _deployment_ profiles and _platform_ profiles, and using both is recommended.
+
+- _deployment_ profiles are intended to provide good defaults for a given deployment topology (`default`, `remote`, `ambient`, etc).
+- _platform_ profiles are intended to provide necessary platform-specific defaults, for a given target platform (`eks`, `gke`, `openshift`, etc).
+
+For example, if you are installing `default` sidecar dataplane on `gke`, we recommend using the following deployment and platform profiles to get started:
```
dhawton, line 29:
```suggestion
For example, if you are installing `default` sidecar dataplane on GKE, we recommend using the following deployment and platform profiles to get started:
```

dhawton (commented):
One minor nit since it's not being used as a value, it should be used as the name and capitalized.

dhawton (approved):

PR #16107, opened 2024-12-17
Remove Style Conflicts from Hamburger Menu SVGs

mike-weiner:
## Description

Fixes https://github.com/istio/istio.io/issues/16011. 

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


mike-weiner:
Hey, @craigbox and @Arhell!

After a couple hours of tearing my hair out, I'm pretty sure we finally have a fix for the root cause with this one. I don't think this is an issue of the SVG not displaying on Chrome. It **does** display, the color of the SVG was set to be the blue color and it is hard to make it out against the same blue background of the header bar. 

To prove this theory, I begin experimenting with the `icon` partial: https://github.com/istio/istio.io/blob/675a2183a7724020f71b3098f5a9ed2c7b91ac7e/layouts/partials/icon.html#L1

I tweaked it to instead be:

```
{{- $name := . -}}<svg style="fill: red; stroke: red;" class="icon {{ $name }}"><use xlink:href="/img/icons.svg#{{ $name }}"/></svg>
```

After this change, you will see the hamburger menu icon display on Safari, FireFox, and Chrome in an extremely obnoxious red color.

So, that led me down the path of some series of conflicting CSS styles. If you look at the code changes to these 2 SVGs you will see that I removed some styling from the SVG code. 

That `.cls-1` class was introducing the conflict.

mike-weiner:
From left to right: Safari, FireFox, Chrome

Screenshot 2024-12-16 at 22 43 20

craigbox:
Nice.  I had tried to override those values in the inspector but had not had any success.  
If you reinstate the file removed in #16077. does that work?

mike-weiner:
@craigbox - Sure did. I've reverted the changes from https://github.com/istio/istio.io/pull/16077 in this PR while making the same tweak to the SVG to fix the problem. 

Screenshot 2024-12-17 at 05 53 24

Arhell (commented):
@mike-weiner great work
Thanks

craigbox (approved):

PR #16169, opened 2025-01-14
[zh] Sync #16044 set prev and next for Ambient get start doc into Chinese

wilsonwu:
## Description

Sync #16044 set prev and next for Ambient get start doc into Chinese

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/zh/docs/ambient/getting-started/_index.md
+++ b/content/zh/docs/ambient/getting-started/_index.md
@@ -6,8 +6,9 @@ aliases:
   - /zh/docs/ops/ambient/getting-started
   - /zh/latest/docs/ops/ambient/getting-started
 owner: istio/wg-networking-maintainers
-skip_list: true
 test: yes
+skip_list: true
+next: /zh/docs/ambient/getting-started/deploy-sample-app
 ---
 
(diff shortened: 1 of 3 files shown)


AdeshGhadage (commented):
@wilsonwu I think you should remove "/zh" from prev and next url then it will. i got same issue in english as well.

wilsonwu:
> @wilsonwu I think you should remove "/zh" from prev and next url then it will work. i got same issue in english as well.

Not sure it will go to correct Chinese next or prev page by remove the "/zh"?

AdeshGhadage:
> 
> Not sure it will go to correct Chinese next or prev page by remove the "/zh"?

it will direct to correct chinese next or prev i have checked it locally also

wilsonwu:
> @wilsonwu I think you should remove "/zh" from prev and next url then it will work. i got same issue in english as well.

Done, thank you so much for your help and test.

windsonsea (approved):

PR #16178, opened 2025-01-16
Fix a couple of typos in Ambient docs

antoninbas:
## Description

* iptables rules for ztunnel ingress processing match on the destination port, and not the source port as was previously described.
* the NetworkPolicy example was incorrect in the document describing how to update policies to accomodate HBONE traffic (policy was missing a name and app port number was not the same in both versions).

## Reviewers

- [x] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


howardjohn (approved):

howardjohn:
thanks!

/ok-to-test

dhawton (approved):

PR #16185, opened 2025-01-20
remove unnecessary semicolon which breaks zsh on mac

### LINKED ISSUE #16162: Fix our install doc on gw API
Ran into this today while following the [doc](https://istio.io/latest/docs/ambient/install/istioctl/#install-or-upgrade-the-kubernetes-gateway-api-crds): 

 ~/Downloads/ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
  { kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.0/standard-install.yaml\; }

error: unable to read URL "https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.0/standard-install.yaml;", server reported 404 Not Found, status code=404

Fixed by this:

 ~/Downloads/ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
  { kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.0/standard-install.yaml }

customresourcedefinition.apiextensions.k8s.io/gatewayclasses.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/gateways.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/grpcroutes.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/httproutes.gateway.networking.k8s.io created
customresourcedefinition.apiextensions.k8s.io/referencegrants.gateway.networking.k8s.io created

craigbox:
zsh, in many default configurations, escapes special characters on paste.  The `;` in the Gateway API example is replaced with `\;` which causes the command to fail.

As it's not a load-bearing semi-colon, lets just remove it.

Fixes #16162

### DIFF
--- a/content/en/boilerplates/gateway-api-install-crds.md
+++ b/content/en/boilerplates/gateway-api-install-crds.md
@@ -5,5 +5,5 @@ installed before using the Gateway API:
 
 {{< text syntax=bash snip_id=install_crds >}}
 $ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-  { kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/{{< k8s_gateway_api_version >}}/standard-install.yaml }
+  kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/{{< k8s_gateway_api_version >}}/standard-install.yaml
 {{< /text >}}
--- a/content/en/boilerplates/gateway-api-install-crds.md
+++ b/content/en/boilerplates/gateway-api-install-crds.md
@@ -5,5 +5,5 @@ installed before using the Gateway API:
 
 {{< text syntax=bash snip_id=install_crds >}}
 $ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-  { kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/{{< k8s_gateway_api_version >}}/standard-install.yaml; }
+  { kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/{{< k8s_gateway_api_version >}}/standard-install.yaml }
 {{< /text >}}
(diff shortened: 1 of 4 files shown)


craigbox:
you're kidding me.  [it _is_ a load bearing semi-colon?](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/16185/lint_istio.io/1881260491049799680)

kfaseela (approved):

kfaseela:
saw the lint failure only after approving :D

craigbox:
note to self, make Faseela an approver for Ukranian too

bleggett:
> you're kidding me. [it _is_ a load bearing semi-colon?](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/16185/lint_istio.io/1881260491049799680)

It's required for bash, we are using `shellcheck` in `bash` syntax mode, and the snippet is `bash` syntax.

The actual way to handle stuff like this is have a `syntax=sh` (POSIX sh subset only) and have shellcheck validate POSIX sh rules versus defaulting to `bash` syntax (which lots of shells emulate, but not exactly).

POSIX sh is the actual "common" subset to target if you want stuff that runs in any shell.

craigbox:
I don't think it's "required for bash" in the sense that things work if you remove it, but only because the only remaining command is a `}`, which is treated as a literal.

I also think our shellcheck scripts aren't smart enough to distinguish between POSIX sh and bash at this point?

Should we go for the trivial fix of changing the command?

```kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.0/standard-install.yaml```

...seems like it will do what we want?

**Comment by ilrudie** [comment]
yeah, It seems to me that something like this should work
```
{{< text syntax=bash snip_id=install_crds >}}
$ kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/{{< k8s_gateway_api_version >}}/standard-install.yaml
{{< /text >}}
```

bleggett:
> I also think our shellcheck scripts aren't smart enough to distinguish between POSIX sh and bash at this point?

Fair enough - shellcheck definitely can distinguish, but our stuff may not be able to.

Issue #16199, opened 2025-01-23
Update footer copyright year and Implement automation


AdeshGhadage:
![Image](https://github.com/user-attachments/assets/1c2a54ba-ca03-442f-b9b3-bd676fe760b1)

AdeshGhadage:
Yes

kfaseela:
> Yes

Please go ahead

PR #3071, opened 2019-01-10
Switch to discuss.istio.io

geeknoid:

PR #16202, opened 2025-01-24
automated footer copyright year

AdeshGhadage:
## Description

Implemented dynamically update footer copyright year for issue #16199.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/data/args.yml
+++ b/data/args.yml
@@ -7,9 +7,6 @@ full_version: "1.25.0"
 # The previous Istio version identifier the docs describe, used for upgrade documentation
 previous_version: "1.24"
 
-# The year to display in copyright notices
-copyright_year: 2025
-
 # Controls the incarnation of the site
 #
 # when preliminary=false and archive=false, we're building for istio.io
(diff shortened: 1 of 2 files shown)


Arhell (commented):
/ok-to-test

data/args.yml
```diff
@@ -7,9 +7,6 @@ full_version: "1.25.0"
 # The previous Istio version identifier the docs describe, used for upgrade documentation
 previous_version: "1.24"
 
-# The year to display in copyright notices
-copyright_year: 2025
```
craigbox, line 11: Can you check the git blame to see why this was added, confirm it was only used in that one place?

It seems like an obvious thing to use "now", which makes it feel like there's a reason - probably so the archive sites have a copyright date of when they were released, and not always "now".

AdeshGhadage: The copyright_year field was first added in PR #3071 . It was originally introduced for the copyright footer and has only been updated since then.

PR #16261, opened 2025-02-21
prepare for v1.25 as istio source is already branched

kfaseela:
## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


Makefile.core.mk
```diff
@@ -77,7 +77,7 @@ baseurl := "$(URL)"
 endif
 
 # Which branch of the Istio source code do we fetch stuff from
-export SOURCE_BRANCH_NAME ?= master
+export SOURCE_BRANCH_NAME ?= release-1.25
```
bleggett, line 80: Shouldn't we leave `master` refs as-is? Typically we do, right?

Or is this a temporary chore we need to do, and then afterwards flip it back from `release-1.25` to `master`?

kfaseela: Yes, I think we move this back later

kfaseela:
@dhawton when should this PR be merged?

Makefile.core.mk
bleggett: Oh yep: https://github.com/istio/istio.io/pull/15848/files

bleggett (approved):

dhawton (approved):

PR #16299, opened 2025-03-05
Fix incorrectly-flipped flag in 1.25 relnotes

bleggett:
## Description

this should be `true` not `false`, it's correct in the upgrade notes.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


craigbox (approved):
false that, double false

PR #16303, opened 2025-03-06
Migrated Ruby linter to NodeJS linter.

### LINKED ISSUE #7906: Istio pages don't abide by the requirements of our linter
We use markdown linter at different levels in tools and istio.io. The same docs break in tools that would normally pass istio.io. For example, the release notes break because there's no level 1 headers in the istio docs.

Ajay-singh1:
## Description

This PR replaces the Ruby linter to a modern NodeJS linter which is more flexible and advanced.

The linter passes all the docs test successfully and also it resolves #7906.

This PR fixes issue No. #15148 

Attached Screenshot showing linter in action.
Screenshot_20250306-130055_Chrome

## Reviewers

- [ ] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


craigbox:
/ok-to-test

Your output shows linting of 0 files?

Ajay-singh1:
@craigbox Thanks for reviewing.Yes their was a minor mistake I have fixed it now the linter is working fine.
Screenshot from 2025-03-06 17-33-33

scripts/lint_site.sh
```diff
@@ -82,8 +82,12 @@ check_content() {
         error "To learn how to address spelling errors, please see https://istio.io/about/contribute/build/#test-your-changes"
         FAILED=1
     fi
-
-    if ! mdl --ignore-front-matter --style mdl.rb .; then
+    
+    # Updated to use markdownlint-cli2
+    if ! command -v markdownlint-cli2 >/dev/null 2>&1 ; then
+        npm install -g markdownlint-cli2
```
dhawton, line 88: I feel like we should make this be part of the build-tools image so we're not constantly running this. I'd also prefer to see something non-node, honestly.

Ajay-singh1: Yeah! I think a Go based linter would be much better.What do you say?

Ajay-singh1:
@craigbox I can see MD007 indentation errors in 21 files , what should be the right approach should I go on and fix those files or should I just disable the MD007 rule so that the lint passes successfully after this step I will open a PR in istio/tools and update the Dockerfile so that the new linter is installed automatically when someone spins up a Docker container.I will update the outdated node utilities too.

dhawton:
> @craigbox I can see MD007 indentation errors in 21 files , what should be the right approach should I go on and fix those files or should I just disable the MD007 rule so that the lint passes successfully after this step I will open a PR in istio/tools and update the Dockerfile so that the new linter is installed automatically when someone spins up a Docker container.I will update the outdated node utilities too.

The most correct way is to install it into the container rather than have it install automatically when the container is started.

craigbox:
I suggest you use this PR and the manual installation method to check
everything works and propose all the changes we need to the markup or
rules.  Then, we build it into the proper container, and rebase the PR on
the updated files when they land.

We can talk you through how this will work.

Ajay-singh1:
Hey @craigbox I have tested this PR through manual installation and everything is working fine.I dont want to change the rules so I have updated the markup of all the .md files where the linter was throwing the error and now the linter checks are successful.What should be the next steps?.Yes it would be very helpful if you guide through this process.

dhawton:
> Hey @craigbox I have tested this PR through manual installation and everything is working fine.I dont want to change the rules so I have updated the markup of all the .md files where the linter was throwing the error and now the linter checks are successful.What should be the next steps?.Yes it would be very helpful if you guide through this process.

As he mentioned:
- Commit them to the PR
- Update the build-tools image in the `tools` repo to install the tool globally
- Once that merges, give about an hour for the build-tools image to be built and begin being applied across the repos
- Trigger a retest here which will use the new build-tools image

Ajay-singh1:
@dhawton Everything is working fine here are the attached screenshots:-
Screenshot from 2025-03-11 20-14-50
Screenshot from 2025-03-11 20-14-25
Screenshot from 2025-03-11 20-19-26

Ajay-singh1:
This PR is cluttered up closing this.

PR #16316, opened 2025-03-12
Migrated the Ruby Linter to NodeJS Linter

Ajay-singh1:
## Description

This PR replaces the Ruby linter to a modern NodeJS linter which is more flexible and advanced.

The linter passes all the docs test successfully and also it resolves https://github.com/istio/istio.io/issues/7906.

This PR fixes issue No. https://github.com/istio/istio.io/issues/15148

## Reviewers

- [ ] Ambient
- [x] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


Ajay-singh1:
@craigbox The previous PR got cluttered up so I had to create a new one.Regarding the MD007 rule the ruby based linter doesnt strictly enforce the MD007 rule  and allows mixed indentation while the node based  linter enforces strict MD007 rule so a much better solution will be to disable the MD007 rule.

dhawton:
> @craigbox The previous PR got cluttered up so I had to create a new one.Regarding the MD007 rule the ruby based linter doesnt strictly enforce the MD007 rule and allows mixed indentation while the node based linter enforces strict MD007 rule so a much better solution will be to disable the MD007 rule.Once the check passes here I will update the common files so that the change is applied to all istio repos.The linting is sucessful now:-

I actually strongly disagree with disabling MD007. This should be something we enforced, it's been something I have been fixing in PRs that I review, it'd be much better to enforce and resolve the issues.

Ajay-singh1:
@dhawton I need a couple of days to come up with a solution.Will be right back.

Arhell (commented):
/ok-to-test

Ajay-singh1:
@dhawton markdownlint-cli2 is not being recognized.

Ajay-singh1:
Complex Issue unable to resolve closing this.

craigbox:
@Ajay-singh1 are you intending to pick this up elsewhere?  I think you have everything you need to proceed, and I can help steward it.  Otherwise I can move it onto another maintainer.

Ajay-singh1:
@craigbox Yes , this will work and will pass all the lint checks.I am waiting for the PR to be merged in istio/tools once it is approved the checks will pass here and we are good to go.

craigbox:
I don't see a PR.  Can you link one if it exists?  Else, do you want to add it to https://github.com/istio/tools/pull/3161 ?
(I would also add both for the meantime)

Ajay-singh1:
Yes this specific PR I was talking about istio/tools#3161.It is in pending state once it merges the lint checks will pass here I am completely confident about it.

craigbox:
You may have to wait a day or so for the automator to build a new container and update this repo to use it.

Ajay-singh1:
👍

craigbox:
Try pushing an empty commit. it will probably retest on an existing image
otherwise

craigbox:
Nice!  The build is working.

The gencheck error is because you added a `<!-- markdownlint-disable -->` block to a code snippet, but not to the generated `snips.sh` file.  However, I would suggest we shouldn't have an ignore block anywhere but the front matter of a document.

I would also like to standardize on using a `markdownlint-disable: XXX` block in the doc front matter.  Can we see if this is natively supported or how we might need to introduce it?

content/en/docs/ops/configuration/security/security-policy-examples/snips.sh
```diff
@@ -81,6 +81,7 @@ spec:
   mtls:
     mode: STRICT
 ---
+<!-- markdownlint-disable-file MD007 -->
```
craigbox, line 84: This content is now in the YAML files that are used for testing, which will almost certainly cause all the tests to fail.

Ajay-singh1: I agree with you I will fix this within a day.Yes markdownlint-disable works for doc front matter it is natively supported.

craigbox: Yeah, that's confirmed here: https://github.com/DavidAnson/markdownlint/issues/1250

unfortunate.

.markdownlint.json
craigbox, line 1: Why is this file both here and in `common/config`?  Can we have just the one and have it be copied?

.markdownlint.json
```diff
@@ -0,0 +1,13 @@
+{
+    "default": false,
+    "MD002": { "level": 2 },
+    "MD007": { "indent": 4 },
```
craigbox, line 4: This rule is disabled in many files.   Can you look into why they are failing, and if we can just change the markup to pass correctly?  Did we not have this rule in the old config?

.markdownlint.json
```diff
@@ -0,0 +1,13 @@
+{
+    "default": false,
```
craigbox, line 2: Does this turn off the default rules?  Wouldn't we want them on?

content/en/blog/2017/0.1-canary/index.md
```diff
@@ -114,6 +114,8 @@ spec:
         subset: v2
       weight: 10
 ---
+<!-- markdownlint-disable-file MD007 MD026 -->
+<!-- markdownlint-disable-file MD007 -->
```
craigbox, line 118: The first line should render the second line unnecessary

craigbox: These comments. if they're truly needed, should only exist at the top of a file (directly under the front matter).  It looks like you might have done something like taking all the failing files and tried to add a comment under the second instance of `---`  ?

.markdownlint.json
Ajay-singh1: Agreed!

.markdownlint.json
Ajay-singh1: Yes this will turn off the default rules , no turning them on is not a good way because all the rules will be applied by default which is used by the linter and we dont want them right?We will explicitly define the rules which we want to use to lint our markdown files.

content/en/blog/2017/0.1-canary/index.md
Ajay-singh1: Agreed! I am also of the same view actually this is the result of a bash script , it automatically added these lines in the files.

.markdownlint.json
Ajay-singh1: We did have this rule in the old configuration.It is failing because markdownlint-cli2 enforces strict MD007 rules while the previous linter allows for mixed indentation for the MD007 rule.Changing markup will have several breaking changes.

Ajay-singh1:
I have addressed all the issues , We can cleanly merge this PR , If you have any further questions do let me know.Thanks @dhawton and @craigbox for your assistance.

Ajay-singh1:
@craigbox Can I get a thumbs up here?

PR #16352, opened 2025-03-25
Fix image captions using Hugo figure shortcode

sbegin0:
## Description

- Replaced  tags with Hugo's {{< figure >}} shortcode so that img captions are accurately shown.

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/ops/deployment/performance-and-scalability/index.md
```diff
@@ -111,18 +111,10 @@ We obtained these results using the [Istio benchmarks](https://github.com/istio/
...
-    caption="P90 latency vs client connections"
-/>
-<br><br>
-<img width="90%" style="display: block; margin: auto;"
-    src="istio-1.24.0-fortio-99.png"
-    alt="P99 latency vs client connections"
-    caption="P99 latency vs client connections"
-/>
-<br>
+{{< figure src="istio-1.24.0-fortio-90.png" caption="P90 latency vs client connections" width="90%" >}}
```
craigbox, line 114: We have a custom `{{< image >}}` shortcode.  We should use that, and then later on try and figure out if we need to maintain it or can just move to using the Hugo provided variant.

 ```suggestion
{{< image link="./istio-1.24.0-fortio-90.png" caption="P90 latency vs client connections" width="90%" >}}
```

(The `./` is probably not needed but many other examples use it)

File: content/en/docs/ops/deployment/performance-and-scalability/index.md [hunk 2]
```diff
@@ -111,18 +111,10 @@ We obtained these results using the [Istio benchmarks](https://github.com/istio/
...
+
+{{< figure src="istio-1.24.0-fortio-99.png" caption="P99 latency vs client connections" width="90%" >}}
+
```
craigbox, line 117: Lint is failing here because there are two blank lines in a row.

craigbox (changes requested):

PR #16394, opened 2025-04-08
Add DocSearch v3 Integration

AdeshGhadage:
## Description

integrated algolia for site search

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [x] User Experience

### DIFF
--- a/assets/sass/misc/_header.scss
+++ b/assets/sass/misc/_header.scss
@@ -112,11 +112,8 @@
         margin-left: 1rem;
     }
 
-    #search-show {
-        display: none;
-        @media (min-width: $bp-lg) {
-            display: inline-block;
-        }
+    #docsearch button {
+        background-color: var(--docsearch-searchbox-background);
     }
 
     #search-form {
@@ -493,9 +490,19 @@
     }
 }
 
-@media (min-width: $bp-xl) {
+@media (min-width: $bp-lg) {
     .main-navigation-footer {
-        #try-istio { margin-left: 2.75rem; }
+        display: flex;
+        flex-direction: row;
+        align-items: center;
+    }
+
+    .main-navigation button {
+        margin: 0;
+    }
+
+    .main-navigation-links {
+        display: flex;
     }
 }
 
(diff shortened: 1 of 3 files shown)


Arhell (commented):
/ok-to-test

craigbox:
You're about 10 steps ahead here, @AdeshGhadage!

1. results are not yet scoped to versions

Screenshot 2025-04-09 at 4 33 46 PM

2. The default UI looks out of place on the page; we should retain the single magnifying glass in the first instance.

data/versions.yml
```diff
@@ -2,4 +2,35 @@
 main: "1.25"
 
 # The version of Istio currently documented in preliminary.istio.io
-preliminary: "1.26"
+preliminary: "1.27"
```
craigbox, line 5: You should probably rebase this PR against the current master.  Let me know if you need help doing that.

data/versions.yml
```diff
@@ -2,4 +2,35 @@
...
+
+supported_versions:
+  - latest
+  - v1.25
+  - v1.24
+  - v1.23
+  - v1.22
+  - v1.21
+  - v1.20
```
craigbox, line 14: (a) we shouldn't need to list all these
(b) we only want to search N-5; so for 1.26, that would be 1.21 and up

layouts/_default/search.html
```diff
@@ -3,57 +3,189 @@
...
+        if (!results[2]) return '';
         return decodeURIComponent(results[2].replace(/\+/g, " "));
     }
-    const q = getParameterByName('q', window.location.href);
-    document.getElementsByName('q')[0].value = q;
+
+    function extractVersionAndLangFromPath() {
+        const pathParts = window.location.pathname.split('/');
+        let version = 'latest';
+        let lang = 'en';
```
craigbox, line 54: Did you write this code, or is it based on something from Algolia?
Should it be tested?  Should it be in its own JS/TS file?

layouts/_default/search.html
```diff
@@ -3,57 +3,189 @@
...
+    document.getElementById('version-filter').addEventListener('change', runSearch);
+    document.getElementById('lang-filter').addEventListener('change', runSearch);
+
+    runSearch();
 </script>
 
-{{ .Content }}
+<style>
+    .search-results {
+        padding: 1rem;
```
craigbox, line 153: note, will need to be in its own SCSS file soon enough

PR #16447, opened 2025-04-29
re-enable multicluster tests

dhawton:
Depended on https://github.com/istio/istio/pull/56120. That should fix the issues docs tests had, so let's test it and re-enable multicluster.


craigbox:
Great work!

craigbox (approved):

PR #16456, opened 2025-05-07
Azure Load Balancer notes

Stevenjin8:
## Description

Band aid for https://github.com/istio/istio/issues/46769

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -910,6 +910,7 @@ NLBs
 no-brainer
 Node.js
 nodeAffinity
+non-200
 non-conformant
 non-L4
 non-mTLS
--- a/content/en/docs/setup/platform-setup/azure/index.md
+++ b/content/en/docs/setup/platform-setup/azure/index.md
@@ -72,7 +72,7 @@ infrastructure:
 
 where `<http[s] port>` is the port number of your HTTP(S) listener.
 If you have multiple HTTP(S) listeners, you need to add an annotation for each listener.
-This annotation is required for Azure Load Balancer health checks to work when the `/` path responds with a non 200-code.
+This annotation is required for Azure Load Balancer health checks to work when the `/` path does not respond with a 200.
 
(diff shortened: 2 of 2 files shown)


content/en/docs/ambient/getting-started/deploy-sample-app/index.md
```diff
@@ -61,6 +61,14 @@ bookinfo-gateway   istio   bookinfo-gateway-istio.default.svc.cluster.local   Tr
 
 Wait for the gateway to show as programmed before continuing.
 
+{{< tip >}}
```
craigbox, line 64: We try very hard to keep the Getting Started guides to be simple, and not full of "if this then that'.  Is there any other way we can communicate this to keep it out of the "happy path"?

Stevenjin8: That's fair. I was going to justify it by saying "we do similar stuff for GKE", but I think I hallucinated that. We definitely want to add something on in the Azure specific section

Stevenjin8: aha! I found it: https://istio.io/latest/docs/examples/bookinfo/#start-the-application-services. Though I agree that this is not be a good precendent to follow

craigbox: The same requirement is probably true of AKS, which probably didn't exist when the page was written 🤣

keithmattix:
What more is left to do on this PR? Seems like something that would be pretty helpful to users

dhawton:
Leaving for @craigbox to review as he had some comments. Dropped the CP release-1.27 label as in docs that branch won't exist until .0 is about to publish, it's created from master at that point.

craigbox (approved):
LGTM, scope has changed to only have this on the Azure page and not in the general Bookinfo guide.

(All those pages need review and updating in some nebulous future!)

PR #16478, opened 2025-05-12
Issue 16404 - Documentation version selector dropdown 

sbegin0:
## Description

This PR tests a version selector dropdown under "Documentation" and includes previous logic for implementing a netlify edge function. 

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


data/versions.yml
```diff
@@ -1,5 +1,26 @@
-# The version of Istio currently documented in istio.io
 main: "1.26"
 
-# The version of Istio currently documented in preliminary.istio.io
 preliminary: "1.27"
+
+# List of all versions to be included in the dropdown selector and for archive logic
+versions:
```
craigbox, line 6: I think you can infer all of this programatically.

We know what the version is, we can subtract N from it, and when the site is built, it's built with a flag showing if content is archive, preliminary or current.  (At least that used to be true; we can just infer anything that is not latest is archive.)

kfaseela:
Looking at the screenshot shared in slack, why is v1.27 listed above v1.26? Do we switch to preliminary.istio.io when we select v1.27?

sbegin0:
> Looking at the screenshot shared in slack, why is v1.27 listed above v1.26? Do we switch to preliminary.istio.io when we select v1.27?

I was debating adding preliminary versions to the dropdown but decided against it, new deploy preview should show the updated menu that starts with latest (1.26)

craigbox:
This should definitely show preliminary _when running on preliminary_, but it would be OK to not show it when running on latest or older.

(It should also highlight the current version if it's an archive site you're looking at)

Issue #16494, opened 2025-05-15
What works in Ambient?


Stevenjin8:
Naturally, the answer is "it depends", but we should do a better job explaining this in the docs/integrating ambient into our docs in general. [Authn policy is a great example](https://istio.io/latest/docs/tasks/security/authentication/authn-policy/).

It makes no mention of Ambient mode, so its not clear what when policy applies. And the answer is a little complicated since L4 auth policy works if ambient is enabled, while L7 policy only applies when there is a waypoint. So its more complicated than a sidecar vs ambient question. (There are also other factors too, such as whether DNS proxying is enabled).

Another problem is that a lot of the code is sidecar specific:

```bash
$ kubectl create ns foo
$ kubectl apply -f <(istioctl kube-inject -f [samples/httpbin/httpbin.yaml](https://raw.githubusercontent.com/istio/istio/release-1.26/samples/httpbin/httpbin.yaml)) -n foo
$ kubectl apply -f <(istioctl kube-inject -f [samples/curl/curl.yaml](https://raw.githubusercontent.com/istio/istio/release-1.26/samples/curl/curl.yaml)) -n foo
```

What I really want to avoid is having a bunch of banners in saying "ambient mode this, ambient mode that" in our tutorials. Users should be able to copy and paste instructions line by line and have things work.

So at a high level, this is my proposal. For tutorials (i.e. tasks), add a tab to code blocks for sidecar vs ambient mode, much like how we have our own ingress APIs vs Kubernetes' Gateway APIs. This way, beginners can scroll down the page and copy-paste code blocks without having to even know what ambient is.

Reference docs are a little trickier, since they are more specific. @therealmitchconnors suggested going field by field and saying if/when it is supported, but that seems extremely tedious. Another approach we could take is going through [the feature page](https://istio.io/latest/docs/releases/feature-stages/) and adding an ambient/ambient+sidecar column, but that feels too general.

cc @therealmitchconnors @grnmeira @craigbox

Stevenjin8:
The dataplane comparison page might be a good starting point https://istio.io/latest/docs/overview/dataplane-modes/

grnmeira:
Regarding the feature page, it does feel a bit general, but I think we need that, independently of what we decide to do with the reference docs. It's a good place for understanding the maturity of certain features in different data-plane modes. I'd suggest actually referencing that in [Overview Data-plane modes](https://istio.io/latest/docs/overview/dataplane-modes/), and unifying both if people think that makes sense.

[...]

PR #16500, opened 2025-05-19
Docs for bookinfo-ambient

Stevenjin8:
## Description

Part of #16494. A lot of tasks/tutorials say "set up bookinfo", so this might be  a good place to start and get early feedback, as later PRs will probably look like this.

For the bookinfo-waypoint.yaml file, see https://github.com/istio/istio/pull/56357

## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/examples/bookinfo/index.md
+++ b/content/en/docs/examples/bookinfo/index.md
@@ -57,8 +57,12 @@ in the [installation guide](/docs/setup/).
 
 To run the sample with Istio requires no changes to the
 application itself. Instead, you simply need to configure and run the services in an
-Istio-enabled environment, with Envoy sidecars injected along side each service.
-The resulting deployment will look like this:
+Istio-enabled environment.
+{{< tabset category-name="dataplane-mode" >}}
+
+{{< tab name="Sidecar Mode" category-value="sidecar-mode" >}}
+
+In Sidecar mode, Envoy sidecars are injected along side each workload, and the resulting deployment will look like this:
 
 {{< image width="80%" link="./withistio.svg" caption="Bookinfo Application" >}}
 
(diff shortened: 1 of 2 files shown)


content/en/docs/examples/bookinfo/index.md
```diff
@@ -67,29 +71,88 @@ and outgoing calls for the services, providing the hooks needed to externally co
...
+
+{{< /tab >}}
+
+{{< /tabset >}}
+
 ### Start the application services
 
 {{< tip >}}
-If you use GKE, please ensure your cluster has at least 4 standard GKE nodes. If you use Minikube, please ensure you have at least 4GB RAM.
+If you using a cloud provider, please ensure your cluster has at least 4 nodes, each with 4GB of memory and 4 CPUS.
```
Stevenjin8, line 97:
```suggestion
If you using a cloud provider, please ensure your cluster has at least 4 nodes, each with at least 4GB of memory and 4 CPUS.
```

Stevenjin8:
TODO: diagram with ambient

content/en/docs/examples/bookinfo/index.md
craigbox: "CPUs".

TBH this is really high for an app that does ~nothing.  Is this still a relevant recommendation, and is it to do with reservations?

craigbox:
https://deploy-preview-16500--preliminary-istio.netlify.app/latest/docs/examples/bookinfo/

I like the idea of  "Sidecar mode' and "Ambient mode" tab.
I feel we start getting confused when we later on have an "Istio Gateway" and "Gateway API" tab.  Which do I use based on which mode?

Is it possible to land the ambient mode changes without having a (three?) way matrix about "Sidecar/Ingress", "Sidecar/Gateway", "Ambient/Gateway" — not to mention if we also support VirtualService in the docs.

content/en/docs/examples/bookinfo/index.md
Stevenjin8: Honestly, I just threw a number out there that I thought would be around that of a "Standard GKE Node". Open to suggestions

Stevenjin8:
> https://deploy-preview-16500--preliminary-istio.netlify.app/latest/docs/examples/bookinfo/
> 
> I like the idea of "Sidecar mode' and "Ambient mode" tab. I feel we start getting confused when we later on have an "Istio Gateway" and "Gateway API" tab. Which do I use based on which mode?
> 
> Is it possible to land the ambient mode changes without having a (three?) way matrix about "Sidecar/Ingress", "Sidecar/Gateway", "Ambient/Gateway" — not to mention if we also support VirtualService in the docs.

I'm not sure I understand. There are four possible views of the page: {sidecar,ambient} x { ingress, gateway}. The dataplane mode tab group is independent of the gateway/ingress api tab group.

I do think its fair to say that this introduces quite a bit of complexity to a getting started page, but I think its a good compromise. 

* From the POV of a first time user, the can scroll down the page and not worry about dataplane modes or gateway/ingress apis.
* From the POV of a user starting out from ambient, the docs are one click away from what they are looking for, and they can just copy paste
* Same for ingress/gateway api

The page contains all the information to get started with bookinfo with various combinations of our API without overwhelming first time users.

At some point, we have multiple APIs that do the same thing, and unless we want do omit some APIs from parts of the docs, there will be some complexity in the docs. (But maybe making some omissions isn't too bad).

Stevenjin8:
@craigbox I think I see what you mean now because we don't support istio ingress apis with ambient. @keithmattix suggested that we get rid of istio ingress apis all together in getting started docs. However, we would still have this issue in later pages such as https://istio.io/latest/docs/tasks/traffic-management/traffic-shifting/ because we do support istio apis and gateway apis for

Stevenjin8:
It seems like we already have a whole getting started flow with istio ambient

* https://istio.io/latest/docs/ambient/getting-started/
* https://istio.io/latest/docs/ambient/getting-started/deploy-sample-app/
* https://istio.io/latest/docs/ambient/getting-started/secure-and-visualize/
* https://istio.io/latest/docs/ambient/getting-started/enforce-auth-policies/

Might have to rethink how we're approaching this...

keithmattix:
> Likewise, if there was a point we could make the Gateway API the default for sidecar mode configuration we would be in a much better place

I think we're approaching the point where we need to make this decision. It'll simplify things for new users and improve our documentation

Stevenjin8:
That's fair. So stepping back the question that started all this is what works in ambient, and going through all the examples and having them work with both dataplane modes is too ambitious (or more ambitious than I would like to be). Maybe we are doing better than I think we are with the dataplane comparison page

craigbox:
[RFC for improvement](https://docs.google.com/document/d/1JjsPzOMJfu_evzgiRp-aJf1KDpa9ImyYrew__y5CTao/edit?tab=t.0)

PR #16513, opened 2025-05-23
fix(refdocs): convert URLs to markdown

### LINKED ISSUE #16102: remove URLs from proto description
https://istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-HTTPSettings

> Maximum number of requests that will be queued while waiting for a ready connection pool connection. Default 2^32-1. Refer to https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking under which conditions a new connection is created for HTTP2. Please note that this is applicable to both HTTP/1.1 and HTTP2.

Change these URLs to markdown throughout the refdocs.  This will need to be done on the `istio/api` repo (in `master` and `release-1.24`) but as there is not currently any automation to move the refdocs here, it can be done here simultaneously.

(While you're there, standard use of "HTTP/2" instead of "HTTP2".)

AritraDey-Dev:
## Description

this fixes #16102 

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/reference/config/networking/destination-rule/index.html
+++ b/content/en/docs/reference/config/networking/destination-rule/index.html
@@ -916,8 +916,7 @@ the speed of traffic ramp-up increases non-linearly.</p>
 <h2 id="ConnectionPoolSettings">ConnectionPoolSettings</h2>
 <section>
 <p>Connection pool settings for an upstream host. The settings apply to
-each individual host in the upstream service.  See Envoy&rsquo;s <a href="https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking">circuit
-breaker</a>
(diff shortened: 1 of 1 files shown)


dhawton (changes requested):
Please see the second line of the file:

```WARNING: THIS IS AN AUTO-GENERATED FILE, DO NOT EDIT. PLEASE MODIFY THE ORIGINAL SOURCE IN THE 'https://github.com/istio/api' REPO```

AritraDey-Dev:
Thanks @dhawton for clarifying this.Should I close this as the issue seems to be solved in the upstream repo?

PR #16515, opened 2025-05-23
remove use of istioctl x wait

### LINKED ISSUE #16429: remove use of `istioctl x wait`
In `tests/util/helpers.sh` we're using `istioctl experimental wait` which was removed https://github.com/istio/istio/pull/51866. We need to remove its use in our doc testing, ideally with another method of checking whether a config was applied before continuing on. I think this might be why tests have become more flakey because without this command, we're starting tests for config has been fully applied and distributed.

AritraDey-Dev:
## Description
fixes #16429 

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/tests/util/helpers.sh
+++ b/tests/util/helpers.sh
@@ -108,19 +108,31 @@ _wait_for_statefulset() {
 # Wait for Istio config to propagate
 # usage: _wait_for_istio <kind> <namespace> <name>
 _wait_for_istio() {
-    #local kind="$1"
-    #local namespace="$2"
-    #local name="$3"
-    local start=$(date +%s)
-    sleep 1s
-    # @TODO: Rewrite this to *NOT* use istioctl experimental wait, since it was removed
-    # https://github.com/istio/istio.io/issues/16429
-    #if ! istioctl experimental wait --for=distribution --timeout=10s "$kind" "$name.$namespace"; then
-        #echo "Failed distribution of $kind $name in namespace $namespace"
-        #istioctl ps
-        #echo "TEST: wait for failed, but continuing."
-    #fi
-    echo "Duration: $(($(date +%s)-start)) seconds"
+    local kind="$1"
+    local namespace="$2"
+    local name="$3"
+    local max_attempts=30
+    local attempt=1
+    local sleep_time=2
+    local found=0
+    while (( attempt <= max_attempts )); do
+        # Check if the resource exists
+        if kubectl get "$kind" "$name" -n "$namespace" >/dev/null 2>&1; then
+            found=1
+            # Optionally, check status.conditions if available (for Gateway API, etc.)
+            # For most Istio CRDs, just existence is the best we can do
+            break
+        fi
+        sleep "$sleep_time"
+        ((attempt++))
+    done
+    if [[ $found -eq 0 ]]; then
+        echo "Timed out waiting for $kind $name in namespace $namespace to be created."
+        kubectl get "$kind" -n "$namespace"
+        return 1
+    fi
+    # Optionally, add more advanced checks for config distribution if needed
+    echo "$kind $name in namespace $namespace is present."
 }
 
 # Wait for named Gateway API gateway to be ready


tests/util/helpers.sh
```diff
@@ -108,19 +108,31 @@ _wait_for_statefulset() {
...
-    local start=$(date +%s)
-    sleep 1s
-    # @TODO: Rewrite this to *NOT* use istioctl experimental wait, since it was removed
-    # https://github.com/istio/istio.io/issues/16429
-    #if ! istioctl experimental wait --for=distribution --timeout=10s "$kind" "$name.$namespace"; then
-        #echo "Failed distribution of $kind $name in namespace $namespace"
-        #istioctl ps
-        #echo "TEST: wait for failed, but continuing."
-    #fi
-    echo "Duration: $(($(date +%s)-start)) seconds"
```
dhawton, line 123: Can we keep this? Useful when we need to determine what tests are taking the longest and why

tests/util/helpers.sh
```diff
@@ -108,19 +108,31 @@ _wait_for_statefulset() {
...
+    local kind="$1"
+    local namespace="$2"
+    local name="$3"
+    local max_attempts=30
+    local attempt=1
+    local sleep_time=2
```
dhawton, line 116:
```suggestion
    local sleep_time=1
```

2 might be too long, given how frequently we call this

tests/util/helpers.sh
AritraDey-Dev: Yes, we can keep this. Logging the duration helps identify which tests or waits are taking the longest. I’ll leave the commented-out istioctl experimental wait part for historical context, and keep the timing log to assist with test diagnostics.

Arhell (commented):
/ok-to-test

craigbox:
Does this PR cause a meaningful change to the time the tests take to run?

tests/util/helpers.sh
```diff
@@ -108,19 +108,35 @@ _wait_for_statefulset() {
...
+    local max_attempts=30
+    local attempt=1
+    local sleep_time=1
+    local found=0
+    local start
+    start="$(date +%s)"
+
+    while (( attempt <= max_attempts )); do
+        # Check if the resource exists
+        if kubectl get "$kind" "$name" -n "$namespace" >/dev/null 2>&1; then
```
craigbox, line 123: Aren't you basically doing

`kubectl wait --for=create -n $namespace $kind/$name --timeout=30s`

with this whole function?

Perhaps that would be cleaner?

craigbox: That's what some of the other functions do.

The thing here is, that the goal of this wasn't so much to ensure that the configuration was created in Kubernetes, which either method checks for, but to _ensure all the sidecars were programmed with it_.  They generally are, but in the instance that the object is created but is not translated to XDS and sent to the other proxy servers, there's no obvious way to tell.

So `kubectl wait` is probably necessary but not sufficient.

craigbox:
Congratulations, your testing has found a bug!

In #13208 the doc was changed to not require `samples/bookinfo/networking/virtual-service-all-v1.yaml`, so checking for `productpage` and `details` will fail as those are not actually created.

I'll change the test to only check for the services that are used by this example.  The cleanup will remove all of them, whether they exist or not.

craigbox:
That one feels a little more like a flake I've seen before.

/rest

AritraDey-Dev:
> I'll change the test to only check for the services that are used by this example. The cleanup will remove all of them, whether they exist or not.

Thank you for catching that and updating the test accordingly!

craigbox:
np.  Before we merge, please change this function to use `kubectl wait` and maybe even rename the function to `_wait_for_resource` throughout all the tests?

craigbox:
That one is the known flake again.

/retest

craigbox:
Every failure is about `httpbin, and smells like it isn't being given enough time to start.
Previously we always slept for at least one second; this PR removed that.  That might be the div we are not doing that.  I assume that is the difference.

Our options here are (a) always sleep 1s, and hope that is enough (b) sleep something like 5s in every test where we install `httpbin`.  I'd like feedback from other docs maintainers here.

Also - you've changed all the files in the `archive` directory, which you probably shouldn't have.  For now, I recommend you revert all the changes to the function name, and we can address that in a different PR.

craigbox:
Now you're truly starting to understand what it is to be an open source maintainer.

Endless /retests of something which seemed at first like a one line PR 😁

tests/util/helpers.sh
```diff
@@ -105,22 +105,24 @@ _wait_for_statefulset() {
...
+    local start_time
+    start_time="$(date +%s)"
+
+    if ! kubectl wait --for=create -n "$namespace" "$kind/$name" --timeout 30s; then
+        echo "Timed out waiting for $kind $name in namespace $namespace to be created."
+        kubectl get "$kind" -n "$namespace"
+        echo "Duration: $(( $(date +%s) - start_time )) seconds"
+        return 1
+    fi
+    echo "$kind $name in namespace $namespace is present."
```
craigbox, line 123: Don't need this output (or duration) if we succeed

tests/util/helpers.sh
```diff
@@ -105,22 +105,24 @@ _wait_for_statefulset() {
...
+    local namespace="$2"
+    local name="$3"
+    local start_time
+    start_time="$(date +%s)"
+
+    if ! kubectl wait --for=create -n "$namespace" "$kind/$name" --timeout 30s; then
+        echo "Timed out waiting for $kind $name in namespace $namespace to be created."
+        kubectl get "$kind" -n "$namespace"
+        echo "Duration: $(( $(date +%s) - start_time )) seconds"
+        return 1
```
craigbox, line 121: Use `exit 1` like in other functions?

tests/util/helpers.sh
```diff
@@ -105,22 +105,24 @@ _wait_for_statefulset() {
     fi
 }
 
-# Wait for Istio config to propagate
+# Wait for istio to be created
```
craigbox, line 108: "wait for a resource to be created"

tests/util/helpers.sh
```diff
@@ -105,22 +105,24 @@ _wait_for_statefulset() {
...
+    local kind="$1"
+    local namespace="$2"
+    local name="$3"
+    local start_time
+    start_time="$(date +%s)"
+
+    if ! kubectl wait --for=create -n "$namespace" "$kind/$name" --timeout 30s; then
+        echo "Timed out waiting for $kind $name in namespace $namespace to be created."
+        kubectl get "$kind" -n "$namespace"
+        echo "Duration: $(( $(date +%s) - start_time )) seconds"
```
craigbox, line 120: the duration will always be 30 seconds in this case

tests/util/helpers.sh
```diff
@@ -105,22 +105,24 @@ _wait_for_statefulset() {
...
-    echo "Duration: $(($(date +%s)-start)) seconds"
+    local kind="$1"
+    local namespace="$2"
+    local name="$3"
+    local start_time
+    start_time="$(date +%s)"
+
+    if ! kubectl wait --for=create -n "$namespace" "$kind/$name" --timeout 30s; then
+        echo "Timed out waiting for $kind $name in namespace $namespace to be created."
+        kubectl get "$kind" -n "$namespace"
```
craigbox, line 119: not that useful to know what else is there, we're testing that something is _not_ there

craigbox:
Interesting error

```
+_wait_for_istio gateway default mygateway
+_wait_for_istio(): local kind=gateway
+_wait_for_istio(): local namespace=default
+_wait_for_istio(): local name=mygateway
+_wait_for_istio(): kubectl wait --for=create -n default gateway/mygateway --timeout 30s
+_wait_for_istio virtualservice default helloworld-v1
+_wait_for_istio(): local kind=virtualservice
+_wait_for_istio(): local namespace=default
+_wait_for_istio(): local name=helloworld-v1
+_wait_for_istio(): kubectl wait --for=create -n default virtualservice/helloworld-v1 --timeout 30s
+_wait_for_istio(): echo 'Timed out waiting for virtualservice helloworld-v1 in namespace default to be created.'
+_wait_for_istio(): exit 1
```

Either this is a flake, or it's confused by (a) there being two different types of Gateway (b) waiting on `create` where a CRD should be `condition=established` ?

tests/util/helpers.sh
dhawton: Personally disagree here.. it was in the original, and I've used it before when I've needed to troubleshoot why a test was taking too long. It's a clue into why a specific step took long, was it waiting a long time for a certain resource to be ready? I don't think it'd hurt the output to leave it, but might help leave a breadcrumb that might help someone again later on.

AritraDey-Dev:
> Our options here are (a) always sleep 1s, and hope that is enough (b) sleep something like 5s in every test where we install `httpbin`. I'd like feedback from other docs maintainers here.


As we're encountering edge cases where config propagation is still flaky, I guess adding an optional advanced wait mechanism in the future (e.g., `polling Envoy config` or using `istioctl proxy-config`).
For now, I'm adding a sleep 2s just to check the logs of this test. A simple sleep 1s might be sufficient, but this is for initial validation.

dhawton:
These aren't flakes it seems..

AritraDey-Dev:
Screenshot from 2025-06-02 14-17-45


The real issue appears to be Istiod control plane health or authentication problems, not the `_wait_for_istio` helper. I believe the root cause is unrelated to this script and likely due to Istiod not being healthy or reachable during the tests.

look at the logs here...
https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/16515/doc.test.profile-default_istio.io/1929316089271947264/build-log.txt

tests/util/helpers.sh
AritraDey-Dev: Using `return 1` not causing tests to fail
While `exit 1 ` causing the tests to immediate failure.

AritraDey-Dev: An important thing to note here is that previously it did not use `exit 1`. Adding `exit 1` now might cause the test to fail because it immediately terminates the entire script when a resource isn't created, while the previous behavior (without `exit 1`) allowed the test to continue running even if the resource creation failed, potentially masking underlying issues.

AritraDey-Dev:
Hi @craigbox @dhawton,
Just checking in on this PR as it has been on hold for a while.Any thoughts on this?

tests/util/helpers.sh
dhawton: @AritraDey-Dev @craigbox there's this comment still

AritraDey-Dev: Yeah, I mentioned earlier that keeping this line helps make debugging easier. Please let me know if I should add it back.

dhawton: Mostly for craig since he contradicted my ask to put it back, but you asked for an update so I tagged you so you could see too this is still waiting on craig.

craigbox: If Daniel wants to keep it in, please keep it in.
(I suggested removing it because it is noise in any success situation, but then again so are all these logs)

craigbox:
> The real issue appears to be Istiod control plane health or authentication problems, not the `_wait_for_istio` helper. I believe the root cause is unrelated to this script and likely due to Istiod not being healthy or reachable during the tests.
> 
> look at the logs here... https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/16515/doc.test.profile-default_istio.io/1929316089271947264/build-log.txt

I read that error as "test failed, trying to clean up, error logged while cleaning up"

I'm not 100% sure how best to progress on this.  Right now it seems like we've just slowed everything down by 1s per test?
Daniel asked that we put back the time check.

tests/util/helpers.sh
AritraDey-Dev: Added the logs back.

AritraDey-Dev: Tests are passing 🎉

dhawton:
> > The real issue appears to be Istiod control plane health or authentication problems, not the `_wait_for_istio` helper. I believe the root cause is unrelated to this script and likely due to Istiod not being healthy or reachable during the tests.
> > look at the logs here... https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/16515/doc.test.profile-default_istio.io/1929316089271947264/build-log.txt
> 
> I read that error as "test failed, trying to clean up, error logged while cleaning up"
> 
> I'm not 100% sure how best to progress on this. Right now it seems like we've just slowed everything down by 1s per test? Daniel asked that we put back the time check.

I agree that we slowed everything down.. unfortunately, whichout a replacement for the old `x wait` command of istioctl, I don't know of a way to properly apply and wait for programming, deployment, etc. It's really unfortuante that that experimental command was dropped,.

I'm ok with implementing this for now... it makes things slower but hopefully will address some of the flakes that happen because we checked or tried to use something that hadn't yet applied.

craigbox (approved):

craigbox:
Ok, approving, and we'll see what the impact is to other PRs.

PR #16526, opened 2025-05-28
Add override for redirects

craigbox:
Add `?redirect=false` to a URL to stop the edge function from redirecting `v1.xx` paths to `latest`.


dhawton (approved):
@craigbox added DNM as #16540 is approved but has DNM

PR #16457, opened 2025-05-07
Added Out-Of-Date Documentation Header for v1.24

sbegin0:
## Description

Added a warning header to all docs pages warning user that this version is out-of-date with an option to go to /latest

## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

PR #16529, opened 2025-05-28
Regenerate 1.24 and 1.25 archive with header

craigbox:
Applied #16457 to `release-1.25` branch and regenerated archive.

This should result in all docs pages (in all languages) having a localised header.

I would prefer it didn't touch the other pages, but we have a minified CSS file with a name that has changed.


craigbox:
* https://deploy-preview-16529--preliminary-istio.netlify.app/v1.25/docs/ : no header
* https://deploy-preview-16529--preliminary-istio.netlify.app/v1.25/docs/overview: no header

I built the archive with:

```
export SCRIPT_DIR=scripts
export PREV_MINOR=1.25
DRY_RUN=1 ISTIOIO_GIT_SOURCE=https://github.com/istio/istio.io.git "${SCRIPT_DIR}"/build_old_archive.sh "build-old-archive-${PREV_MINOR}.0"
```

1.25 was never marked as archived (compare [1.25](https://github.com/istio/istio.io/blob/release-1.25/data/args.yml) and [1.24](https://github.com/istio/istio.io/blob/release-1.24/data/args.yml).

This looks like something John did with a commit, rather than a PR, and will have to go into the new release process. /cc @dhawton 

I'll build 1.24 in the meantime and send a PR to archive 1.25.

kfaseela:
> https://deploy-preview-16529--preliminary-istio.netlify.app/v1.24/docs/ 🎉 

When I click on the "read latest", it says page requested cannot be found.

dhawton:
> > https://deploy-preview-16529--preliminary-istio.netlify.app/v1.24/docs/ 🎉
> 
> When I click on the "read latest", it says page requested cannot be found.

I see the same.. looks like the read latest link retains the /v1.24 path prefix and just appends /latest/docs.

craigbox:
Hmm, it wasn't supposed to do that.  
/cc @sbegin0

craigbox:
OK, here's the deal.  

* `build_old_archive.sh` calls `make archive` which is `archive_version.sh`
* that calls `build_site.sh /"${VERSION}"`
* that runs Hugo with `--baseURL "$1"`

The link we add is `"/latest{{ .RelPermalink | safeURL }}"` 

Thus any link added to this page is going to be relative to /v1.24/.

Workarounds:

* We could change the URL to start `https://istio.io/latest/`, which will almost certainly break another linter
* We could special-case /vX.XX/latest/ URLs in the netlify handler
* ...something else?

craigbox:
Trying `//latest` on the v1.25 archive to see what that makes the browser do

craigbox:
(Status update: this is waiting on the baseURL work that Ajay is doing at present.)

PR #15700, opened 2024-09-17
Update index.md

AleksFirsta:
all cofnig align of lest and when you copy/paste you need to update it manually

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Performance and Scalability
- [ ] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [ ] Developer Infrastructure
- [ ] Localization/Translation

PR #16547, opened 2025-05-29
doc: fix formatting in mirroring task

msahihi:
## Description

Corrected improper formatting in bash code snippets within the Creating a default routing policy and Mirroring traffic to httpbin-v2 sections. Previously, copy-pasting these snippets would result in parsing errors like 
```bash
error: error parsing STDIN: error converting YAML to JSON: yaml: line 6: did not find expected key. 
```
This prevents users from directly using the provided examples.

Screenshot 2025-05-29 at 19 52 01

This PR ensures the commands are directly usable.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/traffic-management/mirroring/index.md
+++ b/content/en/docs/tasks/traffic-management/mirroring/index.md
@@ -22,6 +22,7 @@ you will apply a rule to mirror a portion of traffic to `v2`.
 ## Before you begin
 
 1. Set up Istio by following the [Installation guide](/docs/setup/).
+
 1. Start by deploying two versions of the [httpbin]({{< github_tree >}}/samples/httpbin) service that have access logging enabled:
 
     1. Deploy `httpbin-v1`:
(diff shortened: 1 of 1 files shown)


dhawton (changes requested):
This breaks the formatting of the page and does not actually change the copied text. I'm not sure how you're utilizing clipboard to get different text, but using both highlight + CTRL+C or the inline copy button added to the codeblock results in properly formatted scripts.

Look at how your changes break the style of the page between `3. Deploy the curl workload you’ll use to send requests to the httpbin service:` and `3. Deploy the httpbin Kubernetes service:` (your removal of the indents is also breaking the ordered numbering).

msahihi:
Thank you for the feedback. 

>This breaks the formatting of the page and does not actually change the copied text

I am checking https://deploy-preview-16547--preliminary-istio.netlify.app/latest/docs/tasks/traffic-management/mirroring/ and it is perfectly working. Code format for `Creating a default routing policy` &`Mirroring traffic to httpbin-v2` is fixed.
Is your observation different ?

> I'm not sure how you're utilizing clipboard to get different text, but using both highlight + CTRL+C or the inline copy button added to the codeblock results in properly formatted scripts.

On macOS, the `Copy to clipboard` button is not working as expected for me in either Safari or Chrome. It copies the rendered text with a broken format.
 
> Look at how your changes break the style of the page between `3. Deploy the curl workload you’ll use to send requests to the httpbin service:` and `3. Deploy the httpbin Kubernetes service:` (your removal of the indents is also breaking the ordered numbering).

Not sure how I borked the format. This is the same format as the published one. Checking [the published mirroring task](https://istio.io/latest/docs/tasks/traffic-management/mirroring/ ) I can see both `3. Deploy the httpbin Kubernetes service:` and `3. Deploy the curl workload you’ll use to send requests to the httpbin service:` . The first one is the third step of the `2. Start by deploying two versions of the httpbin service that have access logging enabled:` and the second one is the 3rd step for `Before you begin`  Here is also the screenshot from the published page
Screenshot 2025-06-01 at 14 05 17

but I will check failing lint test

dhawton:
Because you unindented code blocks, it moves the code blocks to not be "under" the number... for example (and in the prelim further down the page), this is what it is supposed to look like

expected

Versus what your change looks like

broken

We indent code blocks to be *under* the ordered list item to keep the numbering consistent and to visually reflect that code block belongs to that step.

craigbox:
As an aside: we don't expect this to be easy, or fixable in this file. It's probably a bug in the tabset implementation.

kfaseela:
> As an aside: we don't expect this to be easy, or fixable in this file. It's probably a bug in the tabset implementation.

Yes, see : https://github.com/istio/istio.io/pull/15700#issuecomment-2361117293

Issue #16559, opened 2025-06-05
fix the site baseURL


craigbox:
While working on #16558 I noticed that our sitemaps are invalid.  To fix this, we will need to update our deployment scripts to pass the correct baseURL into Netlify.

This was attempted once (#14725) but reverted (#14757) because we didn't get the right URLs into the system.

The URL needs to be:

* https://preliminary.istio.io/ when on prelim
* https://istio.io/ when on prod
* https://deploy-preview-16378--preliminary-istio.netlify.app/ (or appropriate for version) when deployed from a PR preview

[This might be useful context](https://discourse.gohugo.io/t/correct-baseurl-on-netlify-previews/10429) as how to address this.

PR #16568, opened 2025-06-09
Fix: Site BaseURLs

Ajay-singh1:
## Description

Related Issue:-#16559

This PR fixes incorrect baseURL values used during site generation, which were causing invalid or broken URLs in sitemap.xml.

## Summary of changes

Added logic for baseURLs based on different netlify contexts.

## How can we check if the baseURLs are pointing correctly?

<ul>
<li>
Really can`t tell until the PR is merged into the master branch.
</li>
</ul>

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -72,18 +72,18 @@ JUNIT_REPORT := $(shell which go-junit-report 2> /dev/null || echo "${ISTIO_BIN}
 ISTIO_SERVE_DOMAIN ?= localhost
 export ISTIO_SERVE_DOMAIN
 
-# Determine the baseURL for the site depending on the context
+# Determine the base URL for the Netlify-hosted site depending on the Hugo context.
+# 'production' context: Use site URL (istio.io, preliminary.istio.io or istio-staging.netlify.app)
+# 'deploy-preview' context: Use per-build URL (deploy-preview-16568--preliminary-istio.netlify.app)
+# any other context: use relative URLs
 ifeq ($(CONTEXT),production)
-  baseurl := $(URL)/latest
-
-# For deploy-preview, use DEPLOY_PRIME_URL
+  NETLIFY_URL := $(URL)/latest
 else ifeq ($(CONTEXT),deploy-preview)
-  baseurl := $(DEPLOY_PRIME_URL)/latest
-
-# Default base URL
+  NETLIFY_URL := $(DEPLOY_PRIME_URL)/latest
 else
-  baseurl := /latest
+  NETLIFY_URL := /latest
 endif
+export NETLIFY_URL
 
 
 # Which branch of the Istio source code do we fetch stuff from
@@ -156,8 +156,7 @@ netlify_install:
 
 netlify: netlify_install
 	@scripts/gen_site.sh
-# Dynamically set the baseurl based on the Netlify Context
-	@scripts/build_site.sh "${baseurl}"   
+	@scripts/build_site.sh "${NETLIFY_URL}"
 	@scripts/include_archive_site.sh
 
 # ISTIO_API_GIT_SOURCE allows to override the default Istio API repository, https://github.com/istio/api@$(SOURCE_BRANCH_NAME)


craigbox:
Can you ping me on the Istio Slack please Ajay?

netlify.toml
```diff
@@ -1,4 +1,5 @@
 [build]
+    command = "bash ./scripts/set_baseurl_and_build.sh"
```
craigbox: I think you want to test this, but I'm not immediately sure why we need a new script and can't just do this in `build_site.sh`

craigbox:
/ok-to-test

(I think you can join as a member now btw)

netlify.toml
Ajay-singh1: Sure!

scripts/build_site.sh
```diff
@@ -14,16 +14,31 @@
...
 
+# Determine BASEURL based on Netlify context
+
+if [[ "$CONTEXT" == "production" ]]; then
+  BASEURL="https://istio.io/"
+elif [[ "$CONTEXT" == "deploy-preview" ]]; then
+  BASEURL="${DEPLOY_PRIME_URL}/"
+else
+  BASEURL="https://preliminary.istio.io/"
+fi
```
craigbox: this will set BASEURL to prelim when you're running on localhost, as this script is used there too

scripts/build_site.sh
```diff
@@ -14,16 +14,31 @@
...
 # limitations under the License.
 
+
+#!/bin/bash
 set -ex
 
+# Determine BASEURL based on Netlify context
+
+if [[ "$CONTEXT" == "production" ]]; then
+  BASEURL="https://istio.io/"
```
craigbox: I'd like to look at what $URL is set to when you're running on netlify in production; you shouldn't need to hardcode this URL in the script, it should be able to be set by netlify

scripts/build_site.sh
Ajay-singh1: Okay Understood.

scripts/build_site.sh
Ajay-singh1: Yeah I saw , it was already their in the script.

Ajay-singh1:
@craigbox This should fix the base URL issue.Here is the sitemap.xml for this PR:-

Screenshot from 2025-06-09 20-18-56

netlify.toml
```diff
@@ -1,6 +1,14 @@
 [build]
-    publish = "public"
-
+  publish = "public"
```
craigbox: netlify.toml supports different sections for the different contexts already.

https://github.com/gohugoio/hugoDocs/blob/master/netlify.toml#L11

Hugo best practices say that `baseURL` should be set to "https://istio.io/" (probably with `/latest/`?  I don't quite recall where that is added)

From a quick read of the docs, the environment in development _should_ use localhost instead of the `baseURL`.

netlify.toml
```diff
@@ -1,14 +1,11 @@
...
-      hugo --baseURL=\"https://preliminary.istio.io/\"
-    fi
-  """
+    publish = "public"
+[context.production]
+    command="hugo --baseURL='https://istio.io/latest'" 
+[context.deploy-preview]
+    command="hugo --baseURL='${DEPLOY_PRIME_URL}/'"
+[context.branch-deploy]
+    command="hugo --baseURL='https://preliminary.istio.io/'"
```
craigbox: feels like you're just spraying things here rather than doing the research and pushing them through thoughtfully fwiw.

* note how preliminary.istio.io has /latest/ URLs too, so if you add that suffix to one, you have to add it to all three.
* the comment before suggested that we shouldn't need to set a baseURL for production but we should set it in the Hugo config file in general.
* I'm not sure the contexts are correct here (and this may require me to find this out in netlify) - preliminary may or may not be a production context on a different site.
The comment

Ajay-singh1: You are right I have to do some research before pushing , Will be back soon.

Ajay-singh1:
@craigbox I have got a working solution now.We can check the sitemap for preliminary deployment and it is generating those sitemaps correctly.See :- <a href="https://deploy-preview-16568--preliminary-istio.netlify.app/sitemap.xml"> Sitemap for Preliminary deployment</a>.

Ajay-singh1:
We have to cherry-pick release to see it in production.

craigbox:
we could do that, assuming we have a roll back plan.

One thing I was wondering is if we should set up a custom page which just shows the variables, push that through all the sites first, and verify the values are as expected?

Ajay-singh1:
I’ve reviewed the changes and I’m confident they won’t break anything. We have a rollback plan — we can revert the commit or use Netlify’s rollback option. I think we’re safe to proceed without the custom debug page.

craigbox:
This isn't working as expected.

Look at https://deploy-preview-16568--preliminary-istio.netlify.app/ for example - all the links are now to preliminary.istio.io.

Ajay-singh1:
This is actually really very complicated.

Screenshot_20250627-192540_Chrome

Netlify logs tell the baseURL is /latest which actually is the major issue here.Preliminary and Main site both have the CONTEXT set to production so netlify configuration will fail if we try to set the baseURL.Maybe updating sitemap layout will work though.

/cc @dhawton @craigbox

Ajay-singh1:
@craigbox Up for review! Their won`t be any room for errors this time. :)

Makefile.core.mk
```diff
@@ -72,8 +72,21 @@ JUNIT_REPORT := $(shell which go-junit-report 2> /dev/null || echo "${ISTIO_BIN}
 ISTIO_SERVE_DOMAIN ?= localhost
 export ISTIO_SERVE_DOMAIN
 
-ifeq ($(CONTEXT),production)
-baseurl := "$(URL)"
+# Determine current Git branch
+CURRENT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
```
craigbox: Ooooh, that's smart!

Makefile.core.mk
```diff
@@ -72,8 +72,21 @@ JUNIT_REPORT := $(shell which go-junit-report 2> /dev/null || echo "${ISTIO_BIN}
...
+  baseurl := https://preliminary.istio.io/latest
+
+# If the current branch is release branch, we use the istio.io/latest as the base URL
+else ifneq ($(filter release-%, $(CURRENT_BRANCH)),)
+  baseurl := https://istio.io/latest
+
+# If the current branch is not master or release branch, we use /latest as the base URL
+# This is useful for feature branches or pull requests and deploy-preview
+else
+  baseurl := /latest
```
craigbox: I would still like to get fully qualified URLs in previews if possible. 
Is there a variable we can add here?

Makefile.core.mk
```diff
@@ -72,8 +72,21 @@ JUNIT_REPORT := $(shell which go-junit-report 2> /dev/null || echo "${ISTIO_BIN}
...
+CURRENT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
+
+# If the current branch is master, we use preliminary.istio.io/latest as the base URL
+ifeq ($(CURRENT_BRANCH),master)
+  baseurl := https://preliminary.istio.io/latest
+
+# If the current branch is release branch, we use the istio.io/latest as the base URL
+else ifneq ($(filter release-%, $(CURRENT_BRANCH)),)
+  baseurl := https://istio.io/latest
+
```
dhawton: This isn't always true. When we do release branch cutting, we'll create release-1.xx, point preliminary at the release branch to test the docs are working correctly, etc. before pointing the main site to this branch.

If this procedure is changing, we'll need to update `README.md` with the new flow.

Ajay-singh1: Yes ,correct!

Makefile.core.mk
Ajay-singh1: Yes we can! ```DEPLOY_PRIME_URL```

Makefile.core.mk
craigbox: We don't use preliminary for that, do we?  We use [istio-staging](https://istio-staging.netlify.app/), which we would have to include some kind of case for

Ajay-singh1: Thanks @craigbox If that is the case then we don't have a problem here, @dhawton I have checked netlify configuration and didn't find the release branch is getting deployed from their 😑😑

craigbox: I'll check it tomorrow.
We do have to think this through. We should perhaps make sure those values are all set on Netlify and use their environment vars.

But all the links on istio-staging should be relative to that; it's true today because all are always relative anyway!

Ajay-singh1: Thanks!

dhawton: > Thanks @craigbox If that is the case then we don't have a problem here, @dhawton I have checked netlify configuration and didn't find the release branch is getting deployed from their 😑😑

we do use prelim, and it's one that craig or faseela sets temporarily to check, and then sets it back once the main site is moved over.

Ajay-singh1: Okay! Thanks. In that case README.md must be updated to avoid confusion.

craigbox: > we do use prelim

I've only ever done this on `istio-staging` which is a separate configuration.  (Not to say that it might never happen)

Ajay-singh1:
@craigbox We are very close! Fully qualified URLs from the deployment-preview.
Attached PNG for reference:-

Screenshot from 2025-07-09 14-28-02

Makefile.core.mk
Ajay-singh1: I agree with Daniel my approach has some issue as the staging environment also uses the release branch for deployment.

Ajay-singh1:
This will work correctly if the environment variables are set correctly in netlify.

craigbox:
https://docs.netlify.com/configure-builds/environment-variables/#deploy-urls-and-metadata

> **URL**: URL representing the main address to your site. It can be either a Netlify subdomain or your own custom domain if you set one; for example, https://petsof.netlify.app/ or https://www.petsofnetlify.com/.

Primary domains for the three sites are: istio.io, preliminary.istio.io, istio-staging.netlify.app

Ajay-singh1:
Thanks! Craig , Let's merge this PR now.This will work 😀.

When on ```https://istio.io``` the sitemap will be generated like ```https://istio.io/latest/sitemap.xml```

When on ```preliminary.istio.io``` the sitemap will be generated like ```https://preliminary.istio.io/latest/en/sitemap.xml```

When on ```istio-staging.netlify.app``` the sitemap will be generated like
```https://istio-staging.netlify.app/latest/en/sitemap.xml```

When on```deploy-previews``` the sitemap will be generated like this
```https://$(deploy-preview-url)/latest/en/sitemap.xml```

All the URLs will be set correctly and Hugo will use this different baseURLs when building and deploying the site.

File: Makefile.core.mk [hunk 9]
```diff
@@ -72,10 +72,20 @@ JUNIT_REPORT := $(shell which go-junit-report 2> /dev/null || echo "${ISTIO_BIN}
 ISTIO_SERVE_DOMAIN ?= localhost
 export ISTIO_SERVE_DOMAIN
 
+# Determine the baseURL for the site depending on the context
```
craigbox: ~I went back and forward on deciding if I should suggest you reuse ISTIO_SERVE_DOMAIN here.  Probably not, because the below would then override people who are trying to set the domain for running `make serve`.~

~(Then I went back to this comment and changed to the suggestion below)~

(Then I redid it all a third time)

I've also tried to figure out how we can get `/latest` into the site config for `baseURL`.  When you run `hugo serve` (not `make serve`) it doesn't use `/latest/` in the URL; you have to specify that yourself.  (I get bitten by this a lot because I run a local `hugo`, not the containerised version that comes with `make serve`.)

Right now I don't think it can be done, because the baseURL that you set in hugo.toml is for the `production` environment, but the `development` environment (as used by `hugo serve`) doesn't obey that value. I'll open an issue on the Hugo issue tracker for that.

Instead, I'm going to change the variable name to match the format used in the rest of the file.

Makefile.core.mk
```diff
@@ -72,10 +72,20 @@ JUNIT_REPORT := $(shell which go-junit-report 2> /dev/null || echo "${ISTIO_BIN}
...
-baseurl := "$(URL)"
+  baseurl := $(URL)/latest
+
+# For deploy-preview, use DEPLOY_PRIME_URL
+else ifeq ($(CONTEXT),deploy-preview)
+  baseurl := $(DEPLOY_PRIME_URL)/latest
+
+# Default base URL
+else
+  baseurl := /latest
```
craigbox:
```suggestion
  NETLIFY_URL := $(URL)/latest
else ifeq ($(CONTEXT),deploy-preview)
  NETLIFY_URL := $(DEPLOY_PRIME_URL)/latest
else
  NETLIFY_URL := /latest
```

Makefile.core.mk
```diff
@@ -72,10 +72,20 @@ JUNIT_REPORT := $(shell which go-junit-report 2> /dev/null || echo "${ISTIO_BIN}
...
 endif
```
craigbox, line 85:
```suggestion
endif
export NETLIFY_URL
```

craigbox: to match the rest of the vars; in case we use this somewhere else later

Makefile.core.mk
craigbox:
```suggestion
# Determine the base URL for the Netlify-hosted site depending on the Hugo context.
# 'production' context: Use site URL (istio.io, preliminary.istio.io or istio-staging.netlify.app)
# 'deploy-preview' context: Use per-build URL (deploy-preview-16568--preliminary-istio.netlify.app)
# any other context: use relative URLs
```

Makefile.core.mk
```diff
@@ -146,7 +156,8 @@ netlify_install:
 
 netlify: netlify_install
 	@scripts/gen_site.sh
-	@scripts/build_site.sh "/latest"
+# Dynamically set the baseurl based on the Netlify Context
+	@scripts/build_site.sh "${baseurl}"
```
craigbox:
```suggestion
	@scripts/build_site.sh "${NETLIFY_URL}"
```

Makefile.core.mk
```diff
@@ -146,7 +156,8 @@ netlify_install:
 
 netlify: netlify_install
 	@scripts/gen_site.sh
-	@scripts/build_site.sh "/latest"
+# Dynamically set the baseurl based on the Netlify Context
```
craigbox:
```suggestion
```

netlify.toml
```diff
@@ -21,4 +21,4 @@
 
 [[edge_functions]]
   path = "/v*"
-  function = "redirect-to-latest"
\ No newline at end of file
+  function = "redirect-to-latest"
```
craigbox, line 24: removed CRLF at end of line?

craigbox:
deploy preview log:

> 1:42:21 PM: + hugo --minify --baseURL https://deploy-preview-16568--preliminary-istio.netlify.app/latest

craigbox (approved):

craigbox:
preliminary log:

> 1:50:50 PM: + hugo --minify --baseURL https://preliminary.istio.io/latest

PR #16574, opened 2025-06-12
LFX'25:Update the pipeline to achieve clean build without any errors

Ajay-singh1:
## Description

This PR replaces Babel with esbuild as the build tool for this repository. The main goals of this is  to improve build performance, reduce configuration complexity, modernize the toolchain, and generate cleaner builds.

## Overview

Babel has served as a popular JavaScript transpiler for years, but it comes with some downsides in the context of modern library development:
<ul>
     <li>Performance bottlenecks: Babel is significantly slower, especially on large codebases or repeated builds</li>
     <li>Maintenance concerns: Many Babel plugins and presets are either deprecated, minimally maintained, or lag behind the evolving JavaScript standard.</li>
     <li>Unnecessary complexity: Babel's plugin-based architecture leads to verbose and often brittle configurations</li>
</ul>

<b>esbuild</b>, on the other hand, is designed for speed and simplicity, with modern JavaScript/TypeScript support and built-in minification and bundling.

## Technical Changes
<b>Removed</b>
<ul>
    <li><b>Babel related dependencies</b></li>
   <ul><li><code>@babel/core</code></li></ul>
   <ul><li><code>@babel/preset-env</code></li></ul>
   <ul><li><code>Anyother babel related dependency</code></li></ul>
</ul>

## Added
<ul>
    <li><b><code>esbuild</code></b>as a dev dependency</li>
</ul>
<ul>
     <li>A new build script using esbuild, capable of:</li>
      <ul> 
           <li>Transpiling modern JavaScript (ES6+) to target environments </li>
           <li>Minimification of Javascript files to achieve faster builds</li>
       </ul>
</ul>
  
## Testing

1. The output structure and exports matches with previous Babel generated output
2. The site is build successfully without having any errors or warnings 

## Result
1. Simplified build setup with fewer and less boilerplate dependencies
2. More future-proof and modern toolchain

 
    

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/zh/docs/ops/best-practices/security/index.md
+++ b/content/zh/docs/ops/best-practices/security/index.md
@@ -61,6 +61,21 @@ Istio 新的 Ambient 数据平面模式引入了一种新的拆分数据平面�
 需要注意的是，`GatewayClass` 是集群范围的资源，将命名空间范围的策略绑定到它需要特别小心。
 Istio 要求绑定到 `GatewayClass` 的策略位于根命名空间中，通常是 `istio-system`。
 
+对于 waypoint，标准的不允许任何行为的策略是：
+
+{{< text yaml >}}
+apiVersion: security.istio.io/v1
+kind: AuthorizationPolicy
+metadata:
+  name: allow-nothing-istio-waypoint
+  namespace: istio-system
+spec:
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: GatewayClass
+    name: istio-waypoint
+{{< /text >}}
+
 {{< tip >}}
 当在 waypoint 中使用默认拒绝模式时，除了“经典”默认拒绝策略外，
 还应使用绑定到 `istio-waypoint` `GatewayClass` 的策略。


craigbox:
Great initiative. 
The new modules would need to be in the build container before the lint will pass.
I don't have enough context to say yes or no without doing a bunch more work.

Can you comment on (or show a diff of) the before/after of the generated JS?

You're asking for approval to merge something that transpiles/touches the site JavaScript; there's no point in showing a video which scrolls a bunch of random pages, we would need to see the before/after for things that are executed by these scripts.

(p.s. generating the site takes ~20 seconds on my laptop.  Do I read right that yours takes 440 seconds - 20 times longer? We need to do something about that!)

AritraDey-Dev:
I found that this PR is causing the `headerAnimation` script and the `expandable` list functionality to break. If you check the js folder, the `headerAnimation.js` file is likely missing from the build. Additionally, a few other files might also be missing from the js directory.
If you compare the builds generated using `Babel` and `esbuild`, you'll notice that the file structures differ significantly.

Ajay-singh1:
@craigbox The work is not completed yet I'll ping you when it does. @AritraDey-Dev Yeah I agree with you that the file structure is different but I have resolved this I will push the changes in the meantime.
Thanks

Ajay-singh1:
@craigbox This PR is up for review , after this PR is merged <a href="https://github.com/istio/tools/pull/3222">istio/tools PR </a>.Hopefully the lint will pass here.

craigbox:
https://github.com/istio/tools/pull/3222 LGTM but I'm not in the reviewer set.  Daniel is, so I think that will go through today his time.

scripts/gen_site.sh
```diff
@@ -18,36 +18,50 @@ set -ex
...
-
-babel --source-maps --minified --no-comments --presets minify \
-  tmp/js/headerAnimation.js \
-  --out-file generated/js/headerAnimation.min.js
-
-babel --source-maps --minified --no-comments \
-  tmp/js/themes_init.js \
-  --out-file generated/js/themes_init.min.js
+#Entrypoint for esbuild to bundle and minify through an entrypoint.js file
+cat <<EOF > tmp/js/entrypoint.js
```
craigbox: Why generate this every time and not check it in?

(This question doesn't presuppose I know the answer)

scripts/gen_site.sh
```diff
@@ -18,36 +18,50 @@ set -ex
...
+  --minify \
+  --sourcemap \
+  --target=es6 \
+  --outfile=generated/js/all.min.js
+
+esbuild tmp/js/headerAnimation.js \
+  --minify \
+  --sourcemap \
+  --target=es6 \
+  --outfile=generated/js/headerAnimation.min.js
```
craigbox: Is it still relevant that this be a separate file?
(Some digging may be required to find out what it does and why it was done this way)

scripts/gen_site.sh
```diff
@@ -18,36 +18,50 @@ set -ex
...
+
+esbuild tmp/js/themes_init.js \
+  --minify \
+  --sourcemap \
+  --target=es6 \
+  --outfile=generated/js/themes_init.min.js
```
craigbox: same here, previously with Babel this was `--minify` but not `--presets minify` for some reason?
Does it still make sense to be its own thing?

src/ts/overlays.ts
```diff
@@ -11,18 +11,21 @@
...
 // limitations under the License.
-
+import { listen } from "./utils";
+import { click } from "./constants";
 declare class Popper {
     constructor(a: HTMLElement, b: HTMLElement, c: any);
     public destroy(): void;
 }
+declare var iconFile: string;
+// Path to the SVG icon sprite file
```
craigbox: comments should be above the line?

src/ts/callToAction.ts
```diff
@@ -11,7 +11,6 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
```
craigbox: were changes such as this required to lint, or something?

src/ts/codeBlocks.ts
```diff
@@ -11,14 +11,16 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
+import { listen , copyToClipboard , getById} from "./utils";
+import { button , ariaLabel , mouseenter , mouseleave , click , active} from "./constants";
+import { readLocalStorage } from "./themes_init";
```
craigbox: review top tip: I don't know anything about Typescript but it would be worth providing a little context as to why these need to be added

scripts/gen_site.sh
Ajay-singh1: esbuild needs an entrypoint for bundling the files and it does when the files are ESModules and the typescript files which is used in this repo is not ESModules but are legacy old standard type global scripts.esbuild follows the import and build a dependency graph so this temporary file is needed for esbuild for bundling the files together and minify them.

You can check the tsconfig.json the target is set to es6.

When you compile a .ts file for ex:- ```tsc src/ts/faq.ts --target es6``` It will throw an error because the files are not following the ES6 standard.

This file is just an entrypoint and nothing else for bundling and minification.

src/ts/callToAction.ts
Ajay-singh1: Yes ```make lint-typescript``` was throwing errors.

scripts/gen_site.sh
Ajay-singh1: You run ```babel presets-minify``` it will minify multiple files together in isolation and generate a single output file which doesn't raise global conflicts such as function , var scoping conflicts etc.esbuild does the same thing but is more advanced like it removes any unused code from the bundle through a process called as tree shaking and minifies aggressively.

scripts/gen_site.sh
Ajay-singh1: Yeah it needs to be a seperate file otherwise it can break the site.I have to do a bunch more test.

src/ts/codeBlocks.ts
Ajay-singh1: This has to be added to make them ESModules and follow ES6 standards otherwise the .ts files will be complied as global scripts and global scripts are now legacy code which doesn't follow ES6 standard.

Ajay-singh1:
@dhawton Can you kindly also review this PR and share your comments?It would be really helpful.

scripts/gen_site.sh
dhawton: Why not have that file exist in say  src/ts/entrypoint.js rather than creating it arbitrarily?

For future maintainers, it would not be intuitive to look in gen_site.sh for a magic file.

src/ts/overlays.ts
dhawton:
```suggestion
// Path to the SVG icon sprite file
declare var iconFile: string;
```

Ajay-singh1:
Thanks @dhawton and @craigbox for the detailed review I have done the necessary changes. @keithmattix Can you kindly also review this PR? just  to ensure everything looks good.

AritraDey-Dev (commented):
While running make serve, if I don't delete the tmp/js folder and try to run make serve again, I encounter the following error:
```
src/ts/callToAction.ts:14:7 - error TS2451: Cannot redeclare block-scoped variable 'callToActionDelayMs'.

14 const callToActionDelayMs = 250;
         ~~~~~~~~~~~~~~~~~~~

  tmp/js/callToAction.js:15:7
    15 const callToActionDelayMs = 250;
             ~~~~~~~~~~~~~~~~~~~
    'callToActionDelayMs' was also declared here.

src/ts/callToAction.ts:16:10 - error TS2393: Duplicate function implementation.

16 function handleCallToAction(): void {
            ~~~~~~~~~~~~~~~~~~

src/ts/kbdnav.ts:15:7 - error TS2300: Duplicate identifier 'KbdNav'.

15 class KbdNav {
         ~~~~~~

  tmp/js/kbdnav.js:15:7
    15 class KbdNav {
             ~~~~~~
    'KbdNav' was also declared here.


Found 3 errors in 2 files.

Errors  Files
     2  src/ts/callToAction.ts:14
     1  src/ts/kbdnav.ts:15
make[1]: *** [Makefile.core.mk:83: site] Error 2
make: *** [Makefile:44: serve] Error 2
```
this was not the case before

Although this is dev-specific and won't occur in prod

Ajay-singh1:
Nice catch! The error happened because of leftover compiled files from the previous build you won't get this error before because how babel handles files is different then how esbuild handles them.

BTW this is a quick fix.

AritraDey-Dev:
I just gave an example that running `make serve` is throwing errors. If you noticed, it's also throwing errors for other make targets.
Adding the clean target to every make command is not a good solution just to resolve the errors.If you add clean to every `make` target, it unnecessarily takes time to clean before performing the actual task.

Also, after you build it, you’ll see a bunch of red squiggly lines — this shouldn’t happen.
I'm also not sure why you added so many import and export statements. Previously, when we were using a Babel-based build, the functions or variables that needed to be compiled globally were already there.

craigbox:
I think it's reasonable for a build to clean up any temporary files after using them?

However the `entrypoint.js` file should not contain reference to tmp files that aren't normally there:

```
import "../../tmp/js/constants.js";
import "../../tmp/js/utils.js";
import "../../tmp/js/feedback.js";
```

craigbox:
I think it's reasonable for a build to clean up any temporary files after using them?

However the `entrypoint.js` file should not contain reference to tmp files that aren't normally there:

```
import "../../tmp/js/constants.js";
import "../../tmp/js/utils.js";
import "../../tmp/js/feedback.js";
```

Makefile.core.mk
```diff
@@ -128,7 +128,7 @@ lint-fast: clean_public build_nominify lint-copyright-banner lint-python lint-ya
 lint-md: clean_public build_nominify
 	@SKIP_LINK_CHECK=true scripts/lint_site.sh en
 
-serve: site
+serve: clean site
```
craigbox: doesn't seem a good idea

Ajay-singh1:
@AritraDey-Dev esbuild isn't a transpiler like babel it also does bundling of files so this has to happen anyways.Regarding the export and imports I already told Craig that the repo is using global legacy code and converting them to ES6 modules is a good idea.Legacy code was used before ES6 and it was a standard then.

It is obvious to clean temporary files before each build so that isn't a major issue here.

For ex:- Try to compile a .ts file with estarget set to es6 you will get errors.

Makefile.core.mk
Ajay-singh1: Completely Agree!

Makefile.core.mk
```diff
@@ -128,7 +128,7 @@ lint-fast: clean_public build_nominify lint-copyright-banner lint-python lint-ya
 lint-md: clean_public build_nominify
 	@SKIP_LINK_CHECK=true scripts/lint_site.sh en
 
-serve: clean site
+serve:  site
```
craigbox: I think there is one more space here than there used to be and that might cause a problem

scripts/gen_site.sh
```diff
@@ -18,36 +18,30 @@ set -ex
 
 mkdir -p generated/js generated/img tmp/js
 
+
```
craigbox: unneeded?

I'm also trying to figure out why we have the tmp/js dir.  Is that because tsc compiles to js in that folder?  Is that default behaviour? 

It also looks like esbuild can build typescript files itself, so maybe we can drop this step and just use esbuild to build?

src/ts/entrypoint.js
```diff
@@ -0,0 +1,19 @@
+import "../../tmp/js/constants.js";
```
craigbox: a. should this be .ts?
b. Could these just be imports of the ts files without a path?

scripts/gen_site.sh
Ajay-singh1: Yes tsc compiles .ts files in tmp/js folder you can check ```tsconfig.json``` file. esbuild can transpile the typescript files but don't do type-checking for that we have to rely on typescript compiler.

src/ts/entrypoint.js
Ajay-singh1: No this shouldn't be typescript files they have to be javascript files because we are bundling and minifying javascript files.

craigbox:
In this case the entrypoint file should be generated by the build process IMO.

It's just a listing of all the files to compile. Better we don't use one at all?

Ajay-singh1:
Yes you are correct these are for compilation but having an entrypoint.js file seems like a good idea as Daniel mentioned it would be convenient for future maintainers and contributors to look into this file and get the idea of how things are working.

Ajay-singh1:
@dhawton @craigbox Can we please merge this PR if everything looks good?

keithmattix (approved):
LGTM

craigbox:
I'm still looking for a change in the entrypoint file.  I had assumed it would contain links to the existing scripts, but it seems it has to have links to `../tmp/` to work.  Right now, someone will see it, it has links to files that don't exist until the build time, and you have to dig through the build script to figure out why.

Given that I think it's best that it is generated when the build is done and removed afterwards.

Ajay-singh1:
I'll look into this and update you.

src/ts/entrypoint.js
dhawton: Does esbuild not transpile typescript to javascript? How does someone who come to the project know where the actual files are? I feel like we should be importing the ts and letting the transpiler bundle and convert for us. This feels like we're adding a special magic step somewhere.

scripts/gen_site.sh
```diff
@@ -20,34 +20,27 @@ mkdir -p generated/js generated/img tmp/js
 
 tsc
```
dhawton:
```suggestion
```

scripts/gen_site.sh
```diff
@@ -20,34 +20,27 @@ mkdir -p generated/js generated/img tmp/js
...
+  --sourcemap \
+  --target=es6 \
+  --outfile=generated/js/headerAnimation.min.js
+
+esbuild tmp/js/themes_init.js \
+  --bundle \
+  --minify \
+  --sourcemap \
+  --target=es6 \
+  --outfile=generated/js/themes_init.min.js
```
dhawton:
```suggestion
esbuild src/ts/entrypoint.ts \
  --bundle \
  --minify \
  --sourcemap \
  --target=es6 \
  --outfile=generated/js/all.min.js

esbuild src/ts/headerAnimation.js \
  --minify \
  --sourcemap \
  --target=es6 \
  --outfile=generated/js/headerAnimation.min.js

esbuild src/ts/themes_init.ts \
  --bundle \
  --minify \
  --sourcemap \
  --target=es6 \
  --outfile=generated/js/themes_init.min.js
```

esbuild doesn't need you to use tsc, so let's not to make it more readable and maintainable.

src/ts/entrypoint.js
```diff
@@ -0,0 +1,19 @@
...
+import "../../tmp/js/prism.js";
+import "../../tmp/js/codeBlocks.js";
+import "../../tmp/js/links.js";
+import "../../tmp/js/resizeObserver.js";
+import "../../tmp/js/scroll.js";
+import "../../tmp/js/overlays.js";
+import "../../tmp/js/lang.js";
+import "../../tmp/js/callToAction.js";
+import "../../tmp/js/events.js";
+import "../../tmp/js/faq.js";
```
dhawton:
```suggestion
import "./constants.ts";
import "./utils.ts";
import "./feedback.ts";
import "./kbdnav.ts";
import "./themes.ts";
import "./menu.ts";
import "./header.ts";
import "./sidebar.ts";
import "./tabset.ts";
import "./prism.js";
import "./codeBlocks.ts";
import "./links.ts";
import "./resizeObserver.js";
import "./scroll.ts";
import "./overlays.ts";
import "./lang.ts";
import "./callToAction.ts";
import "./events.ts";
import "./faq.ts";
```

Ajay-singh1:
@dhawton Thanks! I'll push the changes ASAP.

PR #16579, opened 2025-06-13
chore: fix npm warnings

### LINKED ISSUE #16573: fix npm warnings in site build
Building on #15059 

```
12:29:40 AM: npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
12:29:40 AM: npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
12:29:40 AM: npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
12:29:40 AM: npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
12:29:42 AM: added 31 packages, removed 72 packages, and changed 238 packages in 9s
12:29:42 AM: 28 packages are looking for funding
12:29:42 AM:   run `npm fund` for details
12:29:43 AM: up to date, audited 1 package in 897ms
12:29:43 AM: found 0 vulnerabilities
12:29:45 AM: added 33 packages, and audited 34 packages in 2s
12:29:45 AM: 1 package is looking for funding
12:29:45 AM:   run `npm fund` for details
12:29:45 AM: found 0 vulnerabilities
```

@AritraDey-Dev might like to look at this one.

AritraDey-Dev:
## Description
This PR resolves npm warnings caused by deprecated glob@7.2.3 and inflight@1.0.6 by overriding them.
few observations:
### Before 
```
$ npm explain inflight glob
inflight@1.0.6
node_modules/inflight
  inflight@"^1.0.4" from glob@7.2.3
  node_modules/glob
    glob@"^7.2.0" from @babel/cli@7.23.0
    node_modules/@babel/cli
      @babel/cli@"^7.23.0" from the root project
    glob@"^7.2.0" from svgstore-cli@2.0.1
    node_modules/svgstore-cli
      svgstore-cli@"^2.0.1" from the root project

glob@7.2.3
node_modules/glob
  glob@"^7.2.0" from @babel/cli@7.23.0
  node_modules/@babel/cli
    @babel/cli@"^7.23.0" from the root project
  glob@"^7.2.0" from svgstore-cli@2.0.1
  node_modules/svgstore-cli
    svgstore-cli@"^2.0.1" from the root project
```

 ### Now (After  the fix)
 ```
$  npm explain inflight glob
glob@10.4.5
node_modules/glob
  overridden glob@"^10.0.0" (was "^7.2.0") from @babel/cli@7.27.2
  node_modules/@babel/cli
    @babel/cli@"^7.27.2" from the root project
  overridden glob@"^10.0.0" (was "^7.2.0") from svgstore-cli@2.0.1
  node_modules/svgstore-cli
    svgstore-cli@"^2.0.1" from the root project
```
 
 

Fixes #16573

## Reviewers

- [ ] Ambient
- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -138,9 +138,11 @@ archive-version:
 # to what is included in the tools repo in docker/build-tools/Dockerfile.
 netlify_install:
 	@npm init -y
+	@jq '. + {overrides: {"glob": "^10.0.0", "inflight": false}}' package.json > tmp.$$ && mv tmp.$$ package.json
 	@npm install --omit=dev --global \
 	    sass@v1.89.1 \
-	    typescript@v5.8.3 \
+	    typescript@v5.8.3 
+	@npm install --omit=dev \
 	    svgstore-cli@v2.0.1 \
 		@babel/core@v7.27.4 \
 		@babel/cli@v7.27.2 \

### CODE REVIEW

**Comment by dhawton** [comment]
Why not place the override directly in packages.json? I don't like having things be different in tests vs deployment.. they should be the same.

**Comment by AritraDey-Dev** [comment]
> Why not place the override directly in packages.json? I don't like having things be different in tests vs deployment.. they should be the same.

Yes, my first thought was to do that in package.json itself. But there is no package.json in the project — it is created by the Makefile.core.mk  here
https://github.com/istio/istio.io/blob/95c8d11507e36712bb7e63ae988d27bc04c01f96/Makefile.core.mk#L140

So if we need to make changes to package.json, we first need to create it and then apply the override.

**Comment by dhawton** [comment]
Might be worth changing... I don't like the thought of having production be different from dev. I'll defer to @craigbox here though as there are a number of plans he has for docs that I am not in the loop on.

**Comment by AritraDey-Dev** [comment]
yes i agree it’s better to have a consistent package.json checked into the project so it don’t have differences between dev and prod.

**Comment by craigbox** [comment]
If we need the files to exist, then we should do it the idiomatic way.
Could we then have the Dockerfile include the dependencies from the packages.json file instead?

(It might be worth a quick look through the code history to see if we can figure out why it wasn't done like that from the start)

**Comment by AritraDey-Dev** [comment]
After looking at the previous commit logs, it seems that [PR #3242](https://github.com/istio/istio.io/pull/3242) removed package.json, and then they used npm install separately for each package.
Later, the Docker build process was migrated to the istio/tools repository via [PR #4722](https://github.com/istio/istio.io/pull/4722).

I think the best approach would be to add the overrides in the Dockerfile of the tools repository as well — specifically here:
https://github.com/istio/tools/blob/97af7d3c6f6057a41dab45e8780a6d3d56820f15/docker/build-tools/Dockerfile#L463-L479
By adding the following line:
```
@jq '. + {overrides: {"glob": "^10.0.0", "inflight": false}}' package.json > tmp.$$ && mv tmp.$$ package.json
```
This would make the behavior consistent across both dev and prod environments, addressing @dhawton's concern.
 @craigbox Would you agree that this is the right step to take next?

**Comment by craigbox** [comment]
It looks like Netlify used to use the website builder container?
Don't lean into the history too much here.  We do things how we do them today.  It may be that the change Martin made back then was probably to get the most recent versions of each package installed each time.  We don't want that, and I don't see a compelling reason not to use packages.json at this point.

The principle of simplicity for future maintainers should apply - I don't think that a JQ line and temporary generated files achieves that.

(An even more idiomatic answer might be [hugo mod npm pack](https://gohugo.io/commands/hugo_mod_npm_pack/) and [packages.hugo.json](https://discourse.gohugo.io/t/how-should-i-use-package-hugo-json/43055/3) but I think that's for a later date)

**Comment by craigbox** [comment]
(Daniel had a similar thought [here](https://github.com/istio/istio.io/pull/16574#discussion_r2154812818).)

**Comment by AritraDey-Dev** [comment]
Hmm, your point makes sense—having a package.json is a good idea. In the future, other packages might also require overrides, so I'll add those files.

File: package.json [hunk 1]
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "work",
```
AritraDey-Dev, line 2: Since the package.json is generated by` npm init -y`, the name is still set to "work". Should we change it to "istio.io", or is it okay to keep it as is?

Ajay-singh1:
Thanks for your effort, but I wanted to clarify that the current pipeline is already clean after PR #16574 where we added esbuild as a bundler & minifier.Their are no existing errors and warning everything is neat and clean.

Suppressing or overriding warnings globally could mask future issues and is not an ideal long term solution.Let's focus on meaningful improvement that genuinely help the project.

Thanks.

AritraDey-Dev:
I believe both issues are different. This PR attempts to address the warnings caused by deprecated glob and inflight.
 I suggest running `make netlify `on your PR branch
— you will see the warnings related to glob and inflight,
```
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
````
 which are also mentioned in issue #16573. 
So, even if we switch to `esbuild`, these warnings will still persist unless we use some versions of glob or inflight that do not trigger the deprecation warnings.

**Comment by Ajay-singh1** [comment]
No need to override , these are transitive dependency and the warning is caused by the deprecated svgstore-cli npm package you can replace it with modern npm packages.

See:-
Screenshot from 2025-06-20 07-55-31

**Comment by AritraDey-Dev** [comment]
In your previous comment, you claimed that your PR fixes this issue too — I think it's clear now that it doesn't.
Regarding the overrides: sure, switching to other packages might be better, but that’s something we should confirm with the maintainers first. And in this PR  if you check the earlier comments, the idea was to keep overrides in package.json and so followed that approach

**Comment by Ajay-singh1** [comment]
Sure! I'm not in favour of suppressing this warnings also overriding npm warnings especially for deprecated or unmaintained dependency - can hide known vulnerabilities (CVEs).

I would also like to escalate to maintainers for the best path forward.

File: Makefile.core.mk [hunk 2]
```diff
@@ -140,16 +140,7 @@ netlify_install:
...
-	    svgstore-cli@v2.0.1 \
-		@babel/core@v7.27.4 \
-		@babel/cli@v7.27.2 \
-		@babel/traverse@7.25.9 \
-		@babel/preset-env@v7.27.2
-	@npm install --omit=dev --save-dev \
-		babel-preset-minify@v0.5.2
-	@npm install --save \
-		core-js@3.42.0
+	    typescript@v5.8.3
```
craigbox, line 143: Why would these not be in `packages.json`?

Makefile.core.mk
```diff
@@ -140,16 +140,7 @@ netlify_install:
 	@npm init -y
```
craigbox, line 140: would this still be needed?

package.json
craigbox: I would change it

package.json
```diff
@@ -0,0 +1,38 @@
...
+  "version": "1.0.0",
+  "description": "| Site | Status |------|------- | istio.io | [![Netlify Status](https://api.netlify.com/api/v1/badges/c98435af-5464-4ac3-93c2-9c98faeec9b6/deploy-status)](https://app.netlify.com/sites/istio/deploys) | preliminary.istio.io | [![Netlify Status](https://api.netlify.com/api/v1/badges/a1cfd435-23d5-4a43-ac6d-8ec9230d9eb3/deploy-status)](https://app.netlify.com/sites/preliminary-istio/deploys) | archive.istio.io | [![Netlify Status](https://api.netlify.com/api/v1/badges/f8c3eecb-3c5c-48d9-b952-54c7ed0ece8f/deploy-status)](https://app.netlify.com/sites/archive-istio/deploys)",
```
craigbox, line 4: ...and this

package.json
```diff
@@ -0,0 +1,38 @@
...
+  "main": "index.js",
+  "directories": {
+    "test": "tests"
```
craigbox, line 7: we don't have any of these?

package.json
```diff
@@ -0,0 +1,38 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/istio/istio.io.git"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
craigbox, line 18: Apache 2.0

package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "dependencies": {
+    "@babel/cli": "^7.27.2",
+    "@babel/core": "^7.27.4",
+    "@babel/preset-env": "^7.27.2",
+    "@babel/traverse": "^7.25.9",
+    "core-js": "^3.42.0",
+    "svgstore-cli": "^2.0.1"
+  },
+  "devDependencies": {
```
craigbox, line 35: why is this a devDependency and not a dependency, do you know?

craigbox:
> Sure! I'm not in favour of suppressing this warnings also overriding npm warnings especially for deprecated or unmaintained dependency - can hide known vulnerabilities (CVEs).
> 
> I would also like to escalate to maintainers for the best path forward.

We are pinning the version of the svgstore package so it's reasonable to also pre-acknowledge the warnings at that version.  I'm happy to update to something better supported if we can find something (discussed in the other PR).

Makefile.core.mk
AritraDey-Dev: Because these packages need to be installed globally, we can either include them in the Makefile as we do now, or add them to the scripts section of package.json.

Makefile.core.mk
AritraDey-Dev: Just kept it for now as netlify still depends on it...can be replaced by `npm install` once package.json Is created

package.json
AritraDey-Dev: Entire package.json is generated by `npm init -y` .so I didn't touch it yet.

package.json
AritraDey-Dev: https://github.com/istio/istio.io/pull/16579#discussion_r2163080718
Same for this also

package.json
AritraDey-Dev: https://github.com/istio/istio.io/blob/b5c0c98cc3da557d751a01db3da244b4bda6ef83/Makefile.core.mk#L149
the `--save dev` flag install it as a devDependencies.
And this dependencies are required for build only not for runtime,so often used as devDependencies.
can be put inside dependencies section,but for sure keeping it in devDependencies is preferred always

AritraDey-Dev:
Once the esbuild PR is merged, and since we have already migrated from svgstore to svg-sprite-generator, do we still need the package.json file for the project?
If not, I believe this can be closed?

PR #16598, opened 2025-06-22
migrate svgstore to svg-sprite-generator

AritraDey-Dev:
## Description

Instead of applying overrides as done in #16579, it's better to migrate to npm dependencies that don't rely on deprecated versions.
This PR replaces `svgstore-cli` with the `svg-sprite-generate` package, which resolves the related deprecation warnings.

Additionally, there are deprecation warnings due to babel-cli, which @Ajay-singh1 is currently addressing in #16574.
So, while this PR removes deprecated dependencies introduced by svgstore-cli, #16574 will handle those caused by babel based build.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [x] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/scripts/gen_site.sh
+++ b/scripts/gen_site.sh
@@ -50,4 +50,4 @@ babel --source-maps --minified --no-comments \
   tmp/js/themes_init.js \
   --out-file generated/js/themes_init.min.js
 
-svg-symbol-sprite -i src/icons -o generated/img/icons.svg
+svg-symbol-sprite -i src/icons -o generated/img/icons.svg --prefix ""


craigbox:
This looks like it was last touched 10 years ago: 
https://www.npmjs.com/package/svg-sprite-generator

I'm not too keen on replacing something with something _less_ well-maintained.  Nothing better immediately jumps out at me - https://github.com/Dynation/tiny-isprite is modern but very new.

AritraDey-Dev:
Hmm, I’ve actually tried a bunch of npm packages that offer similar functionality to svgstore, including the one you suggested. However, `tiny-isprite` doesn't seem to be working properly — something appears to be broken with it.

https://github.com/scriptex/svg-symbol-sprite
https://www.npmjs.com/package/svg-symbol-sprite
 this looks actively maintained and fairly recent. It also don't  use glob or inflight, might be a good choice. Would love to hear your thoughts!

craigbox:
If it produces directly comparable output, let's give it a go

craigbox:
Can you please share the before and after sprites in a GitHub gist perhaps?
We would need a change to the build container to get this to pass testing, and based on previous discussion we should be sure we want to do this before we progress.

AritraDey-Dev:
gist - https://gist.github.com/AritraDey-Dev/b20da613b23595d69fb60b7f161b680c

PR for tools repo -  https://github.com/istio/tools/pull/3234

Ajay-singh1:
Good Job! AritraDey.😊🤝

Ajay-singh1:
/ test lint

Ajay-singh1:
This isn't working somehow.

The Difference:-

On Preliminary we have lost the SVGs.

Screenshot from 2025-06-26 19-05-01

Main Site:-

Screenshot from 2025-06-26 19-05-20

craigbox:
Look at https://deploy-preview-16598--preliminary-istio.netlify.app/latest/img/icons.svg#tick

It contains

```
<symbol xmlns="http://www.w3.org/2000/svg" id="icon-tick" viewBox="0 0 17.217 12.521"><defs><style>.a442__a{fill:none;stroke:#fff;stroke-width:3px}</style></defs><path d="m1.061 4.644 6.14 6.14M5.766 11.461l10.39-10.4" class="a442__a"/></symbol>
```

Note the `id` is `icon-tick`.

The id is `tick` as expected on istio.io.

AritraDey-Dev:
Thanks @craigbox for spotting this! I was searching for diffs to find where it was breaking, but never thought the issue was with the symbol ID prefix. Appreciate the help!

craigbox:
https://deploy-preview-16598--preliminary-istio.netlify.app/latest/docs/releases/contribute/shortcodes/ looks good so I'll use it as a proxy.

the file no longer has a 

```
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```

at the top, but that doesn't look to be a requirement (only the w3c validator cares about it.)

Issue #16584, opened 2025-06-17
Replace deprecated document.execCommand in utils.ts with modern Clipboard API


AritraDey-Dev:
The `copyToClipboard() `function in utils.ts currently uses the deprecated `document.execCommand('copy')` API, may not be supported in future browser versions.

https://github.com/istio/istio.io/blob/48a72cef5a4af08af122d932033a6daf5349c604/src/ts/utils.ts#L56

Replace `document.execCommand('copy')` with `navigator.clipboard.writeText(str)` for modern, secure clipboard handling.

gmarav05:
hello @AritraDey-Dev can you please assign this issue to me

gmarav05:
@craigbox can you please assign me this issue.

craigbox:
Hi @gmarav05,

We don't have a culture of asking to be assigned an issue.  Anyone is welcome to comment that they are willing to work on an issue, or to send a PR with a proposed fix.  If you want to make sure it's likely to be accepted you are welcome to discuss the proposed fix first.

Reading what @AritraDey-Dev wrote, it's a one line PR to make this change, and the author will have to show you have tested/verified it works.

gmarav05:
yeah thank you for the information craig i will make changes and verify it

PR #16604, opened 2025-06-24
Replaced document exec Command ("copy");  

gmarav05:
Replaced deprecated document.execCommand in utils.ts with modern Clipboard.

issue no #16584 

Replaced document.execCommand('copy') with navigator.clipboard.writeText(str) for modern, secure clipboard handling.

### DIFF
--- a/src/ts/utils.ts
+++ b/src/ts/utils.ts
@@ -45,7 +45,7 @@ function copyToClipboard(str: string): void {
         const oldSelection = sel.getRangeAt(0);
 
         el.select();                                     // Select the <textarea> content
-        document.execCommand("copy");                    // Copy - only works as a result of a user action (e.g. click events)
+        navigator.clipboard.writeText(str);              // Copy - only works as a result of a user action (e.g. click events)
         document.body.removeChild(el);                   // Remove the <textarea> element
 
         // restore the previous selection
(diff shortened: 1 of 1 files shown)


craigbox:
Please now show some evidence of before/after testing everywhere this code it called.

Tip: use the Netlify deploy preview link under Checks.

AritraDey-Dev (changes requested):
The issue might seem like it’s just about replacing that particular line, but if you’ve researched the implementation, you’ll notice there are other places in the function where document is used for selection and fallback logic. 
When switching to` navigator.clipboard.writeText`, you don’t need any of the document-based selection or textarea logic—so the whole function should be refactored, not just that single line

gmarav05:
should we refactor that for entire codebase? bro @AritraDey-Dev all tests are passing correctly?

src/ts/utils.ts
```diff
@@ -53,7 +53,7 @@ function copyToClipboard(str: string): void {
         sel.addRange(oldSelection);
     } else {
         el.select();                                     // Select the <textarea> content
-        document.execCommand("copy");                    // Copy - only works as a result of a user action (e.g. click events)
+        navigator.clipboard.writeText(str);                    // Copy - only works as a result of a user action (e.g. click events)
```
craigbox: The indenting no longer matches.

BHAVISHYA2005:
@gmarav05 ? Please do as mentioned

PR #16617, opened 2025-06-26
update destination rule docs to ensure consistent use of oneof tag

gilmoregrills:
## Description

I was reading through these docs earlier today, and all the previous sections tag attributes of objects where only one of them may be provided with little `(oneof)` tags, but this one didn't. The should just update the table that was missing these tags to include the tags, and also removes some now-redundant bits of the descriptions of those attributes.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/reference/config/networking/destination-rule/index.html
+++ b/content/en/docs/reference/config/networking/destination-rule/index.html
@@ -1710,32 +1710,30 @@ and similarly us-west should failover to us-east.</p>
 </tr>
 </thead>
 <tbody>
-<tr id="LocalityLoadBalancerSetting-distribute">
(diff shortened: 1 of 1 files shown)


gilmoregrills:
> From the file:
> 
> ```
> WARNING: THIS IS AN AUTO-GENERATED FILE, DO NOT EDIT. PLEASE MODIFY THE ORIGINAL SOURCE IN THE 'https://github.com/istio/api' REPO
> ```
> 
> You'll need to have it changed in the istio/api repo.

Loool my bad 😅 will do

PR #16662, opened 2025-07-14
cherry-pick baseURL fix to istio.io

craigbox:
Manual cherry-pick of #16568.


craigbox:
https://github.com/istio/istio.io/pull/16568#issuecomment-3067534392 shows that the PR worked on preliminary, which uses the same context variable as istio.io (with a different URL variable.)

craigbox:
I don't think I can make the PR ignore the master tests having failed, so I'll open another one.

PR #16663, opened 2025-07-14
cherry-pick baseURL fix to istio.io

craigbox:
Manual cherry-pick of #16568.

https://github.com/istio/istio.io/pull/16568#issuecomment-3067534392 shows that the PR worked on preliminary, which uses the same context variable as istio.io (with a different URL variable.)

Second attempt; this time against the right base!


Ajay-singh1:
Thanks! Craig.

dhawton (approved):

PR #16664, opened 2025-07-14
set baseURL

craigbox:


craigbox:
Pros: `hugo serve` will work
Cons: given we've overridden everything in the `Makefile`, I don't see this breaking anything

dhawton (approved):
I don't see a path where it's not set, unless someone manually runs `hugo serve` outside of make.

PR #16680, opened 2025-07-17
make /latest the latest

craigbox:
Changing the baseURL from `/latest` to `https://istio.io/latest/` caused two related problems:

1. redirects stopped working
2. search stopped working because it relied on redirects

The latter is fixed by way of updating `/search` to `/latest/search`. For now at least, I assume we are going to only have one search index.

For the former, I have removed a redirect for every. single. page. (~2000 lines) and replaced with a wildcard redirect for the five paths we care about.

This only fixes links that don't include a version, which should only be links that were broken (like point 2), or older than 5 years, when we introduced `/latest/` to all the URLs.

### DIFF
--- a/content/en/_index.html
+++ b/content/en/_index.html
@@ -21,7 +21,7 @@ description: A service mesh for observability, security in depth, and management
         "url": "https://istio.io/",
         "potentialAction": {
             "@type": "SearchAction",
-            "target": "https://istio.io/search?q={search_term_string}",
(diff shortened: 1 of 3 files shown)


dhawton (approved):

craigbox:
click 🔍 on https://deploy-preview-16680--preliminary-istio.netlify.app: sent to /latest/search?q=test
go to https://deploy-preview-16680--preliminary-istio.netlify.app/news/: sent to /latest/news/

Issue #15689, opened 2024-09-14
tabset partial breaks mirroring task page


find-arka:
Starting from [here](https://istio.io/latest/docs/tasks/traffic-management/mirroring/#creating-a-default-routing-policy) the yamls have incorrect indentation currently. Copying and pasting the same leads to `error converting YAML to JSON` error messages.

![Screenshot 2024-09-13 at 10 02 42 PM](https://github.com/user-attachments/assets/0eccdfbf-965b-4eb8-a0ad-1e0c9b8dee3e)

## Istio APIs updated

Section 1

```bash
kubectl apply -f - <<EOF
apiVersion: networking.istio.io/v1
kind: VirtualService
metadata:
  name: httpbin
spec:
  hosts:
  - httpbin
  http:
  - route:
    - destination:
        host: httpbin
        subset: v1
      weight: 100
---
apiVersion: networking.istio.io/v1
kind: DestinationRule
metadata:
  name: httpbin
spec:
  host: httpbin
  subsets:
  - name: v1
    labels:
      version: v1
  - name: v2
    labels:
      version: v2
EOF
```

Section 2

```bash
kubectl apply -f - <<EOF
apiVersion: networking.istio.io/v1
kind: VirtualService
metadata:
  name: httpbin
spec:
  hosts:
  - httpbin
  http:
  - route:
    - destination:
        host: httpbin
        subset: v1
      weight: 100
    mirror:
      host: httpbin
      subset: v2
    mirrorPercentage:
      value: 100.0
EOF
```

## Gateway API updated

Section 1

```bash
kubectl apply -f - <<EOF
apiVersion: v1
kind: Service
metadata:
  name: httpbin-v1
spec:
  ports:
  - port: 80
    name: http
  selector:
    app: httpbin
    version: v1
---
apiVersion: v1
kind: Service
metadata:
  name: httpbin-v2
spec:
  ports:
  - port: 80
    name: http
  selector:
    app: httpbin
    version: v2
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: httpbin
spec:
  parentRefs:
  - group: ""
    kind: Service
    name: httpbin
    port: 8000
  rules:
  - backendRefs:
    - name: httpbin-v1
      port: 80
EOF
```

Section 2

```bash
kubectl apply -f - <<EOF
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: httpbin
spec:
[...]

PR #16688, opened 2025-07-17
Fixed Identation for faster access

ccabello1:
I notice that some example has wrong identation resulting in error if you copy paste

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


Arhell (commented):
/ok-to-test

craigbox:
Thanks for looking at this.  The issue isn't how the code is written in the page, which is what you have changed, it's how it is rendered.  See #15689 for the error.

Changing the code will break the tests, so for now, until we work on the actual rendering error, we can't do that.

PR #16727, opened 2025-07-31
prepare for v1.27 as istio source is already branched

zirain:


dhawton:
There's errors in the scripts, which is why docs maintainers typically run these.. I'll work on this since there seems to be broken scripts again, and have a PR update today or tomorrow.

zirain:
> There's errors in the scripts, which is why docs maintainers typically run these.. I'll work on this since there seems to be broken scripts again, and have a PR update today or tomorrow.

Thanks for helping.

PR #16746, opened 2025-08-03
feat: Add Version selector dropdown

Ajay-singh1:
## Description

This PR adds a version selector drop down for Istio documentation which allows user to navigate to older docs.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


Ajay-singh1:
@craigbox After this PR is merged should I go on and regenerate the archive docs from v1.25-v1.22 with the banner.

craigbox:
Quick thoughts from https://deploy-preview-16746--preliminary-istio.netlify.app/:

* this should only appear on Docs pages
* needs styling like the rest of the heading
* I think we settled on "latest" (preliminary), "stable" (current version) and n-5 releases
* "Try Istio" should remain top-right, so we will have to put this to the left of that
* need a responsive version of it

Screenshot 2025-08-04 at 4 10 31 PM

Ajay-singh1:
Took note of the suggestion, working on it.

Ajay-singh1:
Good to Go!

layouts/partials/header.html
```diff
@@ -23,7 +23,26 @@
                 {{ $currentPage := .}}
                 {{ range .Site.Menus.main }}
                     {{ $active := or (eq $currentPage.Title .Name) (or ($currentPage.HasMenuCurrent "main" .) ($currentPage.IsMenuCurrent "main" .)) }}
-                    {{ if .HasChildren }}
+                    {{ if and (eq .Identifier "docs") (eq $.Section "docs") (eq hugo.Environment "production") }}
```
craigbox: "production" means "not archive" in this case?
I thought that we did production builds for all sites and "development" for localhost, or something

layouts/partials/header.html
```diff
@@ -23,7 +23,26 @@
...
+
+                    <ul class="main-navigation-links-dropdown">
+                        {{ range .Children }}
+                        <li class="main-navigation-links-dropdown-item">
+                            <a href="{{ .URL | relLangURL }}" class="main-navigation-links-link">{{ .Name }}</a>
+                        </li>
+                        {{ end }}
+
+
+                        {{ partial "version-selector.html" $ }}
```
craigbox, line 41: what does the $ mean?

layouts/partials/version-selector.html
```diff
@@ -0,0 +1,42 @@
+{{ if and (eq .Section "docs") (eq hugo.Environment "production") }}
```
craigbox: you already had an "if" like this on the page that spawned this?

layouts/partials/version-selector.html
```diff
@@ -0,0 +1,42 @@
...
+{{ $versions := site.Data.versions }}
+{{ $mainVerStr := $versions.main }}
+{{ $preliminaryVerStr := $versions.preliminary }}
+{{ $mainVerParts := split $mainVerStr "." }}
+{{ $major := index $mainVerParts 0 | int }}
+{{ $minor := index $mainVerParts 1 | int }}
+{{ $versionCount := 5 }}
+
+<!-- Preliminary version -->
+{{ $preUrl := printf "https://preliminary.istio.io/v%s/docs" $preliminaryVerStr }}
```
craigbox, line 10: Do we have these URLs in vars somewhere?

layouts/partials/version-selector.html
```diff
@@ -0,0 +1,42 @@
...
+<li class="main-navigation-links-dropdown-item">
+  <a href="{{ $preUrl }}"
+     class="main-navigation-links-link {{ if in $.Permalink "/preliminary/" }}active{{ end }}">
```
craigbox: I don't think any pages have `/preliminary/` in their permalink

scripts/archive_version.sh
```diff
@@ -23,6 +23,8 @@ NUMBER=$(grep -w 'version:' data/args.yml | grep -oE '[^ ]+$' | tr -d '"')
 VERSION+="${NUMBER}"
 
 rm -rf public/
+#set the Hugo environment to development for preventing production code in local builds
```
craigbox: Ah.  We probably don't want that.  Look at `data/args.yml` for `archive: true` etc

layouts/partials/header.html
Ajay-singh1: In hugo when a command is executed through ```hugo``` hugo sets the environment to ```production``` directly.When ```hugo serve``` is executed hugo sets the environment to development.

The live site and preliminary sites directly execute ```hugo --minify``` so the environment is set to ```production```.

When you make a release the bash script ```archive-version.sh``` is called and that script calls the build script ```build_site.sh``` and that runs ```hugo --minify``` so whether you build the site locally it will act like it is in production mode.

So therefore this logic.

layouts/partials/header.html
Ajay-singh1: Provides top-level page context when you are looping through menu items etc it is needed.

layouts/partials/version-selector.html
Ajay-singh1: Yeah yeah it is necessary if you will make an archive the drop-down wont be shown because of this condition.

But I think we can make use of the ```data/args.yml``` that you suggested , much better.

layouts/partials/version-selector.html
Ajay-singh1: These are dynamically generated from the data/versions.yml file.

scripts/archive_version.sh
Ajay-singh1: Yeah! Thanks. This makes job much easier.

layouts/partials/version-selector.html
Ajay-singh1: Will remove this!

layouts/partials/version-selector.html
craigbox: Yeah, the versions are, I meant the string "preliminary.istio.io".  It isn't in a variable anywhere, I've just checked how we do it elsewhere.

layouts/partials/version-selector.html
craigbox: the right thing to do is have the partial always draw the dropdown, as that's its purpose, but only include the dropdown if we want it.

layouts/partials/header.html
craigbox: ack

layouts/partials/header.html
craigbox: See below, the right thing to do is check for `archive` being true rather than the environment

Ajay-singh1: Yes I agree!

layouts/partials/version-selector.html
Ajay-singh1: Yes Thanks!

Ajay-singh1:
Thanks a lot @craigbox for the review really helped a lot.I have addressed all your comments.

layouts/partials/version-selector.html
```diff
@@ -0,0 +1,42 @@
+{{ if and (eq .Section "docs") (not $.Site.Data.args.archive) }}
```
craigbox: reminder: remove this logic from here and put it in the code that calls the partial

Ajay-singh1:
Ping: @craigbox

layouts/partials/version-selector.html
craigbox: still waiting on this one fix

craigbox:
```suggestion
```

layouts/partials/version-selector.html
```diff
@@ -0,0 +1,42 @@
...
+    {{ $url := printf "https://istio.io/%s/docs" $ver }}
+    <li class="main-navigation-links-dropdown-item">
+      <a href="{{ $url }}"
+         class="main-navigation-links-link">
+        {{ $ver }}
+      </a>
+    </li>
+  {{ end }}
+{{ end }}
+{{ end }}
```
craigbox:
```suggestion
```

Ajay-singh1:
Done!

PR #16749, opened 2025-08-04
Add note about istio-cni security

keithmattix:
## Description

Add note about Istio CNI's security implications in ambient mode

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/en/docs/ops/deployment/security-model/index.md
```diff
@@ -40,6 +40,10 @@ The implications of this are discussed [below](#node-compromise).
 Because this consolidates the elevated privileges required to setup networking into a single pod, rather than *every* pod,
 this option is generally recommended.
 
+##### Ambient Mode
+
+In ambient mode, the Istio CNI plugin (and the associated node agent) manages mesh enrollment for pods living on its node. Due to limitations in the Kubernetes API, it is not currently possible for the CNI plugin or its node agent to prevent pods from being scheduled on the node before the CNI plugin is installed and configured. In these rare cases (e.g. on node restart or new node scale out), it is possible that a pod that is labeled for mesh enrollment may come up before the CNI's traffic redirection rules are applied, meaning that policies won't be enforced until after the CNI comes up and that pod is restarted. The Istio community is working with the upstream Kubernetes community to address this limitation, but in the meantime, you can [configure an initcontainer](TODO) enabled [owned CNI mode](TODO) to mitigate these race conditions.
```
craigbox: TODO: link these todos

keithmattix: Yeah I imagine we can link to some pilot env vars or helm values once @jaellio finishes up implementation

linsun: Do you think we should link to our docs here and mention that this can still occur during node cordon? https://istio.io/latest/docs/ambient/upgrade/helm/#understanding-ambient-mode-upgrades 

At least that was one of the first thoughts i had.

linsun: Also, should we mention if the init-container approach requires one time redeploy of the pod as pod is added to ambient but no need to restart the pod when upgrading to newer Istio ambient (ztunnel, cni, control plane)?

linsun:
Left 2 comments that may be worthwhile pointing out. Otherwise LGTM.

content/en/docs/ops/deployment/security-model/index.md
keithmattix: +1 on saying node cordon doesn't help with restarts. I figure the links for initcontainer could talk about the one-time redeploy; WDYT?

linsun: Re redeploy - yep that works as long as it is documented.

keithmattix:
@craigbox @linsun Updated the doc; PTAL when you get a chance

content/en/docs/ops/deployment/security-model/index.md
```diff
@@ -40,6 +40,10 @@ The implications of this are discussed [below](#node-compromise).
 Because this consolidates the elevated privileges required to setup networking into a single pod, rather than *every* pod,
 this option is generally recommended.
 
+#### Ambient Mode
```
craigbox:
```suggestion
#### Ambient mode
```

content/en/docs/ops/deployment/security-model/index.md
```diff
@@ -40,6 +40,10 @@ The implications of this are discussed [below](#node-compromise).
...
+
+In ambient mode, the Istio CNI plugin (and the associated node agent) manages mesh enrollment for pods living on its node. Due to limitations in the Kubernetes API, it is not currently possible for the CNI plugin or its node agent to prevent pods from being scheduled on the node before the CNI plugin is installed and configured. This can occur even if using node cordons + taints as described [in our upgrade documentation](/docs/ambient/upgrade/helm#cni-node-agent). In these rare cases (e.g. on node restart or new node scale out), it is possible that a pod that is labeled for mesh enrollment may come up before the CNI's traffic redirection rules are applied, meaning that policies won't be enforced until after the CNI comes up and that pod is restarted. The Istio community is working with [various](https://github.com/containernetworking/cni/pull/1052) [upstream](https://github.com/kubernetes/kubernetes/issues/130594) communities to address this limitation, but in the meantime, you can enable [owned CNI mode](https://github.com/jaellio/istio/blob/master/releasenotes/notes/55968.yaml) to mitigate these race conditions.
```
craigbox, line 45:
```suggestion
In ambient mode, the Istio CNI plugin (and the associated node agent) manages mesh enrollment for pods living on its node. 

Due to limitations in the Kubernetes API, it is not currently possible for the CNI plugin or its node agent to prevent pods from being scheduled on the node before the CNI plugin is installed and configured. This can occur even if [using node cordons and taints](/docs/ambient/upgrade/helm#cni-node-agent). In rare cases (e.g. on node restart or new node scale-out), it is possible that a pod that is labeled for mesh enrollment may come up before the CNI's traffic redirection rules are applied, meaning that policies won't be enforced until after the CNI comes up and that pod is restarted. 

The Istio community is working with [the CNI](https://github.com/containernetworking/cni/pull/1052) and [Kubernetes communities](https://github.com/kubernetes/kubernetes/issues/130594) to address this limitation, but in the meantime, you can enable [owned CNI mode](https://github.com/istio/istio/blob/master/releasenotes/notes/55968.yaml) to mitigate these race conditions.
```

Reformatted a little for clarity.  I wouldn't have thought a pod restart was required anywhere along the way, but one is mentioned? (In the context of the node taint controller, @ilrudie told me that it was only a problem until the CNI agent claimed the pod)

craigbox: p.s. please fast-follow by documenting that owned CNI mode feature somewhere other than a relnote!

keithmattix: Yeah unfortunately a pod restart is required due to the nature of the bug 😓

keithmattix: Where should we document an env var based feature like this?

ilrudie: Would we really not handle all existing pods when the CNI starts? We used to be able to do that.

keithmattix: @jaellio or @howardjohn can correct me here - I may be wrong but I think because the pod already has the meshed annotation, we don't go In and repair

jaellio: There is an additional race condition with our current reconciliation logic. If a pod is ambient enabled and fully enrolled (annotated as "enabled"), on istio-cni agent start up it will attempt to reconcile (add inpod iptable redirection rules) to all fully enrolled pods. However, if at the time of reconciliation, the pod's net ns hasn't been created the reconciler will skip the pod and log an error. If the istio-cni binary hasn't been installed (which occurs after reconciliation) when the pods net ns is configured it will bypass the mesh.

The istio owned cni config doesn't entirely resolve this issue since on node reboot the file binary and config and inconsistently cleaned up.

jaellio: @keithmattix What are you thoughts on including this update in 1.27.1 after we address the additional bugs above and have an alternative solution?

jaellio: > Would we really not handle all existing pods when the CNI starts? We used to be able to do that.

@ilrudie (not positive this is what you meant, so correct me if I misunderstood :) )

Currently, our reconciliation on istio-cni happens before we install the istio cni. binary and config, so if a pod doesn't have a net ns when the snapshot is taken for reconciliation (we ignore it in this case) and has a net ns configured before istio-cni is installed then the pod will only be meshed after a restart. 

We could fix this by changing the ordering of reconciliation and install as discussed [here](https://istio.slack.com/archives/C049TCZMPCP/p1753229017168319)

keithmattix: > What are you thoughts on including this update in 1.27.1 after we address the additional bugs above and have an alternative solution?

This sounds good to me - I wasn't aware/didn't fully understand the reconciliation aspect.

craigbox: > Where should we document an env var based feature like this?

Somewhere in the ambient documentation relating to the CNI.
Perhaps https://istio.io/latest/docs/ambient/architecture/traffic-redirection/ for now?

We need to separate out the page under the "sidecar" section at some point

ilrudie: @jaellio, that makes sense. I thought we did reconcile after installing the binary but if we don't then I can totally see how we'd sometimes wind up in limbo like this.

Thanks for the explanation.

keithmattix:
@jaellio When you get a chance, can you let me know how/when to proceed with this

content/en/docs/ops/deployment/security-model/index.md
sridhargaddam: > it is not currently possible for the CNI plugin or its node agent to prevent pods from being scheduled on the node before the CNI plugin is installed and configured. This can occur even if using node cordons + taints as described [in our upgrade documentation]

The [CNI node agent upgrade section](https://istio.io/latest/docs/ambient/upgrade/helm/#cni-node-agent) notes that the agent includes mechanisms to prevent pod scheduling issues and steps to prevent unsecured traffic leakage. We may also want to add a brief note covering this edge case.

```
...and while being upgraded or restarted will prevent new pods from being started on that node until an instance of the agent is available on the node to handle them, in order to prevent unsecured traffic leakage.
```

sridhargaddam:
> meaning that policies won't be enforced until after the CNI comes up and that pod is restarted.

@keithmattix @jaellio If we run into such an issue, is there a straightforward way to identify the pods where traffic is being bypassed? For example, if we run `istioctl zc workloads`, would the affected pods show `HBONE` or `TCP`?

content/en/docs/ops/deployment/security-model/index.md
keithmattix: Good call; we should definitely update that since restarts are the Wild West

keithmattix:
@sridhargaddam it won't be based on config; it would be a lack of iptables rules in the pod netns. @jaellio can correct me if I'm wrong

sridhargaddam:
> @sridhargaddam it won't be based on config; it would be a lack of iptables rules in the pod netns. @jaellio can correct me if I'm wrong

It’d be useful to have a command that shows pods marked for the mesh but bypassing it (due to missing iptables rules). Maybe `istioctl analyze...` could be extended with an option to catch and report this situation?

keithmattix:
Hmm that could work if the person running the command had kubectl debug permission

sridhargaddam:
> it is possible that a pod that is labeled for mesh enrollment may come up before the CNI's traffic redirection rules are applied, meaning that policies won't be enforced until after the CNI comes up and that pod is restarted.

I don’t expect this issue to occur in sidecar mode with istio-cni (since the istio-validation container checks that traffic redirection rules are in place). Can someone please confirm?

keithmattix:
Correct, the istio-validation init container prevents this from happening in sidecar mode

sridhargaddam:
> Correct, the istio-validation init container prevents this from happening in sidecar mode

Cool, thanks for confirming, Keith.

PR #16761, opened 2025-08-11
add a workaround to traffic mirroring task page

craigbox:
Well-meaning people keep trying to fix the traffic mirroring page by changing the indentation.  The problem is actually in the tabset template, as explained in #15689.  This PR adds a warning to that effect, as well as instructions to users to go to the source MD file directly to copy the examples.

(This is the only page with this level of indenting as far as I could tell.)


dhawton (approved):

dhawton:
Added CP, I know 1.27.0 is "imminent", but let's get this on ASAP.

PR #16769, opened 2025-08-11
fix create_minor_version.sh and update README

zirain:
cc @dhawton @craigbox

### DIFF
--- a/scripts/create_minor_version.sh
+++ b/scripts/create_minor_version.sh
@@ -149,9 +149,9 @@ parse_versions() {
 
 step0() {
     echo -e "\nStep 0: Check and prepare the environment\n"
+    parse_input "$1"
     prepare_env
     echo "Current working directory: $(pwd)"
-    parse_input "$1"
     parse_versions
 }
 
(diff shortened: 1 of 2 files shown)


scripts/create_minor_version.sh
```diff
@@ -149,9 +149,9 @@ parse_versions() {
 
 step0() {
     echo -e "\nStep 0: Check and prepare the environment\n"
+    parse_input "$1"
```
zirain, line 152: to prevent:
```
+ echo 'DRY_RUN: _gitpush origin master'
DRY_RUN: _gitpush origin master
+ gh repo set-default ''
expected the "[HOST/]OWNER/REPO" format, got ""
+ cleanup_env
+ popd
```

craigbox: will `gh repo` accept `user/istio.io.git` as a repo format?

dhawton: This is already being done in #16770. And yes Craig, it will. Mentioned in the thread I was going to fix them in the master bump

PR #16837, opened 2025-09-02
[zh] Fix ambient mc verify doc link and localize arch image

wilsonwu:
## Description

Fix ambient mc verify doc link and localize arch image

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


content/zh/docs/ambient/install/multicluster/multi-primary_multi-network/arch.svg
zirain, line 1: do we really need this? I cannot image what a developer need such kind of translation.
we'd better ignore these task in the future, and save your time in other tasks.

PR #16843, opened 2025-09-03
spire: update injection template to customize native sidecars

jewertow:
## Description

The injection template for SPIRE does not work with Istio 1.27 that enables native sidecars by default.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [X] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/integrations/spire/index.md
+++ b/content/en/docs/ops/integrations/spire/index.md
@@ -171,7 +171,7 @@ Below are the equivalent manual registrations based off the automatic registrati
               labels:
                 spiffe.io/spire-managed-identity: "true"
               spec:
-                # NOTE: Use `containers` instead of `initContainers`
+                # NOTE: Use containers instead of initContainers
                 # if native sidecars are disabled or not supported in your Kubernetes cluster.
                 initContainers:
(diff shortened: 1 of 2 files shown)


prow/integ-suite-kind.sh
```diff
@@ -113,18 +113,18 @@ elif [[ "$IP_FAMILIES" =~ "IPv6" ]] && [[ "$IP_FAMILIES" =~ "IPv4" ]]; then
    KIND_IP_FAMILY="dual"
 fi
 export KIND_IP_FAMILY
-export NODE_IMAGE="gcr.io/istio-testing/kind-node:v1.31.0"
+export NODE_IMAGE=""
```
dhawton, line 116: Why is this being changed?

jewertow: Native sidecars are not enabled on KIND 1.31, so the integration tests for spire worked fine and didn't capture the issue with patching containers, instead of initContainers, so I upgraded the node image, but now other tests fail.

jewertow:
Closed in favor of #16866

PR #16844, opened 2025-09-03
spire: remove init container wait-for-spire-socket

jewertow:
## Description

I believe that the init container `wait-for-spire-socket` is an obsolete workaround, because Spire CSI driver requires the volume mount to be read-only and it also enforces it internally. I wasn't able to reproduce a scenario where the istio-agent created the socket on its own when spire-agent was not available, because it simply failed on write operation:

```
2025-09-03T10:12:04.958407Z     error   sds     SDS grpc server for workload proxies failed to set up UDS: failed to remove unix://var/run/secrets/workload-spiffe-uds/socket: remove var/run/secrets/workload-spiffe-uds/socket: read-only file system
2025-09-03T10:12:20.959464Z     warn    sds     SDS grpc server could not be started
```

So I think it does not make any sense to still keep this workaround in the documentation.

Alternatively, we might document the following configuration as a replacement:

```
kind: IstioOperator
metadata:
  namespace: istio-system
spec:
  profile: default
  meshConfig:
    trustDomain: example.org
    defaultConfig:
       proxyMetadata:
         WORKLOAD_IDENTITY_SOCKET_FILE: "spire-agent.sock"
  components:
    ingressGateways:
      - name: istio-ingressgateway
        enabled: true
        label:
          istio: ingressgateway
        k8s:
          overlays:
            - apiVersion: apps/v1
              kind: Deployment
              name: istio-ingressgateway
              patches:
                - path: spec.template.spec.volumes.[name:workload-socket]
                  value:
                    name: workload-socket
                    csi:
                      driver: "csi.spiffe.io"
                      readOnly: true
                - path: spec.template.spec.containers.[name:istio-proxy].volumeMounts.[name:workload-socket]
                  value:
                    name: workload-socket
                    mountPath: "/run/secrets/workload-spiffe-uds"
                    readOnly: true
```

The `WORKLOAD_IDENTITY_SOCKET_FILE` ensures that istio-agent will always wait for the SDS socket, instead of trying to create it on its own. That socket name is used by spire-agent by default, so users don't need to customize spire installation to make it work.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [X] Installation
- [ ] Networking
- [X] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ops/integrations/spire/index.md
+++ b/content/en/docs/ops/integrations/spire/index.md
@@ -208,27 +208,6 @@ Below are the equivalent manual registrations based off the automatic registrati
                         name: workload-socket
                         mountPath: "/run/secrets/workload-spiffe-uds"
                         readOnly: true
-                    - path: spec.template.spec.initContainers
-                      value:
-                        - name: wait-for-spire-socket
-                          image: busybox:1.36
-                          volumeMounts:
-                            - name: workload-socket
-                              mountPath: /run/secrets/workload-spiffe-uds
-                              readOnly: true
-                          env:
-                            - name: CHECK_FILE
-                              value: /run/secrets/workload-spiffe-uds/socket
-                          command:
-                            - sh
-                            - "-c"
-                            - |-
-                              echo "$(date -Iseconds)" Waiting for: ${CHECK_FILE}
-                              while [[ ! -e ${CHECK_FILE} ]] ; do
-                                echo "$(date -Iseconds)" File does not exist: ${CHECK_FILE}
-                                sleep 15
-                              done
-                              ls -l ${CHECK_FILE}
     EOF
     {{< /text >}}
 
--- a/content/en/docs/ops/integrations/spire/snips.sh
+++ b/content/en/docs/ops/integrations/spire/snips.sh
@@ -150,27 +150,6 @@ spec:
                     name: workload-socket
                     mountPath: "/run/secrets/workload-spiffe-uds"
                     readOnly: true
-                - path: spec.template.spec.initContainers
-                  value:
-                    - name: wait-for-spire-socket
-                      image: busybox:1.36
-                      volumeMounts:
-                        - name: workload-socket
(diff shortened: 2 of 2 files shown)


jewertow:
`make lint` is not idempotent, because it does not fail in my other PR, and the errors are not related to spire at all.

dhawton (approved):

jaellio:
@jewertow could you explain how we avoid the above failure when removing the init container? My understanding was if the socket hasn't been created, istio-agent will assume it should create its own SDS server. Does a user then need to redeploy the deployment to resolve this? If `WORKLOAD_IDENTITY_SOCKET_FILE` is required, we should document it.

jewertow:
Hey @jaellio, it took me a while to remember why I've removed that init container. So here is the context:

I was wondering why the init container `wait-for-spire-socket` is only needed for gateways, and I have found the original [issue](https://github.com/istio/istio/issues/41308) - it says that if Spire agent is installed after Istio, then the Istio agent will create the SDS socket first.

I spent some time experimenting with this issue to see the impact on sidecars and gateways, but I wasn't able to reproduce the situation where the SDS socket was created by the Istio agent, because of the errors which you can see in this PR description.

Then I thought that the reported issue is no longer valid and we can remove that init container to simplify this integration. However, now I think that I might misunderstood what the issue says in this sentence: "so this may cause the Istio agent to incorrectly create the socket if SPIRE is started after Istio" - I understood that the Istio agent created its own SDS socket, but maybe the author meant returning errors when the agent tries to create the socket... If that was the intention we may need to revert this removal.

Btw. the init container improves a pod lifecycle, because it will stay in `Initializing` state as long as the spire agent didn't create the socket. On the other hand, when the init container is not used, istio-proxy will stay in `Running`, but its readiness probe will fail. So in both cases istio-proxy will be isolated from handling network traffic.

> Does a user then need to redeploy the deployment to resolve this?

I was testing it again, and I see that the gateways eventually becomes ready when spire agent creates the socket.

PR #16847, opened 2025-09-04
Bump WorkloadEntry autoregistration to alpha

keithmattix:
## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/setup/install/virtual-machine/index.md
+++ b/content/en/docs/setup/install/virtual-machine/index.md
@@ -105,7 +105,7 @@ Install Istio and expose the control plane on cluster so that your virtual machi
 
     {{< tab name="Automated WorkloadEntry Creation" category-value="autoreg" >}}
 
-    {{< boilerplate experimental >}}
+    {{< boilerplate alpha >}}
 
     {{< text syntax=bash snip_id=install_istio >}}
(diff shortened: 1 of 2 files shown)


dhawton:
I noticed this is not listed on the Features list, which comes from here: https://github.com/istio/enhancements/blob/master/features.yaml -- is this feature being tracked elsewhere?

keithmattix:
Yeah I checked for it there too and didn't see anything. I suspect it's an env var only feature unfortunately. It's been on by default for 3 years though

dhawton:
> Yeah I checked for it there too and didn't see anything. I suspect it's an env var only feature unfortunately. It's been on by default for 3 years though

Wondering if we should fix that. Does this qualify, then, for alpha? Per https://istio.io/latest/docs/releases/feature-stages/, alpha features are off by default with pointers to issues and/or design docs.

I'll approve this, but we should probably look at bumping it to beta sooner than later to line up with the feature stages.

keithmattix:
Yes we discussed this in slack a bit. A flaky integ test is the only beta blocker

dhawton:
> Yes we discussed this in slack a bit. A flaky integ test is the only beta blocker

Perfect, thanks. Must have missed that slack convo.

PR #16882, opened 2025-09-19
document validations in validateHTTPRetry

ksraj123:
## Description

Validations enforced on HttpRetry here - https://github.com/istio/istio/blob/master/pkg/config/validation/validation.go#L2610 are not documented. When the retry attempts are set to 0 then these validations are enforced.

Please feel free to propose improvements.

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


dhawton (changes requested):
Apologies for this missed review.. we cannot edit this here, from the top of the file:

```
```

PR #16981, opened 2025-11-07
update tip for TopologyLocality label

zirain:
xref: https://github.com/istio/istio/pull/58186


dhawton (approved):
No CP: landed in master only (1.29)

PR #17034, opened 2025-12-11
Modify Kiali deployment commands and add Prometheus

yuvraj9:
## Description

Add Prometheus installation as a default step in the Istio Getting Started documentation. Kiali relies on Prometheus for metrics, so the dashboard does not work without it. This is especially important for kind clusters, where Prometheus must be installed explicitly.

## Reviewers

- [ ] Ambient
- [X] Docs
- [X] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


sridhargaddam (approved):
LGTM

PR #17043, opened 2025-12-18
Update metric expiry and cardinality docs to mention sidecar.istio.io/statsEvictionInterval and Telemetry resource

ringerc:
## Description

* Update the FAQ for metric expiry to mention `sidecar.istio.io/statsEvictionInterval` now that support for metric expiry was added by https://github.com/istio/istio/pull/57736
* Mention the `Telemetry` resource's features for overriding metrics and dimensions in the FAQ for metric cardinality
* Link to the `Telemetry` resource and the docs for customising metrics in the reference docs for standard labels
* Warn that dropping labels scrape-side in Prometheus will have undesired results

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [x] Extensions and Telemetry
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/about/faq/metrics-and-logs/telemetry-v1-vs-v2.md
+++ b/content/en/about/faq/metrics-and-logs/telemetry-v1-vs-v2.md
@@ -41,7 +41,7 @@ v2 which are listed below:
   in Mixer-based telemetry. However, more buckets are available by default
   in in-proxy telemetry for latency metrics at the lower latency levels.
 
-* ~~**No metric expiration for short-lived metrics**~~
+* **No metric expiration for short-lived metrics by default**
   Mixer-based telemetry supported metric expiration whereby metrics which were
   not generated for a configurable amount of time were de-registered for
   collection by Prometheus. This is useful in scenarios, such as one-off jobs,
@@ -50,5 +50,6 @@ v2 which are listed below:
   future, thereby reducing network traffic and storage in Prometheus.
 
   The [`sidecar.istio.io/statsEvictionInterval` annotation](/docs/reference/config/annotations/)
-  provides equivalent functionality in newer Istio versions. See also
+  provides equivalent functionality in newer Istio versions, but metric expiry
+  is not enabled by default. See also
   [FAQ: metric expiry](/about/faq/#metric-expiry).


Ajay-singh1:
The tests are failing you need to run ```make lint``` to fix the linting errors.

ringerc:
@Ajay-singh1 Fixed, thanks

content/en/about/faq/metrics-and-logs/metric-expiry.md
```diff
@@ -4,15 +4,38 @@ weight: 20
...
 Short-lived metrics can hamper the performance of Prometheus, as they often are a large source of label cardinality. Cardinality is a measure of the number of unique values for a label. To manage the impact of your short-lived metrics on Prometheus, you must first identify the high cardinality metrics and labels. Prometheus provides cardinality information at its `/status` page. Additional information can be retrieved [via PromQL](https://www.robustperception.io/which-are-my-biggest-metrics).
+
 There are several ways to reduce the cardinality of Istio metrics:
 
+* On Istio 1.28.0 and above, add the
+  [`sidecar.istio.io/statsEvictionInterval`](/docs/reference/config/annotations/)
+  annotation to workload pods to expire metrics for inactive peers. This will
+  help prevent endless growth of the metric scrape responses from the Istio
+  proxy and the resulting in large `scrape_samples_scraped` and
+  `scrape_response_size_bytes` for job instances. This will _not_ prevent
```
dhawton:
```suggestion
  `scrape_response_size_bytes` for job instances. This will not prevent
```

content/en/about/faq/metrics-and-logs/metric-expiry.md
```diff
@@ -4,15 +4,38 @@ weight: 20
...
+  with high cardinality is not needed, you can drop it from metric generation
+  via
+  [metric customization](/docs/tasks/observability/metrics/customize-metrics/)
+  using a `Telemetry` resource's
+  [`metricsOverrides`](/docs/reference/config/telemetry/#MetricsOverrides).
+  See [Telemetry API](/docs/tasks/observability/telemetry/) for examples.
+* Normalize label values through federation or classification.
+  If the information provided by the label is desired, you can use [Prometheus federation](/docs/ops/best-practices/observability/#using-prometheus-for-production-scale-monitoring), Istio workload k8s labels like [`service.istio.io/workload-name`](/docs/reference/config/labels/index.html), or [request classification](/docs/tasks/observability/metrics/classify-metrics/) to normalize the label.
+
+It is _not_ recommended to use Prometheus scrape-time label rewriting to reduce
```
dhawton:
```suggestion
It is not recommended to use Prometheus scrape-time label rewriting to reduce
```

content/en/about/faq/metrics-and-logs/telemetry-v1-vs-v2.md
```diff
@@ -41,11 +41,15 @@ v2 which are listed below:
...
+  that generate short-lived metrics. De-registering
   the metrics prevents reporting of metrics which would no longer change in the
   future, thereby reducing network traffic and storage in Prometheus.
-  This expiration mechanism is not available in in-proxy telemetry.
-  The workaround for this can be found [here](/about/faq/#metric-expiry).
+
+  This expiration mechanism was not available in in-proxy telemetry until
+  Istio 1.28.0 with the addition of the
+  [`sidecar.istio.io/statsEvictionInterval` annotation](/docs/reference/config/annotations/).
+  See also [FAQ: metric expiry](/about/faq/#metric-expiry).
```
dhawton: Since these docs are for 1.28, we should just remove this section.

ringerc: If that's the Istio convention and not negotiable I can make that change but please hear me out first.

I personally find it very confusing to see docs that mention features with no indication of which versions they are supported in. Particularly when they're silently ignored on older versions, and especially when jumping to older versions of docs can have pages reorganised or where you have to re-find the docs page via the index, so I've always appreciated mentioning.

For example, this page does not show a prominent istio version, nor does it have a quick-switch pull-down for the subject version. It's not at all clear which version it applies to except by looking at the URI.

image

Furthermore other parts of the docs mention versions - which I argue is important since it's a FAQ, and would not necessarily be expected by readers to be part of the version-specific reference documentation.

image

I strongly argue that the effective version should be retained here, especially since the annotation is silently ignored if present on an older Istio version.

ringerc: @dhawton Updated the rest, please see my comments on this one

dhawton: The docs are for the latest minor version of Istio. At the bottom, you can see what version they target and switch to older iterations of the docs.

The problem with doing this, very often it remains, and we end up referencing it 3-4 years later (or longer).. as indicated by the 2nd screenshot you provided.

ringerc: OK. I'll update it. I think the overall docs page design/layout could use work to make it _much_ clearer that these docs (including seemingly generic ones like the FAQ) are for a specific version though, especially if policy is to omit version information. Most places with version-specific docs use a prominent display in the top left.

Updated.

zirain (approved):
Thanks for adding this.

content/en/about/faq/metrics-and-logs/telemetry-v1-vs-v2.md
```diff
@@ -41,11 +41,14 @@ v2 which are listed below:
   in Mixer-based telemetry. However, more buckets are available by default
   in in-proxy telemetry for latency metrics at the lower latency levels.
 
-* **No metric expiration for short-lived metrics**
+* ~~**No metric expiration for short-lived metrics**~~
```
dhawton: We do not do strikethroughs. If you're striking through, typically means it no longer applies and should be removed.

ringerc: Updated. Better? I'd rather not remove the section completely, especially since expiry is not enabled by default.

ringerc:
Rebased to current `master` and I think I should've addressed the review comments @dhawton and @zirain

PR #17049, opened 2025-12-22
adds domain override example to global rate limit advanced case

jufantozzi:
## Description

adds domain override example to global rate limit advanced case. This change resolves istio/istio#58574
## Reviewers

- [ ] Ambient
- [X] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/policy-enforcement/rate-limit/snips.sh
+++ b/content/en/docs/tasks/policy-enforcement/rate-limit/snips.sh
@@ -188,6 +188,10 @@ spec:
         operation: MERGE
         value:
           route:
+            typed_per_filter_config:
+              envoy.filters.http.ratelimit:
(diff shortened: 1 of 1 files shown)


jufantozzi:
@dhawton I was thinking about this change, and from my understanding the testing scenario for the advanced case should either have stopped working or the test should have failed, given the domain used in the call defined in the route filter won't match the domain defined in the RLS configuration. Thoughts?

dhawton:
> @dhawton I was thinking about this change, and from my understanding the testing scenario for the advanced case should either have stopped working or the test should have failed, given the domain used in the call defined in the route filter won't match the domain defined in the RLS configuration. Thoughts?

Not sure. This is not an area I am well versed in.

PR #17057, opened 2025-12-23
Fix Global Rate limit test scenarios after adding domain override 

jufantozzi:
## Description

Complementary change of #17049 to address automated testing scenarios of these docs

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/tasks/policy-enforcement/rate-limit/index.md
+++ b/content/en/docs/tasks/policy-enforcement/rate-limit/index.md
@@ -133,7 +133,7 @@ A [reference implementation](https://github.com/envoyproxy/ratelimit) of the API
 
 1. Apply another `EnvoyFilter` to the `ingressgateway` that defines the route configuration on which to rate limit.
     This adds [rate limit actions](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-ratelimit)
-    for any route from a virtual host named `bookinfo.com:80`.
+    for any route from a virtual host named `bookinfo.com:80` and a domain override through the [RateLimitPerRoute filter extension](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto#extensions-filters-http-ratelimit-v3-ratelimitperroute)
 
     {{< text bash >}}
     $ kubectl apply -f - <<EOF


content/en/docs/tasks/policy-enforcement/rate-limit/index.md
```diff
@@ -159,6 +159,10 @@ A [reference implementation](https://github.com/envoyproxy/ratelimit) of the API
             operation: MERGE
             # Applies the rate limit rules.
             value:
+              typed_per_filter_config:
+                envoy.filters.http.ratelimit:
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.ratelimit.v3.RateLimitPerRoute
+                  domain: product # domain override
```
dhawton, line 165: Given the bullet above, this might need some explanation why you're adding a domain override here that doesn't fit the description.

content/en/docs/tasks/policy-enforcement/rate-limit/index.md
```diff
@@ -53,7 +53,7 @@ A [reference implementation](https://github.com/envoyproxy/ratelimit) of the API
       name: ratelimit-config
     data:
       config.yaml: |
-        domain: ratelimit
+        domain: product # domain name is overridden at route filter level
```
dhawton, line 56: Same here, why is this being overridden? The numbered list item does not provide any context here.

content/en/docs/tasks/policy-enforcement/rate-limit/index.md
jufantozzi: For the domain override to work in this testing setup we have to do one of the following, as the current setup relies on all filters/calls to rate limit service using the same domain:
1. Have all route filters configured to use the same domain override 
2. Have a new Rate limit service configuration applied, with the new domain name for the advanced use case
3. Have the domain override just as a placeholder, and make it override the same domain name

I can see pros/cons with all these alternatives, not sure which one is the best to pick from, as

1: This option does not need to change testing setup and should work as is, but as you mentioned, we would need to apply the override everywhere, even in the more simple scenario where it may not be as suitable
2: Probably the most reasonable change, but requires an additional file (svc config) and structural changes to the test. This also increases the overall complexity for the end user
3: Simplest change, wouldn't need any additional changes other than overriding to the same, already applied name, but this could be missleading or not as "correct" as it's missing the entire point of what the domain override is used for

Arhell (commented):
/ok-to-test

PR #17083, opened 2026-01-08
[ES] Proposal to include an i18n conventions file

josunect:
## Description

Translated documents are usually quite long, and it’s common to find different terms used for the same concept. Because of that, it would be useful to have a document where we define conventions and guidelines per language.
Since these translations often rely on AI, we could create a file like i18n-agents-locale.md per language to capture those conventions and help keep things consistent.

Also normalize terms based on the conventions.

Note: The Spanish terms used are a proposal, any suggestion is welcome.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience
- [X] Localization/Translation

### DIFF
--- a/I18N-AGENTS-ES.md
+++ b/I18N-AGENTS-ES.md
@@ -9,20 +9,6 @@ If an existing ES page conflicts with these conventions, **do not “mass-fix”
 
 - Terminology guide (docs contribution): `/es/docs/releases/contribute/terminology/` (**Guía de terminología**)
 
-## Notes from scanning the existing Spanish corpus
-
-The current ES corpus mixes some translations for historical reasons. **Do not mass-edit old pages** to enforce a new standard.
-Instead, when you touch a page, try to make terminology consistent *within that page* and prefer the conventions below.
-
-High-signal inconsistencies observed in `content/es/**`:
-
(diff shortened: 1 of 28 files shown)


I18N-AGENTS-ES.md
```diff
@@ -0,0 +1,103 @@
...
+- **mesh**: “la mesh” (avoid “malla” and “malla de servicios”)
+- **namespace**: “namespace” (avoid “espacio de nombres” in technical contexts)
+- **control plane / data plane**: “control plane” / “data plane” (avoid “plano de control” / “plano de datos”)
+- **gateway**: “gateway” (avoid “puerta(s) de enlace”)
+- **distributed tracing**: “trazado” / “trazado distribuido” (avoid “rastreo” for tracing; keep YAML keys like `tracing` unchanged)
+
+## Style guidelines (Spanish)
+
+- **Voice**: prefer second-person informal already used across the site (“tú/tu”), unless the page clearly uses another voice.
+- **Clarity**: short sentences; avoid overly literal calques from English.
```
keithmattix, line 72: calques?

PR #17111, opened 2026-01-23
remove confusing warning about ztunnel upgrades

therealmitchconnors:
This warning has caused substantial confusion for prospective users, and the issue it describes has been documented as less disruptive than the alternatives.

### DIFF
--- a/content/en/docs/ambient/upgrade/helm/index.md
+++ b/content/en/docs/ambient/upgrade/helm/index.md
@@ -14,12 +14,6 @@ test: yes
 Follow this guide to upgrade and configure an ambient mode installation using
 [Helm](https://helm.sh/docs/). This guide assumes you have already performed an [ambient mode installation with Helm](/docs/ambient/install/helm/) with a previous version of Istio.
 
-{{< warning >}}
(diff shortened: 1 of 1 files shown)


keithmattix:
So I just did some spelunking through history and it looks like we have tests (added in istio/istio#51710) that confirm 90% success rate for connections during a restart. That's pretty good, but maybe we want to link to that test and the follow up issue (istio/istio#51952) for transparency?

therealmitchconnors:
See [#51952](https://github.com/istio/istio/issues/51952) for latest test results.

keithmattix:
That's the follow up issue I linked; is that what you're referring to?

therealmitchconnors:
I think the linked issue is about TCP connect failures, rather than TCP disconnect.  While there is some possibility of long-running TCP connections being disconnected, the impact is much lower than, say, restarting a pod to cordon and drain a node, or to upgrade a sidecar.  Since this is an overall improvement in upgrade disruption, I think the warning is counter-productive.

keithmattix:
I think that's probably true that overall disruption risk is lower. Approving

keithmattix (approved):

PR #17118, opened 2026-01-30
Clarify ResourceQuota from helm

tomasmota:
## Description
It's not immediately clear from the [values](https://github.com/istio/istio/blob/master/manifests/charts/istio-cni/values.yaml) that setting the platform will create the resourcequota. Maybe this will save someone else trying to deploy duplicate resourcequotas like happened to me.

## Reviewers

- [ x] Docs

### DIFF
--- a/content/en/docs/ambient/install/platform-prerequisites/index.md
+++ b/content/en/docs/ambient/install/platform-prerequisites/index.md
@@ -43,7 +43,7 @@ spec:
 
 #### Platform profile
 
-When using GKE you must append the correct `platform` value to your installation commands, as GKE uses nonstandard locations for CNI binaries which requires Helm overrides.
+When using GKE you must append the correct `platform` value to your installation commands, as GKE uses nonstandard locations for CNI binaries which requires Helm overrides. This will also automatically add the necessary ResourceQuota mentioned above.
 
 {{< tabset category-name="install-method" >}}
 


content/en/docs/ambient/install/platform-prerequisites/index.md
```diff
@@ -43,7 +43,7 @@ spec:
 
 #### Platform profile
 
-When using GKE you must append the correct `platform` value to your installation commands, as GKE uses nonstandard locations for CNI binaries which requires Helm overrides.
+When using GKE you must append the correct `platform` value to your installation commands, as GKE uses nonstandard locations for CNI binaries which requires Helm overrides. This will also automatically add the necessary ResourceQuota mentioned above.
```
kfaseela, line 46: Small suggestion: the sentence here feels a bit late because the ResourceQuota is introduced earlier in the “Namespace restrictions” section. To avoid confusion (and potential duplicate ResourceQuota creation), could you move a short note immediately after the ResourceQuota YAML example that says Helm can create that ResourceQuota?

Suggested wording to insert right after the YAML example:

> Note: Installing the istio-cni Helm chart with `--set global.platform=gke` (see the Helm example below) will automatically create the ResourceQuota shown above.

If you add that there, please remove the duplicate sentence from the “Platform profile” here (or replace it with a short cross-reference like “(See note above about ResourceQuota creation.)”).

PR #17174, opened 2026-02-23
Add and update feature status update table

fjglira:
## Description
Updating and adding the Features status according to the latest release notes. Also, Inference to beta was removed for some reason here: https://github.com/istio/istio.io/commit/b43640698f9c9926a557c7f4e77036e169344e63 by the automator

## Reviewers

- [ ] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/data/features.yaml
+++ b/data/features.yaml
@@ -113,6 +113,21 @@ features:
        maturity: Alpha
        nextExpectedPromotion: ""
     area: Traffic Management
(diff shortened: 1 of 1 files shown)


fjglira:
> https://github.com/istio/enhancements/blob/master/features.yaml is the source of the document, where this needs to be changed and is replaced by automator. We don't update it here directly.

That is why the previous update was deleted. THanks

PR #17258, opened 2026-03-25
[Docs] Add Migrate from Sidecar to Ambient documentation

### LINKED ISSUE #17000: Document migration from Sidecar to ambient
We need to add a specific section in the upstream documentation to point the user to the general steps to move from sidecar to ambient mode. I'll start to work on this because I already have some tests and experience with this migration. I'll update this issue with the sections that need to be added

fjglira:
## Description
**This is still work in progress**. Creating the PR to check the format and start to get some feedback on the documentation
Adding specific documentation to migrate from sidecar to ambient to help users guide on this kind of migrations

Fixes #17000 
## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/migrate/_index.md
+++ b/content/en/docs/ambient/migrate/_index.md
@@ -21,7 +21,7 @@ The migration follows a step-by-step approach:
 1. **Install ambient components:** Add ztunnel and update the CNI to support ambient mode,
    while leaving all existing sidecar workloads unchanged.
 1. **Migrate policies:** Convert `VirtualService` resources to `HTTPRoute`, update
-   `AuthorizationPolicy` resources to target waypoints where needed, and retarget
+   `AuthorizationPolicy` resources to target waypoints where needed, and attach
    `RequestAuthentication` and `WasmPlugin` resources to waypoints. This step should be
    skipped if you only use L4 policies.
 1. **Enable ambient mode per namespace:** Label namespaces to join the ambient mesh,
@@ -40,10 +40,10 @@ The following table summarizes how sidecar-mode resources map to their ambient e
 | `DestinationRule` (traffic policies: connection pool, outlier detection, TLS) | No change; waypoints apply traffic policies |
 | `DestinationRule` (routing subsets used with `HTTPRoute`) | Create version-specific Kubernetes Services as `backendRefs` for `HTTPRoute` |
 | `AuthorizationPolicy` with L4 rules | No change; ztunnel enforces L4 policies directly |
-| `AuthorizationPolicy` with L7 rules | Retarget to waypoint using `targetRefs` |
-| `RequestAuthentication` | Retarget to waypoint using `targetRefs` |
+| `AuthorizationPolicy` with L7 rules | Attach to waypoint using `targetRefs` |
+| `RequestAuthentication` | Attach to waypoint using `targetRefs` |
 | `EnvoyFilter` | Not supported on waypoints |
-| `WasmPlugin` | Retarget to waypoint using `targetRefs` |
+| `WasmPlugin` | Attach to waypoint using `targetRefs` |
 | `Gateway` (networking.istio.io/v1) | No change required; Istio Gateway resources continue to work in ambient mode. Add `istio.io/ingress-use-waypoint` to route ingress traffic through a waypoint. |
 
 ## Do you need waypoint proxies?
--- a/content/en/docs/ambient/migrate/_index.md
+++ b/content/en/docs/ambient/migrate/_index.md
@@ -20,20 +20,37 @@ The migration follows a step-by-step approach:
 
 1. **Install ambient components:** Add ztunnel and update the CNI to support ambient mode,
    while leaving all existing sidecar workloads unchanged.
-1. **Migrate policies:** Convert `VirtualService` and `DestinationRule` resources
-   to Gateway API (`HTTPRoute`) equivalents, and update `AuthorizationPolicy` resources to
-   target waypoints where needed. This step should be skipped if you only use L4 policies.
+1. **Migrate policies:** Convert `VirtualService` resources to `HTTPRoute`, update
+   `AuthorizationPolicy` resources to target waypoints where needed, and retarget
+   `RequestAuthentication` and `WasmPlugin` resources to waypoints. This step should be
+   skipped if you only use L4 policies.
 1. **Enable ambient mode per namespace:** Label namespaces to join the ambient mesh,
    activate waypoints, remove sidecar injection, and restart pods.
 
 Each step is independently reversible. There is no requirement to migrate all
 namespaces at once.
 
+## Resource migration overview
+
+The following table summarizes how sidecar-mode resources map to their ambient equivalents:
+
+| Sidecar resource | Action in ambient mode |
+|---|---|
+| `VirtualService` | Migrate to `HTTPRoute` (`VirtualService` support is Alpha in ambient) |
+| `DestinationRule` (traffic policies: connection pool, outlier detection, TLS) | No change; waypoints apply traffic policies |
+| `DestinationRule` (routing subsets used with `HTTPRoute`) | Create version-specific Kubernetes Services as `backendRefs` for `HTTPRoute` |
+| `AuthorizationPolicy` with L4 rules | No change; ztunnel enforces L4 policies directly |
+| `AuthorizationPolicy` with L7 rules | Retarget to waypoint using `targetRefs` |
+| `RequestAuthentication` | Retarget to waypoint using `targetRefs` |
+| `EnvoyFilter` | Not supported on waypoints |
+| `WasmPlugin` | Retarget to waypoint using `targetRefs` |
+| `Gateway` (networking.istio.io/v1) | No change required; Istio Gateway resources continue to work in ambient mode. Add `istio.io/ingress-use-waypoint` to route ingress traffic through a waypoint. |
+
 ## Do you need waypoint proxies?
 
 {{< tip >}}
 Waypoint proxies are **optional**. If you only need mTLS and L4 authorization policies,
-you can migrate entirely to ztunnel without deploying any waypoint proxies. This allows you to benefit from ambient mode's simplified operations and improved performance without needing to change your existing policies or traffic management configuration.
+you can migrate to ztunnel without deploying waypoints and without changing any existing policies.
 {{< /tip >}}
 
 You need waypoint proxies if your workloads use any of the following:
@@ -47,6 +64,10 @@ If you are unsure, the [migrate policies](/docs/ambient/migrate/migrate-policies
 helps you audit your existing resources.
 
 ## What is not supported
+
+{{< tip >}}
+The limitations listed below reflect the current stable Istio release. Ambient mode continues to evolve and some of these constraints may be lifted in later versions. Check the [release notes](/news/releases/) for updates specific to your Istio version.
+{{< /tip >}}
 
 The following are hard blockers, migration is not possible until these are resolved:
 
@@ -63,11 +84,13 @@ The following are known limitations that affect behavior during or after migrati
   `EnvoyFilter` for advanced Envoy configuration on your sidecar proxies, those
   configurations cannot be carried over to waypoints. This API may be supported in a
   future release.
-- **Traffic from sidecar mode workloads and ingress gateways does not go through
-  waypoint proxies**. During an incremental migration, if a sidecar mode workload calls
-  an ambient mode workload that has a waypoint, the traffic bypasses the waypoint entirely.
-  L7 policies on the waypoint are not enforced for that traffic until the source is also
-  migrated to ambient mode.
+- **Traffic from sidecar mode workloads bypasses waypoint proxies**. During an incremental
+  migration, if a sidecar mode workload calls an ambient mode workload that has a waypoint,
+  the traffic bypasses the waypoint entirely. L7 policies on the waypoint are not enforced
+  for that traffic until the source workload is also migrated to ambient mode.
+- **Ingress gateways bypass waypoints by default**, but can be configured to route traffic
+  through a waypoint by adding the `istio.io/ingress-use-waypoint` label to the Gateway
+  resource.
 - **Mixing `VirtualService` and `HTTPRoute` for the same workload is not supported** and
   leads to undefined behavior. Migrate each workload fully to one API before proceeding.
 
--- a/content/en/docs/ambient/migrate/_index.md
+++ b/content/en/docs/ambient/migrate/_index.md
@@ -1,0 +1,77 @@
+---
+title: Migrate from Sidecar to Ambient
+description: Migrate an existing sidecar-based mesh to ambient mode.
+weight: 12
+owner: istio/wg-networking-maintainers
+test: no
+skip_list: true
+next: /docs/ambient/migrate/before-you-begin
+---
+
+This guide walks you through migrating an existing Istio deployment from
+{{< gloss >}}sidecar{{< /gloss >}} mode to {{< gloss "ambient" >}}ambient mode{{< /gloss >}}.
+The migration is designed to be gradual and reversible: sidecar and ambient workloads can
+coexist in the same mesh during the process, allowing you to migrate one namespace at
+a time with no downtime.
+
+## Migration strategy
+
+The migration follows a step-by-step approach:
+
+1. **Install ambient components:** Add ztunnel and update the CNI to support ambient mode,
+   while leaving all existing sidecar workloads unchanged.
+1. **Migrate policies:** Convert `VirtualService` and `DestinationRule` resources
+   to Gateway API (`HTTPRoute`) equivalents, and update `AuthorizationPolicy` resources to
+   target waypoints where needed. This step should be skipped if you only use L4 policies.
+1. **Enable ambient mode per namespace:** Label namespaces to join the ambient mesh,
+   activate waypoints, remove sidecar injection, and restart pods.
+
+Each step is independently reversible. There is no requirement to migrate all
+namespaces at once.
+
+## Do you need waypoint proxies?
+
+{{< tip >}}
+Waypoint proxies are **optional**. If you only need mTLS and L4 authorization policies,
+you can migrate entirely to ztunnel without deploying any waypoint proxies. This allows you to benefit from ambient mode's simplified operations and improved performance without needing to change your existing policies or traffic management configuration.
+{{< /tip >}}
+
+You need waypoint proxies if your workloads use any of the following:
+
+- L7 `AuthorizationPolicy` rules (matching on HTTP methods, paths, or headers).
+- L7 traffic routing (retries, fault injection, header manipulation, traffic splitting) via `HTTPRoute`. If you currently use `VirtualService` for this, you will need to migrate to `HTTPRoute`, `VirtualService` support in ambient is Alpha.
+- `RequestAuthentication` (JWT validation).
+- L7 telemetry enrichment.
+
+If you are unsure, the [migrate policies](/docs/ambient/migrate/migrate-policies/) page
+helps you audit your existing resources.
+
+## What is not supported
+
+The following are hard blockers, migration is not possible until these are resolved:
+
+- **VM workloads** in the mesh. VM-based workloads cannot join the ambient mesh.
+- **SPIRE** as the certificate provider. Ambient mode does not support SPIRE integration.
+- **`PeerAuthentication` with `mode: DISABLE`**. Ambient always enforces mTLS between
+  mesh workloads. Policies with `DISABLE` mode will be ignored and cannot be migrated.
+- **Primary-remote multicluster configurations**. Only multiple-primary clusters are
+  supported. Deployments with one or more remote clusters will not work correctly.
+
+The following are known limitations that affect behavior during or after migration:
+
+- **`EnvoyFilter` resources targeting waypoints are not supported**. If you rely on
+  `EnvoyFilter` for advanced Envoy configuration on your sidecar proxies, those
+  configurations cannot be carried over to waypoints. This API may be supported in a
+  future release.
+- **Traffic from sidecar mode workloads and ingress gateways does not go through
+  waypoint proxies**. During an incremental migration, if a sidecar mode workload calls
+  an ambient mode workload that has a waypoint, the traffic bypasses the waypoint entirely.
+  L7 policies on the waypoint are not enforced for that traffic until the source is also
+  migrated to ambient mode.
+- **Mixing `VirtualService` and `HTTPRoute` for the same workload is not supported** and
+  leads to undefined behavior. Migrate each workload fully to one API before proceeding.
+
+## Next steps
+
+Start with [Before you begin](/docs/ambient/migrate/before-you-begin/) to verify
+your environment and back up your configuration.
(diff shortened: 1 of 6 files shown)


content/en/docs/ambient/migrate/before-you-begin/index.md
```diff
@@ -0,0 +1,84 @@
...
+prev: /docs/ambient/migrate
+next: /docs/ambient/migrate/install-ambient-components
+---
+
+Before migrating from sidecar to ambient mode, verify that your environment meets the
+requirements and create a backup of your current configuration.
+
+## Requirements
+
+- Istio 1.24 or later (1.25+ recommended for full feature support)
```
keithmattix: We should recommend/require a supported version at least

content/en/docs/ambient/migrate/before-you-begin/index.md
```diff
@@ -0,0 +1,84 @@
...
+{{< text syntax=bash snip_id=none >}}
+$ kubectl get namespaces -l 'istio.io/rev'
+{{< /text >}}
+
+## Audit existing resources
+
+List the Istio resources in use across your cluster:
+
+{{< text syntax=bash snip_id=none >}}
+$ kubectl get virtualservice,destinationrule,authorizationpolicy,requestauthentication,peerauthentication -A
```
keithmattix: EnvoyFilters missing. Same with WASMPlugin

content/en/docs/ambient/migrate/migrate-policies/index.md
```diff
@@ -0,0 +1,282 @@
...
+test: no
+prev: /docs/ambient/migrate/install-ambient-components
+next: /docs/ambient/migrate/enable-ambient-mode
+---
+
+{{< tip >}}
+**You may be able to skip this page.** If you only use L4 `AuthorizationPolicy` rules
+(with no `methods`, `paths`, or `headers` matching) and have no `VirtualService` or
+`DestinationRule` resources, your existing policies will work in ambient mode without
+changes. Go directly to [Enable ambient mode](/docs/ambient/migrate/enable-ambient-mode/).
```
keithmattix, line 16: Not true if you have EnvoyFilter,WasmPlugin, or RequestAuthentication

content/en/docs/ambient/migrate/migrate-policies/index.md
```diff
@@ -0,0 +1,282 @@
...
+weight: 3
+owner: istio/wg-networking-maintainers
+test: no
+prev: /docs/ambient/migrate/install-ambient-components
+next: /docs/ambient/migrate/enable-ambient-mode
+---
+
+{{< tip >}}
+**You may be able to skip this page.** If you only use L4 `AuthorizationPolicy` rules
+(with no `methods`, `paths`, or `headers` matching) and have no `VirtualService` or
```
keithmattix: CUSTOM or AUDIT authz policy also requires waypoint

content/en/docs/ambient/migrate/migrate-policies/index.md
fjglira: Added a specific comment about this

content/en/docs/ambient/migrate/before-you-begin/index.md
fjglira: Added to the get command

content/en/docs/ambient/migrate/migrate-policies/index.md
fjglira: Added a comment about this

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
```diff
@@ -0,0 +1,193 @@
...
+$ kubectl label namespace <namespace> istio.io/rev-
+{{< /text >}}
+
+{{< warning >}}
+Removing the injection label alone does not remove existing sidecars. Pods must be
+restarted for the change to take effect. Do not restart pods until you have confirmed
+that ambient mode is active (Step 2 above).
+{{< /warning >}}
+
+## Step 4: Restart pods
```
krinkinmu: Can you clarify a bit what happens during transitional state when some pods have been migrated, but some have not?

To be specific, let's say we have L7 policies that affect backend selection and/or rewrite the request (i.e., HTTPRoute that matches the request and route to different destination and/or modify request headers). We don't want those L7 policies to be applied multiple times (e.g., by sidecar and by waypoint).

Do I understand correctly, that to avoid this issue, we basically need to make sure that sidecar, after applying L7 policies never sends a request to waypoint to avoid applying any L7 policies the second time (IOW, service selected by the sidecar as a destination for the request is treated as a backend service and no L7 policies will be applied to it again and modify the request again or change the destination of the request)?

And if my understanding is correct, can you clarify who is responsible for ensuring that? Does `ISTIO_META_ENABLE_HBONE` guarantee that requests will never be routed to waypoints or does the user have to make sure to adjust the policies in the right way?

And finally, how would that interact with the L4 policy suggest elsewhere in this guide that makes sure that ambient-enabled pods don't accept requests from non-waypoints via an L4 authorization policy?

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Post-migration observability changes
+
+After migrating to ambient mode, be aware of the following changes to telemetry:
+
+**Metrics**: In sidecar mode, metrics are reported with `reporter="source"` and
+`reporter="destination"`. In ambient mode, metrics from ztunnel use `reporter="source"`,
+while metrics from waypoint proxies use `reporter="waypoint"`. Update any dashboards or
+alerting rules that rely on the `reporter` label.
+
```
frittentheke, line 216: A big thing are likely merged metrics provided by the proxy-agent, see https://istio.io/latest/docs/ops/integrations/prometheus/#option-1-metrics-merging

This solely lives off of the "well known" `prometheus.io` annoations which then allow for e.g. a single Prometheus Operator `PodMonitor` to scrape all workload Pods making use of metrics merging.

Merged metrics have issues:

* https://github.com/istio/istio/issues/49950
* https://github.com/istio/istio/issues/59567

but it's likely a very common feature people use ...



And a well intended recommendation of a change during ambient migration solely depends on the outcome of discussions like https://github.com/istio/istio/issues/59567 about the future of this feature (for sidecar mode). Especially since it's inherently unavailable when using ambient mode.

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
fjglira: Hey, sure. Thanks for raising these questions @krinkinmu! I'll update the docs to fill these gaps, as these are definitely points that other users will find during their migration.

* What happens during that transitional state when some pods are in sidecar mode and others are in ambient? We don't want L7 policies applied multiple times.

Once `ISTIO_META_ENABLE_HBONE` is enabled in the sidecars, the sidecar will be able to communicate with the ambient mesh using HBONE. Because it talks now HBONE, it knows how to tunnel that traffic directly to the destination pod's ztunnel. This effectively bypasses the destination's waypoint for L7 processing, ensuring policies aren't applied twice. The sidecar treats the ambient destination as a standard backend once the L7 logic is resolved.

* How would that interact with the L4 policy suggested elsewhere in this guide that ensures ambient-enabled pods don't accept requests from non-waypoints?

You found an error in my documentation. This is a critical point in a step-by-step migration. If you apply a strict L4 `AuthorizationPolicy` that requires all traffic to come from a waypoint, it will reject the traffic coming from pods still in Sidecar mode (since they bypass the waypoint).

I’ll update the guide to clarify that during this state, your L4 policies must be inclusive to accept both traffic. I think the best approach is to define policies that allow traffic from either the Waypoint or sidecars. We can transition to waypoint only enforcement once the migration for that specific service path is complete. Also I'll add a note to make user it's clear for users

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
fjglira: Hey @frittentheke, thanks for bringing this up. I’ll update the documentation to explicitly mention that the metrics merging is inherently unavailable in Ambient mode. Users currently relying on a single scrape target for both Istio and their application will need to decouple their monitoring:

* Ztunnel/Waypoint: These must be scraped as their own entities. I didn't specify this in the current doc

* Application metrics: Prometheus must now scrape the application pods directly, as there is no longer a sidecar agent to perform the proxy-merge.

Regarding the recommendation, I agree that since the future of this feature in sidecar mode is still under discussion (istio/istio#59567), we shouldn't mention any strategy yet. However, warning users as I mentioned before, I think is good enough.

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
krinkinmu: Thanks for clarifying things. I have a followup question here.

In sidecar mode, roughly, client-side sidecar does endpoint selection and load-balancing and the server-side sidecar does authorization policy enforcement.

With that in mind, does it mean that during the transient state, if client is still using sidecar and the server is already in ambient mode, the L7 authorization policies don't get enforced?

If my understanding is correct, I can see how it might be ok for a short period of time for somebody, but in general it seems somewhat problematic.

mikemorris: Agreed @krinkinmu, I would expect the more desirable migration path might be "making the sidecar dumb" i.e. stop applying L7 policies at sidecar egress and start routing traffic from sidecar clients to waypoints which will apply L7 policies instead (this also frontloads potential breaking changes where client-side policies not supported in ambient are dropped before a full cut-over).

content/en/docs/ambient/migrate/_index.md
```diff
@@ -0,0 +1,77 @@
...
+
+- L7 `AuthorizationPolicy` rules (matching on HTTP methods, paths, or headers).
+- L7 traffic routing (retries, fault injection, header manipulation, traffic splitting) via `HTTPRoute`. If you currently use `VirtualService` for this, you will need to migrate to `HTTPRoute`, `VirtualService` support in ambient is Alpha.
+- `RequestAuthentication` (JWT validation).
+- L7 telemetry enrichment.
+
+If you are unsure, the [migrate policies](/docs/ambient/migrate/migrate-policies/) page
+helps you audit your existing resources.
+
+## What is not supported
```
sridhargaddam, line 76: Shall we add a note like "as of Istio release 1.29..." as this can become stale in future?

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
```diff
@@ -0,0 +1,201 @@
...
+$ kubectl rollout status deployment -n <namespace>
+{{< /text >}}
+
+## Step 5: Remove old sidecar policies
+
+{{< warning >}}
+Do this immediately after the pod restart, before running any validation. Any
+`AuthorizationPolicy` using a workload `selector` with L7 rules that remains active
+after sidecars are removed will be enforced by ztunnel as a `DENY` policy for all
+traffic to that workload, regardless of the HTTP method or path.
```
sridhargaddam: Since an AuthorizationPolicy can be either ALLOW or DENY, the behavior is slightly different between the two.

If an ALLOW AuthorizationPolicy targets a workload and its rules include L7 attributes, ztunnel drops those rules entirely. If every rule in the policy relies on L7 attributes, the resulting policy ends up with no rules at all and allows nothing. In practice, this blocks all traffic to the workload. It’s stricter than intended, but it’s still treated as an ALLOW policy with no matches, not as a DENY policy.

If a DENY AuthorizationPolicy includes L7 attributes, ztunnel simply removes those L7 conditions and enforces whatever L4 conditions remain, such as ports, IPs, or service accounts. If a rule doesn’t have any L4 conditions to begin with (for example, if it only uses request principals or other L7 checks), then removing the L7 parts leaves an empty match, which applies to all traffic. That effectively turns the policy into a deny-all for that workload.

content/en/docs/ambient/migrate/migrate-policies/index.md
```diff
@@ -0,0 +1,327 @@
...
+changes. Go directly to [Enable ambient mode](/docs/ambient/migrate/enable-ambient-mode/).
+{{< /tip >}}
+
+In ambient mode, L7 traffic management is handled by {{< gloss >}}waypoint{{< /gloss >}}
+proxies rather than sidecar proxies. This changes how policies are expressed and enforced:
+
+- **`VirtualService`** support with waypoints is **Alpha**. While it may work in limited
+  cases, migrating to `HTTPRoute` is strongly recommended. Mixing `VirtualService` and
+  `HTTPRoute` for the same workload is not supported and leads to undefined behavior.
+- **`DestinationRule`** subset-based routing is not supported by waypoints. Subsets must be
```
sridhargaddam: AFAICT, DestinationRule subset-based routing is **supported** by waypoints. Only east-west ambient gateways might have some gaps.

content/en/docs/ambient/migrate/before-you-begin/index.md
```diff
@@ -0,0 +1,84 @@
...
+Check which `AuthorizationPolicy` resources contain L7 rules. These will require waypoint
+proxies to function in ambient mode:
+
+{{< text syntax=bash snip_id=none >}}
+$ kubectl get authorizationpolicy -A --no-headers | while read ns name rest; do
+    if kubectl get authorizationpolicy "$name" -n "$ns" -o yaml | grep -qE "(methods:|paths:|headers:)"; then
+      echo "$ns/$name"
+    fi
+  done
+{{< /text >}}
```
sridhargaddam, line 97:
```suggestion
{{< text syntax=bash snip_id=none >}}
$ kubectl get authorizationpolicy -A --no-headers | while read ns name rest; do
    if kubectl get authorizationpolicy "$name" -n "$ns" -o yaml | grep -qE "(methods:|paths:|headers:|action: CUSTOM|action: AUDIT)"; then
      echo "$ns/$name"
    fi
  done
{{< /text >}}
```

content/en/docs/ambient/migrate/_index.md
```diff
@@ -0,0 +1,77 @@
...
+- **Primary-remote multicluster configurations**. Only multiple-primary clusters are
+  supported. Deployments with one or more remote clusters will not work correctly.
+
+The following are known limitations that affect behavior during or after migration:
+
+- **`EnvoyFilter` resources targeting waypoints are not supported**. If you rely on
+  `EnvoyFilter` for advanced Envoy configuration on your sidecar proxies, those
+  configurations cannot be carried over to waypoints. This API may be supported in a
+  future release.
+- **Traffic from sidecar mode workloads and ingress gateways does not go through
```
sridhargaddam: The statement about ingress gateways is not correct.

Sidecars: Always bypass waypoints
Ingress gateways: Bypass waypoints by default, but can be configured to use them via the `istio.io/ingress-use-waypoint` label

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
fjglira: Ok ok, sounds good to me. I'll do today some updates

content/en/docs/ambient/migrate/_index.md
fjglira: Yeah, I think it can be good to have a note. I'll add it

content/en/docs/ambient/migrate/_index.md
```diff
@@ -0,0 +1,77 @@
...
+  mesh workloads. Policies with `DISABLE` mode will be ignored and cannot be migrated.
+- **Primary-remote multicluster configurations**. Only multiple-primary clusters are
+  supported. Deployments with one or more remote clusters will not work correctly.
+
+The following are known limitations that affect behavior during or after migration:
+
+- **`EnvoyFilter` resources targeting waypoints are not supported**. If you rely on
+  `EnvoyFilter` for advanced Envoy configuration on your sidecar proxies, those
+  configurations cannot be carried over to waypoints. This API may be supported in a
+  future release.
```
voipp, line 96: may be add some notes on probable replacement rules

content/en/docs/ambient/migrate/_index.md
```diff
@@ -0,0 +1,77 @@
...
+- **Traffic from sidecar mode workloads and ingress gateways does not go through
+  waypoint proxies**. During an incremental migration, if a sidecar mode workload calls
+  an ambient mode workload that has a waypoint, the traffic bypasses the waypoint entirely.
+  L7 policies on the waypoint are not enforced for that traffic until the source is also
+  migrated to ambient mode.
+- **Mixing `VirtualService` and `HTTPRoute` for the same workload is not supported** and
+  leads to undefined behavior. Migrate each workload fully to one API before proceeding.
```
voipp, line 105: what about migrating Kind: Gateway apiVersion: networking.istio.io/v1

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
fjglira: I added a whole new section at the beginning of the file called `### Phase 1: Make the client sidecar passive`, the idea is to explain to users how to move those L7 policies to waypoints making as you mentione the "sidecar dumb", I rephrased to passive xD

content/en/docs/ambient/migrate/_index.md
fjglira: Added a note

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
fjglira: I added a explanation about this

content/en/docs/ambient/migrate/_index.md
fjglira: Fixed

fjglira:
Hi @mikemorris @krinkinmu @keithmattix @sridhargaddam I added an update with changes from the suggestion. I'm still going over the Netlify version to see that everything looks good

content/en/docs/ambient/migrate/enable-ambient-mode/index.md
krinkinmu: I'm a bit confused, we agreed earlier that sidecar never sends to the waypoint, right?

If you remove the policies from the sidcar to waypoint, doesn't it create the same problem again - during the transient state the L7 policies may not be enforced. Did I misunderstood something?

krinkinmu: Actually, forget about that. My understanding is still that authorization policies in sidecar mode are enforced by the server-side sidecars (at least the authorization part of the policies), so if sidecars never send to the waypoint, it does not seem to matter how the policies are changed - we still kind of have the situation when the authoirzation policies are not enforced if the client is sidecar.

krinkinmu: FWIW, @fjglira we are quite interested in migration to ambient ourselves, if you think it would be helpful to discuss it over a higher bandwidth channel we can set up a vide call or start a chat in slack and I'd be happy to help.

fjglira: We chat more about this on slack. Let's connect and find the best path together :)

content/en/docs/ambient/migrate/migrate-policies/index.md
fjglira: I added a note about this

fjglira:
After today's discussion around the best approach for this topic, I'll do some modifications to the current documentation, and tomorrow I'll submit an update. The main idea is to focus first on the L4-only migration use case, and add warning about gaps on migration with L7 policies. At the same time we will work into fill those gaps

content/en/docs/ambient/migrate/migrate-policies/index.md
```diff
@@ -0,0 +1,337 @@
...
+
+- **`VirtualService`** support with waypoints is **Alpha**. While it may work in limited
+  cases, migrating to `HTTPRoute` is strongly recommended. Mixing `VirtualService` and
+  `HTTPRoute` for the same workload is not supported and leads to undefined behavior.
+- **`DestinationRule`** traffic policies (connection pool settings, outlier detection, TLS)
+  are supported by waypoints and require no changes. However, `HTTPRoute` uses Kubernetes
+  Services as `backendRefs` for routing rather than DestinationRule subsets, so
+  version based traffic splitting in `HTTPRoute` requires separate Services per version.
+- **`AuthorizationPolicy`** resources that use L7 rules (HTTP methods, paths, or headers),
+  or that use `action: CUSTOM` or `action: AUDIT`, must target waypoint proxies via
```
krinkinmu: They could also target a service, or is that not supported?

content/en/docs/ambient/migrate/migrate-policies/index.md
```diff
@@ -0,0 +1,337 @@
...
+    -o jsonpath='{.items[0].spec.serviceAccountName}'
+{{< /text >}}
+
+{{< warning >}}
+Do not use `targetRefs` for this policy. A `targetRefs` based DENY policy is enforced by
+the waypoint, which sees the original client identity not the waypoint's own identity.
+This would cause the waypoint to deny all client traffic before the ALLOW policy can run.
+{{< /warning >}}
+
+### Bypass prevention during incremental migration
```
krinkinmu: I feel like these options should be explained before bypass prevention.

content/en/docs/ambient/migrate/before-you-begin/index.md
krinkinmu, line 1: I think it might be benefitial to explain a little bit the differences between ambient and sidecar relevant for migration to give the reader background needed to understand the actual instructions.

A few things that come to mind that makes sense to elaborate on upfront:

1. How policies are enforced in sidecar and in ambient:
    - AuthorizationPolicy in sidecar always apply to the destination pods matching the labels and are enforced by the destination sidecar
    -  In ambient we have policy enforcement split between ztunnel and waypoints
    -  Policies enforced by waypoint are applied to service or gateway and not a destination pod, so you cannot just re-use the same policies across sidecar and ambient, you have to create functionally equivalanet policies for ambient.
2. Supported and unsupported features:
    - VirtualService support in waypoints not being production ready (the instructions will go into details on how to migrate from VirtualService to HTTPRoute, but even so, it makes sense to outline the scope of work upfront).

krinkinmu (commented):
@fjglira just checking on this to make sure it didn't fall through the cracks?

We discussed in the meeting a few changes to the PR, I also went over the docs again and left a few minor comments.

fjglira:
> @fjglira just checking on this to make sure it didn't fall through the cracks?
> 
> We discussed in the meeting a few changes to the PR, I also went over the docs again and left a few minor comments.

Hey, thanks a lot for checking again. Yes, I'll be making some more changes to the PR this week. I have been busy dealing with some internal CI issues that are driving me crazy xD. Hope to finish that today, and I'll start making changes tomorrow

content/en/docs/ambient/migrate/migrate-policies/index.md
fjglira: Yes, I just added a better explanation

content/en/docs/ambient/migrate/migrate-policies/index.md
fjglira: Done, moved before the bypass prevention

content/en/docs/ambient/migrate/before-you-begin/index.md
fjglira: Ok, I added an explanation for this

krinkinmu, line 1: @fjglira can you add a section or a warning here calling out potential issues when L7 policies are involved.

It seems like we could have cases where we either will have some downtime during the migration or some L7 policies may not be enforced all the time during the migration.

I'd like the warning to be prominent here just to avoid any impressions that migration for L7 policies is currently well supported.

krinkinmu (changes requested):

content/en/docs/ambient/migrate/before-you-begin/index.md
fjglira: Sure, I just added a warning section at the beggining of the file

krinkinmu (approved):
Thank you, @fjglira!

keithmattix (approved):

PR #17324, opened 2026-04-15
docs: note that OutlierDetection automatically activates locality LB failover

yasaswyk:
## Description

Adds a tip callout to the locality load balancing overview page explaining that Istio's default mesh config enables locality LB (`LocalityLbSetting.enabled: true`). As a result, configuring `OutlierDetection` in a `DestinationRule` automatically activates locality-aware failover behavior, even without an explicit `localityLbSetting` in the rule. Documents the opt-out: set `localityLbSetting.enabled: false` in the DR's load balancer settings.

## Reviewers

- [x] Networking
- [x] Docs


Arhell (commented):
/ok-to-test

yasaswyk:
@Arhell mind approving this PR?

keithmattix (approved):

PR #17329, opened 2026-04-18
docs: add note about expected IST0173 warnings when applying Bookinfo destination rules

Ahmedaltu:
## What this PR does

Adds a tip note to the Bookinfo documentation explaining that applying
`destination-rule-all.yaml` may produce IST0173 warnings from
`istioctl analyze` for the `v2-mysql` and `v2-mysql-vm` subsets of
the ratings service.

## Why

These subsets are defined in `destination-rule-all.yaml` but the
corresponding pods (`ratings-v2-mysql`, `ratings-v2-mysql-vm`) are not
deployed as part of the standard Bookinfo installation. Users following
the getting started guide who run `istioctl analyze` will see:

Error [IST0173] (DestinationRule default/ratings) The Subset v2-mysql
defined in the DestinationRule does not select any pods. Which may
lead to 503 UH (NoHealthyUpstream).

Without any explanation, this causes confusion about whether the
installation is healthy. This note clarifies the errors are expected
and safe to ignore for a standard deployment.

## Type of change

- [x] Documentation fix

### DIFF
--- a/content/en/docs/examples/bookinfo/index.md
+++ b/content/en/docs/examples/bookinfo/index.md
@@ -236,6 +236,15 @@ To enforce mutual TLS, use the destination rules in `samples/bookinfo/networking
 
 Wait a few seconds for the destination rules to propagate.
 
+{{< tip >}}
+The `destination-rule-all.yaml` file defines subsets for all Bookinfo service versions,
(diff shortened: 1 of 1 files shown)


Arhell (commented):
/ok-to-test

PR #17344, opened 2026-04-24
test: improve kubectl context handling in test utilities

irontony70345-code:
Only pass --context to kubectl when context is provided.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience


Arhell (commented):
/ok-to-test

fjglira (commented):
Same here, I think the gain from this change is minimal and it is even a step back, hiding a well-used flag across some utils files

PR #17345, opened 2026-04-24
test: factor addon apply and wait helper

irontony70345-code:
This PR factors repeated addon apply-and-wait logic into a small helper.

The change keeps the existing behavior unchanged while reducing duplication in addon deployment test utilities.

## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/tests/util/addons.sh
+++ b/tests/util/addons.sh
@@ -16,27 +16,32 @@
 # limitations under the License.
 
 source "tests/util/helpers.sh"
+# Apply an addon manifest and wait for the deployment to be ready.
+# usage: _apply_addon_and_wait <manifest> <deployment>
+function _apply_addon_and_wait() {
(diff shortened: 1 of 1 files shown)


Arhell (commented):
/ok-to-test

fjglira (commented):
To be honest, I don't think this change is necessary. It replaces a 2-line pattern that appears only 4 times in a test utility file; the gain is minimal, and it also leaves kiali inconsistent (which intentionally applies twice due to a race condition, so the helper wouldn't fit there anyway).

I don't want to discourage you from contributing to the project, quite the opposite! But I think there are better areas to start. 

I'm happy to point you toward something more impactful if you'd like

PR #17425, opened 2026-06-06
fix: correct android favicon size metadata

immanuwell:
## Description
Fixes a typo in the favicon metadata for `android-96x96.png`.

Right now the built site renders `sizes="96xW96"`, so the icon size hint is invalid. This patch changes it to `96x96`. Tiny fix, but it ships on every page rn.

Repro:
1. Run `hugo -d /tmp/istioio-public`
2. Open `/tmp/istioio-public/index.html`
3. Find `android-96x96.png` and note `sizes="96xW96"`

After this change, the same line renders `sizes="96x96"`.

## Reviewers
- [x] Docs
- [x] User Experience


Arhell (commented):
/ok-to-test

craigbox (approved):

craigbox:
/cherry-pick release-1.30

is that that old bug where you can't use the label if the owner isn't an org member?

PR #17463, opened 2026-07-01
updating support table for 1.28 EoL

grnmeira:
## Reviewers

- [ ] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/data/compatibility/supportStatus.yml
+++ b/data/compatibility/supportStatus.yml
@@ -22,7 +22,7 @@
 - version: "1.28"
   supported: "No"
   releaseDate: "Nov 05, 2025"
-  eolDate: "Jun 28, 2026"
+  eolDate: "Jul 1, 2026"
   k8sVersions: ["1.30", "1.31", "1.32", "1.33", "1.34"]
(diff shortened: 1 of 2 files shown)


data/compatibility/supportStatus.yml
```diff
@@ -20,7 +20,7 @@
   k8sVersions: ["1.31", "1.32", "1.33", "1.34", "1.35"]
   testedK8sVersions: ["1.26", "1.27", "1.28", "1.29", "1.30"]
 - version: "1.28"
-  supported: "Yes"
+  supported: "No"
   releaseDate: "Nov 05, 2025"
   eolDate: "Jun 28, 2026"
```
dhawton: Given 1.28.10 published today, shouldn't this be updated?

dhawton (changes requested):
We also need the 1.28 equivalent of https://istio.io/latest/news/support/announcing-1.27-eol-final/

data/compatibility/supportStatus.yml
grnmeira: We could do that.

dhawton (approved):

PR #17471, opened 2026-07-06
Fix TCPRoute apiVersion for Gateway API v1.6.0

dhawton:
Bumps the Kubernetes Gateway API to v1.6.0. In v1.6.0, `TCPRoute` graduated to the standard channel as `v1` and `v1alpha2` is deprecated, so `kubectl get tcproute -o yaml` in the tcp-traffic-shifting task now returns `apiVersion: gateway.networking.k8s.io/v1`. Updates the expected snippet output and localized display blocks to match.

Supersedes #17470 (includes its v1.6.0 bump plus the doc fix that was failing `doc.test.profile-minimal`).


keithmattix (approved):

kfaseela (approved):

fjglira (approved):
lgtm

PR #17485, opened 2026-07-16
Fix markdown style issues in English content

kassvl:
Part of #15148. First batch of the plan described there: content fixes that are valid under both the current mdl linter and markdownlint-cli2, so they can land ahead of the engine swap. Covers list marker consistency (MD004), list nesting depth (MD007), code block shortcode bodies that weren't aligned with their list indentation, and a handful of bare URLs and email addresses.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -1587,7 +1587,6 @@ whitelists
 whitespace
 Wikipedia
 wikipedia.org
-www.wikipedia.org
 wildcard
 wildcarded
 wildcards
@@ -1599,6 +1598,7 @@ workload
 WorkloadEntry
 workstream
 www.google.com
+www.wikipedia.org
 x-envoy-upstream-rq-timeout-ms
 X.509
 X.509.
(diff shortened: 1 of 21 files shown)


Arhell (commented):
/ok-to-test

dhawton:
A lot of these changes are unnecessary. We actually prefer `-` over `*`. Some places you remove `-` and replace with `*`, and in others, you do the opposite.

kassvl:
Fair point, thanks. The earlier commits normalized each mixed file toward its dominant marker, which is where the dash to star conversions came from. Fixed in ecf7010: every touched file now uses dash only, so nothing is converted away from dash anymore.

The lint failure was separate: wrapping the bare www.wikipedia.org as a link made it visible to the spell checker, so it is now in .spelling next to the existing wikipedia.org entry. The rendered page does not change either way, since GFM already autolinked the bare hostname.

PR #17506, opened 2026-08-03
docs: document ztunnel response_flags values for ambient L4 metrics

### LINKED ISSUE #17488: Docs: ztunnel's response_flags metric values are undocumented (metrics reference points only to Envoy's flags)
### What's the issue?

The [Standard Metrics reference](https://istio.io/latest/docs/reference/config/metrics/) documents the `response_flags` label only for Envoy:

> Response Flags (`response_flags`): Additional details about the response or connection from proxy. In case of Envoy, see `%RESPONSE_FLAGS%` in Envoy Access Log.

In ambient mode, ztunnel (Rust, not Envoy) emits its own distinct set of `response_flags` values on the L4 TCP metrics such as `istio_tcp_connections_closed_total`. `response_flags` is an exported metric label that operators scrape and alert on, so these are user-facing, observable values rather than internal details. They are defined in [`src/proxy/metrics.rs`](https://github.com/istio/ztunnel/blob/master/src/proxy/metrics.rs), and none of them are Envoy flags:

| value | meaning |
| --- | --- |
| `-` | none |
| `DENY` | authorization policy denied |
| `CONNECT` | connection failure |
| `TLS_FAILURE` | TLS failure |
| `H2_HANDSHAKE_FAILURE` | HTTP/2 handshake failure |
| `NETWORK_POLICY` | network policy error |
| `IDENTITY_ERROR` | identity error |

None of these values appear anywhere in the current docs, and Envoy's `%RESPONSE_FLAGS%` list (which the docs point to) does not contain them, so there is no way to interpret ztunnel's L4 metrics from the documentation. This matters for diagnosing ambient incidents from metrics: a strict-mTLS denial, for example, surfaces only as `istio_tcp_connections_closed_total{response_flags="DENY"}` and appears in no L7 request metric, but nothing in the docs indicates that.

### Proposed fix

At minimum, a note that ztunnel emits its own `response_flags` values distinct from Envoy's, with a pointer to where they are defined. Ideally, a short table like the one above, alongside the Envoy note in the Standard Metrics reference or in the [ztunnel troubleshooting guide](https://istio.io/latest/docs/ambient/usage/troubleshoot-ztunnel/).

Would an addition like this be welcome? I am happy to open the PR.

kassvl:
Fixes #17488

The [Standard Metrics reference](https://istio.io/latest/docs/reference/config/metrics/) documents the `response_flags` label only in terms of Envoy's `%RESPONSE_FLAGS%`. In ambient mode the L4 TCP metrics are reported by ztunnel, which is not based on Envoy and emits its own distinct set of values, so there was no way to interpret them from the documentation.

**What this changes**

- **Metrics reference:** adds the set of values ztunnel emits, alongside the existing Envoy note, plus a sentence explaining that an L4 denial shows up only in the TCP metrics and in no request-level metric.
- **ztunnel troubleshooting guide:** adds a short "Diagnosing failed connections from metrics" section showing how to use the flag to tell an authorization denial apart from a transport failure.

**Source of the values**

Taken from `EncodeLabelValue for ResponseFlags` in [`src/proxy/metrics.rs`](https://github.com/istio/ztunnel/blob/master/src/proxy/metrics.rs), verified against ztunnel master rather than copied from the issue: the issue I filed listed seven values, but master now also emits `CERT_REVOKED`, which was added with CRL support, so the table here has eight. The label is carried by `CommonTrafficLabels`, which is used by `tcp_connections_opened`, `tcp_connections_closed`, `tcp_received_bytes` and `tcp_sent_bytes`.

I deliberately left `tcp_connections_failed` out: it is registered as unstable in ztunnel, so it did not seem appropriate for the reference page.

Happy to move the table to a different page, trim it to a pointer, or adjust the wording if you would prefer a different shape.


Arhell (commented):
/ok-to-test

PR #17507, opened 2026-08-03
Add blog post: Istio Gives You Progressive Delivery. Who Gives It to Istio?

1991santhu:
## Description

This PR adds a new blog post: **"Istio Gives You Progressive Delivery. Who Gives It to Istio?"**

Istio makes it straightforward to roll application changes out gradually — shift a percentage of traffic, watch, promote or roll back. Changes to the mesh's *own* configuration get none of that. A `VirtualService` or `DestinationRule` applies mesh-wide as soon as istiod accepts it: no percentage, no bake time, no automatic revert. The resource that expresses a careful rollout is itself rolled out in one shot.

The post argues that config pipelines validate whether a change is well-formed and whether it is permitted, but never whether its *effect* is plausible — and that this is why mesh-config incidents so often present as a healthy system doing the wrong thing rather than as an outage.

It then walks a four-rung maturity ladder (manual → health-gated → outcome-gated → autonomous) in Istio terms, and closes with four things a reader can do today with no new features:

- diff the delta rather than the document, and separate "is this change large?" from "was a large change expected?"
- give mesh config its own canary namespace
- watch `istio_requests_total` by `destination_workload` / `destination_version` rather than error rate, since a stale subset produces healthy proxies and flat error rates
- treat config rollback as a rehearsed path

It also notes the same gap in OpenTelemetry's OpAMP protocol and in Flux, linking the live discussions, to frame it as an ecosystem problem rather than an Istio-specific criticism.

Written from experience building configuration-safety and progressive-delivery systems for large-scale infrastructure. No vendor content and no product references.

## Notes

- `.spelling` updated with `Flagger` and `OpAMP`.
- Happy to adjust the framing, trim length, or add diagrams if reviewers would prefer.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -601,6 +601,7 @@ fine-grained
 FIPS
 Firebase
 FitStation
+Flagger
 Fluentd
 foo.yaml
 Forbe
@@ -1074,6 +1075,7 @@ OneCloud
 onramp
 onsite
 onwards
+OpAMP
 OpenAPI
 OpenCensus
 OpenID
--- a/.spelling
+++ b/.spelling
@@ -1266,6 +1266,7 @@ Salmond
 sandboxed
 sandboxing
 sanitization
+Santhosh
 Sathish
 Savcı
 sayin
@@ -1335,6 +1336,7 @@ SoC2
 SolarWinds
 Solo.io
 Solo.io.
+Somarapu
 SongYang
 Specko
 Speedrun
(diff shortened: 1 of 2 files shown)


Arhell (commented):
/ok-to-test

1991santhu:
Pushed some corrections before this takes up reviewer time. One of them reverses a claim in the original draft.

The post cited opamp-spec#384 as an open gap. That issue is now closed. Tigran Najaryan's answer was that the omission is deliberate: the spec covers the protocol, and how you sequence a rollout or when you roll back is left to each implementation. I agreed and closed it myself. Publishing the old wording would have described another CNCF project as having a gap its maintainers had just explicitly rejected. Rewritten to say what they actually said, which I think makes the point better anyway: the signals are in the protocol, what's missing is something that reads them and stops.

The CI sample was worse. `yq` evaluates once per document, so on a multi-document manifest the count came back as several lines, the `[ -lt ]` test errored instead of comparing, and the check exited 0 without printing anything. Same thing for a file that isn't on the base branch yet. A guardrail that fails open is the exact failure the post is about. It now collects the count across documents and handles the missing file explicitly. I tested it in a scratch repo on four cases: unannotated reduction fires, annotated one passes, new file passes, increase passes. My first attempt at the fix aborted with git's exit 128 on the new-file case, which is why that check is spelled out.

Two smaller things. `destination_version` is the workload's version label, not the DestinationRule subset name, so I've said so rather than implying they're the same. And a subset matching no pods was listed as a silent failure when it's the loud one, since no endpoints means 503s straight away. Swapped it for genuinely quiet cases.

Also linked Flagger on first use.

One question for reviewers: the post ends with a short author bio. I can't find another istio.io post that has one, and the `attribution` field already credits the author. Happy to drop that paragraph.

PR #17508, opened 2026-08-03
Pull the KinD node image from registry.istio.io

dhawton:
Every `doc.test.*` job fails during KinD cluster creation:

```
ERROR: failed to create cluster: failed to pull image "gcr.io/istio-testing/kind-node:v1.35.0"
Command Output: error getting credentials - err: exit status 1, out: `You do not currently have an active account selected.`
```

Since istio/test-infra#5991 moved these jobs to the AWS Prow cluster, there is no
active GCP account, so the gcloud credential helper registered for `gcr.io` exits
non-zero and the pull fails rather than falling back to anonymous.

`registry.istio.io/testing/kind-node:v1.35.0` pulls without credentials, and it is
already the default in `common/scripts/kind_provisioner.sh` — this override was
defeating it. Matches istio/istio#59364.

### DIFF
--- a/content/en/docs/tasks/traffic-management/egress/http-proxy/index.md
+++ b/content/en/docs/tasks/traffic-management/egress/http-proxy/index.md
@@ -52,6 +52,8 @@ This example uses [Squid](http://www.squid-cache.org) but you can use any HTTPS 
     http_access allow all
 
     coredump_dir /var/spool/squid
+
+    max_filedescriptors 1024
     EOF
     {{< /text >}}
 
--- a/content/en/docs/tasks/traffic-management/egress/http-proxy/snips.sh
+++ b/content/en/docs/tasks/traffic-management/egress/http-proxy/snips.sh
@@ -38,6 +38,8 @@ http_access deny manager
 http_access allow all
 
 coredump_dir /var/spool/squid
+
+max_filedescriptors 1024
 EOF
 }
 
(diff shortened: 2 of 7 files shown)


Stevenjin8 (approved):

kfaseela (approved):

keithmattix (approved):

dhawton:
depends on https://github.com/istio/common-files/pull/1298 due to an unbound variable issue

dhawton:
/retest

test flakes

craigbox (approved):
LGTM to unblock but https://istio.io/latest/blog/2026/retirement-of-gcr.io-follow-up/#what-to-do-today says this will break later this year also

dhawton:
> LGTM to unblock but https://istio.io/latest/blog/2026/retirement-of-gcr.io-follow-up/#what-to-do-today says this will break later this year also

This is to unblock master and mirror what we do in istio/istio. https://hub.docker.com/u/istio -- we do not push our modified kind images there.

Stevenjin8:
@craigbox @dhawton I meant that more for registry.istio.io/**release**. I think we can still keep registry.istio.io/testing.

dhawton:
/retest

Pod deleted unexpectedly.. Prow is struggling.

Stevenjin8:
@dhawton this seems like a legit failure. Let me investigate

fjglira (approved):

PR #17520, opened 2026-08-10
added sections on requiring traffic to traverse a waypoint

jgkawell:
## Description

Added sections on requiring traffic to traverse a Waypoint using AuthorizationPolicies. This was discussed during the July 29th WG call.

## Reviewers

- [x] Ambient
- [x] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/usage/l7-features/index.md
+++ b/content/en/docs/ambient/usage/l7-features/index.md
@@ -52,7 +52,7 @@ In ambient mode, authorization policies can either be *targeted* (for ztunnel en
 
 The ztunnel cannot enforce L7 policies. If a policy with rules matching L7 attributes is targeted with a workload selector (rather than attached with a `targetRef`), such that it is enforced by a ztunnel, it will fail safe by becoming a `DENY` policy.
 
-A policy attached to a waypoint is only enforced for traffic that actually reaches the waypoint. If a service or namespace is labeled with `istio.io/use-waypoint` but the waypoint does not exist or is not ready, ztunnel sends traffic straight to the destination and the waypoint's Layer 7 policies are not applied. To require that traffic traverses the waypoint, pair the waypoint policy with an `AuthorizationPolicy` enforced by ztunnel that admits only the waypoint's identity. See [Require traffic to traverse the waypoint](/docs/ambient/usage/waypoint/#require-waypoint).
+A policy attached to a waypoint is only enforced for traffic that actually reaches the waypoint. Traffic can bypass the waypoint, and its Layer 7 policies, when the waypoint does not exist or is not ready, or when a client addresses a workload directly rather than the service that a default (service-scoped) waypoint fronts. To require that traffic traverses the waypoint, pair the waypoint policy with an `AuthorizationPolicy` enforced by ztunnel that admits only the waypoint's identity. See [Require traffic to traverse the waypoint](/docs/ambient/usage/waypoint/#require-waypoint).
 
 See [the L4 policy guide](/docs/ambient/usage/l4-policy/) for more information, including when to attach policies to waypoints for TCP-only use cases.
 
--- a/content/en/docs/ambient/usage/l7-features/index.md
+++ b/content/en/docs/ambient/usage/l7-features/index.md
@@ -52,6 +52,8 @@ In ambient mode, authorization policies can either be *targeted* (for ztunnel en
 
 The ztunnel cannot enforce L7 policies. If a policy with rules matching L7 attributes is targeted with a workload selector (rather than attached with a `targetRef`), such that it is enforced by a ztunnel, it will fail safe by becoming a `DENY` policy.
 
+A policy attached to a waypoint is only enforced for traffic that actually reaches the waypoint. If a service or namespace is labeled with `istio.io/use-waypoint` but the waypoint does not exist or is not ready, ztunnel sends traffic straight to the destination and the waypoint's Layer 7 policies are not applied. To require that traffic traverses the waypoint, pair the waypoint policy with an `AuthorizationPolicy` enforced by ztunnel that admits only the waypoint's identity. See [Require traffic to traverse the waypoint](/docs/ambient/usage/waypoint/#require-waypoint).
+
 See [the L4 policy guide](/docs/ambient/usage/l4-policy/) for more information, including when to attach policies to waypoints for TCP-only use cases.
 
 ## Observability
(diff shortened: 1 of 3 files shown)


content/en/docs/ambient/usage/l7-features/index.md
```diff
@@ -52,6 +52,8 @@ In ambient mode, authorization policies can either be *targeted* (for ztunnel en
 
 The ztunnel cannot enforce L7 policies. If a policy with rules matching L7 attributes is targeted with a workload selector (rather than attached with a `targetRef`), such that it is enforced by a ztunnel, it will fail safe by becoming a `DENY` policy.
 
+A policy attached to a waypoint is only enforced for traffic that actually reaches the waypoint. If a service or namespace is labeled with `istio.io/use-waypoint` but the waypoint does not exist or is not ready, ztunnel sends traffic straight to the destination and the waypoint's Layer 7 policies are not applied. To require that traffic traverses the waypoint, pair the waypoint policy with an `AuthorizationPolicy` enforced by ztunnel that admits only the waypoint's identity. See [Require traffic to traverse the waypoint](/docs/ambient/usage/waypoint/#require-waypoint).
```
ilrudie: One common reason for a bypass is traffic type, which isn't mentioned. Normal waypoints handle to-service traffic so any client which dials the workload directly would also by pass a default waypoint (unless protected by L4 auth policy.

krinkinmu (approved):

content/en/docs/ambient/usage/waypoint/index.md
```diff
@@ -213,6 +213,32 @@ The original destination type of the traffic is used to determine if a service o
 For instance, traffic which is addressed to a service, even though ultimately resolved to a pod IP, is always treated by the ambient mesh as to-service and would use a service-attached waypoint.
 {{< /tip >}}
 
+### Require traffic to traverse the waypoint {#require-waypoint}
+
+The `istio.io/use-waypoint` label records your intent to send traffic through a waypoint, but on its own it does not guarantee that this happens. If the named waypoint does not exist or is not ready, ztunnel routes traffic directly to the destination rather than failing the request. Any Layer 7 policy that the waypoint would have enforced never takes effect, and traffic flows as though no waypoint were configured.
+
+If enforcing a waypoint's Layer 7 policies is a security requirement, make the waypoint mandatory with an `AuthorizationPolicy` that admits only the waypoint's identity. A waypoint uses the service account named after its `Gateway`, so a policy on the destination workloads that allows only that identity denies any client that reaches them without first passing through the waypoint. Continuing with the `reviews-svc-waypoint` waypoint from above:
```
ilrudie: allow/deny is common wording, so I would prefer that over admits here.
```suggestion
If enforcing a waypoint's Layer 7 policies is a security requirement, make the waypoint mandatory with an `AuthorizationPolicy` that allows only the waypoint's identity. A waypoint uses the service account named after its `Gateway`, so a policy on the destination workloads that allows only that identity denies any client that reaches them without first passing through the waypoint. Continuing with the `reviews-svc-waypoint` waypoint from above:
```

content/en/docs/ambient/usage/waypoint/index.md
```diff
@@ -213,6 +213,32 @@ The original destination type of the traffic is used to determine if a service o
 For instance, traffic which is addressed to a service, even though ultimately resolved to a pod IP, is always treated by the ambient mesh as to-service and would use a service-attached waypoint.
 {{< /tip >}}
 
+### Require traffic to traverse the waypoint {#require-waypoint}
+
+The `istio.io/use-waypoint` label records your intent to send traffic through a waypoint, but on its own it does not guarantee that this happens. If the named waypoint does not exist or is not ready, ztunnel routes traffic directly to the destination rather than failing the request. Any Layer 7 policy that the waypoint would have enforced never takes effect, and traffic flows as though no waypoint were configured.
```
ilrudie: Same as for l7-features, traffic-type is a commons reason for bypass too

ilrudie (commented):
Thanks for tackling this. I added some notes but over all this looks good

content/en/docs/ambient/usage/l7-features/index.md
jgkawell: Yeah good point. Added wording to that effect here and on the other page

content/en/docs/ambient/usage/l7-features/index.md
```diff
@@ -52,7 +52,7 @@ In ambient mode, authorization policies can either be *targeted* (for ztunnel en
...
 
-A policy attached to a waypoint is only enforced for traffic that actually reaches the waypoint. If a service or namespace is labeled with `istio.io/use-waypoint` but the waypoint does not exist or is not ready, ztunnel sends traffic straight to the destination and the waypoint's Layer 7 policies are not applied. To require that traffic traverses the waypoint, pair the waypoint policy with an `AuthorizationPolicy` enforced by ztunnel that admits only the waypoint's identity. See [Require traffic to traverse the waypoint](/docs/ambient/usage/waypoint/#require-waypoint).
+A policy attached to a waypoint is only enforced for traffic that actually reaches the waypoint. Traffic can bypass the waypoint, and its Layer 7 policies, when the waypoint does not exist or is not ready, or when a client addresses a workload directly rather than the service that a default (service-scoped) waypoint fronts. To require that traffic traverses the waypoint, pair the waypoint policy with an `AuthorizationPolicy` enforced by ztunnel that admits only the waypoint's identity. See [Require traffic to traverse the waypoint](/docs/ambient/usage/waypoint/#require-waypoint).
```
ilrudie: While this isn't wrong per se, it's may be over specific to describe the defaults. I think you could just say something like "mismatched traffic-type" and maybe include a link to discussion of traffic types.

ilrudie: On second reading, "is not ready" should probably be changed to something like "has no address". It doesn't matter if there are zero ready pods once the address has been assigned we will try to send traffic to the waypoint.

jgkawell: Yeah, simplified it a bit

jgkawell: Yeah good point. Changed the wording a bit

ilrudie (approved):

dhawton (approved):

PR #17579, opened 2026-09-01
clarify sidecar+ambient multi-cluster ingress

ilrudie:
## Description

https://github.com/istio/istio/issues/60609 reported some confusion over running sidecar and ambient multi-network side by side with regards to how ingress gateways send traffic to remote networks. I think the behavior reported is expected, but isn't very intuitive. Trying to make this more clearly documented

## Reviewers

- [X] Ambient
- [ ] Docs
- [ ] Installation
- [ ] Networking
- [ ] Security
- [ ] Test and Release
- [ ] User Experience

### DIFF
--- a/content/en/docs/ambient/install/multicluster/multi-primary_multi-network/index.md
+++ b/content/en/docs/ambient/install/multicluster/multi-primary_multi-network/index.md
@@ -23,6 +23,21 @@ across cluster boundaries.
 Before proceeding, be sure to complete the steps under
 [before you begin](/docs/ambient/install/multicluster/before-you-begin).
 
+{{< warning >}}
+In a mesh where ambient and sidecar multi-network coexist, enabling ambient
+multi-network changes how an ingress gateway reaches workloads on another
+network. Even when the destination workload uses sidecar mode, ingress traffic
+uses double HBONE through the ambient east-west gateway on port `15008`, rather
+than the classic east-west gateway on port `15443`. The destination sidecar must
+therefore have inbound HBONE support enabled with
+`ISTIO_META_ENABLE_HBONE=true` in its proxy metadata. This setting is enabled by
+default in `MeshConfig` when using the ambient profile, but existing sidecar
+workloads must be restarted to pick it up. Sidecar-to-sidecar traffic can
+continue to use the classic east-west gateway. See the
+[migration guide](/docs/ambient/migrate/install-ambient-components/)
+for restart and verification instructions.
+{{< /warning >}}
+
 {{< boilerplate multi-cluster-with-metallb >}}
 
 In this configuration, both `cluster1` and `cluster2` observe the API Servers


keithmattix:
@krinkinmu @grnmeira doss this reflect the current state? I think so but douvle check me

krinkinmu:
I believe that it can work, though I haven't tested it personally yet. I'm not opposed to the PR in principle if it was tested, but if not - give me a day or two to get it tested and I will get back to you.

Non blocking for this PR, but there is a broader question of how we want it to work ideally. The problem with ingress specifically is that it's not exactly ambient or sidecar and it has to handle multiple use cases, but maybe we can somehow figure out based on the destination endpoint if it's sidecar or ambient?

@grnmeira WGYT?

ilrudie:
Actually, we don't explicitly integration test sidecar endpoints in the mc ambient suite I don't think. Closest is remote waypoints.

krinkinmu:
I'm trying to reproduce the issue but failing at the moment, basically in my case I don't get any endpoints from the remote cluster at all. I'm sure that it's because I'm doing something wrong, but it will take me sometime to reproduce.

ilrudie:
I have a rough integration test locally. I think there is a real issue. dhbone to the e/w gateway is ok and the inner arrives at the sidecar, addressed to the service, but the sidecar doesn't have a service match on main_internal.

krinkinmu:
Hmm... by any chance could it be that we don't add matches based on the service name and only add matches based on the IP? We can probably get rid of all the IP matches for services all together when it comes to HBONE now.

ilrudie:
https://github.com/istio/istio/compare/ambient-sidecar-multi-network-testing?expand=1, if you want to take a look

ilrudie:
I think there isn't service-oriented chain/match in the dst sidecar at all.

| Chain | `prefix_ranges` | `destination_port` | Matches? |
|---|---|---|---|
| `0.0.0.0_18080` (HTTP workload port) | `10.30.0.52/32` | `18080` | ❌ - workload addressed on wl port |
| `virtualInbound-catchall-http` | `10.30.0.52/32` | *any* |  ❌ - workload addressed |
| `virtualInbound` (catch-all) | `10.30.0.52/32` | *any* | ❌ - workload addressed |

krinkinmu:
I looked into it a bit more and here is my current understanding...

When we cross network boundary with HBONE, we always put service name in the HBONE CONNECT authority (we never put service VIP or specific pod IP there, mostly because we don't typically know those).

Sidecars with HBONE enabled, when they search for an appropriate filter chain match based on the original destination IP address; the way we generate configuration currently the IP address must be the address of the pod (e.g., it cannot be service VIP even at the moment)
  - we are currently use filter_chain_match API in the Envoy configuration, and that API is quite limited in terms of what it could match, so it basically only works with IP addresses
  - we could switch to the Matcher API instead, which is more flexible, but even if we do that, we still have to figure out all the services the pod is part of to generate a match that would allow HBONE connection from another network.

I think it makes sense to pursue this path in general, to allow ambient workloads talk to sidecar workloads in principle. Though such communication is a bit more complicated than just enabling HBONE in sidecar (e.g. consider how L7 policies should be applied in this case?).


For the problem in https://github.com/istio/istio/issues/60609 specifically though, I think we should consider an alternative way of dealing with it. Currently, when we generate EDS for ingress and waypoints, when we decide whether to use double-HBONE or not we only look at the type of proxy and enabled features. E.g., if it's not a sidecar and ambient multi-network is enabled we use double-HBONE and treat everything as ambient.

Naturally that creates a problem for ingress, because it's not a sidecar and it could route to both ambient and sidecar endpoints.

With that in mind, I think we should also consider the type of the endpoint when we generate EDS. We currently do not track in IstioEndpoint structure if the endpoint is behind sidecar or not, but I'd imagine when we create an IstioEndpoint from the pod, we could take a look at the labels and annotations and tell if the sidecar is injected or not.

Once we have that, when we generate EDS we could do the following decision tree:

If source is sidecar or ingress:
    Yes: check if destination is sidecar
          Yes: connect to it as if it was a sidecar
          No: check if endpoint supports HTTP tunnel
                Yes: connect to it as if it was ambient
                No: connect to it as if it was a sidecar
    No: check if the source is waypoint
          Yes: connect to the destination as if it was a sidecar
          No (this is some unknown type of proxy): check if the destination supports HTTP tunnel
                 Yes: connect to it as if it was ambient
                  No: connect to it as if it was sidecar

To summarize, the rules are:
- if source is an ambient proxy (waypoint) - connect to destination as if it was an ambient endpoint
- if source is sidecar or ingress and destination is sidecar - connect to destination as if it was sidecar
- In remaining cases, check if the destination supports HTTP tunnel and if so, connect to it as if it was ambient, otherwise treat it as sidecar.


Compared to changing the sidecar configuration, I think this option offers a nice property - traffic going between sidecars or between ingress and a sidecar stays undisturbed when ambient is enabled. I think this property is quite useful if we want folks currently using sidecar to try out ambient safely and potentially migrate to ambient in the future.



I'm going to work on the second option (I need to figure out how we can conclusively tell if the endpoint has a sidecar or not, but after that, I think it's not really difficult), but let me know what you think about it.

krinkinmu:
Sorry, the decision tree all collapsed in the md, here is what I wanted to write there actually:

If source is sidecar or ingress:
  * Yes: check if destination is sidecar
     * Yes: connect to it as if it was a sidecar
     * No: check if endpoint supports HTTP tunnel
       * Yes: connect to it as if it was ambient
       * No: connect to it as if it was a sidecar
   * No: check if the source is waypoint
     * Yes: connect to the destination as if it was a sidecar
     * No (this is some unknown type of proxy): check if the destination supports HTTP tunnel
        * Yes: connect to it as if it was ambient
        * No: connect to it as if it was sidecar

krinkinmu:
Here I drafter a proof of concept of a sort for this: https://github.com/istio/istio/pull/61548, PTAL.

ilrudie:
So, we should not merge this given the test findings. Should we add a warning that it's not supported until work on the use case concludes?

krinkinmu:
I was sure that we call it out already somewhere, but I cannot find it, so yes, let's add a warning there. We have a section with known limitations in content/en/docs/ambient/install/multicluster/_index.md, maybe that's a good place to call it out.

krinkinmu:
I think https://github.com/istio/istio/pull/61548 is ready for a review now. I still need to put it through the whole test suite, but it fixes the ingress gateway issue in the test.

PR #17638, opened 2026-09-22
update support k8s versions

zirain:
@jewertow can you confirmed this?


jewertow:
Adding DNM, because I have to migrate kubeadm config to v1beta4 to fix integration tests.

